PR #64, opened 2012-04-29
Update World.pmc to use Parrot's PackfileView PMC

Whiteknight:
this works in current master (after a similar pull request is merged for NQP) and will future-proof rakudo against the upcoming merge of the eval_pmc branch in Parrot.


labster:
Closing this ticket because, 3 years later, Parrot support is on hiatus.  At this point, merging this PR would require re-enabling Parrot, which would take a great deal of work.

PR #71, opened 2012-06-24
IO Buffering

ajs:
I added two methods for doing buffering of IO handles and turned on line buffering by default on new handles.


labster:
Because this PR reaches down into Parrot internals, and Parrot support has been suspended, I'm closing this pull request.  If Parrot support returns in the future, this should probably be implemented with nqp op(s).

ajs:
I gave up on this one a long time ago, and am happy to see it closed. Real IO support in Perl6 is going to be a major undertaking, and I don't think that it behooves the language to take it on piecemeal.

PR #106, opened 2013-02-22
Switch to new packfile API and build using --target=pbc

gerdr:
This is a cleaned up version of #102. Needs to be applied together with the corresponding changes to Parrot and NQP.


lizmat:

PR #107, opened 2013-03-01
get() stalls when called from HTTP::Easy

drKreso:
I am not sure if this makes any sense at all. I did run the full spec test suite. It did not break any more than usual. IO tests all passed.

This makes HTTP::Easy work as advertised (and consequently Bailador).


labster:
I'm going to tag this ticket as Parrot specific, because it seems to be related to the readline bugs noted above.  And there's not enough information to otherwise reproduce the bug here.  Let me know if this affects other VMs.

lizmat:

PR #170, opened 2013-06-21
Fix .clone() when providing new array/hash attributes

samuraisam:
This fixes https://rt.perl.org/rt3/Ticket/Display.html?id=118559, passing spectests.

~~It accomplishes this by cloning hash and array attributes, identified by their sigil, when deciding whether to clone attributes.~~

It accomplishes this by ensuring that new lists are assigned to the new object rather than updating that objects list, which points at the list of the original object.


lizmat:
Can this pull request be closed now?

zhuomingliang:
I think so, since it already fixed.

teodozjan:
I kindly remind that Bug#118559 is marked as closed :)

coke:
Based on comments from 2015, closing this PR. Thanks.

PR #171, opened 2013-06-23
make semilists in postcircumfix [] work (a bit)

timo:
These two commits make the following Todos pass:

t/spec/S02-types/multi_dimensional_array.rakudo.parrot        (Wstat: 0 Tests: 41 Failed: 0)
  TODO passed:   16-18, 27-29, 36-38

Otherwise, they do not regress.

Something that is missing is assigning to a slice from a list, as in:

    my @a = (1, 2, 3), (4, 5, 6);
    my @a[0, 1;1] = <foo bar>;
    # a is supposed to be (1, foo, 3), (4, bar, 6) now

I'd like some assistance with that.


timo:
superceded by work by someone, ISTR it was Mouq

PR #180, opened 2013-07-17
Implement nodal operators/methods, mark Any, Iteratable and Parcel as nodal.

timo:
This gives us an "is nodal" for classes as well as for Routines, turns $!rw for Routine into $!flags that keeps "is nodal" in the second and "is rw" in the first bit.

I'm not sure if marking Parcel as nodal is sensible and wether or not marking Hash or Associative or more different things as nodal would be necessary.

There's a few tests for this in the perl6/roast in the nodal_tests branch. I'd like to hear more suggestions for cases that ought to be tested.


timo:
superceded by work (and maybe spec changes?) done by TimToady

PR #184, opened 2013-07-19
properly remove left margin from raw blocks

timo:
delimited_raw, abbreviated_raw and paragraph_raw will now have all spaces from the beginning removed from all lines.

```
=begin pod
    =begin code
    foo
    bar
    =end code
=end pod
```

used to result in 4 spaces left of "foo" and "bar" and no longer does.


AlexDaniel:
Can has updates? 4 years no love.

JJ:
Ping?

JJ:
Since there are so many conflicts now, can this maybe be closed?

PR #232, opened 2014-01-06
Address a variety of issues brought up by jnthn

Arcterus:


jnthn:
Thanks for submitting this. In the end, fixed size and multi-dim arrays ended up being done with VM-level support, which was completed last year prior to the 6.c release.

PR #236, opened 2014-01-18
give the user a way to disable perl5 warnings (only one so far)

timo:
when writing 

```
no warnings :p5;
```

you can get rid of the warning that numbers starting with 0 are not parsed as octals.

I'd like to get some feedback; Especially what else to change. Maybe I should have a look at the P6Regex grammar/actions for more things?

### DIFF
--- a/src/Perl6/Grammar.nqp
+++ b/src/Perl6/Grammar.nqp
@@ -1290,6 +1290,8 @@ grammar Perl6::Grammar is HLL::Grammar does STD {
                     $*P5_WARNINGS := 0;
                     $*GENERAL_WARNINGS := 0;
                 }
+            } else {
+                $/.CURSOR.panic("'no " ~ $<module_name>.Str ~ "' not implemented");
             }
         }
         <.ws>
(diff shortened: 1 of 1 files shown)


lizmat:
Now that we have lexical pragma's, this can be done easier.  One approach to this was taken by cf65f5463fc12043b00

AlexDaniel:
Hey @zoffixznet, what do you think about this? Maybe we should start doing this more, and then it'll be easy to flip the default so that you'd need to `use warnings :p5` if you're coming from perl 5.

zoffixznet:
> Hey @zoffixznet, what do you think about this? 

Most of the things are errors not warnings and include langs other than Perl 5.

As I've said before, these don't belong in core and should be ecosystem slang modules designed for converts from specific languages. Then they can be both more targeted and more extensive.

AlexDaniel:
Actually, we do have something for that nowe. @lizmat, what do you think?

timo:
i would say the "use isms" mechanism implements the idea behind this PR.

PR #254, opened 2014-03-24
Fix for changed parrot 5.0 libsearch path: Add . to the front

rurban:
Required if older rakudo or parrot libs are already installed.

Pull #93 went too old.


lizmat:

PR #267, opened 2014-05-08
Fix config for sometests target. add j-sometests

dwarring:
A couple of Makefile generation tweaks:
`j-sometests` target has been added to run JVM specific tests.
`sometests` target now run tests on all configured backends, e.g.:
 `% make sometests TESTFILES=t/spec/integration/advent2012-day*.t`


dwarring:
Makefile generation seems to have been fixed for jvm

PR #306, opened 2014-09-08
Enable symmetric difference for LoLs

ab5tract:
A stab at LoL-ifying the symmetric difference set operator. I implemented it also for Bag, because it had the base operators that symmetric difference is based on. That doesn't mean it makes any sense to do so.

I'm quite new to all this, so feedback on the approach is much appreciated.

### DIFF
--- a/src/core/set_operators.pm
+++ b/src/core/set_operators.pm
@@ -108,16 +108,33 @@ only sub infix:<<"\x2216">>(|p) {
     infix:<(-)>(|p);
 }
 
-proto sub infix:<(^)>($, $ --> Setty) {*}
-multi sub infix:<(^)>(Any $a, Any $b --> Setty) {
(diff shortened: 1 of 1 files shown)


ab5tract:
This has been superseded by commit 3f59306fe3f.

I've pushed a potential patch for a QuantHash version of symmetric difference: https://github.com/rakudo/rakudo/pull/362

AlexDaniel:
Am I getting it right that this pull request can be closed in favor of https://github.com/rakudo/rakudo/pull/362 ?

ab5tract:
I think @lizmat already addressed all the quant hash / set operator
weirdnesses, but maybe she can chime in on it.

lizmat:
This PR predates the GLR and the re-implementation of set operators.  Closing.

PR #317, opened 2014-10-02
Implement tr

peschwa:
...ok rather, implement a few more features that tr/// and .trans should have.
Those are adverbs for tr/// and .trans as well as closures as replacement. Additionally I have fixed the bug discovered/discussed around [here](http://irclog.perlgeek.de/perl6/2014-09-29#i_9424693).

Remaining TODO in t/spec/S05-transliteration/trans.t is `not ok 23 - ambiguous ranges combined# TODO disambiguate ranges`, which I hereby defer judgement over.


peschwa:
I've merged and fixed this myself now.

PR #324, opened 2014-10-23
Cliparser with grammar

peschwa:


peschwa:
Rebased onto nom as of today, to clean this PR up a little.

moritz:
I'm not at all happy with the approach of joining the command line arguments with a zero-byte and then re-parsing them. Even if we can ensure that all our platforms disallow zero-bytes within the command line arguments (and I have no idea if maybe windows has an alternative API that allows it), it still feels very wrong.

Maybe this is the road to world domination, but I don't see it yet.

I'd welcome comments from other Rakudo developers.

timo:
i'd like to see this in rakudo. we can use a private use unicode character instead of a nullbyte to get around that particular piece of worry.

AlexDaniel:
> private use unicode character instead of a nullbyte

I don't think it's better.

JJ:
This is way outdated, and in conflict with a good bit of the codebase. Also, the approach has raised a certain amount of polemic, and it does not really solve any issue. If conflicts are not solved _and_ there's some degree of support for this I'm closing it in the next few days/weeks.

JJ:
Closed for [this issue in the problem-solving repo](https://github.com/Raku/problem-solving/issues/187)

AlexDaniel:
So I'm looking at this and it feels like it's just not ready yet. At first I didn't see any problem with joining things with a null byte (as indeed you cannot have null bytes in your arguments), but then I saw this:
```
[ '--' <.ws>+ { $*STOPPED := 1 } ]
```
So if you have empty arguments it'll just eat right through them.

And then:
```
<.ws>?
```
So we don't even know if you're in between the arguments or not?

We can of course fix it, but the whole idea of squishing things together only to reparse them later is indeed an anti-pattern of some sort.

Why does it have to be a grammar that parses the whole thing at once? Just don't do it, it'll be simpler and less buggy.

lizmat:
I think this PR can be closed.  Too much has changed since, looks like.  If nobody objects, I will close this PR in the coming days.

coke:
Closing based on last several comments. Thank you for the effort

PR #330, opened 2014-11-16
Add perl6-b binary target

bdw:
I've wanted a binary runner for perl6 for a long time,
starting with the need to sometimes run the thing under
gdb. If we want to be fully 'production ready', I think
this should be a necessity too.

Note that this does not currently make a 'fake-excutable'
(like pbc2exe) does, rather relying on the installation
of moarvm files in their respective directories. Since we're
likely to want a 'static' perl6 binary eventually, this should
be investigated.

Also note that it uses hardcoded moarvm include dependencies,
which isn't really necessary since nqp --show-config can
provide us with this information.

Note that this only creates the target, but doesn't do anything
to the (regular) build yet.


moritz:
If we want to merge this, please rename it to `perl6-m`. We use those postfixes to denote the VM we run under (-m for moarVM, -p for Parrot, -j for the JVM), so adding -b for binary would be very confusing. Instead the current perl6-m shell script wrapper should be abolished.

PR #333, opened 2014-11-25
Change Range excludes_ methods to dashes

lucasbuchala:
Please review. Thank you.


moritz:
I think that extra constructor would have lots of potential for ambiguous dispatch (if no named arguments are provided).

moritz:
Merged in 1f3bb95f28a560f514094f1b5d7b9e587e1a4bd7 (and migrated the new `infix:<eqv>(Range, Range)` too)

PR #335, opened 2014-11-30
Coerce set comparator operands 'up'

ab5tract:
This patch presents the position that comparator ops could be more
DWIM-y by always coercing their arguments up the chain of complexity.

The current behavior is to coerce Setty things to Sets. This puts Bags
and Mixes in an awkward situation, as .values become True and so only
.keys comparisons matter. The approach of having (<+) and the like be
Baggy specific versions of the operator smells overcomplicated to me.

This approach preserves the original behavior by asking our dev to
be explicit about any "downgrading" of any .values inasmuch as they are
present in the operation:

```
$set (<) $bag.Set
```

This demand to a more explicit statement is the only perceived
tradeoff*, the reward of which is that one does not need to think twice
about:

```
my $reference-bag = bag <1 1 1 2 2 3>;
my $random-mix = mix <1 2 2 3 3 3>;

# Existing syntax:
# Bag --> Set (<=) Mix --> Set
> $reference-bag (<=) $random-mix
True

# Coercion of least surprise:
# Bag --> Mix (<=) Mix
> $reference-bag (<=) $random-mix
False
```

There is another option to implement this, using multi methods and
signatures. However, this implementation is way DRY-er, and thus a
better example of code (which I take it the core settings are intended
to be.. and that's judging mainly from the looks of things!).

Notes:

*Which is also arguably a better demand to make than to ask our dev to
remember a separate-but-strange (<+))

### DIFF
--- a/src/core/set_operators.pm
+++ b/src/core/set_operators.pm
@@ -123,13 +123,22 @@ only sub infix:<<"\x2296">>($a, $b --> Setty) {
 #     $a == $b and so $a.keys.all (elem) $b
 # }
 
-proto sub infix:<<(<=)>>($, $ --> Bool) {*}
-multi sub infix:<<(<=)>>(Any $a, Any $b --> Bool) {
-    $a.Set(:view) (<=) $b.Set(:view);
+### Set Comparison Operators
+#
+# Note that these are implemented according to subroutine signature
+# combinatorics because of the performance implications of junctions
+
+only sub infix:<<(<=)>>(Any $a, Any $b --> Bool) {
+    my ($c,$d);
+    if      $a ~~ Mix or $b ~~ Mix { $c = $a.Mix(:view); $d = $b.Mix(:view); }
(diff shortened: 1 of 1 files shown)


timo:
i believe we have something like this in rakudo now, though i forgot who exactly implemented what we have now. @lizmat, i think you touched that code a bunch?

lizmat:
All of the set operators have been almost completely reimplemented since then.  They now occupy several source files, with prefix `/src/core/set_`, each for a logic set: such as `(elem)` and `(cont)`.

The principle is that neither side of a set operator needs to be a `QuantHash`, but that the operator will coerce the operands under the hood if that is necessary.  Which for instance in the case of `42 (elem) ^100` it isn't.

Furthermore, bag and mix are contagious: as soon as one of the parameters of a set operator is a `Bag`, then Bag semantics will be used.  As soon as one of the parameters is a `Mix`, then `Mix` semantics will be used.

Closing this now.

PR #351, opened 2015-01-11
Get Mixes working properly with the QuantHash ops

ab5tract:
Up until now, mixes would get baggified, leading to some serious anti-DWIM.

This PR was suggested by @colomon in another of my Set-related PRs as a necessary first step: https://github.com/rakudo/rakudo/pull/335

See https://github.com/perl6/roast/pull/52 for the relevant spec tests.

### DIFF
--- a/src/core/set_operators.pm
+++ b/src/core/set_operators.pm
@@ -230,7 +230,7 @@ only sub infix:<(.)>(**@p) {
               for $mixhash.keys;
         }
         $mixhash.Mix(:view);
-    } elsif @p.grep(Baggy) {
+    } else {  # go Baggy by default
         my $baghash = nqp::istype(@p[0], BagHash)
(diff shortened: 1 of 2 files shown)


lizmat:
Please don't use a .grep to find out if any of the elements is a Mixy: I would suggest using .first instead, which will short-circuit as soon as the first Mixy is found.  Either that or doing a "last" in the grep on the first match.

lizmat:
From a performance point of view, you probably want to use nqp::istype() (like in other parts of that code) rather than ~~ , e.g. in the "if $a ~~ Mixy && $b ~~ Mixy"

ab5tract:
Done :D

ab5tract:
Which checks has it failed? It all looked green when I checked "Details" ...

PR #357, opened 2015-01-26
Add git reference option into Configure.pl

atroxaper:
Now we can use --git-reference=<path> option to tell Configure.pl do not clone corresponding repo from the web but use the reference to existing local repo. It reduces time for configuring and increases disk space.

This pull request depends on that https://github.com/perl6/nqp/pull/220

### DIFF
--- a/Configure.pl
+++ b/Configure.pl
@@ -39,7 +39,7 @@ MAIN: {
                'gen-parrot:s', 'parrot-option=s@',
                'parrot-make-option=s@',
                'make-install!', 'makefile-timing!',
-               'git-depth=s',
+               'git-depth=s', 'git-reference=s',
     ) or do {
         print_help();
         exit(1);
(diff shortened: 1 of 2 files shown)


Configure.pl
```diff
@@ -371,7 +371,11 @@ sub print_help {
     --git-protocol={ssh,https,git}
                        Protocol used for cloning git repos
     --git-depth=<number>
-                       Use the --depth option for git clone with parameter number
+                       Use the --git-depth option for git clone with parameter number
+    --git-reference=<path>
+                       Use --git-reference option to identify local path where git repositories are stored
+                       For example: --git-reference=/home/user/repo/for_perl6
+                       Folders 'nqp', 'moar', 'parrot' with corresponding git repos should be in for_perl6 folder
```
mj41: s/moar//MoarVM/ # see line 489

PR #362, opened 2015-02-10
Symmetric difference as a quanthash-y op

ab5tract:
# (^+) -- symmetric difference for quant hashes

I'm still fairly convinced that it makes sense to unify set and quanthash
operators, but this patch is designed to open discussion about 'symmetric
difference' behavior for quanthashes as a separate issue from that. I've
thought about it quite a bit, maybe too much.. Either way, I'm not as sure about it
as when I started.

Outside of LoL context, (^+) makes a fair amount of sense -- give back what was
left after removing any intersecting values:

```
> ( x => 1, y => 2, <z> ).Bag (^+) ( <w>, x => 2, y => 4 ).Bag
Bag.new(w,x(1),y(2),z)
```

The implementation uses the absolute value of the subtraction of the
intersecting keys. This allows for symmetry, but because subtraction is not
commutative any claim to symmetry is broken by the order of LoLs.

It could also be argued that the absolute value symmetry is flawed to begin
with, and the ordering should matter in both (^+) and [(^+)] forms.

I'm not as convinced anymore of how important, let alone correct, either approach is,
but I wanted to publish it to see how others might feel. I haven't found spec
information on set operators, but I remember looking when I first started on
this months ago.

_(Also, I wanted to note that since starting this work, lizmat++ has contributed a
pretty freaking sweet patch to add LoL to (^). It makes me wonder what might be 
possible here, but "implement first, apologize second, golf third" seems like a  better
approach for the current tuits.)_
# CAUTION: NOT WORKING

However, this branch isn't just about that. This works from the repl (but crashes from a file, see the gist linked from the bottom):

```
(rakudo/quant-sym-diff) $ ./perl6
> my $b = ( x => 1, y => 2, <z> ).Bag
bag(x, y(2), z)
> my $b2 = ( <w>, x => 2, y => 4 ).Bag
bag(w, x(2), y(4))
> $b (^+) $b2
bag(x, y(2), z, w)
```

But this doesn't.

```
(rakudo/quant-sym-diff) $ ./perl6
> ( x => 1, y => 2, <z> ).Bag (^+) ( <w>, x => 2, y => 4 ).Bag
===SORRY!=== Error while compiling <unknown file>
Two terms in a row
at <unknown file>:1
------> ( x => 1, y => 2, <z> ).Bag ⏏(^+) ( <w>, x => 2, y => 4 ).Bag
    expecting any of:
        infix stopper
        infix or meta-infix
        statement end
        statement modifier
        statement modifier loop
```

This gist contains a test file which will fail with the same issue: https://gist.github.com/ab5tract/572342901a4561c53dbc

### DIFF
--- a/src/core/set_operators.pm
+++ b/src/core/set_operators.pm
@@ -144,6 +144,41 @@ only sub infix:<<"\x2296">>($a, $b --> Setty) {
     $a (^) $b;
 }
 
+
+only sub infix:<<(^+)>>(**@p) {
+    if @p.first(Mixy) {
+        my $mixhash = nqp::istype(@p[0],MixHash)
+                            ?? MixHash.new-from-pairs(@p.shift.pairs)
+                            !! @p.shift.MixHash;
+        for @p.map(*.Mix(:view)) -> $mix {
+            my $i = $mixhash (|) $mix;
+            for $i.keys -> $k {
+                $mixhash{$k}  = $mixhash{$k}
+                                  ?? abs($mixhash{$k} - $mix{$k})
+                                  !! $mix{$k};
+            }
+        }
+        $mixhash.Mix(:view);
+    } else {
+        my $baghash = nqp::istype(@p[0],BagHash)
+                ?? BagHash.new-from-pairs(@p.shift.pairs)
+                !! @p.shift.BagHash;
+        for @p.map(*.Bag(:view)) -> $bag {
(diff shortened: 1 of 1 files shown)


lizmat:
I think the points of this PR have been addressed by my work on set operators.  Please correct me if you think I'm wrong  :-)

PR #363, opened 2015-02-15
Die in skip_rest if no plan is set

paultcochrane:
Current behaviour dies with "use of uninitialized value of type Any in
numeric context", which is confusing to the user of the Test module.  This
change tells the user what is required in order to use skip_rest correctly.


FROGGS:
There is a typo in the error message, it should be skip_rest, not skip_test.

paultcochrane:
Sorry about that!  Corrected in db560a0.

PR #370, opened 2015-02-21
Main single positional

peschwa:
This PR brings handling of arguments to the programm more in line with what Getopt::Long would do. I.e. a parameter that is declared as Scalar (like in «sub MAIN(:$foo) { ... }») cannot be dispatched to by multiple occurences of a command line switch of the same name.

### DIFF
--- a/src/core/Main.pm
+++ b/src/core/Main.pm
@@ -150,24 +150,37 @@ my sub MAIN_HELPER($retval = 0) is hidden_from_backtrace {
     @matching_candidates .=grep: {!has-unexpected-named-arguments($_.signature, $n)};
     # If there are still some candidates left, try to dispatch to MAIN
     if +@matching_candidates {
-        return $m(|@($p), |%($n));
-    }
-    # if there aren't any, try to see if we can find a candidate that wants a named array param
-    # where we only got one occurence on command line invocation
-    else {
-        my $found = False;
-        for %($n).keys -> $named_arg {
-            for $m.candidates -> $cand {
-                for $cand.signature.params -> $param {
-                    if $param.named_names eq $named_arg && $param.type ~~ Positional {
-                        $found = True;
-                        %($n){$named_arg} = [%($n){$named_arg}];
-                        last;
+        # verify that we don't dispatch multiple occurences of a command line argument
+        # to a parameter that is a Scalar
+        if %($n).grep: { .value ~~ Positional } -> @pairs {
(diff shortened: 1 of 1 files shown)


src/core/Main.pm
```diff
@@ -150,7 +150,37 @@ my sub MAIN_HELPER($retval = 0) is hidden_from_backtrace {
     @matching_candidates .=grep: {!has-unexpected-named-arguments($_.signature, $n)};
     # If there are still some candidates left, try to dispatch to MAIN
     if +@matching_candidates {
-        return $m(|@($p), |%($n));
+        # verify that we don't dispatch multiple occurences of a command line argument
+        # to a parameter that is a Scalar
+        if %($n).grep: { .value ~~ Positional } -> @pairs {
+            for @pairs -> $pair {
+                for @matching_candidates -> $cand {
+                    my @param = $cand.signature.params.grep: { .named_names eq $pair.key };
```
Mouq: `.named_names` returns an array. E.g.:
```
sub foo(:name1(:name2(:$name3))) {}
say &foo.signature.params[0].named_names.perl
```
outputs `Array.new("name3", "name2", "name1")`

src/core/Main.pm
```diff
@@ -150,7 +150,37 @@ my sub MAIN_HELPER($retval = 0) is hidden_from_backtrace {
     @matching_candidates .=grep: {!has-unexpected-named-arguments($_.signature, $n)};
     # If there are still some candidates left, try to dispatch to MAIN
     if +@matching_candidates {
-        return $m(|@($p), |%($n));
+        # verify that we don't dispatch multiple occurences of a command line argument
+        # to a parameter that is a Scalar
```
Mouq: My thought would be that we can still dispatch to them, but should tie-break. So `sub MAIN(:$arg) {}` could still receive `program --arg=1 --arg=2`, but `multi MAIN(:@arg) {}` beats `multi MAIN(:$arg) {}` with that invocation… OTOH, invocations like `-n=1,2` are still NYI, which in my mind is different than `-n=1 -n=2`, and there doesn't seem to be spec on the latter.

peschwa: I'm not fully satisfied with how this PR solves the dispatch, and as you say there's still insufficient speculation on some cases.

peschwa:
Closing this for now.  I don't think the approach is the best, and as Mouq mentions there's a few cases that have to be speculated about first anyway.

PR #381, opened 2015-03-06
Better diagnostics in Test::is for strings differing only in whitespace

niner:
If the expected value differs from the got value only by whitespace at
the end of a line, it's very hard to find this difference. Using .perl
for the diag output on the other hand makes it trivial to find
whitespace differences.

With this patch is("foo \nbar baz", "foo\nbar baz"); gives:
expected: "foo\nbar baz"
     got: "foo \nbar baz"


perlpilot:
Why not output diagnostics like this all the time instead of only when there are whitespace differences?

PR #387, opened 2015-03-17
Do not flatten when SEQUENCE returns an array

usev6:
otherwise test 3 - Pascal's triangle from integration/advent2012-day04.t does not work


usev6:
Alternatively, please revert commit cd51b3401cc1d01b84399413fbcede8a8af9c25e.

PR #388, opened 2015-03-17
Better handling of generator function for sequence

usev6:
fixes example ```'10,9,8, { $_ - 1 || last } ... *'``` from S03

also implement the following from S03: "If the signature of the [generator] function does not match the existing values, the sequence terminates."


usev6:
I'll close this PR for now and instead open another one to clean up the murky flattening for SEQUENCE.

PR #392, opened 2015-03-20
Angle-quoted Rat literals

raydiak:
Implements \<1/2> syntax.  This patch only supports the bare no-space Rat literal and does not attempt to implement val() semantics or alter the behavior of qw-ish operators.  Rough first attempt, feedback welcome.

Spectest clean plus some passing TODOs in t/spec/S32-num/rat.t

http://design.perl6.org/S02.html#Rational_literals
http://irclog.perlgeek.de/perl6/2015-03-20#i_10315220


Mouq:
Instead of `+~$<nu>`, I think `$<nu>.ast` would be better (and probably faster). Same with `$<de>`

raydiak:
Thanks for the tip, I've update the PR

Mouq:
Looks good, thanks :)

PR #394, opened 2015-03-23
Terminate sequence when values do not match signature of generator function

usev6:
S03 states: "A sequence generated from an explicit function places no type constraints on the sequence other than those constraints implied by the signature of the function. If the signature of the function does not match the existing values, the sequence terminates."

Currently the exception X::TypeCheck::Binding (from the violation of type constraint of the generator function) is not handled. The following example dies: ```sub f (Int $n) { $n > 3 ?? "liftoff!" !! $n + 1 }; say 1, &f ... *```


usev6:
As discussed with TimToady++ the current behaviour is expected: http://irclog.perlgeek.de/perl6/2015-04-09#i_10413657

PR #398, opened 2015-03-31
Fix "Cannot look up attributes in a type object" in generated scripts

niner:
The panda script failed with "Cannot look up attributes in a type object
in method files at src/gen/m-CORE.setting:26732". Seems like files()
expects to be called on an object. CompUnitRepo objects are to be found
in @*INC.

### DIFF
--- a/src/core/CompUnitRepo/Local/Installation.pm
+++ b/src/core/CompUnitRepo/Local/Installation.pm
@@ -34,9 +34,10 @@ __END__
 ';
     my $perl_wrapper = '#!/usr/bin/env #perl#
 sub MAIN(:$name, :$auth, :$ver, *@pos, *%named) {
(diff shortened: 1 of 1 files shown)


niner:
@*INC contains CompUnitRepo::Local::File and CompUnitRepo::Local::Installation objects. No idea if that is a problem?

FROGGS:
That approach would break what was fixed in: https://github.com/rakudo/rakudo/commit/a56d8b0ba3472c4d099b694fa325506c3e77808a

FROGGS:
Can you provide information to reproduce the issue?

FROGGS:
Instead of `@*INC».files` we should probably call `@*INC.grep(CompUnitRepo::Local::Installation)».files`

PR #400, opened 2015-04-01
Fix RT #124204 in squish: Do not .Str the return value of &as, in order ...

edwinst:
...to

keep objects distinguishable for the comparison operator used (=== by default).

### DIFF
--- a/src/core/List.pm
+++ b/src/core/List.pm
@@ -503,11 +503,11 @@ my class List does Positional { # declared in BOOTSTRAP
     proto method squish(|) {*}
     multi method squish( :&as!, :&with = &[===] ) {
         my $last = @secret;
-        my str $which;
+        my $as;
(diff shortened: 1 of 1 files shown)


moritz:
Currently this branch doesn't merge cleanly, probably because the function was switched from gather/take to .map for performance reasons.

Could you please rebase it against current nom?

AlexDaniel:
[The ticket](https://rt.perl.org/Ticket/Display.html?id=124204) is marked as resolved. Hey @usev6, can you take a look?

usev6:
Yes, this PR is no longer needed. Commit 24931c4307 from 2015-09 did exactly the same thing as this PR -- except that $which was not renamed to $as.

Even so, thanks to @edwinst .

AlexDaniel:
@usev6 thank you for checking!

PR #402, opened 2015-04-03
Fix NullPointerException on JVM in find-reducer-for-op

usev6:
... fixes spectest breakage on JVM


usev6:
Currently there are spectest failures with rakudo.jvm because commands like the following fail (from S03-operators/range.t):

```say (1..3).reduce({ $^a + $^b})```

PR #403, opened 2015-04-07
Make .subst-mutate on non-strings return match object(s)

usev6:


usev6:
S05 states that "the s/// form always returns either a Match object to indicate to smartmatch that it was successful, or a Nil value to indicate that it was not." Since s/// is implemented in terms of "subst-mutate", that method should return a Match object (it already does for strings).

```
$ perl6 'my $a = 707; my $ret = $a.subst-mutate(7,2); say $ret; say $ret.WHAT'
207
(Str)
$ perl6 'my $a = "707"; my $ret = $a.subst-mutate(7,2); say $ret; say $ret.WHAT'
｢7｣
(Match)
```

peschwa:
I've merged this and added a test to S05-substitution/subst.t, thanks!

peschwa:
According to tadzik, github should realize this has been merged already, and but I overzealously closed it.  I'm not sure reopening was a good idea either.  I just hope this doesn't cause any problem, and sorry for the spam.

peschwa:
For posterity: I merged it manually but apparently forgot --no-ff, which caused github not to realize that this PR is what I pushed.  With this document I'll now close manually, but rest assured this PR has been merged. Sorry again for spam and potential confusion.

PR #404, opened 2015-04-09
The -K flag to uname wasn't introduced until FreeBSD 10.0

jonathanstowe:
use -r as next best thing

Found while testing on FreeBSD 8.4 and 9.1


FROGGS:
Btw, `-K` would give `10.0` where `-r` gives `10.0-RELEASE`

PR #405, opened 2015-04-11
Gracefully handle .gist on unthrown exceptions

skids:
If an exception is created and not thrown, and gist is called on it, the backtrace op is called without a VM-level exception object, and a long backtrace about not being able to make a backtrace is produced and appended to the gist.  This patch makes .gist simply print the Exception's message if it is called before the exception is thrown.

Spectested fine.


lizmat:
Could not be merged automatically, so used is as inspiration for https://github.com/rakudo/rakudo/commit/91eaacff36  .   Thank you!

PR #407, opened 2015-04-13
.resolve: return IO::Path instead of str

ugexe:


ugexe:
From the specs:

    method resolve(--> IO::Path)

Currently:

    nickl@localhost:~/.rakudobrew$ perl6 -e 'my IO::Path $path = "/home".IO; say $path.resolve.WHAT; my IO::Path $resolved = $path.resolve'
    (Str)
    Type check failed in assignment to '$resolved'; expected 'IO::Path' but got 'Str'

labster:
New .resolve passes spectest (and the newio branch is already returning an IO::Pathy) and doesn't even break my personal script using `.resolve`, so this looks great to me.

PR #408, opened 2015-04-14
add infix:<ne>(Str,Str) subs, skip metaop for ne(Any, Any)

labster:
Makes string `ne` much closer in speed to `eq` by using `nqp::isne_s()`.
28-47% faster, depending on how many of the args are strings.

The reason I'm creating a PR is because I screwed this up like 2 years ago, but I think I have a better understanding of multi-dispatch now.  The key issue is that with junctions, `!eq` and string `ne` are not equivalent.  My understanding is that with Any, there is only one value to consider, so the `!eq` is equivalent to the simple comparision `a.Stringy ne b.Stringy` in this simpler subset.  If this is true, go ahead and merge.

Passes spectests.

### DIFF
--- a/src/core/Str.pm
+++ b/src/core/Str.pm
@@ -1453,6 +1453,13 @@ multi sub infix:<eq>(str $a, str $b) returns Bool:D {
     nqp::p6bool(nqp::iseq_s($a, $b))
 }
 
+multi sub infix:<ne>(Str:D \a, Str:D \b) returns Bool:D {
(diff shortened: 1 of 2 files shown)


labster:
On deeper reflection, I think this can be merged.  The ! metaop is concerned with booleability, and Any objects only have a single value in string context and thus only a single boolean value.  Therefore the optimized `infix:<ne>(Any, Any)` candidate should be correct.

PR #413, opened 2015-04-23
Deprecated chop with count param

ZeleiAdi:


FROGGS:
Looks like we changed our mind and are allowing an argument to chop now...

https://github.com/rakudo/rakudo/commit/0c0eb13533
https://github.com/perl6/specs/commit/0f49a8544f
https://github.com/perl6/roast/commit/c4328db38e

But thank you for you work anyway.

PR #414, opened 2015-04-23
Add rightfold to credits

rightfold:


rightfold:
@perlpilot Ah, I see! Added it.

PR #418, opened 2015-04-26
Preserve $.timezone in DateTime.later()

cjepeway:
These changes make the tests in roast/date-time-slices-tz pass.

My weekend time to work on this is about up, so I won't be coding again until next Saturday.

This fix is pretty straightforward, but there could be other "timezone gets sliced" fixes coming.  I dunno whether it makes sense to apply this one now, or wait to see if there are others.

### DIFF
--- a/src/core/Temporal.pm
+++ b/src/core/Temporal.pm
@@ -341,7 +341,7 @@ my class DateTime does Dateish {
 
         given $unit {
             when 'second' | 'seconds' {
-                return self.new(self.Instant + $amount, :$.timezone);
+                return self.new(self.Instant + $amount, :$.timezone, :&.formatter);
             }
 
             when 'minute' | 'minutes' { $minute += $amount; proceed }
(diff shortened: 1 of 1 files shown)


b2gills:
Should probably copy `&.formatter` as well

src/core/Temporal.pm
```diff
@@ -341,7 +341,7 @@ my class DateTime does Dateish {
 
         given $unit {
             when 'second' | 'seconds' {
-                return self.new(self.Instant + $amount);
+                return self.new(self.Instant + $amount, :timezone($.timezone));
```
vendethiel: `:$.timezone` should work here

cjepeway:
Will fix &.formatter, too.

cjepeway:
Hey, thanks, really.

There’s some .say’s in the utc() method that should go, though.

Chris.

cjepeway:
Arg, I forgot about this one: the $.timezone.new(0) won’t work unless somebody made Int.new().  Did that happen?

Chris.

PR #419, opened 2015-04-28
Don't print ", or" if @.available.elems == 1

ugexe:
# Corrects this: Please install  or OpenSSL::fake for openssl support.

    require OpenSSL::fake;
    CATCH {
        default { X::NYI::Available.new(:available("OpenSSL::fake"), :feature("openssl")).throw };
    }

### DIFF
--- a/src/core/Exception.pm
+++ b/src/core/Exception.pm
@@ -503,7 +503,7 @@ my class X::NYI::Available is X::NYI {
     method available-str {
         my @a = @.available;
         my $a = @a.pop;
-        (@a.join(', ') || (), $a).join(" or ")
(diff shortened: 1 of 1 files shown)


skids:
Just an idiomaticity nitpick, no biggie:

skids: m: my $a = "B"; my @a = (); say (flat (@a.join(', ') || (), $a) ).join(" or ") # That X::NYI::Available code did used to work IIRC, and a more DRY fix.
skids: m: my $a = "B"; my @a = (); say @a ?? (@a.join(', '), $a).join(" or ") !! $a; # or keeping the ternary, you can lose the || () because its function was entirely to remove the " or ".

PR #422, opened 2015-05-08
Fix undeclared variable error

raydiak:
Previously it would show the cursor in a wrong (later) position and give inappropriate "expecting any of" suggestions based on that position. I'm a bit unclear on the details of this precursor thing, so please do review carefully before merging in case this is the wrong way to fix it, but what I saw seemed to suggest that precursor may be intended for these kinds of cases.

### DIFF
--- a/src/Perl6/Grammar.nqp
+++ b/src/Perl6/Grammar.nqp
@@ -310,7 +310,7 @@ role STD {
 
                             CATCH {}
                         }
-                        $*W.throw($var, ['X', 'Undeclared'], symbol => $name, suggestions => @suggestions);
+                        $*W.throw($var, ['X', 'Undeclared'], symbol => $name, suggestions => @suggestions, precursor => '1');
                     }
                 }
                 else {


AlexDaniel:
It feels like this pull request lacks the example:

```
@arr = <hello world>;
```

```
===SORRY!=== Error while compiling ./test.pl
Variable '@arr' is not declared
at ./test.pl:9
------> @arr⏏ = <hello world>;
    expecting any of:
        postfix
```

raydiak:
Thanks, good call...I had meant to at least include a link to the discussion at http://irclog.perlgeek.de/perl6/2015-05-08#i_10570744 but apparently forgot.  The above example after this patch renders as:

    Variable '$test' is not declared
    at -e:1
    ------> use strict; ⏏$test = 1;

raydiak:
Figured it couldn't hurt to add one more example of misbehavior to show the confusion a little more fully:

    > perl6 -e 'use strict; $foo.bar'
    Variable '$foo' is not declared
    at -e:1
    ------> use strict; $foo.bar⏏<EOL>
        expecting any of:
            method arguments

raydiak:
wrt $/.PRECURSOR, I found https://github.com/perl6/nqp/blob/5e9f7b8c563d3601961751aec03228325e33d76d/src/QRegex/Cursor.nqp#L731 which seems to fit the vague assumptions I was making

PR #424, opened 2015-05-10
sub spurt can take :$bin like method spurt

ugexe:


labster:
I'm not sure why this patch is needed.  The `|c` passes any arguments sub spurt gets on to method spurt, which means I don't think behaviour would change.

ugexe:
You are correct. I should clarify this was to make the no signature match error a tad more consistent with slurp.

    nickl@localhost:~$ perl6 -e 'say slurp(Str);'
    Cannot call slurp(Str); none of these signatures match:
        (IO::ArgFiles:D $io = { ... }, Any :bin($bin), Any :enc($enc) = { ... }, Any |c)
        (IO::Handle:D $io = { ... }, Any :bin($bin), Any :enc($enc) = { ... }, Any |c)
        (Cool:D $path, Any :bin($bin) = { ... }, Any :enc($enc) = { ... }, Any |c)
      in block <unit> at -e:1
    nickl@localhost:~$ perl6 -e 'say spurt(Str);'
    Calling spurt(Str) will never work with any of these multi signatures:
        (IO::Handle $fh, Any $what, Any :enc($enc) = { ... }, Any |c)
        (Cool $path, Any $what, Any :enc($enc) = { ... }, Any |c)
    at -e:1
    ------> say ⏏spurt(Str);

PR #426, opened 2015-05-13
Add support for additional mode flags to IO::Handle.open

cygx:
In addition to `:r`, `:w`, `:a` and the (previously ignored) `:rw`, there's now also `:ra`, `:wx`, `:wr`, `:wrx`, `:ar`.

With the exception of `:ra` (added for consistency), these correspond to the modes of `fopen()` as specified by C11 - you get to the Perl6 name by replacing `+` with `r` or `w` as appropriate.

The `$llmode` parameter to the private `!open` method translates directly to POSIX flags:

    r  O_RDONLY
    -  O_WRONLY
    +  O_RDWR
    a  O_APPEND
    c  O_CREAT
    x  O_EXCL
    t  O_TRUNC

`w` is a special compatibility alias that translates to `-ct` on its own and to `-c` as part of a string. This ensures that `w` and `wa` still map to their expected values.

Adding Perl5 into the mix, we end up with the following:

    p6      c11     moar    compat  p5
    ************************************
    :r      r       r               <
    :rw     r+      +               +<
    :ra             +a
    :w      w       -ct     w       >
    :wx     wx      -cx     wx
    :wr     w+      +ct             +>
    :wrx    w+x     +cx
    :a      a       -ca     wa      >>
    :ar     a+      +ca             +>>

On the JVM, unsupported modes will blow up, but everything that used to work should continue to work (with the caveat that `:rw` now blows up instead of silently getting adjusted to `:w`).

On Parrot, unsupported modes will lead to undefined behaviour as unknown characters get silently dropped from mode strings.

### DIFF
--- a/src/core/IO/Handle.pm
+++ b/src/core/IO/Handle.pm
@@ -13,11 +13,30 @@ my class IO::Handle does IO {
         self.open(:p, |c);
     }
 
-    method open(IO::Handle:D:
-      :$r is copy,
-      :$w is copy,
-      :$rw,
-      :$a,
+    proto method open(|) { * }
+
+    multi method open(IO::Handle:D: :$r!, :$w!, *%_) {
+        die "Cannot use both :r and :w - use :rw, :wr or :wrx instead";
+    }
+
+    multi method open(IO::Handle:D: :$r!, :$a!, *%_) {
+        die "Cannot use both :r and :a - use :ra or :ar instead";
+    }
+
+    multi method open(IO::Handle:D: :$r!  , *%_) { self!open('r'  , 'r'  , |%_) }
+    multi method open(IO::Handle:D: :$rw! , *%_) { self!open('rw' , '+'  , |%_) }
+    multi method open(IO::Handle:D: :$ra! , *%_) { self!open('ra' , '+a' , |%_) }
+    multi method open(IO::Handle:D: :$w!  , *%_) { self!open('w'  , 'w'  , |%_) }
+    multi method open(IO::Handle:D: :$wx! , *%_) { self!open('wx' , '-cx', |%_) }
+    multi method open(IO::Handle:D: :$wr! , *%_) { self!open('wr' , '+ct', |%_) }
+    multi method open(IO::Handle:D: :$wrx!, *%_) { self!open('wrx', '+cx', |%_) }
+    multi method open(IO::Handle:D: :$a!  , *%_) { self!open('a'  , 'wa' , |%_) }
+    multi method open(IO::Handle:D: :$ar! , *%_) { self!open('ar' , '+ca', |%_) }
+    multi method open(IO::Handle:D:         *%_) { self!open(''   , 'r'  , |%_) }
+
+    method !open(IO::Handle:D:
+       $mode,
+       $llmode,
       :$p,
       :$bin,
       :$chomp = True,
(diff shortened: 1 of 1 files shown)


donaldh:
This feels like the wrong path to extending the open method. I think it would be better to introduce adverbs like :truncate and :exclusive rather than a string representation that is neither POSIX nor Perl 5 but more complex than both. 

I also think it would be beneficial to push the implementation down to NQP and Moar.

cygx:
@donaldh `&nqp::open` essentially *is* just a wrapper over POSIX. The difference in API between Perl6 `&open` and `&nqp::open` is basically the same one as between libc `fopen()` and `open()`

donaldh:
My feeling is that we should be migrating from mode strings to flags for the interface between rakudo and nqp. The moar and jvm backends both use implementations that are compatible with the open system call and its flags.

I think the interface we define for Perl 6 should not overload the :r :w adverbs but instead introduce :create :truncate and :exclusive. I think that would align better with what has been specced so far in http://design.perl6.org/S32/IO.html

cygx:
@donaldh That's certainly one way to go. If we go that way, it would probably be a good idea to make `:rw` identical to `:r, :w` and make the *opposite* of `:create` into an adverb (`:exists`?).

I'll have to think about it some more, but the basic building blocks would be

    :r :w :exists :append :truncate :exclusive

with the possibility of shortcuts, eg

    :rw --> :r :w
    :a   --> :w :append

donaldh:
Well O_RDONLY, O_WRONLY and O_RDWR are actually exclusive. It's debatable whether we should adjust the semantics to allow :r and :w to be composable, or just keep :r, :w and :rw as exclusive flags.

http://pubs.opengroup.org/onlinepubs/9699919799/functions/open.html

labster:
I feel like `:wr` is bad huffmaning, because it's an uncommon mode that can clobber your files if you typo `:rw`.  Tokens that do different things shouldn't be spelled so similarly.

cygx:
I think I came up with a scheme that I'm comfortable with that incorporates the criticism:

    multi method open(IO::Handle:D: :$p!, *%_) {
        self.open(:mode<p>, |%_);
    }

    multi method open(IO::Handle:D: :$a!, :$r!, *%_) {
        self.open(:mode<rw>, :create, :append, |%_);
    }

    multi method open(IO::Handle:D: :$a!, *%_) {
        self.open(:mode<wo>, :create, :append |%_);
    }

    multi method open(IO::Handle:D: :$r!, :$w!, *%_) {
        self.open(:mode<rw>, :create, |%_);
    }

    multi method open(IO::Handle:D: :$rw!, *%_) {
        self.open(:mode<rw>, :create, |%_);
    }

    multi method open(IO::Handle:D: :$w!, *%_) {
        self.open(:mode<wo>, :create, |%_);
    }

    multi method open(IO::Handle:D: :$r?, *%_) {
        self.open(:mode<ro>, |%_);
    }

    multi method open(IO::Handle:D:
      :$mode!,
      :$create,
      :$append,
      :$truncate,
      :$exclusive,
      :$bin,
      :$chomp = True,
      :$enc   = 'utf8',
      :$nl    = "\n",
    ) { ... }

That should be less cryptic and still DWIM. For some reason, my code breaks the build, so no commits yet.

FROGGS:
:+1:

FROGGS:
though I would think that every combination like `:rwx` should be allowed to be passed as `:r :w :x`

cygx:
Superseded by #430 - third time's the charm, right?

PR #429, opened 2015-05-14
unbust warning for type objects as Range endpoint for JVM

usev6:
At the moment something like ```1..Any``` fails with "Cannot unbox a type object" on JVM.


usev6:
Instead of a special case for JVM one could combine the checks like below. For some reasons on Moar it suffices to check "nqp::isnull_s" whereas on JVM it suffices to check "nqp::isconcrete".

```
%anno := callframe(++$i).annotations while nqp::isconcrete(%anno<file>) && !nqp::isnull_s(%anno<file>) && %anno<file> ~~ /\.setting$/;
```

PR #432, opened 2015-05-17
Adding test demonstrating structs with nested structs are not working

zostay:
For @FROGGS: I've added a test to highlight the kind of issues I'm having with the CStruct repr. At this point CUnion is not the problem. 

Working with the Gumbo library, it has some structs like this with nested structs that are allocated with the parent, mixed with primitive variables. It would appear that moar is basically ignoring the nested struct when calculating address lookups for later values stored within the struct.

I've been reading the C code in MoarVM to see if I can figure it out, but I need to learn a lot moar vocabulary yet before I understand what it's doing well enough to consider fixing it.

I think I might be able to fake it out and make it work by adding the primitives from the nested struct directly into the parent, but I haven't tried it. Even if I do, I don't think that's the right solution.

### DIFF
--- a/lib/NativeCall.pm
+++ b/lib/NativeCall.pm
@@ -155,6 +155,7 @@ my %repr_map =
     'CStruct'   => 'cstruct',
     'CPointer'  => 'cpointer',
     'CArray'    => 'carray',
+    'CUnion'    => 'cunion',
     'VMArray'   => 'vmarray',
     ;
 sub type_code_for(Mu ::T) {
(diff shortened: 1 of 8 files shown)


zostay:
I tried copying the contents of the nested struct into the parent and that does work.

FROGGS:
Ahh, you must know that until now we treat structs in attribute slots as struct*.
The cunion branch has an 'is inlined' trait that we could make use of here...

labster:
Is there a reason why these tests are here rather than in perl6/roast?

arnsholt:
The NativeCall tests depend on CompileTestLib (in the same dir), which is
unashamedly Rakudo/NQP specific.

The problem is figuring out how to compile C libraries (which we need to
test things properly), without implementing half of autoconf.

FROGGS:
@labster: Also I am not sure if NativeCall is something that belongs to the Perl 6 language itself.

jonathanstowe:
Well in the sense that it is described as such in S21 I would say either it is or the wording of the first sentence of the synopsis should be altered somewhat :-)

FROGGS:
Well, S21 is a draft and mixes specification and synopsis :o)
But I think I can be convinced that NativeCall is meant to be part of the Perl 6 spec (roast).

FROGGS:
I cherry-picked your commit, fixed MoarVM and added a needed `is inlined` trait to your test structure.

PR #438, opened 2015-06-10
Had missed kebab casing is_approx to is-approx

jonathanstowe:
When the underscore names were deprecated in favour of the kebab cased ones it appear is_approx was missed.


lizmat:
The is_approx was deliberately  not kebab-cased to allow for a smooth migration to a better is-approx with more options.  So rejecting this pull request.

jonathanstowe:
That's fine, I think I knew that originally and then forgot.

PR #440, opened 2015-06-15
wrap paragraph lines at 75 chars if POD_TO_TEXT_WRAP is set

stmuk:


stmuk:
going to tweak following IRC feedback

PR #446, opened 2015-06-22
Add .sigil and .twigil methods to Parameter

skids:
This is an implementation-independent way expose more of rakudo's $!flags
  It is also a convenience versus piddling around with output from .perl

I found this to be the main shortcoming in introspection when playing with a better .assuming
implementation.  Not sure how far I will get with the that but this seemed to be an obvious
step forward at least.

### DIFF
--- a/src/core/Parameter.pm
+++ b/src/core/Parameter.pm
@@ -36,6 +36,44 @@ my class Parameter { # declared in BOOTSTRAP
         nqp::isnull_s($!variable_name) ?? Nil !! $!variable_name
     }
 
+    method sigil() {
+        my $sigil = '';
+        my $name = $.name;
+        if $name {
+            $sigil = $name.substr(0,1);
+            if $!flags +& $SIG_ELEM_IS_CAPTURE {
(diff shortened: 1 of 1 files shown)


FROGGS:
Hmmmm, to keep consistency I'd argue that we also want .sigil/.twigil on variables:

```perl6
$ my $a; say $a.VAR.^methods
<anon> <anon> name of default dynamic WHICH
```

skids:
I wouldn't be averse to hunting that down and adding it if that's desired, but probably in a different fork as it is not ".assuming" related.

One thing that is not entirely clear to me right at the moment is the distinction between "$a is parcel" and "\a", or what "@a is parcel" does.  .perl will turn "$ is parcel" or "@ is parcel" into "Any \" and I did the same for .sigil but... I'm befuddled there.

skids:
Oh yeah, another set of flags that are buried a bit are DEFINED_ONLY and UNDEFINED_ONLY (:D and :U) but I can't think of what a implementation-independent method for those should be named.  Suggestions welcome.

PR #454, opened 2015-07-01
Handle "is parcel" better in Parameter.perl and Parameter.sigil

skids:
This modifies:

  Parameter.sigil to only be '\' when it would otherwise be '$' but
  the parameter 'is parcel'

  Parameter.perl to emit 'is parcel' in the cases where unspace
  would mess up a re-parse, not to emit '\%' or '\&' anymore
  and use 'is parcel' instead for those, and to correctly add an
  'is parcel' to '|c is parcel'.  I don't know if the last one makes
  any sense, but if it doesn't it probably should not parse in the
  first place.

### DIFF
--- a/src/core/Parameter.pm
+++ b/src/core/Parameter.pm
@@ -260,11 +260,8 @@ my class Parameter { # declared in BOOTSTRAP
                and not $rest
                and nqp::isnull($!post_constraints) 
                and not $default
-               and nqp::isnull($!sub_signature)
-               {
-                if $name eq '$' and not $rest {
+               and nqp::isnull($!sub_signature) {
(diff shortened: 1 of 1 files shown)


skids:
Due to unspace a lot of stuff misparses when using an anonymous
parcel parameter expressed as '\' instead of '$ is parcel'.

There are observable differences between '\a' and '$a is parcel' but
not between '\' and '$ is parcel' since you cannot reference those
inside the block body anyway.  We can tell the difference between
the former but not the latter from inside a Paramter object.

There was some suspicious logic regarding the '@' sigil that treated
it in many ways the same as the '$' sigil (which deserves it.)

PR #457, opened 2015-07-02
Obey use :from even when a similar named Perl 6 module is installed i…

niner:
…n a CUR

There's no need to even look for a Perl 6 module in CompUnitRepo when there's
a :from adverb present. So don't do that and let the Perl6ModuleLoader handle
the case straight away.

Fixes for example use lib:from<Perl5>


drforr:
This will allow 'use lib:from<Perl5> "./lib";' in scripts to properly reference the perl5 embedded interpreter when available, just to bolster the case.

PR #463, opened 2015-07-07
  .assuming now produces closures with populated signatures.

skids:
... (fixes RT#78240/RT#77744 noting that 2 tests in RT#77744 are
  ...  outdated due to trying to use named parameters to fill positionals)
  .assuming now supports * and Nil primers with special behaviors.
  .assuming will try to bind primer arguments to a partial signature:
  ... currently it will mix most failures into the returned closure.
  ... eventually we can phase-in throwing them instead.
  For now, subsignatures are mostly ignored during this bind.
  For now, constraints pointing to captured types are stripped.
  Where clauses and default value exprs are not evaluated until runtime invoke.
  This currently uses EVAL to construct the closure, which is LTA,
  but it gives us something functional/testable to work forward from.
  Roast tests available at https://github.com/skids/roast/tree/assuming

### DIFF
--- a/src/core/Block.pm
+++ b/src/core/Block.pm
@@ -35,10 +35,314 @@ my class Block { # declared in BOOTSTRAP
         ()
     }
 
-    method assuming(Block:D $b: |curried) {
-        anon sub CURRIED (|direct) {
-            $b(|curried, |direct)
+    method assuming(Block:D $self: |primers) {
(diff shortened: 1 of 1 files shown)


skids:
astoundingly, works under JVM too and seems to make RT#125207 go away as well

PR #465, opened 2015-07-07
Proposed IO::Path slash operator

zostay:
I've seen this operator defined in Ruby code someplace. It seems pretty slick to me to say something like:

    my $homedir = %*ENV<HOME>.IO;
    my $bindir = $homedir/"bin";

### DIFF
--- a/src/core/IO/Path.pm
+++ b/src/core/IO/Path.pm
@@ -635,4 +635,8 @@ my class IO::Path::Win32 is IO::Path {
     method new(|c) { IO::Path.new(|c, :SPEC(IO::Spec::Win32) ) }
 }
 
+multi sub infix:</> (IO::Path:D $base, Str:D $child) is assoc<left> returns IO::Path:D {
+    $base.child($child);
+}
+
 # vim: ft=perl6 expandtab sw=4


b2gills:
Generally the Perl 6 built in operators do only one thing.  
Since that is the division operator it is unlikely that it will get merged.

zostay:
I figured as much, but I like it a lot, so thought it was worth a shot. :)

FROGGS:
:+1: to put that in a module, but then you might also want to handle the `Str/IO::Path` case.

colomon:
I think it's a lovely idea for a module, but even there I think it would
still be prudent to find a symbol for the operator that doesn't already
mean something.

Related idea: maybe a prefix op to make a string into a path?  Analogously
to prefix:<+> for numbers and prefix:<~> for strings?  Then the directory
path extension infix could be that symbol followed by a slash, and you
might also have a directory concatenation op (which would just add to the
directory name without adding a slash).

Don't have any coherent notion what a good symbol would be, though.  This
early in the morning, all I can think of is prefix:<C:>, but that would be
very Windows-centric and ugly...

On Wed, Jul 8, 2015 at 2:25 AM, Tobias Leich <notifications@github.com>
wrote:

> [image: :+1:] to put that in a module, but then you might also want to
> handle the Str/IO::Path case.
>
> —
> <https://github.com/rakudo/rakudo/pull/465#issuecomment-119451824>.
>



-- 
Solomon Foster: colomon@gmail.com
HarmonyWare, Inc: http://www.harmonyware.com

vendethiel:
maybe `~/`? somewhat of a string concat, somewhat of a directory separator...

jonathanstowe:
This could be somewhat alleviated by a bunch of multis on the '~' where one or both parts are an IO::Path

ugexe:
`~` / `method Str` already stringifies an IO::Path

    perl6 -e 'my $p = "/home".IO; say $p; say ~$p'
    "/home".IO
    /home

If one considered adding an extra character, it would then be longer than what can currently be invoked:

    perl6 -e 'my $p = "/home".IO; my $p2 = "lib"; say "$p/$p2".IO;'
    "/home/lib".IO

    perl6 -e 'my $p = "/home".IO; my $p2 = "lib"; say $p~"/$p2".IO;'
    /home/lib

It *would* be nice to have a concat that gives back an `IO::Path.rel2abs($p2, $p)` or `$*SPEC.catdir($p, $p2)` though, so as not to lose $*SPEC's CWD

vendethiel:
Perl6 operators will usually convert their operands to the correct type ;-). (see "==" vs "eq")

jonathanstowe:
As an example:

multi sub infix:<~> (IO::Path $l, IO::Path $r) { $*SPEC.catdir($l, $r) } 

does work (though it probably wants to return an IO::Path itself, ) just throwing some ideas in - I'm cool with the way it is right now to be honest.

zostay:
As alternate names go, I like `/~` myself. It conveys the notion of a modified concatenation. It's not quite as idiomatic as `/`, but it doesn't mix up the existing Perl semantics either. I don't like `~/` as it could be confused with the Unix home directory notation, especially if used as a prefix.

I also think the idea of a prefix operator to create a path is a nice idea, but not I'm not sure how often I would use it. The problem with a prefix is that it's precedence will not be high enough to make it as useful as `.IO`.

    /~ 'bin' .mkdir; #!> Method 'mkdir' not found for invocant of class 'Str'
    (/~'bin').mkdir; # works, but meh
    'bin'.IO.mkdir; # works

It's more typing to use in a case like this at least.

And just to see it in an infix situation:

    if (my $p = $root /~ 'bin' /~ 'perl6-m').x {
        run $p, '-v';
    }

I could cope with that.

The reason I did not create alternatives for `:(IO::Path, IO::Path)` or `:(Str, IO::Path)` was that I was confused into thinking that paths are always absolute. I got this impression because `.gist` always returns the abs-path, which seems wrong to me when `.perl` and `.Str` don't do that.

By using `/~` or similar, a `:(Str, Str)` version becomes safer, so it's probably the way to go.

labster:
Just what Perl 6 needs, more operators!

An earlier spec suggested `qp//` as a path quoting operator but, uh, no one cared enough to implement it.  Which we took to mean that it was probably not important.

It's not like it `.IO` is a long Huffmaning anyway -- prefix `/~` just seems like a less readable IO coercer.   I'm not sold on infix `/~` either.  I think I'd prefer `~/` because it feels more like a concatenation, otherwise it seems like a slash (pathy?) metaop on concatenation.  But not sure either are good ideas.

However, if we use `/~`, that leaves string division `~/` available.  Which I guess means we will still be able to solve "MONEY" ~/ "SEND" == "MORE" automatically.

moritz:
infix `/` is our operator for numerical division; overloading it for a totally different purpose is not acceptable.

jonathanstowe:
phew

PR #468, opened 2015-07-13
dead code removal: `.absolutify()`

ugexe:
The extra keystrokes are to work around dying if `$path eq ''` is theoretically set from `PARSE-INCLUDE-SPEC`


lizmat:
Disagree.  The .absolutify method should allow any CUR implementation control over how a path should be absolutified.  Therefore, I'm rejecting this patch.  Alternately, maybe method absolutify should be implemented in CompUnitRepo::Locally.

PR #470, opened 2015-07-13
Enhancements to Failure/Exception/fail/die

skids:
This branch contains various enhancements to failure/exception code.  Some of them are minor spec changes and some may merit a bit of bikeshedding over exact text.  Some may be cherry-pickable.

Most of the improvements give better messages and/or increase the leniency of constructors and
fail/die.  One implements behavior defined in S04 but which rakudo either never implemented or was bitrotted out.  One RT is fixed.  Two tests break, one of which should be tweaked as its expectations are off, and the other is an RT test which hits one of the spec changes.

A script that shows the expected behaviors after the changes is at
https://gist.github.com/skids/2639325b87d5d55840fd

I may add some more commit to this set, but the issues that remain may be more formidable given exception throwing goes layers deep.  The remaining issues I see are:

1) Backtraces are not consistently shown sometimes, and sometimes show a lot of cruft from inside the failure throwing code and stop before getting to useful information.
2) A tweak of "use fatal" that causes Failures created with unthrown exceptions or adhoc text to take a backtrace at the failure creation time and emit two backtraces -- one from that point and one from the actual throw -- is needed.
3) Since 2) may be useful during debugging an additional non-lexical control to cause that behavior may be needed.

### DIFF
--- a/src/core/Backtrace.pm
+++ b/src/core/Backtrace.pm
@@ -80,11 +80,11 @@ my class Backtrace {
         return nqp::atpos($!frames,$pos) if nqp::existspos($!frames,$pos);
 
         my int $elems = $!bt.elems;
-        return Nil if $!bt-next == $elems;
+        return Nil if $!bt-next >= $elems; # bt-next can init > elems
 
         my int $todo = $pos - nqp::elems($!frames) + 1;
+        return Nil if $todo < 1; # in case absurd $pos passed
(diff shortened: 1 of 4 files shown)


skids:
This PR picked up a merge conflict with later commit 03d13c1.  I don't know how best to deal with that.

PR #479, opened 2015-07-26
Workaround RT#120919 to fix Date/Temporal on JVM backend

skids:
The VALID-UNITS method cannot be both private and in a role on JVM
due to RT#120919.  Also the crash caused by this bug is nasty because
it kills the stack trying to build exceptions inside exceptions due
to repeated failure to bind '$?CLASS' to typed exception attributes.

### DIFF
--- a/src/core/Temporal.pm
+++ b/src/core/Temporal.pm
@@ -8,7 +8,7 @@ my role Dateish {
 
     method IO(|c) { IO::Path.new(self) }
 
-    method !VALID-UNIT($unit) {
+    method VALID-UNIT($unit) {
         state %UNITS =  # core setting doesn't build if it is a my at role level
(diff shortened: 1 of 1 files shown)


labster:
Can we leave a code comment somewhere to revert this when RT#120919 -- otherwise someone is going to wonder why private-looking methods are leaking.

skids:
I was going to upload such a comment to roast.  But then I figured maybe best to post it in the RT?  Have not decided yet.  In addition to src/core/Temporal.pm there is a pre-existing workaround in src/core/IO/Local.pm.

skids:
OK, I added a coet in the RT ticket saying what to do to remove workarounds.

skids:
Bug long since resolved and it looks like this workaround was removed as VALID-UNIT is back to a private method now.

PR #481, opened 2015-07-27
Various fixes to Signature introspection and take safety CATCH off .assuming

skids:
These patches are mostly fixes to problems with .perl methods noticed while/since writing
the new .assuming.  In addition since there have been no complaints with .assuming since it was 
merged, the final commit removes the nerf from the part that checks binding of primer arguments.


moritz:
Thanks!

jonathanstowe:
BTW I had tested the new .assuming and it worked :)

PR #487, opened 2015-08-04
Complain when trying to override smartmach operator

usev6:


usev6:
This follows the suggestion from https://rt.perl.org/Ticket/Display.html?id=125745#txn-1359531.

ab5tract:
It might be useful to also mention that what the user is probably looking for involves tweaking .ACCEPTS on whatever they are putting on the LHS.

jonathanstowe:
:+1:  May look at a way of documenting the ACCEPTS thing better

usev6:
For the records: jnthn++ added a hint '(consider implementing an ACCEPTS method)' to the error message: https://github.com/rakudo/rakudo/commit/16d366a2d227789655c7b53c30edc842eb57b3eb

PR #493, opened 2015-08-14
GLR Array.unshift corrections and optimization

skids:


skids:
This seems to handle many of the cases that the current code gets wrong,
with the exception (perhaps) of .unshift([1,2],) which is a more general issue
as .([1,2]), will choose a :($) candidate over a :(**@) candidate, and that may not
be the intended GLR behavior of dispatch.

skids:
Closing to resolve merge conflict with is-lazy changes.

PR #494, opened 2015-08-14
GLR Array.unshift corrections and optimization

skids:


skids:
Third time's the charm I hope.

With this implementation S32-array/unshift passes all tests except for
the typed array test (not implemented yet) and the test that breaks when
it tries to assign a completion to an array, before actualy testing unshift.

PR #495, opened 2015-08-15
Add temporary short-circuits to "array" preventing some test hangs.

skids:


skids:
This just prevents an infinite loop in S02-types/array-shapes and dies instead

PR #502, opened 2015-08-20
Start to make new GLR code Failure-transparent

skids:


skids:
The second of the two commits will need some debate.

The pros of diverging from the normal convention that push returns
what you just pushed are that it fixes many locations in both current
and future code where push is used in sink context, at least for those
who have not made their own IterationBuffer subclass or fed alternative
target objects into the API.  It also encourages anyone subclassing
IterationBuffer to do the same.  And who knows, having .elems handy
after a push may actually be useful.

The cons are that it sweeps problems with Failure transparency under the
rug so they only crop up in rare and unexpected places, affecting users
who may not be inclined to invest in debugging.

zhuomingliang:
I think it is wrong. At least, it is ugly hack, and being used everywhere, which is not encouraged.  :)
If it is right, we should fix it in the codegen step of nqp, instead of using ugly hack everywhere.

skids:
zhuomingliang, I assume you are referring to the alteration to IterationBuffer.push?  The thought had occurred to me that we might need a separate op for "push that does not return the pushed value" (whether it returns elems or something else being up for grabs.)

I don't see any way codegen can fix the general need to avoid placing Failures in sink context unless you intend to throw them -- that's part of the design of Failure.  We'd need a higher level construct to indicate that, or to rearrange the code somehow to prevent it.

There are actually some strange behaviors related to this which may be codegen/optimizer related.
RT#125769 points out a discrepency between binding and assignment of Failures when returned as the result of an statement.  Getting to the bottom of that might simplify things.  I've seen a few other places where binding and/or sigilless behave differently.  

For that particular RT, whether the if statement is semantically considered a loop (and should not sink its result because it is a 1-list) needs to be nailed down first to know which way to fix it.

skids:
It appears that Seqs of Seqs needed many of the same changes as this PR, per commit c87c2037018b.  I'll keep this open for discussion on IterationBuffer.push's return value and as a resource in case c87c2037018b missed some spots.

PR #504, opened 2015-08-20
add .absolutify() to CUR::Locally

ugexe:
A first crack at lizmats reponse in #468 

Allows CUR::Installation to install from non-$*CWD directories (related to copying files, and getting the proper file modification time for the MANIFEST).

It seems like the base path (in this case, :$from added to CURLI.install) might be better off as an attribute of Distribution. If such a solution is better, a proper name for the attribute will be needed (it may contain a file path, a url, or not exist at all)

### DIFF
--- a/src/core/CompUnitRepo/Locally.pm
+++ b/src/core/CompUnitRepo/Locally.pm
@@ -32,6 +32,10 @@ role CompUnitRepo::Locally {
         self.short-id ~ '#' ~ $!IO.abspath;
     }
 
+    method absolutify(CompUnitRepo::Locally:_: |c) {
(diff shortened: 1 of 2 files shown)


src/core/CompUnitRepo/Locally.pm
```diff
@@ -32,6 +32,10 @@ role CompUnitRepo::Locally {
         self.short-id ~ '#' ~ $!IO.abspath;
     }
 
+    method absolutify(CompUnitRepo::Locally:_: |c) {
```
FROGGS: Can you please remove the `:_`? It adds no meaning but potentially confusion.

PR #505, opened 2015-08-21
CURLI uninstall

ugexe:
Adds an uninstall method.
Treats `$repo<file-count>` more literally, and  obtains $file-id by looking for @files.max


lizmat:
install takes a dist to install, shouldn't uninstall do that as well?

PR #508, opened 2015-08-21
Allow `0` as a valid file name

ugexe:


labster:
Uh... isn't "0" True?  So is an attempt to make `0.IO` work?

ugexe:
Yes, it is to get 0.IO to work.

    $ perl6 -e 'my $module = 0; my $path = $module.IO.absolute(%*CUSTOM_LIB<site>); say $path'
    Must specify something as a path: did you mean '.' for the current directory?
      in block <unit> at -e:1

Sure, it could be stringified by the user, but `1.IO` works.

lizmat:
Re-imagined with 28faf1918068ffb710f4d14 , thanks for the nudge!

PR #509, opened 2015-08-22
Remove say/note (\x) multi candidate too

zhuomingliang:


FROGGS:
What is the motivation to remove them?

jnthn:
They are there for performance reasons; I'd prefer they stay.

zhuomingliang:
@jnthn how about https://github.com/rakudo/rakudo/commit/5c0c7bcfe0670ad51d6ee029cbb29ca2711efd89 ?

moritz:
I removed the `print` candidate, because with it, `my @array = <a b>; print @array` produces a space between `a` and `b`, which seems wrong. Having the slurpy handle the array case fixes it.

jnthn:
Makes sense for print, but isn't changing it for say/note not just foolish
consistency (e.g. silly)?

PR #510, opened 2015-08-22
Handle :0x and :nth(*) correctly and always return Positional

skids:


skids:
Shoud pass S05-modifiers/counted.t with this.

PR #511, opened 2015-08-23
strip_path_prefix() should not return early when searching PATH for the script

carbin:
This prevents the full path being returned, even though the file exists
in PATH, when there is a directory with the same name as the script in a
location that appears earlier in PATH.

The real-world situation where this happens is if PATH contains '.' before where panda is installed (because PATH already contained '.' at the end and panda's path was appended to PATH) and your home directory contains a checkout of panda:

    ~$ panda
    Usage:
      /home/user/rakudo/install/share/perl6/site/bin/panda [--notests] [--nodeps] install [<modules> ...] -- Install the specified modules
      [...]
    ~$ cd rakudo/ 
    ~/rakudo$ panda
    Usage:
      panda [--notests] [--nodeps] install [<modules> ...] -- Install the specified modules
      [...]

With this patch:

    ~$ panda
    Usage:
      panda [--notests] [--nodeps] install [<modules> ...] -- Install the specified modules


ugexe:
What the removed comment means is that the command `$cmd arg1 arg2` can end up running a different program if you invoke it from anywhere outside of the first instance of it in `PATH`. The real world situation for this is much broader than you note, such as multiple copies of rakudo installed (moarvm, jvm) and invoking programs with their full path (like bin wrappers generated by https://github.com/rakudo/rakudo/blob/nom/src/core/CompUnitRepo/Local/Installation.pm#L68)

carbin:
In that case the right way to fix this would be to check that the executable found is a file?

ugexe:
No. `perl6` the command may point to `/home/perl6-moarvm/perl6` or `/home/perl6-jvm/perl6`, and depending on which one is first in `PATH` is the one that owns the alias `perl6`. If I invoke a script via `/home/perl6-jvm/perl6 script.pl6 --help` it should not tell me `perl6 script .pl6 [some help message]`, because if I ran that command it would actually run `/home/perl6-moarvm/perl6 script.pl6`. The problem is simply the result of using multiple perl6 installs (or rakudobrew)

ugexe:
Lets assume you use rakudobrew. So, the first PATH it stops at is probably `~/.rakudobrew/bin` so `~/.rakudobrew/bin/perl6` is invoked by `perl`. That itself runs another command, `/some/path/$moar-or-jvm/perl6 "@"`, and that instance has no idea it was invoked by a perl6 wrapper. When it reaches the point you are inspecting in the code it is impossible for it to know that the first PATH it matches against $base has any relation at all to the current `$EXECUTABLE` (because again, we are running a completely different file path that was *started* by a $base of the same name earlier in the `PATH` order).

carbin:
I may be misunderstanding something but the updated patch shouldn't change that.
This should only prevent false positives with a directory with the same name as the script occurring earlier in PATH and causing the full path to be printed when it doesn't need to be.

Say PATH looks like /home/user/bin:/usr/bin:/opt/bin

In /opt/bin you have a script named 'less', because there is already a less in /usr/bin, when run it prints the full path:

    $ /opt/bin/less
    Usage:
      /opt/bin/less test

This should still happen with the updated patch.

You have another script in /opt/bin called 'foobar', there is no other script named 'foobar' in /usr/bin or /home/user/bin so it doesn't print the complete path:

    $ foobar
    Usage:
      foobar test

However, if you have a directory named 'foobar' in /home/user/bin

    $ mkdir ~/bin/foobar
    $ foobar
    Usage:
      /opt/bin/foobar test
    $ which foobar
    /opt/bin/foobar

That's what this aims to fix.

ugexe:
I was stuck referring to the broader problem. Your new commit indeed fixes a bug.

AlexDaniel:
Thanks! Merged.

PR #512, opened 2015-08-25
Make push and unshift iterate on single arguments in containers.

skids:
Per http://irclog.perlgeek.de/perl6/2015-08-24#i_11106557
itemization only protects against flattening, not single-arg-rule.

When push and unshift were implemented they assumed the opposite.

This brings their behavior in line and fixes as much of the
fallout from that in rakudo.  Even with those fixes we lose
progress on the test suite -- 7 files and some more tests in
already failing files need adjustment because they relied on
the previous behavior.

Some of the internal fixes may be superstitious, but can be
cleaned up later (you must be able to know you will not get any
Iterables to determine what is safe, and that is time consuming.)

### DIFF
--- a/lib/Test.pm
+++ b/lib/Test.pm
@@ -604,7 +604,7 @@ sub _init_vars {
 }
 
 sub _push_vars {
-    @vars.push: item [
+    @vars.push: ([
       $num_of_tests_run,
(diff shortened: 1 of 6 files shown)


skids:
GLRR removed the need for this PR.

PR #522, opened 2015-09-01
Slice assignment improvements

skids:
The first patch fixes some issues.
The second is experimental, and kept separate so it may be reverted separately.

### DIFF
--- a/src/core/array_slice.pm
+++ b/src/core/array_slice.pm
@@ -1,12 +1,5 @@
 # all sub postcircumfix [] candidates here please
 
-# Generate up to N positions
-sub NPOSITIONS(\pos, \elems) {
-    my \indexes = IterationBuffer.CREATE;
-    pos.flat.iterator.push-exactly(indexes, elems);
-    nqp::p6bindattrinvres(List.CREATE, List, '$!reified', indexes)
-}
-
 # Generates list of positions to index into the array at. Takes all those
 # before something lazy is encountered and eagerly reifies them. If there
 # are any lazy things in the slice, then we lazily consider those, but will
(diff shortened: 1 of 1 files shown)


zhuomingliang:
It would be nice to have some tests in roast too. ;)

skids:
Yeah, I'll be writing some tests for the Positional stuff at least.  If I add any for the lazy Seq stuff I'll be sure to mark them experimental/tenuous.

src/core/array_slice.pm
```diff
@@ -221,37 +289,37 @@ multi sub postcircumfix:<[ ]>(\SELF, Callable:D $block, :$SINK!, *%other) is rw
...
 multi sub postcircumfix:<[ ]>(\SELF,Callable:D $block,:$exists!,*%other) is rw {
-    my $pos := $block(|(SELF.elems xx $block.count));
+    my $pos := $block(|(SELF.elems xx ($block.count == Inf ?? 1 !! $block.count)));
     nqp::istype($pos,Int)
       ?? SLICE_ONE_LIST(  SELF,  $pos, :$exists, |%other )
       !! SLICE_MORE_LIST( SELF, @$pos, :$exists, |%other );
 }
 multi sub postcircumfix:<[ ]>(\SELF, Callable:D $block, :$kv!, *%other) is rw {
-    my $pos := $block(|(SELF.elems xx $block.count));
+    my $pos := $block(|(SELF.elems xx ($block.count == Inf ?? 1 !! $block.count)));
```
moritz: I think it would better to write that as `my $count = $block.count; my $pos := $count == Inf || $block == 1 ?? $block(SELF.elems) ?? $block(|(SELF.elems xx $count)`

(and likewise in the other multis).

postcircumfixes are a rather hot path, and we should avoid the infix:<xx> call for a single element, as well as the double `.count` calls for the common path (.count == 1).

skids:
Sure.  I'll leave that for future optimization efforts, though.

tony-o:
Please ignore 6fa42c1, it is unrelated to this PR.

PR #524, opened 2015-09-03
Restore bare Whatever candidate to map method.

skids:
Seems to have gotten removed during GLR.


jnthn:
It was intentionally removed, replaced by Seq.from-loop )which will also be wired up to handle l-value loops).

PR #525, opened 2015-09-05
s/sleep-till/sleep-until/ and coerce arg to Instant

ab5tract:
I'm not sure whether 'sleep-till' was meant to mirror some Perl 5-ish way of writing it, but it seems quite unnecessary.

### DIFF
--- a/src/core/Temporal.pm
+++ b/src/core/Temporal.pm
@@ -729,11 +729,6 @@ sub sleep-until (Instant() $until --> Bool) {
     True;
 }
 
-sub sleep-till (Instant $till --> Bool) {
-    DEPRECATED('sleep-until', |<2015.09 2015.09>);
-    sleep-until($till);
-}
-
 # =begin pod
 #
 # =head1 SEE ALSO
--- a/src/core/Temporal.pm
+++ b/src/core/Temporal.pm
@@ -721,12 +721,16 @@ sub sleep-timer (Real $seconds = Inf --> Duration) {
     }
 }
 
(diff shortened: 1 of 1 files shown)


lizmat:
I think the rename is not needed...  but maybe TimToady can be convinced.  In any case, at this point, we wouldn't deprecate, I don't think.

I think the coercion warrants keeping...

ab5tract:
[this might become a double post, but the web is working at the hotel again and the email did not show up]

On the contrary, as we saw in ptc++'s lightning talk, the description of how this sub works is "it sleeps until..." If it had documentation, I would expect it to be written this way as well. When someone thinks "I would like this to sleep until", the discoverability of sleep-until resolves the annoyance of first trying the reasonable version (sleep-until) and then having to chase down the cute version (sleep-till).

The 'till' is (IMO) awkward and overly English-slang-y. Yes, this is a very "minor" issue, but the more of these cutesy, not-actually-the-way-people-talk-or-discuss we have in core, the less consistent it feels.

I know I can't convince everyone for all of the consistency issues I raise, but this one seems particularly suited for change because a) it is not a heavily used function, and b) there is only one character difference but a significant increase in discoverability.

lizmat:
Well, I'm leaving that to TimToady to decide: about 2 years ago sleep-till was explicitely specced in fcd17ad8, and implemented in f7f3bb82e9aac6a0dc7f

lizmat:
Implemented with 781c02cf4c17d6f7b8a, ab5tract++

PR #526, opened 2015-09-05
Removed capture_and_parcel and its callers

cognominal:


lizmat:
Shouldn't we not just change the name from Parcel to List ???

cognominal:
Yes, one must go further. I am not sure for the next one :

SIG_ELEM_IS_PARCEL should be renamed SIG_ELEM_IS_LIST or SIG_ELEM_IS_SEQ?

lizmat:
Leaving that to jnthn

cognominal:
Not needed anymore. TimToady pushed a better patch :
https://github.com/rakudo/rakudo/commit/e0fc14d37cac3b6be2c769db9c302148e4601714

PR #527, opened 2015-09-05
fix post-glr install /bin wrapper

ugexe:


lizmat:
Shouldn't that be just | instead of flat ?

ugexe:
`s/flat/|/` does not work

PR #529, opened 2015-09-11
Add ddt which is like dd with type.

stmuk:
eg.

> my $foo=1
1
> ddt $foo
Int $foo = 1


lizmat:
Usurped with 49e9745dadd01bf65847 , thanks for the inspiration!   :-)

PR #535, opened 2015-09-16
Process and bind subsignature before running where clauses (RT#123596)

skids:
It seems ordering of the code chunks only affects the unwanted behavior.

I don't know if all where clauses hit the bootstrap binder, so this
might not be a complete fix; consult some guts folks before closing ticket.

### DIFF
--- a/src/Perl6/Metamodel/BOOTSTRAP.nqp
+++ b/src/Perl6/Metamodel/BOOTSTRAP.nqp
@@ -446,6 +446,41 @@ my class Binder {
             nqp::bindkey($lexpad, 'self', nqp::decont($oval));
         }
 
+        # If it has a sub-signature, bind that.
+        my $subsig := nqp::getattr($param, Parameter, '$!sub_signature');
+        unless nqp::isnull($subsig) {
+            # Turn value into a capture, unless we already have one.
+            my $capture;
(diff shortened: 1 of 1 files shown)


FROGGS:
I think that also needs to be done in the Binder.java.

jnthn:
Also to check there's consensus this is the right thing, because I'm not
convinced it is. Now you can do a where before the sub-sig so you don't go
unpacking things that don't meet constraints, and a where inside the
sub-sig that asserts things on the unpack. With this patch we'd lose being
able to do the first.

skids:
One arguments in favor of changing it is that the subsignature appears visually before the where clause, so it is natural to expect it to work.

But jnthn's is a pretty persuasive argument in that moving the checks inside the where can get the same effective results plus the ability to prevent the inside clauses from running via an external clause.

patzim:
I just submitted this to [problem-solving](https://github.com/perl6/problem-solving/issues/24) to get consensus on the desired behaviour. Once that's clear we can close the bug report and either merge or close this PR.

JJ:
Conflicts need to be cleared first...

JJ:
Since this is already in the problem-solving repo, I'm closing this PR. It can be reopened, salvaged or a new one can be created when that's cleared.

PR #539, opened 2015-09-18
Delay attributive parameter binding ops until after bind will succeed

skids:
This fixes one of the cases in [RT#125437](https://github.com/Raku/old-issue-tracker/issues/4332).  Also a flag is now provided that
could be used if we can identify certain types of bind operations
which should not implicitly alter self.

Note that $!param inside a where clause will return self's current value
after this patch, and also note that the patch does not protect against
explicit side-effect-modification of self's attributes in where
clauses during a failed bind.  If we can make a temporary lexpad shadow of
. or !-twigiled variables for the where clause lexpad, I don't know how
(and am not sure we'd want to, anyway.)

### DIFF
--- a/src/Perl6/Actions.nqp
+++ b/src/Perl6/Actions.nqp
@@ -6932,6 +6932,7 @@ Compilation unit '$file' contained the following violations:
         my @p_objs := nqp::getattr($sig, $Sig, '$!params');
         my int $i  := 0;
         my int $n  := nqp::elems(@params);
+        my @attrs;
         while $i < $n {
             # Some things need the full binder to do.
             my %info      := @params[$i];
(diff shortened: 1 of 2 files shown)


JJ:
Can you please try and fix the conflicts, and run tests again?

JJ:
It refers to [this issue](https://github.com/Raku/old-issue-tracker/issues/4332)

JJ:
Can someone take a look at this?

AlexDaniel:
So the original ticket and this PR are about side-effects in `where` clauses, specifically about changing the object itself. It is more or less common knowledge that `where` blocks can fire at any moment (and more times than you expect) and you shouldn't rely on them for side effects ever.

It hurts? Don't do it.

I'm not sure how this PR can resolve anything, but I guess it does help in some specific cases?

lizmat:
This feels like it will be moot once the RakuAST branch lands.  I will therefore close this PR in the coming days unless someone objects.

PR #540, opened 2015-09-25
heredoc related fixmas list work

skids:
Proposed fixes for a couple of RTs in the blockers list.


lizmat:
I see two fails:

t/spec/S32-str/indent.t                                     (Wstat: 512 Tests: 62 Failed: 2)
  Failed tests:  43, 54
  Non-zero exit status: 2

is that to be expected?

skids:
Yes, the semantics change for dedent and there are two tests that assume spaces will be deleted from the right.  Masak said in the ticket he wanted to review any proposed patches from a language level so I'll hit him up to do so on IRC.

PR #543, opened 2015-09-29
One more fixmas indentation fix and improvement on previous fix

skids:


skids:
Worth noting is that $?TABSTOP seems destined to be a (localizable)
compile-time constant, which is necessary for heredoc, but not so great
for users of .indent who may want to change behavior at runtime.

zoffixznet:
*ping* does this still need to go in or can this PR be closed?

There are some conflicts.

JJ:
Well, there were conflicts already in 2016. I don't think this can be fixed, closing.

AlexDaniel:
This PR claims to resolve RT#125543 but it is already closed: https://github.com/Raku/old-issue-tracker/issues/4378
So I guess there's no need in this indeed.

PR #544, opened 2015-10-01
fix gen-usage for getting correct basename

mattn:
On windows, filenames in usage always become absolute paths.

$*SPEC.path() may contains "." on windows. So need to correct path to absolute. And IO.x won't be true on filenames which don't have executable file extension like exe. It should be checked with case insensitive on windows or macosx.

### DIFF
--- a/src/core/Main.pm
+++ b/src/core/Main.pm
@@ -45,13 +45,20 @@ my sub MAIN_HELPER($retval = 0) {
     my sub gen-usage() {
         my @help-msgs;
 
+        my $is-win = $*DISTRO.is-win;
+        my $is-insensitive = $is-win or $*DISTRO ~~ /^macosx/;
         my sub strip_path_prefix($name) {
             my $SPEC := $*SPEC;
             my ($vol, $dir, $base) = $SPEC.splitpath($name);
             $dir = $SPEC.canonpath($dir);
             for $SPEC.path() -> $elem {
-                if $SPEC.catpath($vol, $elem, $base).IO.x {
-                    return $base if $SPEC.canonpath($elem) eq $dir;
+                my $io = $SPEC.join("", $SPEC.rel2abs($elem), $base).IO;
+                if $io.x or ($is-win and $io.e) {
+                    if $is-insensitive {
+                        return $base if lc $SPEC.canonpath($elem) eq lc "$vol$dir";
+                    } else {
+                        return $base if $SPEC.canonpath($elem) eq "$vol$dir";
+                    }
                     # Shadowed command found in earlier PATH element
                     return $name;
                 }


lizmat:
Technically, it's the file system that determines case-insensitivty,  On OS/X it's the default, but it can be changed.  Even for a boot device.

Other than that, checking for OS/X is easier: $*DISTRO.name eq 'macosx'

mattn:
@laben then, how about to remove all prefixes from the path? I don't understand why doing checking on this part. `pl2bat` remove all prefixes like below.
```perl
$0 =~ s|.*[/\\]||;
```

JJ:
There are conflicts with the current version. Also, not clear there is a consensus this is needed. So unless conflicts are fixed *and* some consensus reached, this PR might be closed in the near future.

AlexDaniel:
@JJ conflicts are never an issue, just resolve them.

JJ:
@AlexDaniel I don't seem to have privileges for that, or even see them... 
Captura de pantalla de 2020-05-03 12-53-17

JJ:
OK, had a look at this. The conflict can't be solved, because code has drifted, but code can be salvaged. However, I'm not sure this is an issue; there's no issue complaining about this anywhere but here. So can someone like @AlexDaniel or even @lizmat take a look and some action on it?

src/core/Main.pm
```diff
@@ -45,13 +45,20 @@ my sub MAIN_HELPER($retval = 0) {
     my sub gen-usage() {
         my @help-msgs;
 
+        my $is-win = $*DISTRO.is-win;
+        my $is-insensitive = $is-win or $*DISTRO ~~ /^macosx/;
```
AlexDaniel, line 49: > there are both case sensitive and case insensitive variations on HFS

So this line is just wrong?

AlexDaniel: Moreover, this is using `$*DISTRO ~~ /^macosx/` in sink context (use `&&` instead).

src/core/Main.pm
```diff
@@ -45,13 +45,20 @@ my sub MAIN_HELPER($retval = 0) {
...
             my $SPEC := $*SPEC;
             my ($vol, $dir, $base) = $SPEC.splitpath($name);
             $dir = $SPEC.canonpath($dir);
             for $SPEC.path() -> $elem {
-                if $SPEC.catpath($vol, $elem, $base).IO.x {
-                    return $base if $SPEC.canonpath($elem) eq $dir;
+                my $io = $SPEC.join("", $SPEC.rel2abs($elem), $base).IO;
+                if $io.x or ($is-win and $io.e) {
+                    if $is-insensitive {
+                        return $base if lc $SPEC.canonpath($elem) eq lc "$vol$dir";
```
AlexDaniel, line 58: I guess `fc` will be better here?

PR #549, opened 2015-10-12
Fix MAIN's usage message in case of slurpy hash parameters

cygx:
A parameter `*%flags` will now appear as `[--<flags>=...]` in the message.


zhuomingliang:
Hello, I would like to get this merged. Thanks.

PR #550, opened 2015-10-14
Fix bugs in split(:all)

cygx:
Splitting on regex delimiters returned Match objects instead of strings, splitting on Cool delimiters lacked a Slip.


cygx:
Note: I haven't even compiled the changes, so please do so before applying the patch :p

cygx:
Returning match objects is apparently [by design](http://design.perl6.org/S32/Str.html#line_264), so scratch the first part of the request...

PR #558, opened 2015-10-23
Add IO::Spec::Any

Leont:
In the old design an IO::Spec::Win32 isa IO::Spec::Unix, this violation of the Liskov substitution principple causes various issues. This patch adds a IO::Spec::Any role containing the shared interface, as well as shared implementations. I would have preferred to add this to IO::Spec itself, but that is not possible while it's exposing a (documented) class method (select).

### DIFF
--- a/src/core/IO/Spec/Any.pm
+++ b/src/core/IO/Spec/Any.pm
@@ -1,0 +1,81 @@
+my role IO::Spec::Any {
+
+    method canonpath { ... }
+
+    method dir-sep  { ... }
+    method curdir   { '.'  }
+    method updir    { '..' }
+    method curupdir { none('.','..') }
+    method rootdir  { ... }
+    method devnull  { ... }
+
+    method basename { ... }
+
+    method extension(\path) {
+        my str $str = nqp::unbox_s(path);
+        my int $index = nqp::rindex($str,'.');
+        nqp::p6bool($index == -1)
+          ?? ''
+          !! substr(path,nqp::box_i($index + 1,Int) );
+    }
+
+    method is-absolute { ... }
+    method path { ... }
+    method splitpath { ... }
+    method splitdir { ... }
+    method split { ... }
+    method join { ... }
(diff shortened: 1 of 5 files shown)


labster:
I think that just adding these to IO::Spec would be the better choice, and also the correct choice.  S32::IO notes:

    This set of modules was inspired by Perl 5's C<File::Spec>.  An implementation
    may choose to inherit from C<IO::Spec>, or any of its subclasses, if that
    helps in avoiding code duplication.

So, just add the methods to IO::Spec, and fix the following line of the spec ("The select method is the only method provided by IO::Spec itself.") which is obviously incompatible with the line above.

Is there a reason why IO::Spec can't provide a select method as well as define the interface for IO::Spec subtypes?  Like File::Spec, there's no need to instantiate the classes, so an extra method won't make subtypes noticeably heavier.

Leont:
> Is there a reason why IO::Spec can't provide a select method as well as define the interface for IO::Spec subtypes? Like File::Spec, there's no need to instantiate the classes, so an extra method won't make subtypes noticeably heavier.

When calling a method on a role it gets punned, but that isn't possible on abstract roles

Leont:
Actually, getting rid of all abstract methods may also be a solution, but that just feels wrong to me.

My prefered solution would be to get rid of IO::Spec.select, really.

zoffixznet:
Thanks for the work, however, doesn't look like this will be changed at this point, 1.5 years after the release. My [brief attempt at adding suggested changes](https://gist.github.com/zoffixznet/59e87483894ef8efe00ad38523995646) doesn't compile because Win32 spec wants more methods from Unix. Moreover, the QNX spec is Unix with a single method overriden, needing more design changes that probably can't be made due to 6.c language spec frozen now.

FWIW, the whole `IO::Spec::*` stuff today is treated as low-level, you-shouldn't-really-touch-this stuff and there's some motion to remove it from core entirely in future language versions.

PR #565, opened 2015-10-28
Try harder to diagnose problem instead of reporting undeclared type

usev6:
Fixes RT #125902

### DIFF
--- a/src/Perl6/Grammar.nqp
+++ b/src/Perl6/Grammar.nqp
@@ -2513,8 +2513,22 @@ grammar Perl6::Grammar is HLL::Grammar does STD {
           <DECL=multi_declarator>
         | <DECL=multi_declarator>
         ]
-        || <.ws>[<typename><.ws>]* <ident> <?before <.ws> [':'?':'?'=' | <.terminator> | $ ]> {}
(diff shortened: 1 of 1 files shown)


AlexDaniel:
Can we have this rebased to resolve the conflict? Ping @usev6

usev6:
I'll test my changes with current nom and push an updated version.

AlexDaniel:
I see no problem with this PR. It may be over-engineered as you say, but it gives awesome error messages. I was able to reproduce all of them:

```perl6
my Str where ‘foo’ $test;
```
```
Malformed my (found type followed by constraint; did you forget a variable in between?)
at -e:1
------> my Str ⏏where ‘foo’ $test;
```

```perl6
my Int is default(42)
```
```
Malformed my (found type followed by trait; did you forget a variable in between?)
at -e:1
------> my Int ⏏is default(42)
```

```perl6
my Str;
```
```
Malformed my (did you forget a variable after type?)
at -e:1
------> my Str⏏;
```

Without this PR it simply says “malformed my” which is arguably LTA. Spectest clean, merging.

jonathanstowe:
:+1:

PR #567, opened 2015-10-30
Add :close argument to IO::ArgFiles.lines iterator

lucasbuchala:
A possible fix for https://rt.perl.org/Public/Bug/Display.html?id=126494


lizmat:
Have you verified that that solves the problem?

lucasbuchala:
Only in a superficial test, yes.

```
$ cat /tmp/foo
foo
```

```
$ perl6 -e '.say for lines' /tmp/foo /tmp/foo /tmp/foo 
foo
foo
foo
```

Without this, P6 hangs on the second "foo".

PR #568, opened 2015-10-31
teach EVAL about nqp

cygx:
The language is called lower-case `nqp` as that's what `nqp::getcomp` expects.

~~It might be a good idea to add another candidate that dispatches on `:lang<NQP>` for consistency with with `use :from<NQP>`...~~

For consistency, the spelling `NQP` is also allowed

### DIFF
--- a/src/core/control.pm
+++ b/src/core/control.pm
@@ -216,7 +216,7 @@ proto sub EVAL(Cool $code, Str() :$lang = 'perl6', PseudoStash :$context, *%n) {
     $compiled();
 }
 multi sub EVAL(Cool $code, Str() :$lang! where 'nqp' | 'NQP', *%_) {
-    nqp::loadbytecode("nqp.{ $*VM.precomp-ext }");
+    once nqp::loadbytecode("nqp.{ $*VM.precomp-ext }");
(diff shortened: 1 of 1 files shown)


lizmat:
Letting this one for jnthn to look into

src/core/control.pm
```diff
@@ -215,6 +215,10 @@ proto sub EVAL(Cool $code, Str() :$lang = 'perl6', PseudoStash :$context, *%n) {
     nqp::forceouterctx(nqp::getattr($compiled, ForeignCode, '$!do'), $eval_ctx);
     $compiled();
 }
+multi sub EVAL(Cool $code, Str() :$lang! where 'nqp', *%_) {
```
vendethiel: the where should probably follow the sub below with `$lang // ''`? or maybe it should be changed to use `:$lang!`?

cygx:
Turns out there is the potential for fallout: Calling `EVAL` at `BEGIN`-time can mess up precompilation.

I have not had any issues using it at run-time.

JJ:
Can you fix conflicts before we go any further?

lizmat:
Looking at the code, it looks like the EVAL would be recursing infinitely?

PR #572, opened 2015-11-04
Let NativeCall map `int` to `longlong`

cygx:
This gives the correct result on Win32 - the former mapping to `long` resulted in a 32-bit type being used despite `nativesizeof(int)` returning 8.

Needs to be checked for fallout in roast and possibly the ecosystem.


FROGGS:
One should not use `int` in NativeCall code anyway... Every `int` there is error prone.

I'd favour if we would disallow `int` and `num` in NativeCall routine signature and attribute declarations.

cygx:
While `int` in nativecall declarations might be somewhat confusing, mapping it to the same thing as p6-land native `int` would be consistent - in normal situation, `int` would just be an alias for `int64` and `num` for `num64`.

On standard architectures, this patch should achieve this.

On the other hand, the current situation in untenable as values do not round-trip (which is how I discovered the issue in the first place).

cygx:
If decision-making were up to me, I'd just specify that `int` and `int64` (as well as `num` and `num64`) alias ech other, and give the C types a `c...` prefix, ie `int = int64 = cllong` and `int32 = cint` on non-exotic platforms.

FROGGS:
Is i386 not mainstream anymore?

FROGGS:
No, int might very well mean int128 on some platforms... Like it means int32 and int64 already.

cygx:
@FROGGS right.  But anything other than

    cchar = int8
    cshort = int16
    cint = int32
    clong = int32|int64
    cllong = int64

would be 'exotic' and isn't currently supported anyway.

Which type a raw `int` maps to would be a configure-time option.

cygx:
TL;DR:

  * `int` should *always* refer to a p6 native `int`, which means we need another name like `cint` to refer to the C-level type.

  * the other C-level types should follow the same naming convention for consistency

  * any correspondence between these 'logical' and the fixed-sized integer types is incidental

FROGGS:
Are there any platforms still in existance where C's `int` does not mean `int32`?

cygx:
Mostly just mainframes (yes, they still exist) and DSPs

Skarsnik:
I think the c prefix on NC type could make everything less confusing when writing binding.

AlexDaniel:
This was reverted in https://github.com/rakudo/rakudo/commit/c362bd8295.

AlexDaniel:
Please discuss this further in this ticket: https://github.com/Raku/problem-solving/issues/182

PR #573, opened 2015-11-04
Add support for 1^..^10 non inclusive range quantifiers in regexs

ungrim97:
A recent change in NQP https://github.com/perl6/nqp/pull/249 added support for non inclusive range syntax as quantifiers in a regex. Rakudo overrides the action method that provides that functionality. As such that feature needs to be ported to rakudo as well.


ungrim97:
The spec tests for this have been merged https://github.com/perl6/roast/pull/73

PR #578, opened 2015-11-07
Switch to pure-perl6 TAP harness

Leont:
This patch switches the test harness from a perl 5 based one to a perl6 based one. Currently still WIP because of a two major issues that probably should be fixed first:

1. Parallelization is buggy; I'm often lacking some of the output of the program. I suspect this is an issue in moarvm's Proc::Async implementation, but it might also be in my code (though I really think I'm doing the correct dance). I hope to be able to come up with a minimal test case.

2. The eval-server/client on JVM isn't working. This is probably a small mistake in my code, but I haven't gotten around to finding/fixing it yet. (haven't enabled it on jvm yet because of this, it seems to work without eval-server but is unusably slow)

Feedback is most welcome


jonathanstowe:
:+1:

Leont:
> The eval-server/client on JVM isn't working.

Actually, the client side is probably better implemented as a SourceHandler than as an external program, that would also get rid of another perl 5 dependency.

magnolia-k:
:+1:

niner:
What advantages will this have over the current test harness?

lizmat:
nine:
- dogfooding, for one!
- one step closer to not needing P5 to run P6
- potentially better throughput for testing in the future

jnthn:
Also, potentially better throughput on parallel tests on Windows already, though I'd need to check that.

Leont:
> * dogfooding, for one!
> * one step closer to not needing P5 to run P6

Dogfooding was my prime motivation, though the P5 dependency has another dimension too. The original target was panda not rakudo; a runtime dependency on perl5 is a much bigger issue than a buildtime dependency. Later I realized it made more sense to fix it at the root first so I ported rakudo's harness to it too (which was also excellent for finding a few missing features and even a parsing bug).

Leont:
I've rebased this all. Concurrent testing still doesn't work, but with very different errors.

When run with `TEST_JOBS=1`, I get errors like this:
~~~~
Unhandled exception in code scheduled on thread 7
Unhandled exception in code scheduled on thread 4
Can only perform big integer operations on concrete objects
~~~~
(both messages always appear at least once, sometimes they appear twice)

Adding --ll-exception doesn't add any information, and there doesn't seem to be a backtrace anymore at that point. This smells like a serious bug.

Leont:
On my machine, synchronous parsing seems to work fine, though it's still hitting a rakudobug every 200 `run()` calls. Asynchronous runs are still not quite stable yet, for reasons that I suspect are on the rakudo/moar side (but I haven't ). And there seems to be a minor incantation issue on Windows that needs someone with more Windows clues than me.

To move beyond, I've changed this patch to support both p5 and p6 harness, defaulting to the latter (so `make spectest HARNESS_STYLE=6` will run with my harness). This should make this mergeable without breaking anything.

moritz:
This broke running a single test file through make:

    moritz@pete:~/p6/rakudo$ make t/spec/S06-other/main-usage.t 
    rm -f -- perl6
    cp -- perl6-m perl6
    chmod -- 755 perl6
    verbosity=1 t/spec/S06-other/main-usage.t
    sh: 1: t/spec/S06-other/main-usage.t: Permission denied
    Makefile:609: recipe for target 't/spec/S06-other/main-usage.t' failed
    make: [t/spec/S06-other/main-usage.t] Error 126 (ignored)

Seems it tries to invoke the test file directly, not rakudo

Leont:
Seems I missed that case, and I can't easily make it switch between harnesses but the fix is otherwise just a oneliner.
~~~~
commit 67b6491ba63256927f948642ac6d373060d2b3e0
Author: Leon Timmermans <fawaka@gmail.com>
Date:   Sun Apr 24 22:00:45 2016 +0100

    Fix make t/foo.t

diff --git a/Configure.pl b/Configure.pl
index f2d02da..6c50a1d 100644
--- a/Configure.pl
+++ b/Configure.pl
@@ -309,7 +309,7 @@ MAIN: {
     sorry(@errors) if @errors;
 
     my $l = uc substr($default_backend, 0, 1);
-    print $MAKEFILE qq[\nt/*/*.t t/*.t t/*/*/*.t: all\n\t\$(${l}_HARNESS_WITH_FUDGE) --verbosity=1 \$\@\n];
+    print $MAKEFILE qq[\nt/*/*.t t/*.t t/*/*/*.t: all\n\t\$(${l}_HARNESS5_WITH_FUDGE) --verbosity=1 \$\@\n];
 
     close $MAKEFILE or die "Cannot write 'Makefile': $!";
~~~~

PR #585, opened 2015-11-15
improve permutations

grondilu:
using @permutations avoids computing permutations($n - 1) several times


moritz:
Applied as 57b83f4ad7803bb2dc96313c42636229ca76bf6d, thanks!

PR #594, opened 2015-11-19
Import a new CREDITS based on a new script CREDITS.p6

stmuk:


vendethiel:
From a quick glance, grondilu appears twice, and lue thrice

stmuk:
@vendethiel I will manually fix

stmuk:
see #595

PR #596, opened 2015-11-20
use a Seq for permutations

grondilu:
use a proper iterator, with the lexicographic method.  See:  L<https://en.wikipedia.org/wiki/Permutation#Generation_in_lexicographic_order>


lizmat:
Are you aware of the :last named param for first?  Seems we don't need a reverse then?  Or is that to make it lazier?

grondilu:
No, I was not aware of the :last named param.  I'll update the patch.

zhuomingliang:
Any chance to merge this? @lizmat

PR #598, opened 2015-11-28
Avoid NullPointerException on rakudo.jvm

usev6:
'say my @a' lead to NPE, timotimo++ for suggesting nqp::existskey


peschwa:
I'm unclear what the better reaction to an nqp::atkey call without verifying key existance beforehand would be, but it's agreed that checking first should be done anyway.

PR #603, opened 2015-11-29
avoid installing %?LANG too early

LLFourn:
%*LANG was copied to %?LANG too early causing EVAL not work sometimes (because it was copied before actions were applied).


LLFourn:
tests https://github.com/perl6/roast/pull/80

PR #604, opened 2015-11-29
add a candidate for infix:<**>(\a, 2)

grondilu:
hopefully that would allow things like i**2 == -1 to return a sensible result


grondilu:
sorry this actually does not work.

PR #609, opened 2015-12-02
RT#126785: ensure make *test uses -Ilib rather than preinstalled libs

ungrim97:


niner:
This can't be right. $(PERL) is the perl 5 executable while lib contains Perl 6 libs. perl -Ilib -e 'use Test;' fails accordingly.

ungrim97:
This ensures that the right libs are passed to tap harness which seems happy to pass them on when it runs:

/home/mfrancis/dotfiles/anyenv/envs/plenv/versions/5.22.0/bin/perl5.22.0 -Ilib t/harness --moar t/01-sanity t/04-nativec
all

ungrim97:
Hmm looking at t/harness however it would seem that this is already being done by PERL6LIB on line 44?

jnthn:
Wouldn't this require you to have to "make install" before you can "make
test"?

PR #611, opened 2015-12-03
Alternative to peppering the filesystem with .precomp dirs

flussence:
This'll hopefully solve the problem @ugexe had earlier today (http://irclog.perlgeek.de/perl6/2015-12-02#i_11644081). I admit it's not pretty code by any means, but it seems like an improvement over "not working at all" in that one case.

There are some issues with the locking in `CompUnit::PrecompilationStore::File` being too coarse-grained; this change exacerbates them pretty badly. Only one perl6-m process can be compiling at a time _on the entire OS_ (sorry), and if one of them deadlocks (that's a pre-existing problem, try `panda install Hinges`) they all get wedged. That's probably fixable but I don't have enough tuits to go poking around in low-level locking code.

I've left a safety valve in, in case it needs to be disabled for any reason: put a plain file where it expects the "perl6" cache dir to be and it'll fall back to the existing .precomp behaviour.

### DIFF
--- a/src/core/CompUnit/Repository/FileSystem.pm
+++ b/src/core/CompUnit/Repository/FileSystem.pm
@@ -110,7 +110,10 @@ class CompUnit::Repository::FileSystem does CompUnit::Repository::Locally does C
         $!precomp := CompUnit::PrecompilationRepository::Default.new(
             :store(
                 CompUnit::PrecompilationStore::File.new(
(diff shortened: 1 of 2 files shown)


flussence:
Withdrawing this PR for now to come up with a better plan; I've been hitting write-permission errors when moar/nqp are installed to system dirs in spite of this patch, so I think it's trying to fix the problem at the wrong layer.

PR #613, opened 2015-12-03
Fix RT #126759

LLFourn:
create a sneaky lexical EXPORTHOW package so if you do:

```
use MyDeclarator;
EVAL 'my-declarator my-class { }'
```

it will work because my-declarator is available in %*HOW in EVAL

Fixes the RT and makes the fudged tests here pass: 
https://github.com/perl6/roast/blob/master/S12-meta/exporthow.t

### DIFF
--- a/src/Perl6/World.nqp
+++ b/src/Perl6/World.nqp
@@ -587,6 +587,12 @@ class Perl6::World is HLL::World {
                     }
                 }
             }).eager;
+
+            # copy modified %*HOW into lexical EXPORTHOW symbol so new/superceded
+            # declarators will work inside EVAL.
+            my $lex_EXPORTHOW := nqp::knowhow().new_type(:name('EXPORTHOW'));
+            nqp::setwho($lex_EXPORTHOW,self.p6ize_recursive(%*HOW));
+            self.install_lexical_symbol(self.cur_lexpad,'EXPORTHOW',$lex_EXPORTHOW);
         }
     }
 


AlexDaniel:
https://github.com/Raku/old-issue-tracker/issues/4797

niner:
Are there any tests for this? Since the PR was made, we made module loading (or rather symbol importing) lexical. It may be that while it still compiles, it doesn't actually work as intended anymore.

Ironically, I came across the same issue recently, so I know that the issue is still there and understand what this PR is supposed to fix. I kinda sorta have an idea what the proposed solution does. I don't know if it's the best solution, but I'd take it anyway if there are tests that show it works.

LLFourn:
Hi Rakudo squad! Good feelings come from seeing that your five year old PR might still be useful :tada:.

> Are there any tests for this?

I added this test for this PR it seems: https://github.com/Raku/roast/blob/212042d57830f86886cc4d41ffe44ee501db22d9/S12-meta/exporthow.t#L19-L26

lizmat:
Alas, my merge was premature: it breaks `Inline::Perl5`:

===SORRY!===
Representation for EXPORTHOW must be composed before it can be serialized

And also it unskipping the tests in roast breaks testing.  :-(

Reverting.

LLFourn:
Damn. RIP old PR. You did your best.

lizmat:
@LLFourn yeah, sorry to see that.  On the brighter side, @jnthn is now assigned to the ticket, as his work on dispatchers and RakuAST probably needs to re-visit this problem space before long.

PR #616, opened 2015-12-03
Update Makefile-common.in to work with mingw

ronaldxs:
Fix "/usr/bin/sh: C:Strawberryperlbinperl.exe: command not found" type errors under windows the same way we did for nqp about a week ago by effectively setting the SHELL to cmd under windows and 'sh' for posix environments.


zhuomingliang:
I'm sure this will fix the build with mingw, so I would like to merge this one, thanks.

PR #617, opened 2015-12-04
Improve X::Syntax::Adverb error message

MadcapJake:
Previously strings would just get a `You cannot adverb that` which could be seen as implying that the actual name of the adverb `:foo` could not be "adverbed".

If you now have an adverb following a string (or anything that doesn't have a `name`), it will print the string e.g., `You cannot adverb 'foo'`.

I think this could probably remove [Exception.pm#L1403](https://github.com/rakudo/rakudo/blob/nom/src/core/Exception.pm#L1403)'s `// "that"` too, right?


MadcapJake:
I removed `// "that"` and ran `make test` (passed) and `make spectest` (seems to be all the same current failing tests).

PR #619, opened 2015-12-05
Some IO-related cleanup and an actually useful binary mode

cygx:
In particular, this removes `&pipe` (which was defunct as the backend method [had already been removed](https://github.com/rakudo/rakudo/commit/03b5a0abb58905f8e8e6e768186696f10e37729e)).

It also adds the ability for binary line-by-line access when opening files in mode `:bin` that previously had been [specced](http://design.perl6.org/S32/IO.html#line_114) and [documented](http://docs.perl6.org/routine/open#Encoding_options), but apparently never been implemented.


cygx:
This will be superseded by a more comprehensive bin mode refactor I'm working on.

PR #633, opened 2015-12-13
make some Cool methods on IO::Path fail instead of returning a handled failure

cygx:
Submitted via github web UI, **untested**


cygx:
Would it be better to remove the `CATCH` blocks so that the failure returned by `.open` will throw? That way, we would fail fast even if we bind the return value...

cygx:
superseded by #643

PR #635, opened 2015-12-16
Add fast .elems to .combinations

zoffixznet:
This PR lets us evaluate the total number of `.combinations` without actualy calculating them all. This mirrors doing so with `.permutations`.

I believe we can get rid of `( ).Int` if we change `    return ((),) if k < 1;` on line 10 to `    return ((),) if k < 1 or k > n`, but that's for another PR ATM :)

Relevant IRC discussion: http://irclog.perlgeek.de/perl6/2015-12-16#i_11730331

### DIFF
--- a/src/core/List.pm
+++ b/src/core/List.pm
@@ -41,6 +41,10 @@ my sub combinations(\n, \k) {
             }
             IterationEnd
         }
+
+        method count-only {
+            (  ([*] 1..$!n) / (([*] 1..$!k) * ([*] 1..$!n-$!k))  ).Int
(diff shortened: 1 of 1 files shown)


grondilu:
This is not the best way to compute combinations, as it will compute the whole factorials which can be stupidely large.

I suggest:

    ([*] ($n ... 0) Z/ 1..$k).Int

as in RosettaCode:  http://rosettacode.org/wiki/Combinations

lizmat:
replaced by #637

PR #638, opened 2015-12-17
Add support for subseconds support in ISO 8601, see #126948

Skarsnik:


niner:
While a step in the right direction, there is more. From Wikipedia:
"A decimal mark, either a comma or a dot (without any preference as stated in resolution 10 of the 22nd General Conference CGPM in 2003,[14] but with a preference for a comma according to ISO 8601:2004)[15] is used as a separator between the time element and its fraction."

PR #639, opened 2015-12-18
Add support for specifing the api/abi version to native (NC) routine

Skarsnik:
Like I said in RT 126645 having a way to specify the api (abi?) version is nice on some unix as it prevents to install the devel package to have the  .so (according to doc linked in the RT it should not be used).

What the patch does?
You can declare your routine like this sub foo is native('libfoo:4.2')
I had the idea of having is native('libfoo' , v4.2) but I am not really a fan
I tried being the less intrusive possible as you can see looking at the test.

On win32
This should do nothing.

On unix
if (and only) you write the lib name without something that look like path/full file name it will replace libfoo with libfoo.so.version

On mac os X
libfoo will became libfoo.version.dynlib

Notice that NC will only complain about a missing API version if the lib is not a path

### DIFF
--- a/lib/NativeCall.pm6
+++ b/lib/NativeCall.pm6
@@ -159,7 +159,7 @@ my role NativeCallSymbol[Str $name] {
     method native_symbol()  { $name }
 }
 
-sub guess_library_name($lib) {
+sub guess_library_name($lib) is export(:TEST) {
     my $libname;
     if ($lib ~~ Callable) {
         $libname = $lib();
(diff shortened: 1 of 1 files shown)


jonathanstowe:
:+1: on this, it does seem odd requiring a -devel package on Linux when you're not actually compiling any code against it.

Skarsnik:
This should be removed. Using : under win32 is not a good idea. Etheir change to something else or chabge the code to not bother with it under win32

FROGGS:
@Skarsnik What are you replying to?

PR #640, opened 2015-12-18
Fix CLONE-HASH-DECONTAINERIZED(%(:a(Any))

ugexe:
Removes warning messages for `my %x = :a(Any); CLONE-HASH-DECONTAINERIZED(%x)` which is called anytime `run`, `shell`, or `QX` is used

`CLONE-HASH-DECONTAINERIZED` is currently only used on `%*ENV` to be passed to `nqp::spawn`, so stringifying `Any` to `Str.new` should be safe


ugexe:
note this fixes the warnings from panda

PR #645, opened 2015-12-21
RFC: implement basic binary file mode

cygx:
Makes line-by-line access work in terms of `Buf` as specced. Also lets `IO::Handle.getc` in binary mode return a single byte, same as the libc function.

### DIFF
--- a/src/core/IO/Handle.pm
+++ b/src/core/IO/Handle.pm
@@ -5,9 +5,15 @@ my class Proc { ... }
 my class IO::Handle does IO {
     has $.path;
     has $!PIO;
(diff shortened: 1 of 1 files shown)


JJ:
Code base has drifted away, can you please fix the conflicts before we run any other check?

lizmat:
It looks to me that conceptually this changes the semantics of the `$!decoder` attribute: before this, having a falsey value for `$!decoder` means that reading is in binary mode.  So to change this, would also mean changing where `$!decoder` is used as a boolean.  And that's a lot of places.

Feels to me this is the wrong approach.  So I will close this PR in a few days, unless someone objects.

PR #646, opened 2015-12-21
Small IO cleanup

cygx:
Removes defunct `&pipe`, simplifies `&open`.

### DIFF
--- a/src/core/io_operators.pm
+++ b/src/core/io_operators.pm
@@ -108,13 +108,6 @@ multi sub open($path, :$chomp = True, :$enc = 'utf8', |c) {
     $handle.open(:$chomp,:$enc,|c);
 }
 
-proto sub pipe(|) { * }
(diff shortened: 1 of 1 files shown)


zhuomingliang:
I'd like to see this get merged, thanks. :+1:

lizmat:
Until we have clarity on how we're going to handle version, it would seem wise not to change / remove stuff, especially like this.  So, not going to do that right now.

cygx:
@lizmat `&pipe` is backed by `IO::Handle.pipe` which was removed [on September 4th](https://github.com/rakudo/rakudo/commit/03b5a0abb58905f8e8e6e768186696f10e37729e#diff-836232d4364a7df1cf3ff9c1e022cf0bL34).

lizmat:
Removed pipe() with c0bb692

cygx:
now obsolete

PR #651, opened 2015-12-24
Untwine Pod

LLFourn:
tl;dr `=pod C<foo>` now becomes `[ FormattingCode ]` instead of `["",FormattingCode,""]`

After merge, merge:
1. https://github.com/perl6/roast/pull/90 (tests)
2. https://github.com/perl6/doc/pull/261

long story
I started out trying to fix stray empty strings being everywhere: 

https://rt.perl.org/Public/Bug/Display.html?id=126966

Turns out empty strings were a symptom of a _twine_ algorithm (every FormattingCode has to be _intertwined_ with a string). This implementation detail side effect then leaked out everywhere:
- In the tests: https://github.com/perl6/roast/pull/90 (depends on odd array indexes because of empty strings)
- in Pod::To::HTML: https://github.com/perl6/Pod-To-HTML/pull/9 (caused by empty strings)
- in the htmlifying code: https://github.com/perl6/doc/blob/master/htmlify.p6#L379
- it sorta meant it had to serialize strings that never got used

The purpose of this commit is to expunge this algorithm. Because it is a relatively large commit I did some due diligence and tested it on a pod file containing the concatenation of all the .pod in docs and then putting `$=pod.perl.say` at the end.

```
bash-3.2$ /usr/bin/time -l perl6 test.pod  | wc -c
       58.72 real        58.26 user         0.35 sys
 705699840  maximum resident set size
...
 3672024
bash-3.2$ /usr/bin/time -l p6 test.pod  | wc -c #patched
       59.07 real        58.29 user         0.54 sys
1181294592  maximum resident set size
...
 3663120
```

I also did a precompile with `--target=mbc` to see the difference:

```
1.5M 24 Dec 13:59 test.pod
6.8M 24 Dec 23:08 test.pod.moarvm
6.1M 24 Dec 23:03 test.patched.pod.moarvm
```

This shows:
1. This commit increases the amount of memory used during compilation from 700Mb => 1200Mb
2. that there are less characters in the output (due to less "" everywhere)
3. The patched one has a smaller image size due to less unnecessary serialization
4. They are roughly the same speed

Despite 1. I suggest merging this. It's not caused by my code (I think) but my code exacerbates whatever issue is causing rakudo to need 700Mb to compile a 1.5Mb pod file in the first place. I made a RT:
https://rt.perl.org/Public/Bug/Display.html?id=127020

Given we have precomp now :tada:  this patch only benefits real world non-compiling-large-amounts of pod-code-at-runtime stuff. Also from here it's easier to implement NYI tags like S<> T<> etc.

Oh and `=pod B< < B<foo> > >` will work instead of dying with LTA error now

Merry christmas!
:santa: :santa:

### DIFF
--- a/src/Perl6/Actions.nqp
+++ b/src/Perl6/Actions.nqp
@@ -834,26 +834,26 @@ class Perl6::Actions is HLL::Actions does STDActions {
     method pod_block:sym<delimited_code>($/) {
         my $config  := $<pod_configuration>.ast;
         my @contents := $<delimited_code_content>.ast;
-        my $twine   := Perl6::Pod::serialize_array(@contents).compile_time_value;
-        make Perl6::Pod::serialize_object(
-            'Pod::Block::Code', :contents($twine),
-            :config($config),
-        ).compile_time_value
+        @contents  := Perl6::Pod::serialize_array(@contents).compile_time_value;
+        make Perl6::Pod::serialize_object('Pod::Block::Code',
+                                          :@contents,:$config).compile_time_value
     }
 
     method delimited_code_content($/) {
-        my @t := [];
+        my @contents := [];
         for $/[0] {
             if $_<pod_string> {
-                nqp::splice(@t, Perl6::Pod::merge_twines($_<pod_string>), +@t, 0);
-                nqp::push(@t, $*W.add_constant(
+                nqp::splice(@contents,
+                            Perl6::Pod::pod_strings_from_matches($_<pod_string>),
+                            +@contents, 0);
+                nqp::push(@contents, $*W.add_constant(
                     'Str', 'str', ~$_<pod_newline>
                 ).compile_time_value);
             } else {
-                @t.push($*W.add_constant('Str', 'str', "\n").compile_time_value);
+                @contents.push($*W.add_constant('Str', 'str', "\n").compile_time_value);
             }
         }
-        make @t;
+        make @contents;
     }
 
     method pod_block:sym<paragraph>($/) {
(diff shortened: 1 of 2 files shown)


zhuomingliang:
It looks good, I'd like to see this get merged, thanks. :+1:

lizmat:
[Coke]: as release manager, what are your thoughts ?

LLFourn:
When I made this patch I didn't know about @ShimmerFairy  SUPERNOVA. IMO further efforts in pod refactoring should be abandoned in favour of the "pod as another language" solution.

Though I still recommend merging this patch so we can at least get the spectests right before trying to tackle the more long term solution. I have already updated all ecosystem things (that I could spot) including htmlify.p6 from docs to be compatible with this patch.

Since there is now some interest in this I am working on patch for htmlify.p6 to use precomp which will make the +50% compile time memory thing irrelevant. Precomp uses another process so RT #127020 will be confined to that and will not build up over the process.

LLFourn:
The docs PR:
https://github.com/perl6/doc/pull/334

Here are the benchmarks

run| time (s) |peak mem (MB)
-----|-----------|---------------
no-patch,no-precomp| 594|1117
no-patch,precomp 1st|634|770
no-patch,precomp 2nd|584|730
patch,precomp 1st| 631| 733
patch,precomp  2nd| 552|690

So with this patch + the precomp patch to `htmlify.p6` we reduce memory usage ~45% and speed up by ~8%. The speed increase here comes from having to process much less `""`.

It is only the case where you are using `EVAL slurp ...` that this patch makes things worse but that is because of RT #127020. 

And to top it of this patch is more correct (passes more tests) and leads to smaller 10% smaller precompiles for `.pod`.

coke:
I have no strong opinions either way; Before the next release we're going
to have to review everything on nom since the last release, regardless.

If 6.c spectest is clean with no changes, then no worries. If it requires
spectest updates, then making it part of 6.d is probably the way to go for
now.

We might have to revisit it after we decide on the release policy either
way.

On Sat, Jan 9, 2016 at 6:22 AM, lizmat <notifications@github.com> wrote:

> [Coke]: as release manager, what are your thoughts ?
>
> —
> <https://github.com/rakudo/rakudo/pull/651#issuecomment-170226772>.
>



-- 
Will "Coke" Coleda

LLFourn:
This certainly requires spectest updates, the changes are here:

https://github.com/perl6/roast/pull/90/files

LLFourn:
@coke I was thinking from what TimToady explained the other day:
`TimToady	6.c.1 will be minor tweaks for buggy or missing tests that don't really change the language (much)`
http://irclog.perlgeek.de/perl6/2016-01-14#i_11888757

This is more a candidate for 6.c.1 than 6.d because it fixes incorrect tests, and fits the "doesn't change the language (much)" criteria.

tbrowder:
I have some ideas for handling pod table cell formatting which I believe would be made much easier if this were merged.

softmoth:
This PR can be closed now, as it is handled in #1300 .

PR #652, opened 2015-12-26
I appreciate the thanks, but once is enough

autarch:


lizmat:
Will merge this, but not to break the pristine v6.c state  :-)

PR #662, opened 2015-12-29
RFC: make v5 pluggable

cygx:
For the foreseeable future, there will not be an official Perl5 re-implementation on top of Rakudo. Therefore, it's probably a good idea to add some hooks to let users decide what should happen on `use v5` instead of hardcoding loading the `Perl5` module.

The new system would be used like this:

```
use Slang::Perl5; # exports sub LOAD-V5(Version $) { ... }
say $*DISTRO.name; # still perl6

{
    use v5.42; # calls LOAD-V5(v5.42) which installs a Perl5 slang
    print $^O, "\n";
}
```

### DIFF
--- a/src/Perl6/Grammar.nqp
+++ b/src/Perl6/Grammar.nqp
@@ -1499,9 +1499,16 @@ grammar Perl6::Grammar is HLL::Grammar does STD {
         [
         | <version> [
                     ||  <?{ $<version><vnum>[0] == 5 }> {
-                            my $module := $*W.load_module($/, 'Perl5', {}, $*GLOBALish);
-                            $*W.do_import($/, $module, 'Perl5');
-                            $*W.import_EXPORTHOW($/, $module);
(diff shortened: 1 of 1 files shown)


JJ:
Again, there are some conflicts here. Can you please fix them before we do any further work on this?

lizmat:
Especially after the rename, it appears unlikely that anybody wants to spend time on implementing Perl in Raku.  I *definitely* will not do that.  Furthermore, Perl has turned out to be a moving target, so what version of Perl to implement.  Finally, `Inline::Perl5` is very close to what was intended with the `v5` project.  So I'm closing this PR.  Please re-open if disagreeing with this view.

niner:
And anyway since it only looks for a short name, there's still the possibility 
to distinguish multiple "Perl5" modules by auth, api and version. So in a 
sense, it's already pluggable.

PR #665, opened 2016-01-01
Disable or partially disable precompilation globally at runtime

ChoHag:
Comments requested on the variable name and its contents, neither of which I'm fully satisfied with.


lizmat:
I understand the code.  I do not understand the reason for wanting this.  Especially since we've been able to fix *many* issues with regards to pre-compilation in the past years.  I'm therefore thinking of closing this PR as there does not seem to be a reason for needing this.  Please correct me if I'm wrong :-)

PR #668, opened 2016-01-03
Bypass path attribute caching in CompUnit

ugexe:
* Fixes issues where installation of modules fails on the
first run but works on subsequent tries.
* Allows `rm -rf /path/to/repo` uninstallation of modules
while still being able to recreate the path structure with-
out "No writeable path found" errors.

`method prefix` fixes instances of `$path.mkdir unless $path.e`
that pass around an IO::Path where `!$path.e` or `!$path.w`
erroneously (`perl6 -e 'my $path = "non-existent".IO; say $path.e; $path.mkdir; say $path.e'` -> `False False`). This avoids adding stringification code throughout
the rest of the codebase (or adding .IO all over)

`.mkdir` in BUILD mimicks prior art in PrecompilationStore::File

Alternatively mkdir could (and probably should) break the cache

### DIFF
--- a/src/core/CompUnit/Repository/Installation.pm
+++ b/src/core/CompUnit/Repository/Installation.pm
@@ -4,16 +4,18 @@ class CompUnit::Repository::Installation does CompUnit::Repository::Locally does
     has $!precomp;
     has $!id;
 
-    submethod BUILD(:$!prefix, :$!lock, :$!WHICH, :$!next-repo) {
-        $!prefix.mkdir;
-    }
+    submethod BUILD(:$!prefix, :$!lock, :$!WHICH, :$!next-repo) { }
 
     method writeable-path {
         $.prefix.w ?? $.prefix !! IO::Path;
     }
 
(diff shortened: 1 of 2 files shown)


ugexe:
So what happened when you tested this to make you think that? When I tested it on a non-writable location it still gives the "No writeable path found," error

ugexe:
This seems to solve the original issue and the issue @niner brings up with the original PR. It can probably be rewritten with more appropriate method names. This may raise the question if `can-install()` should do a `mkdir` instead of putting it directly into the `install` method, but that would mean `can-install()` could return `False` to an external installer when it really can install; this PR assumes a call to `can-install()` is an appropriate time to prime the directory path. The other alternative I came up with involves a recursive check of `$.prefix` (something like (untested) `?$.prefix.parts.map({ state $p = $_; $p.IO.e && !$p.IO.w ?? (last and False) !! $p = $p.IO.child($_) }).all`, but i'm not sure thats a real representation of writeable.

PR #671, opened 2016-01-03
Use an uncached IO::Path for $.prefix

ugexe:


lizmat:
Why?

niner:
From the original commit message:
`method prefix` fixes instances of `$path.mkdir unless $path.e`
that pass around an IO::Path where `!$path.e` or `!$path.w`
erroneously. This avoids adding stringification code throughout
the rest of the codebase (or adding .IO all over)

niner:
Pushed as commit 8150686271af8e6b7fef3068ad6d3a06e735ec24

PR #673, opened 2016-01-04
Fix RT #126566

LLFourn:
Exporting multiple custom declarators made them clobber their changes to
`$*ACTIONS`. Now use `%*LANG<MAIN-actions>` to chain the changes.

Test: https://github.com/perl6/roast/pull/93
RT: https://rt.perl.org/Public/Bug/Display.html?id=126566

I have a module that I would like to release that depends on multiple custom declarators so I would be happy if this got merged. 

PS I have a wholistic patch that cleans up a lot of the mixing into LANG stuff but it seems that the less ambitious my PRs are the more likely they are to be merged :P

### DIFF
--- a/src/Perl6/World.nqp
+++ b/src/Perl6/World.nqp
@@ -615,14 +615,15 @@ class Perl6::World is HLL::World {
             %*LANG<MAIN> := $cursor.WHAT;
         }
 
+        my $actions := %*LANG<MAIN-actions>;
         # Add action method if needed.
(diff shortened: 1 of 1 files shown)


LLFourn:
Thanks!

lizmat:
> On 04 Jan 2016, at 12:13, LLFourn <notifications@github.com> wrote:
> 
> Thanks!

You’re welcome.  And indeed, smaller patches have a greater chance of not getting warnocked!



Liz

PR #674, opened 2016-01-05
Implement support for auto-generating private accessors.

autark:
'has $!foo' will auto-generate accessors in the same manner as 'has
$.foo' does, but the resulting accessor will be private. No accessor
is auto-generated should a privatly defined method with the same name
already exist. 'is rw' and 'is readonly' now makes sense for private
attributes as well.

### DIFF
--- a/src/Perl6/Actions.nqp
+++ b/src/Perl6/Actions.nqp
@@ -3102,6 +3102,7 @@ class Perl6::Actions is HLL::Actions does STDActions {
             my %config := hash(
                 name => $attrname,
                 has_accessor => $twigil eq '.',
+		has_private_accessor => $twigil eq '!',
                 container_descriptor => $descriptor,
                 type => %cont_info<bind_constraint>,
                 package => $*W.find_symbol(['$?CLASS']));
--- a/src/Perl6/Metamodel/BOOTSTRAP.nqp
+++ b/src/Perl6/Metamodel/BOOTSTRAP.nqp
@@ -1104,6 +1104,7 @@ BEGIN {
     #     has str $!name;
     #     has int $!rw;
     #     has int $!has_accessor;
+    #     has int $!has_private_accessor;
     #     has Mu $!type;
     #     has Mu $!container_descriptor;
     #     has Mu $!auto_viv_container;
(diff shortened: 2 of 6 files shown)


moritz:
I'm sorry if I'm being dense here, but what's the use case?

autark:
I have sent a patch proposal to rakudobug including a use case.

coke:
What's the RT number?

ugexe:
From the RT https://rt.perl.org/Public/Bug/Display.html?id=127172

    class MyClass {
        has $!secret = 5;
        method sauce(MyClass $other) {
            return self!secret() + $other!secret();
        } 
    } 

    say MyClass.new().sauce( MyClass.new() );

But doesn't this defeat the purpose of a private attribute from an untrusted class instance?

jnthn:
It certainly needs a bit of language design thought as to whether we wish to add this. Certainly you'd need to "trusts" to access it from outside of the scope of the class in question so it's not *that* much of an issue, but it does mean you can have access to the private state of other instances, which so far there has not been a way to do short of reflection (and sometimes it's the things you know can't happen that enable you to refactor, so this matters).

The other thing is that it adds code size overhead for each private attribute you declare, whether you use the feature or not.

Would be worth getting Larry's thoughts. I can argue it both ways, so would rather a little more time to think on it also.

autark:
The patch is only meant to give syntactic relief to the following example (which works in current Rakudo):

    class MyClass {
        has $!secret = 5;

        method !secret() { return $!secret; }
        method sauce(MyClass $other) {
            return $!secret + $other!secret;
        } 
    } 

    say MyClass.new().sauce( MyClass.new() );

It does not make private accessors public, so trying to access a private method of instance A in another class B should still be done with introspection + 'use MONKIES :lots-of'.

Another possible solution would be to allow direct access of private attributes within the class, something like $other.$!secret, though it doesn't look very neat.

pierre-vigier:
Waiting for a decision, with a module, something similar:
https://github.com/pierre-vigier/Perl6-AttrX-PrivateAccessor

lizmat:
Feels to me this issue has been solved in module space.  I therefore intend to close this PR in the coming days unless someone objects.

PR #679, opened 2016-01-10
decent infix:<**> for Complex

dankogai:
By switching to integer arithmetics and `sqrt()` when appropriate, `$z ** $int` and `$z ** ($int/2)` are more accurate, or as accurate as other platforms.   The following is how I noticed:

% perl6
````shell
> my $z = 8+6i
8+6i
> sin($z)**2+cos($z)**2
0.999999999978172+2.00088834390044e-11i
> sin($z)*sin($z)+cos($z)*cos($z)
1.00000000001455+0i
````
% irb
````shell
irb(main):001:0> require 'cmath'
=> true
irb(main):002:0> include CMath
=> Object
irb(main):003:0> z=8+6.i
=> (8+6i)
irb(main):004:0> sin(z)**2+cos(z)**2
=> (1.000000000014552+0.0i)
irb(main):005:0> sin(z)*sin(z)+cos(z)*cos(z)
=> (1.000000000014552+0.0i)
````

% python
````
Python 2.7.10 (default, Oct 23 2015, 18:05:06) 
[GCC 4.2.1 Compatible Apple LLVM 7.0.0 (clang-700.0.59.5)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> from cmath import *
>>> z=8+6j
>>> sin(z)**2+cos(z)**2
(1.000000000014552+0j)
>>> sin(z)*sin(z)+cos(z)*cos(z)
(1.000000000014552+0j)
````

%./perl6-m # after the patch
````shell
> my $z = 8+6i
8+6i
> sin($z)**2+cos($z)**2
1.00000000001455+0i
````

Dan the Complex Perl Monger

PR #680, opened 2016-01-11
make infix:<**>(Int:D,Int:D) failable

dankogai:
As suggested in https://github.com/perl6/roast/pull/95 , `inline:<**>` should raise `X:Numeric::Overflow` when it should but the current implementation does not.

The problem is that `nqp::pow_I()` overflows, it returns `1` (or `-1` for negative bases), not `0`.  So I ruled out all cases where `a ** b` == `1|-1` beforehand.

* `make test` is all okay.
* `make t/spec/S32-num` ( no more gotchas like https://github.com/rakudo/rakudo/pull/679 !) since all affected tests in `t/spec/S32-num/power.t` are marked `TODO`.
* `make spectest` is all okay!

````
Test Summary Report
-------------------
t/spec/S32-num/power.rakudo.moar                            (Wstat: 0 Tests: 76 Failed: 0)
  TODO passed:   12
Files=1092, Tests=51518, 204 wallclock secs (11.72 usr  4.15 sys + 1179.95 cusr 127.78 csys = 1323.60 CPU)
Result: PASS
````

Dan the Perl6 Monger

### DIFF
--- a/src/core/Exception.pm
+++ b/src/core/Exception.pm
@@ -1360,7 +1360,7 @@ my class X::Syntax::Perl5Var does X::Syntax {
       '$^O' => '$?DISTRO.name or $*DISTRO.name',
       '$^R' => 'an explicit result variable',
       '$^S' => 'context function',
-      '$^T' => '$*BASETIME',
+      '$^T' => '$*INITTIME',
       '$^V' => '$*PERL.version or $*PERL.compiler.version',
       '$^W' => '$*WARNING',
       '$^X' => '$*EXECUTABLE-NAME',


colomon:
Huh?  What is the alleged problem this is supposed to fix?

colomon:
I'm presuming it's

   > (-1) ** 324789274582374832749827348732084
   Numeric overflow

Seems like that might be better fixed in NQP and/or Moar?

colomon:
More generally, adding a long list of if statements into the code for fundamental math operations is a code smell.

dankogai:
* `(-1) * $n` does not over flow for `Int $n`.  It just returns `-1` for odd, `1` for even.  `2 ** 4553535345364535345634543534` does overflow.
* Unlike `Num`, `Int` does not seem to have have such exceptional values as `NaN` and `Infinity`.  So you have to check if the result of `nqp::pow_I()` is valid anyhow.
* IMHO you can hardly deodorize this operation.  Here's what `man pow` gave me.

````
SPECIAL VALUES
     pow(+-0, y) returns +-infinity and raises the "divide-by-zero" floating-
     point exception for y an odd integer < 0.

     pow(+-0, y) returns +infinity and raises the "divide-by-zero" floating-
     point exception for y < 0 and not an odd integer.

     pow(+-0, y) returns +-0 for y an odd integer > 0.

     pow(+-0, y) returns +0 for y > 0 and not an odd integer.

     pow(-1, +-infinity) returns 1.

     pow(1, y) returns 1 for any y, even a NaN.

     pow(x, +-0) returns 1 for any x, even a NaN.

     pow(x, y) returns a NaN and raises the "invalid" floating-point exception
     for finite x < 0 and finite non-integer y.

     pow(x, -infinity) returns +infinity for |x| < 1.

     pow(x, -infinity) returns +0 for |x| > 1.

     pow(x, +infinity) returns +0 for |x| < 1.

     pow(x, +infinity) returns +infinity for |x| > 1.

     pow(-infinity, y) returns -0 for y an odd integer < 0.

     pow(-infinity, y) returns +0 for y < 0 and not an odd integer.

     pow(-infinity, y) returns -infinity for y an odd integer > 0.

     pow(-infinity, y) returns +infinity for y > 0 and not an odd integer.

     pow(+infinity, y) returns +0 for y < 0.

     pow(+infinity, y) returns +infinity for y > 0.

     A domain error occurs if x is finite and negative and y is finite and not
     an integer.

     A domain error can occur if x is 0 and y less than or equal to 0.

     Range errors may occur.
````

Deodorizing rakudo-side would only odorize nqp side.  And when it comes to case handling, rakudo side should be easier.

pmichaud:
FWIW, this is starting to sound more and more like a "Perl 6 language spec" question and less of a "rakudo" one.

In particular, Perl 6 doesn't blindly duplicate what the C libraries do, so while "man pow" is a good starting point, it's not the final authority on how infix:<**> should work.

Pm

colomon:
I dunno, I'd say the main issue here is at which level we should fix problems with `nqp::pow_I`.  Acting like it's an unfixable black box and instead trying to paste over its problems at the Rakudo level strikes me as a bad idea.

PR #685, opened 2016-01-14
Fix several issues with List.squish

skids:
Only evaluate :&as and :&with the minimum number of times
  Prevent IterationEnd from leaking into :&as and :&with calls
  Use intuitive ordering of :&with args for non-commutative :with
  Always pass :&with consecutive elements of the original list
  Once committed to doing a push-all, no need to mutate anymore
  Might possibly fix r-j RT#126527 -- needs someone to test
  Will commit roast tests for these behaviors after merge

### DIFF
--- a/src/core/Any-iterable-methods.pm
+++ b/src/core/Any-iterable-methods.pm
@@ -1238,7 +1238,7 @@ Did you mean to add a stub (\{...\}) or did you mean to .classify?"
             has Mu $!iter;
             has &!as;
             has &!with;
-            has $!last;
+            has $!last_as;
             has int $!first;
             method BUILD(\list, &!as, &!with) {
                 $!iter  = list.iterator;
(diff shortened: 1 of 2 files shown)


skids:
Passes both 6.c (minus known failing tests) and master roast sets.

The biggest change here is the order of :with arguments.  A grep of the
ecosystem showed nobody using :with, so odds are low that current
behavior is relied upon.

When writing a custom :with this preserves the right-to-left ordering
of the arguments in the list so one can write :with({ $^last > $^next })

The second biggest change is that consecutive elements are always compared,
so in the case of 1, 2, 3 we do not compare 1 with 3.  This is essential if, for
example, you want to squish monotonically increasing sequences.

$ perl6 -e '(1,2,3,4,4,5,6,7,1,2,3).squish(:with({ $^last == $^next.pred })).say'
(1 4 1)

skids:
Tests for this are in roast PR#105

PR #687, opened 2016-01-21
Fix RT#127252: missing compliance with latest JSON spec

zoffixznet:
Newer JSON spec (RFC7159) does not restrict top level items to just objects/arrays and any valid value is acceptable. See:  http://rfc7159.net/rfc7159#rfc.section.2


coke:
We need to determine if this is 6.c safe.

moritz:
More to the point, rakudo's json methods should be only available internally; they aren't tested in t/spec/ and are vendor extensions.

b2gills:
This no longer applies cleanly, as it was moved to be under Rakudo::Internals

PR #689, opened 2016-01-22
Exposes Failure.handled as documented

salortiz:
For conformance with S32 and http://doc.perl6.org/type/Failure#method_handled

use Test; plan 2;
sub f() { fail 'boom' };
my $f = f;
ok not $f.defined, "Handle";
ok $f.handled

### DIFF
--- a/src/core/Failure.pm
+++ b/src/core/Failure.pm
@@ -2,7 +2,7 @@ my class Failure is Nil {
     has $.exception;
     has $.backtrace;
 #?if moar
-    has int $.handled;
+    has int $!handled;
 #?endif
 #?if jvm
     has $!handled;
--- a/src/core/Failure.pm
+++ b/src/core/Failure.pm
@@ -2,7 +2,7 @@ my class Failure is Nil {
     has $.exception;
     has $.backtrace;
 #?if moar
-    has int $!handled;
+    has int $.handled;
 #?endif
 #?if jvm
     has $!handled;


src/core/Failure.pm
```diff
@@ -2,7 +2,7 @@ my class Failure is Nil {
     has $.exception;
     has $.backtrace;
 #?if moar
-    has int $!handled;
+    has int $.handled;
 #?endif
 #?if jvm
     has $!handled;
```
b2gills: The JVM version should have the same change

jonathanstowe:
Actually, I would say that it wants an accessor that coerces it to a Bool rather than exposing a native int in the API i.e. : 
```
method handled() returns Bool {
   Bool($!handled);
}
```
or something like.

salortiz:
In fact, inside the class all uses of $!handled occurs  in Bool context, I don't know  why an int was chosen.

I can propose a more general patch, but first I wanted to try with a minimal approach for easy consensus.

timo:
On 22/01/16 22:08, Salvador Ortiz wrote:
>
> In fact, inside the class all uses of $!handled occurs in Bool 
> context, I don't know why an int was chosen.
>
> I can propose a more general patch, but first I wanted to try with a 
> minimal approach for easy consensus.
>
> —
> <https://github.com/rakudo/rakudo/pull/689#issuecomment-174049130>.
>

Bool is a full object, whereas an int is just a little slot that goes 
into the containing class' body. It also allows us to generate better 
bytecode on the VMs.

With a Bool, you'll end up allocating an extra Bool object for each 
Failure object and the Failure object then just has a pointer to the 
Bool object.

Hope that clarifies things.
   - Timo

salortiz:
New approach in #693, jonathanstowe's spirit

PR #690, opened 2016-01-23
Typo in X::Seq::Consumed exception

tommystanton:
I triggered the exception myself. :blush: I was presented with the following message (notice the typo of "addding"):

```text
This Seq has already been iterated, and its values consumed
(you might solve this by addding .cache on usages of the Seq, or
```


tommystanton:
(_Feel free to reject the pull request if another Rakudo contributor wants to make the tiny fix themselves._ :smile_cat:)

PR #692, opened 2016-01-23
Fix for RT#127354

zoffixznet:
Proposal that makes the warning disableable.

See also: #691


lizmat:
"note" should never be used in core code.  You "warn" or you don't  :-)

The other PR just removed the "note", which I think is appropriate.

PR #693, opened 2016-01-23
Implement Failure.handled as documented

salortiz:
After the discussion in #689, a new patch that add simply add the method handled --> Bool
Closing original PR

### DIFF
--- a/src/core/Failure.pm
+++ b/src/core/Failure.pm
@@ -55,6 +55,7 @@ my class Failure is Nil {
     }
     multi method Bool(Failure:D:) { $!handled = 1; Bool::False; }
 
+    method handled(Failure:D:)    { Bool($!handled); }
 
 #?if moar
     method Int(Failure:D:)        { $!handled ?? Int !! self!throw(); }
--- a/src/core/Failure.pm
+++ b/src/core/Failure.pm
@@ -2,10 +2,10 @@ my class Failure is Nil {
     has $.exception;
     has $.backtrace;
 #?if moar
-    has int $!handled;
+    has int $.handled is rw;
 #?endif
 #?if jvm
-    has $!handled;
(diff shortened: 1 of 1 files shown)


lizmat:
That should probably be:

+    method handled { nqp::p6bool($!handled) }

I'm not pushing / merging that until we have clarity on what being on 6.c means: if it means including all the warts of the 2015.12 release, then this should go into the 6.d.TOBE code.

src/core/Failure.pm
```diff
@@ -55,6 +55,7 @@ my class Failure is Nil {
     }
     multi method Bool(Failure:D:) { $!handled = 1; Bool::False; }
 
+    method handled(Failure:D:)    { Bool($!handled); }
```
b2gills: In the setting usually `nqp::p6bool` is used to coerce to a Bool

salortiz:
According to S04/Exceptions and roast in https://github.com/perl6/roast/blob/master/S04-exceptions/pending.t#L42
"The .handled  method is rw"
I'm now convinced that the best approach is the original one plus 'is rw', i.e. a simple
```diff
-    has int!handled;
+    has int $.handled is rw;
```
That allows to unfudge ten tests with:
```diff
--- a/S04-exceptions/pending.t
+++ b/S04-exceptions/pending.t
@@ -27,7 +27,7 @@ sub fail_it { fail $^a }
 
     ok all(@fails) ~~ Failure, '@fails is full of fail';
     ok $! !~~ Exception, 'fails do not enter $!';
-    #?rakudo 11 skip '$!.pending'
+    #?rakudo 1 skip '$!.pending'
     is +($!.pending.grep( ! *.handled )), 4,
        '$!.pending has three unhandled exceptions';
 
```

PR #695, opened 2016-01-28
Add osx to travis test matrix

ugexe:


ugexe:
This would vibe well with running the spec tests in travis

PR #696, opened 2016-01-28
add code to implement feature request (bug #127427)

tbrowder:
No tests or doc changes yet, do not know if feature requests for new methods are acceptable.


lizmat:
shouldn't the basename be returned if no extension could be found ?

tbrowder:
Fixed per Liz's comment.

tbrowder:
I'm now reconsidering this pull request as I think about the situation is a larger context.  Opinions, please:

I think the splitting of the basename is more general than applying to IO::Path, but in that context the name "stemname" I don't think is correct. A better name is "stem" so a possible "basenameparts" method would split basename into "stem" and "extension".

However, I think that the method really is a specialization of Str.split so maybe the Str class needs a method called something like "bisect" or "split-in-two" which would return either or both parts.  Then IO::path could have appropriate aliases for that method.  I have to look at class hierarchies and Roles and refine my proposal.

PR #698, opened 2016-01-30
Fix for RT#127402: Set.hash stringifies its objects

salortiz:
Now returns a parameterized Hash that reflect the Setty semantic

### DIFF
--- a/src/core/Setty.pm
+++ b/src/core/Setty.pm
@@ -27,9 +27,9 @@ my role Setty does QuantHash {
     multi method Bool(Setty:D:) { %!elems.Bool }
 
     multi method hash(Setty:D: --> Hash) {
-        my %e;
-        %e{$_} = True for %!elems.values;
-        %e;
+        my \e := Hash.^parameterize(Bool, Any).new;
+        e{$_} = True for %!elems.values;
+        e;
     }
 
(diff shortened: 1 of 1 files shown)


timo:
looks good, thanks!

salortiz:
Seen applied, and then reverted by @lizmat in 2e17820, waiting for definition on bugfixes.
BTW, the syntax my %e{Any} isn't available at this stage, so I use direct parameterization.
Should reopen?

jnthn:
Yes, reopen. Expect to have the fix guidelines out around the start of next week.

lizmat:
> On 30 Jan 2016, at 21:06, Salvador Ortiz <notifications@github.com> wrote:
> Seen applied, and then reverted by @lizmat in 2e17820, waiting for definition on bugfixes.
> BTW, the syntax my %e{Any} isn't available at this stage, so I use direct parameterization.

I see now.   Hmmm… I had compiled the setting with %e{Any}  (which it did, without problem), but apparently it is indeed not doing the right thing.  So Salvador, I stand corrected: your original pull request should be done once we agree on the handling of bug fixes in rakudo vis-a-vis language levels.


Salvador Ortiz++



Liz

PR #701, opened 2016-02-01
Add .shortname to DefiniteHOW

salortiz:
Since DefiniteNOW don't "does" Metamodel::Naming. it needs a shortname
method, unconditionally used by Mu.gist

See http://irclog.perlgeek.de/perl6/2016-02-01#i_11970457
Test:
<sortiz> m: Int:D.gist
<camelia> rakudo-moar a5fe34: OUTPUT«Method 'shortname' not found for invocant of class 'Perl6::Metamodel::DefiniteHOW'␤  in block <unit> at /tmp/3vJXmqV_i_ line 1␤␤»

### DIFF
--- a/src/Perl6/Metamodel/DefiniteHOW.nqp
+++ b/src/Perl6/Metamodel/DefiniteHOW.nqp
@@ -46,6 +46,17 @@ class Perl6::Metamodel::DefiniteHOW
         }
     }
 
+    method shortname($definite_type) {
+	if nqp::isnull(nqp::typeparameterized($definite_type)) {
+            '?:?'
+        }
+        else {
+            my $base_type := nqp::typeparameterat($definite_type, 0);
+            my $definite  := nqp::typeparameterat($definite_type, 1);
+            $base_type.HOW.shortname($base_type) ~ ':' ~ ($definite ?? 'D' !! 'U')
+        }
+    }
+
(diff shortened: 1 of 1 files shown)


salortiz:
After this PR was generated, I found RT#127059, similar but with CurriedRoleHOW, and I thought that may be better a more general approach:

```patch
--- a/src/core/Mu.pm
+++ b/src/core/Mu.pm
@@ -369,7 +369,10 @@ Please refactor this code using the new Iterator / Seq interface.
     }
 
     proto method gist(|) { * }
-    multi method gist(Mu:U:) { '(' ~ self.^shortname ~ ')' }
+    multi method gist(Mu:U:) {
+        my \n := nqp::can(self.HOW, 'shortname') ?? self.^shortname !! self.^name;
+        '(' ~ n ~ ')'
+    }
     multi method gist(Mu:D:) { self.perl }
 
     method perlseen(Mu:D \SELF: $id, $perl, *%named) {
```

Comments?

moritz:
Thanks @salortiz.

IMHO there is room for both improvements: give DefinitiveHOW a shortname, but also make Mu.gist more robust.

Cheers,
Morit

jnthn:
I think all the MOP objects should support shortname as well as name. It's
OK to make .gist more robust too, though.

PR #704, opened 2016-02-03
Use nqp::stat_time on the JVM backend as well as moar.

donaldh:


donaldh:
Ugh, that 'Unbust parse_revision" commit is in a different PR. Why'd it end up in here too?

niner:
Because your branch had that commit while nom doesn't. Please rebase your branch onto the current nom and open a new PR. I merged your parse_revision fixes.

PR #706, opened 2016-02-06
Fixes Rat.perl to actually generate valid code

autarch:


autarch:
See https://github.com/perl6/roast/pull/101 for the corresponding spec test addition.

lizmat:
Added as separate commits, some slightly re-imagined, autarch++

PR #707, opened 2016-02-07
Just call $<nu>.ast.compile_time_value in Perl6::Actions.bare_rat_number

autarch:
The value of $<nu>.ast can either be a QAST::Want or a QAST::WVal depending on
the value of the numerator. Any value of 2**31 or greater gives us a
QAST::WVal.

Previously the code called "@($<nu>.ast)[0].compile_time_value" - this works
fine for a QAST::Want, but a QAST::WVal does not have any children, and so
this would lead to an error of "Cannot find method 'compile_time_value'".

Internally QAST::Want.compile_time_value is implemented as
"self[0].compile_time_value()" so it seems like there's no good reason for us
to do same thing in Perl6::Actions. Yay polymorphism!

### DIFF
--- a/src/Perl6/Actions.nqp
+++ b/src/Perl6/Actions.nqp
@@ -7226,7 +7226,7 @@ class Perl6::Actions is HLL::Actions does STDActions {
     method rat_number($/) { make $<bare_rat_number>.ast }
 
     method bare_rat_number($/) {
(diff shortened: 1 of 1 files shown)


autarch:
Creating a PR for the sake of Travis.

autarch:
See https://github.com/perl6/roast/pull/101 for the corresponding spec test fix.

PR #708, opened 2016-02-07
Fix Instant to round trip properly via .perl.EVAL

autarch:
Previously it was generating a string something like this:

  Instant.from-posix((42, 0))

The second set of parentheses caused the argument to passed as a single List
rather than individual items. We also need to force the second argument to be
a Bool rather than 0 or 1.

Tested in https://github.com/perl6/roast/pull/103

### DIFF
--- a/src/core/Instant.pm
+++ b/src/core/Instant.pm
@@ -32,7 +32,9 @@ my class Instant is Cool does Real {
         'Instant:' ~ $!tai
     }
     multi method perl(Instant:D:) {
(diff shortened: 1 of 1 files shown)


lizmat:
Feels to me that to-posix should maybe return a bool rather than a 0/1 .  That would make to-posix more correct, and it would simplify the .perl to "Instant.from-posix{self.to-posix.perl}"

lizmat:
Reimagined with bc50164e686fb32a9fae3 , thanks for the nudging!

PR #712, opened 2016-02-15
Add named argument :initial-elems to Blob.new

salortiz:
Create an allocated, empty (zeroed) Blob of the given size.
Works also with Buf and subclases.

### DIFF
--- a/src/core/Buf.pm
+++ b/src/core/Buf.pm
@@ -5,8 +5,10 @@ my class X::Cannot::Lazy        { ... }
 my class X::Experimental        { ... }
 
 my role Blob[::T = uint8] does Positional[T] does Stringy is repr('VMArray') is array_type(T) {
-    multi method new() {
-        nqp::create(self)
+    multi method new(:$elems) {
+        my $buf := nqp::create(self);
+        nqp::setelems($buf, nqp::unbox_i($elems.Int)) if $elems;
+        $buf
     }
     multi method new(@values) {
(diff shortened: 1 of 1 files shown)


jnthn:
+1 to the idea, though I think -1 to :elems(...). With shaped arrays we called it :shape(...) instead, and while a multi-dimensional Buf doesn't really make sense, it makes sense to use the same API if we're going to actually *fix* the size (though that'll need a more extensive patch to use the MultiDimArray REPR instead). Furthermore, that'd mean you could simply say:

    my @a[1024] is Buf;

If the idea is to initialize it with a certain number of elements but then allow resizing, then I'd rather go with something that's a bit more explicit about that.

salortiz:
@jnthn, thanks for the comment.

I don't use `:shape` because I don't want to *fix* the size, only pre-allocate some slots so:

`Blob.new:100elems.elems == 100`

ugexe:
You can pre-allocate pretty nicely this way too: `Blob.new: 0 xx 5`

salortiz:
Yes @ugexe, but the difference in speed is really big:
Blob.new(0 xx 10000) vs Blob.new(:initial-elems(10000)) -> 200.2908/s  vs 38052.1614/s

salortiz:
And with FSA enabled in MoarVM even bigger: 200.9773/s vs 58411.0240/s!

salortiz:
With lizmat++ Blob.allocation this can be closed.

PR #714, opened 2016-02-17
Fix the behaviour of require WRT to global symbols

jonathanstowe:
Recent changes in require meant that code that did a require in an inner lexical scope were not having the symbols merged in such a way that they could be found by an indirect lookup, this broke for instance, the way that Panda dealt with the Build.pm files

### DIFF
--- a/src/core/operators.pm
+++ b/src/core/operators.pm
@@ -584,7 +584,7 @@ sub REQUIRE_IMPORT($compunit, *@syms,:$target-package) {
     }
     # Merge GLOBALish from compunit.
     # XXX: should probably use CALLER::UNIT:: but RT #127536
-    CALLER::LEXICAL::GLOBALish::.merge-symbols($GLOBALish);
+    GLOBAL::.merge-symbols($GLOBALish);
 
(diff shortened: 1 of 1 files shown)


ugexe:
This PR allows `IO::Socket::SSL` to be installed again, which reveals more incorrect behavior:

    nickl@li685-90:~/perl6$ perl6 -v
    This is Rakudo version 2015.12-323-gbb2953c built on MoarVM version 2016.01-28-g2136293
    implementing Perl 6.c.

    nickl@li685-90:~/perl6$ perl6 -e 'require IO::Socket::SSL; say 42; say (try require IO::Socket::SSL);'
    42
    (SSL)

    nickl@li685-90:~/perl6$ perl6 -v
    This is Rakudo version 2015.12-380-g6159b5e built on MoarVM version 2016.01-28-g2136293 implementing Perl 6.c.

    nickl@li685-90:~/perl6$ perl6 -e 'require IO::Socket::SSL; say 42; say (try require IO::Socket::SSL);'
    42
    Nil

PR #716, opened 2016-02-18
NC - Add sub to register libraries

Skarsnik:
Currently NativeCall does not provide a nice way to give information on a nice way at run time. 

The native trait that receive library information is sadly resolved at compile time, even if it's really determined at the first run of the sub, it's fine for small case or when you provide the library (so the location is 'fixed' at compile time). But if you want to give a way for the user to change a way to change the location (eg an ENV variable, think of precompil with lib) you will need to give a Callable to the trait.
There are 2 issues with that. The first is that the Callable will be run only when the sub is called, it could be very bad if you need to do something heavy to find the library information (run pkg-config?) it's probably more acceptable to do heavy stuff on 'init' in your code.
The second one is my fault. when I rewritten the guess_library_name I choose that when the user give a Callable he have to do everything himself. So there is no 'magic' that transform a 'foo' in libfoo.so

To give you an example on how to do the %*ENV solution, you need to use NC 'internal'
`
use NativeCall :ALL; 
sub foo is native(sub {%*ENV<MY_LIBFOO> || guess_library_name(('foo', v42)) } )
`

This PR add a main function. `register-native-library` that return a special type (NC-Library-Handle) that you can give to the native trait.

This is a full example. (ENV is optionnal)
`my NC-Library-Handle $LIBFOO = register-native-library('Foo', 'foo', :version(v42), :ENV<MY_LIBFOO>);
sub foo is native($LIBFOO)`

The NC-Library-Handle is a role (internally NC will read the library-file attribute), it allow for people to write other mean to get the library.

You will probably note that it has a extra argument (Foo), it's a keyname used to identify the library. It's mostly useless in most case. But there are other functions to add/remove/list registered lib. One usage if for example, if you want to reuse the Library Handle to call cglobals on your modules using a binding/raw module. You can call `get-native-library('Key')`

The only downside of this is you need to type the variable givent to the native trait. I did not want to remove the compile check of a proper type passed to the trait to avoid that.

I also put back the warning (instead of a note this time) about the missing version. A shared lib should have an ABI/APi version (see the ticket about the version addition, there are link about *bsd/os X policy about them), it's an ecxeptionnal case to NOT have one. And it was also to issue developper to not just go with 'Oh I have the .so on my system it's enought'. The register-native-lib has an explicit argument to not provide a version.


Skarsnik:
This an example of what I could write with a multi file binding for a lib https://gist.github.com/Skarsnik/8687bba8063422798d1a2e83e859c195

jonathanstowe:
Actually this is something I had a use for at some point and either gave up or hacked around it ;-)

So yeah :+1: from me.

Skarsnik:
Also as nine mentioned on IRC:
<nine> Skarsnik: I wonder, if we should rethink the whole NativeCall interface and make that available via use NativeCall :api<2>;

Could be a good idea to have some versioning in NC

JJ:
First, conflicts need to be solved... And then, new features are now discussed in the problem-solving repo before committing them here.

Skarsnik:
This PR exists before the problem-solving repo was created and I posted also in the old 6.d or 6.e repo that existed to discuss changes. 

Sure I can probably spend another hour solving conflict and fighting with git so it gets ignored for another 4 years x)

JJ:
> This PR exists before the problem-solving repo was created and I posted also in the old 6.d or 6.e repo that existed to discuss changes.
> 
> Sure I can probably spend another hour solving conflict and fighting with git so it gets ignored for another 4 years x)

Sorry to hear that, and I understand you perfectly. We intend precisely to do the opposite, address all outstanding PRs, which is why we would really appreciate if you take some time to fix the conflict. But we understand whatever you decide to do...

Skarsnik:
I will give it a try tomorrow, want me to post something also on the prob-solving repo?

JJ:
> I will give it a try tomorrow, want me to post something also on the prob-solving repo?

It would be awesome if you would do so. Some kind of consensus on what needs to be done needs to be reached first, and that's the way we do that now... Thanks a lot.

lizmat:
Am not sure of the status of this PR now.  Can someone enlighten?

Skarsnik:
Well, I am waiting for a reply on https://github.com/Raku/problem-solving/issues/186 to maybe try again to adapt the code to newest code.

lizmat:
Ok, I'll keep this open then.

PR #719, opened 2016-02-27
IO::Proc: store command line and use it in X::Proc::Unsuccessful

rudis:
This allows the user of a program which doesn't catch the exception to
debug the problem or provide a good bug report.

There was some discussion in #perl6 if dumping the whole command line arguments is a good idea or not, but to help debugging unknown programs I'd really like to see the full output so I know exactly what's going on. Also helps with spurious errors.

### DIFF
--- a/src/core/Exception.pm
+++ b/src/core/Exception.pm
@@ -2426,7 +2426,8 @@ my class X::Language::Unsupported is Exception {
 my class X::Proc::Unsuccessful is Exception {
     has $.proc;
     method message() {
-        "The spawned process exited unsuccessfully (exit code: $.proc.exitcode())"
+        "The spawned process exited unsuccessfully " ~
+        "(exit code: $.proc.exitcode(), command line: $.proc.args())"
     }
 }
 
--- a/src/core/Proc.pm
+++ b/src/core/Proc.pm
@@ -5,6 +5,7 @@ my class Proc {
     has $.exitcode = -1;  # distinguish uninitialized from 0 status
     has $.pid;
     has $.signal;
+    has @.args;
 
     has $!in_fh;
     has $!out_fh;
@@ -95,6 +96,7 @@ my class Proc {
 
     method spawn(*@args ($, *@), :$cwd = $*CWD, :$env) {
         my %env := $env ?? $env.hash !! %*ENV;
+        self.args = @args;
         self.status(nqp::p6box_i(nqp::spawn(
             CLONE-LIST-DECONTAINERIZED(@args),
             nqp::unbox_s($cwd.Str),
@@ -107,6 +109,7 @@ my class Proc {
 
     method shell($cmd, :$cwd = $*CWD, :$env) {
         my %env := $env ?? $env.hash !! %*ENV;
+        self.args = [$cmd];
         self.status(nqp::p6box_i(nqp::shell(
             nqp::unbox_s($cmd),
             nqp::unbox_s($cwd.Str),


lizmat:
Looks good to me.  I just wonder about any security implications...

rudis:
You mean like passwords in the command line? I don't think this will be an issue often, as the running user will have access to the data as well. Writing it to log files might be a problem, but that's already true when the password is present in the command-line as ps and similar can extract it.

lizmat:
more worried about P6 errors being exposed in a web frontend, but error logging is also an issue

It therefore feels to me that this is really a debugging aid that needs to be activated only with some environment variable

rudis:
The environment variable bit came up during the discussion as well and I think it's problematic. My goal is to allow users to find the root cause of an issue quickly and I can't expect a normal user to know Perl6 or the environmental variables it uses. In the same way I expect a better error than "file not found" I want to get a better error then "program failed to run".

All error messages can leak sensitive data and I think it's the task of the programmer to sanitize the output if they expose such messages to the public. Or maybe we could use a environment variable/setting which performs the sanitization if requested. But I think the default should be to display the full data to allow proper debugging/bug reporting.

rudis:
Any news on this?

moritz:
I'd like to merge this.

It's best practice not to pass passwords on the command line anyway, because on standard Linux systems, everybody can read them in the process list anyway. And even if that's disallowed across user boundaries, in the context of a web server other processes of the web server user can still see it.

zoffixznet:
:-1: on this change from me.

### 1) The already-mentioned security issue.

> I think it's the task of the programmer to sanitize the output if they expose such messages to the public

Why would that be programmer's task? Programmers makes software and users run it. These errors would show up in user's environment that's out of the programmer's control. Some users may not even be aware of where the output log from their software will show up.

I think there are too many possibilities for this data to end up in the wrong place, and that's the price to bring a rather small amount of benefit (the programmer can see what the args are in code, or dump them if they're dynamic).

### 2) You're exposing `.args` attribute to the users

It's poorly named, since the command being run is stored in them as well. .ags is short for arguments, but... arguments to what?

ugexe:
Without also dumping the $*CWD and $*ENV passed in to IO::Proc its not even very useful for debugging. Nor would one want to wade through all of ENV in most cases if they were. So it seems like the type of debugging the programmer should code themselves as needed

AlexDaniel:
@ugexe either all or nothing, huh? No, it is actually pretty useful.

By the way, see this: https://rt.perl.org/Ticket/Display.html?id=127282. I think that it is **very** related.

moritz:
How about we store all the arguments, but in method `message` only use the name of the command (so `args[0]`)? The command itself should never be a secret, but knowing the command is already a help in diagnosing the error. Better name suggestions than `args` are welcome, maybe `@.commandline`?

zoffixznet:
I like `@.command`

ugexe:
@AlexDaniel There is no way for the end user to disable the output of this type change vs the current behavior that allows the programmer to dump this information with an additional line of code. The reasoning for this PR vs having the programmer themselves do this proper was: `Because the developer is/was stupid or lazy.` and `I really really like to program without ever catching any exception which will "never" occour`. I don't find these reasons compelling.

Last time this was discussed most people seemed on board with just the first argument option (except the PR author). Additionally we now have [RAKUDO_EXCEPTIONS_HANDLER](https://github.com/rakudo/rakudo/commit/aaf7c3cf573814a7027b514000c7505146f19da6) that could be used to for dumping specific/all things

PR #720, opened 2016-03-04
Fix handling of sigilles params in USAGE

kalkin:
Currently `perl6 -e sub MAIN(\bar) {}` will output bar parameter as `<ar>`,
this fixes it. Added also the apropriate test to
[perl6/roast](https://github.com/perl6/roast/pull/107)

### DIFF
--- a/src/core/Main.pm
+++ b/src/core/Main.pm
@@ -70,7 +70,7 @@ my sub MAIN_HELPER($retval = 0) {
                 my $argument;
                 if $param.named {
                     if $param.slurpy {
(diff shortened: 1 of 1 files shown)


lizmat:
Thank you bringing this to our attention.  I didn't like the relative expensive .subst approach: I also felt the logic for creating a usage name needed to live in Parameter.pm.  So I added the method "usage-name" there with 676ad669c9ae92f5c9890 .  And then I used this in the main helper with 7bdebae05e3adc .

Again, thank you for your work!  Keep more of that coming!

PR #726, opened 2016-03-16
IO::Handle - Fix warning on .gist method on empty path name

azryelryvel:
Currently, "say IO::Handle.new.gist;" warns :

Use of uninitialized value $!path of type Any in string context
Any of .^name, .perl, .gist, or .say can stringify undefined things, if needed. in block at -e line 1
IO::Handle<>(closed)

This commit should correct it =)


lizmat:
Did you actually *try* this patch???

The .gist will only be executed if you add ()  (so: <$!path.gist()>)

So implemented with 4ed9fa9fbc7b8037d60

PR #727, opened 2016-03-18
Fix .slurp-rest(:!bin)

ugexe:


lizmat:
Good catch!

ugexe: is there a reason why you added the :$bin in the second candidate?

ugexe:
So `:!bin` is not treated as an unexpected named parameter

lizmat:
ugexe: but that is only for subs, not methods.  Unexpected named parameters disappear into thin air.

lizmat:
in any case, adding a "where" clause in a signature generally slows down things a lot, so I'm coming up with another solution

PR #730, opened 2016-03-29
$*VM.platform-library-name: only return IO::Path

ugexe:
Previously this would return a `Str` or `IO::Path` depending on if the passed in `IO::Path` was absolute or relative:

```
perl6 -e 'say $*VM.platform-library-name($*CWD.child("xxx").relative.IO)'
"/home/nickl/perl6/rakudo/libxxx.so".IO

perl6 -e 'say $*VM.platform-library-name($*CWD.child("xxx").absolute.IO)'
/home/nickl/perl6/rakudo/libxxx.so
```

Both now return an IO::Path. A few further changes were required for areas that used $*VM.platform-library-name

### DIFF
--- a/lib/NativeCall.pm6
+++ b/lib/NativeCall.pm6
@@ -202,8 +202,7 @@ sub guess_library_name($lib) is export(:TEST) {
     return '' unless $libname.DEFINITE;
     #Already a full name?
     return $libname if ($libname ~~ /\.<.alpha>+$/ or $libname ~~ /\.so(\.<.digit>+)+$/);
-    my $pln = $*VM.platform-library-name($libname.IO, :version($apiversion || Version));
-    return $libname.IO.is-absolute ?? $pln.absolute !! $pln.relative;
+    $*VM.platform-library-name($libname.IO, :version($apiversion || Version));
 }
 
 sub check_routine_sanity(Routine $r) is export(:TEST) {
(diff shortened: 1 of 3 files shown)


lizmat:
Wouldn't it make more sense to always return a string, as apparently, we're interested in the *name*?

ugexe:
Sure, I wasn't sure which way to go with it. Since its capable of returning parts of the path besides the modified basename it just didn't _feel_ like a name, so I went the other way with it

niner:
Even though it's just the name, even that is a relative path and you certainly want to use it as a path.

lib/NativeCall.pm6
```diff
@@ -403,7 +390,7 @@ sub cglobal($libname, $symbol, $target-type) is export is rw {
     Proxy.new(
         FETCH => -> $ {
             nqp::nativecallglobal(
-                nqp::unbox_s(guess_library_name($libname)),
+                nqp::unbox_s(guess_library_name($libname).IO.absolute),
```
salortiz: This worries me, `guess_library_name` can properly return `''` when we request a symbol for an already loaded library.

ugexe: If that is true I'd say there are bigger problems, like a method named "guess_library_name" trying to do magic, or indexing %lib on such return values

Skarsnik: guess_library_name returns something dlopen and co can use to load the right library. That why a path or a name is valid.

Also some things outside NC.pm use guess_library_name (it's not ideal, but it's the only way to avoid rewriting it when we want run time lib resolution see: #716 for more details) so it's not that easy to change it

For the missing test on win32/os X. I just did not have these platforms set up to write tests for it.

PR #731, opened 2016-03-30
Fix for WhateverCodes in sink context

ab5tract:
This is my first NQP patch. My apologies without surprise if it turns out to be too boneheaded to merge.

The intent of the patch is to silence this warning:

```
perl6 -e 'my $f = <25 25 25>; $f ~~ s:nth(*-1)[\d+] = 42; say $f'
WARNINGS for -e:
Useless use of "-" in expression "*-1" in sink context (line 1)
25 42 25
```

### DIFF
--- a/src/Perl6/Optimizer.nqp
+++ b/src/Perl6/Optimizer.nqp
@@ -1388,13 +1388,13 @@ class Perl6::Optimizer {
                 elsif $op.node && $!void_context {
                     my str $op_txt := nqp::escape($op.node.Str);
                     my str $expr   := nqp::escape(widen($op.node));
-                    unless $op_txt eq '/' && $op[1].has_compile_time_value && $op[1].compile_time_value == 0 {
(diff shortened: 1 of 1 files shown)


lizmat:
wouldn't this check imply that anything with a multiplication wouldn't warn either ?

ab5tract:
Lizmat, excellent point. And so obvious now that you've mentioned it...

Is it a sane approach to iterate over the nodes of the expression and type check for `$*W.find_symbol(['Whatever'])`.

ab5tract:
It appears that there is no problem with multiplication, as this warning is not triggered in the current Rakudo for

```
> my $f = <25 25 25>; my $i = 3; $f ~~ s:nth(1 * $i)[\d+] = 42; say $f
25 25 42
```

ab5tract:
I have pushed the 'check first and last indices' approach which appears to work just fine for multiple `*` characters in an expression.

Please let me know if there is anything else blocking this patch.

PR #732, opened 2016-04-01
Combinations fixes

skids:
Fix sub combinations to return () not ((),) for absurd $k
 Fix List.combinations(Range) to use right values with careted Range endpoints
 Fix List.combinations(Range) to return () on an iterably-empty range
 Ensure only one empty list appears in result, only when a k==0 is involved
 Use a map/map instead of gather/take, no discernable performance impact
 ...someone may wish to weigh in on which will optimize better in future.
 ...this may help a plan to avoid pulling during .elems on the resulting Seq

### DIFF
--- a/src/core/List.pm
+++ b/src/core/List.pm
@@ -4,6 +4,7 @@ my class Supply { ... }
 my class Supplier { ... }
 
 my sub combinations(Int() $n, Int() $k) {
+    return () if $k < 0;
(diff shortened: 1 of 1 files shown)


skids:
Passes current roast.  Did not check 6.c.

skids:
Test cases in roast PR#110

PR #733, opened 2016-04-01
Add kv adverb to first

0racle:

The docs define first as
````
```
... however there is no `:kv` adverb implemented. Although [S32](http://design.perl6.org/S32/Containers.html) does not mention a `first-kv`, there's a reasonable user expectation that `first` would have same adverbs as `grep`.

This patch adds an additional conditional check for `%a<kv>` to the private worker method `first-result` and returns `(index,value)`

Tests `first-kv.t` and `first-end-kv.t` have been written and will be submitted to roast accordingly.

### DIFF
--- a/src/core/Any-iterable-methods.pm
+++ b/src/core/Any-iterable-methods.pm
@@ -508,7 +508,7 @@ Did you mean to add a stub (\{...\}) or did you mean to .classify?"
                     value
                 }
                 elsif %a<kv> {
-                    (index,value).Seq
+                    index,value
                 }
                 else {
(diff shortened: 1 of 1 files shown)


lizmat:
You probably also would need to cover the case of :!kv   :-)

0racle:
I messed up and deleted my fork. I couldn't figure out how to fix it, so I [resubmitted](https://github.com/rakudo/rakudo/pull/734) my pull. Can this one be deleted/rejected?

Emeric54:
@0racle you can use the `Close Pull Request` button. :)

PR #734, opened 2016-04-01
Add kv adverb to first

0racle:

The docs define first as
````
```
However there is no `:kv` adverb implemented. Although [S32](http://design.perl6.org/S32/Containers.html) does not mention a `first-kv`, there's a reasonable user expectation that `first` would have same adverbs as `grep`.

This patch adds an additional conditional check for `%a<kv>` to the private worker method `first-result` and returns `(index,value)`. It also added the negated `:!kv` to the other negated adverbs that just return the value itself.

Tests `first-kv.t` abd `first-end-kv.t` have been written and will be submitted to roast accordingly.

### DIFF
--- a/src/core/Any-iterable-methods.pm
+++ b/src/core/Any-iterable-methods.pm
@@ -507,9 +507,12 @@ Did you mean to add a stub (\{...\}) or did you mean to .classify?"
                 elsif %a<v> {
                     value
                 }
+                elsif %a<v> {
+                    (index,value)
+                }
                 else {
                     my $k = %a.keys[0];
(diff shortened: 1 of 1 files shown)


src/core/Any-iterable-methods.pm
```diff
@@ -507,9 +507,12 @@ Did you mean to add a stub (\{...\}) or did you mean to .classify?"
                 elsif %a<v> {
                     value
                 }
+                elsif %a<v> {
```
AlexDaniel: Same condition as above?

PR #736, opened 2016-04-02
Add kv adverb to first

0racle:

The docs define first as
````
```
However there is no `:kv` adverb implemented. Although [S32](http://design.perl6.org/S32/Containers.html) does not mention a `first-kv`, there's a reasonable user expectation that `first` would have same adverbs as `grep`.

This patch adds an additional conditional check for `%a<kv>` to the private worker method `first-result` and returns `(index,value).Seq`.

A Seq is return for 2 reasons. One being that `grep :kv` returns a Seq, so this is consistent. The second reason is, as [jnthn pointed out](http://irclog.perlgeek.de/perl6/2016-04-01#i_12271679), most things that produce values you can iterate return a Seq. 

Tests `first-kv.t` and `first-end-kv.t` have been written and will be submitted to roast accordingly.

### DIFF
--- a/src/core/Any-iterable-methods.pm
+++ b/src/core/Any-iterable-methods.pm
@@ -507,9 +507,12 @@ Did you mean to add a stub (\{...\}) or did you mean to .classify?"
                 elsif %a<v> {
                     value
                 }
+                elsif %a<kv> {
+                    (index,value).Seq
+                }
                 else {
                     my $k = %a.keys[0];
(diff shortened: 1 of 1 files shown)


0racle:
Test files attached here for review/comments

[first-end-kv.txt](https://github.com/rakudo/rakudo/files/200795/first-end-kv.txt)
[first-kv.txt](https://github.com/rakudo/rakudo/files/200796/first-kv.txt)

PR #737, opened 2016-04-09
disallow dimensions smaller or equal 0 in shapes

spebern:
suggestion for fix of Bug #126800

### DIFF
--- a/src/core/Rakudo/Internals.pm
+++ b/src/core/Rakudo/Internals.pm
@@ -6,6 +6,7 @@ my class X::Assignment::ToShaped { ... };
 my class X::Str::Sprintf::Directives::BadType { ... };
 my class X::Str::Sprintf::Directives::Count { ... };
 my class X::Str::Sprintf::Directives::Unsupported { ... };
+my class X::IllegalDimensionInShape { ... };
 
 my class Rakudo::Internals {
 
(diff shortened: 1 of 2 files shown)


src/core/Rakudo/Internals.pm
```diff
@@ -402,8 +403,12 @@ my class Rakudo::Internals {
             if nqp::istype($_, Whatever) {
                 X::NYI.new(feature => 'Jagged array shapes');
             }
+            my $dim = $_.Int;
+            if $dim <= 0 {
+                X::IllegalDimensionInShape.new(dim => $dim).throw;
```
vendethiel: `dim => $dim` is usually written `:$dim`

lizmat:
935059ffb5a239d958cbc3ef addressed this, and some other stuff

AlexDaniel:
Thanks, @spebern! However, #126800 cannot be closed without tests. Perhaps you could write some?

If I'm not mistaken, https://github.com/perl6/roast/blob/master/S09-multidim/decl.t is the right file for such tests.

spebern:
Ok thanks for linking the file! I will write them today or tomorrow.

AlexDaniel:
Already done by @usev6 in https://github.com/perl6/roast/commit/a2617480f4, but feel free to add more.

AlexDaniel:
It disallows zero-sized arrays, why? ([RT #131965](https://rt.perl.org/Ticket/Display.html?id=131965))

PR #738, opened 2016-04-09
Loopy REPL

awwaiid:
Lift more of the control for the REPL directly into REPL.pm, such as the main loop itself.


lizmat:
Looks like there's some debug code left behind in this PR?

> say "hello world"
Inside of partial-eval
Just about to call compiler eval
hello world
True

lizmat:
also, ctrl-D-ing out of the REPL (instead of doing "exit" gives:

> ===SORRY!===
Cannot unbox a type object

awwaiid:
@lizmat thanks for taking a look -- the current state of this PR is looking for comments on the overall thing. The ctrl-D issue and also I think if you do any error it doesn't catch it right and exits -- I want to fix all of those before I remove the "WIP" (Work In Progress) status.

I'm kinda surprised that the tests passed :)

One of the goals is for there to be zero "nqp::" calls inside of REPL.pm, and advise on getting rid of the existing ones is welcome.

awwaiid:
@lizmat @hoelzro I have squashed up and pushed up a version of this that seems to work. Please git it a spin!

hoelzro:
@awwaiid I'll give it a try and let you know how it goes!

hoelzro:
I tried it out, and it works great!  One quick thing - I noticed you removed `interactive_result` and `interactive_exception` - we should probably leave those in, in case loading the Perl 6 REPL fails and we need to fall back to the REPL that NQP provides us.

src/Perl6/Compiler.nqp
```diff
@@ -77,38 +68,16 @@ class Perl6::Compiler is HLL::Compiler {
...
+        my $stdin    := nqp::getstdin();
+        my $encoding := ~%adverbs<encoding>;
+        if $encoding && $encoding ne 'fixed_8' {
+            nqp::setencoding($stdin, $encoding);
         }
 
-        my $super := nqp::findmethod(HLL::Compiler, 'interactive');
-        my $result := $super(self, :interactive(1), |%adverbs);
-        self.teardown();
+        my $result := $!p6repl.repl-loop(:interactive(1), |%adverbs);
```
hoelzro: Keep in mind that the REPL6 load may have failed, and `$!p6repl` can be `Mu` here - we need to gracefully fallback if that's the case.

src/core/REPL.pm
```diff
@@ -57,14 +59,14 @@ do {
...
-        method readline(Mu \SELF, Mu \super, Mu \stdin, Mu \stdout, Mu \prompt) {
+        method repl-read(Mu \prompt) {
             my $line = readline(prompt);
 
             if $line.defined {
                 add_history($line);
             }
 
-            $line // nqp::null_s()
+            $line;
```
hoelzro: This is a total nitpick on my part and by no means should you feel compelled to follow it - I'm just pointing out my style and attempting to justify it.

I prefer that if implicit returns are used (as `$line` is used here), that the semicolon be omitted from the final line of the block.  This may seem silly, but to me, it a) serves as a hint to the reader, communicating that
the intent is to return that value, and b) if someone is working on the code and adds a line after the implicit return (for debugging or whatever), they get a compiler error rather than suddenly having the routine they just modified return a wildly different value.

src/core/REPL.pm
```diff
@@ -277,27 +282,118 @@ do {
...
+                }
 
-                    when X::Comp::FailGoal {
-                        if $!multi-line-enabled && .pos == code.chars {
-                            return needs_more_input();
-                        }
+                when X::Comp::FailGoal {
+                    if $!multi-line-enabled && .pos == code.chars {
+                        return needs_more_input();
+                    } else {
```
hoelzro: Nice catch!

src/core/REPL.pm
```diff
@@ -277,27 +282,118 @@ do {
...
+            return $result;
+        }
+
+        method repl-eval($code, *%adverbs) {
+            my $needs_more_input = False;
+            %adverbs<needs_more_input> := sub () {
+                $needs_more_input = True;
+            };
+            my $result := self.partial-eval($code, %adverbs);
+            if $needs_more_input {
```
hoelzro: This isn't a blocker for your work; just something to keep in mind for the future, but since `repl-eval`'s invocation is completely contained within the Rakudo project, we have more flexibility in how we indicate
"more input needed" to the caller rather than being forced to use a sentinel value.

awwaiid:
@hoelzro RE removing `interactive_result` and `interactive_exception` from Compiler.nqp -- my idea here is to fully embrace REPL.pm and never fallback to nqp's REPL in any circumstance -- if it doesn't load I think we should die and not provide a REPL (and maybe I should have it die/exit when it fails to load).

Are there some particular circumstances where the p6-level REPL might/should not work that I'm not realizing?

src/core/REPL.pm
awwaiid: Yep. Similarly I think my next goal is going to be to remove all nqp:: from REPL.pm.

hoelzro:
@awwaiid It *should* always work; I've just been guarding against breakage so that users aren't without a REPL if I had overlooked something.  As REPL6 gets more mature, we can rely on it more.

awwaiid:
Fear not! At this point I don't know of any cases where it doesn't work, so I propose the time is now! :)

awwaiid:
One regression: This PR doesn't watch stdout and skip printing the output if anything the user did has stdout written. How important is that?

coke:
@awwaiid - I'd like to keep that feature if we can.

hoelzro:
@awwaiid I'd never noticed that feature, but it sounds like it would be nice to keep if we can

jnthn:
:+1: to keeping the "did we output anything" feature from me also; I've found it quite nice.

awwaiid:
Cool, it was easy to add that back in -- now it watches STDOUT just like it did in the nqp land.

src/core/REPL.pm
```diff
@@ -258,46 +257,137 @@ do {
...
             my $self = self.bless();
             $self.init(compiler, $multi-line-enabled);
             $self = mixin-line-editor($self);
-
             $self
         }
 
         method init(Mu \compiler, $multi-line-enabled) {
-            $!compiler = compiler;
+            $!compiler := compiler;
```
awwaiid: @hoelzro Switching this to a hard-bind lets perl6-j work!

awwaiid:
Found another bug -- doing `&:[+]` displays an error about string context (but then goes on to print out the .gist without issue). Will fix.

awwaiid:
Fixed the string context warning.

awwaiid:
@hoelzro what is the suggested strategy for getting this merged? I can put back in the fallback mode if that would help.

hoelzro:
@awwaiid Sorry for the delay; if you put the fallback code back in, I don't see why we couldn't merge this tomorrow and see how it does for people before 2016.05 goes out!

awwaiid:
@hoelzro I replaced the fallback, and tested by purposely renaming `class REPL` to `class REPLx` to see it fail.

    $ ./perl6
    To exit type 'exit' or '^D'
    Couldn't load Rakudo REPL.pm: Undeclared name:
        REPL used at line 1. Did you mean 'Real', 'REPLx'?
    
    Falling back to nqp REPL.
    > 5+5
    10
    > sub foo($x, $y) { $x + $y }
    sub foo ($x, $y) { #`(Sub|80135880) ... }
    > foo(2, 5)
    7
    > die "hmm"
    hmm
    
    > 

I have it clearly label that it is falling back, so that hopefully we can get some reports if this ever happens.

PR #739, opened 2016-04-10
Match infix eqv

Xliff:
infix:<eqv> implementation for Match objects.

### DIFF
--- a/src/core/Match.pm
+++ b/src/core/Match.pm
@@ -62,6 +62,17 @@ my class Match is Capture is Cool {
         }
     }
 
+    multi sub infix:<eqv>(Match:D $a, Match:D $b) {
+        [&&] (
+            $a.to   eqv $b.to,
+            $a.from eqv $b.from,
+            $a.orig eqv $b.orig,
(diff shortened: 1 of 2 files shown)


lizmat:
although .to/.from/,orig are easy to check, I'm not sure whether they should be part of eqv for Matches

Also, to short-circuit Matches eqvied with themselves, a very cheap a =:= b might be in order

Xliff:
@lizmat, you are correct, I had the identity check in an earlier version but it was lost and I forgot to include it in the re-implementation. I will add that back in, now. Thanks for the reminder.

in your opinion, what should be part of eqv for Matches?

lizmat:
Feels to me this should be more about *what* is matched, not *where* it was matched?

moritz:
@lizmat I disagree. `eqv` is a pretty strict comparison. If you an distinguish two objects by looking at their attributes or their `.perl` output, they shouldn't be equivalent.

moritz:
Thanks @Xliff, squash-merged.

PR #741, opened 2016-04-13
Move permutations() back to List.pm

MasterDuke17:


niner:
What's the rationale for this change? Your commit message only mentions what the commit does (which is rather obvious from the patch itself) but not why you do it.

lizmat:
perhaps because it apparently doesn't need native arrays anymore?

Which makes me wonder why the native arrays have been removed?

MasterDuke17:
I was looking at the combinations() implementation in List.pm, and then
went to look at permutations() and it was off in operators.pm. I looked
back at the history and commit messages (and yes, mine was pretty bad) and
the reason didn't seem valid anymore, so I moved it back to List.pm, where
it passed all tests.

PR #742, opened 2016-04-13
Duration Improvements

pprince:
This is my first attempt to contribute something to p6; feel free to point out any errors/misunderstandings on my part, or further improvements; I am happy to incorporate suggestions and re-submit.

Description:
---
    Add missing ops * / for Duration; fixes RT #127339
    
    Also see corresponding PR/commits against perl6/roast.
    
    Has tests to cover new functionality, and all tests pass.
    
    Still, I am new to Perl6, and look forward to one or more
    rounds of requested changes before this gets into Perl6.
---
    Move Duration and Instant closer to S02 spec.
    
    Duration and Instant now behave (more?) like value types.
    
    Also, Duration now becomes a Rat in explicit numeric context.
    i.e., +$dur will return a Rat, however $dur1/$dur2 still performs
    Num/Num artithmetic, even when it loses precision :(
---

### DIFF
--- a/src/core/Duration.pm
+++ b/src/core/Duration.pm
@@ -10,17 +10,38 @@ my class Duration is Cool does Real {
     method narrow(Duration:D:) { $!tai.narrow }
 
     multi method Str(Duration:D:) { ~$.tai }
-    multi method Numeric() { $!tai }
+    multi method Numeric() { $!tai } # (Unary plus)
 
     multi method WHICH(Duration:D:) { self.^name ~ '|' ~ $!tai.WHICH }
 
     multi method perl(Duration:D:) { "Duration.new({$.tai.perl})" }
 }
 
+# For Addition and Subtraction operations, dimensionless number operands are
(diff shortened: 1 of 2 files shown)


pprince:
**SEE ALSO:  [perl6/roast PR#112](https://github.com/perl6/roast/pull/112)**

pprince:
I am going to withdraw this pair of PR's for now, as I:

1. have re-considered the ideal corner-case behavior,
2. would like to submit cleaner PR's that don't e.g. leave commit-squashing up to the person merging,
3. would like to not litter roast with extraneous test cases, and
3. would like to simultaneously offer a PR to update/clarify the Duration-related parts of S02 & S32.

You should look forward to my re-submssion... I believe we can specify Duration such that everyone gets what they want, no one gets any surprises, "weird" cases Just Work™, and all while avoiding having to do anything you'd call complicated.  Wish me luck :)

PR #743, opened 2016-04-13
Implement alternate ways to call `subtest` from Test.pm6

zoffixznet:
This implements the proposal in [RT #127890](https://rt.perl.org/Ticket/Display.html?id=127890) (both `desc, code` and `desc => code` forms)

In addition, it adds a `diag $desc` at the start of the subtest, to make it easier to see it in the output.

I also piggy-backed on already-existing tests to test out the functionality. They also had an old name for `dies-ok`, which I fixed.

### DIFF
--- a/lib/Test.pm6
+++ b/lib/Test.pm6
@@ -289,7 +289,9 @@ sub skip-rest($reason = '<unknown>') is export {
     $time_before = nqp::time_n;
 }
 
-sub subtest(&subtests, $desc = '') is export {
+multi sub subtest(Pair $what)            is export { subtest($what.value,$what.key) }
+multi sub subtest($desc, &subtests)      is export { subtest(&subtests,$desc)       }
+multi sub subtest(&subtests, $desc = '') is export {
     _push_vars();
     _init_vars();
     $indents ~= "    ";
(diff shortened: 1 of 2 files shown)


ugexe:
It might be worth pointing out #684 for notes on when this functionality was merged and reverted previously

zoffixznet:
Ah, thanks. Do NOT merge this yet. I'll take a crack at it tonight, to see why the failure is occuring.

zoffixznet:
Here's the results on my investigation: the issue is with that particular failing test in the roast and I submitted [a PR to fix it](https://github.com/perl6/roast/pull/113). When that PR is merged, it's safe to merge this PR.

The issue is the test uses `throws-like` sub to test whether `({nextsame})()` dies with `X::NoDispatcher`
exception. Under the hood, `throws-like` actually uses `subtest` to perform an `EVAL`. Now, depending on whether `subtest` is a multi, a different exception is returned, which is why this PR's change of `subset` to a multi makes the test fail:
```
<ZoffixWin> m: multi foo { EVAL 'nextsame', context => CALLER::CALLER:: }; foo; CATCH { default { .^name.say }};
<camelia> rakudo-moar a45224: OUTPUT«X::ControlFlow::Return␤»
<ZoffixWin> m: sub foo { EVAL 'nextsame', context => CALLER::CALLER:: }; foo; CATCH { default { .^name.say }};
<camelia> rakudo-moar a45224: OUTPUT«X::NoDispatcher␤»
```

jnthn++ also explained why the `::Return` exception is thrown:

```
<jnthn> nextsame and nextwith are equivalent to return callsame and return callwith
<jnthn> Which is why the "return outside of routine" thing
```

moritz:
Thanks @zoffixznet, I've squash-merged this PR.

PR #744, opened 2016-04-16
Fix RT #127883: Missing line/column number information when using a missing module

azawawi:
For more information, please see https://rt.perl.org/Public/Bug/Display.html?id=127883


azawawi:
Fixed the attribute name to a more descriptive name `source-line-number` after getting feedback from moritz++.

azawawi:
Thanks :+1:

PR #745, opened 2016-04-17
Call :done/:quit on Supplies from Proc::Async

Leont:
Currently :done and :quit aren't called on stdout/stderr supplies from Proc::Async, this is a major blocker for #578


timo:
this is probably correct. thanks a lot!

PR #747, opened 2016-04-18
Avoid NullPointerException in method 'sort'

usev6:


peschwa:
This allows R-J to build again, which is good, but it seems a bit brittle to fix things in Rakudo-land just because one backend is a bit more picky.

On the other hand, moar and jvm both exhibit the same behavior, but the moar implementation of p6sort apparently deals with it somehow. Consider the following:

    $ ./perl6-j -e'use nqp; \
    my \indices := nqp::setelems(nqp::list, 2); \
    my int $i; nqp::bindpos(indices, $i, nqp::decont($i)) \
        while nqp::islt_i($i = nqp::add_i($i, 1), 2); say indices[0]'
    (Mu)

    <psch> m: use nqp; my \indices := nqp::setelems(nqp::list, 2); my int $i; nqp::bindpos(indices, $i, nqp::decont($i)) while nqp::islt_i($i = nqp::add_i($i, 1), 2); say indices[0]
    16:32 <camelia> rakudo-moar 1aabef: OUTPUT«(Mu)␤»

On those grounds I'm okay with passing proper data in the first place, instead of relying on the backend to make it work anyway.

peschwa:
Well, AppVeyor didn't finish, but I'm merging anyway. Travis seems the more trustworthy of those two, currently.

peschwa:
Thanks!

PR #748, opened 2016-04-19
Avoid NPE when parameter has no coerce_type

usev6:
This lead to NPE in some files under S32-trig/
One golfed example for dying code:

  class Foo is Real { }

### DIFF
--- a/src/core/Parameter.pm
+++ b/src/core/Parameter.pm
@@ -389,21 +389,23 @@ multi sub infix:<eqv>(Parameter \a, Parameter \b) {
     return True if a =:= b;
 
     # different nominal or coerce type
+    my $acoerce := nqp::getattr(a,Parameter,'$!coerce_type');
+    my $bcoerce := nqp::getattr(b,Parameter,'$!coerce_type');
     return False
       unless nqp::iseq_s(
           nqp::getattr(a,Parameter,'$!nominal_type').^name,
           nqp::getattr(b,Parameter,'$!nominal_type').^name
         )
       && nqp::iseq_s(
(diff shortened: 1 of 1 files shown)


usev6:
jnthn++ mentioned -- as an alternative --, that we could introduce the concept of VMNull from Moar to reduce the chance of busting rakudo-j with NPE like in this case: http://irclog.perlgeek.de/p6dev/2016-04-18#i_12357833

psch++ was positive on that idea as well: http://irclog.perlgeek.de/perl6/2016-04-19#i_12359494

I'll leave this PR open, since it would make a lot of tests pass (again) on rakudo-j.

usev6:
Actually, the changes to lines 405 and 406 (s/getattr/getattr_i/) are no longer needed (since nqp commit 5981bf0439, I think).

Is it acceptable to add another commit removing those changes or better to open a new PR?

peschwa:
I'd be fine with another commit to this PR adjusting the relevant line. 

Also, in reference to the VMNull proposal (because I saw it come up in the [irclog](http://irclog.perlgeek.de/perl6/2016-04-28#i_12406710)) I also support merging this PR now. I wasn't aware it was that many tests that we can pass again, and VMNull isn't as approachable as I had hoped.

usev6:
Actually I was a bit unprecise in the linked irc discussion: We have about 5500 failing tests, because whole test files are dying due to the NPE. One could try to find and fudge the relevant tests, to make those test files run to completion -- that would probably help as well.

peschwa:
Thanks!

PR #749, opened 2016-04-19
Disable JVM CI builds on Windows (AppVeyor) for performance reasons

azawawi:
JVM build takes way too long and an AppVeyor build worker is limited to 60 minutes.


azawawi:
Thanks :+1:

PR #751, opened 2016-04-19
Make usage text fit in 80 columns

sjn:
Also, update running.pod usage text example to match actual usage text.

### DIFF
--- a/docs/running.pod
+++ b/docs/running.pod
@@ -1,3 +1,5 @@
+=begin pod
+
 =head1 NAME
 
 perl6 - Rakudo Perl 6 Compiler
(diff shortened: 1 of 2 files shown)


docs/running.pod
```diff
@@ -18,15 +20,18 @@ compiled code.
...
+  --optimize=[level]   use the given level of optimization (0..3)
   --encoding=[mode]    specify string encoding mode
-  --output=[name]      specify name of output file
+  -o, --output=[name]  specify name of output file
   -v, --version        display version information
   --stagestats         display time spent in the compilation stages
   --ll-exception       display a low level backtrace on errors
   --profile            write profile information as HTML file (MoarVM)
-  --profile-filename=[name]
-                       specify an alternate profile output file
```
niner: --profile-filename would be the only option with an argument where we don't show that it takes an argument. This may be confusing.

sjn: Yeah, good point. I'll make another pull request with that, and look through the code to see if there are any other useful/important command-line options that might be worth adding.

PR #752, opened 2016-04-20
Ability to use a custom debugger module

Altai-man:
I looked at how Perl 5 `Devel::Trace` works and saw that it uses a compiler hook for a custom debugger module. But now in Rakudo we have only `Debugger::UI::CommandLine` module hardcoded. So I decided to fix it to gain an ability to write proper `Devel::Trace` implementation for Perl 6 too.

This commit adds a support for a new command line argument: `-Dsomedebuggermodule`, like
`perl6-debug-m -DCustomDebugger file.p6`
In the case if we don't have this argument, we try to use default `Debugger::UI::CommandLine`.

Also my knowing of proper nqp style is less than awesome, so suggestions are strongly appreciated.

It passes `make spectest` and I doubt it reduces performance of perl6-debug-m.

### DIFF
--- a/src/perl6-debug.nqp
+++ b/src/perl6-debug.nqp
@@ -484,6 +484,7 @@ sub MAIN(*@ARGS) {
 	if @ARGS[$i] ~~ /^\-D/ {
 	    $debugger := "-M" ~ nqp::substr(@ARGS[$i], 2);
 	    nqp::splice(@ARGS, [], $i, 1);
+	    last;
 	}
 	$i++;
     }
--- a/src/perl6-debug.nqp
+++ b/src/perl6-debug.nqp
@@ -479,12 +479,19 @@ sub MAIN(*@ARGS) {
 
     # Force loading of the debugger module.
     my $debugger;
-    if @ARGS[1] ~~ /^\-D/ {
-	$debugger := "-M" ~ nqp::substr(@ARGS[1], 2);
-	@ARGS.shift();
-    } else {
+    my $i := 1;
+    while @ARGS[$i] ~~ /^\-/ {
+	if @ARGS[$i] ~~ /^\-D/ {
+	    $debugger := "-M" ~ nqp::substr(@ARGS[$i], 2);
+	    nqp::splice(@ARGS, [], $i, 1);
+	}
+	$i++;
+    }
+
+    if !(nqp::defined($debugger)) {
 	$debugger := '-MDebugger::UI::CommandLine';
     }
+
     my $pname := @ARGS.shift();
     @ARGS.unshift('-Ilib');
     @ARGS.unshift($debugger);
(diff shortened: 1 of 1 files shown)


coke:
Just based on code review, not running the patch, it seems reasonable to me. 👍

src/perl6-debug.nqp
```diff
@@ -476,13 +476,20 @@ sub MAIN(*@ARGS) {
     
     # Set up END block list, which we'll run at exit.
     nqp::bindhllsym('perl6', '@END_PHASERS', []);
-    
+
     # Force loading of the debugger module.
+    my $debugger;
+    if @ARGS[1] ~~ /^\-D/ {
+	$debugger := "-M" ~ nqp::substr(@ARGS[1], 2);
+	@ARGS.shift();
```
jnthn: This seems a little fragile, in so far as:

    perl6-debug -MSomeModule -Ilib -DDebug::Trace foo.p6

Would fail to spot the -DDebug::Trace, because it's not the first option. Maybe scanning through the list until a non-option is found (something not starting with -) would be better. Then splice that argument out (`nqp::splice`).

Altai-man:
@jnthn,
I really not intended to accept a debugger module if it's not the first option, that's why I just hardcoded the position of argument. Otherwise, we surely can loop over ARGS, find our `-D`, extract and push it into `$debugger`. Will play with it a bit now...
But, I'm not sure about understanding of your variant, why would we need to find a non-option(filename, obviously)? We just need to search anything with `-D` start characters, extract/put and try to use it, aren't we?

jnthn:
@Altai-man Because in an invocation such as:

    perl6 -Ilib foo.p6 --some-option

That `--some-option` is going to the script, not to be processed by the compiler. So, we should not consider arguments after the script name; they're not "ours".

Altai-man:
Oh, this case really slipped off my mind, sorry. I'll prepare better version now.

Altai-man:
Some tests of the new commit:
`./perl6-debug-m -Ilib -DTracer  example.p6 --some-option --another`
`./perl6-debug-m -DTracer -Ilib example.p6 --some-option --another`
passes options to the script, invokes Tracer.
`./perl6-debug-m -Ilib example.p6 --some-option --another`
passes options to the script, invokes default debugger module.

src/perl6-debug.nqp
```diff
@@ -476,13 +476,27 @@ sub MAIN(*@ARGS) {
...
     nqp::bindhllsym('perl6', '@END_PHASERS', []);
-    
+
     # Force loading of the debugger module.
+    my $debugger;
+    my $i := 1;
+    while @ARGS[$i] ~~ /^\-/ {
+	if @ARGS[$i] ~~ /^\-D/ {
+	    $debugger := "-M" ~ nqp::substr(@ARGS[$i], 2);
+	    nqp::splice(@ARGS, [], $i, 1);
```
vendethiel: this should probably `last;` as well?

Altai-man: To save several checks time? Yeah, seems reasonable.

PR #754, opened 2016-04-23
fail() when trying to concatenate two strings and the resulting strin…

MasterDuke17:
…g would be too large.


lizmat:
Feels to me we shouldn't pollute such a hot code path at the Perl 6 level.  If this is indeed a problem of a backend, then the backend should be generating the exception, *not* the client of the backend.

I'm leaning towards closing this pull request.  Perhaps this should be turned into an RT ticket instead?

lizmat:
http://irclog.perlgeek.de/p6dev/2016-04-23#i_12381309

PR #755, opened 2016-04-23
Use less Perl 5 in our build process

Leont:
The first patch renames $(PERL) to $(PERL5) in the Makefiles, for clarity.

The second patch replaces ExtUtils::Command calls by their native equivalents, when possible.

The third patch converts create-moar-runner.pl from perl5 to perl6


lizmat:
I get a fail on "make install" with:

tools/build/create-moar-runner.pl "/Users/liz/Github/rakudo.moar/install/bin/moar" perl6.moarvm /Users/liz/Github/rakudo.moar/install/bin/perl6-gdb-m "/Users/liz/Github/rakudo.moar/install/share/perl6/runtime" "gdb" "/Users/liz/Github/rakudo.moar/install/share/nqp/lib" "/Users/liz/Github/rakudo.moar/install/share/perl6/lib" "/Users/liz/Github/rakudo.moar/install/share/perl6/runtime"
sh: tools/build/create-moar-runner.pl: Permission denied

Leont:
It seems I had missed something (and hadn't tested a make install), fixed pushed now.

coke:
531a044db162e2d0189c22df91ff2af8daac81a4 breaks the build.

It replaces perl5 with a call to perl6, but doesn't properly capture all the dependencies needed for this replacement; if you "make -j perl6-m", you'll see that it eventually dies with a missing file.

coke:
I reverted that commit; it will need review before merging.

Leont:
First guess (will test tomorrow): those rules need to depend on the setting too.

PR #756, opened 2016-04-25
elaborate on different kinds of profiles for --profile

timo:
I'd like someone to review this; it's now the only option that will wrap lines and I wasn't sure how to format it.


timo:
I imagine an alternative way to format what --profile looks like could be like this:

```
    --profile=instrumented      do performance measurements
    --profile=heap              do heap snapshots
````

and perhaps then the --profile-filename could go after each of these? perhaps indented a little bit?

PR #757, opened 2016-04-25
Make short-id2class accessible

MadcapJake:
Previously it was impossible to add a custom CUR but with this change, you can add your CUR to the registry of short-ids.

I know that the plan is to change how CURs are loaded so this is a stopgap until that implementation.

### DIFF
--- a/src/core/CompUnit/RepositoryRegistry.pm
+++ b/src/core/CompUnit/RepositoryRegistry.pm
@@ -207,7 +207,7 @@ class CompUnit::RepositoryRegistry {
         }
     }
 
-    sub short-id2class(Str:D $short-id) {
+    our sub short-id2class(Str:D $short-id) {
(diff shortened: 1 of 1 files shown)


niner:
This makes short-id2class a public interface that we will have to support. I'd much rather see work towards a real designed solution.

MadcapJake:
We could mark it as deprecated. I think there needs to at least be a solution until the real one is made available. Right now it just doesn't work without this.

MadcapJake:
Perhaps if it was moved it to an experimental package we could merge this.  @niner, would that be acceptable? This way it would imply that the feature should not be depended on because it will be changing.
```perl6
use experimental :cur;
use CompUnit::Repository::GX;
use Foo::Bar;
```

JJ:
Closed after opening Raku/problem-solving#189

PR #764, opened 2016-05-09
Some optimizations to List.pm

MasterDuke17:


lizmat:
for the foreseeable future, ++$i will be faster than $i++ in sink context

PR #766, opened 2016-05-13
remove false statement until it is made true

tbrowder:


coke:
https://rt.perl.org/Ticket/Display.html?id=128087

PR #769, opened 2016-05-16
Prevent a segfault on deref of null typed pointer

jonathanstowe:
Stops

perl6 -e 'use NativeCall; my $a = Pointer[int32].new; say $a.deref'
Segmentation fault (core dumped)

Some libraries may quite happily not set a passed pointer if there
isn't any data, this just handles it in a nicer way (obviously
one could do what this patch does in the user code.)

### DIFF
--- a/lib/NativeCall.pm6
+++ b/lib/NativeCall.pm6
@@ -399,18 +399,17 @@ multi refresh($obj) is export(:DEFAULT, :utils) {
     1;
 }
 
-sub nativecast($target-type, $source) is export(:DEFAULT) {
-    if $target-type ~~ Signature {
-        my $r := sub { };
-        $r does Native[$r, Str];
-        nqp::bindattr($r, Code, '$!signature', nqp::decont($target-type));
-        nqp::bindattr($r, $r.WHAT, '$!entry-point', $source);
-        $r
-    }
-    else {
-        nqp::nativecallcast(nqp::decont($target-type),
-            nqp::decont(map_return_type($target-type)), nqp::decont($source));
-    }
+multi sub nativecast(Signature $target-type, $source) is export(:DEFAULT) {
+    my $r := sub { };
+    $r does Native[$r, Str];
+    nqp::bindattr($r, Code, '$!signature', nqp::decont($target-type));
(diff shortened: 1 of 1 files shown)


salortiz:
:+1:

moritz:
Shouldn't it rather be an exception to dereference an uninitialized pointer?

jonathanstowe:
Possibly but a common use for a pointer is where it's passed to a function to fill and the function returns a status, it is entirely possible that it indicates okay but has no data so doesn't populate the pointer.

moritz:
But shouldn't you check before dereferencing then? It sounds to me like the extra check wouldn't be very expensive, but debugging a potential programming error where a pointer isn't supposed to be NULL would be much more work.

jonathanstowe:
I'm relaxed either way, but it shouldn't segfault whatever, it would be better if there was a Bool coercion on the Pointer to shorten the check if the exception is chosen.

moritz:
:+1: to a better Bool coercer.

jonathanstowe:
Shall I just change this so it throws an exception, and add a Bool?

arnsholt:
Yeah, I think `.deref` should probably throw an exception when dereferencing a NULL. Alternatively, it could return failure, which is much the same thing of course.

jonathanstowe:
:+1: A failure sounds like a good idea :) I'll do that.

jonathanstowe:
Gah, should have rebased against nom :-\

src/core/Match.pm
```diff
@@ -96,6 +96,19 @@ my class Match is Capture is Cool {
     }
 }
 
+multi sub infix:<eqv>(Match:D $a, Match:D $b) {
+    [&&] (
+	$a 	=:= $b,
```
vendethiel: can `$a =:= $b` (container check, right?) be True while `$a.to` and `$b.to` are different?

jonathanstowe:
@vendethiel that is already in nom afaik, I screwed up the rebase and added these to this PR by mistake - https://github.com/rakudo/rakudo/pull/769/commits/a2188ecdcfffccd68411a51328798afd95d391ca is the only pertinent one :)

jonathanstowe:
I'm going to do this again

PR #770, opened 2016-05-18
Remove superfluous diag() in subtest

ugexe:
Test.pm `diag()` is used for displaying errors everywhere else

See:
https://gist.github.com/ugexe/0fd759225aefbeac9c08d1cc87603206

PR #771, opened 2016-05-18
Remove diag() from subtest()

zoffixznet:
It does not parse correctly and the default verbosity of prove() already includes the name of the subtest, even if at the end of its run. Relevant IRC conversation: http://irclog.perlgeek.de/perl6/2016-05-18#i_12502193


zoffixznet:
There's already #770

PR #772, opened 2016-05-22
Fix RT #128073

MasterDuke17:
Also some further cleanup of IO::Path.dir to remove some VM specific code.


MasterDuke17:
Requires NQP PR #287

PR #773, opened 2016-05-22
Implement RT #125489

MasterDuke17:
Adds a .mode method to IO::[Path,Handle,Special] that returns the file's permissions as either an Int or Str.

### DIFF
--- a/src/core/IO/Handle.pm
+++ b/src/core/IO/Handle.pm
@@ -800,9 +800,10 @@ my class IO::Handle does IO {
     method r(IO::Handle:D:) { $!path.r }
     method w(IO::Handle:D:) { $!path.w }
     method x(IO::Handle:D:) { $!path.x }
-    method modified(IO::Handle:D:) { $!path.modified }
-    method accessed(IO::Handle:D:) { $!path.accessed }
-    method changed(IO::Handle:D:)  { $!path.changed  }
+    method modified(IO::Handle:D:)   { $!path.modified    }
+    method accessed(IO::Handle:D:)   { $!path.accessed    }
+    method changed(IO::Handle:D:)    { $!path.changed     }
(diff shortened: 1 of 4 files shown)


lizmat:
I like it in general, but I'm not so sure about the :Str parameter.  This sorta seems like a break from the rest.

MasterDuke17:
What about getting rid of it and leaving just the sprintf path, but with '.Int' on the end?

lizmat:
I'm not sure what that would give us?

MasterDuke17:
It might just be me, but I've never in my life wanted the mode/permissions as an unmasked number (so I would argue against returning the raw value of the nqp::stat() call). But I guess I don't really have strong feelings about whether the method should return an Int or a Str.

labster:
Why not return a Str but Int mixin?

lizmat:
Isn't that just an IntStr ?

MasterDuke17:
I "Cannot mix in non-composable type Int into object of type Str". I can do Int but Stringy, but I'm not sure what that buys me.
```
my int $mode = nqp::stat($!abspath, nqp::const::STAT_PLATFORM_MODE) +& 0o7777;
return new($mode, sprintf('%04o', $mode))
```
This seems nice, but Rakudo won't compile, it seems it doesn't know about IntStr at that point.

labster:
> Isn't that just an IntStr ?

Maybe.  Let me look...
/me files perl6/doc#544

I really need to learn modern Perl 6.

labster:
The IntStr types look OK, but you still have these signatures:

> mode(IO::Handle:D: :$Str)

MasterDuke17:
Right, thanks for catching that.

MasterDuke17:
Tests added in perl6/roast PR #126 (https://github.com/perl6/roast/pull/126). Documentation added in perl6/docs PR #565 (https://github.com/perl6/doc/pull/565).

PR #775, opened 2016-05-23
Fix RT #128214

MasterDuke17:
While the docs could be read to imply that finding a non-existent part during IO::Path.resolve isn't an error, it is an error in regular Bash directory traversal.

### DIFF
--- a/src/core/IO/Path.pm
+++ b/src/core/IO/Path.pm
@@ -161,14 +161,13 @@ my class IO::Path is Cool {
             # Normal part, set as next path to test
             my str $next = nqp::concat($resolved, nqp::concat($sep, $part));
 
-            # Path part doesn't exist; handle rest in non-resolving mode
+            # Path part doesn't exist, so bail out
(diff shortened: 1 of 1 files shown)


smls:
I think this needs an explicit design decision first - see [the discussion on the RT page](https://rt.perl.org/Ticket/Display.html?id=128214#txn-1402333).

MasterDuke17:
How do I comment on the RT page? The 'reply' and 'comment' buttons don't do anything.

coke:
If you're having trouble with RT, please ask on #perl6 on IRC on freenode; if it turns out there's a bug, someone can forward you to the rt admins.

zoffixznet:
The author replied on the RT ticket. Removing `pending user` tag.

MasterDuke17:
@zoffixznet Do you think you could take another look at this during your grant work?

zoffixznet:
This has now [been resolved](https://rt.perl.org/Ticket/Display.html?id=128214#txn-1460586).

PR #777, opened 2016-05-29
Fix RT#128284: warnings on failing Test::cmp-ok

zoffixznet:
Instead of attempting to stringify the matcher, we try to call .name (that will give us name for named subs) or call ^.name, which will give us the name of the thing, such as Block, Whatever, etc.

https://rt.perl.org/Ticket/Display.html?id=128284


peschwa:
Thanks!

PR #778, opened 2016-06-02
Access packages in their CUR::FS $base

ugexe:
Makes `META6.json` and `META.info` follow the same behavior in regards to access modules in a package $base
See: http://irclog.perlgeek.de/perl6-toolchain/2016-06-02#i_12593703


lizmat:
Why the ? in "unless ?$found" ?  Feels a bit superstitious to me, or am I missing something?

ugexe:
Feel free to remove it (or tell me to), that's just the way I usually would write it.

Realistically this is a temporary thing anyways, since the method itself needs to be refactored entirely (for instance: slurping the META file once for each unloaded module requested, or possibly setting $found multiple times in the for loop).

PR #780, opened 2016-06-04
Avoid NPE for arrays with deleted elements

usev6:
fixes RT #128320

### DIFF
--- a/src/core/List.pm
+++ b/src/core/List.pm
@@ -476,7 +476,7 @@ my class List does Iterable does Positional { # declared in BOOTSTRAP
                 method push-all($target) {
                     my int $elems = nqp::elems($!reified);
                     my $no-sink;
-                    $no-sink := $target.push(nqp::atpos($!reified,$!i))
+                    $no-sink := $target.push(nqp::ifnull(nqp::atpos($!reified,$!i),Nil))
                       while nqp::islt_i($!i = nqp::add_i($!i,1),$elems);
                     IterationEnd
                 }


lizmat:
I think this patch is fixing the wrong thing.  This is the *List* iterator failing for an *Array*.  The Array.iterator should have been called, and that one *does* know how to handle nulls.

So the problem is really: *WHY* is the List.iterator class used for an Array on the JVM ???  Perhaps this is indicative of a deeper issue on the JVM???

usev6:
It's likely that I'm wrong, but I think that List.iterator is used because we use '+values' in the signature of 'sub map' (one-argument slurpy binding: https://design.perl6.org/S06.html#One-argument_slurpy_binding). This was made so with commit d9c21e99f5

To demonstrate my point, I made the following change -- which also made the NPE go away (but probably other things don't work as expected with this change).

```
diff --git a/src/core/Any-iterable-methods.pm b/src/core/Any-iterable-methods.pm
index 6f1ca40..57f7ef9 100644
--- a/src/core/Any-iterable-methods.pm
+++ b/src/core/Any-iterable-methods.pm
@@ -1368,7 +1368,7 @@ multi sub infix:<minmax>(+args) { args.minmax }
 sub minmax(+args, :&by = &infix:<cmp>) { args.minmax(&by) }
 
 proto sub map(|) {*}
-multi sub map(&code, +values) { my $laze = values.is-lazy; values.map(&code).lazy-if($laze) }
+multi sub map(&code, *@values is raw) { my $laze = @values.is-lazy; @values.map(&code).lazy-if($laze) }
 
 proto sub grep(|) {*}
 multi sub grep(Mu $test, +values, *%a) {
```

usev6:
As a small experiment I did s/Nil/Int/ in my patch:

```
diff --git a/src/core/List.pm b/src/core/List.pm
index 0c65a6f..d4646ed 100644
--- a/src/core/List.pm
+++ b/src/core/List.pm
@@ -476,7 +476,7 @@ my class List does Iterable does Positional { # declared in BOOTSTRAP
                 method push-all($target) {
                     my int $elems = nqp::elems($!reified);
                     my $no-sink;
-                    $no-sink := $target.push(nqp::atpos($!reified,$!i))
+                    $no-sink := $target.push(nqp::ifnull(nqp::atpos($!reified,$!i),Int))
                       while nqp::islt_i($!i = nqp::add_i($!i,1),$elems);
                     IterationEnd
                 }
```

This compiles and gives the following result on rakudo-m:

$ perl6-m -e 'my @a = 0..1; @a[0]:delete; say map { $_ }, @a'
((Int) 1)

So it's not specific for rakudo-j to use the List iterator in this example.

However, my patch could very well be bogus nonetheless.

usev6:
We discussed this issue on #perl6-dev (http://irclog.perlgeek.de/perl6-dev/2016-09-19#i_13238477) and arnsholt++ also suggested to fix this problem in the JVM specific code (add a check for null at the appropriate place). I'm closing this pull request.

PR #781, opened 2016-06-09
Remove as many uses of $no-sink as possible

MasterDuke17:
This is an exploration of whether the use of $no-sink can be removed.See http://irclog.perlgeek.de/perl6-dev/2016-06-08#i_12628220 for some discussion. As the PR stands, it passes all the spec tests.

Removing the one remaining use in List.pm caused a whole bunch of failures (I'd guess about 50 different test files).

Removing the one remaining use in Any-iterable-methods.pm caused failures in three test files:
- t/spec/S04-statements/label.rakudo.moar
- t/spec/S04-statements/last.t
- t/spec/S04-statements/redo.t

### DIFF
--- a/src/core/Any-iterable-methods.pm
+++ b/src/core/Any-iterable-methods.pm
@@ -161,7 +161,6 @@ Did you mean to add a stub (\{...\}) or did you mean to .classify?"
                           );
                         ) unless $!did-init;
 
-                        my $no-sink;
                         my int $redo;
                         my int $running = 1;
                         my $value;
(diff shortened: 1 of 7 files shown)


lizmat:
Thanks for pursuing this!

All no-sink's have been removed, and replaced by structures that are guaranteed to still not sink in the future.  Most notably, postfix if/while/until currently do not sink, which could be considered a bug.

Related commits:
   d292d76daa4214d
  57d66ee76aa727b7ed
  3b9b6132dd351081e0b29
  2095ed65d1858864076
  8cfa6c76ece381e900
  8887a7c4caae98c8ecb6
  87e91ccc64815479f36
  9cfcb96a2fe77cc9ef9
  74d7ac05eef72f1dbc833ec
  5b4470bd19702186e08b
  652e8f0f6a872aee75e

PR #783, opened 2016-06-11
Rework is-approx() in Test.pm6

zoffixznet:
This PR reworks the implementation of `is-approx` to fix bugs, improve usefulness, and improve displayed messages when the test fails. The implementation follows my proposal I shared a couple of weeks ago, salient parts of which I'm replicating below. First, I'll write some of the responses/explanations for discussions over the last year with regard to `is-approx`:

* **Should `is-approx` use `≅`?** The `≅` is already available to users via `cmp-ok` and `is-approx` can avoid ≅'s more-complex behaviour. So I've not used it.
* **Should the default tolerance be relative?** The absolute tolerance is what's described in speculation, the documentation, and it's much easier to explain and understand. If someone is already using `is-approx`, they'd expect the behaviour that was specced/documented. The relative tolerance is still available via `:rel-tol` named parameter.
* **Should is-approx have different behaviour for super small values?** One of the old commits used [this Python function](https://www.python.org/dev/peps/pep-0485/) as justification for implementing more complex behaviour for `is-approx`. That document is 13 pages long. Considering the new implementation gives the user full control over relative and absolute tolerances, I don't see any good reason to make the default behaviour complex, in an attempt to guess the usecase of the user's code.

# Incompatible changes

`is_approx` now triggers deprecation warning.

At this point, the only behaviour present in speculation and documentation is `is-approx $got, $expected, $desc?` that uses absolute tolerance of `1e-5`. **That behaviour is preserved.** The roast uses `is_approx` which, aside from adding a deprecation warning, this PR does not change. The spectest fully passes without any changes.

Thus, any changes of extra features would only break the code of users who learned of the extra behaviour from the source code. More over, the extra behaviour is partially broken, which makes it less likely it's actually in use anywhere.

# Bug Fixes

This PR will fix these bugs:

* [RT#128063](https://rt.perl.org/Ticket/Display.html?id=128063) *[BUG] is-approx multi that takes just two numbers never called*
* [RT#128282](https://rt.perl.org/Ticket/Display.html?id=128282) *[RFC] Use kebab case for named parameters in Test.pm6*
* [RT#128303](https://rt.perl.org/Ticket/Display.html?id=128303) *[BUG] is-approx calculates relative tolerance, instead of absolute*

# Usability Changes

* It will now be possible to use `rel-tol` or `abs-tol` separately
* On failures, the errors will be more descriptive and include tolerances

# Description of New Behaviour

The subroutine can be called in numerous ways.
`$rel-tol` is relative tolerance and `$abs-tol` is absolute tolerance:

    my Numeric ($got, $expected, $abs-tol, $rel-tol) = ...

    is-approx $got, $expected;
    is-approx $got, $expected, 'test description';

    is-approx $got, $expected, $abs-tol;
    is-approx $got, $expected, $abs-tol, 'test description';

    is-approx $got, $expected, :$rel-tol;
    is-approx $got, $expected, :$rel-tol, 'test description';

    is-approx $got, $expected, :$abs-tol;
    is-approx $got, $expected, :$abs-tol, 'test description';

    is-approx $got, $expected, :$abs-tol, :$rel-tol;
    is-approx $got, $expected, :$abs-tol, :$rel-tol, 'test description';

### No tolerance specified

If no tolerance is provided, it's assumed to be set as absolute tolerance
of `1e-5`

### Absolute tolerance specified

When absolute tolerance is set, it's used as the actual maximum value by
which the `$got` and `$expected` can differ. For example:

    is-approx 3, 4, 2; # success
    is-approx 3, 6, 2; # fail

    is-approx 300, 302, 2; # success
    is-approx 300, 400, 2; # fail
    is-approx 300, 600, 2; # fail

Regardless of values given, the difference between them cannot be more
than `2`

### Relative tolerance specified

When relative tolerance is set, the test checks relative difference between
values. Given the same tolerance, the larger the numbers given, the larger the
number they can differ by can be.

For example:

    is-approx 10, 10.5, :rel-tol<0.1>; # success
    is-approx 10, 11.5, :rel-tol<0.1>; # fail

    is-approx 100, 105, :rel-tol<0.1>; # success
    is-approx 100, 115, :rel-tol<0.1>; # fail

Both versions use `0.1` for relative tolerance, yet the first can differ
by about `1` while the second can differ by about `10`. The function used
to calculate the difference is:

                  |got - expected|
    rel-diff = ──────────────────────
               max(|got|, |expected|)

And the test will fail if `rel-diff` is higher than `$rel-tol`

### Both absolute and relative tolerance specified

When both absolute and relative tolerances are specified, each will be
tested independently, and the `is-approx` test will succeed only if both pass.

------


tbrowder:
Most excellent! I'll fiddle with a redo of the docs.

PR #793, opened 2016-06-18
Release 2016.06

zoffixznet:


zoffixznet:
Please wait before merging. There's an issue with make install when DESTDIR is set: http://irclog.perlgeek.de/perl6-dev/2016-06-18#i_12690874

zoffixznet:
The blockage has been cleared and it's now safe to merge and tag. There's still possibly an issue with NativeCall tests failing when **building an rpm package**, as reported by nine++, and I simply mentioned that in the announcement file.

PR #794, opened 2016-06-19
Add IO::[Path,Handle].mode to Changelog

MasterDuke17:


lizmat:
Alas, too late for mention in this release.

PR #796, opened 2016-06-20
Throw when $limit to split is NaN

zoffixznet:
When $limit to split is NaN, depending on which multi is called, it is
either treated as Inf, or an unboxing-to-native-int error is thrown.
This commit throws a meaningful error instead.

Fixes [RT#128034](https://rt.perl.org/Ticket/Display.html?id=128034)

### DIFF
--- a/src/core/Str.pm
+++ b/src/core/Str.pm
@@ -617,6 +617,12 @@ my class Str does Stringy { # declared in BOOTSTRAP
         my int $any = self!split-sanity($v,$k,$kv,$p);
 
         my $limit = nqp::istype($parts,Whatever) ?? Inf !! $parts;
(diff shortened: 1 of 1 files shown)


lizmat:
Somehow this approach with quietly feels wrong for such a relative hot path.  Will look at it again tomorrow after some good night sleep.

lizmat:
Fixed with 57f8f8f0faa81ee90c59 , thanks for the inspiration.

I think part of the missing info was, that NaN == NaN is indeed False, but NaN === NaN is True, so simply checking for NaN is better.  On the fly, I also usurped the handling of Whatever and used Inf as the default, so it will generally need to do less that way.

PR #798, opened 2016-06-21
Make assuming preserve return type

zostay:
Small change (possibly naïve) for making .assuming preserve the return type.


lizmat:
This patch appears to break t/spec/S06-currying/positional.t , test #36 .  Could you find out why?

zostay:
I looked at it and threw my arms up. I have no idea why using --> in the proto would strip is raw from the multi, at least that's what I think happened. I switched to using returns instead, but just on the multi and that seems to be better.

zoffixznet:
Thank for the work, but based on comments in #857, seems it superseeded this PR and was merged before this one. Sorry.

PR #800, opened 2016-06-24
Change required version to 5.014. Since Perl::OSType was added in 5.013

andreoss:


zoffixznet:
:-1: We don't require 5.14 perl. We only require a single module. First, the user can download install that module with cpan. And an even better solution is to include that module.

One of the hosts I manage still has 5.10. Not everyone is running latest and greatest and bumping wrong prereqs as an easy way out just adds confusion over the requirements.

lizmat:
so why do we need Perl::OSType ?

zoffixznet:
> so why do we need Perl::OSType ? 

The module maps what Perl 5's `$^O` (operating system name) into generic types (unix, Windows, etc).

I'm not familiar with Rakudo codebase to answer why we need it, but it can really be replicated with a single hash, without bumping perl prereq version: https://metacpan.org/source/DAGOLDEN/Perl-OSType-1.010/lib/Perl/OSType.pm#L17

AlexDaniel:
According to the first page we *must* support perl v5.6.

+1 for bumping perl version. It's time to move on.

zoffixznet:
> According to the first page we must support perl v5.6.

No, we mustn't. The page shows Perl &lt;5.14 is still in use, nothing more. We can either disenfranchise them with fake prereqs on flimsy pretext that it's "time to move on," with full acknowledgement that the reason they still use ancient perl is beyond their control. Or, we can copy-paste a dozen lines of code, which I'll do in about an hour when we get home.

I don't think practicing planned obsolecense in this case has any benefits to the Rakudo project.

PR #815, opened 2016-07-08
min/max on Failure must fail, not return ±Inf

zoffixznet:
Fixes [RT#128573](https://rt.perl.org/Ticket/Display.html?id=128573), namely:

```
<Zoffix> m: say Failure.new.min: &infix:<cmp>
<camelia> rakudo-moar 29a110: OUTPUT«Inf␤»
<Zoffix> m: say Failure.new.max: &infix:<cmp>
<camelia> rakudo-moar 29a110: OUTPUT«-Inf␤»
<Zoffix> m: say Failure.new.max
<camelia> rakudo-moar 29a110: OUTPUT«-Inf␤»
<Zoffix> m: say Failure.new.min
<camelia> rakudo-moar 29a110: OUTPUT«Inf␤»
<Zoffix> m: say min +'a'
<camelia> rakudo-moar 29a110: OUTPUT«Inf␤»
<Zoffix> m: say max +'a'
<camelia> rakudo-moar 29a110: OUTPUT«-Inf␤»
```

### DIFF
--- a/src/core/Any-iterable-methods.pm
+++ b/src/core/Any-iterable-methods.pm
@@ -902,7 +902,7 @@ Did you mean to add a stub (\{...\}) or did you mean to .classify?"
         todo = $elems;
         my $value;
 
-        return $value
+        return nqp::istype($value, Failure) ?? $value.exception.throw !! $value
           if nqp::isconcrete($value := self.AT-POS(i))
             while nqp::islt_i(++i,todo);
 


zoffixznet:
To consider before merging. The added behaviour contains inconsistency, as the current Rakudo actually throws for when there are more than one argument:

```
zoffix@VirtualBox:~/CPANPRC/rakudo$ ./perl6-m -e 'my $x = min +"a"; say 42'
42
zoffix@VirtualBox:~/CPANPRC/rakudo$ ./perl6-m -e 'my $x = min +"a", +"a"; say 42'
Cannot convert string to number: base-10 number must begin with valid digits or '.' in '⏏a' (indicated by ⏏)
  in block <unit> at -e line 1

Actually thrown at:
  in block <unit> at -e line 1

zoffix@VirtualBox:~/CPANPRC/rakudo$ 
```

zoffixznet:
Updated to throw as per chat, but I now see reference to `in any  at gen/moar/m-Metamodel.nqp line 3055` for the case this PR address, which isn't present in the currently existing exception thrown when multiple args are given:

```
cpan@perlbuild2:~/CPANPRC/rakudo$ ./perl6-m -e 'say min +"a"'                                                                                                                                                                                                  
Cannot convert string to number: base-10 number must begin with valid digits or '.' in '⏏a' (indicated by ⏏)
  in block <unit> at -e line 1

Actually thrown at:
  in any  at gen/moar/m-Metamodel.nqp line 3055
  in block <unit> at -e line 1

cpan@perlbuild2:~/CPANPRC/rakudo$ ./perl6-m -e 'say min +"a", +"a"'
Cannot convert string to number: base-10 number must begin with valid digits or '.' in '⏏a' (indicated by ⏏)
  in block <unit> at -e line 1

Actually thrown at:
  in block <unit> at -e line 1

cpan@perlbuild2:~/CPANPRC/rakudo$
```

zoffixznet:
PR Updated with sortiz++'s version that now throws a consistently-looking version of the exception:

```
zoffix@VirtualBox:~/CPANPRC/rakudo$ ./perl6 -e 'say max +"a", +"a"'
Cannot convert string to number: base-10 number must begin with valid digits or '.' in '⏏a' (indicated by ⏏)
  in block <unit> at -e line 1

Actually thrown at:
  in any  at gen/moar/m-Metamodel.nqp line 3055
  in block <unit> at -e line 1

zoffix@VirtualBox:~/CPANPRC/rakudo$ ./perl6 -e 'say max +"a"'                                                                                                                                                                                                  
Cannot convert string to number: base-10 number must begin with valid digits or '.' in '⏏a' (indicated by ⏏)
  in block <unit> at -e line 1

Actually thrown at:
  in any  at gen/moar/m-Metamodel.nqp line 3055
  in block <unit> at -e line 1

zoffix@VirtualBox:~/CPANPRC/rakudo$ 
```

salortiz:
@zoffixznet On second thought, after reading [RT#125334](https://rt.perl.org/rt3//Public/Bug/Display.html?id=125334) and the last two paragraphs of [S03#Comparison_semantics](http://design.perl6.org/S03.html#Comparison_semantics) I'm not so convinced that we took the correct approach.

So I now have two questions that should be answered at the language design level before this can be merged:
- `Any min Str` *really* should returns `Inf`?
- `Failure.new cmp Any` should throw or returns the Failure?

Worries me cementing on roast an improper semantic.

zoffixznet:
As per [lizmat++'s suggestion](http://irclog.perlgeek.de/perl6-dev/2016-07-08#i_12806261) I've modified the PR once again to the version where single-arg min/max with a Failure returns a Failure.

salortiz:
As this pretends to close [PR#125334](https://rt.perl.org/rt3//Public/Bug/Display.html?id=125334) the case `foo min bar` is relevant because it is the same that `min(foo,bar)`.

OTOH, the current approach propagates (without throwing) a Failure only if it is the only concrete argument, so not much consistent.

IMO, the goal for close both tickets should be a consistent  "throw vs propagate" behavior.

zoffixznet:
> As this pretends to close PR#125334 the case foo min bar is relevant because it is the same that min(foo,bar).

No, this PR closes only [RT#128573](https://rt.perl.org/Ticket/Display.html?id=128573) and doesn't affect the more-than-one-arg behaviour.

The current behaviour for 2+ args—described in [RT#125334](https://rt.perl.org/Ticket/Display.html?id=125334) as "fixed"—was done before 2015.10 release. I merely added tests for it last night.

The inconsistency makes sense to me, since a Failure is not compared with anything in 1-arg case, so propagating it makes sense. In the 2+ args version, we use `cmp` for comparison. And it's `cmp` that throws:

```
$ perl6 -e 'my $v = +"a"; say 42'
42
$ perl6 -e 'my $v = +"a"; $v cmp 1; say 42'
WARNINGS for -e:
Useless use of "cmp" in expression "$v cmp 1" in sink context (line 1)
Cannot convert string to number: base-10 number must begin with valid digits or '.' in '⏏a' (indicated by ⏏)
  in block <unit> at -e line 1

Actually thrown at:
  in block <unit> at -e line 1
```
The propagation of Failures for 2+ args suggests we can find min/max values without using any actual values and implies that `cmp` needs to propagate Failures too.

salortiz:
Ok, so this holds `min(+'a') ~~ Failure` is `True`

What we expect with `min(Any, +'a')`, that should throw or propagate?

zoffixznet:
> What we expect with min(Any, +'a'), that should throw or propagate?

Personally, I'd expect that to throw, as I'm giving it two things and I expect it to use their values to compare the two and give the minimum and trying to see Failure's value explodes it. The current answer `Inf` hides the error in my code. The PR version gives a `Failure` in that case. So yeah, the PR does affect 2+ arg versions, contrary to what I stated earlier.

With this it does seem like throwing for some cases and Failuring for others is inconsistent, especially if you consider things like `min @any-number-of-things-inside`.

RAKUDO:
```
$ perl6 -e 'say min(Any, +"a");'
Inf
```

PR:
```
cpan@perlbuild2:~/CPANPRC/rakudo$ ./perl6 -e 'say min(Any, +"a");'
Cannot convert string to number: base-10 number must begin with valid digits or '.' in '⏏a' (indicated by ⏏)
  in block <unit> at -e line 1

Actually thrown at:
  in block <unit> at -e line 1

cpan@perlbuild2:~/CPANPRC/rakudo$ ./perl6 -e 'my $x = min(Any, +"a"); say 42'
42
cpan@perlbuild2:~/CPANPRC/rakudo$ 
```

salortiz:
Indeed, that's why I'm searching a more consistent/general approach. We should not forget that the one-arg case is just a degenerated one.

Moreover, I suspect AlexDaniel assumed [RT#125334](https://rt.perl.org/Ticket/Display.html?id=125334) "fixed" because he tested with `.say`, that masks the real behavior, and pmichaud and Coke seems to agree on Failure propagation. (when I try to reopen it, I can't)

AlexDaniel:
@salortiz I assumed that it was fixed because the original one-liner was ``min(+'a', +'a').say``, and I am almost 100% sure that it used to print Inf. I didn't think about it thoroughly. I reopened the ticket for you.

salortiz:
My recent thoughts.

- Per S03  `… and the general policy on unthrown exceptions is that the exception is thrown as soon as you try to use the exception as a real value.`
- Any use of `min`/`max` with more that one argument actually throws at `cmp` time when a Failure is involved.
- `min`/`max` with only one argument is a degenerate case that should be consistent.

So, the best approach IMO is, indeed, throw in !first-concrete if a Failure is detected.

zoffixznet:
Updated the PR per latest comments. We now throw failures, even for single-arg calls.

PR #829, opened 2016-07-22
Fix RT#127958 on moarvm... :D smiley containers assigned from Nil

skids:
There is no sensible default default for :D types, blame Kurt Gödel.
  This breaks the assumption that a type's default is a legal value.
  So, check the assigned type even when assigning from the_default.

### DIFF
--- a/src/core/Exception.pm
+++ b/src/core/Exception.pm
@@ -1937,30 +1937,36 @@ my class X::TypeCheck::Binding is X::TypeCheck {
     has $.symbol;
     method operation { 'binding' }
     method message() {
-        if $.symbol {
-            self.priors() ~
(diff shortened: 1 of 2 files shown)


skids:
Passed roast master and 6.c-errata.
JVM/JS fixes may also be needed.

gfldex:
judgment was spoken see: https://rt.perl.org/Public/Bug/Display.html?id=127958

lizmat:
No todoed tests passed  :-(

LLFourn:
Tests: https://github.com/perl6/roast/commit/fa9a6b05f665bef3b5b45cb35d4018b2a8368c31

PR #844, opened 2016-08-22
Reject signatures with both slurpy parameters and default values.

LemonBoy:
Closes MoarVM's [145 ticket](https://github.com/MoarVM/MoarVM/issues/145).
Comes with [roasted tests](https://github.com/perl6/roast/pull/144).


jnthn:
Looks good. You'd never be able to use the default value here anyway (since an empty capture is still a valid capture, and so it would be happily bound). Better to catch it here blowing up at code-gen time.

PR #845, opened 2016-08-24
Print the right type name when the type constraint check fails.

LemonBoy:
[RT#126124](https://rt.perl.org/Public/Bug/Display.html?id=126124).


lizmat:
I guess we still need tests for this  :-)

zoffixznet:
Tests added in https://github.com/perl6/roast/commit/c0feebdbb1

PR #849, opened 2016-08-26
fix p5ism: $*DISTRO names are lowercase

cygx:
Unbreaks the windows build.

Note that we could in principle use `$*DISTRO.is-win` instead. However, that would trigger for Cygwin as well where we should be creating shell scripts instead of batch files.


lizmat:
Sigh.   Basically did this in ebf34aa946edacdbf , then changed it in 0f1d0394bdc4401e , then changed the test back in 6a8278cd5d92e87c7c .  I really should first check for PR's before checking backlog on #perl6-dev

PR #851, opened 2016-08-26
Avoid leaking file handles during precompilation

cygx:
Sadly, this does not fix an issue I'm having on win32 where `nqp::rename` fails during precompilation :(


lizmat:
Thank you!  unfortunately I noticed this PR too late  :-(

Implemented by ede7c60b214b5d1 and 070497028938981c5

PR #853, opened 2016-08-26
REPL Cleanup

awwaiid:
* Remove now unused fallback-repl methods
* Extract control command check into method


zoffixznet:
Thanks a lot!

PR #854, opened 2016-08-30
Fix RT #118607

MasterDuke17:
Give the same suggestion for a sub redeclaration as for a method.

### DIFF
--- a/src/core/Exception.pm
+++ b/src/core/Exception.pm
@@ -947,7 +947,8 @@ my class X::Redeclaration does X::Comp {
     has $.postfix = '';
     has $.what    = 'symbol';
     method message() {
-        "Redeclaration of $.what $.symbol$.postfix";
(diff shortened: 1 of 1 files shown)


zoffixznet:
Are you sure this doesn't break the cases when the $.what is **not** a routine? This small tests shows the $m is not returned on false cases:

```
$ perl6 -e '{ my $m = "foo"; $m ~= "bar" if 42 }().say'
foobar

$ perl6 -e '{ my $m = "foo"; $m ~= "bar" if 0 }().say'
()
```

MasterDuke17:
Good catch. It passed the spectest, but I guess that just means there should be a test added. I have a test for the 'routine' case I'll add when this is merged, I'll also try to create a test for the non-'routine' case.

zoffixznet:
Thanks a lot!

PR #856, opened 2016-09-01
Print a newline after the message when using 'warn'

LemonBoy:
Streamline the behaviour so now it's consistent across die, note and
warn.


zoffixznet:
Thanks a lot!

PR #862, opened 2016-09-03
Catch another class of compile-time errors

awwaiid:
Resolves https://rt.perl.org/Public/Bug/Display.html?id=129179


zoffixznet:
Thanks a lot!

PR #863, opened 2016-09-04
Fix RT #129100

titsuki:
SEE ALSO: https://rt.perl.org/Public/Bug/Display.html?id=129100

### DIFF
--- a/src/core/Range.pm
+++ b/src/core/Range.pm
@@ -612,7 +612,7 @@ my class Range is Cool does Iterable does Positional {
 
     method rand() {
         fail "Can only get a random value on Real values, did you mean .pick?"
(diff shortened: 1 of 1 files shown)


zoffixznet:
As [commented on IRC](http://irclog.perlgeek.de/perl6/2016-09-04#i_13145737), this PR looks as a no-op to me. Are you sure it actually fixes the problem?

zoffixznet:
The issue stems from a coercion in [one of the `.new` multies](https://github.com/rakudo/rakudo/blob/aceb4af2/src/core/Range.pm#L51) that appears to have a sideeffect. This PR does not address that, so I'm going to close it. I'll attempt to debug this a bit further and will comment on the ticket.

PR #864, opened 2016-09-04
Warn the user if the desired coercion cannot be performed.

LemonBoy:
Emit a 'can' before invoking the coercion method.


niner:
What are the performance implications of this change?

PR #868, opened 2016-09-05
Repl compiletime error continuation

awwaiid:
Need to reset the input upon CATCH so further things can be executed.


awwaiid:
Bah. @zoffixznet beat me to it :)

PR #871, opened 2016-09-07
Disable grammar errors when running under -e

MasterDuke17:
This disables the 'expecting any of' error message when running under
-e. E.g., running 'p6 -e "/sadf\/"' usually produces this error:
===SORRY!===
Regex not terminated.
at -e:1
------> /sadf\/⏏<EOL>
Unable to parse regex; couldn't find final '/'
at -e:1
------> /sadf\/⏏<EOL>
    expecting any of:
        infix stopper
    term

With this patch the error is instead just:
===SORRY!===
Regex not terminated.
at -e:1
------> /sadf\/⏏<EOL>
Unable to parse regex; couldn't find final '/'
at -e:1
------> /sadf\/⏏<EOL>

However, it also adds a command line switch '--grammar-errors' which
can be used to turn them back on. Behavior when running a file, e.g.,
'perl6 script.p6' is unchanged.

Implements RT #128969

### DIFF
--- a/src/Perl6/Compiler.nqp
+++ b/src/Perl6/Compiler.nqp
@@ -84,6 +84,7 @@ class Perl6::Compiler is HLL::Compiler {
       -v, --version        display version information
       --stagestats         display time spent in the compilation stages
       --ll-exception       display a low level backtrace on errors
+      --grammar-errors     display grammar errors when running under -e
       --profile[=kind]     write profile information to an HTML file (MoarVM)
                              instrumented - performance measurements (default)
                              heap - record heap snapshots after every
--- a/src/core/Exception.pm
+++ b/src/core/Exception.pm
@@ -570,7 +570,7 @@ my role X::Comp is Exception {
                 !! "$.message\nat line $.line"
               !! "$.message\nat $.filename():$.line";
             $r ~= "\n------> $green$.pre$yellow$eject$red$.post$clear" if defined $.pre;
-            if $expect && @.highexpect {
+            if $expect && @.highexpect && !($*PROGRAM-NAME eq '-e' and !Rakudo::Internals.GRAMMAR-ERRORS) {
                 $r ~= "\n    expecting any of:";
                 for @.highexpect {
                     $r ~= "\n        $_";
--- a/src/core/Rakudo/Internals.pm
+++ b/src/core/Rakudo/Internals.pm
@@ -811,6 +811,11 @@ my class Rakudo::Internals {
              }
           !! nqp::list()
     }
+    method GRAMMAR-ERRORS() {
+        nqp::existskey($compiling-options, 'grammar-errors')
+          ?? '--grammar-errors'
+          !! Empty
+    }
 
 #?if moar
     method PRECOMP-EXT()    { "moarvm" }
--- a/src/main.nqp
+++ b/src/main.nqp
@@ -26,6 +26,7 @@ my @clo := $comp.commandline_options();
 @clo.push('c');
 @clo.push('I=s');
 @clo.push('M=s');
+@clo.push('grammar-errors');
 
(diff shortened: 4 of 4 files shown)


MasterDuke17:
This passes make spectest.

niner:
What's the rationale for this change? The commit message mentions only what is changed but not why we want to withhold useful information in that case?

AlexDaniel:
Keep in mind there's also [RT #128969](https://rt.perl.org/Ticket/Display.html?id=128969) where we can discuss it.

zoffixznet:
:-1: from me.

I don't see why `-e` code should produce different errors than code from a file and bloating the executable with more command line options just to placate people who want the original errors is not a solution.

As for the ticket:

> but the "expecting any of" I ignore. I understand that people doing fancy grammar related things probably do find them very useful.

I saw [a conversation about 'infix stoppers'](http://irclog.perlgeek.de/perl6-dev/2016-08-16#i_13030963) that talked about possibly filtering out such opaque things. So I think rather than adding command line switches and logic to hide information generated by `-e` code, we should just improve the usefulness of the `expected any` message.

vendethiel:
+1 to @zoffixnet's message.

MasterDuke17:
nine: Your comment here (
http://irclog.perlgeek.de/perl6-dev/2016-08-16#i_13030929) is part of the
rational. The RT has some more explanation. Tldr; a couple people have said
they don't find that part of the error message useful, so I removed them in
the circumstance where I felt they were least useful.

Zoffix: OK, we could disable it for -e *and* for a file, that way the error
messages are the same. This may just be a matter of opinion, but I don't
feel that adding just one command line option is bloating. There already is
some precedent, the added output you get from --ll-exception was put behind
a command line option.

All told, I'm not married to this. If we can improve the existing message
that would be great. There hadn't been any comments on the ticket, and I
wasn't sure how complicated it would be to implement. I looked at the code,
found out it was pretty simple, so went ahead and did it, figuring a PR
would spark some discussion.

(As somewhat of an aside, and I don't have access to a way to test this
right now, but I believe I noticed that the error message from running the
example I gave with -e was not exactly the same as when I echoed it to a
file and ran it that way

MasterDuke17:
Right, so `p6 -e '/sadf\/'` outputs:
===SORRY!===
Regex not terminated.
at -e:1
------> /sadf\/⏏\<EOL>
Unable to parse regex; couldn't find final '/'
at -e:1
------> /sadf\/⏏\<EOL>
    expecting any of:
        infix stopper
        term

and `p6 a`, where `cat a` = `/sadf\/` outputs:
===SORRY!===
Regex not terminated.
at /home/dan/Source/perl6/ForMasterDuke/a:2
------> \<BOL>⏏\<EOL>
Unable to parse regex; couldn't find final '/'
at /home/dan/Source/perl6/ForMasterDuke/a:2
------> \<BOL>⏏\<EOL>
    expecting any of:
        infix stopper

I actually notice three differences/problems. The first is that it says the error is at line 2, but there's only one line in the file (and if I add a blank line either before or after it says the error is on line 3). The second is that it doesn't show the bad regex, just '\<BOL>'. The third is that it doesn't say it's missing `term`.

zoffixznet:
Thank you for the work, but I'm rejecting this PR and the associated ticket.

I don't think the added maintenance burden and extra option the users have to read through justify the presence of this very minor feature that might not be desired by all.

We have since implemented [`RAKUDO_EXCEPTIONS_HANDLER` env var](https://docs.perl6.org/programs/00-running#Error_Message_Verbosity_and_Strictness) that lets users use custom classes for error reporting (that allow filtering for custom handling for just specific exceptions; letting default handler handle the rest). So there's a possibility of implementing this feature entirely in a module; or expanding the idea further: `Exceptions::Simple` for example that give just the basics.

PR #872, opened 2016-09-07
Name the <sym> token when parsing a ternary operator.

LemonBoy:
Some functions such as 'can_meta' might need this information to print
more informative error messages.

RT#129080

Executing
```perl6
1 R?? 2 !! 3
```
now gives
```Cannot reverse the args of ?? because conditional operators are too fiddly```.

Not spectest'd yet.


zoffixznet:
Thanks a lot!

PR #875, opened 2016-09-11
Use '+=' and '-=' instead of postfix ops

usev6:
This is a workaround for rakudo-j, because postfix:<++>
and postfix:<--> are dying here:
https://rt.perl.org/Ticket/Display.html?id=128123#txn-1422445


zoffixznet:
Thanks.

PR #876, opened 2016-09-12
Add Nil return constraints where possible

MasterDuke17:
This was a quick first pass through the code, I tried to only make a change where it was obviously ok. Passes `make spectest`. Each commit is an individual file. Inspired by timotimo's comment here http://irclog.perlgeek.de/perl6-dev/2016-09-10#i_13184818

### DIFF
--- a/src/core/Any-iterable-methods.pm
+++ b/src/core/Any-iterable-methods.pm
@@ -957,7 +957,7 @@ Did you mean to add a stub (\{...\}) or did you mean to .classify?"
                 has $!arity;
                 has $!count;
 
-                method set-cheat($new-arity, $new-count) {
+                method set-cheat($new-arity, $new-count --> Nil) {
                     $!arity = $new-arity;
                     $!count = $new-count;
                 }
--- a/src/core/Any.pm
+++ b/src/core/Any.pm
@@ -28,13 +28,13 @@ my class Any { # declared in BOOTSTRAP
     multi method EXISTS-KEY(Any:D: $) { False }
 
     proto method DELETE-KEY(|) is nodal { * }
-    multi method DELETE-KEY(Any:U: $) { Nil }
+    multi method DELETE-KEY(Any:U: $ --> Nil) { }
     multi method DELETE-KEY(Any:D: $) {
         Failure.new("Can not remove values from a {self.^name}")
     }
 
     proto method DELETE-POS(|) is nodal { * }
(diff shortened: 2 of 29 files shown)


zoffixznet:
I reviewed all of the changes. Below are the things I pick for a second pair of eyes to review, as I don't know what those guts are doing. All the rest of the changes are OK.

---

All the changes here used to return the value of .add_phaser() that is [a complex condition](https://github.com/rakudo/rakudo/blob/15532dbc19b2328fa8916599a1a1d71f6fded641/src/Perl6/World.nqp#L3186). Are those not needed in here? https://github.com/rakudo/rakudo/pull/876/commits/d5ed2f4b30449618b3db3475d882b38a254a0d98

Do any conditions in the set_value's if() chain return anything we want? https://github.com/rakudo/rakudo/pull/876/commits/8f69d017ed15c73c5aa46712a5878778f77e1758

This currently returns True if the cancelation was done and falsy if not. Do we want to keep that? https://github.com/rakudo/rakudo/pull/876/commits/781a19d5f4a202b16d2519a35e5fbeaf5b2ad044

This returns the value of nqp::ctxcaller() when $!module_ctx is falsy; is that needed?
https://github.com/rakudo/rakudo/pull/876/commits/3e42a6c2f45cdf7f72c1c6c471ba3f6fc9f28de2

This returns True if inilization occurs and False if it already is initilized; do we need that?
https://github.com/rakudo/rakudo/pull/876/commits/692c7c1b020fec7835fdc4101fd3b8c257e9ca81

This always returned True, not will always return False; is that OK?
https://github.com/rakudo/rakudo/pull/876/commits/dd05735c7e205edef12fffca8631813a34ec80e8

The bottom two here used to return the resultant arrays, now return Nil; is that OK?
https://github.com/rakudo/rakudo/pull/876/commits/35a77321641fb9d39eb6b45725c94e64cab8ca80

The first two changes used to return the result of conditional; now return Nil. is that OK?
https://github.com/rakudo/rakudo/pull/876/commits/5d4621f2fc256e153f541a7c62a5e50bd0d8db68

ugexe:
It may be worth noting that

    src/core/CompUnit/PrecompilationRepository.pm
    src/core/CompUnit/RepositoryRegistry.pm

define interfaces, and that forcing a return type on them limits future changes to the CUR API

MasterDuke17:
Redone in #1026

PR #878, opened 2016-09-16
Allow Unicode digits in radix bases, match variables, and type constraints

MasterDuke17:
Fixes RT #129279


MasterDuke17:
Passes `make spectest`, I'll add tests to roast when this is merged.

PR #884, opened 2016-09-26
Revert "Error if No/Nl chars are used in colon pairs/radix"

MasterDuke17:
Reverts rakudo/rakudo#883

See the conversation with @TimToady starting [here](http://irclog.perlgeek.de/perl6-dev/2016-09-26#i_13280449)

tl;dr We don't need to teach people how Unicode works, but if a lot more similar tickets to [RT #129319](https://rt.perl.org/Ticket/Display.html?id=129319) show up we can revisit


zoffixznet:
Thanks.

PR #892, opened 2016-10-01
Mention typename in X::Assignment::RO on JVM

usev6:
Makes two tests in S32-basics/xxKEY.t pass (fudged 'todo' now).


zoffixznet:
Thanks!

PR #894, opened 2016-10-04
Fail with typed errors if nqp::open dies in IO::Handle.open.

cygx:
Also removes an expensive check necessary to avoid opening directories:
This is now handled at the VM level, which still needs to happen on the JVM (cf https://github.com/perl6/nqp/issues/310).

It could be made even nicer if there was a way to communicate error types to the HLL, eg via error codes provided in addition to the error message. Currently, we're forced to (non-atomically) repeat checks that already have been performed at the VM level (or parse error messages).

### DIFF
--- a/src/core/Exception.pm
+++ b/src/core/Exception.pm
@@ -467,6 +467,14 @@ my class X::IO::DoesNotExist does X::IO {
     }
 }
 
+my class X::IO::DoesExist does X::IO {
+    has $.path;
+    has $.trying;
+    method message() {
+        "Found preexisting file '$.path' while trying to do '.$.trying'"
+    }
+}
+
 my class X::IO::NotAFile does X::IO {
     has $.path;
     has $.trying;
--- a/src/core/IO/Handle.pm
+++ b/src/core/IO/Handle.pm
@@ -94,7 +94,9 @@ my class IO::Handle does IO {
         {
             CATCH {
                 fail X::IO::DoesNotExist.new(:$!path, :trying<open>)
-                    unless $!path.e;
+                    unless $create || $!path.e;
+                fail X::IO::DoesExist.new(:$!path, :trying<open>)
+                    if $exclusive && $!path.e;
                 fail X::IO::Directory.new(:$!path, :trying<open>)
                     if $!path.d;
                 fail X::IO.new(:os-error(.message));
--- a/src/core/IO/Handle.pm
+++ b/src/core/IO/Handle.pm
@@ -73,8 +73,10 @@ my class IO::Handle does IO {
             return self;
         }
 
+#?if jvm
         fail (X::IO::Directory.new(:$!path, :trying<open>))
           if $!path.e && $!path.d;
+#?endif
 
         my $llmode = do given $mode {
             when 'ro' { 'r' }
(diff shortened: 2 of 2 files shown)


niner:
It's quite possible to throw a typed exception from the VM level as I've recently learned. It's just a little more involved. Look for P6EX in the source.

OTOH the additional checks are done only in an error case anyway. And how much performance are you gonna need there?

cygx:
> It's quite possible to throw a typed exception from the VM level as I've recently learned.

Wouldn't that introduce a dependency on the specific HLL?

> And how much performance are you gonna need there?

It's not about performance, but having to duplicate the checks independently on the p6-side. Such a repetition just introduces another chance to get it wrong, which in fact I already managed to do (cf https://github.com/rakudo/rakudo/pull/894/commits/191f68e19844da23c60827877371a6c9f5af5b5c).

It's also a bit iffy in so far as in principle, the file in question could have been modified between the failure of the call to `MVM_file_open_fh()` and the checks in the `CATCH` block, so incase of concurrent modification, it's theoretically possible to throw thw wrong exception...

cygx:
Conflicts aside, it should still be good(-ish). The `#?if jvm` can likely go due to https://github.com/perl6/nqp/issues/310 getting resolved. The new exception type could use a better name (`DoesAlreadyExist`? something completely different?). It would still be preferable if the reason for the failure could be communicated upwards from the VM to avoid race conditions (eg `nqp::open` fails due to missing file, file gets created between the attempt to open and the checking of the error conditions)...

zoffixznet:
> It would still be preferable if the reason for the failure could be communicated upwards from the VM to avoid race conditions 

Oh, whoops. Yeah, I've just actually looked at the code and not too happy with such decisions what what exceptions to throw. There's definitely a way to throw typed VM exceptions... I'll look into that.

lizmat:
I am unsure as to the status of this PR.  Can someone enlighten?

niner:
The #?if jvm can go. It would indeed be better if the VM threw the typed exception. The mechanism for this does not create a dependency on the HLL (it's already in use in several places after all without trouble). Doing those checks twice does indeed create a race condition where the cause of the error may change between the checks. Nevertheless, if no one volunteers to implement throwing typed exceptions at the VM level, then we'd still be better off with this merged. Race condition and all.

vrurg:
> The #?if jvm can go. It would indeed be better if the VM threw the typed exception. The mechanism for this does not create a dependency on the HLL (it's already in use in several places after all without trouble).

What kind of mechanism do you mean here?

niner:
There are several fields for handlers that generate typed exceptions in the 
HLL config:

https://github.com/rakudo/rakudo/blob/master/src/Perl6/bootstrap.c/
BOOTSTRAP.nqp#L3902

Though I'd like to have an even easier way of throwing typed exceptions from 
the VM level. If all it'd take is an entry here and an entry there, we 
wouldn't use MVM_exception_throw_adhoc all the time and never have to parse 
exception messages in Rakudo...

vrurg:
@niner I have implemented a replacement for `P6EX`: https://github.com/rakudo/rakudo/blob/12b49b8343e86633b0f94d826cbac3469a03c2e2/src/Perl6/Metamodel/Configuration.nqp#L73

When used it looks much tidier and readable than `P6EX`: https://github.com/rakudo/rakudo/blob/12b49b8343e86633b0f94d826cbac3469a03c2e2/src/Perl6/Metamodel/CoercionHOW.nqp#L205. And it doesn't require extra boilerplate to register an exception which is looked up by its name by traversing the stash of package `X`.

But use it from inside a VM would still be complicated by requiring named parameters for exception constructor.

lizmat:
Is this PR still salvageable ?

PR #898, opened 2016-10-07
Remove bandaid for JVM

usev6:
Looks like back in 2015 those unbound closures caused NPE.
This seems to be fixed now.


zoffixznet:
Thanks!

PR #900, opened 2016-10-10
Simplify rad_number token a bit

MasterDuke17:
Pull out some copied code into their own tokens.

Passes `make spectest`


lizmat:
Are there any performance issues involved with this?

MasterDuke17:
Spectest took just about the same time, some tests of doing radix conversion in a loop took the same time before and after.

PR #903, opened 2016-10-12
handle uneven rows lengths in Pod::To::Text::table2text()

dwarring:
Fixes RT129862, e.g.:

=begin table
a | b | c
l | m | n
x | y
=end table

### DIFF
--- a/lib/Pod/To/Text.pm6
+++ b/lib/Pod/To/Text.pm6
@@ -67,8 +67,9 @@ sub table2text($pod) {
     my @rows = $pod.contents;
     @rows.unshift($pod.headers.item) if $pod.headers;
     my @maxes;
-    for 0..(@rows[1].elems - 1) -> $i {
-        @maxes.push([max] @rows.map({ $_[$i].chars }));
+    my $cols = [max] @rows.map({ .elems });
+    for 0..^$cols -> $i {
+        @maxes.push([max] @rows.map({ $i < $_ ?? $_[$i].chars !! 0 }));
     }
     my $ret;
     if $pod.config<caption> {


tbrowder:
Lizmat, I don't think that fix should be permanent one (even though the test I added does work).  According to our agreed-upon changes in <https://github.com/perl6/doc/blob/master/doc/Language/tables.pod6>, Restriction 5, the "bad" table is clearly malformed (uneven rows), and any change should fail with a message saying so and point to the offending table.  My (much delayed work on the tables) should fix that.  Until  that time, I recommend the issue stay open.

lizmat:
Hmmm... doesn't seem I have the rights to open this issue again.  Or is it open now?

zoffixznet:
It's a merged PR, not an Issue, so I don't think one can reopen it.

jonathanstowe:
There is however a 'revert' button by the commit merged message above, if that is appropriate ;-)

zoffixznet:
I'm a bit confused about what we're doing here:

1. This PR fixed [RT#129862](https://rt.perl.org/Ticket/Display.html?id=129862) allowing for uneven rows
2. https://github.com/perl6/roast/pull/168 added tests for it
3. However, @tbrowder's comments above indicate uneven rows should be forbidden?

If they're forbidden, then this PR should be reverted and the test changed to a TODO, expecting an exception. If they're permitted, then which issue needs to remain open?

dwarring:
Another way of looking at this PR is that it allows trailing empty cells to be omitted from a row.

tbrowder:
On Fri, Oct 14, 2016 at 12:37 PM, Zoffix Znet <notifications@github.com> wrote:
> I'm a bit confused about what we're doing here:
>
> This PR fixed RT#129862 allowing for uneven rows
> perl6/roast#168 added tests for it
> However, @tbrowder's comments above indicate uneven rows should be
> forbidden?

Yes.  See <https://docs.perl6.org/language/tables#Restrictions>, number 5:

<quote>
5. Having unequal numbers of columns in table rows is illegal.
</quote>

> If they're forbidden, then this PR should be reverted and the test changed
> to a TODO, expecting an exception.

That's the proper step IMHO.

zoffixznet:
And what is the reason for such a restriction? I can see it being problematic when whitespace is used for column separators and right-most cell of a particular row is meant to be empty.

tbrowder:
> And what is the reason for such a restriction? I can see it being
> problematic when whitespace is used for column separators and
> right-most cell of a particular row is meant to be empty.

That's the primary reason.  When we were discussing the limitations it
seemed reasonable to make that a requirement for all tables.

Whether the table is manually or machine generated, there should be no
missing columns.

The problem that started the issue is no clue about where the error
occurred--my solution will cure that hopefully.

zoffixznet:
> That's the primary reason. When we were discussing the limitations it seemed reasonable to make that a requirement for all tables.

What do you mean? What I stated is the reason *against* such a restriction. Basically, you're saying that anyone who uses whitespace for column separators should get ready to re-draw their entire table the moment they have an empty cell.

> Whether the table is manually or machine generated, there should be no missing columns.

In my experience, it's pretty common to have empty cells in tables. I don't even have to go far to get an example. Our own compiler features table has tons of empty cells (right-most column): http://perl6.org/compilers/features

> The problem that started the issue is no clue about where the error occurred--my solution will cure that hopefully.

And why not use max-columns, just as this PR already did. It'd be pretty obvious to the person creating the table when they mess up, since the first time that table is rendered there'd be an extraneous column—this worked well for HTML for decades.

Based on information you provided for the reasoning, it's :-1: from me: you're effectively making white-space column separators a discouraged feature because you want the parser to complain about defects that are easily detectable without it anyway.

dwarring:
Imho additional checking, such as this  pod6checker + testing tools, rather
than putting all the onus or Rakudo.

Rakudo should be a bit more lenient render the POD, if at all possible.

Pretty well how it works in Perl 5.

dwarring:
I'm inclinded to update the Perl 6 Most Wanted list. I think we need Perl 6
equivalents to Perl 5's Pod::Checker and Test::Pod

tbrowder:
On Wednesday, October 19, 2016, Zoffix Znet <notifications@github.com>
wrote:

> And why not use max-columns, just as this PR already did. It'd be pretty
> obvious to the person creating the table when they mess up, since the first
> time that table is rendered there'd be an extraneous column—this worked
> well for HTML for decades.
>
> No one objected when I made the restrictions, but I'm happy for you to
remove restriction 5.

>

zoffixznet:
Thanks!

tbrowder:
I am close to a major overhaul PR for pod table handling and will warn about empty trailing cells only if the user requests debugging.  In any event, the table will be handled gracefully, even with no trailing whitespace as is NOT the case on master at the moment.

PR #905, opened 2016-10-16
Fix pod syntax error in running.pod

dod38fr:
Added the missing `=back` found by `podchecker`

All the best


zoffixznet:
Thanks a lot!

PR #906, opened 2016-10-25
Remove BOOTSTRAPATTRs where possible

MasterDuke17:
Passes `make test` on moar and jvm. Passes `make m-spectest`. Has one or two failures in `make j-spectest`, but I get the exact same failures with `make j-spectest` for a rakudo built from nom at the exact commit I branched from.


FROGGS:
Thank you!

PR #910, opened 2016-10-30
Revert "At least temporary fix for RT #127161"

usev6:
This reverts commit d80c728e257dbe17384c48c9a6e516158e6627a4.

This temporary fix is no longer needed, since a workaround
for the underlying problem was added with commit 986891f7ea


zoffixznet:
Thanks.

PR #913, opened 2016-11-07
Suggested shortcut method similar to the Date methods in Datish.pm

Xliff:
If yyyy-dd-mm is a thing, then this at least adds consistency. ;)


zoffixznet:
Thanks a lot.

PR #914, opened 2016-11-07
Add shortcut method dd-mm-yyyy for American date format, similar to t…

Xliff:
…he yyyy-dd-mm method.

For those who really want to make their string dates harder to deal with (and I'm American and biased!)

### DIFF
--- a/src/core/Dateish.pm
+++ b/src/core/Dateish.pm
@@ -107,7 +107,7 @@ my role Dateish {
     }
 
     method yyyy-mm-dd() { sprintf '%04d-%02d-%02d',$!year,$!month,$!day }
-    method mm-dd-yyyy() { sprintf '%02d-%02d-%02d',$!month,$!day,$!year }
+    method mm-dd-yyyy() { sprintf '%02d-%02d-%04d',$!month,$!day,$!year }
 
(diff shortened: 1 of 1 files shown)


src/core/Dateish.pm
```diff
@@ -107,6 +107,7 @@ my role Dateish {
     }
 
     method yyyy-mm-dd() { sprintf '%04d-%02d-%02d',$!year,$!month,$!day }
+    method mm-dd-yyyy() { sprintf '%02d-%02d-%02d',$!month,$!day,$!year }
```
jonathanstowe: should that last one be %04d ?

AlexDaniel:
So what's the reason to add this particular format?

labster:
This seems like extra clutter for the core.  I'm American and I never use this date format, except when filling out weird paper forms by hand.  I don't understand why this wouldn't be in a module, or even an augment block.

zoffixznet:
:-1: From me on this one. It's not a standard way to write a day and as others pointed out, just adds clutter.

Xliff:
Fair enough. Close it.

zoffixznet:
Thanks anyway 😸

PR #915, opened 2016-11-08
RT#129363: duckmap preserve structure types

cpin:
Implements `deepmap` and `duckmap` in terms of slight modification to the existing `deepmap` code.

Corresponding changes to perl6/roast made in that repo (perl6/roast#180)

Comments welcome!

### DIFF
--- a/src/core/Rakudo/Internals.pm
+++ b/src/core/Rakudo/Internals.pm
@@ -2120,6 +2120,93 @@ my class Rakudo::Internals {
           while nqp::islt_i(++$i,$elems);
         $target
     }
+
+    proto method coremap(|) { * }
+
+    multi method coremap(\op, Associative \h, Bool :$deep) {
+        my @keys = h.keys;
+        hash @keys Z self.coremap(op, h{@keys}, :$deep)
+    }
+
+    multi method coremap(\op, \obj, Bool :$deep) {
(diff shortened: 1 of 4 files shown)


lizmat:
is coremap supposed to be a public method?  If not, could you rework it as a private method?

cpin:
No, it need not be public. Will update.

cpin:
I may be being dim here, but I don't seem to be able to see a way (easily) to make `coremap` "hidden" from the rest of the setting whilst maintaining the "bare" use of `deepmap` and `duckmap`? Lizmat made the very helpful suggestion of using a private method - difficulty being a method of _what_? :) I may be misunderstanding (very probable!)

Looking through the rest of the setting, all other "implementation detail" type subs are contained within classes which it makes sense to instantiate (such as `DateTime` for example). Admittedly, I lack the perl6 chops to really appreciate all of my options - feel free to set me straight!

zoffixznet:
> Lizmat made the very helpful suggestion of using a private method - difficulty being a method of what? :) 

I think it would be of Rakudo::Internals: https://github.com/rakudo/rakudo/blob/nom/src/core/Rakudo/Internals.pm

> Looking through the rest of the setting, all other "implementation detail" type subs are contained within classes

I think the plan is to eventually move all of that stuff into Rakudo::Internals

zoffixznet:
Thanks a lot!

PR #917, opened 2016-11-11
Solaris needs to use gnu make instead of /usr/bin/make

vlmarek:
Standard Solaris make is not capable to build Rakudo. Gnu make is known under the name 'gmake'

### DIFF
--- a/Configure.pl
+++ b/Configure.pl
@@ -134,6 +134,13 @@ MAIN: {
     $config{'runner_suffix'} = $win ? '.bat' : '';
 
     my $make = 'make';
+    if ($^O eq 'solaris') {
+        if (not -X '/usr/bin/gmake') {
+            die "gmake is required to compile rakudo. Please install by 'pkg install gnu-make'";
+        }
+        $make = 'gmake';
+    }
+
     if ($win) {
         my $has_nmake = 0 == system('nmake /? >NUL 2>&1');
         my $has_cl    = `cl 2>&1` =~ /Microsoft Corporation/;


FROGGS:
Awesome! vlamerk++. Does nqp and rakudo pass its `make test`?
What architecture does you machine have? Is it sparc, sparc64, i386 or x86_64?

vlmarek:
>Awesome! vlamerk++.

Thanks, that was just a small thing. rakudo has similar problem, but ...

> Does nqp and rakudo pass its `make test`?

No. When I tried first time, I was able to compile rakudo, nqp and create perl6
binary with some manual hacking. I was able to run the "Forrest Fire" from
Roseta Code (minus 'stty size' does not work on Solaris, 'tput lines', 'tput
cols' does). But half of the perl 6 tests core dumped. Then I started to play
with various compilers and settings and I am not even able to compile rakudo
now :-/ It makes studio compiler crash too :) The problem for me is
consistency - I think that 'make clean', distclean, realclean are not reliable
and don't really erase everything. I think that switching compilers (Solaris
now has three of them) made the whole build system sad. I have to fix that
first. 

>What architecture does you machine have? Is it sparc, sparc64, i386 or
>x86_64?

x86_64 at the moment. A bit of context - I work for Oracle and I do fix bugs in
Solaris. At the moment I am playing on Solaris development release (since the
build toolchain is mostly up to date there) but that might be another factor.
Once I get consistent results I want to consult the problems I am seeing. And I
can test on all the architectures once we get going.

Thanks for working on Perl 6 guys!
__ 
	Vlad

FROGGS:
There is no chance that I'd get shell access to solaris machines so that I could make nqp and rakudo work?

jonathanstowe:
I may try my annual "attempt to run solaris x86 on a VM" at some point :)

vlmarek:
>There is no chance that I'd get shell access to solaris machines so that I could make nqp and rakudo work?

It is theoretically possible and we have done that in the past, but it's process heavy, and there are NDAs involved.  Solaris 11.3 is free to download, so the simplest thing might be if I fix the build infrastructure to just work there and you can give it a go yourself.

Other option would be if I prepare VirtualBox image, but I have to consult that internally first. I'm sure there's ton of internal regulations.

Thank you
-- 
  Vlad

FROGGS:
Okay, I'm downloading 11.3 now... Thank you!

vlmarek:
>I may try my annual "attempt to run solaris x86 on a VM" at some point :)

There's no harm doing that :) I should at least file new ticket to share where am I with my fixes now.

__ 
  Vlad

vlmarek:
Ok, I'm attaching the diff I have, hopefully it helps.

__ 
  Vlad

PR #918, opened 2016-11-13
Check for concreteness in p6scalarfromdesc on JVM

usev6:
That method looks whether the decontainerized value of desc is
a type object. It handles the 'null' case as well.

Fixes RT #128341 and newly failing tests in S09-multidim/assign.t.


zoffixznet:
Thanks a lot.

PR #919, opened 2016-11-15
Changes needed to support line directives

MasterDuke17:
An extra blank line after a directive throws off the numbering.

The m-CORE.setting file is no longer seen in a backtrace, so we need
another way of determining if we are in the settings. This isn't a
great solution, but should work for now.

Requires https://github.com/perl6/nqp/pull/319

With the NQP PR, passes `make spectest`

### DIFF
--- a/src/core/Backtrace.pm
+++ b/src/core/Backtrace.pm
@@ -59,7 +59,7 @@ my class Backtrace::Frame {
         nqp::p6bool(nqp::istype($!code,Routine))
     }
     method is-setting(Backtrace::Frame:D:) {
-        $!file.ends-with("CORE.setting")
+        $!file.contains("src/core")
     }
 }
 
--- a/src/core/Backtrace.pm
+++ b/src/core/Backtrace.pm
@@ -59,7 +59,7 @@ my class Backtrace::Frame {
         nqp::p6bool(nqp::istype($!code,Routine))
     }
     method is-setting(Backtrace::Frame:D:) {
-        $!file.contains("src/core")
+        $!file.starts-with("SETTING::")
     }
 }
 
--- a/tools/build/gen-cat.nqp
+++ b/tools/build/gen-cat.nqp
@@ -20,7 +20,7 @@ sub MAIN(*@ARGS) {
     }
     my $stderr := nqp::getstderr();
     for @ARGS -> $file {
-        say("#line 1 $file\n");
+        say("#line 1 $file");
         my $fh := open($file, :r);
         my int $in_cond := 0;
         my int $in_omit := 0;
--- a/tools/build/gen-cat.nqp
+++ b/tools/build/gen-cat.nqp
@@ -20,7 +20,7 @@ sub MAIN(*@ARGS) {
     }
     my $stderr := nqp::getstderr();
     for @ARGS -> $file {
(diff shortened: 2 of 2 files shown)


zoffixznet:
IMO just checking the path for `.contains("src/core")` is too generic. Any project that follows Rakudo's directory structure would fail to report error message backtraces. And any other combinations are also matched: `src/core-utils`, `/home/tombrozresrc/corespondents/some-other/code/perl6.p6`, etc.

Looking at changes in [https://github.com/perl6/nqp/pull/319/files](https://github.com/perl6/nqp/pull/319/files)... can't we just add some sort of special prefix to core files and filter them out that way instead?

MasterDuke17:
FWIW, some discussion with @niner and @lizmat about `.contains("src/core")` [here](https://irclog.perlgeek.de/perl6-dev/2016-11-12#i_13555788).

zoffixznet:
This one?

```
<lizmat> contains('src/core') ?
<lizmat> MasterDuke: I think ^^ will be sufficient for now
<nine> So people should not develop their modules in a src/core directory?
<MasterDuke> nine: FWIW, there is already an error message somewhere if you try to compile a file ending in '.setting'
```

TBH, I'm not a fan of "for now" solutions. We're not in a rush; why not do it right the first time around?

In the [NQP PR](https://github.com/perl6/nqp/pull/319/files) this line just finds the filename that the [gen-cat script](https://github.com/rakudo/rakudo/blob/ee8ae92e423f27a4df55ce98b0797a9c556e8bc2/tools/build/gen-cat.nqp#L23) makes, right?:
```
+    regex comment:sym<line_directive> {
 +        ^^ '#' \s* 'line' \s+ $<line>=(\d+) [ \s+ $<filename>=(\S+) ]? $$
 +    }
 +
```

So what if we [change the gen-cat script line 23](https://github.com/rakudo/rakudo/blob/ee8ae92e423f27a4df55ce98b0797a9c556e8bc2/tools/build/gen-cat.nqp#L23) from `say("#line 1 $file\n");` to something like `say("#line 1 Rakudo::Internals/$file\n");?

Then instead of `src/core` to detect a core file, we can use `.starts-with('Rakudo::Internals/')`. 

Is there a problem with this approach?

MasterDuke17:
I don't really have a problem with that approach. However, you'd need to remove the 'Rakudo::Internals/' when printing the backtrace to reflect the actual file paths. @jnthn didn't like modifying the values in Backtrace.pm (which my first attempt at implementing this did), but maybe simply removing a static string would be fine.

zoffixznet:
I would be fine with leaving Rakudo::Internals/ in when prining..

lizmat:
I'm n favour of adding a marker, but would be against using Rakudo::Internals, to prevent confusion with the Rakudo/Internals.pm file.

How about just prefixing it with SETTING:: ?  That would be sort of appropriate anyway  :)

zoffixznet:
> How about just prefixing it with SETTING:: ? That would be sort of appropriate anyway :)

Sounds good to me.

MasterDuke17:
Change made, @lizmat++, @zoffixznet++

zoffixznet:
Thanks a lot.

PR #921, opened 2016-11-19
Remove no longer required special case for r-j

usev6:
I was unable to find out why this special case was needed back in 2015.
A spectest run looked good on rakudo-j (no new failures).


zoffixznet:
Thanks a lot.

PR #925, opened 2016-11-22
perl6 --doc=Text show signature parameter pod

kalkin:


lib/Pod/To/Text.pm6
```diff
@@ -113,9 +113,12 @@ sub declarator2text($pod) {
 
 sub signature2text($params) {
       $params.elems ??
-      "(\n\t" ~ $params.map({ $_.perl }).join(", \n\t") ~ "\n)" 
+      "(\n\t" ~ $params.map(&param2text).join("\n\t") ~ "\n)" 
       !! "()";
 }
+sub param2text($p) {
+    $p.perl ~ ',' ~ ( $p.WHY ?? ' # ' ~ $p.WHY !! ' ')
```
ugexe: Concatenating a comma like this  will result in possible hanging commas unlike the previous method of using `join`

zoffixznet:
How would I go about triggering the output this PR produces? I built your branch, but I'm not seeing the difference from stock rakudo, so I'm guessing I'm not using the right syntax?

```
cpan@perlbuild2/tmp/tmp.zabPzgifR7/rakudo (nom)$ cat foo.p6 
sub foo(
    #|{ Just some ASD param } 
    $asd
) { $asd.WHY.say }(42)
cpan@perlbuild2/tmp/tmp.zabPzgifR7/rakudo (nom)$ ./perl6 -Ilib --doc=Text foo.p6 
class $asd
Just some ASD param
cpan@perlbuild2/tmp/tmp.zabPzgifR7/rakudo (nom)$ 
```

kalkin:
You have to have pod attached to the subroutine like this:
```perl6
#| THIS MAKES THE DIFFERENCE
sub foo( #|{ Asd  } 
    $ads ) {
    $ads.say
}

```

zoffixznet:
Thanks a lot.

For the trailing comma: IMO it's fine to have it, since it's perfectly Perlish and the way to get rid of it is either messy or requires passing some sort of `:this-is-last-element` to `param2text` which is also yuky, so IMO leaving it in preferred.

ugexe:
The trailing comma can be removed without knowing the last element by passing all the items to be processed

    sub params2text(*@_) {
        @_.map: { qq|{.perl}{.WHY??", # {.WHY}"!!""}| }
    };

    say @params.elems
        ?? "\n\t" ~ @params.&params2text.join(",\n\t")
        !! "()";

PR #927, opened 2016-11-23
Fix add history for readline, since the method is called add-history,…

samcv:
… it's the function that's called add_history


zoffixznet:
Thanks

PR #930, opened 2016-11-28
Avoid calling nqp::substr with negative offset

usev6:
... when we are interested in the preceding character.


usev6:
This makes the following test from S04-statements/terminator.t pass on rakudo-j:

```throws-like '{my $x = 2;', X::Syntax::Missing, 'open closure';```

PR #936, opened 2016-12-05
Fix nqp-lib for installing perl6-{gdb,valgrind}-m

Benabik:
Recent changes (a803a153b3138de944f89637517cff6328b0e793) broke my ability to install Rakudo (OS X, using DESTDIR).  It failed trying to installed perl6-gdb-m.  It looks like the lines for installing the debug runner didn't get updated with the blib changes.


niner:
Thanks for fixing my mess! Turns out, hacking build system code after only 5 hours of sleep may not be the best of ideas ;)

Benabik:
Build systems are the worst.  There's always some combination of stuff that you don't test.  :-D

PR #938, opened 2016-12-07
Stop relying on a bug in infix:<**>(Int, Int).

pmurias:


zoffixznet:
```
<babydrop> There's a point on that topic that bugged me before:
<babydrop> m: say 2e0 ** 99999999999999
<camelia> rakudo-moar 5bacb0: OUTPUT«Inf␤»
<babydrop> m: say 2 ** 99999999999999
<camelia> rakudo-moar 5bacb0: OUTPUT«Numeric overflow␤  in block
        <unit> at <tmp> line 1␤␤Actually thrown at:␤  in block <unit> at <tmp> line 1␤␤»
<babydrop> So... what if we remove all those checks and feed the args straight to pow_I; 
          giving an Inf on overflow?
<babydrop> And then we have consistency between Num, and Int candidates
*babydrop* notes pmurias isn't here, so comments on the PR instead
```

https://irclog.perlgeek.de/perl6-dev/2016-12-07#i_13699034

PR #941, opened 2016-12-08
Empty emit fix

Altai-man:
Fixes #130291: https://rt.perl.org/Public/Bug/Display.html?id=130291

The comment is removed because of https://irclog.perlgeek.de/perl6/2016-12-07#i_13698178


zoffixznet:
Thanks a lot!

PR #943, opened 2016-12-09
Add .gist, .perl, and .Str to BOOTSTRAPATTR

MasterDuke17:
Fixes RT77070, https://rt.perl.org/Ticket/Display.html?id=77070

Passes `make m-spectest`


jonathanstowe:
Totally for this one, if nothing else it makes examining the attributes of some Meta objects easier :)

PR #944, opened 2016-12-10
Make uniprop 500% to 15% faster. Return Bools for Bool Uni properties. Return correct properties for short names.

samcv:
* uniprop is now 500% faster when no arguments are given, 15% faster for valid property names and ~20% faster for invalid ones.

* Return Bool for boolean Unicode properties instead of 1 or 0

* Fix getting properties for Unicode short names
* There is a problem with MoarVM where some of the short unicode names
are not mapped to the right function. For now we should fix this
in Rakudo. (https://github.com/MoarVM/MoarVM/issues/448)

The code was generated with a script, `tools/build/makeUNIPROP.pl6`

That script also includes a test to check that short/alternate Unicode names return the same
result as the primary name. It failed a large number before, but with this change it
passed all of them.


samcv:
I have rebased it for the changes in my last pull request and also made it 15% faster than it was before (see edited PR body text). The old version was 50% slower than before, so now it is 15% faster plus works better :smile:

lizmat:
Cool!

If you have code that generates part of the setting, please include the script in tools/build.  Also make sure that you can read in the code that a part of it has been generated.  See examples of that in tools/build/makeNATIVE_ARRAY.pl6 and src/core/native_array.pm

src/core/Cool.pm
```diff
@@ -362,17 +362,101 @@ multi sub unimatch(|)     { die 'unimatch NYI on jvm backend' }
...
-    state %prefs;  # could prepopulate this with various prefs
-    given %prefs{$propname} // '' {
+    given nqp::atkey(%prefs, $propname) {
         when 'S' { nqp::getuniprop_str($code,$prop) }
         when 'I' { nqp::getuniprop_int($code,$prop) }
-        when 'B' { nqp::getuniprop_bool($code,$prop) }
-        # your ad here
-        default {
+        when 'B' { nqp::p6bool(nqp::getuniprop_bool($code,$prop)) }
+            default {
```
MasterDuke17: Looks like there's an extra indent here.

samcv:
Added the script that generated the code and updated Cool.pm with the updated code (which tells you where generated code starts and ends).

samcv:
Added commit, so when there are no arguments given, it can use a faster codepath, making it 5x faster.

PR #945, opened 2016-12-11
Implement `uniprops` function, also allow both `uninames` and `uniprops` to take a List.

samcv:
S15(DRAFT) says that uninames and uniprops only take strings, but I think that it reasonable to expect the plural functions to accept the same types as the singular functions, which accept either Int's or Str.

Would like peoples thoughts on this, if anybody is opposed.


zoffixznet:
Is there a concrete usecase in mind for these two?

```
+multi sub uniprops(@strs, Stringy:D $propname = "GeneralCategory") {
 +  @strs.map: { uniprop($_, $propname) }
 +}

multi sub uninames(@c) { @c.map: { uniname($_) } }
```

I'm 👎 on them, because I rather not add candidates just because we can (still reeling from the `splice` nightmare). Also, they allow the user to stick any sort of positional in there without any error checking, so I foresee tickets for LTA errors for when the given positional contains an `IntStr` ("ambiguous dispatch doesn't tell which item it is") or LTA behaviour when a multi-grapheme string is given ("only first char is returned for a string with no warning").

So unless there's some decent usecase for these, I rather we not add them.

moritz:
It's also a bit weird to have a scalar argument after an array, we tend to do it the other way 'round, if at all.

samcv:
@zoffixznet uniprop and uniname both accept either an integer or a string. Is there a reason the `IntStr` problem would not be in those two but would affect uniprops and uninames?

samcv:
Did not have a specific use case, I just found it odd that uniname and uniprop both accept integers but having uninames and uniprops not accept integers and only strings is a little weird (at least to me)
```perl6
say (50,11,50).uninames.perl => ("DIGIT TWO", "LINE TABULATION", "DIGIT TWO").Seq
```

zoffixznet:
> Is there a reason the IntStr problem would not be in those two but would affect uniprops and uninames?

The problem is still there, it's just the error references the user's call and the single invocant/argument that's problematic:

```
<Zoffix> m: <42>.uniname
<camelia> rakudo-moar 3d4825: OUTPUT«Ambiguous call to 'uniname'; these signatures all match:␤:(Str:D $str)␤:(Int:D $code)␤  in block <unit> at <tmp> line 1␤␤»
```

Whereas with `uniprops`/`uninames` that error would not make much sense to the user since they're not calling `uniname`/`uniprop` and the invocant/argument would not necessarily be an IntStr but can be a list of things as well.

> odd that uniname and uniprop both accept integers but having uninames and uniprops not accept integers and only strings is a little weird

To me it makes sense that the plurals wouldn't be taking integers, since to get the info for all chars in a string, you'd need to `.comb` it, whereas with Ints you'd already have them in a list and can just call:

```
<Zoffix> m: say (50,11,50)».uniname.perl
<camelia> rakudo-moar 3d4825: OUTPUT«("DIGIT TWO", "LINE TABULATION", "DIGIT TWO")␤»
```

But by adding the candidates that can take `@a`, you're making calls like these also valid:

```perl6
    say (50, "meows", 999, "f", "42").uninames.perl;
```

samcv:
Okay, since people seem opposed to it I have removed the positional accepting multi's. I also renamed "GeneralCategory" to "General_Category" since the name with the underscore is the canonical name of the Unicode property.

zoffixznet:
Thanks a lot.

PR #954, opened 2016-12-13
Fix for RT #130211

Altai-man:
https://rt.perl.org/Public/Bug/Display.html?id=130211

Seems quite simple, but this is nearly my first experience with nqp and a metamodel part in particular, so I'm sure there is a simpler/prettier way to do that.

~~Also, I didn't test performance. As I see it, it shouldn't be too much overhead, but still.~~

DON'T MERGE IT YET.

### DIFF
--- a/src/Perl6/Metamodel/RoleToClassApplier.nqp
+++ b/src/Perl6/Metamodel/RoleToClassApplier.nqp
@@ -46,15 +46,22 @@ my class RoleToClassApplier {
         # role first.
         my $to_compose;
         my $to_compose_meta;
+        my %method_to_role;
         if +@roles == 1 {
             $to_compose := @roles[0];
             $to_compose_meta := $to_compose.HOW;
+            for $to_compose_meta.method_table($to_compose) {
+                nqp::bindkey(%method_to_role, $_.key, $to_compose.HOW.name($to_compose));
+            }
         }
         else {
             $to_compose := $concrete.new_type();
             $to_compose_meta := $to_compose.HOW;
             for @roles {
                 $to_compose_meta.add_role($to_compose, $_);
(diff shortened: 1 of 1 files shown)


arnsholt:
This isn't my area of expertise either, but it looks pretty straightforward and sane to me. I have only comment, which is the case where a method with the same name is required by more than one role. In that case, I think only the last one mentioned in the code is used.

And if you're worried about performance, you can move the generation of the lookup table to inside the error condition (where there's currently just an `nqp::die`). That way, any performance penalty will only be hit if there's actually an error; in which case a few dozen milliseconds don't really matter.

Altai-man:
@arnsholt, you are correct: I tested it with more than one role and in that case only the last one is printed.
If we want to print all not implemented methods, then we need to stack them up in an array, perhaps, on checking time and die afterwards with this array.
But this wasn't mentioned in original issue, so I didn't make any more complex changes unless someone from core team say so.

zoffixznet:
Do you think it's possible to move all of this logic into the that $yada conditional where we die? Currently we do all this work and store all the names of methods just to throw it all away for 99.9999% of code that won't die.

Seems like all the variables used to build %method_to_role are available inside that conditional, so we could bypass creation of %method_to_rol and we could find the name of the role only when we're ready to die.

arnsholt:
Oh, actually we *can* move the whole thing into the error case. I was confused by the error case being inside `compose_method_table`, but that's an inner function to `apply` so `@roles` is directly available inside `compose_method_table` too.

Altai-man:
Yes, you are correct. I'll move it inside error block now.

zoffixznet:
Merged per https://irclog.perlgeek.de/perl6/2016-12-13#i_13732777

PR #955, opened 2016-12-13
Fix Readline REPL not initializing properly.

samcv:
Would break since the last commits to add a history file in.
Fix this.


zoffixznet:
Thanks.

PR #956, opened 2016-12-14
Add missing MVMROOT in finddispatcher.

zhuomingliang:


lizmat:
FWIW, this spectests cleanly but doesn't fix the HARNESS_TYPE=6 instability

zhuomingliang:
It only fixes potential problem, but now it never happened yet.

PR #957, opened 2016-12-15
Show all roles that have missing method in error message

Altai-man:
```
> EVAL 'role R { method overload-this(){...} }; role C { method overload-this(){...} }; class A does R does C {};'; CATCH { default { say .^name, .Str } };
X::Comp::AdHocMethod 'overload-this' must be implemented by A because it is required by roles: C, R.
```
See https://irclog.perlgeek.de/perl6/2016-12-13#i_13732692 for discussion.
cast @zoffixznet.

Tests for https://rt.perl.org/Public/Bug/Display.html?id=130211 depend on approve or reject of this PR, so I'll write one after this PR.

### DIFF
--- a/src/Perl6/Metamodel/RoleToClassApplier.nqp
+++ b/src/Perl6/Metamodel/RoleToClassApplier.nqp
@@ -108,16 +108,18 @@ my class RoleToClassApplier {
                 if $yada {
                     unless has_method($target, $name, 0)
                             || has_public_attribute($target, $name) {
+                        my @needed;
                         for @roles {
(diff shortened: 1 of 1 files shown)


zoffixznet:
Thanks.

> Tests for https://rt.perl.org/Public/Bug/Display.html?id=130211 depend on approve or reject of this PR, so I'll write one after this PR.

Try to write ones that don't restrict the content of the message too much, since other implementation may choose to emit a differently-worded message. Alternatively, you can write a test in https://github.com/rakudo/rakudo/blob/nom/t/05-messages/01-errors.t instead

PR #958, opened 2016-12-16
Update changelog.

samcv:


zoffixznet:
Thank you.

FWIW, it's fine to leave adding stuff to changelog to just me. 

Basically I have a site— http://perl6.fail/release/stats —that tracks what commits I logged (the commits are on the "Commits" tab; the ones that have been added have a line on the left side). And several times a month I go through new commits in order and add them to the changelog, while marking them as added on the site.

P.S.: I saw you asked when Rakudo gets released. It's 3rd Saturday of the month, if there are no delays with the MoarVM release, it's usually starts around 11AM EST and it's done with a bot in #perl6-dev channel, so you can't miss it....

samcv:
Ah ok thanks for the info. I suspected you had some system.

I submitted it partially because you forgot to add Readline history in the additions section, and just went ahead and added the others.

zoffixznet:
Thanks. Indeed, I did. I think I assumed the `inputrc` stuff was part of all that....

PR #960, opened 2016-12-17
Speed up perl6 -ne '' <large file> by ~40%

MasterDuke17:
By re-writing IO::ArgFiles.lines.pull-one and IO::Handle.lines.pull-one
completely in NQP.

Passes `make m-spectest`


zoffixznet:
[From IRC](https://irclog.perlgeek.de/perl6/2016-12-21)

```
11:28        notviki	m: use nqp; dd nqp::defined(Failure.new)
11:28	camelia	rakudo-moar b8df3a: OUTPUT«1␤»
11:28	notviki	m: use nqp; dd Failure.new // 'differ'
11:28	camelia	rakudo-moar b8df3a: OUTPUT«"differ"␤»
11:29	notviki	MasterDuke: ^ looks like nqp::defined is not a direct replacement for // ?
                                                  Can $!handle.get return a Failure?
11:29	lizmat	notviki: that's because Failure.defined is special
11:29		it disables the Failure and makes it disappear by pretending it wasn't defined
11:30		nqp::defined is not so subtle
11:30	notviki	MasterDuke: aslo, in the previous version, $limit could be any numeric,
                                    but now it'll be silently ignored unless it's an Int. I think
                                   we could stick an Int() coercer in the sig
11:33		m: sub (Int()) {}(*)
11:33	camelia	rakudo-moar b8df3a: OUTPUT«No such method 'Int' for invocant of type 'Whatever'␤  in sub  at <tmp> line 1␤  in block <unit> at <tmp> line 1␤␤»
11:33	notviki	MasterDuke: oh, I guess we can't just stick a coercer on $limit
                               'cause it could be Whatever
```

MasterDuke17:
Re nqp::defined() vs .defined(), $io is returned if it's a Failure and won't get to the definedness check at all (same logic as the original code).

Re $limit, at https://github.com/MasterDuke17/rakudo/blob/nom/src/core/IO/ArgFiles.pm#L80 it gets turned into an Inf if it is a Whatever. I could add an ` || nqp::istype($limit, Num) && nqp::isle_n(nqp::unbox_n($limit), 0)`, but slows things down a bunch. Given that roast doesn't have anything to say about Int v Num limits right now, I'd vote for coercing to Int instead.

zoffixznet:
> I could add an || nqp::istype($limit, Num) && nqp::isle_n(nqp::unbox_n($limit), 0)

> I'd vote for coercing to Int instead. This would also mean I could remove the nqp::if(nqp::istype($limit, Int)... before decrementing.

Now I'm really confused about what you're proposing. You can't coerce an Inf to a Int... so how would an Inf limit be represented?

---

The limit can be any `Numeric` type (it can easily be a Rat, for example, if someone is calculating the limit by dividing something) along with `Whatever`. Because it's an integral counter, we are free to coerce almost all values to an Int. There are 4 edge-case values that can't be represented in an Int: `Whatever`, `-Inf`, `+Inf`, and `NaN`. `-Inf` is equivalent to `0`, while the other three are equivalent to `+Inf`.

So in [this conditional](https://github.com/MasterDuke17/rakudo/blob/nom/src/core/IO/ArgFiles.pm#L80), we just need to handle the extra cases:

```perl6
my $l = nqp::istype($limit,Whatever) || $limit == Inf || $limit === NaN ?? Inf
       !! $limit == -Inf ?? 0
            !! $limit.Int;
```

The rest of the code will remain the same. The `istype(.. Int)` check will remain as a way to check for Inf $limit.

zoffixznet:
On further inspection of our other features they don't handle `-Inf`/`NaN` cases, so to maintain consistency we shouldn't handle them here either, so the conditional would just need an extra check for `+Inf` (since we can't coerce it to Int):

```perl6
my $l = nqp::istype($limit,Whatever) || $limit == Inf ?? Inf !! $limit.Int;
```

MasterDuke17:
Sorry, I had two ideas in my head and they got mixed up in the writing. Your proposal is pretty much exactly what I meant. I have some other ideas, but they would be for after this got merged.

MasterDuke17:
Oh, and `$handle.get ` only returns a `str` or `Nil`. `nqp::defined(Nil)` is false, so should be good there.

lizmat:
Good catch!

PR #961, opened 2016-12-18
Convert some bindattrs to direct binding

MasterDuke17:
See https://irclog.perlgeek.de/perl6-dev/2016-12-17#i_13759010 for
reference.

Passes `make m-spectest`


lizmat:
Thanks for the digging!

PR #963, opened 2016-12-19
Fix {List,Seq,Range}.is-lazy method for type objects invocants

lucasbuchala:
Before:

```
> List.is-lazy
Cannot look up attributes in a List type object

> Range.is-lazy
Cannot look up attributes in a Range type object

> Seq.is-lazy
Invocant requires an instance of type Seq, but a type object was passed.  Did you forget a .new?
```

After:

```
> say List.is-lazy
False

> say Range.is-lazy
False

> say Seq.is-lazy
False
```

### DIFF
--- a/src/core/List.pm
+++ b/src/core/List.pm
@@ -998,7 +998,7 @@ my class List does Iterable does Positional { # declared in BOOTSTRAP
         self.new(|c);
     }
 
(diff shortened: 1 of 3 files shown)


zoffixznet:
Thanks a lot.

PR #964, opened 2016-12-19
Use #!/usr/bin/env perl instead of #! perl

samcv:
Unless there is a good reason to keep this, it would be nice
if this was changed for better compatibility.


lizmat:
how does that work on Windows?

jnthn:
On Windows you would just run "perl Configure.pl", or rely on file
associations. Windows itself doesn't pay any attention to #! lines.

PR #966, opened 2016-12-22
Handle IO::ArgFiles.lines(Inf)

MasterDuke17:
Base everything off the common case (much more similar to IO::Handle.lines now).


zoffixznet:
Thanks a lot.

PR #967, opened 2016-12-23
Unbust JVM build (findmethod)

usev6:
With nqp commit 49d6beb34a33780212039f5179b032659582a3f0 one version of findmethod
was removed. Using the remaining version seems to work fine. There might be some
spectest failures related to changes in this area, but at least rakudo-j does not
die while compiling.


pmurias:
The method has been renamed (so that we don't have to with the same name
and arguments in different order) not removed :(

pmurias:
Looking at the changes more closely it seems that the method you used was
generally more correct then the one used before.

PR #970, opened 2016-12-24
Travis CI: build on Ubuntu 14.04 instead of 12.04 and use GCC 6

samcv:
This pull request is ready to be merged if it is wanted.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,4 @@
-language: "perl"
+language: "java"
 
 os:
     - linux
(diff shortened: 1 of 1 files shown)


samcv:
This is now complete. Does anybody have any opinion on this at all?

samcv:
I was mostly wondering if the flapping native call test could stop flapping if we make this change, and we can always go back to it if the test stops flapping (we will know there is some difference).

https://irclog.perlgeek.de/perl6-dev/2016-12-24#i_13792601

zoffixznet:
Thanks.

ugexe:
Seems wrong to change the language from perl to java just to keep from uncommenting the line to install the jdk (which should only be needed for the nvm anyway?)

samcv:
@ugexe if that isn't changed on the Travis CI 14.04 it doesn't have javac in the path at the moment, which is why it was changed. I would have just installed it if it were that simple.

If there is another way to specify the javac path *easily* let me know

ugexe:
I assume thats because you tried `openjdk-7-jre` when you really wanted `openjdk-7-jdk`

ugexe:
@samcv ^ ?

PR #972, opened 2016-12-24
Fix bug in makeUNIPROP.p6 causing it not to generate 2 props

samcv:
I forgot to flush the code generator function at the end, so it
ended up missing the last two properties:
'Full_Composition_Exclusion' and 'Canonical_Combining_Class'


zoffixznet:
Thanks.

PR #974, opened 2016-12-25
uniprop: return Bool's for Emoji properties to prepare for MoarVM adding

samcv:
With this pull: https://github.com/MoarVM/MoarVM/pull/468
We will have all Emoji properties except Emoji_All. With that change to
MoarVM and this change to Rakudo we now pass the roast Emoji property
tests (aside from Emoji_All).


samcv:
Done in this commit:
https://github.com/rakudo/rakudo/commit/3baffe71a8e1b49d709d9bcbce228c023587e69d

PR #975, opened 2016-12-25
remove mass-chars.t from spectest.data

samcv:
To see why this was removed see:
https://github.com/perl6/roast/commit/3baa4db85abfe21d4df1eb9fc1149fd1f30dda15
    
 This test has been superceeded by the recently added GraphemeBreakTest.t


samcv:
Updated the comment/title, which for some reason were accidentally the same ones as my previous pull.

zoffixznet:
We can't remove it from the list, because there's just one t/spectest.data for both 6.c-errata and master roast

samcv:
Oh, got it. That makes sense :)

PR #978, opened 2016-12-30
Make new Proc::Async constructor friendly to subclass attribs.

skids:


skids:
Test case:

$ perl6 -e 'class foo is Proc::Async { has $.fooo; method new (*@stuff) { nextwith(|@stuff, :42fooo) } }; foo.new("/bin/echo").fooo.say;'
42

... $.fooo is "Any" without this patch.

Finding other custom constructors that lack this might make good LHF for new contributors

zoffixznet:
Thanks!

> Finding other custom constructors that lack this might make good LHF for new contributors

It's worth noting just sticking a slurpy on like that is very expensive, so don't want it done that way in all classes, so I wouldn't call that an LHF. Related Ticket: https://rt.perl.org/Ticket/Display.html?id=130020#ticket-history

PR #980, opened 2016-12-31
[JVM] Type Uni is not usable on rakudo-j

usev6:


lizmat:
Thanks!   Should have realized that when I added that :-(

PR #986, opened 2017-01-02
Added new test entry for socket-fail-invalid-values.t

dugword:
Updated entry for tests for RT issues:
https://rt.perl.org/Ticket/Display.html?id=130475
https://rt.perl.org/Ticket/Display.html?id=130473


dugword:
Pull request for tests: https://github.com/perl6/roast/pull/212

PR #987, opened 2017-01-05
Use submethod TWEAK instead of method init

titsuki:
I think `submethod TWEAK` is more suitable than user-defined init method in this case.


titsuki:
Thanks !

PR #988, opened 2017-01-05
Test.pm's is_approx isn't being removed until 6.d

MasterDuke17:
See https://irclog.perlgeek.de/perl6-dev/2017-01-05#i_13860967

Un-spectested, YMMV


zoffixznet:
Thanks.

PR #990, opened 2017-01-06
Handle all Nd digits in regex backslash errors

MasterDuke17:
/\۳/ now suggests $2, before it just didn't recognize it.

Passes `make m-spectest`


zoffixznet:
FWIW, I see a Zefram-type ticket for this in the future: with the new behaviour, the `۳\x[308]3` would parse fine, but suggest the user replaces it with `$0`

Also, since the exclusion uses only ASCII zero, the suggested replacement for `٠` (arabic zero), would end up being `$-1`

MasterDuke17:
@zoffixznet, right, now it excludes them all.

`/\۳\x[308]3/` says `Unrecognized backslash sequence (did you mean $2?)`

MasterDuke17:
Still passes `make m-spectest`.

PR #991, opened 2017-01-06
Handle Nd's in ${} special variables

MasterDuke17:
Passes `make m-spectest`.


MasterDuke17:
Test added in https://github.com/perl6/roast/commit/2667e3edfc

samcv:
Merged. Thanks!

PR #997, opened 2017-01-18
Bandaid for r-j: don't bind nqp::null to attribute

usev6:
Checking for nqp::null is fast, but does not work on JVM (cmp. RT #130532).

This commit should be reverted once the underlying issue in nqp-j is fixed.


zoffixznet:
Thanks

PR #998, opened 2017-01-20
Implement `last` and `redo` for whenever blocks

timo:
AlexDaniel++ for the inspiration on `next` (which is already in nom), and also lizmat wondering what `redo` would do, both of which led me to continue with `last` and`redo`.

The code isn't as pretty as it probably could be, and perhaps it even performs badly. That is for the reviewers to decide! :P


lizmat:
Looks good to me, spectests clean:

Interaction with "next" seems to be ok:

$ 6 'react { whenever Supply.from-list(^10) { print "$_ "; redo unless $++ %% 2; next unless $++ %% 3 } }'
0 1 1 1 1 1 1 2 2 2 2 2 2 3 3 3 3 3 3 4 4 4 4 4 4 5 5 5 5 5 5 6 6 6 6 6 6 7 7 7 7 7 7 8 8 8 8 8 8 9 9 9 9 9 9

However, interaction with "last" seems incorrect?

$ 6 'react { whenever Supply.from-list(^10) { print "$_ "; last } }'
0 1 2 3 4 5 6 7 8 9 

Also, a LAST phaser appears to be called twice:

$ 6 'react { whenever Supply.from-list(^10) { print "$_ "; last; LAST say "goodbye" } }'
0 1 2 3 4 5 6 7 8 9 goodbye
goodbye

zoffixznet:
Pending changes from the OP: https://irclog.perlgeek.de/perl6/2017-03-18#i_14286879

lizmat:
Feels this still needs a bit of work and some discussion?  Or maybe this PR should be closed?

jnthn:
@lizmat Seems to me like this feature already works in `master`, and the couple of issues with `last`/`LAST` mentioned in the PR appear to not be a problem any more. So probably can be closed, though as a bonus could make sure the examples here are test-covered.

PR #1001, opened 2017-01-21
Comment since it warns with rakudo star roast runs which confuses people.

stmuk:
We probably need proper test versioning (v6.c whatever).


zoffixznet:
Yes, we almost got it for last night's release, but I ran out of tuits. The next release will definitely have separate test file lists and so the warnings will not be spurious any more, so I'll close this PR without merging.

Thanks.

PR #1002, opened 2017-01-22
Don't mention *CORE.setting in error message

usev6:
References to CORE.setting.moarvm where removed with 4ae08e0b24,
but rakudo-j points to gen/jvm/CORE.setting (no .jar there).

Fixes RT #130509 and RT #124767 on JVM


lizmat:
Thanks.  But a more generic solution would be having either Moar having no extension either, or having the JVM mention the extension.

usev6:
Yeah, I agree. I took a look, but didn't find the place where that is set. So I went to put the additional alternation in for now.

PR #1004, opened 2017-01-25
Initial implementation of non-blocking await for 6.d.PREVIEW

jnthn:
So far, when you `await` in code running on a thread in the thread pool, it would block a real OS thread. This meant that certain patterns had to be expressed in unnatural ways (using `.then` and a callback), or that the scheduler had to have its maximum number of threads raised really high. For example, sleepsort on a hundred or a thousand values would run into issues there if you wrote it with `await`.

My original intention for `await`, when used on a thread pool thread, was that it would instead hand control back to the scheduler so the thread could be used for other work. This would mean we could have thousands of tasks awaiting stuff, and just a handful of threads required for progress.

This PR implements exactly that. As a bonus, it also makes `await` in the case of an already available result not require concurrency control in the `Promise` case, which should be a speed-up. Furthermore, there's now an interface for things that wish to participate in the `await` system to implement, so while we just use it for `Promise`, `Supply`, and `Channel` today, module space will also be able to create their own things that we can `await`.

Unless you `use v6.d.PREVIEW`, nothing changes as of merging this PR. You have to opt in to get this new behavior of `await`. Once Perl 6.d is released, then Rakudo will switch to 6.d being the default and - unless code has explicitly declared itself as wanting 6.c - this will become the default behavior for all `await`s on thread pool threads. (And `await` in the mainline, or in a bare `Thread`, will continue to have the same blocking behavior as today, just implemented through the new `Awaitable` interface.)

### DIFF
--- a/src/core/Awaitable.pm
+++ b/src/core/Awaitable.pm
@@ -1,2 +1,48 @@
+# An Awaitable is something we can use the `await` operator on. To support
+# this, it requires a `get-await-handle` method be implemented, which returns
+# an `Awaitable::AwaitHandle`.
 my role Awaitable {
+    method get-await-handle() { ... }
+}
+
+# An Awaitable::Handle implementation is an immutable object that conveys the
+# status of the requested asynchronous result at the point we obtain the
+# handle. If the `.already` property is `True`, then there is no need to block
+# or suspend execution; the `.result` or `.cause` of failure can be used right
+# away (depending on the value of `.success). Otherwise, the consumer of the
+# handle should call the `subscribe-awaiter` method with its unblock/resume
+# handler, and then proceed to block/suspend. In this case, the handler will
+# be passed two arguments: a `Bool` success, and a result/cause (result if
+# success is `True`, cause if it's `False`). The `Awaitable::Handle` will
+# *not* have its success/result/cause updated; this would open the door to
+# data races (including subtle ones related to read/write ordering), when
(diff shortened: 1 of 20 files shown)


src/core/Awaitable.pm
```diff
@@ -1,2 +1,48 @@
...
+# data races (including subtle ones related to read/write ordering), when
+# the point of the fast-path is to test if we've got a result already with
+# minimal overhead (and thus minimal concurrency control).
+my role Awaitable::Handle {
+    has Bool $.already;
+    has Bool $.success;
+    has Mu $.result;
+    has Exception $.cause;
+
+    method already-success(\result) {
```
jonathanstowe: Strangely I was thinking about this the other day, there seems to be a pattern in Scala where a method that returns a Future may return a Future.failed directly rather than throwing an exception.

jnthn: I've been considering `Promise.kept(value)` and `Promise.broken(exception)` Promise factories that would cheaply create such `Promise` objects. Ran into a few situations in my own code where I missed them.

jonathanstowe:
For myself I am very excited by this,  there is a fair amount of  work-around in Tinky to stop it being impacted awaiting on lots of validators.

jnthn:
Feedback has been positive, and nobody has registered any objections, so I'll go ahead and merge this. It's low risk for existing code anyway, since you have to write `use v6.d.PREVIEW` to get the changed functionality.

PR #1008, opened 2017-01-31
New fancier fudgeandrun

ronaldxs:
Built for [roast issue #226](https://github.com/perl6/roast/issues/226).  New fudgeandrun is backwards compatible but can also be used to run tests for roast repository without rest of rakudo repository.  This fudgeandrun runs under Windows.  Defaults to running fudged tests with "prove -v -e(./?)perl6 ..." but has a -6|--six option for backwards compatibility.  Also has a -q|--quiet option to turn off '-v' with prove.  Trying to run with no files dumps usage including options.  Handles multiple test files consistently.

### DIFF
--- a/t/fudgeandrun
+++ b/t/fudgeandrun
@@ -2,8 +2,6 @@
 use strict;
 use warnings;
 
-use List::Util qw(first);
-
 ######################################################################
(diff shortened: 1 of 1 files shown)


zoffixznet:
Thanks.

PR #1009, opened 2017-02-01
Add unicmp op which uses unicmp_s nqp op to do string compare

samcv:


b2gills:
I don't like the name, for one the string comparison operator is `leg`.

I also don't have any ideas about what would be a good name.

zoffixznet:
> I don't like the name, for one the string comparison operator is `leg`.

.oO( `lug` ...)

samcv:
leg just calls cmp_s, which is the same as cmp does, and the underlying nqp op is unicmp_s

Any thoughts on better names?

zoffixznet:
> leg just calls cmp_s, which is the same as cmp does, and the underlying nqp op is unicmp_s

But there are other `cmp` candidates for all sorts of types, whereas `leg` is just for strings.

> Any thoughts on better names?

Not really, though I find `unileg` sounding even more amusing than plain `leg` :P

samcv:
Commited this directly and it is hidden behind `use experimental :unicmp`

https://github.com/rakudo/rakudo/commit/6f6f0cf19ed10e8c7e3e828b3a776eb2c3220fc1
https://github.com/rakudo/rakudo/commit/5870ef97d1cd7f94d578a8359a70fddbf57e9820

PR #1010, opened 2017-02-01
replace s///r in fudgeandrun for compatibility with perl 5.10

ronaldxs:
As recommended by @zoffixznet on IRC


zoffixznet:
Thanks.

PR #1011, opened 2017-02-03
RFC: increase readability of REPL.pm by reducing verbosity

wchristian:
Note ahead: I'm not a seasoned Perl6 dev and may have misunderstood syntax. I also am not currently in a situation to run the tests, so i have not verified that those remain working the same.

This does nothing to change the functionality of REPL.pm, i hope. It does however reduce the complexity of the file as far as reading it goes, by flattening out some of the deeper indentation trees and reducing some superfluous syntax.

There's a few more places i could've done, and i'm happy to do a second pass, but i wanted to hear first if you guys would want to merge something like this in.

### DIFF
--- a/src/core/REPL.pm
+++ b/src/core/REPL.pm
@@ -96,10 +96,8 @@ do {
 
         method teardown-line-editor {
             my $err = linenoiseHistorySave($.history-file);
-
-            if $err {
-                note "Couldn't save your history to $.history-file";
-            }
+            return if !$err;
+            note "Couldn't save your history to $.history-file";
         }
 
         method repl-read(Mu \prompt) {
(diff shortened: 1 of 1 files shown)


src/core/REPL.pm
```diff
@@ -96,10 +96,8 @@ do {
 
         method teardown-line-editor {
             my $err = linenoiseHistorySave($.history-file);
-
-            if $err {
-                note "Couldn't save your history to $.history-file";
-            }
+            return if !$err;
```
zoffixznet: Personally, I write `if !...` as `unless ...` or as `if not ...` or as `if ! ...`. I tripped up twice reading this PR 'cause I missed the `!` 😄

zoffixznet:
Thanks. Tested and left some feedback for your second pass, if you wish to do it 😄

PR #1012, opened 2017-02-04
Sort the binaries by version

jonathanstowe:
Assuming we always want to have the latest version of a binary
unless it has been otherwise specified then this seems to be
the simplest way to achieve that.

### DIFF
--- a/src/core/CompUnit/Repository/Installation.pm
+++ b/src/core/CompUnit/Repository/Installation.pm
@@ -84,7 +84,7 @@ sub MAIN(:$name is copy, :$auth, :$ver, *@, *%) {
     }
 
     %*ENV<PERL6_PROGRAM_NAME> = $*PROGRAM-NAME;
-    exit run($*EXECUTABLE, @binaries.sort(*<ver>)[*-1].hash.<files><bin/#name#>, @*ARGS).exitcode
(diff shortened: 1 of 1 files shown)


lizmat:
perhaps .tail instead of [*-1] ??  :-)

niner:
This will improve newly installed scripts but won't help for those which are already there. If instead we sort in the "files" method, it'd apply immediately to all installed scripts. And as right now there is no defined sort order to files' results, it shouldn't hurt to fit it to the scripts' use case.

jonathanstowe:
Yes sorting in .files would be better, I didn't want to do that as I don't know where else it us used.

niner:
Nothing in rakudo uses this method and as I said, the order is currently undefined anyway.

jonathanstowe:
Hmm, though on reflection it seems to me that sorting in .files won't entirely fix it as they may appear in more than one Repository (though I don't know how often that will happen in practice.)

niner:
Oh, that's true. Ok, this code needs some major refactor anyway because let's face it. It's not pretty :)

jonathanstowe:
All good, I may revisit by moving the "find me the binaries" into a method in the CUR at some point :)

PR #1013, opened 2017-02-04
Don't constant fold x op when the result is big

MasterDuke17:
If the resulting string is very large, it can take a long time to encode
it on the string heap and then decode it when needed.

Some discussions:
https://irclog.perlgeek.de/perl6/2017-02-04#i_14040978
https://irclog.perlgeek.de/perl6/2017-02-04#i_14042387

Fixes RT #127972, https://rt.perl.org/Public/Bug/Display.html?id=127972

Passes `make m-spectest`.


zoffixznet:
Thanks.

PR #1014, opened 2017-02-04
Better error message for invalid “return” trait mod

AlexDaniel:
Previously it either complained about a missing block or spewed some
internal gibberish. Now it suggests “returns”.

Resolves RT #130712

Spectest OK.


zoffixznet:
Thanks.

PR #1016, opened 2017-02-07
Emit properly nested blocks for the combination of default value to p…

pmurias:
…arams and subs/->.


JJ:
This conflict does not seem easily fixable, since the code in master introduces several new variables. Can you please check it and see if solving it is doable?

lizmat:
I'm unsure on what this is trying to fix.  So I'm unsure how to test what this is doing, when trying to unbitrot this PR.  Could someone enlighten me?

lizmat:
Nobody enlightened, so closing

PR #1017, opened 2017-02-08
Support smartmatching char Ranges

MasterDuke17:
Before `say ("a" .. "c") ~~ ("a" .. "c")` would give the error:

`Cannot convert string to number: base-10 number must begin with
valid digits or '.' in '⏏a' (indicated by ⏏)`

Passes `make m-spectest`.

Inspired by https://irclog.perlgeek.de/perl6/2017-02-08#i_14061478

### DIFF
--- a/src/core/Range.pm
+++ b/src/core/Range.pm
@@ -426,13 +426,23 @@ my class Range is Cool does Iterable does Positional {
         nqp::istype(($_ := got.Real), Failure) ?? False !! nextwith $_
     }
     multi method ACCEPTS(Range:D: Range \topic) {
-        (topic.min > $!min
-         || topic.min == $!min
-            && !(!topic.excludes-min && $!excludes-min))
-        &&
-        (topic.max < $!max
-         || topic.max == $!max
-            && !(!topic.excludes-max && $!excludes-max))
+        $!is-int && topic.is-int
+          ??
+            (topic.min > $!min
+             || topic.min == $!min
+                && !(!topic.excludes-min && $!excludes-min))
+            &&
+            (topic.max < $!max
+             || topic.max == $!max
+                && !(!topic.excludes-max && $!excludes-max))
+          !!
+            (topic.min after $!min
+             || topic.min eq $!min
+                && !(!topic.excludes-min && $!excludes-min))
+            &&
+            (topic.max before $!max
+             || topic.max eq $!max
+                && !(!topic.excludes-max && $!excludes-max))
     }
 
     multi method AT-POS(Range:D: int \pos) {


samcv:
Awesome! Going to merge.

AlexDaniel:
This barely makes any sense.

```perl6
(topic.min after $!min
    || topic.min eq $!min
```

So you are mixing a string operator (``eq``) with a generic comparison (``after``). This does not look right at all. Did you mean ``gt``?

AlexDaniel:
And then there is [RT #129131](https://rt.perl.org/Ticket/Display.html?id=129131), are you sure that all of the weirdness of character ranges were accounted for?

samcv:
gt is also a generic operator

AlexDaniel:
Except that it stringifies stuff before comparing?

zoffixznet:
It's generally a mistake to use string ops to compare two numerics. `.is-int` only marks integer ranges, but we also have Num, Rat, and FatRat ranges. For example this comparison is now broken and throws:

```
<Zoffix> m: say 1..Inf ~~ -1/0..1/0
<camelia> rakudo-moar 0d8624: OUTPUT«Attempt to divide 1 by zero using div␤  in block <unit> at <tmp> line 1␤␤Actually thrown at:␤  in block <unit> at <tmp> line 1␤␤»

<Zoffix> star: say 1..Inf ~~ -1/0..1/0
<camelia> star-m 2016.10: OUTPUT«True␤»
```

There are probably more examples with Rats: `<=>` compares two rats using their components, while `after` compares them by coercing them to Num first.


I also notice this string ranges made up of string numerics used to DWIM but now give False. Was this intended?:

```
<Zoffix> m: say '2'..'3' ~~ 0..10
<camelia> rakudo-moar 0d8624: OUTPUT«False␤»

<Zoffix> star: say '2'..'3' ~~ 0..10
<camelia> star-m 2016.10: OUTPUT«True␤»
```

samcv:
It would seem not, well at least
```perl6
1 ~~ '1'
```
still works.
We should add these new smart match problems to roast and then revert this commit or fix its issues I think.

zoffixznet:
> well at least 1 ~~ '1' still works.

That's not a range so it wouldn't be affected by this commit.

AlexDaniel:
+1 for a revert. This is a dramatic change, I'd much rather see an RFC ticket before doing something like this.

samcv:
@zoffixznet yes it would not be affected, that was my point (that this is allowed so it should be for ranges).

samcv:
Will revert it now.

samcv:
@zoffixznet should this work? You seemed to indicate it should work fine?
```perl6
say 1..Inf ~~ -1/0..1/0 # True␤
```

samcv:
I have added a test to roast for:
```perl6
ok '2'..'3' ~~ 0..10, "Can smart match string Range's which hold numbers and Ranges which are numbers";
```

zoffixznet:
> should this work? `say 1..Inf ~~ -1/0..1/0 # True␤` You seemed to indicate it should work fine?

Answered on IRC, but for anyone else reading this... Yes, that should work same as `say 1..Inf ~~ -Inf..Inf`. While zero-denominator Rats explode when you try to stringify them, their `.Num` coercion is non-explosive and follows [IEEE 754-2008](http://www.csee.umbc.edu/~tsimo1/CMSC455/IEEE-754-2008.pdf) semantics, per section 7.3. So `0/0` is a `NaN`, while all others are `Inf`, with the same sign as the numerator.

So the behaviour pre-this-commit used numeric comparison which used the Num "view" of Rats (and thus the range was two infinites), but the new Str comparison uses the Str "view" of Rats which does explode with division by zero.

----

There was also a relevant [discussion on IRC](https://irclog.perlgeek.de/perl6-dev/2017-02-08#i_14064592) about this commit. And basically the comparison should base the comparator used based on whether the RHS Range is numeric or stringy, regardless of which type of Range LHS is. So that means:

```perl6
say '0'..'9' ~~ 0..10; # is True, the strings get coerced to numerics
say 0..10 ~~ '0'..'9'; # is True, the number get coerced to strings and '10' is bother
      # before '9' and after '0'; note that on current HEAD (with this commit reverted)
      # this currently incorrectly gives False
say 'a' .. 'c' ~~ 'a' .. 'c' # is True (currently explodes trying to coerce to numeric)
say 'a' .. 'c' ~~ 0 .. 42 # Currently this explodes due to strings coerced to numerics.
       # This matches the
       # behaviour of `'a' ~~ 1` but I'd argue both cases should just give False instead
       # of exploding, with the reasoning being that if I *knew* the types of the 
       # arguments, I'd use either `==` or `eq` accordingly; and smartmatch implies I 
         # don't know and so it should not explode at me when I give it things that can't
          # be compared right.

zoffixznet:
@MasterDuke17 did you want to submit another PR implementing discussed semantics or am I free to do that myself?

I added tests to roast that spec the wanted behavior as discussed above and on IRC: https://github.com/perl6/roast/commit/e152a0ff3a

Unfudged, they crash due to coersion of strings to numeric (those smartmatches must just give False). I tried to use the fudger to fudge them but it just hangs, so I fudged it manually. To unfudge, remove [these two lines](https://github.com/perl6/roast/blob/e152a0ff3a37ef56c4c68af09421b47d5ad9d93c/S03-smartmatch/range-range.t#L29-L30) and [this one](https://github.com/perl6/roast/blob/e152a0ff3a37ef56c4c68af09421b47d5ad9d93c/S03-smartmatch/range-range.t#L86-L87)

MasterDuke17:
I'm probably not going to be able to do a complete implementation right away, please feel free to do it yourself if you want to.

PR #1018, opened 2017-02-12
Make Parameter.[named_names, type_captures] list_s

MasterDuke17:
Even with just `perl6 -e ''` lots of Parameters are created, this reduces
the memory used by these attributes, as well as the CPU needed to
box/unbox them.

Passes `make m-spectest`. Reduces the count of BOOTStrs by 432 in
a heap profile of `perl6 -e ''`.

### DIFF
--- a/src/Perl6/Actions.nqp
+++ b/src/Perl6/Actions.nqp
@@ -2752,7 +2752,7 @@ class Perl6::Actions is HLL::Actions does STDActions {
             my @params := %sig_info<parameters>;
             @params.unshift(hash(
                 is_multi_invocant => 1,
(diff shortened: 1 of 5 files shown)


timo:
i don't see anything wrong with it :+1:

jnthn:
I think the JVM binder will also need an update for this. As it stands, I suspect this will bust the JVM build.

MasterDuke17:
Ahh, @timo reminded me to test my intifying PR with the JVM, but I haven't tested this one yet, I'll do that.

MasterDuke17:
@jnthn Thanks for the help, JVM fix is now committed.

PR #1020, opened 2017-02-17
Revert "forbid to put variable in initializer of its declaration"

AlexDaniel:
Aka “Allow self-referential arrays in declarations”

This reverts commit 3754356d8e4401c2e567f20462adb2ecd97dc7d5.

TL;DR Reverted commit does more harm than good, and with recent
changes is no longer needed as much. There are ways you can shoot
yourself in the foot, yes, and we should find a way to print awesome
error messages in these cases.

The change in the reverted commit achieves very little. In theory, it
prevented obvious ways to create a VMNull (and nulls used to blow
up). An example of the issue that was masqueraded:

    my $a := $a; say $a  # segfaulted previously

At the same time, it didn't prevent any other ways of achieving the
same thing. For example:

    my $a := $::a; say $a  # segfault, even after the fix!

However, the change prevented legitimate creation of self-referential
arrays:

    my @a = 42, @a    # ← complains needlessly

And forced to use workarounds to achieve it:

    my @a = 42, @::a  # ← does not complain and works perfectly.
                      # Congratulations, you've tricked the detection!

A proper way to fix the issue would be to prevent the segfault.
It was done in ae2ae92330e5b861b47500de0d8b0855297757a3 (actual
MoarVM/nqp commit unknown).

There's a good point that the user may have wrong assumptions on how
self-referential stuff should work with binding, and we should have a
compile-time error for this case. DIHWIDT, but would be nice if
somebody came up with a proper way to detect this.

In other words: it does not segfault anymore, so this “fix” is not
needed. And even though the behavior is LTA when you get a VMNull
anywhere, it does not warrant the prevention of a legitimate use of
self-referential arrays (or hashes, pairs, lists, etc.).

The revert had conflicts because of these commits:
* 1458f0ea3ea2943252c4436fd66a22861845f8ad
* ae651044245eabcfe172813ad3f5bab377025f51

I had to tweak it a bit to resolve the issues, should be OK.

Obviously, doesn't pass spectest because of misguided tests for the
reverted commit. This has to be corrected in errata.


AlexDaniel:
https://github.com/perl6/roast/pull/236

zoffixznet:
👍 on the change, but seems this PR would make 6.c-errata fail. These tests all expect X::Syntax::Variable::Initializer to exist, but this PR removes it: https://github.com/perl6/roast/blob/6.c-errata/S04-declarations/my.t#L162-L179

AlexDaniel:
@zoffixznet well, these tests are removed in my roast PR. The PR is not for 6.c-errata, but the same commit can be used to fix that…

As for changing the errata, I guess it's not a problem if others agree that these tests were trying to achieve something completely different (the actual issue with ``is default(…)`` is different from what the error message says; ``my $z = $z`` has always worked just fine; the result of using binding with self-referential things was never actually defined except for this attempt to cover a segfault).

zoffixznet:
> As for changing the errata

I'd prefer to change errata as little as possible, especially in cases where it's not easy to say whether the test was incorrect. Perhaps this can be left to v6.d? We're starting to have infrastructure to support it, so we can probably figure out a way to stick it into v6.d.PREVIEW

Also, noticing the case below doesn't actually work and its error is LTA; perhaps the current error isn't as useless?

```
<Zoffix> m: my @foo := 1..3, (@::foo Z+ 100); say @foo
<camelia> rakudo-moar a162c4: OUTPUT«No such method 'iterator' for invocant of type 'Mu'␤  in block <unit> at <tmp> line 1␤␤»
```

AlexDaniel:
Its error is as LTA as if you do this:
```perl6
say Mu Z 1
```
```
No such method 'iterator' for invocant of type 'Mu'
  in block <unit> at -e line 1
```
> especially in cases where it's not easy to say whether the test was incorrect

Well, luckily this is not the case… ;)

zoffixznet:
[Comments from perlpilot](https://irclog.perlgeek.de/perl6-dev/2017-02-28#i_14181701):
```
19:19 	perlpilot 	IOninja: I don't know how much brain you want connected to the eyes, but it looked good to me.  :)
19:20 	IOninja 	perlpilot: so you're OK with removing 6.c-errata tests that PR breaks?
19:22 	perlpilot 	that or fixing them if they're trying to test something useful and it's just the X::Whatever name that's problematic
19:22 		(I don't know if the X::Syntax::Variable::Initializer was invented for the thing this reverts or not)
19:22 		If it was, then yes, remove them.
```

zoffixznet:
I'm on the side of rejecting this:

1) Because it doesn't fully address the core issue: there are still cases where assignment doesn't work and the error does make sense.
2) It requires modification of 6.c-errata tests; removing one of the Exception types in particular. Considering there *are* cases where the error makes sense, in a fully-developed fix, such removal isn't required.

AlexDaniel:
1. Yes, similarly to how a broken clock is right once a day, there is indeed **one** case when the error is what one can maybe call right (when using binding), and in the rest of the cases the error is wrong.
2. What tests are we talking about exactly? There is one test that tests ``my %h is default(%h<foo>)``, and the error has nothing to do with what actually happens. That test has to be removed anyway. And then there is another test with ``my $z = $z``, which, interestingly, has never been a problem. The last test is for ``my @foo := 1..3, (@foo Z+ 100)``, which, again, is a perfectly valid code except that one should use normal assignment instead of binding. In theory, there could be an error which would suggest normal assignment, that would be pretty nice indeed. Anyway, I've already explained this in #236.

We can wait for 6.d, I guess (but then it is unclear what should be done with wrong behavior right now). Or we can try to somehow keep the error in this one “useful” case mentioned above… Meh.

AlexDaniel:
FWIW holding this until somebody (hopefully me when I get time for it) implements a helpful error message for ``my %h is default(%h<foo>)`` seems like the right thing to do.

AlexDaniel:
Relevant RT tickets:
* [RT#87034](https://rt.perl.org/Ticket/Display.html?id=87034) [open]: [BUG] Either binding an array to a list containing the array itself should work consistently or not at all in Rakudo
* [RT#121807](https://rt.perl.org/Ticket/Display.html?id=121807) [open]: [LTA] is default(…) does not explain that the value should be known at compile time
* [RT#125371](https://rt.perl.org/Ticket/Display.html?id=125371) [open]: Segv on binding to one self and saying it

AlexDaniel:
I know it has been a while, but I'm still planning to do something about it. Sorry for being slow.

AlexDaniel:
Looked into this today, I really hate the fact that it's so easy to introduce a hack that helps with one issue but creates other problems, but it's so hard to revert it back. I don't have enough energy for this now or in the near future, but we'll have to revert it in the end some day, so I'm leaving this open.

lizmat:
I propose closing this PR: it has bitrotted, and will become moot when RakuAST is merged.  I suggest maybe opening a problem-solving ticket about self-referential initializations to be handled properly in the RakuAST branch.

vrurg:
Makes sense closing it.

jnthn:
> will become moot when RakuAST is merged

Yes, it would need implementing in a very different way there.

> I suggest maybe opening a problem-solving ticket about self-referential initializations to be handled properly in the RakuAST branch

Makes sense, given whatever semantics are picked will need to be implemented afresh. The functionality proposed for removal has caught thinkos for me a number of times, while I don't recall a time when I tripped over it. It is smart enough not to get annoyed about mentions of the variable inside of a code object being assigned as the initializing value, for when using it to define recursive things. Thus I consider it more right than wrong; if self-referential arrays are really so common, we could make an exception by sigil, but the number of times that's useful vs. catches a thinko, I don't know.

PR #1021, opened 2017-02-21
Optimize Actions.(dec|hex|oct|bin)int()

MasterDuke17:
Create a string_to_int that uses nqp::radix instead of nqp::radix_I.
Then call that from Actions.(dec|hex|oct|bin)int() instead of
string_to_bigint when the number of chars in the string to be converted
is small enough that it couldn't cause overflow.

Passes `make m-spectest`.

If there was a way to tell whether we're on a 32 or 64 bit system in
Actions.nqp the number of chars at which we switch to nqp::radix_I
could be increased.

### DIFF
--- a/src/Perl6/Actions.nqp
+++ b/src/Perl6/Actions.nqp
@@ -631,10 +631,17 @@ class Perl6::Actions is HLL::Actions does STDActions {
         }
     }
 
-    sub string_to_bigint($src, $base) {
+    sub string_to_int($src, int $base, int $chars) {
+        my $res := nqp::radix($base, ~$src, 0, 2);
+        $src.CURSOR.panic("'$src' is not a valid number")
+            unless nqp::iseq_i(nqp::atpos($res, 2), $chars);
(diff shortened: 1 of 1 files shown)


samcv:
Did you see any difference in speed?

lizmat:
This is the code in src/core/Rakudo/Internals.pm:

my constant $?BITS = do {
    my int $a = 0x1ffffffff;
    nqp::iseq_i($a,8589934591) ?? 64 !! 32
}

I think something like that can work in NQP as well  :-)

MasterDuke17:
@lizmat thanks, I"ll add that.

@samcv in my testing, the inclusive time for decint (which just calls string_to_(big)?int) went down. Its exclusive time increased, but by less than the combination of string_to_int+string_to_bigint exclusive time decreased, so there was an overall savings. decint ended up being something like 20% faster by inclusive time, but it's so fast to begin with that the overall savings aren't very noticeable.

MasterDuke17:
@samcv, @lizmat added a commit that adjusts when to switch to string_to_bigint based on whether the system is 32 or 64 bit.

PR #1023, opened 2017-02-23
Test to cover RT #130712

Altai-man:
https://github.com/perl6/roast/pull/244 - previous conversation.
Closes https://rt.perl.org/Public/Bug/Display.html?id=130712


zoffixznet:
Thanks.

PR #1027, opened 2017-02-25
Change atpos of @!named_names to atpos_s

MasterDuke17:
These were missed in the recent conversion of @!named_names to an
nqp::list_s.

Fixes the recent DBDish install failure spotted by @zoffixznet++.

Passes `make m-spectest`.


zoffixznet:
Thanks.

PR #1029, opened 2017-03-01
add ⁇ ‼ unicode version

FCO:
add ⁇ ‼ unicode version

### DIFF
--- a/src/Perl6/Grammar.nqp
+++ b/src/Perl6/Grammar.nqp
@@ -4250,9 +4250,8 @@ grammar Perl6::Grammar is HLL::Grammar does STD {
     token prefix:sym<->   { <sym> <O(|%symbolic_unary)> }
     token prefix:sym<−>   { <sym> <O(|%symbolic_unary)> }
     token prefix:sym<??>  { <sym> <.dupprefix('??')> <O(|%symbolic_unary)> }
-    token prefix:sym<?>   { <sym> <!before '⁇' || '??'> <O(|%symbolic_unary)> }
-    token prefix:sym<⁇>  { <sym> <.dupprefix('⁇')> <O(|%symbolic_unary)> }
-    token prefix:sym<!>   { <sym> <!before '‼' || '!!'> <O(|%symbolic_unary)> }
+    token prefix:sym<?>   { <sym> <!before '??'> <O(|%symbolic_unary)> }
+    token prefix:sym<!>   { <sym> <!before '!!'> <O(|%symbolic_unary)> }
     token prefix:sym<|>   { <sym>  <O(|%symbolic_unary)> }
     token prefix:sym<+^>  { <sym>  <O(|%symbolic_unary)> }
     token prefix:sym<~^>  { <sym>  <O(|%symbolic_unary)> }
--- a/src/Perl6/Grammar.nqp
+++ b/src/Perl6/Grammar.nqp
@@ -4391,7 +4391,7 @@ grammar Perl6::Grammar is HLL::Grammar does STD {
     token infix:sym<min>  { <sym> >> <O(|%tight_or_minmax)> }
     token infix:sym<max>  { <sym> >> <O(|%tight_or_minmax)> }
 
-    token ternary-explan {
+    token ternary-explain {
         || <?before '::' <.-[=]>> { self.typed_panic: "X::Syntax::ConditionalOperator::SecondPartInvalid", second-part => "::" }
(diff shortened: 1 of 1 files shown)


src/Perl6/Grammar.nqp
```diff
@@ -1742,6 +1742,7 @@ grammar Perl6::Grammar is HLL::Grammar does STD {
     token term:sym<value>              { <value> }
     token term:sym<unquote>            { '{{{' <?{ $*IN_QUASI }> <statementlist> '}}}' }
     token term:sym<!!>                 { '!!' <?before \s> }  # actual error produced inside infix:<?? !!>
+    token term:sym<‼>                 { '‼' <?before \s> }  # actual error produced inside infix:<⁇ ‼>
```
AlexDaniel: Nitpicking, but why misaligned?

src/Perl6/Grammar.nqp
```diff
@@ -4348,6 +4349,22 @@ grammar Perl6::Grammar is HLL::Grammar does STD {
         <O(|%conditional, :reducecheck<ternary>, :pasttype<if>)>
     }
 
+    token infix:sym<⁇ ‼> {
```
AlexDaniel: So this whole thing is just copy-pasted? Any better solution?

src/Perl6/Grammar.nqp
FCO: I didn't see that... fixing...

AlexDaniel: Oh, ‼ is probably double-width in your font. OK.

src/Perl6/Grammar.nqp
FCO: yes, just copy-pasted. I don't know if there is a better solution... any suggestion?

FCO: Not copy-pasted any more...

src/Perl6/Grammar.nqp
```diff
@@ -4332,19 +4333,32 @@ grammar Perl6::Grammar is HLL::Grammar does STD {
     token infix:sym<min>  { <sym> >> <O(|%tight_or_minmax)> }
     token infix:sym<max>  { <sym> >> <O(|%tight_or_minmax)> }
 
+    token ternary-explain {
+        || <?before '::' <.-[=]>> { self.typed_panic: "X::Syntax::ConditionalOperator::SecondPartInvalid", second-part => "::" }
```
samcv: I don't think this should have the leading `||`

AlexDaniel: Why not? That's being used almost everywhere, and it's just a syntactic thing to make the stuff aligned.

samcv: There's no reason for it to be there. Change it to spaces or something. Though are you saying we have useless leading `||` in other places too?

AlexDaniel: Yes. In fact, the syntax itself was added on purpose.

Don't change it.

zoffixznet:
@FCO, did you run `TEST_JOBS=8 make spectest` (or better yet, `TEST_JOBS=8 make stresstest`, since this is a pretty core change)?

See you moved some stuff to a separate token, but it still uses `~$<EXPR>` which isn't available to it anymore, so it's always Nil. If our tests don't catch that, we need to improve them.

FCO:
I made the ?? !! and the ⁇ ‼ a single token. I think that fixes the @zoffixznet, @AlexDaniel and @samcv  issues. I've ran the spec tests and the stress tests, I got 2 errors on the first and 1 on the last one. Both on spec tests worked when I ran only the test (t/spec/S10-packages/basic.rakudo.moar and t/spec/S17-supply/watch-path.t). The stresstest error was on file t/spec/S10-packages/precompilation.rakudo.moar 2 errors on TODOs and 1 real error. @zoffixznet told me on irc that that test "flop on occasion".

AlexDaniel:
Looks nicer indeed.

zoffixznet:
Thanks.

src/Perl6/Grammar.nqp
```diff
@@ -4391,17 +4392,17 @@ grammar Perl6::Grammar is HLL::Grammar does STD {
     token infix:sym<max>  { <sym> >> <O(|%tight_or_minmax)> }
 
     token infix:sym<?? !!> {
-        :my $*GOAL := '!!';
-        $<sym>='??'
+        $<sym>=['⁇' || '??']
+        {}:my $*GOAL := $<sym> eq '??' ?? '!!' !! '‼';
```
vendethiel: why the leading `{}`?

zoffixznet: To update $/

PR #1031, opened 2017-03-02
Implement Str.parse-names / &parse-names

zoffixznet:
Functionality:
- similar to `"\c[BELL, BLACK HEART SUIT]"`, except can be used by
    users at runtime, without requiring `EVAL`/`nqp` in user's code
- Unlike `\c[…]` does not allow comments or anything similarly esoteric.
    The idea is this will process some sort of user input, so we want
    it just as liberal as necessary (whitespace around char names is allowed).

Naming:
- the name follows `Str.parse-base` that parses a string containing numerals
    in some base; whereas `Str.parse-names` parses a string of character names

------

Sample usage:

```perl6
    say parse-names('BELL, BLACK HEART SUIT'); # output: 🔔♥
    say 'BELL, BLACK HEART SUIT'.parse-names; # output: 🔔♥
```

### DIFF
--- a/src/core/Exception.pm
+++ b/src/core/Exception.pm
@@ -1810,6 +1810,14 @@ my class X::Phaser::PrePost is Exception {
     }
 }
 
+my class X::Str::InvalidCharName is Exception {
+    has $.name;
+    method message() {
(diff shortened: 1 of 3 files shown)


coke:
+1

PR #1032, opened 2017-03-05
Add ≤ ≥ ≠ unicode ops

AlexDaniel:
🙌

Tiny IRC discussion:
https://irclog.perlgeek.de/perl6/2017-03-01#i_14183585

⩵ (and especially ⩶) were left out intentionally. For now, let's focus
on operators that were actually requested many times.

Spectest clean.

### DIFF
--- a/src/Perl6/Grammar.nqp
+++ b/src/Perl6/Grammar.nqp
@@ -4343,8 +4343,11 @@ grammar Perl6::Grammar is HLL::Grammar does STD {
     token infix:sym«≅»    { <sym>  <O(|%chaining)> }
     token infix:sym«==»   { <sym>  <O(|%chaining)> }
     token infix:sym«!=»   { <sym> <?before \s|']'> <O(|%chaining)> }
+    token infix:sym«≠»    { <sym>  <O(|%chaining)> }
     token infix:sym«<=»   { <sym>  <O(|%chaining)> }
+    token infix:sym«≤»    { <sym>  <O(|%chaining)> }
     token infix:sym«>=»   { <sym>  <O(|%chaining)> }
+    token infix:sym«≥»    { <sym>  <O(|%chaining)> }
     token infix:sym«<»    { <sym>  <O(|%chaining)> }
     token infix:sym«>»    { <sym>  <O(|%chaining)> }
     token infix:sym«eq»   { <sym> >> <O(|%chaining)> }
--- a/src/core/Numeric.pm
+++ b/src/core/Numeric.pm
@@ -287,6 +287,7 @@ sub infix:<=~=>(|c) { infix:<≅>(|c) }
 proto sub infix:<!=>(Mu $?, Mu $?) is pure  { * }
 multi sub infix:<!=>($?)        { Bool::True }
 multi sub infix:<!=>(Mu \a, Mu \b)   { not a == b }
+sub infix:<≠>(|c) is pure { infix:<!=>(|c) }
 
 proto sub infix:«<»(Mu $?, Mu $?) is pure   { * }
 multi sub infix:«<»($?)         { Bool::True }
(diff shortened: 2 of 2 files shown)


lizmat:
This makes these unicode operators inherently slower than their Texas counterparts.  Why not just create aliases?

  my &infix:<≠> := &infix:<!=>

??

jnthn:
Because the alias is made a startup time if you use `:=`. Thus `&infix:<≠>` is considered a variable rather than a static symbol, and so optimizers must assume it can change. As a result, spesh will refuse to inline it, or even emit optimized calling code for it. You might get it to inline it if you alias it using `my constant ...`. But the way it is in the original patch is the only one I'm certain it's going to be able to inline.

jnthn:
I did some measurements:

```
jnthn@lviv:~/edument/rmtly/try-it-out$ time perl6-m -e 'sub a($x, $y) { }; for ^1000000 { a(1,2) }'
real	0m0.815s
user	0m0.780s
sys	0m0.028s
jnthn@lviv:~/edument/rmtly/try-it-out$ time perl6-m -e 'sub a($x, $y) { }; my constant &b := &a; for ^1000000 { b(1,2) }'

real	0m0.889s
user	0m0.844s
sys	0m0.024s
jnthn@lviv:~/edument/rmtly/try-it-out$ time perl6-m -e 'sub a($x, $y) { }; my &b := &a; for ^1000000 { b(1,2) }'

real	0m1.055s
user	0m1.040s
sys	0m0.012s
jnthn@lviv:~/edument/rmtly/try-it-out$ time perl6-m -e 'sub a($x, $y) { }; sub b($x, $y) { a($x, $y) }; for ^1000000 { b(1,2) }'

real	0m1.278s
user	0m1.228s
sys	0m0.028s
```

The improvement of `constant` over `my` is pretty notable, making it the best way to go. At present, it also seems that the arg passing-on overhead is higher than the cost of the missed inline, so the originally suggested change would in the immediate term be an improvement. However, that is liable to improve, while the variable case is not.

tl;dr alias using constant.

AlexDaniel:
I looked at how other things are implemented, and this is what I found:
* ``×`` and ``÷`` are implemented similarly to this PR ([src](https://github.com/rakudo/rakudo/blob/9cec31afee774e73dc49753e188c712bb450c1d4/src/core/Numeric.pm#L199)).
* All set ops are implemented like this as well ([src](https://github.com/rakudo/rakudo/blob/9cec31afee774e73dc49753e188c712bb450c1d4/src/core/set_operators.pm#L60))
* That's also how ``≅`` is implemented ([src](https://github.com/rakudo/rakudo/blob/9cec31afee774e73dc49753e188c712bb450c1d4/src/core/Numeric.pm#L285))
* ``−`` too ([src](https://github.com/rakudo/rakudo/blob/9cec31afee774e73dc49753e188c712bb450c1d4/src/core/Mu.pm#L965))

The only alternative op that is done differently is ``o`` ([src](https://github.com/rakudo/rakudo/blob/9cec31afee774e73dc49753e188c712bb450c1d4/src/core/operators.pm#L718)):
```perl6
my &infix:<o> := &infix:<∘>;
```

@lizmat do you think I should change it for this PR, or perhaps all alternative ops should be changed at once later?

lizmat:
I would like to see that in a separate PR.  Possibly with timings  :-)

src/Perl6/Grammar.nqp
```diff
@@ -4343,8 +4343,11 @@ grammar Perl6::Grammar is HLL::Grammar does STD {
     token infix:sym«≅»    { <sym>  <O(|%chaining)> }
     token infix:sym«==»   { <sym>  <O(|%chaining)> }
     token infix:sym«!=»   { <sym> <?before \s|']'> <O(|%chaining)> }
+    token infix:sym«≠»    { <sym> <?before \s|']'> <O(|%chaining)> }
```
AlexDaniel: True. I was thinking about that too. Fixed now.

AlexDaniel:
@lizmat yea, I think I can do that.

zoffixznet:
Thanks.

PR #1033, opened 2017-03-07
Bump NQP_REVISION to get bumped MOAR_REVISION

bdw:
This fixes a JIT issue with integer signed division, which would not correctly round to minus infinity in the range [-1,0].


zoffixznet:
Thanks.

PR #1034, opened 2017-03-08
Fix multiple requires & require clobbering pre-existing pacakges

LLFourn:
Fixes:
```
class Test::Foo { }; { require Test; say Test::Foo }
require This;
require This::And::That;
```


lizmat:
I'll leave this to nine to look at :-)

PR #1035, opened 2017-03-12
Add narrower candidates for infix min and max

MasterDuke17:
Add :D candidate for Int, int, Num, and num and implement them in nqp.
Makes `Int:D max Int:D` about 3x faster.

Passes `make m-spectest`.


zoffixznet:
There isn't a :U `int`, so probably don't need `:D` on those candidates... I'm surprised that usage doesn't warn/throw.

MasterDuke17:
Ah, right, I'll remove them.

zoffixznet:
Thanks.

PR #1036, opened 2017-03-12
fix installation path of core dist

dod38fr:
Wihout this patch, all core files are installed in `/usr/share/perl6` (build uses `--prefix=/usr`):

```
$ ls /usr/share/perl6
bin  dist  lib  precomp  repo.lock  resources  runtime  short  site  sources  vendor  version
```

And the directory which is specified in perl6 script (with `--libpath="/usr/share/perl6/lib"` ) is empty:
```
$ ls /usr/share/perl6/lib/
[empty]
```

This patch fixes this issue. When applied, all core files are installed in `/usr/share/perl6/lib`.
This is currently applied in [Debian's rakudo_2017.02-1](https://packages.debian.org/source/experimental/rakudo):
```
$ ls /usr/share/perl6/
lib  runtime  site  vendor  version
$ ls /usr/share/perl6/lib/
dist  precomp  repo.lock  short  sources  version
```

All the best


niner:
What is the issue this patch would fix?

dod38fr:
My bad. I cannot reproduce the test that led me to create this patch. On top of that, it does break loading of core modules.

Sorry for the noise.

PR #1037, opened 2017-03-13
Enhance Pod::To::Text

kalkin:
* Rework handling of FormattingCodes
* Pod::To::Text add handling for mor FormattingCodes
* Pod::To::Text fix twine2text routine
* Pod::To::Text beautify headings
* Fix Pod::To::Text join failing when displaying a positional
* Pod::To::Text fix showing Parameter as class
* Pod::To::Text do not skip enums

### DIFF
--- a/lib/Pod/To/Text.pm6
+++ b/lib/Pod/To/Text.pm6
@@ -157,9 +157,34 @@ my %formats =
 
 sub formatting2text($pod) {
     my $text = $pod.contents>>.&pod2text.join;
-    $pod.type ~~ %formats
-      ?? colored($text, %formats{$pod.type})
-      !! $text
+    given $pod.type {
+        when <L> {
+            if $pod.meta.first {
+                my $meta = $pod.meta.first;
+                if $meta.starts-with: "/language/" {
+                    $meta = "doc:" ~ $meta.comb[10..*].join.split('#').first;
+                } elsif !$meta.contains(":") {
+                    $meta = "doc:" ~ $meta;
+                }
+                $text ~= " <" ~ $meta ~ ">" ;
+            }  else {
+                $text ~= " <doc:" ~ $text ~ ">" ;
+            }
+            colored($text, %formats{$pod.type});
+        }
+        when %formats {
+            colored($text, %formats{$pod.type})
+        }
+        when <N> {
+            " ($text)"
+        }
+        when <E>|<P> {
+            "$pod.type()\<$text\>";
+        }
+        default {
+            $text;
+        }
+    }
 }
 
 sub twine2text($twine) {
--- a/lib/Pod/To/Text.pm6
+++ b/lib/Pod/To/Text.pm6
@@ -161,7 +161,7 @@ sub twine2text($twine) {
     return '' unless $twine.elems;
     my $r = $twine[0];
     for $twine[1..*] -> $f, $s {
-        $r ~= twine2text($f.contents);
+        $r ~= $f.?contents ?? twine2text($f.contents) !! $f.Str;
         $r ~= $s;
     }
     $r;
--- a/lib/Pod/To/Text.pm6
+++ b/lib/Pod/To/Text.pm6
@@ -98,6 +98,7 @@ sub declarator2text($pod) {
         when .HOW ~~ Metamodel::EnumHOW {
             "enum $_.perl() { signature2text $_.enums.pairs } \n"
         }
+        when Parameter { next }
         when .HOW ~~ Metamodel::ClassHOW {
             'class ' ~ $_.perl
         }
--- a/lib/Pod/To/Text.pm6
+++ b/lib/Pod/To/Text.pm6
@@ -95,6 +95,9 @@ sub declarator2text($pod) {
         when Sub {
             'sub ' ~ $_.name ~ signature2text($_.signature.params)
         }
+        when .HOW ~~ Metamodel::EnumHOW {
+            "enum $_.perl() { signature2text $_.enums.pairs } \n"
+        }
         when .HOW ~~ Metamodel::ClassHOW {
             'class ' ~ $_.perl
         }
--- a/lib/Pod/To/Text.pm6
+++ b/lib/Pod/To/Text.pm6
@@ -144,10 +144,15 @@ sub param2text($p) {
 }
 
 my %formats =
-  C => "bold",
-  L => "underline",
-  D => "underline",
-  R => "inverse"
+  C => "green",
+  B => "red",
+  L => "blue",
+  X => "blue",
+  D => "cyan",
+  I => "bold",
+  R => "inverse",
+  T => "on_blue",
+  K => "magenta",
 ;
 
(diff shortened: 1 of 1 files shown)


zoffixznet:
Haven't reviewed this, but I'm a bit 👎  on all the colour additions (or rather their replacement of bold+underline option). The "on_blue" one is unreadable on my terminal color scheme and I flipped through a handful default ones and red is unreadable on another one. Moreover, color deficiency in humans isn't rare. Would be nice if "bold+underline+reverse" color scheme was still available for those who want it.

P.S.: that `use MONKEY-SEE-NO-EVAL;` on line 10 is useless.

untitled

lib/Pod/To/Text.pm6
```diff
@@ -33,14 +33,32 @@ sub pod2text($pod) is export {
...
+            $result ~= "\n$rest" if $rest.defined;
+            $result.trim-trailing;
+        }
         default { '    ' ~ pod2text($pod.contents)  }
     }
 }
 
 sub code2text($pod) {
-    "    " ~ $pod.contents>>.&pod2text.subst(/\n/, "\n    ", :g)
+    my $text = "    " ~ $pod.contents>>.&pod2text.subst(/\n/, "\n    ", :g);
```
zoffixznet: FWIW, `"\n"` instead of `/\n/` in `subst` is about 3x faster.

lib/Pod/To/Text.pm6
```diff
@@ -121,24 +144,54 @@ sub param2text($p) {
...
 sub formatting2text($pod) {
     my $text = $pod.contents>>.&pod2text.join;
-    $pod.type ~~ %formats
-      ?? colored($text, %formats{$pod.type})
-      !! $text
+    given $pod.type {
+        when <L> {
+            if $pod.meta.first {
+                my $meta = $pod.meta.first;
+                if $meta.starts-with: "/language/" {
```
zoffixznet: This appears to be something specific for the docs.perl6.org? This module is a general-use module, so this should be removed.

zoffixznet:
*ping*

Any updates with respect to the requested changes?

kalkin:
@zoffixznet Sorry. I was on sick leave. Now I'm partly back.

> Moreover, color deficiency in humans isn't rare. Would be nice if "bold+underline+inverse" color scheme was still available for those who want it.

Currently by default the colors are not enabled at all, which I think is less then awesome for new users, who don't know that you can enable colors, by exporting a shell variable. I only found out that this is possible through reading the src code. I'm not sure may be it's properly documented now…

### Proposal:
We should check if we are running in a terminal and if so, we should enable colours. The colorblind mode shouldn't be just "bold+underline+inverse". AFAIK there are some colorschemes which work for color blind people. The colorblind and b/w mode should be enabled on terminal through some var export.   `p6doc -h` should document the different shell variables which control the color mode.

lib/Pod/To/Text.pm6
samcv: This seems to just be indenting the code… We **do** have the `.indent` function.
`$pod.contents>>.&pod2text.indent(4)`

samcv: The only case where they won't do the same thing is if there is a trailing "\n", in which case what you are doing could potentially cause unwanted indentation happening.
```perl6
("    " ~ "$var\n".subst(/\n/, "\n    ", :g)).perl
# "    here is some\n    text right here\n    rigth here\n    right there\n    "
```
Compare this to:
```perl6
"$var\n".indent(4).perl
# "    here is some\n    text right here\n    rigth here\n    right there\n"
```

lib/Pod/To/Text.pm6
```diff
@@ -33,14 +33,32 @@ sub pod2text($pod) is export {
 
 sub heading2text($pod) {
     given $pod.level {
-        when 1  {          pod2text($pod.contents)  }
-        when 2  { '  '   ~ pod2text($pod.contents)  }
+        my $content = $pod.contents.first.contents.grep( * !~~ "").first;
```
samcv: Not sure why you are using `.grep` **and** `.first`, you only need the `.first`

Using only `.first` and not `.grep.first` makes it 2x faster. And then we get more speed up doing this:
```perl6
my $content = $pod.contents.first.contents.first( *.Bool);
```
Which will catch empty strings as well as undefined strings.

lib/Pod/To/Text.pm6
```diff
@@ -33,14 +33,32 @@ sub pod2text($pod) is export {
...
+        my $text;
+        if $content ~~ Str {
+            $text = $content;
+        } else {
+            $text = pod2text $content;
+        }
+        my $name = $text.lines[0];
+        my $rest = $text.lines[1…*].join: "\n";
```
samcv: Please change `[1…*]` at least to `[1..*]`, using three dots is 3x+ slower since it creates a sequence not a range.

lizmat:
> On 9 Jun 2017, at 11:18, Samantha McVey <notifications@github.com> wrote:
> 
> @samcv requested changes on this pull request.
> 
> Made a few comments with some changes that need to be made.
> 
> In lib/Pod/To/Text.pm6:
> 
> > @@ -33,14 +33,32 @@ sub pod2text($pod) is export {
>  
>  sub heading2text($pod) {
>      given $pod.level {
> -        when 1  {          pod2text($pod.contents)  }
> -        when 2  { '  '   ~ pod2text($pod.contents)  }
> +        my $content = $pod.contents.first.contents.grep( * !~~ "").first;
> +        my $text;
> +        if $content ~~ Str {
> +            $text = $content;
> +        } else {
> +            $text = pod2text $content;
> +        }
> +        my $name = $text.lines[0];
> +        my $rest = $text.lines[1…*].join: "\n";
> 
> Please change [1…*] at least to [1..*], using three dots is 3x+ slower since it creates a sequence not a range.

Please note that:

  my ($name,$rest) = $text.split(“\n”,2);

is going to be way faster still!


Liz

zoffixznet:
> The colorblind and b/w mode should be enabled on terminal through some var export

Actually disregard that. I now see configuration options in my terminal to use different colours for each of the ANSI ones, so the colorblind solution would be to just change those colours (or disable color support).

AlexDaniel:
Does anybody know what's the status of this today?

kalkin:
@AlexDaniel I haven't touched it for month, feel free to take it, fix it and pr it.

JJ:
So I guess this would also be a candidate for picking up...

samcv:
I'm putting a "Needs Help" label on this as @kalkin seems unable to pick this up. Hopefully this will make this more visible.

JJ:
Took a look at this one, I'm afraid I'm closing it. There's no specific issue it solves, specific code changes may or may not be the best, and some other parts of it need consensus before being accepted. Parts of it can be salvaged and submitted as different PRs as #3360.

kalkin:
I'm ok with it.

JJ:
Thanks, @kalkin

PR #1041, opened 2017-03-16
Fix #130610: Triangular reduce with &infix:<,> and a list of lists sh…

FCO:
Fix #130610: Triangular reduce with &infix:<,> and a list of lists shouldn't flatten


zoffixznet:
Thanks.

PR #1042, opened 2017-03-18
Fix infinite loop #130940

FCO:


lizmat:
Looks ok, but probably want to lift this across the release?

FCO:
I think after the release is great...

src/core/Str.pm
```diff
@@ -265,6 +282,7 @@ my class Str does Stringy { # declared in BOOTSTRAP
           nqp::p6box_i($i)
         )
     }
+    multi method rindex(Str:D: Cool:D $needle, Num:D $pos) {self.rindex: $needle.Str, $pos.Int}
```
zoffixznet: Not sure where this candidate came from. It did not exist before this PR. Did you mean `Cool:D` instead?

src/core/Str.pm
```diff
@@ -178,6 +188,8 @@ my class Str does Stringy { # declared in BOOTSTRAP
         )
     }
 
+    proto method indices(|) {*}
+    multi method indices(Str:D: Cool:D $needle, *%pars) {self.indices: $needle.Str, |%pars}
     multi method indices(Str:D: Str:D $needle, :$overlap) {
```
zoffixznet: The Cool had `multi method indices(Cool:D: Str(Cool) $needle, Int(Cool) $start, :$overlap) {` candidate, but there appears to be no candidate that takes `$start` argument in Str. 

Should it be added or was the Cool one a leftover that should've been removed?

src/core/Str.pm
```diff
@@ -229,6 +243,7 @@ my class Str does Stringy { # declared in BOOTSTRAP
           nqp::p6box_i($i)
         )
     }
+    multi method index(Str:D: Cool:D $needle, Num:D $pos) {self.index: $needle.Str, $pos.Int}
```
zoffixznet: Not sure where this candidate came from. It did not exist before this PR. Did you mean `Cool:D` instead?

src/core/Str.pm
```diff
@@ -1058,8 +1076,12 @@ my class Str does Stringy { # declared in BOOTSTRAP
         )
     }
 
+    proto method subst-mutate(|) {
+        $/ := nqp::getlexdyn('$/');
+        {*}
+    }
     multi method subst-mutate(
-      Str:D $self is rw: $matcher, $replacement,
+      Str:D $self is rw: Any:D $matcher, Cool(Str:D) $replacement,
```
zoffixznet: The `Cool(Str:D)` seems to be wrong ("convert from Str to Cool"). But the type-constraint itself here is wrong, as the `$replacement` can also be a `Callable`. If our spectests don't catch that, we need to add such tests.

FCO:
Yes, that was due to RT#131014.
Just Fixed the issues.

zoffixznet:
Thanks.

PR #1044, opened 2017-03-18
Use uint64 for mask in Buf's `~^` operator

MasterDuke17:
0xFFFFFFFFFFFFFFFF == 2**64-1, which is too big a value to assign to
signed natives and will throw after some fixes to MoarVM are implemented.

Passes `make m-spectest`.


zoffixznet:
Thanks.

PR #1045, opened 2017-03-19
teach &dir to be recursive

gfldex:
With &dir being recursive pretty much all deps on Find::File could be replaced by:
    dir('.', :recursive).grep(*.ends-with: '.html');

Also, by turning IO::Path.dir into a multi, it becomes accessable to modules that want to augment it, without replacing the base method entirely.

roast commit: https://github.com/perl6/roast/commit/ddf9ae1c0331d42845d8c099fee4da02f10d3910

### DIFF
--- a/src/core/IO/Path.pm
+++ b/src/core/IO/Path.pm
@@ -388,7 +388,8 @@ my class IO::Path is Cool {
         True;
     }
 
-    method dir(IO::Path:D:
+    proto method dir(|){ * }
+    multi method dir(IO::Path:D:
         Mu :$test = $*SPEC.curupdir,
         :$absolute,
         :$Str,
@@ -448,6 +449,16 @@ my class IO::Path is Cool {
               )
             );
             nqp::closedir($dirh);
+        }
+    }
+
+    multi method dir(IO::Path:D:
+        :$recursive,
+        |rest
+    ) {
+        gather for callwith(self, |rest) {
+            (take .dir(:recursive, |rest)) if .d;
+            take $_;
         }
     }
 


ugexe:
What about Rakudo::Internals.DIR-RECURSE?

zoffixznet:
A definite 👎 from me.

As [I've mentioned yesterday](https://irclog.perlgeek.de/perl6/2017-03-18#i_14287956) file-finding is a broad area and IMO bringing functionality into core will not cover it entirely or will give core devs a lot more to care for.  Just [look at all the options Perl 5 has to offer](https://metacpan.org/search?q=File%3A%3AFind) and then read all the reviews with various concerns people raise.

For example, you're doing a depth-first search; why? Someone will then propose a `:breadth-first` argument. Then someone will then propose a `:call` `Callable` argument that can be executed to test which directories to descend into. Then someone will propose another argument to make it follow into symlinked directories. These are just the things I can think off the top of my head.

And lastly, this PR does not appear to work. As the multi that handles `:recurse` isn't the one called when `".".IO.dir: :recurse` is used.

gfldex:
Is there anything speaking against having IO::Path.dir being a multi? Then it could be moved into a module.

zoffixznet:
Made it a multi in https://github.com/rakudo/rakudo/commit/fbe7ace6fc

Though, unless you're also adding positional args, augmenting it is a bit round-about, with a subset for the invocant, to make it consider that multi version:

```
cpan@perlbuild2~/CPANPRC/rakudo (nom)$ ./perl6 -e 'use MONKEY; augment class IO::Path { subset Foo of IO::Path:D where {True}; multi method dir (Foo: :$bar!, |c) is default {say "meows" }}; say ".".IO.dir'                                                  
("dynext".IO ".travis.yml".IO "nqp".IO "README.md".IO "perl6-m".IO "perl6.moarvm".IO "perl6".IO "Configure.pl".IO "INSTALL.txt".IO "Makefile".IO "perl6-gdb-m".IO "perl6-debug.moarvm".IO "src".IO "t".IO "zef".IO "CORE.d.setting.moarvm".IO "packages".IO ".gitignore".IO "perl6-valgrind-m".IO "CREDITS".IO "config.status".IO "LICENSE".IO "docs".IO ".precomp".IO "CONTRIBUTING.md".IO ".git".IO "RESTRICTED.setting.moarvm".IO "perl6-debug-m".IO "appveyor.yml".IO "gen".IO "VERSION".IO "blib".IO "tools".IO "CORE.setting.moarvm".IO "install".IO "lib".IO)

cpan@perlbuild2~/CPANPRC/rakudo (nom)$ ./perl6 -e 'use MONKEY; augment class IO::Path { subset Foo of IO::Path:D where {True}; multi method dir (Foo: :$bar!, |c)  {say "meows" }}; say ".".IO.dir: :bar'
meows
True
```

AlexDaniel:
I wouldn't be so eager to reject the general idea of having a recursive dir in core. Maybe not with this particular PR, but still.

@zoffixznet mentioned very good reasons why we might want to delegate this to module space, and he is right. However, maybe that's the case when we can make some parallels with [left-pad](https://www.reddit.com/r/programming/comments/4bjss2/an_11_line_npm_package_called_leftpad_with_only/) issue that was going on some time ago, and also the interesting fact that we have [indent](https://docs.perl6.org/routine/indent) method in core.

If a very simple 11 line (ha!) addition is all it takes to satisfy 90% of the needs, why not have it in core? Nobody is saying that we should go all the way and cover all possible needs, that's what modules are for. But a basic recursive dir sub? I don't see any problem.

Especially considering that we [**already have it in core**](https://github.com/rakudo/rakudo/blob/d444f655fd82df4b71f29c0a04fc2626bc76be53/src/core/Rakudo/Internals.pm#L1323), just not exposed to users in meaningful way and not specced.

zoffixznet:
> takes to satisfy 90% of the needs,

How did you come up with that percentage?

AlexDaniel:
> How did you come up with that percentage?

If somebody (@gfldex++ ?) does a better analysis, sure that will be much better than my random guess. But if it is ≈90% or so, will you agree that the addition is justified?

What I can say, however, is that in the only case I needed recursive ``dir`` I had a great success with ``Rakudo::Internals.DIR-RECURSE``… I don't remember if there were any speed-related issues with modules or something, but it's implementation definitely sufficed (not that I recommend this to anybody).

zoffixznet:
> If somebody does a better analysis [...] my random guess

I was just calling you out for using obviously-made-up numbers to make a case.

The language provides all the tools for this feature and you can cover all the bases with [about 7-lines of code](https://docs.perl6.org/routine/dir) using language's primitives, instead of a set of as many arguments to a routine.

Moreover, `File::Find` comes pre-installed with the Rakudo's distribution. And based on its revdeps being just 2% of ecosystem modules, this doesn't sound like a much-demanded case that ought to be included in core language to add to core developers' maintenance burden.

I'd like to thank @gfldex for the work, but I'm going to reject this PR. If some find this decision unsatisfactory, then I invite them to open an RFC ticket and solicit some approval for the idea first.

zoffixznet:
P.S.: the implementation in this PR also has a file exhaustion bug. It'll crash when running over a tree with more than ~1024-2000 dirs... whatever the system's max open files limit is.

PR #1046, opened 2017-03-21
Fix triangle reduction with listinfix ops

FCO:
The triangle reduction was appending the data inside of a array and applying the
op on that. It was changed to push the data and apply the op on that array with
each item deconted. RT #131009


lizmat:
isn't the decont a different issue from the flattening?  If so, maybe these should be two different PR's ?

FCO:
I think on both as the same issue, but if it isn't, I could make another PR (but each PR will not be working...)

PR #1047, opened 2017-03-26
Fix SEGV in FOREIGN_LANG after nqp fixes in nqp PR#349

skids:


skids:
Along with nqp PR#349 this fixes some issues found while getting Grammar::BNF
back into running condition on current rakudo/nqp.

zoffixznet:
Thanks.

PR #1049, opened 2017-03-26
Remove $.pid from Proc

AlexDaniel:
Not that we should not have it, but as of today it does nothing at
all (making users assume that there is something wrong with their
code).

See https://irclog.perlgeek.de/perl6/2017-03-25#i_14323473

Spectest clean.


zoffixznet:
Thanks.

PR #1050, opened 2017-03-29
Fix for RT #131075

dogbert17:
It was noted by Zoffix++ that the number of tests run by harness6
differed from the number run by harness5 when spec- or stresstesting.
The problem turned out to be the piece of code combining the normal
test files in spectest.data with the ones marked as being 'slow'.
That code used the Z operator on lists that were not the same size,
changing to roundrobin instead solves the problem. Zoffix++ for help.


zoffixznet:
Thanks.

PR #1054, opened 2017-03-31
Make `my @a = ^10; my @b = @a[0..*]` 50-80x faster

MasterDuke17:
By just reusing the zen-slice (`my @a = ^10; my @b = @a[]`)
implementation when we know we have a `Range` with a `.min == 0` and
a `.max == Inf`.

Passes `make m-spectest`.

### DIFF
--- a/src/core/array_slice.pm
+++ b/src/core/array_slice.pm
@@ -211,7 +211,9 @@ multi sub postcircumfix:<[ ]>( \SELF, Any:D \pos, :$v!, *%other ) is raw {
 multi sub postcircumfix:<[ ]>( \SELF, Iterable:D \pos ) is raw {
     nqp::iscont(pos)
       ?? SELF.AT-POS(pos.Int)
-      !! POSITIONS(SELF, pos).map({ SELF[$_] }).eager.list;
+      !! nqp::istype(pos, Range) && pos.min == 0 && pos.max == Inf
+        ?? SELF.ZEN-POS()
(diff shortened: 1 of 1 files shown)


AlexDaniel:
Does it make normal slices slower? I mean, of course it does, because you have 3 extra checks there. But by how much?

MasterDuke17:
I get ~4-6% slower.

zoffixznet:
> (by normal slices I mean something like [0..3])

> I get ~4-6% slower.

Wouldn't those be waaaaay more common, making this optimization detrimental overall? I use ranges all the time in slices, but I can't think of a single time where I've used a `^*` range; I'd just use a zen slice or the array as is, without any slicing.

Also, the current impl has a bug, as it doesn't check whether the 0 `.min` is `.excludes-min`:

```
<Zoffix__> m: say (0^..*).min; say (^10).[0^..*]
<camelia> rakudo-moar 833c29: OUTPUT: «0␤(1 2 3 4 5 6 7 8 9)␤»
```

AlexDaniel:
Yes, 50x faster for some weird case (I'm pretty sure I've never used ``0..*``) does not justify a 4% slowdown.

zoffixznet:
Thanks for the work, however, I think this would be a net-negative to pessimise common case by 4-6% to give a very rare case an improvement.

As well as the change to reificiation is an issue. I recall on IRC it was confirmed Zen slice isn't meant to reify.

PR #1057, opened 2017-04-08
[Test] Add default description.

jsimonet:
Some methods already have default description (can-ok for example).

Add description for like, unlike, use-ok methods.
Rename some parameters for a more standardised naming (desc vs msg) for
isa-ok, does-ok, can-ok and use-ok.


zoffixznet:
Thanks.

PR #1058, opened 2017-04-09
Fix Bags to allow values larger than a native int

MasterDuke17:
I believe this broke in e7e97c7b.

Passes `make m-spectest`.

Before the fix:
`perl6-m -e 'my $bag = {"red" => 200000000000000000019, "blue" => 100000000000000000000}.Bag; dd $bag'`
gave
`Cannot unbox 68 bit wide bigint into native integer`
after the fix:
`Bag $bag = ("red"=>200000000000000000019,"blue"=>100000000000000000000).Bag`

### DIFF
--- a/src/core/Baggy.pm
+++ b/src/core/Baggy.pm
@@ -41,14 +41,14 @@ my role Baggy does QuantHash {
           nqp::while(
             $iter,
             nqp::if(
-              nqp::isle_i(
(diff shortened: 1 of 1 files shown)


zoffixznet:
Thanks.

AlexDaniel:
@MasterDuke17 tests? :)

PR #1061, opened 2017-04-13
Make errors in BOOTSTRAP more awesome

MasterDuke17:
Include relevant information where possible.

Passes `make m-spectest` except for two tests in t/spec/S06-signature/optional.t, which expect the exact text of one of the changed error messages, but have as their description "making an "is rw" parameter optional dies with adequate error message". I would argue that my changes to the error message are improvements, so the tests should be altered to pass.

### DIFF
--- a/src/Perl6/Metamodel/BOOTSTRAP.nqp
+++ b/src/Perl6/Metamodel/BOOTSTRAP.nqp
@@ -1529,19 +1529,19 @@ BEGIN {
             my $SIG_ELEM_IS_RW       := 256;
             my $SIG_ELEM_IS_OPTIONAL := 2048;
             my $dcself := nqp::decont($self);
-            my int $flags := nqp::getattr_i($dcself, Parameter, '$!flags');
-            if $flags +& $SIG_ELEM_IS_OPTIONAL {
-                nqp::die("Cannot use 'is rw' on an optional parameter");
-            }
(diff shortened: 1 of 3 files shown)


zoffixznet:
The failing tests are also part of 6.c-errata.

I'm 👍 on this PR and don't think exact error text should be tested by roast, so I'd like to keep the tests, but remove the `message =>` part, so text itself isn't tested, and move copy the tests to https://github.com/rakudo/rakudo/blob/nom/t/05-messages/01-errors.t and tests the actual text messages  there.

MasterDuke17:
Want me to add a commit copying the tests as part of this PR?

zoffixznet:
> Want me to add a commit copying the tests as part of this PR?

Sounds good. Also, there's a way to throw typed exceptions from nqp/VM, maybe that's even better:
https://irclog.perlgeek.de/perl6-dev/2017-04-13#i_14425191

MasterDuke17:
@zoffixznet @timo there doesn't appear to be any existing exception that is a great match. There's `X::Parameter::RW`, but its message is about expecting a writable container and getting something else instead. Want me to create one?

zoffixznet:
> Want me to create one?

Yes, if possible.

MasterDuke17:
New typed exception added and thrown.

zoffixznet:
Thanks

PR #1063, opened 2017-04-19
RFC Allow the decode of utf* buffers as other encodings

samcv:
If a :force option is used, let the person change the encoding to
another type.

This will allow a person for example to decode a utf8 buffer as
iso-8859-1

### DIFF
--- a/src/core/Buf.pm
+++ b/src/core/Buf.pm
@@ -407,11 +407,11 @@ constant blob32 = Blob[uint32];
 constant blob64 = Blob[uint64];
 
 my class utf8 does Blob[uint8] is repr('VMArray') {
-    method decode(utf8:D: $encoding = 'utf-8') {
+    method decode(utf8:D: $encoding = 'utf-8', Bool :$force) {
         my $enc = Rakudo::Internals.NORMALIZE_ENCODING($encoding);
-        die "Can not decode a utf-8 buffer as if it were $encoding"
-            unless $enc eq 'utf8';
-        nqp::p6box_s(nqp::decode(self, 'utf8'))
+        die "Can not decode a utf-8 buffer as if it were $encoding without :force"
+            if !$force and $enc ne 'utf8';
+        nqp::p6box_s(nqp::decode(self, $encoding))
     }
     method encoding() { 'utf-8' }
     multi method Str(utf8:D:) { self.decode }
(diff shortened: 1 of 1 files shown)


zoffixznet:
How come is it blocked in the first place? It's a bunch of bytes; let the user do whatever they want with them.

If blocking is desired, perhaps adding a .Blob/.Buf coercers to `.utf*` is better than an option? Feels like something a user might try first when getting the error.

samcv:
I agree with what you said @zoffixznet, though I was making an RFC remaining closer to the original in case there was some other reason for wanting to give the user an error on this.

I would be perfectly fine removing any error as well.

jnthn:
The `utf8` type is specified in S02. If we know the data was encoded as `utf8` then, well, we know it's not just any old bunch of bytes. :-) There is already an easy way to get around this: `Buf.new("omg".encode("utf-8")).decode("latin-1")`.

If we're going to add something, I think I'd rather we do it on the encoding end, allowing an alternative type to be specified for the result of the encoding. This in turn would allow the user to request a mutable `Buf` out of any encoding too, And the case here would just become `.encode('utf-8', :blob(Buf))` or so (the name of the argument needs some more thought).

I'm -1 to having a `:force` to do this since there's way a lot too many things that it could mean in the context of decoding, and - for something you'd rarely want to do - it doesn't pull its weight.

smls:
My use-case was dealing with filenames with broken encoding.

Through some screw-ups in the prior toolchain (not Perl 6 related), they ended up on a Linux partition but still encoded in Windows Codepage iso-8859-1 encoding rather than UTF8.

The Perl 6 script that fetched the directory listing naturally interpreted them as UTF8, and stored them in  into `Str`s. It just so happened that the filenames had a valid UTF8 interpretation... just not the string interpretation they were *intended* as... :)

I tried `$filename.encode("utf8").decode("iso-8859-1")` to see what the filenames were *supposed* to be (to help me debug the problem), and was surprised by the `Can not decode a utf-8 buffer as if it were iso-8859-1` error because my understanding was that `.decode` and `.encode` convert freely back and forth between `Buf` and `Str`.

So yeah, its a pretty niche use-case, and since there's a work-around...

    Blob.new( $str.encode("utf8") ).decode("iso-8859-1")

...a `:force` option might not be needed.

But I think at least the error message should be improved. nstead of saying `can not decode`, it should make clear that it is a safety restriction and not a technical limitation, and explain to the user what their being protected from, and how they can work around it if need be.

AlexDaniel:
Hey @samcv, any new thoughts on this?

samcv:
I think we need to have the error tell the user the workaround. That they must turn it into a Buf or Blob. The use case is pretty common, but I'm not convinced at this time we need to do anything more than update the error message.

JJ:
Besides the conflicts, there does not seem to be a clear consensus here. So can we please fix the conflicts and then submit whatever new feature to the problem-solving repo?

lizmat:
Do we have a golf of this issue?  I can't seem to generate the error?

lizmat:
Feels to me this PR has bitrotted, and should therefore probably be closed.

samcv:
@lizmat you are right, the functionality has changed and we can now decode a buffer as whatever we want without error. I am going to close this PR now.

PR #1070, opened 2017-04-24
Add test for RT #118263 fix

MasterDuke17:
The message when trying to pun a role with required methods should have
the names of the child, parent, required methods, and suggest "does".


lizmat:
Perhaps it would be better to test for elements of the message, rather than the exact message?

MasterDuke17:
@lizmat test updated.

PR #1072, opened 2017-04-26
Fix typed optional Array and Hash parameters

MasterDuke17:
When handling typed optional Array and Hash parameters that don't
have a value passed we now nqp::create() the correct empty
parameterized type.

Before, `sub foo(Int @a?) { @a.push(42); dd @a }; foo();` would die with `Type check failed in binding to @a; expected Positional[Int] but got Array ($[])`, but now it prints `Array[Int].new(42)`

Fixes RT #118555

Passes `make m-spectest`.

### DIFF
--- a/src/Perl6/Actions.nqp
+++ b/src/Perl6/Actions.nqp
@@ -8433,15 +8433,31 @@ class Perl6::Actions is HLL::Actions does STDActions {
                 }
                 else {
                     if %info<sigil> eq '@' {
+                        my $type;
+                        if nqp::istype($nomtype, $*W.find_symbol(['Positional'])) && nqp::can($nomtype.HOW, 'role_arguments') {
+                            my @role_args := $nomtype.HOW.role_arguments($nomtype);
+                            $type := $*W.parameterize_type_with_args($/, $*W.find_symbol(['Array']), @role_args, nqp::hash);
+                        }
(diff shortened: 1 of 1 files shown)


zoffixznet:
Are `Positional`/`Associative` and `Array`/`Hash` the only things that differ between the two branches? Feels like half the code could be de-duplicated by stuffing those into variables.

Something like:

```perl6
 if (my $is-array := %info<sigil> eq '@') || %info<sigil> eq '%' {
    my $role      := $is-array ?? 'Positional' !! 'Associative';
    my $base-type := $is-array ?? 'Array'      !! 'Hash';

    $var.default: QAST::Op.new: :op<create>, QAST::WVal.new: value =>
        nqp::istype($nomtype, $*W.find_symbol([$role]))
            && nqp::can($nomtype.HOW, 'role_arguments')
        ?? $*W.parameterize_type_with_args(
            $/, $*W.find_symbol([$base-type]),
            $nomtype.HOW.role_arguments($nomtype),
            nqp::hash)
        !! $*W.find_symbol([$base-type])
}
…
```

MasterDuke17:
@zoffixznet re-factored, but with the paren style used in the rest of the file.

zoffixznet:
Thanks.

PR #1075, opened 2017-05-03
Fix rotate on empty list

MasterDuke17:
`my @a = (); @a.rotate(1)` used to error with `Modulation by zero`
before, but now does nothing.

Passes `make m-spectest`.


zoffixznet:
Thanks.

PR #1076, opened 2017-05-06
Fix stuck precomp on osx

ugexe:


ugexe:
I can't answer that question, as I don't know why it works. I've just done a similar pattern more than once over the years having encountered osx + proc issues before.

In this case `$proc.out.lines.unique;` is where it gets stuck, but reading and/or closing `:err` first seems to correct the behavior.

ugexe:
I golfed this down, and it affects more than just osx:
```
# WORKS
perl6 -e 'my $proc = shell q|perl6 -e "die q!oo! xx 1000000;"|, :out; my @out = $proc.out.lines.unique;'

# WORKS
perl6 -e 'my $proc = shell q|perl6 -e "die q!oo! xx 100;"|, :out, :err; my @out = $proc.out.lines.unique;'

# STUCK
perl6 -e 'my $proc = shell q|perl6 -e "die q!oo! xx 1000000;"|, :out, :err; my @out = $proc.out.lines.unique;'
```

LLFourn:
fwiw I RT'd the above behaviour on OSX a while ago:

https://rt.perl.org/Public/Bug/Display.html?id=127682

PR #1077, opened 2017-05-06
Make feed compile errors better

MasterDuke17:
Fixes https://rt.perl.org/Ticket/Display.html?id=112396.

Some discussion about the new error messages here: https://irclog.perlgeek.de/perl6-dev/2017-05-05#i_14539593

Passes `make m-spectest`.


zoffixznet:
Thanks.

PR #1078, opened 2017-05-07
Convert substr+eq to eqat

MasterDuke17:
nqp::eqat is much faster than doing a substr+eq.

Passes `make m-spectest`.


zoffixznet:
Thanks.

PR #1080, opened 2017-05-10
Test instantiating non-class has its name in error

MasterDuke17:
For RT #122232


zoffixznet:
Thanks.

zoffixznet:
> It might make sense to use the name longer than one character for the first test.

Good point. Fixed in https://github.com/rakudo/rakudo/commit/65c0786d3d

PR #1082, opened 2017-05-14
Throw X::NYI when trying to subset a native type

MasterDuke17:
Native type subsets can be created currently, but don't work, always
saying the type check failed (see RT #131310).

Passes `make m-spectest`.


zoffixznet:
Thanks.

PR #1083, opened 2017-05-15
Add ready Promise to Proc::Async

the-eater:
Adds the ready promise to Proc::Async, which keep's when the Proc is started, which may be useful in cases you don't control the start of the Proc but want to know when it starts

### DIFF
--- a/src/core/Proc/Async.pm
+++ b/src/core/Proc/Async.pm
@@ -42,6 +42,8 @@ my class Proc::Async {
     my class ProcessCancellation is repr('AsyncTask') { }
     my enum  CharsOrBytes ( :Bytes(0), :Chars(1) );
 
+    has $!ready_promise = Promise.new;
+    has $!ready_vow = $!ready_promise.vow;
     has $.path;
     has @.args;
     has $.w;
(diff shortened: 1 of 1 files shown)


lizmat:
Looks like a nice feature to me.

moritz:
I like the idea, thanks!

I think the implementation can be improved. The code should take a vow on that promise (and keep it in a private attribute or so), so that other code cannot accidentally fulfill or break the promise, much like most code in the setting does that returns promises.

the-eater:
@moritz is this what you mean?

src/core/Proc/Async.pm
```diff
@@ -95,6 +97,10 @@ my class Proc::Async {
             $enc, :$translate-nl
     }
 
+    method ready(--> Promise) {
+      $!ready_promise;
```
jnthn: Should use 4-space indent here.

the-eater:
Since "ready" is also the callback name used by MoarVM I thought it was a fitting name :)
I've already created a branch with tests https://github.com/the-eater/roast/tree/feature-proc-async-ready

Only docs still need to be done

src/core/Proc/Async.pm
the-eater: Fixed :) thanks!

moritz:
Thanks @the-eater!

PR #1086, opened 2017-05-22
Update localport if we made bind on port 0

cono:
This is a dependency for: https://github.com/MoarVM/MoarVM/pull/600

There is a common practice to bind on first un-used port, you have to
bind on port 0 and after that with help of getsockname call you can get
what port is in use.


zoffixznet:
Thanks!

PR #1088, opened 2017-05-26
set default thread pool size based on hardware

robertlemmen:


robertlemmen:
related: https://github.com/MoarVM/MoarVM/pull/604, https://github.com/perl6/nqp/pull/362

jnthn:
While this may produce improved results in theoretical tests, we'll cause massive breakage to real-world code if we merge this. I know I've got a load of production code that relies on using more thread pool threads that the system has CPU cores, and thanks to 6.c's `await` blocking then this is an extremely common situation to end up in.

I can imagine using this op max'd with 16 (so if you have a 32-core box then we'll be able to use all 32 of them, but we bottom out at a maximum of 16 to make sure we don't break existing code).

What we really need is a much smarter scheduler that actually monitors progress and can add threads based on queue lengths and throughput. Then we could use the number of CPU cores as our ceiling for threads that seem to be making progress through tasks, but if there's no progress add extras to try to break the "deadlock". That way, many programs can get by on just a handful of threads, but we can cope with those that end up blocking a bunch of the pool too.

robertlemmen:
low-capped to 16 now.

out of curiosity, how do you suggest the smart scheduler should make it's decision? purely based on the existence of a waiting task and the fact that no task has finished in a defined time? or something more intrusive that looks at the waits-for graph?

zhuomingliang:
How about merge this one? https://github.com/MoarVM/MoarVM/pull/604 got merged.

niner:
nqp::cpucores is currently only supported by MoarVM

zhuomingliang:
niner: and JVM too, https://github.com/perl6/nqp/pull/362/files

jnthn:
Not this close to the release. Maybe after but note that as it currently stands the scheduler generally starts all the threads it possibly can, so people doing testing on 64-core boxes or some such might find it ends up trying to make 64 threads per Perl 6 process that does a lot of async stuff.

zhuomingliang:
I guess now it's safe to merge.

MasterDuke17:
@jnthn is this still applicable with the recent scheduler changes?

jnthn:
The new ThreadPoolScheduler now uses the `nqp::cpucores` op as a soft-maximum (possible to go over to try and beat deadlock and to have at least one thread for each queue type, but it only reluctantly goes past it). So, it does what this PR wanted to do but without it being awkward to merge because of how it would interact with the previous scheduler's behavior. Thanks, @robertlemmen, for doing the op for me to use in the new scheduler.

PR #1090, opened 2017-05-29
Error for wrong arity in signature has right value

MasterDuke17:
Tests for https://rt.perl.org/Ticket/Display.html?id=123085


zoffixznet:
Thanks!

PR #1091, opened 2017-05-30
Suggest misspelled method names

MasterDuke17:
Like types, lexicals, and routines/phasers, suggest methods with a
similar name if the given method is not found for the invocant.

Fixes https://rt.perl.org/Ticket/Display.html?id=123078 and
https://rt.perl.org/Ticket/Display.html?id=127395.

Passes `make m-spectest`.

### DIFF
--- a/src/core/Exception.pm
+++ b/src/core/Exception.pm
@@ -176,17 +176,21 @@ my class X::Method::NotFound is Exception {
             %suggestions<encode($encoding).bytes> = 0;
         }
 
-        for $.invocant.^methods(:all)>>.name -> $method_name {
-            my $dist = StrDistance.new(:before($.method), :after($method_name));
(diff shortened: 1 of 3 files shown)


MasterDuke17:
@jnthn how does the most recent commit look?

lizmat:
@MasterDuke17: I get a hang now on something like 'my ::foo $x, say $x'  Could you confirm?

MasterDuke17:
Odd behavior. Not only did I have to add the `nqp::can` check for `methods` and `private_method_table`, but i had to change to `$!invocant`, using `$.invocant` infini-looped.

MasterDuke17:
Rebased and merge conflicts fixed.

zoffixznet:
Thanks.

PR #1092, opened 2017-05-30
add fix for RT #131400

tbrowder:


lib/Pod/To/Text.pm6
```diff
@@ -20,7 +20,25 @@ sub pod2text($pod) is export {
         when Pod::Block::Declarator { declarator2text($pod)     }
         when Pod::Item         { item2text($pod).indent(2)      }
         when Pod::FormattingCode { formatting2text($pod)        }
-        when Positional        { $pod.map({pod2text($_)}).join("\n\n")}
+        when Positional        {
+            =begin comment
+	    # original code
+            $pod.map({pod2text($_)}).join("\n\n")
+            =end comment
```
zoffixznet: We already have git history if ever need previous code. We don't need to keep it in source.

lib/Pod/To/Text.pm6
```diff
@@ -20,7 +20,25 @@ sub pod2text($pod) is export {
...
+
+	    # unwrapping original code for debugging...
+	    # fix based on irc #perl6, 2017-05-30, word descrip from
+	    # kurahaupo at 23:43 and on 2017-05-30 MasterDuke's hint
+	    # at 00:24
```
zoffixznet: Same with this; you can reference this in commit message rather than source.

tbrowder:
On Tue, May 30, 2017 at 11:02 Zoffix Znet <notifications@github.com> wrote:

> *@zoffixznet* commented on this pull request.
>
...

I can add that info for both chunks, but will be a bit later today.

zoffixznet:
Thanks.

tbrowder:
On Tue, May 30, 2017 at 1:01 PM, Zoffix Znet <notifications@github.com> wrote:
> Thanks.

You're welcome!

PR #1094, opened 2017-06-01
Fix for RT #131201 (error message when proto has no implementation)

jsimonet:
I tried to write some code to fix RT #131201.

Basically, it's a check to see if the array is empty or not, but i don't know how to inform user that a prototype method exists and may be implemented.

### DIFF
--- a/src/core/Exception.pm
+++ b/src/core/Exception.pm
@@ -2447,7 +2447,7 @@ my class X::Multi::NoMatch is Exception {
         else {
             @bits.push('...');
         }
-        if @cand[0] ~~ /': '/ {
+        if @cand && @cand[0] ~~ /': '/ {
             my $invocant = @bits.shift;
             my $first = @bits ?? @bits.shift !! '';
             @bits.unshift($invocant ~ ': ' ~ $first);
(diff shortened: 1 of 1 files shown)


src/core/Exception.pm
```diff
@@ -2462,9 +2462,10 @@ my class X::Multi::NoMatch is Exception {
                         ~  join("\n    ", '', sort keys @cand ∖ @un-rw-cand)
                     unless @cand == @un-rw-cand
                 )
-            !! join "\n    ",
-                'none of these signatures match:',
-                @cand
+            !! ( @cand
+                 ??  join "\n    ", 'none of these signatures match:', @cand
+                 !! "No signature match."
```
zoffixznet: The `No signature match` isn't very clear. Perhaps, it should read `Routine does not have any candidates. Is only the proto defined?`

jsimonet: Yes, it's a better message :) I'll commit this one.

zoffixznet:
Thanks!

zoffixznet:
For record, you fixed https://rt.perl.org/Ticket/Display.html?id=131367#ticket-history too

jsimonet:
Perfect :) I'm happy to have contributed to Perl6!

PR #1095, opened 2017-06-08
A worry for RT #126732

AlexDaniel:
I blinked my eyes and the ticket got closed. In any case, here is a
fix.

For those who think that this might cause some kind of breakage, this
is just a worry, so don't worry.

This whole discussion was basically about adding or not adding this
one line of code. I don't see any problem with it. Curious users will
try to use superscripts alone, maybe just for fun, but a “potential
difficulty” will give them a lesser WTF-ish feeling. So let's just
take it.

Example:

Code:
```
say ³²
```

Result:
```
Potential difficulties:
    ³ is parsed as a separate numeric literal. Did you forget a term?
    at -e:1
    ------> say ³⏏²
9
```

### DIFF
--- a/src/Perl6/Grammar.nqp
+++ b/src/Perl6/Grammar.nqp
@@ -3506,6 +3506,7 @@ grammar Perl6::Grammar is HLL::Grammar does STD {
         | 'Inf' >>
         | $<uinf>='∞'
         | <?{ nqp::existskey(nqp::backendconfig(), 'moarlib') }> <unum=:No+:Nl>
+          [ <.after <[⁰¹²³⁴⁵⁶⁷⁸⁹]>> <.before <[⁰¹²³⁴⁵⁶⁷⁸⁹]>> {$/.worry: "$<unum> is parsed as a separate numeric literal. Did you forget a term?"} ]?
         ]
     }
 


zoffixznet:
No, the whole discussion on the *now-rejected* ticket was about not adding useless special casing to well-defined rules. We don't need to follow up with "...except when..." any time we discuss Perl 6 numeric literals.

You're adding a warning that you can't even turn off. You're half-solving a problem that doesn't exist.

You're trying to slip through a PR, hoping some core dev who wasn't closely following the discussion merges it.

AlexDaniel:
No, I was hoping that instead of starting another holy war we can just add one line, solve the problem and move on.

``㆒²`` is not a problem, and the two tickets were not talking about ``㆒²`` (originally, at least). On the other hand, you can easily get ``²²`` if a cat walks over your keyboard. Or if you try it yourself. If you write ``㆒²``, you are actively asking for trouble, while ``²²`` is just a superscript that stands on its own (well, that's how it looks like). Of course you assume that every perl 6 user knows all of its parsing rules, so he/she should not be surprised, but again we gain nothing from this.

> You're adding a warning that you can't even turn off

Yet another derailing. First of all, show me somebody who insists on writing ``²²`` and making it not complain. Then, the fact that you can't turn off worries is part of another problem, so perhaps create a ticket for that?

coke:
-1 from me on merging this

zoffixznet:
> we can just add one line, solve the problem and move on.

Except it's not "just". You want to add one line of code that will help no one and will make thousands of people who talk or write about Perl 6 numerals in the next decade to pointlessly waste breath explaining this special case.

> show me somebody who insists on writing ²² and making it not complain. 

You're looking at them. I insist on writing any `No` char in a term position and have it be into a numeric literal. I don't want to "write any No char, except when I'm also using superscript powers, but I can still can use superscripts as a term otherwise, because it's just a No char, and I can still use `**` power op with them"

> I was hoping that instead of starting another holy war

There isn't any holy war. I'm just positing to you that *"if a cat walks over your keyboard"* is not a sound language design principle.

AlexDaniel:
> is not a sound language design principle

But it is. People make typos, people type stuff from books or slides, people get lost in vim, etc. In the end you get something that you wouldn't ever write consciously. Well, unless your whole goal is to reject some ticket, of course. And now your point is that we should let it produce a completely nonsensical value, just for the purity of the whole thing. In theory, you're right. Practically your point does not get us anywhere.

zoffixznet:
> your whole goal is to reject some ticket

You [keep repeating](https://irclog.perlgeek.de/perl6-dev/2017-04-07#i_14391258) that line without providing any evidence to back it up.

> now your point is that we should let it produce a completely nonsensical value

That's a nice looking strawman you got there, but the value produced is quite the opposite of that.

Per [nine's earlier comments in #perl6-dev](https://irclog.perlgeek.de/perl6-dev/2017-06-08#i_14703885) and Coke's downvote above, rejecting this PR, for the same reasons the ticket was rejected yesterday.

AlexDaniel:
Maybe we should at least create a FAQ entry for this thing (yes, instead of trying to reject it by all means… so far I've seen no problem solving regarding this issue from you, and only rejectionism). It is brought up on the channel recurrently.

zoffixznet:
You created a PR for a ticket you knew was rejected. What did you think was gonna happen?

Your tactic of bullying people into getting your way by reopening rejected Issues and pretending the concerns that led to their rejection don't exist will only get you that far.

lizmat:
> On 9 Jun 2017, at 13:53, Zoffix Znet <notifications@github.com> wrote:
> 
> You created a PR for a ticket you knew was rejected. What did you think was gonna happen?
> 
> Your tactic of bullying people into getting your way by reopening rejected Issues and pretending the concerns that led to their rejection don't exist will only get you that far.

What’s wrong with creating a FAQ entry?



Liz

zoffixznet:
> What’s wrong with creating a FAQ entry?

I never made any comments about any FAQ entries.

AlexDaniel:
I was *expecting* that at some point we will resolve the issue one way or another. The fix was created and the spectest was passed on my way home, the ticket wasn't even rejected at that moment I think. I don't mind the status of the pull request corresponding with the status of the ticket, but I decided to create a PR anyway so that we have some actual code instead of just talk.

Not sure what you're accusing me of though.

“Hello, here's an actual one-line fix for the ticket that is **already closed**”
“you're trying to slip through a PR”

“Maybe we should at least create a FAQ entry for this thing”
“Your tactic of bullying people …”

The issue itself is not that big of a trouble. But the way we handle rejections is. That is:
1. The ticket was closed without any tests.
1. The documentation does not cover this. At least a FAQ entry would do. Yes, [unicode_texas](https://docs.perl6.org/language/unicode_texas) mentions “Numeric values”, but it is too far of a stretch to say that this particular case is explained.
1. The code wasn't changed to produce a helpful error message.
1. **No ticket in any repo was created to handle things I mentioned above.**
1. **No justification was given on why everything above was omitted.**

Bullying? Well, indeed, maybe we should roast people for closing tickets without tests.

Perhaps that can be a checklist for rejections? Mark at least one and feel free to reject. Otherwise, do something and then reject.

.oO( ``2**³²`` anyone? 🤒 )

zoffixznet:
> I was expecting that at some point we will resolve the issue one way or another.

You mean [after I told you I was going to close it](https://irclog.perlgeek.de/perl6/2017-03-06#i_14215347) **THREE MONTHS AGO** or after [you whined about the ongoing discussion and hoped people would "shut up" about it](https://irclog.perlgeek.de/perl6-dev/2017-06-07#i_14701964) you're surprised the resolution was a rejection?

> “Maybe we should at least create a FAQ entry for this thing”
> “Your tactic of bullying people …”

You're now deliberately slandering me by placing a different sentence next to what I replied to and considering this bullshit discussion is going on for third day in a row, you're now just deliberately harassing me because you got nothing better to do.

PR #1096, opened 2017-06-10
Suggest methods when missing method in optimizer

MasterDuke17:
Pass the right thing as the invocant so it can be introspected for
possible suggestions.

Also add a '!' to the beginning of the name of private methods in the
message saying it doesn't exist to be consistent with names in
suggestions.

Should finish up resolving
https://rt.perl.org/Ticket/Display.html?id=123078

Passes `make m-test` and `make m-spectest`.


zoffixznet:
Thanks.

PR #1098, opened 2017-06-11
Say hash name when assigning value of wrong type

MasterDuke17:
Before, `perl6 -e 'my Int %h{Any}; %h<a> = "b"'` would give the error
`Type check failed in binding to assignval; expected Int but got Str ("b")`,
but now it gives `Type check failed in assignment to %h; expected Int but
got Str ("b")`

Addresses at least some of https://rt.perl.org/Ticket/Display.html?id=118023

Passes `make m-test` and `make m-spectest`.

### DIFF
--- a/src/core/Hash.pm
+++ b/src/core/Hash.pm
@@ -579,7 +579,7 @@ my class Hash { # declared in BOOTSTRAP
             )
         }
 
-        method ASSIGN-KEY(::?CLASS:D: TKey \key, TValue \assignval) is raw {
+        method ASSIGN-KEY(::?CLASS:D: TKey \key, Mu \assignval) is raw {
             nqp::if(
               nqp::getattr(self,Map,'$!storage').DEFINITE,
               nqp::if(


zoffixznet:
What's doing the type check now, with the .ASSIGN-KEY type check removed?

Also, this feels like solving 10% of the problem. The error is still same as old one with binding, `:exists`, `:delete`, and when key is of the wrong type in assignment.

MasterDuke17:
Well, now an X::TypeCheck::Assignment is thrown instead of an X::TypeCheck::Binding::Parameter. It says the same line in Hash.pm (which is just the start of ASSIGN-KEY since the whole thing is NQP).

Yeah, I haven't been able to get it to complain the same way with the key. Figured I'd at least do the value while I'm working on the key (and this PR could also be considered fishing for suggestions).

MasterDuke17:
I still haven't gotten any of the other cases to work. Calling @jnthn @lizmat @timo for suggestions.

MasterDuke17:
Closing this after some discussion with @jnthn, there's a more general solution possible.

[18:27] <MasterDuke> jnthn: btw, do you have any suggestions for https://github.com/rakudo/rakudo/pull/1098 ?
[18:30] <jnthn> Well, it's just one of the classic things that happens because we're implementing Perl 6 in Perl 6 :)
[18:32] <jnthn> Maybe a non-awful general solution would be to add some kind of parameter trait
[18:32] <jnthn> will bind-error { ... }
[18:33] <jnthn> And then X::TypeCheck::Binding::Parameter can invoke it to get a better message
[18:33] <MasterDuke> to Hash's key and value?
[18:33] <jnthn> Yeah
[18:33] <jnthn> Note that it's a trait though, not a runtime closure
[18:34] <jnthn> So we'll have to pass in something to the block (perhaps the exception object itself)
[18:34] <jnthn> That'd give us a general solution to this kind of problem

PR #1101, opened 2017-06-15
Fix ModuleLoader.nqp so RAKUDO_MODULE_DEBUG works the same as elsewhere

samcv:
This code emulates the whitespace ignoring code written in Perl 6
in which beginning and trailing whitespace in %*ENV<RAKUDO_MODULE_DEBUG>
are ignored.
Here, the str number is converted to an int. Examples:
    'yes' => 1, '' => 0, null => 0, ' 9 ' => 9, 'no' => 1, 'camelia' => 1 etc.

Additionally it also fixes a bug that would cause Rakudo to die during compilation
if RAKUDO_MODULE_DEBUG were set to anything other than a number. This was
discovered by the Rakudo AppImage builder here:
https://travis-ci.org/samcv/rakudo-appimage-module-test-automation/jobs/242732276#L1204

When given a non-number during normal operation the debugging code in
ModuleLoader.nqp would not be activated, even though other debugging
code in Rakudo codebase would. The original code was added a long time ago,
on 2013-04-21.

This fix only applies to MoarVM, and this bug could still exist on the JVM.
This commit harmonizes the ModuleLoader.nqp code with the Perl 6 based code.

### DIFF
--- a/src/Perl6/ModuleLoader.nqp
+++ b/src/Perl6/ModuleLoader.nqp
@@ -1,4 +1,8 @@
-my $DEBUG := +nqp::ifnull(nqp::atkey(nqp::getenvhash(), 'RAKUDO_MODULE_DEBUG'), 0);
+# $DEBUG is set to 1 for Truey numeric values of ENV 'RAKUDO_MODULE_DEBUG'
+# or for non-numeric strings. All other cases $DEBUG is set to 0
+my $rakudo-module-debug := nqp::atkey(nqp::getenvhash(), 'RAKUDO_MODULE_DEBUG');
+my $DEBUG := nqp::stmts((my $debug-radix := nqp::radix(10, $rakudo-module-debug, 0, 0)),($debug-radix[2] != -1))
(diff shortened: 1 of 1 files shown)


niner:
That's a lot of code for handling edge cases of setting a debug flag. Certainly more code than the actual debugging, as $DEBUG is only used in 3 places in that file.

I could live very well without the debug output from ModuleLoader.nqp. I'm note sure anyone else has a use for it. So I vote for removing rather than extending it.

samcv:
I had another version that was only a few lines, that at least wouldn't die and would also work in more situations than previously even though it doesn't handle whitespace. That may be better to use. Will revise my commit.

zoffixznet:
It's a flag for debugging. IMO it doesn't need to emulate anything and should just check for truthiness.

samcv:
Now:

$DEBUG is set to 1 for Truey numeric values of ENV 'RAKUDO_MODULE_DEBUG'
or for non-numeric strings. All other cases $DEBUG is set to 0

This is much simpler and doesn't handle the corner cases, while acting properly in all of the common cases.

PR #1103, opened 2017-06-17
Fix t/spec/S02-magicals/KERNEL.t on OpenBSD

stmuk:


src/core/Kernel.pm
```diff
@@ -115,7 +115,13 @@ class Kernel does Systemic {
                 # These are the ones libuv emulates on Windows.
                 @names = flat "", <INT BREAK HUP WINCH>;
             } else {
-                @names = flat "", qx/kill -l/.words;
+                if $*KERNEL.name eq 'openbsd' {
```
zoffixznet: Could we change `$*KERNEL` to `self.name` (or even `$!name`) to avoid needless the dynvar lookup?

stmuk: Done and I'm running tests now.

zoffixznet:
Thanks.

PR #1104, opened 2017-06-17
Fix run/shell :merge

ugexe:
Previously this command would deadlock:
`perl6 -e 'my $proc = run("ls", :merge).out.slurp.say'`


zoffixznet:
Thanks.

PR #1107, opened 2017-06-26
add the option --libdir to rakudo Configuration

gerd:
This patch is used for packaging rakudo at Fedora. It would be nice if it could go to the rakudo sources.

### DIFF
--- a/Configure.pl
+++ b/Configure.pl
@@ -58,7 +58,7 @@ MAIN: {
     }
     $options{prefix} = File::Spec->rel2abs($options{prefix});
     unless (defined $options{libdir}) {
-        my $default = File::Spec->catdir($options{prefix}, 'lib');
+        my $default = File::Spec->catdir($options{prefix}, 'share');
         # print "ATTENTION: no --libdir supplied, building and installing architecture specific files to $default\n";
         $options{libdir} = $default;
     }
--- a/Configure.pl
+++ b/Configure.pl
@@ -53,11 +53,13 @@ MAIN: {
 
     unless (defined $options{prefix}) {
         my $default = defined($options{sysroot}) ? '/usr' : File::Spec->catdir(getcwd, 'install');
+        print "ATTENTION: no --prefix supplied, building and installing to $default\n";
         $options{prefix} = $default;
     }
     $options{prefix} = File::Spec->rel2abs($options{prefix});
     unless (defined $options{libdir}) {
         my $default = File::Spec->catdir($options{prefix}, 'share');
+        
         $options{libdir} = $default;
     }
     my $prefix         = $options{'prefix'};
--- a/Configure.pl
+++ b/Configure.pl
@@ -59,7 +59,6 @@ MAIN: {
     $options{prefix} = File::Spec->rel2abs($options{prefix});
     unless (defined $options{libdir}) {
         my $default = File::Spec->catdir($options{prefix}, 'share');
-        
         $options{libdir} = $default;
     }
     my $prefix         = $options{'prefix'};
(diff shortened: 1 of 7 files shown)


Configure.pl
```diff
@@ -53,10 +53,15 @@
 
     unless (defined $options{prefix}) {
         my $default = defined($options{sysroot}) ? '/usr' : File::Spec->catdir(getcwd, 'install');
-        print "ATTENTION: no --prefix supplied, building and installing to $default\n";
+        # print "ATTENTION: no --prefix supplied, building and installing to $default\n";
         $options{prefix} = $default;
     }
     $options{prefix} = File::Spec->rel2abs($options{prefix});
+    unless (defined $options{libdir}) {
+        my $default = File::Spec->catdir($options{prefix}, 'lib');
```
niner: Defaulting to $prefix/lib breaks compatibility with existing installations forcing our users to re-install all modules. We should not do that without an excellent reason.

Configure.pl
```diff
@@ -53,10 +53,15 @@
 
     unless (defined $options{prefix}) {
         my $default = defined($options{sysroot}) ? '/usr' : File::Spec->catdir(getcwd, 'install');
-        print "ATTENTION: no --prefix supplied, building and installing to $default\n";
+        # print "ATTENTION: no --prefix supplied, building and installing to $default\n";
```
niner: Why do you remove this message? Doesn't seem related to the other changes and the message is actually quite helpful.

Configure.pl
```diff
@@ -53,10 +53,15 @@
...
         $options{prefix} = $default;
     }
     $options{prefix} = File::Spec->rel2abs($options{prefix});
+    unless (defined $options{libdir}) {
+        my $default = File::Spec->catdir($options{prefix}, 'lib');
+        # print "ATTENTION: no --libdir supplied, building and installing architecture specific files to $default\n";
```
niner: Committing commented out code is rarely useful. It's either meant to be run or not.

gerd:
There is no need to change the default destination.

The 'print'  instructions in comments could be removed. So that the user will not be surprised about the message.

zoffixznet:
Is this now mergeable or are more changes needed?

gerd:
I think this pull request is mergeable. To play it safe I will do some more installations today with and without the patch to see if I find some unexpected problems or to see if there are differences in installations with default settings.

gerd:
Sorry for being not in time. The default works fine. The 'perl6' executable works fine when the option '--libdir' is used. I recognized a problem with the command 'perl6-gdb-m' and fixed it. It would be nice if the pull request would be merged now.

dod38fr:
@gerd Hi. I'm packaging rakudo for Debian. I'd like to compare the layout of the rakudo package you produce for fedora with the one I have for Debian.

Could you send me (or add here) a link that list the files shipped by Fedora in rakudo package ?

Thanks

dod for Debian

gerd:
Hello,
I am on a trip for a week. After about 8 days I will send you the file list.

--Gerd

-------- Ursprüngliche Nachricht --------
Von: Dominique Dumont <notifications@github.com> 
Datum:02.09.2017  21:11  (GMT+01:00) 
An: rakudo/rakudo <rakudo@noreply.github.com> 
Cc: Gerd Pokorra <pokorra@uni-siegen.de>,Mention <mention@noreply.github.com> 
Betreff: Re: [rakudo/rakudo] add the option --libdir to rakudo Configuration
  (#1107) 

@gerd Hi. I'm packaging rakudo for Debian. I'd like to compare the layout of the rakudo package you produce for fedora with the one I have for Debian.

Could you send me (or add here) a link that list the files shipped by Fedora in rakudo package ?

Thanks

dod for Debian

dod38fr:
@gerd ok, no problem.

dod38fr:
@gerd Hi. Just a reminder, could you send me (or add here) a link that list the files shipped by Fedora in rakudo package ?

All the best

PR #1108, opened 2017-07-06
Fix 'is export' on constants assigned to routines

LLFourn:
The 'is export' exception fallback is higher precedence than the
constant exporting candidate, so this re-dispatches to the correct
candidate.

Fixes "Can't use unknown trait 'is export' in a ... declaration" for
statements like:

constant $constant-regex is export =  /fo*/;
constant &constant-sub   is export  = sub { "win" };

Fixes [RT #131704](https://rt.perl.org/Public/Bug/Display.html?id=131704)

### DIFF
--- a/src/core/traits.pm
+++ b/src/core/traits.pm
@@ -251,13 +251,12 @@ my $!;
 my $/;
 my $_;
 
-multi sub trait_mod:<is>(Routine:D \r, :$export!) {
+multi sub trait_mod:<is>(Routine:D \r, :$export!, :$SYMBOL = '&' ~ r.name) {
     my $to_export := r.multi ?? r.dispatcher !! r;
-    my $exp_name  := '&' ~ r.name;
     my @tags = flat 'ALL', (nqp::istype($export,Pair) ?? $export.key() !!
                             nqp::istype($export,Positional) ?? @($export)>>.key !!
                             'DEFAULT');
-    Rakudo::Internals.EXPORT_SYMBOL($exp_name, @tags, $to_export);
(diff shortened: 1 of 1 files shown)


LLFourn:
tests: https://github.com/perl6/roast/pull/280

zoffixznet:
This feels a bit of a too big a hammer for the job and a bit of a roundabout. What if we add optional `$SYMBOL` param to the `:$export!` Routine:D candidate and use the symbol as the export symbol if it's needed. This also exports the dispatcher for multies, which I think OP patch doesn't do.

```diff
diff --git a/src/core/traits.pm b/src/core/traits.pm
index 46b7cb1..200d6e2 100644
--- a/src/core/traits.pm
+++ b/src/core/traits.pm
@@ -251,9 +251,9 @@ my $!;
 my $/;
 my $_;
 
-multi sub trait_mod:<is>(Routine:D \r, :$export!) {
+multi sub trait_mod:<is>(Routine:D \r, :$export!, :$SYMBOL) {
     my $to_export := r.multi ?? r.dispatcher !! r;
-    my $exp_name  := '&' ~ r.name;
+    my $exp_name  := $SYMBOL // ('&' ~ r.name);
     my @tags = flat 'ALL', (nqp::istype($export,Pair) ?? $export.key() !!
                             nqp::istype($export,Positional) ?? @($export)>>.key !!
                             'DEFAULT');
```

LLFourn:
@zoffixznet 👍  That's better

LLFourn:
@zoffixznet PR changed to your suggestion

zoffixznet:
Thanks.

PR #1110, opened 2017-07-10
Site wide config: working --prefix option

tbrowder:
This branch enables, with the --prefix option:

1. configure, make, make test, etc. as user
2. install as root into the --prefix dir

The installation is portable (see notes in the Configure.pl file). Note the original installation used absolute file names which hampered portability.

Still to do:

1. correct and improve documentation
2. rename Perl 6 build files with .pl file name extensions to .p6
3. add a timestamp capability to minimize rebuilds when not necessary
4. remove some PR comments and dead code
5. make it work for other OSs

### DIFF
--- a/Configure.pl
+++ b/Configure.pl
@@ -51,13 +51,32 @@ MAIN: {
         exit(0);
     }
 
-    unless (defined $options{prefix}) {
+    #unless (defined ;$options{prefix}) {
+    # the value of instdir remains false unless --prefix is used
+    my $instdir = '';
+    if (defined $options{prefix}) {
+        # save the value in another var: $INSTDIR (instdir)
+        $options{instdir} = $options{prefix};
+        $instdir = $options{instdir};
+        my $default = defined($options{sysroot}) ? '/usr' : File::Spec->catdir(getcwd, 'install');
+        $options{prefix} = $default;
+
+        print "ATTENTION: --prefix supplied, building locally and installing to $instdir\n";
+
+    }
+    else {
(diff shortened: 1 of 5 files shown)


niner:
rpmbuild usually unpacks the sources into a BUILD/rakudo-2017.07 directory and uses BUILDROOT/rakudo-2017.07 as directory for the build results and passes that in via DESTDIR:

```
%build
perl Configure.pl --prefix=%{_usr}
make

%install
make install DESTDIR=$RPM_BUILD_ROOT
```

Is this still possible with your patches?

zoffixznet:
With the release on which Rakudo Star will be based just 5 days away, I'd recommend waiting until *after* the release to merge this. So if there is some case where things break, we can fix it much more easily.

tbrowder:
nine: ref rpmbuild, i removed DESTDIR from the build tools but i don't see why that incantation couldn't be made to work.  to me the most important things about my PR are:

1. the underlying build tools for nqo and moar did not change

2. the configure/build steps are alway done in the working dir by ordinary user, install using --prefix must be done by root unless the user owns the prefix dir ((and its parent if not already created)

3. the ./install directory is portable, i.e., a new install target should be easy to create as for the rpm build

tbrowder:
@niner ref rpmbuild incantation, I think it will work with my PR if you change the steps to:

%build
perl Configure.pl --backend=moar --gen-moar --gen-nqp --prefix=$RPM_BUILD_ROOT

%install
make install

NOTE: All with the caveat that both steps are done in the same working directory.

tbrowder:
Oops, not working now. Closing until fixed.

PR #1112, opened 2017-07-13
Call done/quit on all taps - RT #130919

ugexe:
https://rt.perl.org/Ticket/Display.html?id=130919

`perl6 -e 'my $supplier = Supplier.new; my $supply = $supplier.Supply; $supply.tap(done => { say 1 }); $supply.tap(done => { say 2 }); $supplier.done;'`

Actual (current) output:
1

Expected (PR) output:
1
2

### DIFF
--- a/src/core/Supply.pm
+++ b/src/core/Supply.pm
@@ -204,18 +204,14 @@ my class Supply does Awaitable {
                         emit(value) unless $!finished;
                     },
                     done => -> {
-                        unless $!finished {
-                            $!finished = 1;
-                            done();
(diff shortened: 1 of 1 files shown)


jnthn:
I wonder if this way will break the `Supply` protocol; `my $supplier = Supplier.new; my $supply = $supplier.Supply; $supply.tap(done => { say 1 }); $supplier.done; $supplier.done` should always only have its `done` called once.

jnthn:
Just checked that, and no, it's fine in that regard, I guess 'cus the cleanup takes care of it.

PR #1116, opened 2017-07-15
Fix no message printing when using ll-exception

MasterDuke17:
The recent change to only generate an Exception's message when it's
being gisted broke using `--ll-exception`.


MasterDuke17:
This built fine for me, but `./perl6-m tools/build/install-core-dist.pl /home/dan/Source/perl6/install/share/perl6` seemed to hang. Could someone check this before merging?

PR #1117, opened 2017-07-18
Speed up .contains by flattening the haystack

MasterDuke17:
`nqp::indexingoptimized` flattens any strands in the string, allowing it
to be searched faster.

`say (("abcde" x 100_000_000) ~ "foo").contains("foo"); say now - INIT now` went from 9s to 7.2s, `say (("abcde" x 10) ~ "foo").contains("foo"); say now - INIT now` was unchanged.

Passes `make m-spectest`.

### DIFF
--- a/src/core/Str.pm
+++ b/src/core/Str.pm
@@ -177,7 +177,7 @@ my class Str does Stringy { # declared in BOOTSTRAP
     multi method contains(Str:D: Cool:D $needle) {self.contains: $needle.Str}
     multi method contains(Str:D: Str:D $needle) {
         nqp::p6bool(nqp::isne_i(
-          nqp::index($!value,nqp::getattr($needle,Str,'$!value'),0),-1
+          nqp::index(nqp::indexingoptimized($!value),nqp::getattr($needle,Str,'$!value'),0),-1
         ))
     }
     multi method contains(Str:D: Cool:D $needle, Int(Cool:D) $pos) {self.contains: $needle.Str, $pos}
@@ -186,7 +186,7 @@ my class Str does Stringy { # declared in BOOTSTRAP
           nqp::if(
             (nqp::isge_i($pos,0) && nqp::islt_i($pos,nqp::chars($!value))),
             nqp::isne_i(
-              nqp::index($!value,nqp::getattr($needle,Str,'$!value'),$pos),-1)
+              nqp::index(nqp::indexingoptimized($!value),nqp::getattr($needle,Str,'$!value'),$pos),-1)
           )
         )
     }


samcv:
I'm curious if you can try checking the times of strings created other ways? Because strings made with `x` essentially are just repeated so i'd think it could be slow, but I'm curious what the speed would be if you wrote files of varying sizes concatted them and then checked `.contains`

```perl6
spurt "file", "abcde" x 5_000;
my $var = slurp "file";
$var ~= $var;
$var.contains("foo")
```
Thanks. Curious how that effects things as I've been working with strands and the concatenation code lately.

zoffixznet:
So it made it 25% faster, but look at the memory use: went from 63MB up to 2.5**GB**, which is 35x more and not something that's even available on your typical lower-tier VM.

This is a very clear loss to me.

lizmat:
> On 18 Jul 2017, at 12:07, Zoffix Znet <notifications@github.com> wrote:
> 
> So it made it 25% faster, but look at the memory use: went from 63MB up to 2.5GB, which is 35x more and not something that's even available on your typical lower-tier VM.
> 
> This is a very clear loss to me.

If these numbers are correct, than I would have to agree.

zoffixznet:
I forgot to copy-paste the actual numbers:

```bash
zoffix@VirtualBox~$ /usr/bin/time perl6 -e 'say (("abcde" x 100_000_000) ~ "foo").contains("foo");'
True
8.73user 0.02system 0:08.78elapsed 99%CPU (0avgtext+0avgdata 69044maxresident)k
0inputs+0outputs (0major+12716minor)pagefaults 0swaps
zoffix@VirtualBox~$ /usr/bin/time perl6 -e 'use nqp; say nqp::indexingoptimized(("abcde" x 100_000_000) ~ "foo").contains("foo");'
True
7.96user 1.17system 0:09.28elapsed 98%CPU (0avgtext+0avgdata 2487260maxresident)k
13056inputs+48outputs (58major+326834minor)pagefaults 0swaps
zoffix@VirtualBox~$ 
```

MasterDuke17:
Fwiw:
```
dan@hermes:~/Source/perl6$ /usr/bin/time ./install/bin/perl6-m -e 'say (("abcde" x 100_000_000) ~ "foo") ~~ / "foo" /;'
｢foo｣
2.10user 0.69system 0:02.91elapsed 96%CPU (0avgtext+0avgdata 2509676maxresident)k
39008inputs+0outputs (172major+548941minor)pagefaults 0swaps
```
Finishes in 1/4th the time, but the same amount of memory used as my patch (since it also does an nqp::indexingoptimized under the hood).

So we could have `.contains` as the slow, but more memory efficient way, but it seems crazy that it should be slower than starting up the regex engine.

samcv:
I have a PR for MoarVM that was created in response to this. https://github.com/MoarVM/MoarVM/pull/616

See my second post there for the timings I made before/after for 8bit/32bit strings based on @MasterDuke17 's tests.

zoffixznet:
> So we could have .contains as the slow, but more memory efficient way

Your patch also makes other cases of .contains 14x **slower**:

```
zoffix@VirtualBox~$ /usr/bin/time perl6 -e 'use nqp; say ("foo" ~ ("abcde" x 100_000_000)).contains("foo");'
True
0.13user 0.00system 0:00.13elapsed 98%CPU (0avgtext+0avgdata 68468maxresident)k
0inputs+0outputs (0major+12543minor)pagefaults 0swaps

zoffix@VirtualBox~$ /usr/bin/time perl6 -e 'use nqp; say nqp::indexingoptimized("foo" ~ ("abcde" x 100_000_000)).contains("foo");'
True
1.98user 0.34system 0:02.33elapsed 99%CPU (0avgtext+0avgdata 2510656maxresident)k
0inputs+0outputs (0major+63703minor)pagefaults 0swaps
zoffix@VirtualBox~$ 
```

samcv:
@zoffixznet if you test it by using `now - INIT now` it's actually 530x slower.

from 0.0021549s to 1.1439643s

With my MoarVM PR it takes that down from 1.1439643s to 0.7555986s though that is still 350x slower,.

zoffixznet:
> Your issue about it making it slower in those cases was also something I thought of as well

FWIW my biggest issue with this PR is memory usage. My last Perl 6 Web up ended in failure because of memory, not because of speed.

If we're going to flatten ropes everywhere to get 25% speed gains at the cost of exorbitant memory use, then why have ropes in the first place.

samcv:
@zoffixznet my PR to MoarVM reduces the memory usage to store as 8 bit so it would have 1/5 the peak memory usage you saw. Right now when flattening it allocates enough space for a 32bit string, then iterates through the rope until done. Then if it can fit into 8 bits, it allocates another 8bit string on top of the 32bit and then copies from the 32bit string to the 8bit string, then deallocating the 32 bit 

With my changes it allocates an 8 bit string, iterates the rope copying codepoints to the 8bit string. If it sees a codepoint that won't fit, it then allocates a 32bit string, copies the ones already copied from the 8 bit string, deallocates the 8 bit string and then continues iterating until done.

But I do agree with you, we may not want to flatten in every case because it would cause us to use double the storage space until it is done flattening. (with the change, before it would allocate the 32 bit string even if it wasn't needed, so it could be many times more if we're flattening something stored as 8 bit). I think that issue is bigger than an unrelated synthetic condition where storing 600 million codepoints takes a lot of memory. Unless we think people are actually going to be storing 600 million codepoints in a string.

A bigger issue we need to determine is when should be flatten the string and when should we not. We might do it in cases we maybe shouldn't.

MasterDuke17:
Given @jnthn's comment [here](https://irclog.perlgeek.de/moarvm/2017-07-20#i_14897207), and the increased memory use, I'll close this PR in a day or two unless anyone has any suggestions.

samcv:
That seems sane enough, to keep it so only the regex engine does flattening makes it easier to tell when that may be going on. And also reduces how often a string might be flattened by other operations only to need it again later.

PR #1118, opened 2017-07-19
Improve Version smart match with different lengths

ugexe:
Passes spectest, but changes the behavior of the following to pass:
```
# Not in roast
use Test;
nok v6 ~~ v6c;
nok v6 ~~ v6.d;
nok v6 ~~ v6.c+;
nok v6 ~~ v6.d+';
done-testing;
```

### DIFF
--- a/src/core/Version.pm
+++ b/src/core/Version.pm
@@ -80,22 +80,20 @@ class Version {
         }
     }
     multi method ACCEPTS(Version:D: Version:D $other) {
-        my $oparts := nqp::getattr(nqp::decont($other),Version,'$!parts');
-        my $oelems  = nqp::isnull($oparts) ?? 0 !! nqp::elems($oparts);
+        my $oparts       := nqp::getattr(nqp::decont($other),Version,'$!parts');
+        my int $oelems    = nqp::isnull($oparts) ?? 0 !! nqp::elems($oparts);
+        my int $elems     = nqp::elems($!parts);
+        my int $max-elems = nqp::if(nqp::isge_i($oelems,$elems), $oelems, $elems);
 
-        my int $elems = nqp::elems($!parts);
-        my int $i     = -1;
-        while nqp::islt_i(++$i,$elems) {
-            my $v := nqp::atpos($!parts,$i);
+        my int $i = -1;
+        while nqp::islt_i(++$i,$max-elems) {
+            my $v := nqp::if(nqp::isge_i($i,$elems), Whatever, nqp::atpos($!parts,$i));
 
             # if whatever here, no more check this iteration
             unless nqp::istype($v,Whatever) {
-
-                # nothing left to check, so ok
-                return True if nqp::isge_i($i,$oelems);
+                my $o := nqp::if(nqp::isge_i($i,$oelems), 0, nqp::atpos($oparts,$i));
 
                 # if whatever there, no more to check this iteration
-                my $o := nqp::atpos($oparts,$i);
                 unless nqp::istype($o,Whatever) {
                     return nqp::p6bool($!plus) if $o after  $v;
                     return False               if $o before $v;


ugexe:
```perl6
use v6;
use Test;

subtest 'Non-range version ACCEPTS' => {
    ok v6 ~~ v6,    'v6 ~~ v6';
    ok v6c ~~ v6,   'v6c ~~ v6';
    ok v6d ~~ v6,   'v6d ~~ v6';
    nok v6 ~~ v6c,  'v6 ~~ v6c';
    nok v6 ~~ v6.d, 'v6 ~~ v6.d';

    ok v0 ~~ v0,     'v0 ~~ v0';
    ok v0 ~~ v0.0,   'v0 ~~ v0.0';
    ok v0 ~~ v0.0.0, 'v0 ~~ v0.0.0';
    ok v0.0   ~~ v0, 'v0.0   ~~ v0';
    ok v0.0.0 ~~ v0, 'v0.0.0 ~~ v0';

    ok v0.1 ~~ v0.1,      'v0.1 ~~ v0.1';
    ok v0.1 ~~ v0.1.0,    'v0.1 ~~ v0.1.0';
    nok v0.1.0 ~~ v0.2,   'v0.1.0 ~~ v0.2';
    nok v0.1.0 ~~ v0.1.1, 'v0.1.0 ~~ v0.1.1';
}

subtest 'Whatever version ACCEPTS' => {
    ok v6 ~~ v6.*,  'v6 ~~ v6.*';
    nok v6 ~~ v5.*, 'v6 ~~ v5.*';
    nok v6 ~~ v0.*, 'v6 ~~ v0.*';

    ok v0.1.5 ~~ v0.*.5,  'v0.1.5 ~~ v0.*.5';
    nok v0.1.6 ~~ v0.*.5, 'v0.1.6 ~~ v0.*.5';
}

subtest 'Plus version ACCEPTS' => {
    ok v6 ~~ v6+,    'v6 ~~ v6+';
    ok v6c ~~ v6+,   'v6c ~~ v6+';
    ok v6d ~~ v6+,   'v6d ~~ v6+';
    nok v6 ~~ v6.c+, 'v6 ~~ v6.c+';
    nok v6 ~~ v6.d+, 'v6 ~~ v6.d+';

    ok v0 ~~ v0+,      'v0 ~~ v0+';
    ok v0.2 ~~ v0+,    'v0.2 ~~ v0+';
    ok v0.2 ~~ v0.0+,  'v0.2 ~~ v0.0+';
    ok v0.2 ~~ v0.1+,  'v0.2 ~~ v0.1+';
    ok v0.2 ~~ v0.2+,  'v0.2 ~~ v0.2+';
    nok v0.2 ~~ v0.3+, 'v0.2 ~~ v0.3+';
}

done-testing;
```

lizmat:
I would like to see TimToady's decision on that, as the current implementation is per his wishes, afaik.

ugexe:
## QUESTION

Is `v6 ~~ v6c+` better as True or False?

## PREAMBLE

When smart-matching two Versions the RHS is viewed as a constraint on the LHS. This can be read as:

`$rhs-version.ACCEPTS($lhs-version-range)`

When viewed this way we can then reason that its ok to treat LHS version ranges (`v6+ ~~ v6c+`) as a [DIHWIDT](http://design.perl6.org/S99.html#DIHWIDT) (even though I use it in the first CURRENT BEHAVIOR example).

## PROPOSAL

When smart-matching two Version objects where the number of Version parts differ we should pad the Version with fewer parts before comparing.

Pad LHS with `.0`, such that it is always treated as a concrete version.

Pad RHS with `.*`, such that it falls back to acting like a range constraint.

## CURRENT BEHAVIOR

#### LHS shorter, pad with `.*`
```
$ perl6 -e 'say v6   ~~ v6c'
$ perl6 -e 'say v6.* ~~ v6c'
True
```

#### RHS shorter, pad with `.*`
```
$ perl6 -e 'say v6c ~~ v6'
$ perl6 -e 'say v6c ~~ v6.*'
True
```

## PROPOSED BEHAVIOR

#### LHS shorter, pad with `.0`
```
$ perl6 -e 'say v6   ~~ v6c'
$ perl6 -e 'say v6.0 ~~ v6c'
False
```

#### RHS shorter, pad with `.*`
```
$ perl6 -e 'say v6c ~~ v6'
$ perl6 -e 'say v6c ~~ v6.*'
True
```

## CAVEAT

This PR introduces the issue of `v6.* ~~ v6.c.e.f` expanding to `v6.*.0.0 ~~ v6.c.e.f`, which is arguably wrong. I left this unaddressed since, as mentioned before, a LHS range could be considered DIHWIDT. However it would be trivial to modify the rules to pad the LHS with `.*` instead of `.0` if the last LHS Version part is a Whatever such that we expand to `v6.*.*.* ~~ v6.c.e.f`.

ugexe:
http://design.perl6.org/S11.html#Versioning
```
    When specifying the version of your own module, 1.2 is equivalent to 1.2.0, 1.2.0.0, and so on. However use searches for modules matching a version prefix, so the subversions are wildcarded, and in this context :ver<1.2> really means :ver<1.2.*>. If you say:

        use v6;

    which is short for:

    S11-modules/use-perl-6.t lines 3–11

        use Perl:ver<6.*>;

    you're asking for any version of Perl 6. You need to say something like

        use Perl:<6.0>;
        use Perl:<6.0.0>;
        use Perl:<6.2.7.1>;
```
*Note that this PR aligns with the above.*

If this is the intended behavior of Version's smart-match, this PR addresses the problem that is demonstrated with the following (which should arguably fail):
```
perl6 -e 'use Test:ver(v6.c.1.2.3); ok 1'
ok 1 -
```

TimToady:
I'm fine with this.  In cases where we assume the LHS can be a pattern, it's specifically a type that is always a pattern, such as a Signature.  Here we have a single type that might be literal or pattern, so I think forcing the LHS to literal is a good thing, especially since the we have multiple degrees of freedom in how versions match.  If we ever need to the ability to ask whether one version pattern entirely or partially subsumes another, that's a good thing for a module.

To look at it another way, this is the same decision as forcing actual module versions to be literals rather than wildcards.  (Arguably we could have had different version types for literal vs pattern, as we do for Capture vs Signature, but we don't.)

PR #1120, opened 2017-07-22
Centralize CURI search logic

ugexe:
* Use !matching-dist to do any dist searches to remove duplicated
search logic.

* !matching-dist returns a list of matching dists so that it
can serve a wider range of functions. A thin wrapper method
could use this to provide a list of "candidate" distributions
from a specific CUR.

* Take $api into account when searching dists.

* Set all meta<files>.values to absolute paths, not just the one
being searched for.

### DIFF
--- a/src/core/CompUnit/Repository/Installation.pm
+++ b/src/core/CompUnit/Repository/Installation.pm
@@ -372,9 +372,16 @@ sub MAIN(:$name is copy, :$auth, :$ver, *@, *%) {
     }
 
     method script($file, :$name!, :$auth, :$ver) {
-        my $matching-scripts := self.files($file, :$name, :$auth, :$ver);
-        my $scripts = $matching-scripts.map: { .<files>{$file} }
-        return $scripts.head;
+        my $spec = CompUnit::DependencySpecification.new(
+            :short-name($file),
+            :auth-matcher($auth || True),
+            :version-matcher($ver // '*'),
+        );
+
+        with self!matching-dists($spec) {
+            my $matches-file := $_.map: { self!resources-dir.add(.hash.values[0]<source>) }
+            return $matches-file.head;
+        }
     }
 
     method files($file, :$name!, :$auth, :$ver) {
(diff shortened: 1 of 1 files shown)


src/core/CompUnit/Repository/Installation.pm
```diff
@@ -372,83 +372,68 @@ sub MAIN(:$name is copy, :$auth, :$ver, *@, *%) {
...
+            :auth-matcher($auth || True),
+            :version-matcher($ver // '*'),
+        );
+
+        with self!matching-dist($spec) {
+            my $dist-ids       := $_.map: { .hash.keys[0] }
+            my $matches-spec   := $dist-ids.map: { self!read-dist($_) }
+            my $matches-file   := $matches-spec.grep: { .<files>{$file} }
+
+            $matches-file.map: -> $candi is copy {
```
niner: I think an explicit return would not hurt here as we're within a block and the missing semicolon after the closure doesn't have any effect.

src/core/CompUnit/Repository/Installation.pm
```diff
@@ -372,83 +372,68 @@ sub MAIN(:$name is copy, :$auth, :$ver, *@, *%) {
...
+                # absolutify paths
+                $candi<files>{$_} = self!resources-dir.add($candi<files>{$_}) for $candi<files>.hash.keys;
+                $candi;
             }
         }
-        @candi
     }
 
     method !matching-dist(CompUnit::DependencySpecification $spec) {
```
niner: I'd rename this to !matching-dists as we're now returning a list instead of a single dist.

src/core/CompUnit/Repository/Installation.pm
```diff
@@ -372,83 +372,68 @@ sub MAIN(:$name is copy, :$auth, :$ver, *@, *%) {
...
-                $resource-id ||= self!read-dist($_.basename)<files>{$file};
-                (id => $_.basename, ver => Version.new( $ver || 0 ), :$auth, :$api, :$resource-id).hash
-            }).grep({
-                $_.<auth> ~~ $auth
-                and $_.<ver> ~~ $ver
-            });
-        for @dists.sort(*.<ver>).reverse {
-            return self!resources-dir.add($_<resource-id>);
-        }
+        my $matching-scripts := self.files($file, :$name, :$auth, :$ver);
```
niner: This is a no-go. The whole point of having a .script method was to not have to use .files as the latter unconditionally parses the JSON meta data which costs a lot of time.
In the case of a script the short-name lookup file contains the resource-id (the file name in the resources/ directory) while for a module it contains the source-id + checksum. At least for scripts installed after the introduction of this method.

ugexe: I knew I was overlooking something obvious when my implementation made it appear redundant. Although after reexamining the method maybe we should get rid of `:$name`, which is unused.

ugexe:
Superceded by https://github.com/rakudo/rakudo/pull/1125

PR #1121, opened 2017-07-22
Cache compunits by id instead of spec str

ugexe:
Previously this would cache on a spec string, but the spec
used was the *query* and not a concrete spec.

The earliest a concrete identity can be obtained to cache is
after calling !matching-dist, which handles all the range stuff
so we don't cache something like "6.*" when "6.c" is what will be
loaded. While we could also continue to use the string identity
(just from this concrete version) that would require a Distribution
to be created that is othewise saved by caching on the file-id.

### DIFF
--- a/src/core/CompUnit/Repository/Installation.pm
+++ b/src/core/CompUnit/Repository/Installation.pm
@@ -504,7 +504,6 @@ sub MAIN(:$name is copy, :$auth, :$ver, *@, *%) {
     {
         my ($dist-id, $dist) = self!matching-dist($spec);
         if $dist-id {
-            return %!loaded{~$spec} if %!loaded{~$spec}:exists;
             my $source-file-name = $dist<source>
                 // do {
                     my $provides = self!read-dist($dist-id)<provides>;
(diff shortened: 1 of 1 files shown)


niner:
I think, we should do both: cache on spec str (because it's massively cheaper) and cache on compunit id (because it will handle more cases). The vast majority of DependencySpecifications consist just of the name without any more details. Thus the chances for a cache hit based on the given specification alone are quite high. Removing the cache on spec str would make us do costly file system access just to discover, that we have already loaded the module. But yes, there's still something to be gained by caching on the compunit id.

ugexe:
I considered performance, but could not demonstrate any difference in the majority case:
`perl6 -e 'for ^100000000 { use Test; }; say time - INIT time'`
e.g. other mechanisms already take care of the majority case

ugexe:
I'm going to merge this tomorrow if there are no objections since it improves the current behavior without negatively impacting performance. I'll address the suggestion of caching the spec string in [Centralize CURI search logic #1120](https://github.com/rakudo/rakudo/pull/1120) using code similar to that shown below (but adapted to the PR mechanics of returning a list)
```perl6
 class CompUnit::Repository::Installation does CompUnit::Repository::Locally does CompUnit::Repository::Installable {
     has $!cver = nqp::hllize(nqp::atkey(nqp::gethllsym('perl6', '$COMPILER_CONFIG'), 'version'));
     has %!loaded;
+    has %!seen;
     has $!precomp;
     has $!id;
     has Int $!version;
@@ -418,6 +419,10 @@ sub MAIN(:$name is copy, :$auth, :$ver, *@, *%) {
     }
 
     method !matching-dist(CompUnit::DependencySpecification $spec) {
+        $!lock.protect: {
+            return %!seen{~$spec} if %!seen{~$spec}:exists;
+        }
+
         if $spec.from eq 'Perl6' {
             my $repo-version = self!repository-version;
             my $lookup = $.prefix.add('short').add(nqp::sha1($spec.short-name));
@@ -444,7 +449,9 @@ sub MAIN(:$name is copy, :$auth, :$ver, *@, *%) {
                             !! Version.new($spec.version-matcher))
                     });
                 for @dists.sort(*.value<ver>).reverse.map(*.kv) -> ($dist-id, $dist) {
-                    return ($dist-id, $dist);
+                    $!lock.protect: {
+                        return %!seen{~$spec} = ($dist-id, $dist);
+                    }
                 }
             }
         }
```

ugexe:
Superceded by https://github.com/rakudo/rakudo/pull/1125

PR #1125, opened 2017-07-30
Add candidates method as CURI recommendation manager

ugexe:
Add `method candidates` - All distribution searching/resolving go
through this new method. This gives developers insight into how a
CURI resolves a module request, which was not possible with .resolve
(since it would chain to the next CUR, and only returned a single
result). Many NYI s22 features will require something like this, and
.candidates is indeed mentioned in s22
https://design.perl6.org/S22.html#candidates

!matching-dist is now backed by .candidates, but also caches a single
'best match' so it acts similar to before.

.candidates (i.e. !matching-dist) now returns a Distribution object
and not ($dist-id, $dist-data). This is possible because of a role
that allows our lazy distribution to avoid yet additional costly IO
calls to obtain meta data. Previously this was done by passing around
a hash internally, but now we dogfood our Distribution interface
throughout CURI.

Avoid io after setting up various directories once.

Removes .script, and reverts to using just .files - .script was just
a fast path of .files under certain conditions and lacked the ability
to filter by name. Now .files can provide the faster path for other
files automatically. Additionally this fixes an issue where .script
would invoke older version bin/ scripts from ~/.perl6

Adds :$api field wherever it was missing for consistency.

### DIFF
--- a/src/core/CompUnit/Repository/Installation.pm
+++ b/src/core/CompUnit/Repository/Installation.pm
@@ -503,7 +503,7 @@ sub MAIN(:$name, :$auth, :$ver, *@, *%) {
         CompUnit::DependencySpecification $spec,
         --> CompUnit:D)
     {
-        with self.candidates($spec).head {
+        with self!matching-dist($spec) {
             return CompUnit.new(
                 :handle(CompUnit::Handle),
                 :short-name($spec.short-name),
@@ -530,7 +530,7 @@ sub MAIN(:$name, :$auth, :$ver, *@, *%) {
         CompUnit::PrecompilationStore :@precomp-stores = self!precomp-stores(),
         --> CompUnit:D)
     {
-        with self.candidates($spec).head {
+        with self!matching-dist($spec) {
             my $source-file-name = .meta<source>;
             X::CompUnit::UnsatisfiedDependency.new(:specification($spec)).throw
                 unless $source-file-name;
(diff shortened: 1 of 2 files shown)


niner:
My standard benchmark (require GTK::Simple) does not show any change outside the margin of error :)

niner:
But this is not good :(

```
nine@sphinx:~> zef
Use of uninitialized value of type Any in string context.
Methods .^name, .perl, .gist, or .say can be used to stringify it to something meaningful.
  in sub MAIN at /home/nine/rakudo/install/share/perl6/site/bin/zef line 2
Use of uninitialized value of type Any in string context.
Methods .^name, .perl, .gist, or .say can be used to stringify it to something meaningful.
  in sub MAIN at /home/nine/rakudo/install/share/perl6/site/bin/zef line 2
Could not find /home/nine/.perl6/resources in:
    /home/nine/.perl6
    /home/nine/rakudo/install/share/perl6/site
    /home/nine/rakudo/install/share/perl6/vendor
    /home/nine/rakudo/install/share/perl6
    CompUnit::Repository::AbsolutePath<29344704>
    CompUnit::Repository::NQP<32911168>
    CompUnit::Repository::Perl5<32911208>
  in sub MAIN at /home/nine/rakudo/install/share/perl6/site/bin/zef line 2
  in block <unit> at /home/nine/rakudo/install/share/perl6/site/bin/zef line 2
```

Happens to a zef installed on nom after switching rakudo to your branch.

PR #1126, opened 2017-07-31
Add `ACCEPTS` method to Hash for defined hashes.

nicqrocks:
This should make it so that `my %h = ("a" => 1); say %h ~~ %h` works.

### DIFF
--- a/src/core/Map.pm
+++ b/src/core/Map.pm
@@ -105,6 +105,10 @@ my class Map does Iterable does Associative { # declared in BOOTSTRAP
         so self.keys.any.match($topic);
     }
 
+    multi method ACCEPTS(Map:D: Map:D \m --> Bool) {
+    	self eqv m;
+    }
+
     multi method EXISTS-KEY(Map:D: Str:D \key) {
         nqp::p6bool(
           nqp::defined($!storage) && nqp::existskey($!storage,key)


zoffixznet:
Reading that code, it likely fails stuff like `42 ~~ %(:foo)` because the `proto` hides all the other candidates. Just remove the proto, while leaving the method as a `multi`; there's a proto in `Mu` already.

Also, it should be moved to `Map.pm` (and types changed from `Hash` to `Map`), because Maps are basically immutable Hashes, so semantics on them should be the same.

nicqrocks:
Oops, give me a min to fix that.

zoffixznet:
Don't know if you already know this, but run `TEST_JOBS=8 make spectest` to run the spectest and ensure the change doesn't break any existing tests.

nicqrocks:
Yeah, I've been trying to do that before pushing, but I'm running on kind of a crummy machine. So it takes ~35 min to run.

lizmat:
I'm not clear why the proto is needed.

Also, it looks to me this should for `Map`, as `Hash` is a subclass of `Map`.

nicqrocks:
Something like this then?

zoffixznet:
Thanks.

PR #1137, opened 2017-08-23
Add a source-address parameter to IO::Socket::INet.pm

Skarsnik:
See https://github.com/MoarVM/MoarVM/pull/659


src/core/IO/Socket/INET.pm
```diff
@@ -89,6 +90,7 @@ my class IO::Socket::INET does IO::Socket {
                $family == PIO::PF_INET
             || $family == PIO::PF_INET6
         } = PIO::PF_INET,
+        Str :$source-address is copy = "".Str,
```
MasterDuke17: `:$source-address` should probably be indented a couple spaces to line it up with the rest of the parameters. Also, you shouldn't need the `.Str` on the `""`.

Skarsnik: I get an error on the nqp::connect call (can't unbox the Str to str) if I don't put .Str

JJ:
Can you please try and fix conflicts before it's checked any further?

Skarsnik:
This is probably not needed if https://github.com/rakudo/rakudo/pull/3451 is done properly (need to be able to specificy address/proto), the way to add the source address is actually to be able to bind before connect.

lizmat:
@patrickbkr does that have any relevance on your work lately?

patrickbkr:
@lizmat, did you intend to ping @Kaiepi instead of me?

lizmat:
Indeed I did, sorry for the noise  :-)

Kaiepi:
This is address-y, so yes. This can coexist with my work, but needs changes in order to do so.

src/core/IO/Socket/INET.pm
```diff
@@ -129,15 +135,15 @@ my class IO::Socket::INET does IO::Socket {
 #?endif
         }
         elsif $.type == PIO::SOCK_STREAM {
-            nqp::connect($PIO, nqp::unbox_s($.host), nqp::unbox_i($.port));
+            nqp::connect($PIO, nqp::unbox_s($.host), nqp::unbox_i($.port), nqp::unbox_s($.source-address), nqp::unbox_i($.source-port));
```
Kaiepi, line 138: Particularly here. This can be implemented without changing the `nqp::connect` op.

lizmat:
And there we are, a year on again.  @Kaiepi @Skarsnik what shall we do with it?

Skarsnik:
I can try to have a look if the bind operand was added to nqp to try to reimplement it :)

PR #1138, opened 2017-08-25
Test.pm: Don't print an empty line after a failed test

Leont:
Previously Test.pm would output invalid TAP, in particular in subtests (because while empty lines are legal TAP, in subtest they need to be indented too). This patch should fix that.


zoffixznet:
👍 Thanks.

PR #1140, opened 2017-08-27
Supply.batch: handle no args and $elems == 1

aeruder:
In most cases, the returned supply from the batch method emits arrays of
batched elements.  There's no reason why the following should happen:

$s.batch(elems => 1, seconds => 1000)
  tap sees arrays
$s.batch(elems => 1)
  tap doesn't see arrays

I also think that just batch() should do something sane and also return
a supply that produces arrays.  This provides us a very consistent
interface from the tap-side where we know post-batch we should always be
expecting an array of values.

### DIFF
--- a/src/core/Supply.pm
+++ b/src/core/Supply.pm
@@ -849,8 +849,7 @@ my class Supply does Awaitable {
         }
     }
 
-    method batch(Supply:D $self: :$elems, :$seconds ) {
-        return self if (!$elems or $elems == 1) and !$seconds;  # nothing to do
+    method batch(Supply:D $self: Int(Cool) :$elems = 0, :$seconds) {
         supply {
             my @batched;
             my $last_time;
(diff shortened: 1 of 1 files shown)


aeruder:
Note: this would break tests on roast

https://github.com/perl6/roast/blob/master/S17-supply/batch.t#L60

lizmat:
I'd rather see the test for !$elems done once when entering the sub (e.g. by making a lexical int $todo = $elems // 1

aeruder:
There is also some discussion on the IRC channel of properly handling rational numbers, negative numbers.

AlexDaniel:
So ``if !$elems or`` check should actually be ``if $elems <= 0``?

AlexDaniel:
And ``… or @batched.elems == $elems;`` will probably be better as ``… or @batched.elems >= $elems;`` ?

lizmat:
AlexDaniel: is there a code path that can make the batch contain more elements than $elems?  If so, I would like to know about that code path, that shouldn't be possible to happen.

aeruder:
I believe I have integrated all the suggestions

Using Int(Cool0 for elems to perform automatic coercion to Int, the non-timeout case uses >= as a way to handle negative and 0-sized arrays without an intermediate value.

lizmat:
Thank you!

PR #1142, opened 2017-08-29
Throw an X::Syntax on nameless postfix .::  (Fix for RT#130181)

skids:


skids:
tests are in my roast branch of the same name.  I'll push those once this is merged.

lizmat:
I'm slightly worried about parsing performance for what would essentially be an extra check for every method call.  Do you have any idea on that?

skids:
I only compiled a couple of times so I can't offer statistically significant numbers.  stage stats didn't go outside their normal range.  Trying to finesse LTM to get the block to run only if it chose a literal '::' over <longname> is a bit beyond my rx-foo.

vendethiel:
Would adding a leading `| '::' { ... }` be enough, or not?

vendethiel:
A literal always win to a subrule (that has no literals, even) in LTM, iirc

skids:
Right but you *do* want to match the longname in case the '::' is followed by something.  And you *don't* want the block to run if that happens.  I think (?) is there is a '::' {} rule above longname the {} will get run when it first tries to match just the '::', even though it eventually chooses the longname

Now if we added
  <?before '::'> <longname> {}

...above the <logname> I think that would safely only run the code for all cases where the identifier starts with ::, which is rare.

vendethiel:
Thanks for the precision

PR #1145, opened 2017-08-31
Simplify ignore(mark|case) handling in INTERPOLATE

MasterDuke17:
By switching to use the relatively new nqp::index(ic|im|icim) ops.

Passes `make m-spectest`.


lizmat:
Excellent catch!

PR #1148, opened 2017-09-01
Special-case FailGoal to adhere to S05 (Fix RT#130528)

skids:
(...also, clean up whitespace debris produced during parsing)

  The general-case behavior of the rx ~ operator is correctly
  implemented, but it results in an internal variable name
  leaking, which contradicts an example in S05.

  This is an ugly fix, but doing it "right" would impact
  compile-time hot paths.  This solution confines the mess
  entirely to the error handling path.


skids:
Tests are in my roast branch rt130528.

PR #1150, opened 2017-09-03
Fix INTERPOLATE call with wrong args in p5metachar

MasterDuke17:
I'm not sure how to craft a regex to demonstrate the bug, but printing
out the argument values in INTPEROLATE shows that (e.g., `$monkey` is
always set and `$s` is what changes depending on whether a `use MONKEY`
is in scope, which is wrong).

Some discussion about this here
https://irclog.perlgeek.de/perl6-dev/2017-09-01#i_15101225


MasterDuke17:
Full disclosure, I'm not 100% sure those are the correct values to use for `p5metachar:sym<(??{ })>`, I copied them from `p5metachar:sym<var>`.

PR #1154, opened 2017-09-11
Supress line number if throw X::Package::Stubbed

Gnouc:


src/core/Exception.pm
```diff
@@ -1898,11 +1898,13 @@ my class X::Package::UseLib does X::Comp {
 }
 my class X::Package::Stubbed does X::Comp {
     has @.packages;
-    # TODO: suppress display of line number
     method message() {
         "The following packages were stubbed but not defined:\n    "
         ~ @.packages.join("\n    ");
     }
+    multi method gist(::?CLASS:D: :$sorry = True, :$expect = True) {
```
jnthn: Why are these named parameters being given default values but then never used?

jnthn:
Does this still output the `==SORRY==` before the error?

src/core/Exception.pm
Gnouc: @jnthn Because I override one defined in role [X::Comb](https://github.com/Gnouc/rakudo/blob/edac1d687099794ac163549adff1863839994653/src/core/Exception.pm#L644)

Gnouc:
@jnthn No, as I added a test in https://github.com/perl6/roast/pull/310

Gnouc:
cc @jnthn @zoffixznet What can we do to make it better and clearer?

The only way I can think is about overriding the `multi method gist`, but it will make others confuse like @jnthn comment above.

zoffixznet:
> cc @jnthn @zoffixznet What can we do to make it better and clearer?

I'm not sure what this PR is fixing, sorry. Don't we normally *do* want the line numbers?

The `multi method gist(::?CLASS:D: :$sorry = True, :$expect = True)` looks to me like it could just be `multi method gist()`. You mentioned you're overriding something, but I'm not seeing it. Looks like the arguments aren't used at all and need not be specified, unless you're doing some multi-dispatch tie-breaking.

jnthn:
Worth a mention: they're not needed 'cus all methods have an implicit `*%_`

Gnouc:
@zoffixznet the TODO line said that line number should be suppressed. Yes, the other multi method was defined in `X::Comp` role, which I refered in response to @jnthn comment

zoffixznet:
@moritz [added that comment](https://github.com/rakudo/rakudo/commit/4235dfe0f702bc479de9ba80ef228e3d873d282b) 6 years ago... but I don't understand why we'd want to remove the line number.

From what I can tell so far, all this PR should be doing is simply removing that comment.

moritz:
Back when I wrote that, the line number was likely one inside the compiler, or maybe the line number where the compunit closes; that is more confusing than helpful for the user.

zoffixznet:
> Back when I wrote that, the line number was likely one inside the compiler,

Thanks. Now it shows the line number for where the stub is, so I'd say we don't want to suppress anything anymore. @Gnouc, please just remove the comment.

```
$ perl6 -I. -MFoo -e ''
===SORRY!=== Error while compiling /tmp/tmp.nnEFj6tntx/Foo.pm6 (Foo)
The following packages were stubbed but not defined:
    Foo
at /tmp/tmp.nnEFj6tntx/Foo.pm6 (Foo):4
------> <BOL>⏏<EOL>
    expecting any of:
        statement end
```

moritz:
@zoffixznet is that a recent change? With rakudo 2017.08, I get:

```
moritz@pete:~$ cat stubbed.p6 
class A { ... }
# more stuff say

say 42;
moritz@pete:~$ perl6 stubbed.p6
===SORRY!=== Error while compiling /home/moritz/stubbed.p6
The following packages were stubbed but not defined:
    A
at /home/moritz/stubbed.p6:5
------> <BOL>⏏<EOL>
```

The error points at line 5, the end of the file, not line 1 where A was stubbed.

zoffixznet:
> The error points at line 5, the end of the file, not line 1 where A was stubbed.

Oh, sorry, I didn't realize it could be pointing there. Yes, it points to the end of file for me as well :)

zoffixznet:
> Seems in this case they do, thought I don't know if it's a bug. It can be shortened to just multi method gist(::?CLASS:D: :$) though

I did that change in https://github.com/rakudo/rakudo/commit/7ba9b7cd6f

PR #1156, opened 2017-09-14
Implement pred() and succ() for the Enumeration role

book:
These method will walk the enumeration in the declaration order.

Using Order as an example:
- Order::Same.succ is Order::More,
- Order::Same.pred is Order::Less.

Calling pred or succ on the boundaries will fail with X::OutOfBound.
Using the same example, Order::Less.pred fails with this X::OutOfRange:
"Decrement out of range. Is: Less, should be in Order::Less^..Order::More".

### DIFF
--- a/src/core/Enumeration.pm
+++ b/src/core/Enumeration.pm
@@ -36,16 +36,12 @@ my role Enumeration {
     method pred() {
         my @values := self.^enum_value_list;
         my $index   = @values.first( self, :k );
-        return $index <= 0
-            ?? Failure.new( X::OutOfRange.new( what => "Decrement", got => self, range => @values[0] ^.. @values[*-1] ) )
-            !! @values[ $index - 1 ];
+        return $index <= 0 ?? self !! @values[ $index - 1 ];
     }
     method succ() {
         my @values := self.^enum_value_list;
         my $index   = @values.first( self, :k );
-        return $index >= @values.end
-            ?? Failure.new( X::OutOfRange.new( what => "Increment", got => self, range => @values[0] ..^ @values[*-1] ) )
-            !! @values[ $index + 1 ];
+        return $index >= @values.end ?? self !! @values[ $index + 1 ];
     }
 }
 
--- a/src/core/Enumeration.pm
+++ b/src/core/Enumeration.pm
@@ -32,6 +32,21 @@ my role Enumeration {
             ?? $x
             !! self.^enum_from_value($x)
     }
+
+    method pred() {
+        my @values := self.^enum_value_list;
+        my $index   = @values.first( self, :k );
+        return $index <= 0
+            ?? Failure.new( X::OutOfRange.new( what => "Decrement", got => self, range => @values[0] ^.. @values[*-1] ) )
+            !! @values[ $index - 1 ];
+    }
+    method succ() {
+        my @values := self.^enum_value_list;
+        my $index   = @values.first( self, :k );
+        return $index >= @values.end
+            ?? Failure.new( X::OutOfRange.new( what => "Increment", got => self, range => @values[0] ..^ @values[*-1] ) )
+            !! @values[ $index + 1 ];
+    }
 }
 
 # Methods that we also have if the base type of an enumeration is


zoffixznet:
Some discussion on this change: https://irclog.perlgeek.de/perl6/2017-09-14#i_15165696

zoffixznet:
👍 Thanks!

AlexDaniel:
So .pred and .succ are now O(n) ? :S

zoffixznet:
> So .pred and .succ are now O(n) ? :S

More accurately, Enumeration.succ/.pred now give sane values. There's no "now", if you consider previous behaviour didn't actually produce succeeding/preceeding values of the Enumeration.

AlexDaniel:
Right. I see.

zoffixznet:
Also, considering it's Enumeration, the `n` will be very small in vast majority of cases 😸

AlexDaniel:
Err, I wouldn't be so eager to imply that this is a non-issue. For example, ``Signal`` enum already has 35 values (still small, but some enums in the wild can easily have more values), and if we suppose that one day somebody might iterate over enum values with ``.succ`` (which is as easy as ```January ... December``` or the likes), then you get O(n²). Yes, most of the time it shouldn't cause any trouble, but it is definitely LTA.

I'll create a ticket for it shortly.

AlexDaniel:
Also, I think this doesn't work anyway:
```perl6
enum Animal (Cat => 0, Dog => 0, Human => 42);
say Dog.succ
```
```
Dog
```

book:
Is the ordering issue caused by the data structure used to initialize the enum?

book:
The more I think about this, the more I think @TimToady [is right](https://irclog.perlgeek.de/perl6/2017-09-14#i_15160486):
> in general, .succ and .pred make little sense on enums

Having `.pred` and `.succ` respectively get stuck on the beginning and the end of the list, makes them less useful in an iterator context, anyway. When one want to iterate over the enumeration, there are other ways available:

    .say for MyEnum.^enum_value_list;

So after some more thinking about it, maybe it's best to drop these changes entirely.

book:
@AlexDaniel there's indeed a bug: `@values.first( self, :k );` will return the first enumeration item with the same value. So in your example it stops at `Cat` instead of `Dog`.

book:
Also, in this case `Cat === Dog` is `True`.

zoffixznet:
> and if we suppose that one day somebody might iterate over enum values with .succ

That just sounds like premature optimization to me. Adding a tracking attribute isn't free either. Some forms of `.unique` are O(n^2) and we live just fine with it.

> So maybe it's best to drop these changes entirely.

We can't really. You'd still have .pred/.succ from Numeric making these method calls return some random number instead of next enumeration value

> Is the ordering issue caused by the data structure used to initialize the enum?

It's actually a ~~bug~~ limitation in .WHICH. One of the reasons the current .WHICH system is slated for replacement:

```perl6
enum Foo (a => 0, b => 0); say [a.WHICH, b.WHICH]; # OUTPUT: [Foo|0 Foo|0]
```

> Also, in this case Cat === Dog is True.

You can fix it by using `nqp::eqaddr` for your `.first` test instead of the `===` operator:

```bash
$ perl6 -e 'enum Foo (a => 0, b => 0, c =>  1); use nqp; say Foo.^enum_value_list.first: { nqp::eqaddr(a, $_) }, :k'                                                                                                                         
0
$ perl6 -e 'enum Foo (a => 0, b => 0, c =>  1); use nqp; say Foo.^enum_value_list.first: { nqp::eqaddr(b, $_) }, :k'
1
```

book:
Isn't using nqp at this stage akin to cheating?

book:
Regarding the possibly useful use in `January ... December`, I suspect `January ... *` would turn into an infinite list, because of `December.succ` being the same `December`.

lizmat:
Perhaps we need a Seq coercer and/or an iterator method.

lizmat:
> On 15 Sep 2017, at 12:24, Zoffix Znet <notifications@github.com> wrote:
> It's actually a bug limitation in .WHICH. One of the reasons the current .WHICH system is slated for replacement:
> 
> enum Foo (a => 0, b => 0); say [a.WHICH, b.WHICH]; # OUTPUT: [Foo|0 Foo|0]

This just means we have a bug in Enumeration.WHICH.

zoffixznet:
> Isn't using nqp at this stage akin to cheating?

In a perfect world, perhaps. In practical world, we use nqp all over the place; it's often a lot faster. In this case, though, fixing Enumeration.WHICH might be a better alternative 😄

AlexDaniel:
Hm. What problem *exactly* are we trying to solve by implementing .succ/.pred like this? For example:

Code:
```perl6
enum Animal (Cat => 3, Dog => 2, Human => 4);
say Dog.succ
```

Result:
```
Human
```
So ``.succ`` simply gives some *other* enum (which is irrelevant to the enum value). And this kinda makes sense when you consider that enum values can be non-numeric.

>  > in general, .succ and .pred make little sense on enums

Yeah. There is simply no successor or predecessor when we talk about enumerations.

> So maybe it's best to drop these changes entirely.

Maybe. I'm not entirely sure.

zoffixznet:
> Hm. What problem exactly are we trying to solve by implementing .succ/.pred like this?

We unified behaviour between Bool and other Enumeration:Ds as well as made it return a more relevant value—another Enumeration:D from the same Enumeration:U instead of .succ'ing the basic value the Enumeration:D represents.



>>    So maybe it's best to drop these changes entirely.

>Maybe. I'm not entirely sure.

To me, the unified behaviour looks easy to learn and simple to remember and I like that it's not bound to actual values the enums represent.

PR #1157, opened 2017-09-15
Avoid mistaking two enumeration elements with the same value

book:
This is a fix for this issue (shown in #1156):

    enum Animal (Cat => 0, Dog => 0, Human => 42);
    say Dog.succ; # OUTPUT: «Dog␤»

With this change, the behaviour becomes:

    say Dog.succ; # OUTPUT: «Human␤»

### DIFF
--- a/src/core/Enumeration.pm
+++ b/src/core/Enumeration.pm
@@ -46,12 +46,12 @@ my role Enumeration {
 
     method pred(::?CLASS:D:) {
         my @values := self.^enum_value_list;
-        my $index   = @values.first( self, :k );
+        my $index   = @values.first: { nqp::eqaddr( self, $_ ) }, :k;
         return $index <= 0 ?? self !! @values[ $index - 1 ];
     }
     method succ(::?CLASS:D:) {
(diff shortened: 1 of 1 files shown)


zoffixznet:
👍 Thanks.

book:
I just realized that this patch was probably not necessary any more with 69dae1f3be3827a73ef601b8e2cb2ab3223e2f62. Except after testing it apparently is.

zoffixznet:
> I just realized that this patch was probably not necessary any more

Yeah. It's my fault for suggesting nqp::eqaddr instead of fixing .WHICH.

> Except after testing it apparently is.

The .WHICH fixed needed an extra push; done in https://github.com/rakudo/rakudo/commit/8d938461a9

But don't worry. To close RT#132093, I had [to restructure the methods' code a bit anyway](https://github.com/rakudo/rakudo/commit/55aa7f28d3) and changed the nqp::eqaddr use to the `===` op (not 100% sure nqp::eqaddr alone is sufficient for this purpose)

book:
"a bit" ;-)

PR #1159, opened 2017-09-16
use 'time' as default seed for srand

Gnouc:


Gnouc:
Test https://github.com/perl6/roast/pull/319

pmichaud:
Copying the comment I made on the related PR in roast:  https://github.com/perl6/roast/pull/319

Using the system clock time as a random number seed is often considered to be an unwise choice -- especially if random numbers are somehow used for cryptosecurity purposes. (time isn't random, it's highly predictable. If you can narrow the random number seed down to a specific year, that's only 25 bits of randomness to be dealt with, which modern computers can crack pretty quickly. Narrow it down to a month and it's only about 21 bits of randomness.)

It also isn't a good choice if there's a possibility that multiple instances are started within the resolution of the result (in the case of time, more than one process per second), as they will likely all get the same sequence of "random" numbers.

In general, Perl 6 probably shouldn't provide defaults that are widely considered unsafe or substandard practice.

My $0.02,

Pm

lizmat:
"Perl 6 probably shouldn't provide defaults that are widely considered unsafe or substandard practice."

PR #1160, opened 2017-09-17
Fix needless buffering when running prove with -j

AlexDaniel:
See [RT #132108](https://rt.perl.org/Ticket/Display.html?id=132108).

Submitting this pull request just to get more eyes on it. I'm pretty sure that we can come up with a better solution.

### DIFF
--- a/lib/Test.pm6
+++ b/lib/Test.pm6
@@ -39,6 +39,8 @@ my int $done_testing_has_been_run = 0;
 _init_vars();
 
 sub _init_io {
+    nqp::setbuffersizefh(nqp::getstdin(),  0);
+    nqp::setbuffersizefh(nqp::getstdout(), 0);
     $output         = $PROCESS::OUT;
     $failure_output = $PROCESS::ERR;
     $todo_output    = $PROCESS::OUT;
--- a/lib/Test.pm6
+++ b/lib/Test.pm6
@@ -39,8 +39,8 @@ my int $done_testing_has_been_run = 0;
 _init_vars();
 
 sub _init_io {
-    nqp::setbuffersizefh(nqp::getstdin(),  0);
     nqp::setbuffersizefh(nqp::getstdout(), 0);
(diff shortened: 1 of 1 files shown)


lizmat:
Why is the buffering needless?  If it is needless, but it enhances performance, why remove it?  If the buffering reduces perfomance, why?  that seems counter-intuitive?

niner:
On Sonntag, 17. September 2017 11:57:04 CEST lizmat wrote:
> Why is the buffering needless?  If it is needless, but it enhances
> performance, why remove it?  If the buffering reduces perfomance, why? 
> that seems counter-intuitive?

Buffering doesn't reduce performance, it increases latency. It keeps the TAP 
from reaching the harness in a timely manner, so the harness cannot print an 
updated status.

lizmat:
Understood (now)  :-)

lib/Test.pm6
```diff
@@ -39,6 +39,8 @@ my int $done_testing_has_been_run = 0;
 _init_vars();
 
 sub _init_io {
+    nqp::setbuffersizefh(nqp::getstdin(),  0);
```
jnthn: Setting it on STDIN makes no sense; the op is only about output buffering. Setting it on both STDOUT and STDERR might make sense.

AlexDaniel: do'h! That's exactly what I meant…

AlexDaniel:
FWIW Test.pm6 is already using nqp rather heavily.

AlexDaniel:
I've created a separate ticket for Perl 6-level setbuffersize, see [RT #132111](https://rt.perl.org/Ticket/Display.html?id=132111).

jnthn:
Yeah, to be clear I wasn't saying we had to do the "provide a way without nqp ops" as a pre-req for merging this, just that others will likely want to do the same and we should support it more properly. I'll try and get to it sometime in the next weeks, and try and remember to update Test.pm6 to use it also.

PR #1162, opened 2017-09-18
Make `trim*` subroutines work with `Cool` instance

Gnouc:


Gnouc:
Test https://github.com/perl6/roast/pull/324

lizmat:
Have you tried building / running the spectest with these changes?  I would guess not, as you need to provide a `proto` for multi subs in the setting (it is *not* automatically added).

lizmat:
Also, if the only difference between the `Cool` and `Str` candidate, is the signature, you can get rid of the `Str` candidate, as the `Str` is also `Cool`.

lizmat:
But probably you need to stringify the value first.

Gnouc:
@lizmat I did and the building is successful. The test is fail but not because of my changes.

I don't think it's good to remove `trim*` for `Str`. They're dedicated there for string. Other type actually is `Stringy` first, that what we currently do for `Cool`. At first, I also do `$c.Stringy.trim`, but I realize we have `trim` method for `Cool`, so there's no reason to use `$c.Stringy.trim`.

Gnouc:
@lizmat I made the changes, per: https://irclog.perlgeek.de/perl6-dev/2017-09-18#i_15179761

PR #1163, opened 2017-09-18
Add Scalar indicators to Hash[].perl when needed (fix RT#132119)

skids:


skids:
Test in roast PR#325

zoffixznet:
👍 Thanks!

PR #1164, opened 2017-09-19
Make :delete works with lazy Arrays

Gnouc:
Fixes RT#131790


Gnouc:
Test https://github.com/perl6/roast/pull/326

PR #1165, opened 2017-09-19
fix subset isa method on a subset.

dwarring:
perl6 -e'subset S of Int; subset S2 of S; say S2.isa(S)'

now returns True. fully resolves RT #132073

### DIFF
--- a/src/Perl6/Metamodel/SubsetHOW.nqp
+++ b/src/Perl6/Metamodel/SubsetHOW.nqp
@@ -58,7 +58,8 @@ class Perl6::Metamodel::SubsetHOW
     }
 
     method isa($obj, $type) {
-        $!refinee.isa($type);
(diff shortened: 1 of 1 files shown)


zoffixznet:
Does `perl6 -e'subset S of Int; subset S3 of S; subset S2 of S; say S2.isa(S3)'` return `False`

Not very familiar with this part of codebase, but looks like the added condition simply checks whether the given type is a subset… *any* subset.

dwarring:
Yes, that returns False.
I borrowed the added logic from [the type_check method](https://github.com/rakudo/rakudo/blob/0704cd97226e63001943426666c88cef1c5fe711/src/Perl6/Metamodel/SubsetHOW.nqp#L78)

zoffixznet:
👍 Thanks!

PR #1166, opened 2017-09-20
dogbert17++ fix harness6 by only allowing TEST_JOBS=1

stmuk:


zoffixznet:
Does a ticket exist for the breakage? If not, someone should file it, so it doesn't get forgotten.

PR #1168, opened 2017-09-22
Restrict dynamic lookup metasyntax in rx EVAL (security RT#131079)

skids:


lizmat:
It doesn't break any spec tests, but also doesn't pass any todo's.

How can we know whether this solves the security issue?

AlexDaniel:
Well, we can just take it and then mark [RT #131079](https://rt.perl.org/Public/Bug/Display.html?id=131079) with ｢testneeded｣.

skids:
I should have mentioned... a patch to roast is already prepared in my rt131079 roast branch to test this case.

zoffixznet:
This doesn't fix all the cases:

```perl6
$ rakudo -e 'my $re = q|<OUTER::(shell "echo pwnd" and exit)>|; say "foo" ~~ /<$re>/;'
pwnd
```

skids:
Thanks, Zoffix, I'll try to work some new patch up for those.

zoffixznet:
Here's another one; slightly different tack that requires attacker to provide the string to match against as well.

```
$ rakudo -e 'my $user-string = "shell qw/echo pwnd/ and exit"; my $user-re = q|.+ <EVAL>|; $user-string ~~ /<$user-re>/;'
pwnd
```

skids:
From what I can tell the first two cases should get shot down not with a security exception,
but should be caught by an ambiguous syntax check, as these non-rx examples show:

    $ perl6 -e 'my $f =  IO::("File"); $f.say;'
    Combination of indirect name lookup and call not supported
    at -e:1
    ------> my $f =  IO::("File")⏏; $f.say;
        expecting any of:
            argument list

(the correct way to do the above is use leading colons, and the security check would catch those)

I'll try adding similar checks in Actions.nqp in the rx assertion rule as are in the main language rule.

There are two other similar error messages in Grammar, but darned if I can figure out how to trigger them:
    elsif $<arglist> {
                $/.panic("Combination of indirect name lookup and type arguments not supported");
            }
            elsif $<accept> || $<accept_any> {
                $/.panic("Combination of indirect name lookup and coercion type construction not supported");
            }

...I think maybe these syntax situations have become unreachable due to smiley processing or other changes over time?  Or maybe someone can figure out how to trigger those error messages and they might lead to other test cases to put in roast...

I'll also look at the direct calling of EVAL

AlexDaniel:
Dammit. This stuff should be done with a whitelist, not a blacklist.

skids:
Well, given what we are talking about here is allowing user input that contains a rx, that would be one huge whitelist... really the problem here is we are catering to a crazy use case.

That said, one way to do that would be to have an entirely separate rx grammar object for user-provided EVALs.

AlexDaniel:
My point is that blacklisting is simply not going to work. We'll have a security hole there forever.

skids:
I think that is pessimistic, personally, and in addition it isn't a use case one should engage in without taking some precautions on the initiating side.

zoffixznet:
> My point is that blacklisting is simply not going to work. We'll have a security hole there forever.

I never seen the eval monkey as a security feature. It's merely there to advise newbie developers they're doing something potentially dangerous.

skids:
I'll push another patch maybe tonight.  It doesn't handle the `<EVAL>` case, as that is a bigger issue, already noted in RT#128684, which should probably be tagged [SECURITY]

Part of the problem here is some of what should work with indirect references is NYI, and the spec could use a bit of review.  It only mentions the `<::(...)>` form, and it is not clear as to whether mentioning this is just an elaboration of the more general case of an identifier.  As such, I marked one obvious NYI as NYI, and the other (a dynamic alias name) as Syntax::Reserved for now.

PR #1169, opened 2017-09-22
Fix PIO table so they are equal to the kernel values + add family packing

the-eater:
This will allow enforcing IPv4 and IPv6 sockets, but also allow the creation of `AF_UNIX`/`PF_UNIX` sockets

Currently this will renumber the PIO table as it's not the same as their C counterparts

Other part to fix MoarVM/MoarVM#683

### DIFF
--- a/src/core/IO.pm6
+++ b/src/core/IO.pm6
@@ -9,20 +9,21 @@ enum SeekType (
   :SeekFromEnd(2),
 );
 enum ProtocolFamily (
-  :PF_LOCAL(0),
+  :PF_UNSPEC(0),
+  :PF_LOCAL(1),
   :PF_UNIX(1),
   :PF_INET(2),
-  :PF_INET6(3),
-  :PF_MAX(4),
+  :PF_INET6(10),
+  :PF_MAX(44),
 );
 enum SocketType (
-  :SOCK_PACKET(0),
   :SOCK_STREAM(1),
   :SOCK_DGRAM(2),
   :SOCK_RAW(3),
   :SOCK_RDM(4),
   :SOCK_SEQPACKET(5),
-  :SOCK_MAX(6),
+  :SOCK_PACKET(10),
+  :SOCK_MAX(11),
 );
 enum ProtocolType (
   :PROTO_TCP(6),
--- a/src/core/IO/Socket/INET.pm6
+++ b/src/core/IO/Socket/INET.pm6
@@ -1,9 +1,10 @@
 my class IO::Socket::INET does IO::Socket {
     my module PIO {
-        constant PF_LOCAL       = 0;
+        constant PF_UNSPEC      = 0;
+        constant PF_LOCAL       = 1;
         constant PF_UNIX        = 1;
         constant PF_INET        = 2;
-        constant PF_INET6       = 3;
+        constant PF_INET6       = 10;
         constant PF_MAX         = 4;
         constant SOCK_PACKET    = 0;
         constant SOCK_STREAM    = 1;
@@ -24,19 +25,31 @@ my class IO::Socket::INET does IO::Socket {
     has Int $.localport;
     has Int $.backlog;
     has Bool $.listening;
-    has $.family = PIO::PF_INET;
+    has $.family = PIO::PF_UNSPEC;
     has $.proto = PIO::PROTO_TCP;
     has $.type = PIO::SOCK_STREAM;
 
+
+
     my sub split-host-port(:$host is copy, :$port is copy, :$family) {
+        if $family === PIO::PF_UNIX {
+            return ($host, $port);
+        }
+
         if ($host) {
-            my ($split-host, $split-port) = $family == PIO::PF_INET6
-                ?? v6-split($host)
-                !! v4-split($host);
+            my ($split-host, $split-port) = (Str, Int);
+
+            if $family !== PIO::PF_INET6 {
+                ($split-host, $split-port) = v4-split($host);
+            }
+
+            if ! $split-port && $family !== PIO::PF_INET {
+                ($split-host, $split-port) = v6-split($host);
+            }
 
             if $split-port {
                 $host = $split-host.Str;
-                $port //= $split-port.Int
+                $port //= $split-port.Int;
             }
         }
 
@@ -63,7 +76,9 @@ my class IO::Socket::INET does IO::Socket {
         Int    :$family where {
                 $family == PIO::PF_INET
              || $family == PIO::PF_INET6
-        } = PIO::PF_INET,
+             || $family == PIO::PF_UNIX
+             || $family == PIO::PF_UNSPEC
+        } = PIO::PF_UNSPEC,
                *%rest,
         --> IO::Socket::INET:D) {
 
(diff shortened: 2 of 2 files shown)


zoffixznet:
FWIW, [the docs](https://docs.perl6.org/type/IO::Socket::INET#method_new) don't mention the constants at all and instead use Int `2` or `3` for family. (edit: [I swapped them to constants now](https://github.com/perl6/doc/commit/3ecad59a153d765517913876ef1010001e00c7bf))

This PR would break code of any user who followed the docs. I suggest more caution with merging this PR. Perhaps [a notice can be posted on rakudo.org](http://rakudo.org/category/notices/), telling people to switch to `PF_INET` and `PF_INET6` constants.

zoffixznet:
And the plot moistens… [the constants](https://github.com/rakudo/rakudo/blob/73aeee6ce7bfe24e4e57ce97fb6972c3be00c76f/src/core/IO/Socket/INET.pm#L3-L18) in `IO::Socket::INET` don't appear to be available to users at all, instead, there's [a ProtocolFamily enum](https://github.com/rakudo/rakudo/blob/73aeee6ce7bfe24e4e57ce97fb6972c3be00c76f/src/core/IO.pm#L11-L17) in `src/core/IO.pm`.

The values should be changed there too, at the least, but sounds like we need to rethink this duplicate stuff or something.

the-eater:
We could for now translate 3 to 10 and show a notice

`Warning: using 3 for PF_INET6 is deprecated, please use the constant PF_INET6`

this would also make sure peoples code won't break yet, and PF 3 is `PF_AX25` and I don't think that still has much use, let alone is used in Perl6, `PF_LOCAL` and `PF_UNIX` aren't supported yet so I don't think we'll have to worry about those, but I might be wrong

AlexDaniel:
> Perhaps a notice can be posted on rakudo.org, telling people to switch to PF_INET and PF_INET6 constants.

FWIW, it's also a good idea to ping me about things like this so that I aware of stuff (or whoever the release manager is at the moment). There's always a chance I can miss important bits from discussions on PRs.

the-eater:
@zoffixznet I now removed the constants from `IO::Socket::INET::PIO` and fixed them in `src/IO.pm`

src/core/IO/Socket/INET.pm
```diff
@@ -55,18 +53,34 @@ my class IO::Socket::INET does IO::Socket {
         return $host ?? ($host, $port) !! $uri;
     }
 
+    method !correct-family(Int:D $family) {
+        if ($family == 3) {
+            note "You're using the old value (3) for IPv6 / PF_NET6, " ~
+                "please use the constant PF_INET6 to specify IPv6";
+            note Backtrace.new.Str;
+
```
zoffixznet: Here, would you change to use [DEPRECATED() routine](https://github.com/rakudo/rakudo/blob/ebd6440c27ea07e2bfde4809fa4e97dac1491e69/src/core/Deprecations.pm#L54) instead of two `note`s. If you grep the codebase for `DEPRECATED` you'll find a few usage examples.

the-eater: I want to use the DEPRECATED routine, but I've been unable to get DEPRECATED to skip the last 2 functions in the trace (as those are just internal functions).

any tips how to get around that?

zoffixznet: I think that's the `:up` argument. Try giving it some numbers. Grep the source for other uses of DEPRECATED and see how it's used.

the-eater: yes, I've been unable to get that to work, would always keep using the last call (`!correct-family`), even if I put `:up` on a ridiculous high number, maybe it has been fixed now, but this is why this this PR didn't progress after that anymore.

zoffixznet: And `:up(*)` doesn't work either?

the-eater: Didn't try that yet! will do when I have some time on my hands, thanks @zoffixznet

the-eater: Works! thanks @zoffixznet!

src/core/IO.pm6
```diff
@@ -9,20 +9,21 @@ enum SeekType (
...
 );
 enum ProtocolFamily (
-  :PF_LOCAL(0),
+  :PF_UNSPEC(0),
+  :PF_LOCAL(1),
   :PF_UNIX(1),
   :PF_INET(2),
-  :PF_INET6(3),
-  :PF_MAX(4),
+  :PF_INET6(10),
```
the-eater: This should be retrieved from the OS, Linux has `10`, Windows `23`, and BSD iirc `28`. how...

zoffixznet:
@AlexDaniel would you review/merge this? I assume we'd want the DEPRECATED sub to show some date for when the old version would be removed and the change'd need a p6lert issued.

src/core/IO.pm6
the-eater: We can adhere to the Linux numbers inside Rakudo, and transform those to platform specific numbers in the VM. but I really dislike that idea

the-eater:
Currently this would break IPv6 support for Mac, BSD and Windows. altho to be frank. there is no explicit IPv4/6 support yet. but will result in a panic (I guess). when ran there, when explicitily IPv6 needs to be used

src/core/IO.pm6
the-eater: @zoffixznet you have any input on this?

AlexDaniel:
Oops, I *just* noticed that I'm assigned here.

FWIW there is also this set of PRs which is slightly related: https://github.com/rakudo/rakudo/pull/1137.
And also this comment: https://github.com/MoarVM/MoarVM/pull/659#issuecomment-352561379

Ping @Skarsnik.

the-eater:
Note that the family packing is already merged in MoarVM/MoarVM#630, it just doesn't happen at Rakudo's side yet

AlexDaniel:
@the-eater @Kaiepi This sliiped through the cracks… But what is its current status? Can we just merge it, or does it require more work?

JJ:
Plus... conflicts.

AlexDaniel:
> Plus... conflicts.

Just because the same lines were touched doesn't mean I wouldn't be able to merge.

the-eater:
this was 3 years ago, so can't remember much of it, iirc there was the discussion of which codes we had to respect for the PF_* (Linux? BSD?) since they differ, my goal was then to find a way to configure those values at compile time so they always reflect the corresponding OS's, but me then not being very C/C++ savvy couldn't find a way and kind of,, dropped it

I think since this is now @Kaiepi part now it's entirely their call, since I imagine they did some other things with the network stack, that might either work or fail to work with my 3 years ago patch

Kaiepi:
Between when this was made and now, there were a couple changes made to how socket families are handled:
- Families are an independent parameter from ports now for socket ops, so the family packing here's no longer necessary.
- I made it so socket families are independent values from their kernel values for the time being, since the JVM and JS backends complicated things a bit with that WRT consistent values for families, types, protocols, and eventually, socket options, IIRC?

Having socket family, type, and protocol constants be consistent with their kernel values would be great to have so they can be used with NativeCall, just I wasn't sure how it should be implemented and it isn't super critical to completing the IP6NS grant's work. I wouldn't mind this happening on the side.

lizmat:
Should this PR be closed, because of recent developments?

PR #1170, opened 2017-09-23
Look for yada fullfillment in default parent type, too (RT#131676)

skids:
Role interface/requirements should be satisifiable by Any/Mu,
  but at the point we check them, they have not been appended to
  the MRO yet, unless we are a subclass of an already-composed
  class that has them.  So add the default parent type from the
  class HOW to the list of classes we check for fullfillment.

  In the case of "class C is Mu" we go down a different path which
  ends up skipping Any, as it should.

### DIFF
--- a/src/Perl6/Metamodel/RoleToClassApplier.nqp
+++ b/src/Perl6/Metamodel/RoleToClassApplier.nqp
@@ -7,7 +7,18 @@ my class RoleToClassApplier {
             return nqp::existskey(%mt, $name);
         }
         else {
-            for $target.HOW.mro($target) {
+
+            # In an uncomposed class we need to look for methods in
+            # the part of the mro that has not been added yet.
+            my @mro := $target.HOW.mro($target);
+            if $target.HOW.parents($target, :local(1)) == 0
+              && $target.HOW.has_default_parent_type
+              && $target.HOW.name($target) ne 'Mu' {
+                my $dt := $target.HOW.get_default_parent_type;
+                nqp::splice(@mro, $dt.HOW.mro($dt), nqp::elems(@mro), 0);
+            }
+
+            for @mro {
                 my %mt := $_.HOW.method_table($_);
                 if nqp::existskey(%mt, $name) {
                     return 1;
--- a/src/Perl6/Metamodel/RoleToClassApplier.nqp
+++ b/src/Perl6/Metamodel/RoleToClassApplier.nqp
@@ -7,18 +7,7 @@ my class RoleToClassApplier {
             return nqp::existskey(%mt, $name);
         }
         else {
-
-            # In an uncomposed class we need to look for methods in
-            # the part of the mro that has not been added yet.
-            my @mro := $target.HOW.mro($target);
-            if $target.HOW.parents($target, :local(1)) == 0
-              && $target.HOW.has_default_parent_type
-              && $target.HOW.name($target) ne 'Mu' {
-                my $dt := $target.HOW.get_default_parent_type;
-                nqp::splice(@mro, $dt.HOW.mro($dt), nqp::elems(@mro), 0);
-            }
-
-            for @mro {
+            for $target.HOW.mro($target) {
                 my %mt := $_.HOW.method_table($_);
                 if nqp::existskey(%mt, $name) {
                     return 1;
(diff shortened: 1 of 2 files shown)


src/Perl6/Metamodel/RoleToClassApplier.nqp
```diff
@@ -7,7 +7,18 @@ my class RoleToClassApplier {
...
-            for $target.HOW.mro($target) {
+
+            # In an uncomposed class we need to look for methods in
+            # the part of the mro that has not been added yet.
+            my @mro := $target.HOW.mro($target);
+            if $target.HOW.parents($target, :local(1)) == 0
+              && $target.HOW.has_default_parent_type
+              && $target.HOW.name($target) ne 'Mu' {
+                my $dt := $target.HOW.get_default_parent_type;
+                nqp::splice(@mro, $dt.HOW.mro($dt), nqp::elems(@mro), 0);
```
jnthn: This modifies the `@mro` array in-place, but the `mro` method that computes it also, I believe, caches the result. Since `has_method` is called many times, then I suspect this ends up adding the default parent type each time so the thing in `mro` is not only "polluted" but also gets longer. I'd suggest:

1. Do an `nqp::clone` on the result of calling `$target.HOW.mro($target)`
2. Move that logic out of `has_method` and into the caller of it, and pass in an `@mro` parameter, so we only have to do the work once

These details aside, the patch seems reasonable.

skids: As far as I can tell, though the mro method does cache that, the cached copy is either A) on a call to a role that has not been composed in which case it is blown away (ClassHOW line 104) by a new call to compute_mro immediately thereafter, or B) never gets altered because a composed class is either Mu or has parents.  But yeah the loop over the @collisions and %methods would accumulate extra cruft.

I guess the EnumHOW also uses this.

Anyway, is there a reason not to just move the .add_parent call in ClassHOW above the whole role manipulation, and just hold off on setting $!composed until after roles are done?  That would be even more efficient... the only reason Any/Mu are missed by the role applier is that they are not in the .mro.

skids:
Ah I see now... the add_parent call sequence is to get things in a neat order.

The updated patch moves munging the mro to inside ClassHOW which makes
it easier to see that the munged mro is thrown out later.

I can't seem to figure out a way to cause a version of the original bug with
an enum, at  the moment... so this might be sufficient, and more efficient?

JJ:
Just checking in. What seems to be the problem with this PR?

patzim:
I just checked this out and tried to see if the examples in rt131676 are gone. They are. But I also fail to reproduce them without this patch. Did this get fixed somehow otherwise?

patzim:
[bisectable6](https://gist.github.com/Whateverable/3250098e49374f4a8ae0590eefc15eb3) points to [this commit](https://github.com/rakudo/rakudo/commit/fd5d5bdfb97930cfae1f9b2572d2ab6eea7ca92f). That commit touches the same bit of code this PR does.
I haven't dug in deep enough to know the changes actually do the same thing, but it's sad to see work being done twice because of PRs passing by unnoticed... :(

skids:
> That commit touches the same bit of code this PR does.

Probably best to close this pull request, since it appears that the other commit manages to fix the issue while also fixing an issue with "handles" stuff.  This commit didn't address "handles" at all.

However, that commit altered the return value of RoleToClassApplier.apply and did not check to see that that method is also called from the Enum code, so it broke this:

`role FFFF { method gah {...} }; enum f does FFFF <Fee Fie Foo Fum>;`

EnumHow needs the same treatment ClassHow got in that patch, but, a bit of a refactor may be
worth considering to get the error message generation and has_method procedure confined back to RoleToClassApplier.nqp, rather than having copies of it metastasize into EnumHow as well as ClassHow.

Also it's been so long since I touched this I really can't say whether the way that patch works is correct or if it introduced other issues by causing things to get cached early or messing with the order in which stuff appears in the mro.  It took me a couple solid nights just to wrap my head around it way back then. :-)

JJ:
> > That commit touches the same bit of code this PR does.
> 
> Probably best to close this pull request, since it appears that the other commit manages to fix the issue while also fixing an issue with "handles" stuff. This commit didn't address "handles" at all.
> 

OK, closing it then. It's also got conflicts, so it's probably better to drop it. Thanks @skids anyway for your work.

PR #1171, opened 2017-09-24
In slurp, don't call nqp::join if not needed

MasterDuke17:
nqp::join creates a copy of the string(s) during processing and then
creates another copy to return. For slurping large files this can be
very wasteful if there weren't any "\r\n"s to split on.

For a 50mb file of 1million lines, the time to slurp it dropped from
2.3s to 1.8s, and ~230mb of memory didn't need to be allocated.

Rakudo builds fine and passes `make m-test m-spectest`.

### DIFF
--- a/src/core/IO/Path.pm
+++ b/src/core/IO/Path.pm
@@ -609,9 +609,12 @@ my class IO::Path is Cool does IO {
             $handle,
             nqp::stmts(
                 (my $blob := $handle.slurp(:close)),
-                nqp::if($bin, $blob, nqp::join("\n",
-                  nqp::split("\r\n", $blob.decode: $enc || 'utf-8')))
-            ))
+                nqp::if($bin, $blob,
+                    nqp::if(
+                        nqp::isne_i(nqp::elems(my $l := nqp::split("\r\n", $blob.decode: $enc || 'utf-8')), 1),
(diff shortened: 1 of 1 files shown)


AlexDaniel:
This original ``\r\n`` →``\n`` substitution hardly makes any sense to me. We are not doing ``\r`` →``\n``, so it's a half-step in achieving nothing. Then, we have this trap: https://docs.perl6.org/language/traps#Splitting_the_Input_Data_Into_Lines
But not only ``.lines`` splits by whatever it wants to (which is ok), there's a trap inside the trap!  ``slurp.whatever`` is not going to work anyway because slurp itself is mangling newlines. Woah. So you *have* to use *``:bin``* if you are processing arbitrary *text* input line by line. I don't know if you can split a Buf conveniently, but it seems that the easiest way is ``slurp(:bin, ‘foo’).decode.split(“\n”)``, or something.

MasterDuke17:
Fwiw, https://github.com/MoarVM/MoarVM/pull/705/commits/40da3e6e0b315806269226dd1bb23f83f69e10a9 removes the need for this PR.

MasterDuke17:
Closing since https://github.com/MoarVM/MoarVM/pull/705 was merged.

PR #1172, opened 2017-09-26
Make lazy range works as Array index

cuonglm:
Fixes RT#127573

### DIFF
--- a/src/core/array_slice.pm
+++ b/src/core/array_slice.pm
@@ -54,7 +54,9 @@ multi sub POSITIONS(
         }
     }
 
-    my \pos-iter = pos.iterator;
+    my \pos-iter = pos ~~ Range && pos.is-lazy && !SELF.is-lazy                                                    # If the range is lazy but the iterable is not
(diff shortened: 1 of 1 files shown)


zoffixznet:
Briefly looking at this, looks like it only fixes Range endpoints for non-lazy iterables, but none of that needs to be true for the bug to exit:

```
<Zoffix__> m: my @a = lazy 1,2,3,4,5; @a[1]:delete; .say for @a[lazy 0, 1, 2, 3, 4]
<camelia> rakudo-moar 8cf083: OUTPUT: «1␤»
```

Lazy index lists have a feature that they auto-stop when end of the iterable they're indexing is reached. The bug is due to the hole from deletion is making this feature think it reached the end of the iterable.

Gnouc:
@zoffixznet Yes, I only addressed the problem described in ticket, will take a look more deeply when I have time.

zoffixznet:
@Gnouc don't know if you know already, but we have a developers' chat where you can talk with humans and ask questions if you have any: https://webchat.freenode.net/?channels=#perl6-dev

Gnouc:
@zoffixznet Yes, I'm `cuonglm` there 😄

But I'm in UTC+7 timezone, so just conflict working time with most people there 😭

JJ:
The corresponding ticket [is closed](https://github.com/Raku/old-issue-tracker/issues/5144). There are also conflicts. I'm closing this. Please open a fresh PR if you think this is still an issue.

PR #1175, opened 2017-09-28
Move security RT#131079 fix from Grammar to Actions

skids:
This lets us catch compound name cases
  Use appropriate syntax errors before flagging restricted status
  Give prohibition on longname aliases a typed exception


skids:
Tests are in my rt131079 roast branch,

zoffixznet:
👍 Thanks!

PR #1177, opened 2017-10-01
Fix some unspace parsing cases (RT#128462 and degenerate unspace)

skids:
see RT for analysis/explanation


skids:
Additional tests and unfudge are in my rt128462 roast branch

zoffixznet:
👍 Thanks!

PR #1178, opened 2017-10-04
Implement metamethod shorthand syntax (RT#131478)

skids:
Silence Perl5 warning on $.<!alpha>
  Create a faux non-multipart-name desigilname token
  Teach dissect_longname to digest said token
  Teach variable action to generate a p6callmethodhow
  Adjust parsing to handle the meta thrigil

### DIFF
--- a/src/Perl6/Actions.nqp
+++ b/src/Perl6/Actions.nqp
@@ -2706,7 +2706,7 @@ class Perl6::Actions is HLL::Actions does STDActions {
                 setup_attr_var($/, $past);
             }
         }
-        elsif $twigil eq '.' && $*IN_DECL ne 'variable' {
+        elsif ($twigil eq '.' || $twigil eq '.^') && $*IN_DECL ne 'variable' {
             if !$*HAS_SELF {
                 $*W.throw($/, ['X', 'Syntax', 'NoSelf'], variable => $name);
             } elsif $*HAS_SELF eq 'partial' {
@@ -2714,7 +2714,12 @@ class Perl6::Actions is HLL::Actions does STDActions {
             }
             # Need to transform this to a method call.
             $past := $<arglist> ?? $<arglist>.ast !! QAST::Op.new();
-            $past.op('callmethod');
+            if $twigil eq '.^' {
+                $past.op('p6callmethodhow');
+            }
+            else {
+                $past.op('callmethod');
+            }
             $past.name($desigilname);
             $past.unshift(QAST::Var.new( :name('self'), :scope('lexical') ));
             # Contextualize based on sigil.
--- a/src/Perl6/Grammar.nqp
+++ b/src/Perl6/Grammar.nqp
@@ -2097,7 +2097,7 @@ grammar Perl6::Grammar is HLL::Grammar does STD {
     }
 
     token special_variable:sym<$.> {
-        <sym> {} <!before \w | '('>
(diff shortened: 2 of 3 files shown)


skids:
Tests in my rt131478 roast branch.

Note that when you have no 'self' available '$.^method' will throw an error about that, rather than try to run a metamethod on an in-place nameless variable.  I think that doing the latter is of pretty low utility, and if you really need to, you can instead '($).^mro.say'

zoffixznet:
IMO we should just make it generate a better error instead of making it a working syntax. Simply to cut-down on maintenance burden for devs and amount of learning for end users for syntax that's unlikely to be commonly used and has an alternative that's just 3-chars longer.

Because it's not just the `$.^` syntax; none of these work either:

```
m: class Foo { method x { $.?name.say } }.new.x
m: class Foo { method x { $.+name.say } }.new.x
m: class Foo { method x { $.*name.say } }.new.x
m: class Foo { method x { $."name"().say } }.new.x
m: class Foo { method x { $.Any::elems.say } }.new.x
m: class Foo { method x { $.&say } }.new.x
m: class Foo { method x { $.$(&say) } }.new.x
```

skids:
The difference here is that the above examples semantically refer to an anon state variable, whereas $.^thing is a meta-attribute access, so I'd call those a separate issue, really.  The error messages could indeed be improved in those cases to just tell people to use ($). if that is what they meant... but that would be a different bug... no need to get the two issues mixed up.

Note that self.thing and $.thing are syntactically different, container-wise and iter-wise:

$ perl6 -e 'class A { has $.a = (1,2,3); method foo { .perl.say for self.a; .perl.say for $.a } }; A.new.foo;'
1
2
3
$(1, 2, 3)

zoffixznet:
>The difference here is that the above examples semantically refer to an anon state variable, whereas $.^thing is a meta-attribute access

To my eye, `.` and `.?` methodops are much tighter semantically than `.` and `.^`, so I'm failing to see the logic why `.?` should be an error, while `.^` should work. 

> I'd call those a separate issue, really.

To me, looks like the same issue: "things that work with '`$` means `self`' shortcut"

skids:
OK, point taken... whatever is in the set of things defined as twigils is a twigil and is as tight as a name/ident.  For example $=pod is not an assignment operation on an anonymous variable.  (Also it isn't "$ means self", it's "$. / @. / %. / &. means accessor/methodcall on self" in particular.)

So I guess $.say should pretty much never be interpreted as a method call on an anonymous variable in the first place, even where there is no self.

In my view .^name is a syntactical justification to have a special namespace for meta things, and the $.^name form is actually useful given (in code that uses the meta model) you often want to work on self's meta, such functions often produce lists, and those lists can end up in expressions where containerization is needed.

So yes, I have to admit that doing this builds a case for also implementing $.? et allia.  But not every patch fixes all issues in a category, and to a certain extent, this is demand-based... someone else filed an RT about $.^ so I filled that need.  I disagree about developer torment and don't think we should close the door on those forms, though they are not an implementation priority without some user asking for them.  I don't think this burdens the learning curve... inconsistency burdens the learning curve (though it is sometimes necessary) just as much.

As far as compounds, I am not against .?^ (or should that be .^?)  being designated as "maybe we'll think about implementing this sometime down the road if it turns out to fill an idiom niche but no promises, since .^ was just sugar to begin with."

zoffixznet:
👍 Thanks!

PR #1179, opened 2017-10-04
Fix for RT#128054 and RT#126569 ... needs experienced review ...

skids:
Please see RT#128054 for elaboration.  This needs to be assessed
  by someone who knows a lot about what is going on in the QAST tree.


skids:
Integrated into Zoffix++ patch... no longer need this PR.

PR #1181, opened 2017-10-07
Suggest enum values as types

MasterDuke17:
Implements RT #123926, e.g., `enum E <Foo Bar>; sub x(Floo) {}` now says
`Invalid typename 'Floo' in parameter declaration. Did you mean 'Foo'?`


zoffixznet:
👍 Thanks!

PR #1183, opened 2017-10-07
“corresponding … was at” RT #125641

AlexDaniel:
Spectest clean.


AlexDaniel:
Maybe the error message should be printed on two lines.

PR #1185, opened 2017-10-07
Add tests to cover a common NC usage

jonathanstowe:
C functions that consistently take some struct or pointer
as the first argument are declared as methods of that
CStruct or CPointer


zoffixznet:
👍 Thanks!

PR #1188, opened 2017-10-08
Awesomify “malformed loop spec” errors

AlexDaniel:
This is a questionable fix for [RT #127857.](https://github.com/Raku/old-issue-tracker/issues/5220)

Makes each parse of a loop construct 0.04 ms slower. So if you have
one hundred ｢loop｣-s in your code, that's already about four
milliseconds slower.

Also breaks one 6.c-errata test that checks for “loop spec” text.

### DIFF
--- a/src/Perl6/Grammar.nqp
+++ b/src/Perl6/Grammar.nqp
@@ -1202,12 +1202,26 @@ grammar Perl6::Grammar is HLL::Grammar does STD {
     token statement_control:sym<loop> {
         <sym><.kok>
         :s''
-        [ '('
-            [
-            <e1=.EXPR>? ';' <e2=.EXPR>? ';' <e3=.EXPR>?
-            || <.malformed('loop spec')>
-            ]
-        ')' ]?
+        [
+          :my $exprs := 0;
+          '('
+          [     <e1=.EXPR>? {$exprs := 1 if $<e1>}
+          [ ';' <e2=.EXPR>? {$exprs := 2}
+          [ ';' <e3=.EXPR>? {$exprs := 3}
+          ]? ]? ]? # succeed anyway, this will leave us with a nice cursor
+          [
+          || <?{ $exprs == 3 }> ')'
+          || <?before ')'>
+             [
+             || <?{ $exprs == 0 }>
+                <.malformed("loop spec (expected 3 semicolon-separated expressions)")>
+             || <.malformed("loop spec (expected 3 semicolon-separated expressions but got {$exprs})")>
+             ]
+          || <?before ‘;’>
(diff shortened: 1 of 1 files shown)


AlexDaniel:
Looking for some thoughts on how to make this PR suck less.

AlexDaniel:
Some examples of error messages:

```perl6
loop (my $x = 0, $x < 10, $x++) {}
```
```
Malformed loop spec (expected 3 semicolon-separated expressions but got 1)
at -e:1
------> loop (my $x = 0, $x < 10, $x++⏏) {}
```
```perl6
loop () {}
```
```
Malformed loop spec (expected 3 semicolon-separated expressions)
at -e:1
------> loop (⏏) {}
    expecting any of:
        prefix
        term
```
```perl6
loop (my $x = 0; $x < 10; $x++;) {}
```
```
Malformed loop spec (expected 3 semicolon-separated expressions but got more)
at -e:1
------> loop (my $x = 0; $x < 10; $x++⏏;) {}
```

AlexDaniel:
Failing errata test in question: https://github.com/perl6/roast/blob/34577134e02213a3d303f70d307173d873699510/S32-exceptions/misc.t#L783-L784

ugexe:
IMO all the extra code (especially in a hot path) is not worth it just to give a slightly better error message.

AlexDaniel:
@ugexe What kind of hot path are we talking about? This is a compile-time thingy that happens once.

I see a little problem in the way it attempts to match three ‘;’-separated EXPRs. I'm thinking, why not ``$<e>=(<EXPR>?)+ % ';'``, or something like that? It would need a corresponding change in Actions file, but for some reason I couldn't really get it right this way. That said, if it did that, then maybe there'd be no overhead for the code blocks? Or something like that. I'm pretty sure there's a way to do it without introducing any performance hit.

AlexDaniel:
FWIW I measured the performance hit by creating a file with 5000 loops.

ugexe:
I misunderstood, thinking it applied to each iteration instead of each loop. Disregard.

JJ:
The conflict needs to be looked at, it's not directly solvable. Can you check?

AlexDaniel:
The conflict itself is not a big deal, this PR simply needs more work to make the change versioned.

AlexDaniel:
Ah, oops, I mistook this for another PR. I think this one can be merged, actually. I think I even had tests for this somewhere.

AlexDaniel:
OK, I don't have tests for this. FWIW the errata issue can be simply adjusted to test for /^loop spec/ instead.

AlexDaniel:
> The conflict needs to be looked at, it's not directly solvable. Can you check?

```git rebase --strategy-option patience master``` did it.

PR #1189, opened 2017-10-13
Correctly explain why couldn't instantiate a role

MasterDuke17:
This broke when Exceptions were changed to not create their message
unless .gist'ed.

Fixes RT #132285.


zoffixznet:
👍 Thanks!

PR #1191, opened 2017-10-13
optimize recently added AT-POS for typed pointers

dwarring:
Inline the add method and subsequent calls to nativecast and Int and add
methods. Improves performance by 4-5x.


zoffixznet:
👍 Thanks!

PR #1195, opened 2017-10-14
Optimize Regex multi for INTERPOLATE

MasterDuke17:
Move some code into a conditional and don't allocate some variables if
not needed.

`perl6 -e 'my $a = rx/aaaaaab/; "a" x 999999 ~ "b" ~~ /$a/; say now - INIT now'` used to report ~14.3s, now reports ~9.5s. This helps the performance regression spotted in https://rt.perl.org/Ticket/Display.html?id=132294, but still doesn't fix it completely.


AlexDaniel:
Thanks!

PR #1196, opened 2017-10-15
Simplify signatures for Regex and Associative...

MasterDuke17:
multis of INTERPOLATE.

`perl6 -e 'my $a = rx/aaaaaab/; "a" x 999999 ~ "b" ~~ /$a/; say now - INIT now'` used to report ~9.5s, now reports ~3.3s. This pretty much fixes the performance regression spotted in https://rt.perl.org/Ticket/Display.html?id=132294.


AlexDaniel:
Thanks!

PR #1197, opened 2017-10-15
Improve wrong arity error for anonymous subs

MasterDuke17:
If the routine doesn't have a name, use `'<anon>'` instead.

----------
(edit by @zoffixznet: wrap `'<anon>'` to code tags, so it does not get interpreted as HTML)


zoffixznet:
:+1: Thanks!

PR #1201, opened 2017-10-22
[jvm] Fix p6bindattrinvres for native attributes

usev6:
This was inspired by similiar code from nqp's Ops.java where we
call 'get_attribute_boxed' first and if that results in a
BadReferenceRuntimeException we try 'get_attributes_natives'.

Fixes RT #132128.

### DIFF
--- a/src/vm/jvm/runtime/org/perl6/rakudo/RakOps.java
+++ b/src/vm/jvm/runtime/org/perl6/rakudo/RakOps.java
@@ -10,6 +10,7 @@ import org.perl6.nqp.sixmodel.reprs.CallCaptureInstance;
 import org.perl6.nqp.sixmodel.reprs.ContextRefInstance;
 import org.perl6.nqp.sixmodel.reprs.NativeRefInstance;
 import org.perl6.nqp.sixmodel.reprs.VMArrayInstance;
+import org.perl6.nqp.sixmodel.reprs.P6OpaqueBaseInstance;
 
 /**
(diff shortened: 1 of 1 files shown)


usev6:
I'm not really sure, this is a good way to fix the issue. So it would be great to get a review.

usev6:
I thought about that, but unfortunately I don't know of a good way to determine if the attribute in question is native. (If I'm not mistaken tc.native_type is set anyway.) Scanning through the code I found two places where catching an exception is used in similar situations:
* https://github.com/perl6/nqp/blob/4494e704966b0309e07a129c34f6076870253562/src/vm/jvm/runtime/org/perl6/nqp/runtime/Ops.java#L2760
* https://github.com/perl6/nqp/blob/4494e704966b0309e07a129c34f6076870253562/src/vm/jvm/runtime/org/perl6/nqp/runtime/Ops.java#L6668

zoffixznet:
FWIW, I [removed](https://github.com/rakudo/rakudo/commit/e4a5bb17c9) the `Int.new` code that used p6bindattrinvres with native attrs. Looks like [it's not meant to work with natives on MoarVM either](https://irclog.perlgeek.de/perl6-dev/2017-10-23#i_15342138) and just worked with `bigint` for misterious reasons.

usev6:
Given that p6bindattrinvres was not supposed to work with native attributes (pmurias++, timotimo++) and that the Int.new problem is gone on JVM (Zoffix++), I'm closing this PR. It was an instructive thing to look at, nevertheless.

PR #1221, opened 2017-10-30
Normalize proto bodies

MasterDuke17:
Convert all `{ * }` to `{*}` for consistency.

Rakudo builds OK and passes `make m-test m-spectest`.

### DIFF
--- a/src/core/Any-iterable-methods.pm
+++ b/src/core/Any-iterable-methods.pm
@@ -8,7 +8,7 @@
 use MONKEY-TYPING;
 augment class Any {
 
-    proto method map(|) is nodal { * }
+    proto method map(|) is nodal {*}
     multi method map(Hash \h) {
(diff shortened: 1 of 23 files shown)


zhuomingliang:
I prefer `{ * }`. :)

ugexe:
`{*}` has the advantage that it means the same thing if you add scope to the sub like `proto sub foo(|) { {*} }`.

```
$ perl6 -e 'proto sub foo(|) { {*} }; multi sub foo { }; say foo(1)'
Cannot resolve caller foo(Int); none of these signatures match:
    ()
  in sub foo at -e line 1
  in block <unit> at -e line 1
```
```
$ perl6 -e 'proto sub foo(|) { { * } }; multi sub foo { }; say foo(1)'
*
```

zoffixznet:
👍 Thanks!

PR #1242, opened 2017-11-08
RFC: Use IO::Spec in the method resolve of IO::Path

ufobat:
This PR is supposed to fix that resolve did not work correclty on Windows.

C:\Users\martin>perl6 -e "say $*CWD; say $*CWD.resolve; say $*CWD.resolve.e"
"C:\Users\martin".IO
"\C:\Users\martin".IO
False

I'was not able to test that on a windows box, but it seems not to break those command on my linux box.

### DIFF
--- a/src/core/IO/Path.pm
+++ b/src/core/IO/Path.pm
@@ -277,13 +277,14 @@ my class IO::Path is Cool does IO {
             if nqp::iseq_s($part, $up) {
                 next unless $res-list;
                 nqp::pop_s($res-list);
-                $resolved = $res-list ?? $sep ~ nqp::join($sep, $res-list)
-                                      !! $empty;
+                $resolved = $res-list
+                    ?? $!SPEC.catdir(|nqp::p6bindattrinvres((), List, '$!reified', $res-list))
+                    !! $empty;
                 next;
             }
 
             # Normal part, set as next path to test
-            my str $next = nqp::concat($resolved, nqp::concat($sep, $part));
+            my str $next = $!SPEC.catdir($resolved, $part);
 
             # Path part doesn't exist...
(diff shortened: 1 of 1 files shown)


src/core/IO/Path.pm
```diff
@@ -277,13 +277,13 @@ my class IO::Path is Cool does IO {
             if nqp::iseq_s($part, $up) {
                 next unless $res-list;
                 nqp::pop_s($res-list);
-                $resolved = $res-list ?? $sep ~ nqp::join($sep, $res-list)
+                $resolved = $res-list ?? $!SPEC.catdir($res-list.flat)
                                       !! $empty;
```
zoffixznet: Are you sure that `.flat` is needed here? Doesn't look like the list is getting any unflat stuff added into it anywhere

ufobat: I think i need to pass something that matches the signature of catdir. The nqp::list_s() is not. I've copied this code from somewhere else where a List is needed. 

it should work now?

zoffixznet:
Thanks. I built this (on Linux) and unfortunately, seems there are several issues:

1) This patch made .resolve method 1.8x slower, even when using Unix spec.
2) It doesn't actually fix the method for Windows (still has leading slash):

```
    $ ./perl6 -e 'my $*SPEC = IO::Spec::Win32; "C:/tmp/tmp.O4XZSUYPo1/rakudo".IO.resolve.say'
    "\C:\tmp\tmp.O4XZSUYPo1\rakudo".IO
```
3) There are failures in `t/spec/S32-io/io-path.t` spectest that causes it to crash without completing. You can run the spectest with `make spectest`

PR #1246, opened 2017-11-11
RFC: IO::Path.resolve for windows 2nd attempt

ufobat:
The first of the 2 commits actually do break spec tests. I've added some details to the commit message as well.

In my opinion the spec test could be wrong, that's why I am still sending this PR. I am assuming that when you work on a `:volume<C:>` and you create a path with this volume that you must create the \ ( of C:**\\**$whatever ).

This is the result of resolve with IO::Spec::Win32 Spec:
> ./perl6 -e '$*SPEC = IO::Spec::Win32; "C:\\foo\\bar".IO.resolve.say'
"C:\foo\bar".IO

### DIFF
--- a/src/core/IO/Path.pm
+++ b/src/core/IO/Path.pm
@@ -232,7 +232,6 @@ my class IO::Path is Cool does IO {
         self.bless(:path($!SPEC.canonpath($!path)), :$!SPEC, :$!CWD);
     }
     method resolve (IO::Path:D: :$completely) {
-        # XXXX: Not portable yet; assumes POSIX semantics
         my int $max-depth = 256;
         my str $sep       = $!SPEC.dir-sep;
         my str $cur       = $!SPEC.curdir;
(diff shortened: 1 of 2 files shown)


zoffixznet:
> the spec test could be wrong

No, the test is correct. The presence of the volume does *not* imply an absolute path:

```bash
F:\Temp\test>perl6 -e "'F:/Temp/test/foo'.IO.spurt: 'test on F drive'"

F:\Temp\test>perl6 -e "'C:/Temp/test/foo'.IO.spurt: 'test on C drive'"

F:\Temp\test>type F:foo
test on F drive
F:\Temp\test>type C:foo
test on C drive
F:\Temp\test>type C:\foo
The system cannot find the file specified.

F:\Temp\test>type F:\foo
The system cannot find the file specified.

F:\Temp\test>
```

ufobat:
I am assuming that `C:foo` is  the $cwd but on a different volume? 
Stay tuned for a 3rd PR ...

zoffixznet:
> I am assuming that C:foo is the $cwd but on a different volume?

That's what my tests suggest; I didn't look up any proper info on how it works on Windows.

PR #1267, opened 2017-11-24
Make Supply.zip() 'done' when the first supply 'done'

FCO:


jnthn:
Thanks; do you plan to write tests also, or shall I?

FCO:
Ill try to do that when I come home...

FCO:
I was trying to write those tests, but couldn't find good way to do that... I have to go to sleep, but Ill try to continue tomorrow... but if someone want to do that, thats ok for me...

PR #1271, opened 2017-11-25
Fix Supply.throttle()

FCO:


lizmat:
Could you elaborate as to why this needed fixing?  Reversing the meaning of something without any other comment than "fixing it", feels a bit meagre to me.  What was broken exactly?

FCO:
Sorry about that!
The supply generated by throttle wasn't being done when the source supply become done.
This (https://github.com/perl6/roast/pull/359) tests that...

PR #1272, opened 2017-11-25
Pass typenames to two compile time exceptions

usev6:
The typenames are the only things needed for the error messages
in X::Inheritance::Unsupported and X::Composition::NotComposable.

With this change it's no longer necessary to use $! twigils when
generating the error messages (cmp. commit a822bcf9cf).

Also this makes two failing tests for RT #129906 (StackOverflowError,
probably caused by failed typechecks for the parameters in question)
in integration/error-reporting.t pass again on the JVM backend.

### DIFF
--- a/src/core/Exception.pm
+++ b/src/core/Exception.pm
@@ -2161,9 +2161,9 @@ my class X::ControlFlow::Return is X::ControlFlow {
 
 my class X::Composition::NotComposable does X::Comp {
     has $.target-name;
-    has $.composer;
+    has $.composer-name;
     method message() {
-        $!composer.^name ~ " is not composable, so $!target-name cannot compose it";
+        $.composer-name ~ " is not composable, so $.target-name cannot compose it";
     }
 }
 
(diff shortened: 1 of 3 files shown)


usev6:
Unfortunately this change is not spectest clean. We have one test for the method 'composer' of X::Composition::NotComposable.

https://github.com/perl6/roast/blob/00e4de362c4dbec6f679bacf33a26b3b76bed68e/S32-exceptions/misc.t#L335

I thought, I'd open a PR nonetheless -- at least to get some feedback.

moritz:
I'm opposed to this change.

Generating the error message is not the only purpose of the exception object; other use cases might catch these exceptions, and do something else with the objects. We don't know what they might want to do with the exceptions, so we shouldn't artificially limit their options.

For a user it's much easier to go from an type object to the name than the other way round, because it doesn't run in the scope that threw the exception. So this change makes life harder for potential users, with little upsides.

When I introduced typed exceptions in Rakudo, I tried to make them as rich as possible; without this richness, we could go back to throwing just strings as exceptions (ok, slight hyperbole here, but I hope you get the general point).

lizmat:
I concur with moritz on this.  Although I must admit that perhaps we may need to look at all our exceptions more thoroughly to find cases where we actually are expecting type names rather than types.

usev6:
@moritz Thanks for your feedback. Your explanation is convincing (and actually I had a similiar thought "there's more than just .message" before I fell asleep yesterday :-).

I'll try to find the underlying reason why the JVM explodes badly running the tests for RT #129906. (And maybe I'll also understand why we need a822bcf9cf to pass the tests on MoarVM.)

PR #1273, opened 2017-11-26
Fixes #1258

### LINKED ISSUE #1258: Supply.list returns a Seq
`Supply.list` would make a really good (better than inheriting from `Any`) implementation of `Supply.Seq`. Actually, we should just rename that method to `Seq`, since:

    $ perl6-m -e 'say Supply.interval(2).list.WHAT'
    (Seq)

A `list` method returning a `Seq` is less than great, though. I suggest after renaming the current `list` to `Seq`, we then implement `Supply.list` as `self.Seq.list`.

Compiler version (`perl6 -v`): This is Rakudo version 2017.09-479-g7cf5ce7 built on MoarVM version 2017.09.1-614-g1952356 implementing Perl 6.c.

FCO:
Fixes #1258 renaming the Supply.list to Supply.Seq and creating Supply.list that calls Supply.Seq.list

### DIFF
--- a/src/core/Supply.pm
+++ b/src/core/Supply.pm
@@ -671,7 +671,7 @@ my class Supply does Awaitable {
 
     my class ConcQueue is repr('ConcBlockingQueue') { }
     multi method list(Supply:D:) {
-        $.Seq.list
+        self.Seq.list
     }
     method Seq(Supply:D:) {
         gather {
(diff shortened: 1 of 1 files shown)


lizmat:
A little style issue: personally I would have written "self.Seq.list" rather than "$.Seq.list".  Also, the $.foo notation is really intended for attributes, although they also work on methods.  Using it for methods actually introduces a silent ambiguity:

`class A { has $.Seq; method a() { say $.Seq.^name } }; A.new(Seq => 42).a    # Int`

FCO:
Fixed.

lizmat:
Thanks!

Issue #1238, opened 2017-11-06
but-ing a Hash only works the last time ( (%(:42foo) but A) but B )


AlexDaniel:
Code:
```perl6
role Foo { has $.a };
role Bar { has $.b };
say ((%(:66x, :67y) but Foo(5)) but Bar(7)).a
```

Result (2015.12,2016.01.1,2016.02):
```
5
```
Result (2016.03…2017.10,HEAD(ab03b54)):
```
Any
```

<br>

Bisectable points at (2016-03-14) https://github.com/rakudo/rakudo/commit/6b2b34a79150f6e9a275fd39eb5dd96519d6179f

``.b`` always returns 7 (which is right) no matter what version of rakudo you use.

ab5tract:
Returns `(Any)` for base and RakuAST:

```
> set-base; ./rakudo-m issues/r1238.raku
(Any)

> set-rakuast; ./rakudo-m issues/r1238.raku
(Any)

> ./rakudo-m -v
Welcome to Rakudo™ v2024.04-92-g24057e69c.
Implementing the Raku® Programming Language v6.d.
Built on MoarVM version 2024.04-6-g84121f0de.
```

lizmat:
Oddly `Map.clone` appears to be reduced to `self` now.  Not sure whether that is actually correct, as a `Map` may contain mutable elements, such as a `Hash`.

lizmat:
The problem appears to be caused by the fact that the LHS of `but` is getting (re-)instantiated (which it probably shouldn't?)
```raku
role Foo {
    has $.a;
    method TWEAK() { dd }
}

role Bar {
    has $.b;
    method TWEAK() { dd }
}
say ((%(:66x, :67y) but Foo(5)) but Bar(7)).b;
```
```
method TWEAK(Hash+{Foo} $:: *%_)
method TWEAK(Hash+{Foo}+{Bar} $:: *%_)
7
```

lizmat:
Looks like 8f440ea6d4 is to blame, although the old state is a bit iffy as well.

lizmat:
Turns out, the same issue exists for arrays:
```raku
role Foo { has $.a };
role Bar { has $.b };
say ((my @a but Foo(5)) but Bar(7)).a;  #  (Any)
```

lizmat:
As do `Date` and `DateTime`.

lizmat:
Tests added with https://github.com/Raku/roast/commit/dd3202f302

PR #1283, opened 2017-11-30
Issue 1238: Fix Hash clone not retaining all attributes

jstuder-gh:
Addresses [Issue 1238](https://github.com/rakudo/rakudo/issues/1238).

This issue was occurring due to the an operation that takes place in the
Hash 'but' method. A cloning of the Hash obj occurs but not all the
attributes are being replicated in the clone.

Mirroring (almost) the logic of the Mu clone method so that all
attributes are replicated on the new clone obj, not just $!storage.
However, unlike in Mu, not cloning the $!storage Scalar container (the
.VAR on the attribute) to avoid causing a regression for the issue fixed
in commit 6b2b34a79150f6e9a275fd39eb5dd96519d6179f.


jstuder-gh:
Requires tests be added to the roast, if accepted.

lizmat:
Thanks for looking into this!

`clone` is supposed a fast, low level way of cloning an object.  By copying the logic from `Mu`, you're essentially ignoring all of the knowledge we have of the `Hash` object, namely that it also has a `$!descriptor`.  And performance wise, there is now no difference between having this method, and not having this method (aka, removing `Hash.clone` altogether).

I think it should be sufficient to make sure the same `$!descriptor` is used in the cloned `Hash` object.  Please adjust your pull request accordingly, and I'll make sure it will get merged.

jstuder-gh:
I tried just binding the `$!storage` and `$!descriptor` attributes only, but unfortunately the result when calling the `.a` accessor was still `(Any)`, (given the example included in the issue).

I suppose we could query the roles that are mixed into the object and bind the non-standard attributes to the new object. That would make it so that we don't have to bind all the attributes.

Ignore the high-level Perl 6, I'm just thinking in terms of the end result. I'd ultimately implement in a lower-level nqp style, of course.

```
nqp::bindattr($cloned, Map,  '$!storage',    nqp::clone( nqp::getattr(self, Map, '$!storage')) );
nqp::bindattr($cloned, Hash, '$!descriptor', nqp::clone( nqp::getattr(self, Hash, '$!descriptor')) );

my @attrs = self.^roles.grep( * ~~ none Associative, Iterable )
                .map( *.^attributes.unique.Slip );

for @attrs -> $attr {
    unless nqp::objprimspec($attr.type) {
        my $name     := $attr.name;
        my $package  := $attr.package;
        if nqp::attrinited(self, $package, $name) {
            my $attr_val := nqp::getattr($cloned, $package, $name);
            nqp::bindattr($cloned, $package, $name, nqp::clone($attr_val));
        }
    }
}
```

Although, I suspect that after iterating through the roles then through the attributes, it wouldn't really be any faster than iterating through the attributes on the initial object and binding them on the new (as in the Mu implementation).

What do you think?

jstuder-gh:
I realize now that I was coming at this from the wrong angle before.

I had also committed it on top of the master branch. I had forgotten this fact so when I force pushed the real master to my fork it wiped the ref to the of the old commit on Github, so why keep this open.

I might return to this issue again at some point in the future, and at that time I will open another PR.

Issue #1282, opened 2017-11-30
[POD] need to handle table cells with char column separators as data


tbrowder:
example pod table:

    =begin table
    + | plus
    | | pipe
    =end table

The leading pipe char and plus char data elements are parsed as column separators and should be escaped to avoid such interpretation.  The current table pod handling doesn’t handle such escaping.

tbrowder:
I have a fix I will soon introduce as a PR (in both rakudo and roast).

PR #1287, opened 2017-11-30
add code to fix rakudo repo issue #1282

tbrowder:
allows escaping visual column separators

new tests are added in roast PR #361

### DIFF
--- a/src/Perl6/Pod.nqp
+++ b/src/Perl6/Pod.nqp
@@ -334,15 +334,15 @@ class Perl6::Pod {
         # and converts it into a twine data structure. primarily
         # designed for a table cell.
         # note all Z<> comments have already been removed.
-	my $s   := $S; # the raw string to be converted to a twine
-	my $ret := ''; # the cleaned string (i.e., without comments)
-	my $idx := nqp::index($s, '<'); # find the possible beginning of a formatting code
+        my $s   := $S; # the raw string to be converted to a twine
+        my $ret := ''; # the cleaned string (i.e., without comments)
+        my $idx := nqp::index($s, '<'); # find the possible beginning of a formatting code
         if $idx < 0 {
             # no '<' found so return the simple string
             return $s;
         }
 
-	while $idx > -1 {
+        while $idx > -1 {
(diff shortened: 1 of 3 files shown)


src/Perl6/Pod.nqp
```diff
@@ -327,6 +327,39 @@ class Perl6::Pod {
...
+	my $ret := ''; # the cleaned string (i.e., without comments)
+	my $idx := nqp::index($s, '<'); # find the possible beginning of a formatting code
+        if $idx < 0 {
+            # no '<' found so return the simple string
+            return $s;
+        }
+
+	while $idx > -1 {
+            my $idx2 := nqp::index($s, '>', $idx+2); # and the end
+            nqp::die("FATAL:  Non-existent closing '>' for inline pod comment in string '$s' in Table $table_num") if $idx2 < 0;
```
samcv: Maybe: "Couldn't find terminator '>' for inline pod comment..." so it matches closer to the wording we use for other error messages.

src/Perl6/Pod.nqp
```diff
@@ -327,6 +327,39 @@ class Perl6::Pod {
...
+            $ret := nqp::concat($ret, $s0);
+
+            # throw away the orig string up to the end of the found comment
+            $s := nqp::substr($s, $idx2+1); # the trailing chunk
+            # look for another comment in the remaining string
+            $idx := nqp::index($s, 'Z<');
+	}
+
+	# make sure we use up a non-empty string end
+	if $s {
```
samcv: Could maybe be this to be more compact:
```perl6
 $ret := nqp::concat($ret, $s) if $s;
```

src/Perl6/Pod.nqp
tbrowder: ok, will change...

src/Perl6/Pod.nqp
tbrowder: yep...

tbrowder:
Thanks for the review, approval, and merge.

The twine data structure is a weird array interleaving text and Pod::FormattingCode objects.  It's not well documented.

Issue #1268, opened 2017-11-24
Lazy slicing should not stop on the first hole, or something (say @a[lazy 3..*])


AlexDaniel:
After looking at https://github.com/perl6/doc/pull/1681 I noticed this:
```perl6
my @letters = <a b c d e f>;
@letters[4]:delete;
say @letters[3..*]
say @letters[lazy 3..*]
```

Result (2015.09⌁2017.09,[``456358e3^``](https://github.com/rakudo/rakudo/commit/456358e3c380eeeb5fe5bc6260ba6f51c42a52ff^)):
```
(d)
(d)
```

Result (https://github.com/rakudo/rakudo/commit/456358e3c380eeeb5fe5bc6260ba6f51c42a52ff, 2017.10,2017.11,HEAD(3166400)):
```
(d (Any) f)
(d)
```

I don't understand why it should stop at the first hole. The behavior changed after https://github.com/rakudo/rakudo/commit/456358e3c380eeeb5fe5bc6260ba6f51c42a52ff, and that seems to be unintentional. I'm really confused as to what should be the right behavior, but my gut feeling is that ``(d (Any) f)`` is the right answer.

AlexDaniel:
ping @lizmat @zoffixznet

zoffixznet:
> The behavior changed after 456358e, and that seems to be unintentional

It was intentional. The change was anticipated and I used it as the argument against the optimization, but you said the stopping-at-hole was a bug (there's probably an RT for that).

I agree, `(d (Any) f)` is what I'd expect to be shown, but don't know of a good way to implement that because IIRC holes are currently just `nqp::null`s; just as items past end are.

(IMO the whole holes-in-arrays business should not exist at all and `:delete` should not work on Arrays; but I guess that ship's sailed)

AlexDaniel:
Ah, I should've read more as I'm starting to forget things. [RT #127573](https://rt.perl.org/Ticket/Display.html?id=127573) seems to be identical to this ticket.

jnthn:
The tricking thing, I think, is that the various slicing operations are implemented in terms of methods on the target to be indexed, such that implementing the various methods is enough to get all of the slicing behavior. We can't just call `.elems`, because the target might well be lazy/infinite. Stopping at a position that doesn't exist, checked by `EXISTS-POS`, is usually fine - it's just `:delete`, which is rare, that can cause a surprise.

I don't know of any alternative options available with the current API. So unless I've missed one, then I think our options are:

[...]

PR #1299, opened 2017-12-08
Issue 1268: Fix lazy slicing stopping at the first deleted

jstuder-gh:
This commit fixes the issue in which using lazy Iterables as Positional
indices will evaluate until the first deleted (undefined) value but not
continue further.

Within the default eagerize routine, it checks the number of reified
elements upon encountering an undefined value to determine whether to
continue.

Addresses [Issue #1268](https://github.com/rakudo/rakudo/issues/1268)

### DIFF
--- a/src/core/List.pm
+++ b/src/core/List.pm
@@ -477,16 +477,6 @@ my class List does Iterable does Positional { # declared in BOOTSTRAP
         )
     }
 
-    method NUM-REIFIED(List:D:) {
-        nqp::stmts(
-            nqp::if(
-                nqp::defined($!todo) && nqp::not_i( $!todo.fully-reified ),
-                $!todo.reify-until-lazy
-            ),
-            nqp::elems($!reified)
-        )
-    }
-
     multi method AT-POS(List:D: int $pos) is raw {
         nqp::if(
           nqp::isge_i($pos,0) && nqp::isconcrete($!reified),
--- a/src/core/List.pm
+++ b/src/core/List.pm
@@ -477,7 +477,15 @@ my class List does Iterable does Positional { # declared in BOOTSTRAP
         )
     }
 
-    method NUM-REIFIED(List:D:) { nqp::elems($!reified) }
+    method NUM-REIFIED(List:D:) {
+        nqp::stmts(
+            nqp::if(
+                nqp::defined($!todo) && nqp::not_i( $!todo.fully-reified ),
+                $!todo.reify-until-lazy
+            ),
+            nqp::elems($!reified)
+        )
+    }
 
     multi method AT-POS(List:D: int $pos) is raw {
         nqp::if(
--- a/src/core/List.pm
+++ b/src/core/List.pm
@@ -477,6 +477,8 @@ my class List does Iterable does Positional { # declared in BOOTSTRAP
         )
     }
 
+    method NUM-REIFIED(List:D:) { nqp::elems($!reified) }
+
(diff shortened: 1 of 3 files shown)


jstuder-gh:
Will need to write tests if accepted.

zoffixznet:
Is there a way we can we do this without adding a public method?

I also don't understand how this fixes anything for Seqs, since reified part will often just contain whatever was already reified:

```
<Zoffix__> m: with [<a b c d e>] { .[0]:delete; with .Seq { use nqp; nqp::elems( nqp::getattr(.cache, List, '$!reified') ).say } }
<camelia> rakudo-moar 1dbf5f589: OUTPUT: «0␤»
```

And so, IMO this type of fix just muddles up the conditions where a hole would be stopped at. "Stops at first gap or end of Iterable" is easy to understand. But if you also have to care about the kind of iterable you, as a user, have you now have 4 different cases to consider instead of just 2.

IMO we should either fix this entirely or not fix it at all (and document behaviour as described in the [Issue](https://github.com/rakudo/rakudo/issues/1268#issuecomment-347023393)). Doing half-way measures doesn't improve usability or usefulness of the feature.

jstuder-gh:
I added a commit that addresses the issue of pending, unreified elements on the List or Seq as showcased in your example by explicitly reifying until lazy if the Reifier indicates that there is more to do.

As for the public methods, I suppose we could encapsulate all of this logic within the eagerize block. I had just figured this was a bit cleaner. But if the number of reified elements is not something we want to be exposing to the user, I can get rid of them.

zoffixznet:
> But if the number of reified elements is not something we want to be exposing to the user

No we shouldn't, especially not if calling the method *causes*, often complete, reification.

------

BTW is the `Seq` handling needed at all? Looking more at it now and it looks like the hole is filled when an Array is converted to Seq/List and this issue only exists with Arrays, since you can't `:delete` from Seqs/Lists directly:

```
<Zoffix__> m: my @a2 = 4, 5, 6; @a2[1]:delete; say @a2.Seq[lazy 0, 1, 2, 3, 4, 5, 6, 7]
<camelia> rakudo-moar 1dbf5f589: OUTPUT: «(4 (Any) 6)␤»

<Zoffix__> m: my @a2 = 4, 5, 6; @a2[1]:delete; say @a2.List[lazy 0, 1, 2, 3, 4, 5, 6, 7]
<camelia> rakudo-moar 1dbf5f589: OUTPUT: «(4 Nil 6)␤»
```

jstuder-gh:
> No we shouldn't, especially not if calling the method causes, often complete, reification.

Good point. I'll remove the methods and incorporate the logic into the eagerizer in the next iteration. I'll also try removing the Seq handling as well.

jstuder-gh:
As discussed, I've removed the public methods and incorporated the logic into the default "eagerize" block for POSITIONS. It also passes the spectest.

zoffixznet:
I don't see anyone else commenting, but I'm still against this sort of piece-meal fixes. All it's doing is making the behaviour hard to predict.

(the latest version of this PR assumes anything that isn't a Seq is a List, which will make this codepath explode for userland Positionals)

jstuder-gh:
I'm closing this PR as it's clear now that this was not the right path to take in order to fix this problem. There is still discussion at [Issue 1268](https://github.com/rakudo/rakudo/issues/1268) about another approach. If that proves to be the right way forward, I will open another PR at a later date.

PR #1307, opened 2017-12-12
Drop extraneous spaces in code and table blocks

softmoth:
code2text() was missing a .join call, so inserted extra spaces between
chunks (e.g. around embedded formatting codes).

table2text() was padding the final column in each row, which is not
desirable in a plain text format.

### DIFF
--- a/lib/Pod/To/Text.pm6
+++ b/lib/Pod/To/Text.pm6
@@ -36,7 +36,7 @@ sub heading2text($pod) {
 }
 
 sub code2text($pod) {
-    "    " ~ $pod.contents>>.&pod2text.subst(/\n/, "\n    ", :g)
+    $pod.contents>>.&pod2text.join.indent(4)
 }
 
 sub item2text($pod) {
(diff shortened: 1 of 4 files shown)


softmoth:
There are no spec tests for Pod::To::Text. However, I have run this over all the pod in p6doc/doc, and verified that it looks good.

AlexDaniel:
AppVeyor failures are due to another issue (can be ignored).

👍 on the PR, looks good. But can we have some tests for the fixed issues? I think that's important because eventually POD stuff may be refactored, so it'd be sad to lose some of the fixes just because there are no tests.

softmoth:
Alex, thank you for the review. I've added a new t/07-pod-to-text/01-whitespace.t file.

softmoth:
I am pretty sure they don't belong in roast. Initially I put the file in spec/MISC, but it feels wrong to put in roast as the output isn't specified and to me it shouldn't be part of the language definition.

If that's not consensus then I'm okay putting it anywhere, just let me know where it belongs.

AlexDaniel:
We can move it later.

PR #1316, opened 2017-12-16
Add tests for testing almost all Unicode codepoints

samcv:
Tests all codepoints names (except for Hangul). Tests Line_Break and
General_Category. Also tests General_Category for many reserved characters
as well.


zoffixznet:
Eyeballing the code, looks like these would take quite some time to run, no? Perhaps these should be in `make xtest` (doesn't yet exist) instead of regular `make test`. Some `t/05-message/` tests could be moved to `xtest` as well.

zoffixznet:
> samcv │ takes ~14s to run

Saw that in the chat. 14s ain't that bad and can just go in and be moved to `xtest` whenever it's implemented.

AlexDaniel:
That's what I thought too. 14s is a bit too much to add to the regular spectest.

By the way: https://imgur.com/a/xhvrR (data gathered by grepping ZOFVM lines from clog).

samcv:
It is 14s if you don't use any threading fyi

AlexDaniel:
What's the current status of this?

samcv:
@AlexDaniel I fixed the merge conflicts and I'm going to merge this now.

PR #1317, opened 2017-12-17
Add two special cases for infix:<,>

MasterDuke17:
Since these are likely to be common cases, give them a fast path.

`my $a; my $b; ($a,$b) = "a","b" for ^1000000` decreases from ~5.4s to ~4.9s.

### DIFF
--- a/src/core/List.pm6
+++ b/src/core/List.pm6
@@ -1492,6 +1492,12 @@ my class List does Iterable does Positional { # declared in BOOTSTRAP
 # The , operator produces a List.
 proto sub infix:<,>(|) is pure {*}
 multi sub infix:<,>() { nqp::create(List) }
+multi sub infix:<,>(Int \a, Int \b) is default {
+    nqp::p6bindattrinvres(nqp::create(List),List,'$!reified',nqp::list(a,b))
+}
+multi sub infix:<,>(Str \a, Str \b) {
+    nqp::p6bindattrinvres(nqp::create(List),List,'$!reified',nqp::list(a,b))
+}
 multi sub infix:<,>(|) {
 
     # look for a Slip in the parameters


AlexDaniel:
After the release plz. 👍

lizmat:
Why the difference between `Int` / `Str`?  Or does a generic `infix:<,>($,$)` not JIT very well ?

zoffixznet:
> Why the difference between Int / Str? Or does a generic infix:<,>($,$)

Generic `infix:<,>($,$)` would require [handling of `Slip`s](https://github.com/rakudo/rakudo/blob/4fca94743/src/core/List.pm#L1437-L1438).

MasterDuke17:
Hm, I wanted a very small/simple fast path so that it'd inline, but maybe something like this would be small enough, still faster than the completely generic case, but cover more instances than the `Str $, Str $` or `Int $, Int $` cases?

```
multi sub infix:<,>($, $) {
    my \in := nqp::p6argvmarray();
    nqp::istype(nqp::atpos(in,0),Slip) || nqp::istype(nqp::atpos(in,1),Slip)
      ?? nextsame
      !! nqp::p6bindattrinvres(nqp::create(List),List,'$!reified',in)
}
```

zoffixznet:
The updated version makes `&infix:<,>(1, Empty)` 12.78x slower, while making `&infix:<,>(1, 1)` 25% faster.

I think constructs like `my @a = 1, |@b` are common enough to make this PR overall detrimental to performance.

MasterDuke17:
I reverted to just special casing `Int` and `Str`. Now `my $a; my $b; ($a,$b) = "a","b" for ^1000000` decreases from ~4.2s to ~3.6s

zoffixznet:
> Since these are likely to be common cases

I'm wondering if that's really true. Considering `($a, $b, $c) = $x, $y, $z` would not go through the added candidates. So seems like we're adding extra candidates to handle a slim sliver of possibilities, while possibly slightly impacting negatively everything else.

> my $a; my $b; ($a,$b) = "a","b" for ^1000000

It is faster, but that doesn't quite represent what I think you think you're measuring:

```perl6
my $a; my $b; ($a,$b) = "a","b"
````

That ^ won't go through the new candidates. The `"a","b"` gets constant-folded during optimize stage and the `($a,$b)` don't go through the new candidates because they don't yet have `Str` values in them. In your bench, the observed speedup is due to `($a,$b)` containing `Str` objects and thus hitting the extra candidates on anything but the first iteration of the loop.

------

I didn't have much luck measuring it, but I think there's a negative impact in this PR because we currently don't cache candidate lookup for slurpy candidates. This means the more candidates we have, the slower dispatch would be to slurpy candidates. For example, by avoiding the slurpy candidate caching issue in the opt for `dispatch:<var>` [showed](https://github.com/rakudo/rakudo/commit/ff31f0a30d331724e93f2c2450fd05d093390f65) a 2.6x boost to a 1-candidate call, but 43x boost to a 13-candidate call.

Here, we're going from 2 candidates to 4. I don't see any difference with the bench below but is that because the dispatch-to-slurpy issue isn't affected by extra candidates in all cases or because the slowdown drowns in the noise and would actually be visible in real-life programs?

```perl6
use nqp;
my $z = rand;
my int $x;
{
    nqp::while(
      nqp::isgt_i(4000_000, ($x = nqp::add_i($x, 1))),
      nqp::callstatic(&infix:<,>, $z, $z));
    say now - ENTER now;
}
```

------

I'm +⅓ on merging this—it feels like here we're optimizing by shooting with our eyes closed and hoping we hit the target and not someone's foot.

MasterDuke17:
FWIW, there are just shy of 2400 uses of something like `my ($a, $b) = ` in the ecosystem, see https://gist.github.com/Whateverable/1848808edebb31ab442fecf8bcadaeda.

zoffixznet:
> FWIW, there are just shy of 2400 uses of something like my ($a, $b) = in the ecosystem, see https://gist.github.com/Whateverable/1848808edebb31ab442fecf8bcadaeda.

`my ($a, $b) =` doesn't go through the new candidates, though, does it?

It'd need to either be `my Int ($a, $b) = ...` or `my Str ($a, $b) = ...` or `($a, $b) = ...` where both `$a` and `$b` have `Int` or `Str` objects. Going through that list, I count 62 cases that'd hit the new candidates. 

So 2.5% of cases get 37% faster while 97.5% of cases possibly get immeasurably slower (say 0.1% slower ). I think that's a win overall, yeah?

I see Geo::Ellipsoid has 1340 cases that'd hit `(Rat, Rat)` candidate if it existed. Should we add that too?

MasterDuke17:
I just checked, `my ($a, $b) = .words[1,3] for "50k.sql".IO.lines` does hit the `Str,Str` version. I do agree that this is perhaps less than 100% optimal. I'd really rather make the generic case faster, but I tried a couple different things and couldn't get any speedup.

MasterDuke17:
`infix:<,>` isn't even the most expensive routine, but I couldn't find anything to do with the ones ahead of it (included below).

Name | Entries
-- | --
STORE               SETTING::src/core/List.pm6:741 | 1000000
iterator               SETTING::src/core/List.pm6:583 | 2000000
ReifiedList               SETTING::src/core/Rakudo/Iterator.pm6:2533 | 3996070

zoffixznet:
The `my ($a, $b) = .words[1,3] for "50k.sql".IO.lines` is equivalent to this though:

```perl6
my ($a, $b);
($a, $b) = .words[1,3] for "50k.sql".IO.lines
```

So the `Str,Str` candidate is hit only because `$a` and `$b` are set to `Str` objects after the first iteration.

zoffixznet:
👍 Thanks!

AlexDaniel:
FWIW, I think there's only one other use of `is default` routine trait in core:

https://github.com/rakudo/rakudo/blob/65874b15552520a512b67073ec7becf997468a9f/src/core/operators.pm6#L58-L63

MasterDuke17:
There's also some in native_array.pm6

MasterDuke17:
The `is default` is because of allomorphs. If there was a separate `IntStr,IntStr` candidate it wouldn't be needed obviously, but I didn't think that was common enough to need a special case.

PR #1321, opened 2017-12-19
Add argument to ambiguous call exception message

MasterDuke17:
We already had the capture of the call's argument(s) in the exception,
just add it to the message to help debugging.

### DIFF
--- a/src/core/Exception.pm
+++ b/src/core/Exception.pm
@@ -2526,8 +2526,41 @@ my class X::Multi::Ambiguous is Exception {
     has @.ambiguous;
     has $.capture;
     method message {
-        join "\n",
-            "Ambiguous call to '$.dispatcher.name()'; these signatures all match:",
+        my @bits;
+        my @priors;
+        if $.capture {
+            for $.capture.list {
+                try @bits.push(.WHAT.perl);
+                @bits.push($_.^name) if $!;
+                when Failure {
+                    @priors.push(" " ~ .mess);
+                }
+            }
+            for $.capture.hash {
+                if .value ~~ Failure {
+                    @priors.push(" " ~ .value.mess);
+                }
+                if .value ~~ Bool {
+                    @bits.push(':' ~ ('!' x !.value) ~ .key);
+                }
+                else {
+                    try @bits.push(":$(.key)\($(.value.WHAT.?perl))");
+                    @bits.push(':' ~ .value.^name) if $!;
+                }
+            }
+        }
+        else {
+            @bits.push('...');
+        }
+        if @.ambiguous[0].signature.gist ~~ /': '/ {
+            my $invocant = @bits.shift;
+            my $first = @bits ?? @bits.shift !! '';
+            @bits.unshift($invocant ~ ': ' ~ $first);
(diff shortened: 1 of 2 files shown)


lizmat:
do you intend to see this merged before or after release today?

zoffixznet:
Would `$.capture.perl()` have problems with these kinds of arguments?

1) `my @a = 1..*;` crash, because you can't call `.perl` on it
2) `my @b = 1...100000;` huge amount of output, because `.perl()` would spit out every element.

lizmat:
> On 19 Dec 2017, at 12:17, Zoffix Znet <notifications@github.com> wrote:
> 
> Would $.capture.perl() have problems with these kinds of arguments?
> 
> 	• my @a = 1..*; crash, because you can't call .perl on it
> 	• my @b = 1...100000; huge amount of output, because .perl() would spit out every element.

s/perl/gist/ ?

That would limit the output (generally).

MasterDuke17:
Updated. And no, I don't expect this to make the current release.

MasterDuke17:
FYI, `multi z (@a, Int) { say "here" }; multi z (@a, Str) { say "there" }; my @a = 1..*; z(@a, <1>)` now says:
```
Ambiguous call to 'z(Array, IntStr)'; these signatures all match:
:(@a, Int $)
:(@a, Str $)
```

zoffixznet:
Looking at latest version, feels like it's doing a lot of work that should probably be stuffed in `Capture.gist` instead. 

Also this:

```
try @bits.push($where ?? .perl !! .WHAT.perl );
```

Still has the [aforementioned](https://github.com/rakudo/rakudo/pull/1321#issuecomment-352719534) problem with producing too much output on large objects (e.g. `my @b = 1...100000;`)

MasterDuke17:
Not in the text of the error message, `multi z (@a, Int) { say "here" }; multi z (@a, Str) { say "there" }; my @a = 1..100000; z(@a, <1>)` also says:
```
Ambiguous call to 'z(Array, IntStr)'; these signatures all match:
:(@a, Int $)
:(@a, Str $)
```

zoffixznet:
> Not in the text of the error message, multi z (@a, Int) { say "here" }; multi z (@a, Str) { say "there" }; my @a = 1..100000; z(@a, <1>) also says:

That's because in that case the `$where` is still False (`my $where = so first { $_.signature.gist ~~ / where / }, @.ambiguous;`), so it goes through the `.WHAT.perl` and not the `.perl` branch. 

Though, when I tried to add a `where`, it goes through a different exception (X::Multi::NoMatch). Under which cases would `where` would be set in this exception?

MasterDuke17:
I think leaving that conditional in was a thinko. `@.ambiguous` could have signatures with `where` in them, but of course that wouldn't matter when printing the caller.

zoffixznet:
I tried to resolve conflicts but it didn't seem to work. Can you try?

MasterDuke17:
I resolved the conflict and then squashed the commits.

zoffixznet:
👍 Thanks!

PR #1335, opened 2017-12-23
Ensure Configure.pl uses gmake for builds on BSDs

Kaiepi:
BSD's make isn't entirely compatible with GNU's but still gets used on build anyways, unlike with Solaris. Since no version of make may be installed in the first place, and it might not be installed or symlinked as gmake on Linux, this checks if either name for it exists in $PATH and acts like GNU make


zoffixznet:
👍 Thanks!

PR #1339, opened 2017-12-24
Provides for fixes for several POD bugs:

tbrowder:
RT #124281 - colonpairs in POD config options always produce strings

RT #126742 - config items should not include quotes for string values

RT #130477 - Pod config parses colonpairs but simply stringifies
             whatever it matched

RT #132632 - List and hash configuration value formats are not yet
             implemented (NYI)

+ Improved string quote handling from a previous commit.

  - quotes are stripped in the proper step of the process

  - Q/q quotes are handled

+ List element and hash values are typed as expected.

+ Comments added to better explain the %config generation process.

+ The docs have been updated to reflect these changes.

+ Corresponding tests are in a PR to roast.


AlexDaniel:
(https://github.com/perl6/roast/pull/371)

src/Perl6/Pod.nqp
```diff
@@ -1,25 +1,33 @@
...
-    if nqp::existskey(%env, $ddenvvar) {
-        my $val := nqp::atkey(%env, $ddenvvar);
-        $debug := $val;
-    }
-    if nqp::existskey(%env, $duenvvar) {
-        my $val := nqp::atkey(%env, $duenvvar);
-        $udebug := $val;
-    }
+    # for users
+    my $udebug := nqp::ifnull(  nqp::atkey(  nqp::getenvhash(), 'RAKUDO_POD6_TABLE_DEBUG'  ), 0  );
```
AlexDaniel: I wonder, why is it “POD6” sometimes and just “POD” in other cases?

src/Perl6/Pod.nqp
```diff
@@ -116,32 +124,175 @@ class Perl6::Pod {
...
+        }
+        elsif 'q' eq nqp::lc($first) && %h2{$second} && $last eq %h2{$second} {
+            say("DEBUG: found Q quote pair for val |$val|") if $debugp;
+            my $first2 := nqp::concat($first, $second);
+            $val := subst($val, /^ $first2 /, q{});
+            $val := subst($val, / $last $/, q{});
+        }
+
+        if $debugp {
+            say("  leading character is   [{$m[0]}]");
```
AlexDaniel: Maybe that's too verbose? Squish it into one line perhaps?

src/Perl6/Pod.nqp
tbrowder: um, not paying attention, which do you suggest? my original thought for users it should be clear it’s not p5 pod. then when i started working on more than table pod i thought just POD is clear enough for contributers.

what should be used?

src/Perl6/Pod.nqp
tbrowder: ok...

src/Perl6/Pod.nqp
```diff
@@ -116,32 +106,281 @@ class Perl6::Pod {
...
+                my $n := nqp::chars($s);
+                # delete 1 char for leading +-
+                if $s ~~ / <[+-]> / { --$n }
+
+                # convert the int to a bigint if required
+                #   max val int (int32): +2_147_483_647 <= 10 digits
+                #   max val int (int64): +9_223_372_036_854_775_807 <= 19 digits
+                # we convert to bigint after 9 digits
+                # we convert to string after 18 digits
+                if $n > 18 {
```
AlexDaniel: There has to be a better way. Also, I'm assuming this is not going to work on 32-bit systems?

AlexDaniel: What do you mean by “too big for bigint”? I thought the point of bigints is to store unlimitedly huge ints :)

AlexDaniel:
Ping @LLFourn. This resolves a bunch of tickets and adds some tests. Would you agree that we should take this for now, with some hopes that the whole things is reworked later?

src/Perl6/Pod.nqp
tbrowder: well, my limited testing showed a limit of a 64-bit integer...

AlexDaniel: I guess the misunderstanding is that int64 is a bigint? It's not. A bigint is just an `Int`, and I think that the logic that determines if it it fits into a native int or not is unnecessary. Let's keep the code as simple as possible. Or, if not, why can't it be just an Int?

tbrowder: wait, forget what i just said. after looking at the opcodes again i think i see how to do the conversion. it has to get specifically get “boxed” as an Int...

tbrowder:
ok, i’m still working on it...

tbrowder:
i give up on the possibility of adding a bigint at this stage after exhausting attempts at every path I can think of. The best I can do is: (1) create an int32 or int64 (after using the Actions test for $?BITS) and (2) note the limitation in the docs.  I believe that will provide an adequate capability for %config hssh entries. If, in the future, some user is thwarted by the limitation he or she can file a feature request.

tbrowder:
still working...making some progress...

tbrowder:
still have hope for great things :-D ...

tbrowder:
Ready for another review.

src/Perl6/Pod.nqp
tbrowder: Bigints are now handled properly with the latest commit.

src/Perl6/Pod.nqp
```diff
@@ -116,32 +111,280 @@ class Perl6::Pod {
         ).compile_time_value
     }
 
+    sub decint($s) {
+        # code copied from Actions.nqp and locally modified
+        my int $base := 10;
+        my int $chars := nqp::chars($s);
+        $chars > $max-dec-intchars
+          ?? string_to_bigint($s, $base, $chars)
```
AlexDaniel: Why not use a bigint *always*? Will make the code simpler.

tbrowder: Can probably do so...will try it.

tbrowder: So done—ready for another review.

src/Perl6/Pod.nqp
```diff
@@ -1,25 +1,20 @@
 class Perl6::Pod {
 
-    # various helper methods for Pod parsing and processing
-    my $caption := ''; # var to save table caption values between 
+    # various helpers for Pod parsing and processing
+
+    # hardware int checks (from Actions.nqp):
+    my int $?BITS := nqp::isgt_i(nqp::add_i(2147483648, 1), 0) ?? 64 !! 32;
+    my $max-dec-intchars := $?BITS == 64 ?? 16 !! 9;
```
AlexDaniel: Don't need these two lines also now?

tbrowder: Argh! No...

AlexDaniel:
I think this looks good, where “good” is good enough to resolve a few tickets right now and make some tests pass. As mentioned in https://github.com/rakudo/rakudo/pull/1300#issuecomment-351248903, the whole thing needs a rewrite, and I'm hoping that someone will get to it soon. Meanwhile we can enjoy some resolved bugs and features that are now implemented (although you'll probably see bugs when doing more complicated stuff, but that's ok for now).

If anyone has anything against this approach please let me know. If not, I'll merge this tomorrow probably.

PR #1343, opened 2017-12-27
Issue 1305: Throw exception on attempted assign to range pos

jstuder-gh:
Range is immutable as mentioned in the spec
<https://github.com/perl6/specs/blob/master/S02-bits.pod#immutable-types>

Assignment failed before, but this provides a clearer error message.


zoffixznet:
This can be just a single, non-multi method, no? `method ASSIGN-POS(Range:D: |) { … }`

jstuder-gh:
You're right. I've modified the PR.

Issue #1347, opened 2017-12-29
Feature request: U+23E8 DECIMAL EXPONENT SYMBOL for scientific notation


clsn:
## The Problem
Just a small deficiency in awesome.

## Expected Behavior

It's such a great character, it's a shame more places don't use it.  It's right there in its name!

```
> 6.02⏨23
6.02e+23
>2.35⏨5
235000
```

It's so clear, placing the `10` in subscript position so the exponent looks like an _exponent_.  Even cooler would be to make that the normal output for floats, but that would probably break a lot of things. Might as well use Unicode for all it's worth.

## Actual Behavior

Doesn't do that.

AlexDaniel:
Relevant wiki page: https://github.com/rakudo/rakudo/wiki/save-me-from-texas (you should probably add your proposal there as well).

AlexDaniel:
#### “When an established math/compsci notation exists for a Perl 6 feature (and is represented in Unicode).”

I don't think it's well established. See this:

> Decimal Exponent Symbol is part of the Unicode Standard, e.g. 6.022⏨23. It is included as U+23E8 ⏨ DECIMAL EXPONENT SYMBOL to accommodate usage in the programming languages Algol 60 and Algol 68.

That does not sound like a widespread use to me.

#### “When a Perl 6 operator or syntax is basically ASCII art trying to paint a larger “glyph”, and there's a Unicode character for that exact glyph (not just resembling it, but specifically intended for it).”

No.

#### “When it is too hard to implement it in a module (e.g. it is so deep in the parser that trying to reimplement it in a module is completely unreasonable)”

I'm guessing that this should be implementable.

#### “When a particular sequence is often auto-corrected by software.”

No.


<br>
So my understanding is that it should be added to the wiki page, then we can reject the ticket for now. The best ones from the wiki page will get added eventually. Maybe.

clsn:
Fair enough.  I just didn't know enough about the whole environment to pick the right place to post this.  *I* think it's a great idea (and shouldn't that be enough?  Why do you need to consult actual authorities and standards and all?  Isn't my opinion, on all matters, the most important in the world?), and of course it's a one-character change.  I was going to post a pull-request for it but then I came back and saw these comments.  I'll post it to the wiki page; close the ticket if you like.

AlexDaniel:
[...]

PR #1348, opened 2017-12-29
Add ⏨ as exponent notation for floats.

clsn:
To address #1347 in case anyone eventually cares.

### DIFF
--- a/src/Perl6/Grammar.nqp
+++ b/src/Perl6/Grammar.nqp
@@ -3655,7 +3655,7 @@ grammar Perl6::Grammar is HLL::Grammar does STD {
         ]
     }
 
-    token escale { <[Ee]> <sign> <decint> }
+    token escale { <[Ee⏨]> <sign> <decint> }
 
     token sign { '+' | '-' | '−' | '' }
 


zoffixznet:
👎 from me on this.

I doubt many people will use this and it's even in the same token that has the U+2212 minus, which in the past caused tons of bugs, that is still buggy in cases, and that I've never seen anyone use.

This falls squarely into "you think this is cute now" category that we'd be regretting in the future.

zoffixznet:
> which in the past caused tons of bugs

Already spotted one bug were this PR accepted: `&val` (and subsequently `Str.Numeric`) [won't know](https://github.com/rakudo/rakudo/blob/5f48c0695c48f152f0541503fb05151a84bc4bad/src/core/allomorphs.pm#L287) how to handle this syntax and would need fixing. I suspect `sprintf` won't either and that lives in nqp.

AlexDaniel:
Yeah we'll have to put that into the drawer for now. Who knows, maybe we'll get back to it later. In any case, it's now listed on the wiki and I even added a link to this PR. Further discussion is welcome but this won't be implemented in the very near future.

PR #1350, opened 2017-12-30
Issue 1204: Don't wrap quoteword in Slip if one word

jstuder-gh:
When using word quoting with interpolation, interpolated words are
wrapped in a Slip. This is desirable when there are multiple words, as
the interpolated values in the Slip get integrated into the surrounding
List, but is problematic when there is only one word total (prior to
interpolation).

If there is only one word, don't wrap it in a Slip. This fixes the issue
when using both qqww and << >>.

Also added "val" sub variant that handles both Slips and Lists, as it
was there was not a good candidate for dealing with the List returned by
the "words" method (on the << >> path).

### DIFF
--- a/src/Perl6/Actions.nqp
+++ b/src/Perl6/Actions.nqp
@@ -9774,7 +9774,17 @@ class Perl6::QActions is HLL::Actions does STDActions {
             }
         }
         walk($past);
-        +@($result) == 1 ?? $result[0] !! QAST::Stmts.new( $result )
+
+        # Strip out list op and possible Slip if only one resulting word
+        nqp::if(
+            +@($result) == 1,
+            nqp::if(
+                nqp::istype($result[0], QAST::Op) && $result[0].name eq 'Slip',
+                $result[0][0],
+                $result[0]
+            ),
+            QAST::Stmts.new( $result )
+        )
     }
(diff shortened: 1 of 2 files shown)


jstuder-gh:
Will need to add tests to roast, if accepted.

zoffixznet:
Thanks.

Looks like this case where there's some text flush with the variable now produces nested results:

```
cpan@perlbuild4~/CPANPRC/temp/rakudo (issue_1204)$ ./perl6 -e 'my $x = "foo bar"; dd «f$x»'                                                                                                                                                                    
("f", ("foo", "bar"))
cpan@perlbuild4~/CPANPRC/temp/rakudo (issue_1204)$ perl6 -e 'my $x = "foo bar"; dd «f$x»'                                                                                                                                                                      
("f", "foo", "bar")
```

> Will need to add tests to roast

Would you also add a test for the above case as well, please? Not a single test in the stresstest caught this regression.

jstuder-gh:
Updated the branch to include a much better fix that does not contain the regression you pointed out. I will add to the spec test hopefully sometime later today, and include the cases you outlined above

zoffixznet:
👍 Thanks!

PR #1359, opened 2018-01-03
Fix RT#127051

titsuki:
Hi,
I fixed RT#127051: https://rt.perl.org/Public/Bug/Display.html?id=127051

The changes are:

+ Add smileys to make sure that rational calculation only uses instances not type objects [9910b39]
+ Add tests for exception handling of rational calculation [9adc58c]


lizmat:
Thank you!

titsuki:
Thanks for merging!

PR #1363, opened 2018-01-04
Fix RT#127016

titsuki:
Hi,
I fixed RT#127016: https://rt.perl.org/Public/Bug/Display.html?id=127016

The changes are:
+ Use Int:D instead of Int() [1d034df]
+ Cast Match objects to Int objects before calling Date.new [f9f8e44]

roast: https://github.com/perl6/roast/pull/374


AlexDaniel:
What's the output of ```say Date.new(‘2018’, ‘1’, ‘4’)``` now?

AlexDaniel:
Ah. Github decided to hide the important part :) Nevermind!

zoffixznet:
Reading the code, it looks like this PR breaks dates constructed from components calculated with division:  `say Date.new: 2017, 12/2, 3`

Perhaps just `Int:D()` coerser is a better option?

titsuki:
@AlexDaniel @zoffixznet 
Thanks for reviewing!
I fixed the matters pointed out.

zoffixznet:
👍 Thanks!

Issue #1304, opened 2017-12-11
Chainable ops silently fail to be chained when negated


zoffixznet:
Don't know if there's some logic behind this behaviour that I'm missing, but it looks like the ops that can normally be chained with different ops don't chain right when negating them (i.e. the Bool return of one seems to be used as input for the second op in the chain).

IMO it's LTA that it silently gives wrong result. Perhaps there's a way to make it work or make it complain?

```
$ perl6 -e 'say 1 < 3 > 2 '
True
$ perl6 -e 'say 1 !< 3 !> 2 '
True
$ perl6 -v
This is Rakudo version 2017.11-8-g7939014 built on MoarVM version 2017.11
implementing Perl 6.c.
```

zoffixznet:
If this issue is rejected, please rip out the fudged tests in https://github.com/perl6/roast/commit/5cbef9d6825e9e3fbcbc278190a59d5f004417ae

zoffixznet:
> Maybe we should close RT#121987 in favor of this ticket.

👍  I like GitHub's tracker more.

AlexDaniel:
So, more tests here: https://github.com/perl6/roast/blob/5cbef9d6825e9e3fbcbc278190a59d5f004417ae/S03-operators/relational.t#L151-L155 (from RT#121987).

jstuder-gh:
I don't know if this is the right solution for this problem, but I have one possible fix that addresses it and reapplies chaining when using the negation metaoperator.

NQP Commit:
<https://github.com/jstuder-gh/nqp/commit/84eef7cbf71d3194afb509ae4c31d6865211a155>

Rakudo Commit:
<https://github.com/jstuder-gh/rakudo/commit/07909381852d825cc6ccc5b195c735e8e12fad68>

NQP is modified because this modifies the 'chain' op (MoarVM only for now) so that it can use the first child node as the callee (as the negation metaop produces a subtree to determine the operator's behavior; seen below).

    - QAST::Op(call &METAOP_NEGATE)  :is_pure<?>
      - QAST::Var(lexical &infix:«<») <wanted>

Of course something similar would need to be done for the JVM and JS backends (I have not looked into what this would entail). I suppose this could complicate upkeep of the different backends and I don't know how this would affect optimization efforts.

If you think that this is a good solution, I can clean up these commits a bit (reword, add more context) and submit as PRs in their respective repos.

zoffixznet:
[...]

PR #1375, opened 2018-01-08
Issue 1304: Preserve operator chaining when operators negated

jstuder-gh:
This is one part in a two part fix to [Issue 1304](https://github.com/rakudo/rakudo/issues/1304).
The NQP component to this fix can be found [here](https://github.com/perl6/nqp/pull/387).

If the behavior on the base operator is to allow 'chaining' (as it is
for '<' and '>' in the grammar) and the metaop is '!', then preserve
chaining on the op wrapping the metaop and baseop.

Previously, applying the negation metaoperator toward operators that
allowed chaining (eg. '<', or '>') wouldn't preserve the chaining
behavior.

This commit is being made in conjunction with development in NQP
altering the chain op.

### DIFF
--- a/src/Perl6/Actions.nqp
+++ b/src/Perl6/Actions.nqp
@@ -7343,7 +7343,12 @@ class Perl6::Actions is HLL::Actions does STDActions {
             }
             my $t        := $basepast.ann('thunky') || $base<OPER><O>.made<thunky>;
             my $helper   := '';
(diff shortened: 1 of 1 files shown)


jstuder-gh:
Passes the existing spectest when combined with NQP commit.

Are more tests needed in the roast? I saw in [Issue 1304](https://github.com/rakudo/rakudo/issues/1304) that some had been added. Let me know if additional tests are required.

zoffixznet:
👍 Thanks!

Issue #1261, opened 2017-11-20
Silent Failure When Binding To Map Key


zoffixznet:
The second line of this code silently `fail`s to bind a new value:

```perl6
my $h := Map.new: (:42foo);
$h<foo> := 100;
say $h;  # OUTPUT: Map.new((foo => 42))
```

Presumably, the `Failure` doesn't explode because variable assignment/binding doesn't sink the returned values, BUT it does work correctly when trying the same thing with a List:

```perl6
my $h := Map.new: (:42foo);
my $x = ($h<foo> := 100);
say $h;                 # OUTPUT: Map.new((foo => 42))
say $x.^name;           # OUTPUT: Failure
say $x.exception.^name; # OUTPUT: X::Bind
```

```perl6
my $h := Map.new: (:42foo);
sink $h<foo> := 100;

# Cannot bind to Map
#  in block <unit> at test.p6 line 2
```

```perl6
my $list := <foo bar ber>;
$list[0] := 100;

Cannot use bind operator with this left-hand side
  in block <unit> at test.p6 line 5
```

jstuder-gh:
Should binding work if the value at 'foo' is a container (similar to how it works in List) or should binding to a Map key never work, under any circumstance?

jnthn:
It's assignment that would work were it a Scalar container. Assignment should not.

Since a bind is treated as sunk, I think the error here is `BIND-POS` using `fail` instead of `die`.

jstuder-gh:
Thanks for the clarification!

It does look like it is possible to [bind to the Scalar container in a List](https://github.com/rakudo/rakudo/blob/master/src/core/List.pm#L577). Is that a mistake?

jnthn:
Huh, I don't agree with that behavior. `List` is meant to be immutable.

jnthn:
(And the reason assignment to a container works is because it mutates the `Scalar`, not the `List`.)

zoffixznet:
> Huh, I don't agree with that behavior. List is meant to be immutable.

Tho it does mirror the things you can do with Scalars elsewhere, with binding uprooting it and replacing it with a plain value:

```bash
$ perl6 -e 'my $l := $,; say $l; $l[0] := 42; say $l; my $s := $; say $s; $s := 42; say $s'
((Any))
(42)
(Any)
42
```

And if you squint you could say it does still mutate just the Scalar, mutating it to a value 😸

jnthn:
[...]

PR #1396, opened 2018-01-13
Issue 1261: Throw BIND-KEY Exception instead of using Failure

jstuder-gh:
As suggested by zoffix and jnthn, since the result of BIND-KEY does not
sink, the Exception within the Failure is never thrown. Throws now.

Addresses [Issue 1261](https://github.com/rakudo/rakudo/issues/1261)


lizmat:
Thanks!

Issue #1397, opened 2018-01-13
.ords.rotor(2=>-1,:partial) doesn't work


b2gills:
## The Problem

    > 'abc'.ords.rotor( 2=>-1, :partial )
    Cannot invoke this object (REPR: Uninstantiable; Callable)

    > dd 'abc'.ords.rotor( 2=>-1, :partial )
    (low-level Seq)

## Expected Behavior

    > 'abc'.ords.List.Seq.rotor( 2=>-1, :partial )
    ((97 98) (98 99) (99))

## Environment
* Operating system: Linux
* Compiler version: 2017.12-204-g1e8c976

zoffixznet:
Test to cover the bug:

```perl-6
subtest '.rotor: 2 => -1, :partial obeys Iterator protocol' => {
    plan 2;

    my $iterends = 0;
    my $s := Seq.new: class :: does Iterator {
        method pull-one {
            return $++ if $++ < 3;
            $iterends++;
            IterationEnd
        }
    }.new;
    is-deeply $s.rotor(2 => -1, :partial), ((0, 1), (1, 2), (2,)).Seq,
        'got right result';
    is $iterends, 1, 'stopped pulling after receiving IterationEnd';
}
```

PR #1400, opened 2018-01-14
Issue 1397: Reifiy pending elements in Seq prior to rotor call

jstuder-gh:
There are instances where a Seq is newly created and its elements have
not been iterated/cached yet. Calling a method like rotor that
anticipates such elements will fail. Add a method in Seq that intercepts
the Seq and reifies any pending elements prior to dispatching it to the
Any rotor method for processing.

Addresses [Issue 1397](https://github.com/rakudo/rakudo/issues/1397)

### DIFF
--- a/src/core/Rakudo/Iterator.pm
+++ b/src/core/Rakudo/Iterator.pm
@@ -2795,6 +2795,7 @@ class Rakudo::Iterator {
             has $!cycle;
             has $!buffer;
             has int $!complete;
+            has int $!is-exhausted = 0;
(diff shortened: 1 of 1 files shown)


jstuder-gh:
Passes existing spectest. Will need to write additional tests if accepted.

jstuder-gh:
Caching did resolve the error in this instance, but your way is better and resolves the underlying issue.

I've updated the branch with the new fix.

zoffixznet:
👍 Thanks!

PR #1402, opened 2018-01-14
Fix RT#132214

titsuki:
Hi, I Fixed RT#132214: https://rt.perl.org/Public/Bug/Display.html?id=132214

The changes are:

+ Delete an iteration of $<version> from statement_control:sym<need> [7537f89]
+ Make "need" enable to throw an exception when it detects version token [0cdd82a]
+ Recommend "use" instead of "need" [0a73f19]
+ Add tests for "need 6.0" and "need v6.0" [4d5e097]


lizmat:
Thanks!

titsuki:
Thanks for merging :)

Issue #1380, opened 2018-01-09
Incompatible pointer type warning in NativeCall tests


zoffixznet:
This showed up [in travis run](https://travis-ci.org/rakudo/rakudo/jobs/326789558) on v2017.12.144.g.5.f.7.fdfdd.3:

```
t/04-nativecall/01-argless.c: In function ‘ArglessUTF8String’:
t/04-nativecall/01-argless.c:39:12: warning: return from incompatible pointer type [-Wincompatible-pointer-types]
     return my_str;
            ^~~~~~
```

PR #1446, opened 2018-01-26
Fix function signature

CurtTilmes:
Current signature is int*, but it actually returns a char* which gives a
warning. It still works because Perl declares the nativecall as a Str
anyway.  This just avoids the warning.

Address issue https://github.com/rakudo/rakudo/issues/1380


zoffixznet:
👍 Thanks!

PR #1448, opened 2018-01-27
Issue 1304 Continued: Re-enable negated chaining on the JVM

jstuder-gh:
This statement was previously restricted to MoarVM only as the JVM
implementation's chain op was not yet set up to handle a child as the callee.

Now, as of commit 84278100acbf589a37f5baa85ffe27d68de52d13 (in the NQP
repo), the chain op's JVM implementation works the same way that
MoarVM's does, making this negated chaining implementation functional
on the JVM backend as well.

See <https://github.com/rakudo/rakudo/issues/1304>.


jstuder-gh:
Depends upon NQP change in <https://github.com/perl6/nqp/pull/402>

zoffixznet:
👍 Thanks!

PR #1454, opened 2018-01-29
Fix Iterator.Rotor when cycle contains Whatever or Inf

taboege:
Whatever or Inf in the rotor cycle were encoded as -1 in the number of elements to consume. This didn't have the desired effect with /signed/ comparisons.

Two tests will be submitted to roast shortly, see: https://github.com/taboege/roast/commit/fa3dea758f846efdfbe32044498b262496c6c8b9

### DIFF
--- a/src/core/Rakudo/Iterator.pm
+++ b/src/core/Rakudo/Iterator.pm
@@ -2891,12 +2891,17 @@ class Rakudo::Iterator {
                     )
                   ),
                   nqp::until(                          # fill the buffer
-                    nqp::isge_i(nqp::elems($!buffer),$elems)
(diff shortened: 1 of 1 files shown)


lizmat:
How can `$elems` become -1 ?

taboege:
It's done explicitly earlier in the code to signify Whatever and Inf, e.g. src/core/Rakudo/Iterator.pm +2825:

```
nqp::if(
    nqp::istype($size,Whatever),
    nqp::stmts(                    # eat everything
        (my int $elems = -1),
        ($!complete = 0)
    ),
```

lizmat:
Thanks!

Issue #1341, opened 2017-12-26
state variable resets for each loop of a loop expression


Summertime:
## The Problem
```pl6
say do loop { state $x = 0;  last if ++$x > 2;  $x; }
```
## Expected Behavior

outputs `(1 2)`

## Actual Behavior

outputs an infinite list `(1 1 1 1 1 1 ...)`

*aside:* it appears to work with jvm:
```pl6
j: say do loop { state $x = 0; last if ++$x > 2; $x }
<camelia> rakudo-jvm a92950fb4: OUTPUT: «(1 2)␤»
```

## Steps to Reproduce

```
PS C:\[...]> perl6 -e 'say do loop { state $x = 0;  last if ++$x > 2;  $x; }'
(1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 1 ...)
```

## Environment
* Operating system: Windows 10 64bit
* Compiler version (`perl6 -v`):
```
This is Rakudo Star version 2017.10 built on MoarVM version 2017.10
implementing Perl 6.c.
```

jstuder-gh:
This code produces the QAST subtree below:

        - QAST::Block(:cuid(3) :blocktype(declaration_static))  :statement_id<1> :in_stmt_mod<?> :code_object<?> :outer<?>
          - QAST::Stmts  { state $x = 0; last if ++$x > 2; $x }
          - QAST::Stmts 
            - QAST::Block(:cuid(1) :blocktype(immediate)) <wanted> :statement_id<2> :in_stmt_mod<?> :code_object<?> :outer<?> { state $x = 0; last if ++$x > 2; $x }
              - QAST::Stmts  
                - QAST::Var(lexical $x :decl(statevar)) 
                - QAST::Var(lexical $_ :decl(var)) 

The block cuid(3) gets cloned repeatedly by a looping construct and as a result a fresh new container is being created for the nested cuid(1) for each statevar instance. The value in each new container is reinitialized to 0, resulting in the result seen in the ticket when incremented.

[...]

PR #1467, opened 2018-02-02
Issue 1341: Elevate statevars to enclosing block when wrapped

jstuder-gh:
Previously in some blocks following the "do" statement prefix, statevars
would be seemingly reset after each iteration of the loop.

Given the statement:

    say do loop { state $x = 0; last if ++$x > 2; $x }

This statement would result in multiple thunk refs being produced, and
ultimately, would be wrapped in a block that is used as a cloning target
in looping and a new container would be declared and initialized for
each iteration.

The QAST would look like this (trimmed to relevant parts):

    - QAST::Block(:cuid(3) :blocktype(declaration_static))
      - QAST::Stmts  { state $x = 0; last if ++$x > 2; $x }
      - QAST::Stmts
        - QAST::Block(:cuid(1) :blocktype(immediate)) <wanted>
          - QAST::Stmts
            - QAST::Var(lexical $x :decl(statevar))
            - QAST::Var(lexical $_ :decl(var))

After this commit the QAST looks like this:

    - QAST::Block(:cuid(3) :blocktype(declaration_static))
      - QAST::Stmts  { state $x = 0; last if ++$x > 2; $x }
        - QAST::Var(lexical $x :decl(statevar))
      - QAST::Stmts
        - QAST::Block(:cuid(1) :blocktype(immediate)) <wanted>
          - QAST::Stmts
            - QAST::Op(null)
            - QAST::Var(lexical $_ :decl(var))

This way the statevar is declared in the block being cloned and thus the
same container is used in each iteration.

Note the enclosing block is a block newly created via the make_thunk_ref sub,
and contains no child nodes other than the first Stmts (that has no children prior
to this commit) and the second Stmts containing the cuid(1) subtree.

See [Github Issue 1341](https://github.com/rakudo/rakudo/issues/1341)

### DIFF
--- a/src/Perl6/Actions.nqp
+++ b/src/Perl6/Actions.nqp
@@ -130,6 +130,14 @@ sub wanted($ast,$by) {
                 QAST::WVal.new( :value($*W.find_symbol(['Seq']))),
                 block_closure($block) );
 
+            # Elevate statevars to enclosing thunk
+            if $body.has_ann('has_statevar') && $block.has_ann('past_block') {
+                Perl6::Actions::migrate_blocks(
+                    $body, $block.ann('past_block'),
+                    -> $n { nqp::istype($n, QAST::Var) && $n.decl eq 'statevar' }
+                )
+            }
+
             # conditional (if not always true (or if repeat))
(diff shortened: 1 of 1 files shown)


zoffixznet:
👍 Thanks!

PR #1468, opened 2018-02-02
Add more documentation on building MoarVM

Kaiepi:
Ways this could be better written would be appreciated
See https://github.com/rakudo/rakudo/issues/1420#issuecomment-359354469


zoffixznet:
👍 Thanks!

PR #1473, opened 2018-02-03
[RFC] first draft of "datagram"-oriented API for udp sockets

timo:
ask the .Supply to give you :datagrams and you'll
receive instances of IO::Socket::Async::Datagram with
.data being the string or buffer of data and
.hostname and .port being the sender's address.

requires a branch of the same name to be checked out in MoarVM.

I'd like to get some suggestions for the names involved here. We'd probably want to use :datagram instead of :datagrams for example, and i'm not sure about the IO::Socket::Async::Datagram class name, either. The attributes I'm not too worried about, and the "decode gives another datagram" functionality doesn't seem terribly bad - though perhaps it wants to get a function in the other direction as well.

### DIFF
--- a/src/core/IO/Socket/Async.pm
+++ b/src/core/IO/Socket/Async.pm
@@ -42,17 +42,28 @@ my class IO::Socket::Async {
         $p
     }
 
+    my class Datagram {
+        has $.data;
+        has $.hostname;
+        has $.port;
+
+        method decode(|c) {
+            return self.clone(data => $!data.decode(|c));
+        }
+    }
+
     my class SocketReaderTappable does Tappable {
         has $!VMIO;
         has $!scheduler;
         has $!buf;
         has $!close-promise;
+        has $!udp;
 
-        method new(Mu :$VMIO!, :$scheduler!, :$buf!, :$close-promise!) {
-            self.CREATE!SET-SELF($VMIO, $scheduler, $buf, $close-promise)
+        method new(Mu :$VMIO!, :$scheduler!, :$buf!, :$close-promise!, :$udp!) {
+            self.CREATE!SET-SELF($VMIO, $scheduler, $buf, $close-promise, $udp)
         }
 
-        method !SET-SELF(Mu $!VMIO, $!scheduler, $!buf, $!close-promise) { self }
+        method !SET-SELF(Mu $!VMIO, $!scheduler, $!buf, $!close-promise, $!udp) { self }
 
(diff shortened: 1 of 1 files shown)


timo:
oh, and maybe we want to build some kind of shorthand for sending to the hostname/port that a datagram was received from.

abraxxa:
Looking forward to more and better low level socket support! This looks good to me, my only issue might be performance if the construction of the object slows the receiver down too much.
Do you have benchmarks?

timo:
i've tried this:

1.) Generate a firehose of UDP packets: 
`perl6 -e 'my $s = IO::Socket::Async.udp(); my $data = "Hello, World! How are you?".encode("utf8"); for ^1_000_000 { await $s.write-to("localhost", 3333, $data) }'`

2.) measure how long it takes to receive a hundred thousand udp packets with the regular interface
`time perl6 -e 'react { my int $cnt = 0; whenever IO::Socket::Async.bind-udp(\'localhost\', 3333) { $cnt++; done if $cnt >= 100_000 } };'`

3.) measure how long it takes to receive a hundred thousand udp packets with datagrams
`time perl6 -e 'react { my int $cnt = 0; whenever IO::Socket::Async.bind-udp(\'localhost\', 3333).Supply(:datagrams) { $cnt++; done if $cnt >= 100_000 } };'`

4.) revert moarvm and rakudo to master and measure step 2 again (make sure to kill 1 so it doesn't keep the moarvm binary from being updated in "make install") - for reference, my master branches are at 2018.01-59-g6da2afe5a for moarvm and 2018.01-49-gc6cc673de for rakudo.

i'm seeing some strange results:

a) after reverting to master, the regular interface is actually slower. perhaps i accidentally changed the behavior of whether data will be decoded or not when received? 
b) the datagram version maxes out at about a gig of ram, whereas the regular one maxes out at about half a gig in my branch and three quarters of a gig on master

I have to get back to apartment moving business now. i might not have time to continue this until after the 11th, but i've still got a few ideas to speed things up:

1.) teach sockets to keep the last source address and source port around in a little cache so that it doesn't have to re-encode the hostname and re-box the port all the time in the somewhat common case of the same source sending multiple messages in a row
2.) have a flag signal whether hostname and port should be boxed at all - right now they'd always have the perl6 objects (Str and Int, not Datagram) created and immediately become garbage if they're not used.

timo:
merged outside of github in ada26ca18e08c48ad0fbf431cd99f82e2e9723af

PR #1494, opened 2018-02-07
Separate adding child block from creating code obj

jstuder-gh:
Not a necessary change, but one that I think make things a bit clearer.

Previously, the create_simple_code_object method in the Perl6 World
would add the block passed to it to the desired outer block in addition
to returning a code object for it. From the name, it is not clear that
the method has such side-effects. Separating the actions in order to
make it explicit that we performing both of these.

### DIFF
--- a/src/Perl6/Actions.nqp
+++ b/src/Perl6/Actions.nqp
@@ -1084,9 +1084,8 @@ class Perl6::Actions is HLL::Actions does STDActions {
             );
 
             $*W.pop_lexpad();
-            $*W.add_child_to_outer_block($block);
             $*W.add_phaser(
-                $/, 'INIT', $*W.create_simple_code_object($block, 'Block'), $block
+                $/, 'INIT', $*W.create_code_obj_and_add_child($block, 'Block'), $block
             );
         }
     }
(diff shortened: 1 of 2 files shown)


jstuder-gh:
Also wanted to point out that at [line 8338 in Actions](https://github.com/jstuder-gh/rakudo/blob/sep_code_obj_actions/src/Perl6/Actions.nqp#L8338) it is pushing the block onto the outer block twice.

I didn't change that as I don't know the full context of what is going on there, but it does look a bit suspect IMO.

zoffixznet:
👎 on this. We can probably just tweak the original name for clarity. I don't see much point in paying for another lookup of `$*W` variable plus another method call just because original method name doesn't tell the full story.

vendethiel:
> paying for another lookup of $*W variable

Is the cost that high? Would it make sense to alias it in a local variable beforehand, then?

jstuder-gh:
Ok, I see your point. Added another commit that changes the name instead and hopefully makes it a bit more clear that the method performs additional actions.

zoffixznet:
👍 Thanks!

zoffixznet:
> Also wanted to point out that at line 8338 in Actions it is pushing the block onto the outer block twice.

> I didn't change that as I don't know the full context of what is going on there, but it does look a bit suspect IMO.

Yeah, that looks to be a bug and it should be pushed only once. Looking at the [QAST dump](https://temp.perl6.party/s-qast.html) of `my $x ~~ s/foo/bar/` `QAST::Block :cuid(2)` shows up twice.

PR #1495, opened 2018-02-07
Remove extra push of child to outer block

jstuder-gh:
The action should only occur once and already occurs in the method to
create code objects. zoffix++ for verification.


zoffixznet:
👍 Thanks!

Issue #1488, opened 2018-02-06
multi sub infix:<cmp>(\a, \b) is missing decont


zoffixznet:
Fast-path is missing nqp::decont here: https://github.com/rakudo/rakudo/blob/0bf735aaf/src/core/Order.pm#L11

zoffixznet:
tests https://github.com/rakudo/rakudo/pull/1499

PR #1498, opened 2018-02-07
Fix R#1488 - multi sub infix:<cmp>(\a, \b) is missing decont. Zoffix++

dogbert17:
Phixes: https://github.com/rakudo/rakudo/issues/1488

There's every chance that the codebase contains more problems like
this.


zoffixznet:
👍 Thanks!

PR #1505, opened 2018-02-09
Don't add blank lines to REPL history file

0racle:


0racle:
I should also have mentioned, this also applies to your live (buffered) history in the REPL.

zoffixznet:
👍 Thanks!

PR #1522, opened 2018-02-13
Remove the Bridge method from the numeric classes

ash:
The Bridge method is mainly used in the Real role, where it serves a proxy method to get to the Num representation of a value. Other classes (Int, Rat, Num, etc.) use the call of Real.Bridge when they participate in trigonometric or arithmetic (and more, e.g. comparison) operations.

Although I might be missing the proper intention of the Bridge method, it looks it is overcomplicating the process and is not really needed. All the use cases of Bridge for the built-in data types effectively coerce the data to a Num value.

There's a small speed gain in the following case, but this is a side story which is not actually the main goal of Bridge removal:

time ./perl6 -e'for 1..10_000_000 {Int.new(1).sin}'


zoffixznet:
👍 I always found the `.Bridge` method a little baffling.

The [speculations](https://design.perl6.org/S32/Numeric.html#line_220) mention it as a way for custom types to let themselves be coerceable to core numeric types, but we do it via `Numeric` method anyway:

```
with class Foo does Numeric { multi method Numeric { 42 } }.new { say $_ + $_ }
# OUTPUT: 84
```

And, in fact, defining `Bridge` method instead just crashes:
```
with class Foo { method Bridge { 42 } }.new { say $_ + $_ }'
Cannot resolve caller Numeric(Foo: ); none of these signatures match:
    (Mu:U \v: *%_)
  in block <unit> at -e line 1
```

colomon:
Bridge is intended to make it easy to write new numeric classes and have
them interoperate with other numeric classes by providing a default common
representation.  It's an entirely different operation than .Numeric, which
usually just returns self for actual numeric types.   It's more like .Num,
but left intentionally vague to allow implementations to use something else
(FatRat, for instance) as the common number type.

AlexDaniel:
What about all these tests like https://github.com/perl6/roast/blob/master/S32-num/real-bridge.t?

ash:
@AlexDaniel does not pass, obviously (while i did not see during make spectest). Should be fixed if accepted.

AlexDaniel:
@ash we can fix these tests on roast master but this also needs a change in v6.c-errata, and there must be a very good reason to do so. Moreover, there's [at least one module](https://github.com/grondilu/p6-modular/blob/990846d44b11998eef4685f8cc7d1648c031cffb/lib/Modular.pm#L4) using Bridge method.

zoffixznet:
>  It's an entirely different operation than .Numeric [...] It's more like .Num, but left intentionally vague to allow implementations to use something else (FatRat, for instance) as the common number type.

Ah, OK. Now I finally get it. I thought such a case would still end up with a `Num`, but it wouldn't if, say, two custom objects are used:

```
with class :: does Real { method Bridge { FatRat.new: 67, 4 } }.new { dd $_ + $_ }
# OUTPUT: FatRat.new(134, 4)
```

I rescind my 👍 then.

PR #1539, opened 2018-02-17
RT#102994: State init vars if not set on first pass

jstuder-gh:
Given the following subroutine:

    sub f($x) {
        return if $x;
        state $y = 5;
        $y;
    }

Previously, ```$y``` would only be assigned a value if the statement is
reached on the first invocation of the static frame. So if ```f(1)``` is called first,
each subsequent ```f(0)``` call returns ```(Any)``` 

To address this, added an indicator on the static frame (in
corresponding MoarVM/JVM developments) and modified ```p6stateinit``` to take a var
and check whether it has been assigned to. If the assignment was not
reached in the first pass in the life of a static frame, the is_hll_init
value indicates that it requires HLL assignment the next time it is
encountered.

See [RT#102994](https://rt.perl.org/Public/Bug/Display.html?id=102994)

### DIFF
--- a/src/Perl6/Actions.nqp
+++ b/src/Perl6/Actions.nqp
@@ -2343,7 +2343,9 @@ class Perl6::Actions is HLL::Actions does STDActions {
         # generate code that runs the block only once
         make QAST::Op.new(
             :op('if'),
-            QAST::Op.new( :op('p6stateinit') ),
+            QAST::Op.new( :op('p6stateinit'),
+                QAST::Var.new( :name($sym), :scope('lexical') )
+            ),
             QAST::Op.new(
                 :op('p6store'),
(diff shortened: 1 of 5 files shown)


jstuder-gh:
Relies on changes made in <https://github.com/MoarVM/MoarVM/pull/805> and <https://github.com/perl6/nqp/pull/418>. 
 
 
This change passes all the tests in the (MoarVM) spectest except one: 
 
    t/spec/S04-declarations/state.rakudo.moar 
 
which I assume is because of the modification to the MVMStaticFrame struct. I don't know how to update this test.

jstuder-gh:
This is not going to work (See [MoarVM PR#805](https://github.com/MoarVM/MoarVM/pull/805)). Closing this now. I'll submit another PR at a later date if I create something that addresses the entire issue.

PR #1542, opened 2018-02-18
RT#102994: State init vars if not set on first pass (CodeRef edition)

jstuder-gh:
Given the following subroutine:

    sub f($x) {
        return if $x;
        state $y = 5;
        $y;
    }

Previously, ```$y``` would only be assigned a value if the statement is
reached on the first invocation of the closure. So if ```f(1)``` is called first,
each subsequent ```f(0)``` call returns ```(Any)```

To address this, added an indicator on the code ref associated with the block (in
corresponding MoarVM/JVM developments) and modified ```p6stateinit``` to take a var
and check whether it has been assigned to. If the assignment was not
reached in the first pass in the life of a closure, the is_hll_init
value indicates that it requires HLL assignment the next time it is
encountered.

See [RT#102994](https://rt.perl.org/Public/Bug/Display.html?id=102994)

### DIFF
--- a/src/Perl6/Actions.nqp
+++ b/src/Perl6/Actions.nqp
@@ -2358,7 +2358,9 @@ class Perl6::Actions is HLL::Actions does STDActions {
         # generate code that runs the block only once
         make QAST::Op.new(
             :op('if'),
-            QAST::Op.new( :op('p6stateinit') ),
+            QAST::Op.new( :op('p6stateinit'),
+                QAST::Var.new( :name($sym), :scope('lexical') )
+            ),
             QAST::Op.new(
                 :op('p6store'),
                 WANTED(QAST::Var.new( :name($sym), :scope('lexical') ),'once'),
@@ -3231,7 +3233,7 @@ class Perl6::Actions is HLL::Actions does STDActions {
                     }
 
                     $past := QAST::Op.new( :op('if'),
-                        QAST::Op.new( :op('p6stateinit') ),
+                        QAST::Op.new( :op('p6stateinit'), $orig_past ),
                         $past,
                         $orig_past);
                     $past.nosink(1);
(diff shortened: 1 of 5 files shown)


jstuder-gh:
This is different from the PR I submitted yesterday. This PR sets the flags on the CodeRef associated with a closure, making it about 100% more useful.

jstuder-gh:
Relies on changes made in <https://github.com/MoarVM/MoarVM/pull/806> and <https://github.com/perl6/nqp/pull/420>.

This change passes all the tests in the (MoarVM) spectest except one:

    t/spec/S04-declarations/state.rakudo.moar 

which I assume is because of the modification to the MVMCodeBody struct. I don't know how to update this test.

jstuder-gh:
It seems that I may have misread and that it did not fail the spectest? I find the output to be confusing. The output is as follows:

    Test Summary Report
    -------------------
    t/spec/S04-declarations/state.rakudo.moar                       (Wstat: 0 Tests: 44 Failed: 0)
      TODO passed:   38
    Files=1222, Tests=76133, 740 wallclock secs (10.36 usr  1.41 sys + 2197.44 cusr 86.28 csys = 2295.49 CPU)
    Result: PASS

I don't really understand the output or why state.rakudo.moar is there, but it says it passes. I must have read it last time as "44 Failed" before instead of "Tests: 44; Failed: 0".

I have rebased all the branches for all 3 repos to the current HEAD of master and have run the spectest (Roast master is [b68073aa1579d69c7834e62335ae7aadff758061](https://github.com/perl6/roast/commit/b68073aa1579d69c7834e62335ae7aadff758061) as of the time of writing) and it passed with the same output as above.

It also passes the new tests introduced in [Roast PR #398](https://github.com/perl6/roast/pull/398).

zoffixznet:
> why state.rakudo.moar is there,

The line below it says `TODO passed:   38`. So it's reporting that one of the tests is [marked as a TODO](https://docs.perl6.org/language/testing#index-entry-todo-todo%28%24reason%2C_%24count_%3D_1%29) and is expected to fail, but it's now passing.

Presumably your commit fixed some bug and the now-passing test will need to be unfudged (if you don't know what "fudging" is in roast, it's described [here](https://github.com/perl6/roast#introduction))

jstuder-gh:
Oh I see now. The test that is marked todo is unfudged in [Roast PR #398](https://github.com/perl6/roast/pull/398)

jstuder-gh:
I'm closing this PR. I intend to resubmit in the future, but with a rebased and reworked version.

PR #1558, opened 2018-02-23
Speedup Rational.Str

MasterDuke17:
By generating the fractional part two characters at a time instead of
one.

For a FatRat that ends up being a 1571 char string, time before was 2.75s, time after is 1.9s.
For a FatRat that ends up being a 3139 char string, time before was 16.2s, time after is 12.6s.

### DIFF
--- a/src/core/Rational.pm
+++ b/src/core/Rational.pm
@@ -99,7 +99,14 @@ my role Rational[::NuT = Int, ::DeT = ::("NuT")] does Real {
                 ?? 6 !! $!denominator.Str.chars + 1;
 
             my $fract-result = '';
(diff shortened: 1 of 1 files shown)


AlexDaniel:
Heh, but why not 3 or 4 characters at a time? :joy: What's so magical about 2?

zoffixznet:
👍 Thanks!

> What's so magical about 2?

You need a check for roundness and the lower numbers would still have to be handled because `$precision` might not chunk up precisely. Notice the original 1-char loop is still in the code and then 2-char loop has a roundness check. Adding more chars would mean more loops and more checks, which gets progressively messier.

Issue #1567, opened 2018-02-25
Test::like() and Test::unlike() don't coerce anymore


jdv:
The docs say at https://docs.perl6.org/language/testing#By_regex say:

"Marks a test as passed if the $value, when coerced to a string, matches the $expected-regex."

So I was surprised when I tried to do like($io_path_obj, /foo/) it bombed out since
IO::Path has a Str method.  I believe this commit, ba3cf4e5b459479b338dffd5a00ff2a07608b595,
broke the ability for the coercion to work.

I made an attempt to fix and test it.  Here's the fix PR:  https://github.com/rakudo/rakudo/pull/1568

But the file I tried to change for testing seems to be not used and I couldn't find another so here's the diff:

~~~~
[jdv@new-host spec]$ git diff
diff --git a/S24-testing/1-basic.t b/S24-testing/1-basic.t
index 1d59aea..0afbf29 100644
--- a/S24-testing/1-basic.t
+++ b/S24-testing/1-basic.t
@@ -83,6 +83,12 @@ my $foo = Foo.new();
 isa-ok($foo, 'Foo');
 isa-ok(Foo.new(), 'Foo');
 
+class Bar {
+    has $.attr is rw;
+    multi method Str { $.attr; }
+}
+$bar = Bar.new;
+
 ## like
 
 like("Hello World", rx:P5/\s/, '... testing like()');
@@ -92,6 +98,10 @@ like("Hello World", rx:P5/\s/, :desc('... testing like()'));
 like("HelloWorld", rx:P5/\s/, desc => '... testing like()', todo => 1);
 like("HelloWorld", rx:P5/\s/, :todo(1), :desc('... testing like()'));
 
+like($bar, /asdf/, :desc('... testing like()'), :todo(1));
+$bar = 'asdf';
+like($bar, /asdf/, :desc('... testing like()'));
+
 ## unlike
 
 unlike("HelloWorld", rx:P5/\s/, '... testing unlike()');
@@ -101,6 +111,10 @@ unlike("HelloWorld", rx:P5/\s/, :desc('... testing unlike()'));
 unlike("Hello World", rx:P5/\s/, todo => 1, desc => '... testing unlike()');
 unlike("Hello World", rx:P5/\s/, :desc('... testing unlike()'), :todo(1));
 
+unlike($bar, /asdf/, :desc('... testing like()'), :todo(1));
+$bar = Nil;
+unlike($bar, /asdf/, :desc('... testing like()'));
+
[...]

PR #1568, opened 2018-02-25
Unbusts like and unlike coercion.

jdv:
See https://github.com/rakudo/rakudo/issues/1567.

### DIFF
--- a/lib/Test.pm6
+++ b/lib/Test.pm6
@@ -489,7 +489,7 @@ multi sub like(
     $time_after = nqp::time_n;
     $got.defined; # Hack to deal with Failures
     my $ok;
-    if $got ~~ Str:D {
+    if Str($got) ~~ Str:D {
         my $test = $got ~~ $expected;
         $ok = proclaim($test, $desc);
         if !$test {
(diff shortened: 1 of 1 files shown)


lib/Test.pm6
```diff
@@ -489,7 +489,7 @@ multi sub like(
     $time_after = nqp::time_n;
     $got.defined; # Hack to deal with Failures
     my $ok;
-    if $got ~~ Str:D {
+    if Str($got) ~~ Str:D {
```
zoffixznet: Here, you're coercing to a Str and then checking if the result is a Str.... It's extremely unlikely the result would be false, so why keep the `else` branch around?

lib/Test.pm6
```diff
@@ -489,7 +489,7 @@ multi sub like(
...
         my $test = $got ~~ $expected;
```
zoffixznet: Here, you're still smartmatching against the original, not its stringified variant. So `like <foo bar ber>, /^foo$/` would pass the test, despite there being no match, if `$got` is actually stringified.

Issue #1531, opened 2018-02-16
Failure to sink `for` when `Z+=` used as last statement


zoffixznet:
Spotted this on https://stackoverflow.com/questions/45001820/which-context-confuses-this-perl-6-zip-operator

```
my $hash = %(
    wallet   => 100,
    gave     =>   0,
    received =>   0,
    );

for ^1 { $hash<wallet gave> Z+= <-1 1> };

dd $hash;
```

## Expected Behavior

```
Hash $hash = ${:gave(1), :received(0), :wallet(99)}
```

## Actual Behavior

```
Hash $hash = ${:gave(0), :received(0), :wallet(100)}
```

perl6 -v = 2018.01-204-gbeaad7b

PR #1570, opened 2018-02-26
Fix failure to sink last statements of `for` loops

zoffixznet:
**TL;DR: this fixes bugs but tanks performance of for loops by 8% and I'm scared to push it to master.**

-----

Fixes R#1531 https://github.com/rakudo/rakudo/issues/1531
Also fixes lack of explosion in `sub foo {fail}; foo for ^1`

In `for` loops, users would tend to expect last statement to get
sunk, but because our for loops are basically `.map` calls, they
don't. For methods calls, we already specially sink them if they're
the last statement, so `Failure.new for ^1` *does* blow up. But
we don't do anything for nqp::calls, which is why `sub foo {fail};
foo for ^1` doesn't blow up.

This patch fixes that problem and fixes stuff like $a Z+= 1 that
returns a Seq that actually needs to get sunk to perform the work.

There's nothing wrong with the patch as far as I can see, except
for one thing: it degrades performance by 8% when measured with
`sub foo {my $x}; { for ^14000_000 { foo }; say now - ENTER now }`
When I ran the stresstest, it too came out 8% slower than my earlier
runs. So... I'm kinda hesitant about pushing this to master.

### DIFF
--- a/src/Perl6/Actions.nqp
+++ b/src/Perl6/Actions.nqp
@@ -420,6 +420,8 @@ sub unwanted($ast, $by) {
             elsif $node.op eq 'p6for' || $node.op eq 'p6forstmt' {
                 $node := $node[1];
                 if nqp::istype($node,QAST::Op) && $node.op eq 'p6capturelex' {
+                    add-sink-to-final-call($node.ann('past_block'), 1)
+                        unless $*COMPILING_CORE_SETTING;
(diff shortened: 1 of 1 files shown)


zoffixznet:
Actually, slower stresstest run [appears to have been a fluke](https://irclog.perlgeek.de/perl6-dev/2018-02-26#i_15860570) and more runs come out even 2s faster than on master, so doesn't look like it's as a massive perf loss as I originally thought.

Merging.

Issue #1576, opened 2018-02-28
Assigning NativeCall return value to constant segfaults on calling the sub again


Kaiepi:
## The Problem
I have a module where I was exposing a set of a library's constants like this:
```perl6
use v6.c;
use NativeCall;
unit class Net::LibIDN2:ver<0.0.1>:auth<github:Kaiepi>;

sub idn2_check_version(Str --> Str) is native('idn2') { * }
method check_version(Str $version --> Str) { idn2_check_version($version) or '' }

constant IDN2_VERSION = idn2_check_version('');
```

## Expected Behavior
`IDN2_VERSION` should be the version of the library installed and calling `Net::LibIDN2.check_version` should return the version as well.

## Actual Behavior
The constant has the correct value, but MoarVM segfaults on calling the method. Removing the constants makes the method work. This doesn't seem to happen with code like this, so I figure `unit class` may be playing into it:
```perl6
use v6.c;
use NativeCall;

sub strlen(Str --> int32) is native('c') { * }

class Foo {
    method strlen(Str $str --> Int) { strlen($str) }
}

constant ZERO = strlen('');

sub MAIN { say Foo.strlen('ayy') }
```

## Steps to Reproduce

Install [Net::LibIDN2](https://github.com/Kaiepi/p6-Net-LibIDN2) and LibIDN2. Revert [this commit](https://github.com/Kaiepi/p6-Net-LibIDN2/commit/ccf588ba6cda85e4c2a33707e2980feef168760b). Call `Net::LibIDN2.check_version`.

## Environment
* Operating system:
OpenBSD morfent.kennel.qt 6.2 GENERIC.MP#5 amd64
* Compiler version (`perl6 -v`): 
This is Rakudo version 2018.02.1-47-g838782b7d built on MoarVM version 2018.02-22-g25677aeeb
implementing Perl 6.c.

zoffixznet:
> so I figure unit class may be playing into it

IIRC native called routines are loaded on first use. I imagine what's happening is first use occurs during module's precomp and when you call it again after loading the precomped module that first use's loaded routine is missing, hence SEGV.

Golfed repro:

```
$ cat Foo.pm6
use NativeCall;
sub strlen(Str --> int32) is native is export {}
BEGIN strlen ''

$ perl6 -I. -MFoo -e 'strlen ""'
Segmentation fault (core dumped)
```

zoffixznet:
[...]

PR #1577, opened 2018-02-28
Differentiate precomp NC sub setup markers

zoffixznet:
Fixes https://github.com/rakudo/rakudo/issues/1576 R#1576

If a NativeCalled sub used is during precomp, it's marked as
`$!setup`. However, when it's later used in non-precomp environment,
the stuff setup does is not available and so we get a SEGV.

Fix by using a separate marker for $!setup when we're in precomp
mode so that when we switch to non-precomp mode, we setup again.

### DIFF
--- a/lib/NativeCall.pm6
+++ b/lib/NativeCall.pm6
@@ -260,6 +260,7 @@ my Lock $setup-lock .= new;
 # native call.
 our role Native[Routine $r, $libname where Str|Callable|List|IO::Path|Distribution::Resource] {
     has int $!setup;
+    has int $!precomp-setup;
     has native_callsite $!call is box_target;
     has Mu $!rettype;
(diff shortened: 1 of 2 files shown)


lizmat:
Why is the `nqp::istrue()` needed her: https://github.com/rakudo/rakudo/pull/1577/files#diff-46ffdef2be7f008af940be7578df7d89R298

zoffixznet:
> Why is the nqp::istrue() needed

Thanks. Brainfart. I tested using this, which calls .Bool and fails:

```
$ perl6 -e 'BEGIN say ?$*W.is_precompilation_mode'
An exception occurred while evaluating a BEGIN
at -e:1
Exception details:
  No such method 'Bool' for invocant of type 'NQPMu'
```

Update pushed.

PR #1578, opened 2018-02-28
Optimize boolification methods

zoffixznet:
This started out exciting, but after more thorough measurements, I think this should be rejected and instead we need to investigate what makes `nqp::p6bool` vs `Num.Bool` so much faster.

--------------------------------------

Replace .Bool calls with nqp::p6bool() in Bool.pm (other areas
of the compiler still use the method call, which could be changed too)

Using `my $x = …; {for ^3000_000 {my $ = ?$x}; say now - ENTER now}` Makes:
- with Num; 9x faster
- with `class {}.new`; 11% faster
- with `class { method Bool { False } }.new` or `but False` of core types; 4.6x SLOWER
- with an unmixed Int or Str, no change


zoffixznet:
> I think this should be rejected and instead we need to investigate what makes nqp::p6bool vs Num.Bool so much faster.

nm, I see it.

PR #1589, opened 2018-03-04
Don't call Regexes in HyperRace's GrepCode

MasterDuke17:
They should be matched against instead. Fixes the `No such method '!cursor_start' for invocant of type 'Str'` error when doing `say +lines().race.grep(/en/)`.


zoffixznet:
Since `&matcher` doesn't change, we don't need to test if `&matcher` is a `Regex` for each item we're grepping and can do it just once.

FWIW, I have a commit doing that just once; I can just commit that, if you like.

MasterDuke17:
Doh, of course. Feel free to fix and close this.

zoffixznet:
OK, I'll commit tests and fix, once I figure out a working version. FWIW, merely matching with regex still has issues due to parallelaziation: https://irclog.perlgeek.de/perl6-dev/2018-03-04#i_15882720

PR #1600, opened 2018-03-08
Make MONKEY-TYPING imply `no precompilation`

jnthn:
It's not possible to load two precompiled modules that both do a
monkey-patch to the same type. Precompiling them will thus result in
modules that cannot be loaded together with others.

So far we've expected module authors to read the docs and remember to
add `no precompilation`. That's less reliable than the compiler just
doing the right thing for them. Especially given that the problem will
probably never be noticed by a module author when writing tests for
their module, but instead when somebody uses that module together with
others in a larger application.

### DIFF
--- a/src/Perl6/World.nqp
+++ b/src/Perl6/World.nqp
@@ -952,6 +952,12 @@ class Perl6::World is HLL::World {
       'worries',            1,
     );
 
+    # pragmas that imply `no precompilation`
+    my %no_precompilation_pragmas := nqp::hash(
+        'MONKEY',           1,
+        'MONKEY-TYPING',    1,
+    );
+
     # not yet implemented pragmas
     my %nyi_pragma := nqp::hash(
       'internals',  1,
@@ -1076,6 +1082,12 @@ class Perl6::World is HLL::World {
         else {
             $RMD("  '$name' is not a valid pragma") if $RMD;
             return 0;                        # go try module
+        }
+
+        if nqp::existskey(%no_precompilation_pragmas, $name) {
+            unless $*COMPILING_CORE_SETTING {
+                self.do_pragma($/, 'precompilation', 0, NQPMu);
+            }
         }
 
         $RMD("Successfully handled '$name' as a pragma") if $RMD;


jnthn:
This passes test/spectest fine. Filed as a PR to give folks a day or two for input before it goes in.

zoffixznet:
> It's not possible to load two precompiled modules that both do a monkey-patch to the same type. Precompiling them will thus result in modules that cannot be loaded together with others.

But what if my program *doesn't* have *two* such modules? Would I be then paying the penalty of having to recompile—potentially huge—part of my program each time I run it for no reason?

> So far we've expected module authors

It could make sense from point of view of stuff published into the ecosystem, but I imagine the vast majority of Perl 6 modules code that exists is for some private programs, even designed to be used only with those specific programs. I think there's a fairly low chance that in such programs more than one module would monkey with the same type.

This PR feels like we're trying to force a detrimental behaviour on users who *do* read the documentation, simply because we assume that some won't.

holli-holzer:
Would it be possible to add some kind of adverb to `use`?

    use MonkeyModule :always-compile;

niner:
> Would it be possible to add some kind of adverb to `use`?
>     use MonkeyModule :always-compile;

That'd be possible, yes. But a developer knowing that there's an issue may as 
well add the `no precompilation;` to MonkeyModule.

holli-holzer:
If the author realizes that. Wasn't that point? Such an adverb would provide the module user to fix that kind of mistake. It could also be of other use potentially?

holli-holzer:
Why is it, that "It's not possible to load two precompiled modules that both do a
monkey-patch to the same type"? 

What would happen if you would defer all calls to `compose`, maybe "collecting" them, get rid of possible duplicate types and run them in bulk at the end of the compilation phase?

JJ:
Ping

lizmat:
Pong.  Feels that this can be closed on account of lack of consensus.

jnthn:
Yes, this turned out to be more controversial than I expected. I'm not going to push for it; of all the Raku things I could spend my time on, this is a long way down the list (and probably a better solution is a recommendation manager that avoids things that `augment`).

Issue #1588, opened 2018-03-04
Signals do not match host system (NativeCall kill issue)


jstuder-gh:
## The Problem

There isn't anything wrong with the idea of platform-independent signal numbers but when mixed with NativeCall the results are Less Than Awesome.

    use NativeCall;
    sub kill(int32, int32) is native {*}

    my $supply = signal(SIGUSR1);
    my $tap = $supply.tap: { say 'sigusr1 here' };
    kill($*PID, SIGUSR1);
    $tap.close;

    # OUTPUT: «Power Failure␤»

On my system (GNU/Linux), signal 30 (SIGUSR1 in MoarVM) is SIGPWR.

Since NativeCall is a heavily-advertised feature and it's usage is encouraged, I think that we should not use platform-independent signal numbers and instead have the Signal enum's values match the host system.

jstuder-gh:
I do have a proposed fix for this although I wanted to run it by the community to see that it was agreed both that the current functionality should change and that this method of fixing it is a good way to go about it.

MoarVM:
<https://github.com/MoarVM/MoarVM/compare/master...jstuder-gh:sig_hash>

NQP:
<https://github.com/perl6/nqp/compare/master...jstuder-gh:sig_hash>

Rakudo:
<https://github.com/rakudo/rakudo/compare/master...jstuder-gh:sig_hash>

This development maps the MoarVM supported signal names to the proper value on the host system. If a signal is not supported on the host, the number is set to zero (kill's null signal; no signal sent).

It also sets a bit-flag indicating the mapped numbers of the MoarVM supported  signals that we can check before setting the signal handler. Finally, it creates an nqp op (getsignals) that exposes the mapped signal numbers to the HLL when called.

On the Rakudo side, the Signal enum is populated with the data retrieved via the nqp::getsignals op. This also allows us to remove the signal mapping code from the file as well, making it easier to read IMO.

The results of issue's example are below. (Note: the Moar code above does not include the output of update_ops which is necessary for it to compile and for the example below to work)

    use NativeCall;
    sub kill(int32, int32) is native {*}
    
    my $supply = signal(SIGUSR1);
    my $tap = $supply.tap: { say 'sigusr1 here' };
    kill($*PID, SIGUSR1);
    $tap.close;

    # OUTPUT: «sigusr1 here␤»

[...]

PR #1601, opened 2018-03-08
Issue 1588: Populate Signal enum with getsignals op

jstuder-gh:
Previously, Rakudo and it's backends used system-independent signal
values and transformed them to the proper value when submitting a signal
handler. However when used in conjunction with NativeCall and the native
C kill implementation, the fact that the Signal values do not match the
host system becomes a problem.

Making it so that the signal numbers exposed to the user as Signal enums
will match those of their host system. They can be set via the
getsignals op, which will retrieve the valid signals from the host.

See <https://github.com/rakudo/rakudo/issues/1588>

### DIFF
--- a/src/core/signals.pm6
+++ b/src/core/signals.pm6
@@ -1,4 +1,17 @@
-my enum Signal ( |nqp::getsignals() );
+my enum Signal (
+    |nqp::stmts(
+        ( my $res  := nqp::hash ),
+        ( my $iter := nqp::iterator(nqp::getsignals) ),
+        nqp::while(
+            $iter,
+            nqp::stmts(
+                ( my $p := nqp::shift($iter) ),
+                nqp::bindkey($res, nqp::iterkey_s($p), nqp::abs_i(nqp::iterval($p)) )
+            )
+        ),
+        $res
+    )
+);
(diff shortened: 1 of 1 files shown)


jstuder-gh:
Relies on PRs <https://github.com/MoarVM/MoarVM/pull/814> and <https://github.com/perl6/nqp/pull/430>

jstuder-gh:
Now that the [MoarVM component](https://github.com/MoarVM/MoarVM/pull/814) of this development has been reviewed and merged, requesting a merge of the related [NQP PR](https://github.com/perl6/nqp/pull/430) and this. Doing so will expose the mapped signums to P6 users via the Signal enum, addressing [Issue 1588](https://github.com/rakudo/rakudo/issues/1588). Thanks.

jstuder-gh:
@lizmat, I noticed that my NQP PR regarding signals was merged. That PR removes the NQP signal constants (as they are no longer accurate), which is causing the errors seen on Rakudo. Please merge in this PR and that should stop the errors (in addition to setting all the Signal enum values :+1: )

lizmat:
And bumped, rakudo is green again

PR #1608, opened 2018-03-11
Make 2 elem infix:<,> special cases more generic

MasterDuke17:
And extract `infix:<,>` `Slip` handling to a common sub.

### DIFF
--- a/src/core/List.pm6
+++ b/src/core/List.pm6
@@ -1489,6 +1489,19 @@ my class List does Iterable does Positional { # declared in BOOTSTRAP
     }
 }
 
+sub INFIX_COMMA_SLIP_HELPER(\reified, \future) {
+    nqp::stmts(
+      (my $list :=
+        nqp::p6bindattrinvres(nqp::create(List),List,'$!reified',reified)),
+      nqp::bindattr($list,List,'$!todo',
+        my $todo:= nqp::create(List::Reifier)),
+      nqp::bindattr($todo,List::Reifier,'$!reified',reified),
+      nqp::bindattr($todo,List::Reifier,'$!future',nqp::getattr(future,List,'$!reified')),
+      nqp::bindattr($todo,List::Reifier,'$!reification-target',reified),
+      $list
+    )
+}
+
 # The , operator produces a List.
 proto sub infix:<,>(|) is pure {*}
 multi sub infix:<,>() { nqp::create(List) }
@@ -1496,14 +1509,7 @@ multi sub infix:<,>(Slip:D \a, Slip:D \b) {
     nqp::stmts(  # Slip seen, first copy non-slippy things
       (my $reified := nqp::create(IterationBuffer)),
       # now set up the List with a future
-      (my $list :=
-        nqp::p6bindattrinvres(nqp::create(List),List,'$!reified',$reified)),
-      nqp::bindattr($list,List,'$!todo',
-        my $todo:= nqp::create(List::Reifier)),
-      nqp::bindattr($todo,List::Reifier,'$!reified',$reified),
-      nqp::bindattr($todo,List::Reifier,'$!future',nqp::list(a,b)),
-      nqp::bindattr($todo,List::Reifier,'$!reification-target',$reified),
-      $list
+      INFIX_COMMA_SLIP_HELPER($reified, nqp::list(a,b))
     )
 }
 multi sub infix:<,>(Any \a, Slip:D \b) {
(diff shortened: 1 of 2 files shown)


MasterDuke17:
`my $a; my $b; ($a,$b) = "a","b" for ^1000000; say now - INIT now` slows slightly from 3.6s to 3.8s (still faster than the 4.2s it was before https://github.com/rakudo/rakudo/pull/1317).

`my ($a, $b) = .words[1,3] for "50k.sql".IO.lines; say "|$a|,|$b|"; say now - INIT now` is unchanged. 

`my $a = Rat.new; my $b = Rat.new; my $c; my $d; ($c,$d) = $a,$b for ^1000000; say now - INIT now` goes from 5.7s to 4.7s.

src/core/List.pm6
```diff
@@ -1489,13 +1489,45 @@ my class List does Iterable does Positional { # declared in BOOTSTRAP
...
+}
+
 # The , operator produces a List.
 proto sub infix:<,>(|) is pure {*}
 multi sub infix:<,>() { nqp::create(List) }
-multi sub infix:<,>(Int \a, Int \b) is default {
-    nqp::p6bindattrinvres(nqp::create(List),List,'$!reified',nqp::list(a,b))
+multi sub infix:<,>(Slip:D \a, Slip:D \b) {
+    nqp::stmts(  # Slip seen, first copy non-slippy things
+      (my $reified := nqp::create(IterationBuffer)),
```
zoffixznet: Not sure I'm following that comment. Both things are slippy, aren't they?

src/core/List.pm6
```diff
@@ -1489,13 +1489,45 @@ my class List does Iterable does Positional { # declared in BOOTSTRAP
...
+    nqp::stmts(  # Slip seen, first copy non-slippy things
+      (my $reified := nqp::setelems(nqp::create(IterationBuffer),1)),
+      nqp::bindpos($reified,0,a),
+      # now set up the List with a future
+      INFIX_COMMA_SLIP_HELPER($reified, nqp::list(b))
+    )
+}
+multi sub infix:<,>(Slip:D \a, Any \b) {
+    nqp::stmts(  # Slip seen, first copy non-slippy things
+      (my $reified := nqp::create(IterationBuffer)),
```
zoffixznet: This comment too smells of copy-pasta, since it doesn't appear we're doing anything with non-slippy things first.

zoffixznet:
Any time changes for things with slips, like `my ($a, $b) = 42, Empty` ?

MasterDuke17:
> Any time changes for things with slips, like my ($a, $b) = 42, Empty ?

`my ($a, $b) = 42, Empty for ^1000000; say now - INIT now` decreases from 5.1s to 3.9s.

zoffixznet:
Sweet. This code is over my head to review. Maybe @lizmat can.

src/core/List.pm6
```diff
@@ -1489,13 +1489,45 @@ my class List does Iterable does Positional { # declared in BOOTSTRAP
     }
 }
 
+sub INFIX_COMMA_SLIP_HELPER(\reified, \future) {
```
jnthn: Please put this as a method in `Rakudo::Internals`, so that we don't leak it.

src/core/List.pm6
```diff
@@ -1489,13 +1489,45 @@ my class List does Iterable does Positional { # declared in BOOTSTRAP
...
+      # now set up the List with a future
+      INFIX_COMMA_SLIP_HELPER($reified, nqp::list(b))
+    )
+}
+multi sub infix:<,>(Slip:D \a, Any \b) {
+    nqp::stmts(
+      (my $reified := nqp::create(IterationBuffer)),
+      # now set up the List with a future
+      INFIX_COMMA_SLIP_HELPER($reified, nqp::list(a,b))
+    )
```
jnthn: Could this not be simpler as `INFIX_COMMA_SLIP_HELPER(nqp::create(IterationBuffer), nqp::list(a,b))`?

src/core/List.pm6
```diff
@@ -1489,13 +1489,45 @@ my class List does Iterable does Positional { # declared in BOOTSTRAP
...
+multi sub infix:<,>(Slip:D \a, Slip:D \b) {
+    nqp::stmts(
+      (my $reified := nqp::create(IterationBuffer)),
+      # now set up the List with a future
+      INFIX_COMMA_SLIP_HELPER($reified, nqp::list(a,b))
+    )
+}
+multi sub infix:<,>(Any \a, Slip:D \b) {
+    nqp::stmts(  # Slip seen, first copy non-slippy thing
+      (my $reified := nqp::setelems(nqp::create(IterationBuffer),1)),
```
jnthn: I don't think the `setelems` helps here much? It probably stops it rounding up to, I think, 4, which when there's something slipped in after seems like a less than great guess. Sure, it could be `Empty`, but it might not be.

MasterDuke17:
@jnthn new commit pushed with your tweaks. And too bad about `INFIX_COMMA_SLIP_HELPER`, I was really hoping its addition would be the highlight of the next release notes.

lizmat:
I'm surprised that the `my class List::Reifier { ... }` is needed.  Is it?

Oddly enough, I seem to be unable to find where `List::Reifier` actually comes to existence.

zoffixznet:
> Oddly enough, I seem to be unable to find where List::Reifier actually comes to existence.

I think it's just defined as `class Reifier` inside `class List`: 
https://github.com/rakudo/rakudo/blob/master/src/core/List.pm6#L19

lizmat:
Ah, duh.  :-)

No further comments from me then.

PR #1609, opened 2018-03-12
Fix Metamodel shortname assignments

### LINKED ISSUE #1606: Wrong output of in CArray.^shortname
```
<Zoffix__> m: use NativeCall; say CArray[Pointer[int32]]
<camelia> rakudo-moar 26522e8ac: OUTPUT: «(CArray[Pointer)␤»
```

Needs a fix in this code somewhere: https://github.com/rakudo/rakudo/blob/26522e8acd096412325f6c667c823c675f2fbae6/src/Perl6/Metamodel/Naming.nqp#L5-L24

Kaiepi:
WIP. Please review before merging.

Fixes #1606

### DIFF
--- a/src/Perl6/Metamodel/Naming.nqp
+++ b/src/Perl6/Metamodel/Naming.nqp
@@ -5,18 +5,22 @@ role Perl6::Metamodel::Naming {
         $!name := $name;
         nqp::setdebugtypename($obj, $name);
         if $name {
-            my @names := nqp::split('[', $name);
-            if nqp::elems(@names) > 1 {
-                my @main := nqp::split('::', @names[0]);
-                my @sub  := nqp::split('::', @names[1]);
-                $!shortname :=
-                  @main[nqp::elems(@main) - 1]
-                    ~ '[' ~ @sub[nqp::elems(@sub) - 1];
+            while (my int $colon := nqp::rindex($name, '::')) != -1 {
+                my int $paren := nqp::rindex($name, '[', $colon - 1);
+                my int $comma := nqp::rindex($name, ',', $colon - 1);
+                my int $chop-start :=
+                    ($paren < 0 && $comma < 0)
+                    ?? 0
+                    !! ($paren != -1 && $paren < $comma)
+                        ?? $comma + 1
+                        !! $paren + 1;
+                $name := nqp::concat(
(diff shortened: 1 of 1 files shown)


src/Perl6/Metamodel/Naming.nqp
```diff
@@ -5,18 +5,26 @@ role Perl6::Metamodel::Naming {
...
-            }
-            else {
-                my @parts := nqp::split('::', $name);
-                $!shortname := @parts[nqp::elems(@parts) - 1];
+            while (my int $colon := nqp::rindex($name, '::')) != -1 {
+                my int $paren := nqp::rindex($name, '[', nqp::sub_i($colon, 1));
+                my int $comma := nqp::rindex($name, ',', nqp::sub_i($colon, 1));
+                my int $chop-start := nqp::if(
+                    nqp::islt_i($paren, 0) && nqp::islt_i($comma, 0),
+                    0,
```
zoffixznet: Since this is NQP code, there's no benefit from writing `$paren < 0` as `nqp::islt_i`; in NQP it already compiles down to that.

The same applies to `+`, `-`,  `!=`, `~`, and `?? !!` ops in this commit that are written using word-based ops.

You can see they get written into the word-based ops with:

```bash
$ nqp --target=optimize -e 'my int $paren := 0; $paren + 0 < 42'

[...]
- QAST::Op(islt_i &infix:«<»)  <
            - QAST::Op(add_i &infix:<+>)  +
              - QAST::Var(local __lowered_lex_2) 
              - QAST::IVal(0) 
```
Notice the `QAST::Op(islt_i` and `QAST::Op(add_i` in the list.

Kaiepi: Ah. When is it appropriate to use `nqp::if` and the like?

zoffixznet: In the part of the full-blown Perl 6 code of the compiler, they can be used when among other other nqp, to help consistency/readability. But even in full-blown code stuff like `&&`, `||`, and `?? !!` get rewritten to `nqp::if`/`nqp::unless`  (not always the case for `+`, `-` and the rest, which is why you might see word-based ops for those used frequently in full-Perl-6-portion of the compiler).

zoffixznet:
👍 Thanks!

PR #1611, opened 2018-03-12
Add support to Encoding::Decoder and Buf.decode to use replacements

samcv:
This also makes sure we setup decodestream on MoarVM to explicitly use
permissive decoding, so that is set properly when my decodestream patch
lands in MoarVM.

Add support for replacement and strict setting with Buf.decode

Allows the user to set a replacement that is used when encountering an
undecodable character. Allows setting strict to throw (or replace if a
replacement is provided) on an undecodable character.

Replacement support currently only working with windows-1252 and
windows-1251 at the moment.


lizmat:
Wow

PR #1615, opened 2018-03-12
Throw when opening read-only files with truncate and create flags

### LINKED ISSUE #1614: .open with :truncate/:create violates "read only" default open mode
There ~~are~~ [were](https://github.com/perl6/roast/commit/24822cf173cfbdbf859efd4121ccebad9a70a5a3) some proptests in S32-io/open.t that test `:truncate` and `:create` by their lonesome don't throw. But since the default open mode is "read-only", it doesn't make much sense to me that something opened only for reading would be created and especially truncated.

The throwage actually does occur on JVM:

```
$ ./perl6-j -e '"/tmp/zdsdsdsxczxsdzdasdsadsds".IO.open: :truncate, :create;'
java.nio.file.NoSuchFileException: /tmp/zdsdsdsxczxsdzdasdsadsds
  in block <unit> at -e line 1

Actually thrown at:
  in block <unit> at -e line 1
```

And apparently on MoarVM too sometimes: https://irclog.perlgeek.de/perl6-dev/2018-03-12#i_15913886

But on my Bodhi Linux box, it just silently creates/truncates:

```
$ perl6 -e '"/tmp/zdsdsdsxczxsdzdasdsadsds".IO.open: :truncate, :create;'                                                                                                                                    
$ ls /tmp/zdsdsdsxczxsdzdasdsadsds
/tmp/zdsdsdsxczxsdzdasdsadsds
```

Also, depsite JVM refusing to work with non-existant file, it silently fails to :truncate, if the file already exists:
```
$ cat /tmp/foo
abc
2018.02.1 zoffix@VirtualBox~/CPANPRC/judo (master)$ ./perl6-j -e '"/tmp/foo".IO.open: :truncate, :create;'
2018.02.1 zoffix@VirtualBox~/CPANPRC/judo (master)$ cat /tmp/foo
```

-----

So we need to give a closer look at the modes and consistify their behaviour.

-----

Some discussion on IRC: https://irclog.perlgeek.de/perl6-dev/2018-03-12#i_15913886

Kaiepi:
Fixes #1614

### DIFF
--- a/src/core/IO/Handle.pm6
+++ b/src/core/IO/Handle.pm6
@@ -117,6 +117,14 @@ my class IO::Handle {
         );
 
         nqp::if(
+            nqp::if(:$truncate, :$create),
+            nqp::if(
+                nqp::iseq_s($mode, 'ro'),
+                die("Cannot open read-only file with flags ':truncate :create'")
+            )
+        );
+
+        nqp::if(
           nqp::iseq_s($!path.Str, '-'),
           nqp::if(
             nqp::iseq_s($mode, 'ro'),


src/core/IO/Handle.pm6
```diff
@@ -116,6 +116,14 @@ my class IO::Handle {
           ),
         );
 
+        nqp::if(
+            nqp::if(:$truncate, :$create),
```
zoffixznet: `:$truncate` creates a Pair. That's not needed here. Also, this doesn't fix cases when only `$truncate` or only `$create` is specified.

You can also just use `&&` and `||` operators. They get converted to `nqp::if` and `nqp::unless` (the rest of this routine here does use too many `nqp::if`s because when I wrote that stuff I didn't know `&&` and `||` got converted to nqp ops)

So you can just write `($truncate || $create) && die ...` or `die ... if $truncate || $create` (that `if` would also get converted to `nqp::if`)

zoffixznet: To clarify, I think both `:truncate` and `:create`, even when used individually don't make much sense in read-only mode.

Kaiepi: I need to study the internals of nqm more so I can write optimized code the right way.

>To clarify, I think both :truncate and :create, even when used individually don't make much sense in read-only mode.

>  To clarify, I think both :truncate and :create, even when used individually don't make much sense in read-only mode.

I agree.

zoffixznet: > I need to study the internals of nqm more so I can write optimized code the right way.

You can dump QAST trees with `--target=optimize` and see what the stuff you write comes out as. It takes some time to get used to, but once you do, it's easy to see what gets optimized to what. There's an article on that: https://rakudo.party/post/Perl-6-Core-Hacking-QASTalicious

zoffixznet:
Thanks, but I think we'll leave this area as undefined behaviour, at least for now. Full response in https://github.com/rakudo/rakudo/issues/1614#issuecomment-373799149

PR #1617, opened 2018-03-14
Add strict, replacement options for IO::Handle, Str.encode & other fixes

samcv:
* Combine Encoding::Encoder::Builtin::Replacement and
Encoding::Encoder::Builtin
 into one class. They were essentially duplicated, and we can achieve the
same
 goal without adding any additional overhead.

* Encoding.decoder:  make sure :!replacement does not attempt to use a
 replacement. If we get Bool False, we don't want to use a replacement.
 Also make sure the methods accept Bool's and don't assume we will only
 get a Str type. This makes it act identically to how Encoding.encoder
 processes the replacement option.

* Throw during encoder setup on JVM when trying to use a replacement
 or :strict instead of throwing in encode-chars. Will speed things up
 if encode-chars is called many times.

* Add :strict and :replacement options to IO::Handle.encoding()

* Add :strict option to Str.encode

Fix bug mapping windows1252/windows1251 without a dash

This bug did not result in any regression, since we did not accept
windows-1252 or windows-1251 without the dash until very recently. By
accident windows1252 was mapping to windows-1251 and windows1251 was
mapping to windows-1252.

### DIFF
--- a/src/core/Encoding/Encoder/Builtin.pm6
+++ b/src/core/Encoding/Encoder/Builtin.pm6
@@ -1,44 +1,35 @@
 my class Encoding::Encoder::Builtin does Encoding::Encoder {
     has str $!encoding;
     has Blob $!type;
+    has $!replacement;
+    has int $!config;
 
-    method new(Str $encoding, Blob:U $type) {
-        nqp::create(self)!setup($encoding, $type)
+    method new(Str $encoding, Blob:U $type, :$replacement, :$strict) {
+        nqp::create(self)!setup($encoding, $type, :$replacement, :$strict)
     }
-
-    method !setup($encoding, $type) {
+    method !setup($encoding, $type, :$replacement, :$strict) {
         $!encoding = $encoding;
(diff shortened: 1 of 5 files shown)


src/core/Encoding/Encoder/Builtin.pm6
```diff
@@ -1,44 +1,35 @@
...
-        nqp::encoderep($str, $!encoding, $!replacement, nqp::create($!type))
+        nqp::encoderepconf($str,
+            $!encoding,
+            $!replacement,
+            nqp::create($!type),
+            $!config)
 #?endif
 #?if !moar
-        X::NYI.new(feature => 'encoding with replacement').throw
+        nqp::encode($str, $!encoding, nqp::create($!type));
```
jnthn: This seems to result in silent failure to use the replacement rather than an NYI error?

jnthn: Ah, right, good. :-)

PR #1641, opened 2018-03-22
Simplify List's Reifier.new

MasterDuke17:
Use nqp::splice instead of doing nqp::push+nqp::atpos in a loop.

Also use nqp::p6bindattrinvres instead of nqp::bindattr + return.


lizmat:
Thanks!

PR #1643, opened 2018-03-23
Optimize Array's !splice-save

MasterDuke17:
Pull an nqp::getattr out of a loop.

Also use nqp::p6bindattrinvres instead of nqp::bindattr + return.

### DIFF
--- a/src/core/Array.pm6
+++ b/src/core/Array.pm6
@@ -1047,14 +1047,13 @@ my class Array { # declared in BOOTSTRAP
               nqp::bindattr((my $saved:= nqp::create(self)),List,'$!reified',
                 (my $buffer :=
                   nqp::setelems(nqp::create(IterationBuffer),removed))),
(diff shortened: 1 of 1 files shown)


lizmat:
Did you notice a performance improvement when the getattr was moved outside of the loop?

I seem to recall that not having an extra variable actually improved performance in this case.

lizmat:
Thanks!

MasterDuke17:
Hm, I definitely ran a micro-benchmark that showed using getattr in a loop was significantly slower, but I can't find anything in my history that exercised this exact case.

MasterDuke17:
Just tried `my @a = ^200000; my $s = now; my @c = @a.splice($_, 2) for ^60000; say now - $s`. With getattr in the loop I got an average of about 4.03s, with it pulled out I got an average of about 4.02s. But there was enough noise I couldn't swear that any one was actually faster.

Issue #1583, opened 2018-03-02
tools/install-dist.pl is a perl 6 script but has Prolog file extension


AlexDaniel:
See these commits:
* https://github.com/rakudo/rakudo/commit/ade7c4660f46a2b12425de24f66665af809826c2
* https://github.com/rakudo/rakudo/commit/98e575de00db42101d4879e43fe740a34791c418
* https://github.com/rakudo/rakudo/commit/b7df58f25ba47bd404b19005626b6b0058435ec1
* https://github.com/rakudo/rakudo/commit/efd0d6c0516373c1c7a368854b7bc5d43fe463a0

We are now consistently using `.pm6` for modules and `.p6` for scripts, except for [`tools/install-dist.pl`](https://github.com/rakudo/rakudo/blob/efd0d6c0516373c1c7a368854b7bc5d43fe463a0/tools/install-dist.pl) file which is a perl 6 script but confusingly it has a `.pl` extension.

Renaming it is easy, but I'm suspecting that changing it may break some things outside of the rakudo repo. Any ideas on what is affected exactly?

niner:
Distro packages certainly reference the script. E.g. https://build.opensuse.org/package/view_file/devel:languages:perl6/rakudo/rakudo.spec?expand=1 line 50

CurtTilmes:
Alpine distro uses it to install [zef](https://git.alpinelinux.org/cgit/aports/tree/testing/zef/APKBUILD)
but easy enough to update with next release.

AlexDaniel:
@stmuk @robertlemmen @ugexe @nxadm `install-dist.pl` was just renamed to `install-dist.p6` in a PR for https://github.com/rakudo/rakudo/issues/2077 and that went unnoticed by me. Is this affecting you by any chance? I'd say we should leave the rename in there, but what exactly can break and what should we fix now?

stmuk:
@AlexDaniel This shouldn't affect star since I use zef to install itself but I think there are upstream packagers who use it or at least some did ask on IRC about using it (maybe Debian/pkgsrc)?  It means packagers can install modules using base rakudo without zef

A quick search on http://colabti.org/irclogger/irclogger_log/perl6?date=2018-06-05#l340 suggests arch maybe use it as well?

Maybe we could revert this change but warn to stderr that the name will change?

robertlemmen:
I think we also rename it during the installation into the package... 

[...]

PR #1652, opened 2018-03-25
Rename tools/install-dist.pl to install-dist.p6

MasterDuke17:


AlexDaniel:
We have a ticket for it: https://github.com/rakudo/rakudo/issues/1583

I'm not sure how to do this rename safely. Maybe we have to leave a symlink for the old file?

MasterDuke17:
Fixed in https://github.com/rakudo/rakudo/pull/2102, so closing this.

AlexDaniel:
Thanks, @MasterDuke17, and sorry :)

PR #1664, opened 2018-03-28
Implement fails-like routine in Test.pm6

zoffixznet:
Same as throws-like, except checks the code returns an armed Failure.

Failures are a ubiquitous feature of the language, yet there's no
convenient way to test for them:

- isa-ok: only tests for type, not the exception contained within,
    and doesn't mark Failures as handled causing warnings
    when those Failures are GCed
- throws-ok: explodes Failures, making it impossible to differentiate
    between code that throws and code that fails (we already had a bug
    in core that wasn't caught by roast for this reason)
- fails-like helper routine is already used extensively in roast,
    showing there is real need for this routine

### DIFF
--- a/lib/Test.pm6
+++ b/lib/Test.pm6
@@ -628,6 +628,25 @@ sub throws-like($code, $ex_type, $reason?, *%matcher) is export {
     }, $reason // "did we throws-like $ex_type.^name()?";
 }
 
+sub fails-like (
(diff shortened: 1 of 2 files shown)


zoffixznet:
Docs:  https://github.com/perl6/doc/pull/1876
Tests: https://github.com/perl6/roast/pull/408

PR #1665, opened 2018-03-28
Make resolved private method calls callstatic

zoffixznet:
- This makes resolved private method calls 11% faster than public method calls.
- This also passes stresstest and to best of knowledge of staticalization I couldn't repro any negative effects, e.g. `with class Foo { my $x = 42; method !z { say $x }; method z { self!z }; method mod { $x = $++ }; }.new { for ^100 -> $x { .z; .mod; } }` seems to update `$x` just fine
- I've no idea if it's acceptable to actually staticalize these calls. If not, I'd love to know why, to better understand when `callstatic` is suitable

### DIFF
--- a/src/Perl6/Optimizer.nqp
+++ b/src/Perl6/Optimizer.nqp
@@ -1899,7 +1899,7 @@ class Perl6::Optimizer {
                         $op.shift; $op.shift; # name, package (both pre-resolved now)
                         $op.unshift($inv);
                         $op.unshift($call);
-                        $op.op('call');
+                        $op.op('callstatic');
                         $op.name(NQPMu);
                     }
                 }


MasterDuke17:
> This makes resolved static method calls 11% faster than private method calls.

Is that supposed to be "faster than *public* method calls."?

zoffixznet:
> Is that supposed to be "faster than public method calls."?

Yeah. Thanks.

lizmat:
FWIW, it spectests clean for me.  Can't wait to put private methods back!  :-)

src/Perl6/Optimizer.nqp
```diff
@@ -1899,7 +1899,7 @@ class Perl6::Optimizer {
                         $op.shift; $op.shift; # name, package (both pre-resolved now)
                         $op.unshift($inv);
                         $op.unshift($call);
-                        $op.op('call');
+                        $op.op('callstatic');
                         $op.name(NQPMu);
```
jnthn: I'm decidedly confused. :-) Here's the code that compiles `callstatic` differently from `call`:

https://github.com/perl6/nqp/blob/b6d98eac3f0dff1dbd4df252351efa35ab700f7e/src/vm/moar/QAST/QASTOperationsMAST.nqp#L1444

Notice how it is contained within a `if $op.name { ... }` block. All it's doing is instructing the VM that it can cache the lexical lookup, because it will always return the exact same code object (which is why we use it for things from the setting scope and `UNIT` scope - because they can never be closure-cloned).

Given that, I don't currently see how this change could affect the generated code, and so lead to a performance difference. I wonder what I'm missing...

jnthn: (Also, to be clear, I didn't yet attempt to replicate the performance difference locally.)

PR #1673, opened 2018-03-30
use tmpdir for precompilation / REPL history if home directory is readonly

kbucheli:
If the home directory is readonly, then precompilation does not work and the REPL history cannot be saved. In that case I use a per user directory in $*TMPDIR.

This fixes #127031


src/core/CompUnit/RepositoryRegistry.pm6
```diff
@@ -96,7 +96,12 @@ class CompUnit::RepositoryRegistry {
                      ?? nqp::atkey($ENV,'HOMEPATH') !! '')
                  ) -> $home-path {
                 $home = "$home-path/.perl6";
-                my str $path = "inst#$home";
+                nqp::mkdir($home, 0o700) unless nqp::stat($home, nqp::const::STAT_EXISTS);
+
+                if !nqp::stat($home, nqp::const::STAT_ISDIR) {
+                    $home = "$*TMPDIR/.perl6_{+$*USER}";
```
niner: Predictable paths in /tmp are a common security issue.

src/core/REPL.pm6
```diff
@@ -402,9 +402,18 @@ do {
...
 
-            $!history-file = $*ENV<RAKUDO_HIST>
-                ?? $*ENV<RAKUDO_HIST>.IO
-                !! ($*HOME || $*TMPDIR).add('.perl6/rakudo-history');
+            with $*ENV<RAKUDO_HIST> {
+                $!history-file = $_.IO;
+            }
+            else {
+                # looking up the home thingy from the CompUnit Repository Registry would be more elegant
+                my IO::Path $home = "$*HOME/.perl6".IO;
```
zoffixznet: Without commenting on the rest of the PR...  Note that `$*HOME` can be undefined, in which case you'll be attempting to create `.perl6` in the root dir:

```
2018.03.129 zoffix@VirtualBox~$ unset HOME
2018.03.129 zoffix@VirtualBox/home/zoffix$ perl6 -e 'say "$*HOME/.perl6".IO'                                                                                                                                                                                   
Use of Nil in string context
  in block <unit> at -e line 1
"/.perl6".IO
```

src/core/CompUnit/RepositoryRegistry.pm6
```diff
@@ -96,7 +96,17 @@ class CompUnit::RepositoryRegistry {
                      ?? nqp::atkey($ENV,'HOMEPATH') !! '')
                  ) -> $home-path {
                 $home = "$home-path/.perl6";
-                my str $path = "inst#$home";
+                nqp::mkdir($home, 0o700) unless nqp::stat($home, nqp::const::STAT_EXISTS);
```
JJ, line 99: Are we sure we want this side effect?

kbucheli: Do you have any better suggestion to figure out if the home directory can be written to?

treyharris:
On Mon, May 4, 2020 at 16:17 Konrad Bucheli <notifications@github.com>
wrote:

> *@kbucheli* commented on this pull request.
> ------------------------------
>
> In src/core/CompUnit/RepositoryRegistry.pm6
> <https://github.com/rakudo/rakudo/pull/1673#discussion_r419701434>:
>
> > @@ -96,7 +96,17 @@ class CompUnit::RepositoryRegistry {
>                       ?? nqp::atkey($ENV,'HOMEPATH') !! '')
>                   ) -> $home-path {
>                  $home = "$home-path/.perl6";
> -                my str $path = "inst#$home";
> +                nqp::mkdir($home, 0o700) unless nqp::stat($home, nqp::const::STAT_EXISTS);
>
> Do you have any better suggestion to figure out if the home directory can
> be written to?
>

As a general defensive programming technique, one should, whenever
possible, just attempt an operation and catch it if it fails. This protects
not only against race conditions but from permission errors due to ACL’s,
foreign OS, filesystem or fileserver translations, etc. (It’s best if one
has a way to detect “filesystem full” vs. “permission error”, because
ordinarily, you want to just fail in the former exception and gracefully
fall back in the latter.)

src/core.c/REPL.pm6
```diff
@@ -425,9 +425,16 @@ do {
                     $!history-file = $new;
                 }
 
-                without mkdir $!history-file.parent {
-                    note "I ran into a problem trying to set up history: {.exception.message}";
-                    note 'Sorry, but history will not be saved at the end of your session';
+                if  !mkdir($!history-file.parent) || ! $!history-file.parent.d {
```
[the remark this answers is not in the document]
lizmat: The exact semantics of `mkdir` have been under scrutiny many times already.  I've lost track *and* got severely burnt out on anything related to refactoring IO related stuff.

lizmat:
@ugexe @niner Any objections against merging?

src/core.c/CompUnit/RepositoryRegistry.pm6
```diff
@@ -148,6 +148,17 @@ class CompUnit::RepositoryRegistry {
              )
            ) -> $home-path {
             $home = $home-path ~ $sep ~ '.raku';
+            # check if we can write the home directory, else fall back to temp directory
+            nqp::mkdir($home, 0o700) unless nqp::stat($home, nqp::const::STAT_EXISTS);
+            if !nqp::stat($home, nqp::const::STAT_ISDIR) {
+                my $tmp = "$*TMPDIR/.raku_{+$*USER}";
```
niner, line 154: This still suffers from predictable temp directory names. The attacker can know the names of directories and files we are going to create. Thus they will be able to create directories before use and put a symlink in the place of a precomp (or other) file we're going to create. This will cause us to overwrite an arbitrary file chosen by the attacker.

niner (changes requested):

src/core.c/CompUnit/RepositoryRegistry.pm6
jonathanstowe: Yeah, it probably needs something like https://modules.raku.org/dist/File::Temp:zef:zef:rbt/lib/File/Temp.rakumod#L38

PR #1674, opened 2018-03-31
REPL: exit by pressing ^D twice, not once

Kaiepi:
This allows clearing the current line without exiting and prevents typos
from forcing it to exit.

### DIFF
--- a/src/core/REPL.pm6
+++ b/src/core/REPL.pm6
@@ -303,8 +303,8 @@ do {
         method interactive_prompt() { '> ' }
 
         method repl-loop(*%adverbs) {
-
-            say "To exit type 'exit' or '^D'";
+            my $exit_char = ($*VM.osname eq 'win32') ?? '^Z' !! '^D';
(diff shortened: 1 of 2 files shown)


Kaiepi:
Do not merge this. This has only been tested with Linenoise, and I know it's possible to do with Readline, but isn't as it stands.

Kaiepi:
Closing this since it doesn't work with the default line editor, and can't test Readline since it doesn't get used even when I force it to with env flags.

Kaiepi:
Tested Readline. Using ^D like this works

Kaiepi:
Updated with checks for ^D/^Z depending on OS

JJ:
Sorry, seems to have conflicts. Can you please check and fix them?

vrurg:
Considering that `^Z` support for Windows is already implemented, we probably only need lines 317-332 of the patch.

lizmat:
Is this PR salvageable?

PR #1680, opened 2018-04-01
Rewrite List cmp in NQP for faster execution

jstuder-gh:
* Check that @a and @b are not lazy. If they are fall back to slow path
* In fast path, extract '$!reified' elements and use lower-level ops
  like nqp::elems and nqp::atpos on them instead of high-level
  equivalents

In the slow path, add laziness check. This fixes a bug with comparison
of lazy arrays where one is exhausted, and when the the number of elems
is compared a Failure occurs ("Lazy Lists Cannot .elems")

In order for the laziness check to work in instances where two identical
lazy lists are compared, modify "ZipIterable" so that all the sub-
iterators are pulled an equal number of times, even if one of the
earlier iterators pulled "IterationEnd"

### DIFF
--- a/src/core/List.pm6
+++ b/src/core/List.pm6
@@ -1641,7 +1641,7 @@ proto sub prefix:<|>(|) {*}
 multi sub prefix:<|>(\x) { x.Slip }
 
 sub CMP-SLOW(@a, @b) {
-    (@a Zcmp @b).first(&prefix:<?>) || @a <=> @b
+    (@a Zcmp @b).first(&prefix:<?>) || &infix:<cmp>( |do .is-lazy for @a, @b ) || @a <=> @b
 }
 
 multi sub infix:<cmp>(@a, @b) {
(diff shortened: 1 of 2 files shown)


jstuder-gh:
The script I used to test this PR can be found [here](https://gist.github.com/jstuder-gh/d818edb77742437de1d6336661d4f057), and the data found [here (Warning: large file)](https://gist.github.com/jstuder-gh/ae0b434c071aa9ba74924360c480f2eb). There are 150,000 records total being sorted.

Prior to optimization, the program spent 182,627.59ms (~182 seconds) executing the List cmp method. After optimization, it spent 9,578.55ms (~9 seconds) executing the same method.

The fast path is used only if the data is fully reified and has a quantifiable number of elements, but in common cases of list-based sorting based upon multiple criteria, it should be considerably faster.

The total execution time of the application went from 197,481.3ms (~197 seconds) to 21,928.35ms (~21 seconds)

lizmat:
Good catch!   Thanks!

PR #1706, opened 2018-04-08
Update 05-messages/03-errors.t test for RT #132238

ronaldxs:
RT #132238
Clearer error message for unclosed <> hash quote index operator.


zoffixznet:
👍 Thanks!

PR #1712, opened 2018-04-08
Update 05-messages/03-errors.t simpler and clearer tests for RT#132238

ronaldxs:
Simpler and clearer tests for RT#132238 unclosed $< .


zoffixznet:
👍 Thanks!

PR #1715, opened 2018-04-08
Update 05-messages/03-errors.t to add test for RT #122980.

ronaldxs:
Test RT #122980 check for the invalid "missing semicolon" message. The rakudo versions with the problem were over three years old and the Test libraries were not compatible with 2018 03-errors.t so testing with a symptomatic version had to be manually jury rigged. If test is still wanted think about the best that can be done.


lizmat:
Thanks!

PR #1722, opened 2018-04-10
Make (Set|Bag|Mix)(Hash)? stringification the same

MasterDuke17:
perlawhirl++ for the find.

Rakudo build ok and passes `make m-test m-spectest`.

### DIFF
--- a/src/core/Baggy.pm6
+++ b/src/core/Baggy.pm6
@@ -344,10 +344,10 @@ my role Baggy does QuantHash {
           ),
           nqp::if(
             nqp::istype(self,Bag),
-            'bag()',
+            'Bag()',
             nqp::if(
               nqp::istype(self,Mix),
-              'mix()',
+              'Mix()',
               nqp::concat('().',self.^name)
             )
           )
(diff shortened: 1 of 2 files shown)


0racle:
irclog link for context

https://irclog.perlgeek.de/perl6/2018-04-10#i_16027887

lizmat:
No, this is intentional, because the semantics of `Bag()` and `bag()` are different.

`bag(<a b c>)` takes *all* values at face value and puts it in the bag.

`Bag( <a b c> )` looks at the values and tries to interpret them as `Pair`s if possible.

Note the difference in behaviour:

    $ 6 'dd bag((a => 42, a=> 42))'
    (:a(42)=>2).Bag

    $ 6 'dd Bag((a => 42, a=> 42))'
    ("a"=>84).Bag

AlexDaniel:
By the way, we also have this ticket: [RT #132281](https://rt.perl.org/Ticket/Display.html?id=132281). Please do something with it also when there's a final decision on this PR.

lizmat:
I'm going to close this PR on the following grounds:

- .perl bag/mix *must* be lowercase because of their semantics
- .perl on set *could* be uppercase, but seems more consistent to also keep that lowercase
- adding a comma in .gist on set seems to me a loss of real-estate.  As discussed in other places, .gist is for human consumption.  Separating on space seems logical to me.  But on that last point I could be persuaded by others to add the comma.

Issue #1730, opened 2018-04-13
Range.ACCEPTS fails to accept one of two equivalent Ranges


zoffixznet:
```

07:54 | Tison | m: dd (0..9) ~~ (^10); dd (^10) ~~ (0..9);
-- | -- | --
07:54 | camelia | rakudo-moar 08b951c8f: OUTPUT: «Bool::True␤Bool::False␤»
07:54 |   | brrt joined #perl6
07:55 | Tison | Interesting, so (0..9) and (^10) both Range and they contain same data but not same thing


08:02 | Zoffix | looks like just failing to check the opposite of `!topic.excludes-max && $!excludes-max`
-- | -- | --
08:03 |   | m: use nqp; dd nqp::iseq_i(nqp::add_i(nqp::istrue(1), nqp::istrue(1)),1)
08:03 | camelia | rakudo-moar 08b951c8f: OUTPUT: «0␤»
08:03 | Zoffix | m: use nqp; dd nqp::iseq_i(nqp::add_i(nqp::istrue(1), nqp::istrue(0)),1)
08:03 | camelia | rakudo-moar 08b951c8f: OUTPUT: «1␤»
08:04 | Zoffix | m: use nqp; dd nqp::iseq_i(nqp::add_i(nqp::istrue(0), nqp::istrue(1)),1)
08:04 | camelia | rakudo-moar 08b951c8f: OUTPUT: «1␤»
08:04 | Zoffix | prolly that with topic.excludes-max and $!excludes-max (and the same for mins) would fix the bug  

```

```

08:11 | lizmat | m: dd 9.9 ~~ 0.0..^10.0   # can also have non-int in ranges
-- | -- | --
08:11 | camelia | rakudo-moar 08b951c8f: OUTPUT: «Bool::True␤»

```

W4anD0eR96:
https://github.com/rakudo/rakudo/blob/08b951c8f33cfc702c308e64efcb44f3ec725117/src/core/Range.pm6#L438-L458

(^10).max == 10
(0..9).max == 9
so (^10) ~~ (0..9) meets $max < topic.max and then always be False

zoffixznet:
> (^10) ~~ (0..9) meets $max < topic.max

Note that there `^10` is the `topic` and `0..9` is the `self`. So `topic.max < $!max` is False and it tries the `||` branch.

> no idea how about other types

[...]

PR #1731, opened 2018-04-13
Fix Range.ACCEPTS fails on (Range, Range)

W4anD0eR96:
Follow what we do on (Range, Cool)

Co. #1730


W4anD0eR96:
Tests on perl6/roast#416

Issue #1729, opened 2018-04-13
Range.minmax inconsistency / LTA error


zoffixznet:
If you call `.minmax` on an excluded Int range, you get the minmax values; if you call it on a non-Int range, you get an error saying you can't minmax excluded Ranges.

```
<Zoffix__> m: dd ^5 .minmax
<camelia> rakudo-moar 08b951c8f: OUTPUT: «(0, 4)␤»
<Zoffix__> m: dd ^5.1 .minmax
<camelia> rakudo-moar 08b951c8f: OUTPUT: «Failure.new(exception => X::AdHoc.new(payload => "Cannot return minmax on Range with excluded ends"), backtrace => Backtrace.new)␤»
```

Either the error should mention "non-Int" ranges or both types of ranges should fail.

W4anD0eR96:
I create a PR #1734 that makes

`(^10).minmax` raise Exception, since [0, 10) surely has no max value.

We can use `(^10).int-bounds` to get (0 9) if we want to limit domain in Integer.

AlexDaniel:
We should be careful about this change before the release, especially if we want to add more cases that throw. Probably a deprecation warning suggesting to use `.int-bounds` is needed.

PR #1734, opened 2018-04-14
Improve Range.minmax on excluded Ranges

TisonKun:
Co. #1729

Then `(^10).minmax` raise Exception, since [0, 10) surely has no max value.

We can use `(^10).int-bounds` to get (0 9) if we want to limit domain in Integer.

If this PR be approved, I'll write roast.

### DIFF
--- a/src/core/Range.pm6
+++ b/src/core/Range.pm6
@@ -706,11 +706,9 @@ my class Range is Cool does Iterable does Positional {
     }
 
     multi method minmax(Range:D:) {
-        $!is-int
-          ?? self.int-bounds
-          !! $!excludes-min || $!excludes-max
-            ?? Failure.new("Cannot return minmax on Range with excluded ends")
-            !! ($!min,$!max)
+        $!excludes-min || $!excludes-max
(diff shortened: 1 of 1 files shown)


lizmat:
I think this is wrong because in the case of an integer range, you *can* return `minmax` because it's just one more / less than the bound.

One could argue a similar behaviour should occur for ranges like `"a" ..^ "z"`, but I think that's a very uncommon idiom.

zoffixznet:
> I think this is wrong because in the case of an integer range, you can return minmax because it's just one more / less than the bound.

Would the same be true for other numerics? Take the Int value of .max plus fractional portion of .min and subtract 1 if the result is larger than max?

```
<Zoffix__> m: for 1.1..^5.3, 1.3..^5.1 {  .eager.say; my $max := .max.Int + .min - .min.floor; say $max > .max ?? $max - 1 !! $max }
<camelia> rakudo-moar 430266629: OUTPUT: «(1.1 2.1 3.1 4.1 5.1)␤5.1␤(1.3 2.3 3.3 4.3)␤4.3␤»
```

TisonKun:
... close because no more keep on this.

PR #1766, opened 2018-04-24
Condense three IO::Path.new multis down to one...

MasterDuke17:
by using coercers.

### DIFF
--- a/src/core/IO/Path.pm6
+++ b/src/core/IO/Path.pm6
@@ -32,14 +32,8 @@ my class IO::Path is Cool does IO {
     }
 
     proto method new(|) {*}
-    multi method new(IO::Path: Str $path, :$SPEC = $*SPEC, Str:D :$CWD) {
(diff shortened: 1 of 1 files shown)


MasterDuke17:
The benchmarks I ran showed no performance loss:

```
my $i; for ^100_000       { $i = IO::Path.new($_);           Nil }; say $i.perl; say now - INIT now
my $i; for ^100_000       { $i = IO::Path.new($_, :CWD($_)); Nil }; say $i.perl; say now - INIT now
my $i; for "aaaa".."tttt" { $i = IO::Path.new($_);           Nil }; say $i.perl; say now - INIT now
my $i; for "aaaa".."tttt" { $i = IO::Path.new($_, :CWD($_)); Nil }; say $i.perl; say now - INIT now
```

MasterDuke17:
Yeah, but I didn't see any difference when I ran the benchmarks I have above before and after my change. Maybe everything else going on (when actually doing `IO::Path.new`) hides the difference. Could someone else try running them before and after to confirm?

zoffixznet:
:+1: Thanks.

PR #1779, opened 2018-04-27
Implement NativeCall CArray.allocate method

### LINKED ISSUE #1778: Implementing a CArray.allocate method
Currently it's possible to allocate some memory for a `CArray` like this:

```perl6
my CArray[uint8] $ptr .= new;
$ptr[4096 - 1] = 0;
```

Or:

```perl6
my CArray[uint8] $ptr .= new: 0 xx 4096;
```
But this isn't very clear. You could use a `Buf`/`Blob` to allocate memory like this, but I feel like `CArray` should be able to as well to make it easier to pass to other NativeCall functions. Would this be a good idea to add?

Kaiepi:
Fixes #1778

### DIFF
--- a/lib/NativeCall/Types.pm6
+++ b/lib/NativeCall/Types.pm6
@@ -90,7 +90,7 @@ our class Pointer                               is repr('CPointer') {
 
 # CArray class, used to represent C arrays.
 our class CArray is repr('CArray') is array_type(Pointer) {
-    method AT-POS(CArray:D: $pos) { die "CArray cannot be used without a type" }
+    method AT-POS(::?CLASS:D: $pos) { die "CArray cannot be used without a type" }
 
     my role IntTypedCArray[::TValue] does Positional[TValue] is array_type(TValue) {
         multi method AT-POS(::?CLASS:D \arr: $pos) is raw {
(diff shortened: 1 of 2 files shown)


lizmat:
Cool.   But are you aware of `nqp::setelems()`?  Or is there a reason you cannot use that here?

Kaiepi:
I wasn't using it because it doesn't create new `Pointer` or `CStruct` instances, but I can use it for `IntTypedArray`  and `NumTypedArray`

Kaiepi:
Updated

lizmat:
my $arr := nqp::create(type);
    nqp::setelems($arr, $elems);
    $arr;

can be written shorter as:

    nqp::setelems(nqp::create(type),$elems)

`nqp::setelems` returns whatever it was setting the elements of.  That should make those subs shorter and better inlineable.

Kaiepi:
`nqp::setelems` didn't end up working when I tested it again, so I switched back to using `nqp::bindpos`/`nqp::bindpos_i`/`nqp::bindpos_n` instead:

```
This representation (CArray) does not support positional access (for type NativeCall::Types::CArray[int32])
  in method allocate at /home/morfent/Documents/perl6/rakudo/lib/NativeCall/Types.pm6 (NativeCall::Types) line 122
  in block <unit> at t/04-nativecall/05-arrays.t line 173
```

lizmat:
Too bad.  Not sure why, over my head I’m afraid.  :-)

lizmat:
Thanks!

PR #1784, opened 2018-04-29
X::NYI uses optional $.feature in method message

ufobat:
Methods `message` checks wether $.feature is set and returns slightly different error texts. 
But there are still other exceptions that use variables that might not have been set.  I dont know wether some of those variables are ment to have a `is required` on them or not.


zoffixznet:
👍 Thanks!

PR #1791, opened 2018-05-01
GH#1758: Only suggest a submethod if it's local

vendethiel:
This avoids us suggesting a submethod from a parent class, which would
always result into an error.

### DIFF
--- a/src/core/Exception.pm6
+++ b/src/core/Exception.pm6
@@ -171,7 +171,11 @@ my class X::Method::NotFound is Exception {
         }
 
         if nqp::can($!invocant.HOW, 'methods') {
-            for $!invocant.^methods(:all)>>.name -> $method_name {
+            my @invocant_methods = $!invocant.^methods(:local)>>.name;
+            for $!invocant.^methods(:all) -> $method_candidate {
+                my $method_name = $method_candidate.name;
+                # GH#1758 do not suggest a submethod from a parent
+                next if $method_candidate.^name eq 'Submethod' && !@invocant_methods.first($method_name, :k).defined;
                 my $dist = StrDistance.new(:before($.method), :after($method_name));
                 if $dist <= $max_length {
                     %suggestions{$method_name} = $dist;
--- a/t/05-messages/01-errors.t
+++ b/t/05-messages/01-errors.t
@@ -2,7 +2,7 @@ use lib <t/packages/>;
 use Test;
 use Test::Helpers;
 
-plan 48;
+plan 50;
 
 # RT #129763
 throws-like '1++', X::Multi::NoMatch,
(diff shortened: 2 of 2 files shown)


src/core/Exception.pm6
```diff
@@ -171,7 +171,11 @@ my class X::Method::NotFound is Exception {
         }
 
         if nqp::can($!invocant.HOW, 'methods') {
-            for $!invocant.^methods(:all)>>.name -> $method_name {
+            my @invocant_methods = $!invocant.^methods(:local)>>.name;
```
MasterDuke17: This could be `my @invocant_methods = $!invocant.^methods(:local).grep(*.^name eq 'Submethod')>>.name;`, right? That way the list to check inside the loop is smaller.

src/core/Exception.pm6
```diff
@@ -171,7 +171,11 @@ my class X::Method::NotFound is Exception {
...
+            for $!invocant.^methods(:all) -> $method_candidate {
+                my $method_name = $method_candidate.name;
+                # GH#1758 do not suggest a submethod from a parent
+                next if $method_candidate.^name eq 'Submethod' && $method_name eq none(@invocant_methods);
```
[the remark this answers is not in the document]
vendethiel: That `|| 0` wants to be `|| -1`, right?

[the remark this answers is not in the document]
MasterDuke17: Whoops, really should be `// -1`.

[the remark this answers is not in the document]
lizmat: Don't we have `with` to handle these cases?

src/core/Exception.pm6
vendethiel: Whoops, I didn't realize there was a comment here.
Yeah, I can move it there, but it's a bit of a pessimization as well, because then we calculate the list of submethods even if we're never to suggest them.

MasterDuke17: No, I'm fine with it.

vendethiel:
See also #1873.

t/05-messages/01-errors.t
```diff
@@ -228,6 +228,19 @@ throws-like { Blob.splice }, X::Multi::NoMatch,
...
+# https://github.com/rakudo/rakudo/issues/1758
+{
+    throws-like q| class GH1758_1 { submethod x { }; }; class B is GH1758_1 {}; B.new._ |,
+        X::Method::NotFound,
+        :message{ !.contains: "Did you mean 'x'" },
+        'Ancestor submethods should not be typo-suggested';
+
+    throws-like q| class GH1758_2 { submethod x { };}; GH1758_2.new._ |,
+        X::Method::NotFound,
+        message => /"Did you mean 'x'"/,
```
MasterDuke17: For consistency with the previous test, shouldn't this be `:message{ .contains: "Did you mean 'x'" },`?

t/05-messages/01-errors.t
```diff
@@ -228,6 +228,19 @@ throws-like { Blob.splice }, X::Multi::NoMatch,
...
+        'Ancestor submethods should not be typo-suggested';
```
MasterDuke17: And I'd suggest changing this description to something like `'Submethods at the same inheritance level should be suggested'`

t/05-messages/01-errors.t
vendethiel: I use that style because the rest of the file did that for positive matching, but I can change it.

MasterDuke17: Not that big a deal. If it's consistent with something that's fine. All the tests (across NQP, Rakudo, and roast) aren't very consistent on the whole.

t/05-messages/01-errors.t
JJ: Can you please check this?

JJ:
Also, there are conflicts now. Can you please check?

PR #1792, opened 2018-05-02
Fix smartmach exception explosion.

### LINKED ISSUE #1772: Hash ~~ Hash violates "smartmatch never throws" paradigm
Looks like it does `eqv` under the hood and `eqv` cannot handle `Mu`s:

```
m: my %h = a => Mu; say $%h ~~ $%h
rakudo-moar 79ed89ba4: OUTPUT: «Type check failed in binding to parameter '<anon>'; expected Any but got Mu (Mu)␤  in block <unit> at <tmp> line 1␤␤»
```

Possibly the fix would involve just wrapping the potentially-explosive code into a `try` and returning `False` in case of an exception.

jsimonet:
Smartmatch should never explodes.

Fix #1772

### DIFF
--- a/src/core/Map.pm6
+++ b/src/core/Map.pm6
@@ -119,7 +119,7 @@ my class Map does Iterable does Associative { # declared in BOOTSTRAP
     }
 
     multi method ACCEPTS(Map:D: Map:D \m --> Bool) {
-    	self eqv m;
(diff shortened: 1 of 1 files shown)


zoffixznet:
:+1: Thanks!

PR #1796, opened 2018-05-04
Add a CACHEDIR.TAG to precomp dirs

flussence:
The `CACHEDIR.TAG` file is an informal standard for declaring any directory is an expendable cache by putting a file with a magic string in it, handy for software that can't or won't use fixed cache locations (e.g. `XDG_CACHE_DIR`).

It's not required by any means, but polite to have, as it lets generic system tools know this directory is safe to blow away or exclude from things like backups.

This patch plays it safe and only adds the file to new dirs; having it check and backfill existing ones would waste precious cycles during startup and I didn't feel brave enough to attempt it in any case.

### DIFF
--- a/src/core/CompUnit/PrecompilationStore/File.pm6
+++ b/src/core/CompUnit/PrecompilationStore/File.pm6
@@ -191,6 +191,11 @@ class CompUnit::PrecompilationStore::File does CompUnit::PrecompilationStore {
     {
         unless $!prefix.e {
             $!prefix.mkdir or return;
+            try $!prefix.child('CACHEDIR.TAG').spurt:
+                "Signature: 8a477f597d28d172789f06886806bc55\n" ~
+                "# This file is a cache directory tag created by rakudo.\n" ~
+                "# For information about cache directory tags, see:\n" ~
+                "#       http://www.brynosaurus.com/cachedir/\n";
         }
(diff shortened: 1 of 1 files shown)


jnthn:
Seems reasonable to me, both in terms of idea and implementation. @niner, any thoughts?

niner:
I very much like the idea. The implementation is not a 100 % there though. With this the existence of the CACHEDIR.TAG very much depends on how a precomp directory is created. It's created when a precomp dir is created for storing a precomp file, but apparently not when storing a repo-id file and probably not when the precomp dir is created as part of installation of a dist.

Also we should discuss the question if all precomp directories should get this tag. E.g. the files in vendor/precomp will in general be part of a system package. As such I think they should be backed up, as otherwise after a restore the installed package would no longer match its description due to missing files. Also there'd be no way to regenerate the precomp files in vendor/precomp.

JJ:
Unfortunately, code seems to have drifted away. Can you please try and fix the conflict before we proceed with this?

lizmat:
I will take care of this one in a few hours.

lizmat:
This code is unsalvageable, reworking the idea into a different commit.

lizmat:
Implemented with https://github.com/rakudo/rakudo/commit/37646b869b

PR #1802, opened 2018-05-08
Micro-optimize Rational.Str

MasterDuke17:
By pulling an addition out of a loop.

Rakudo builds of and passes `make m-test m-spectest`.

`perl6 -e 'my $s; my $r = FatRat.new(5,2777777777777777); for ^10_000 { $s = $r.Str }; say $s; say now - INIT now;'` does ~90k fewer Int additions and is slightly faster.


lizmat:
Thanks!

PR #1806, opened 2018-05-08
Add test for RT#123380 (REPL tests go in rakudo tree)

skids:


skids:
Bug seems to have been resolved at some point.

zoffixznet:
👍 Thanks!

PR #1807, opened 2018-05-09
Rewrote Rational.Str to speed up stringification, Rats somewhat, FatRats a lot

thundergnat:
<strike>Moved Rational.Str to Rat.Str. Add a fast FatRat.Str multi.</strike> Rewrote Rational.Str to speed up Rational stringification across the board. Speeds up Rats marginally. Speeds up FatRats, especially those with large denominators significantly. Takes from half, to 1/200th the time that it used to take.

Spectest clean.


src/core/Rat.pm6
```diff
@@ -35,6 +71,28 @@ my class FatRat is Cool does Rational[Int, Int] {
...
+    multi method Str(FatRat:D:) {
+        my $whole  = self.abs.floor;
+        my $fract  = self.abs - $whole;
+
+        my $result = nqp::if(
+            nqp::islt_I($!numerator, 0), '-', ''
+        ) ~ $whole;
+
+        if $fract {
+            if $!denominator.chars < 5 {
```
MasterDuke17: Can't this be just `$!denominator < 10000`? That would be faster than converting it to a string to count the chars.

thundergnat: Good point. I'll fix that.

src/core/Rational.pm6
```diff
@@ -94,35 +94,38 @@ my role Rational[::NuT = Int, ::DeT = ::("NuT")] does Real {
         my $whole  = self.abs.floor;
         my $fract  = self.abs - $whole;
 
+        my $reduce = False;
         # fight floating point noise issues RT#126016
-        if $fract.Num == 1e0 { ++$whole; $fract = 0 }
+        if $fract.Num == 1e0 {
+            if nqp::unbox_s(self.^name) eq 'Rat' {
```
MasterDuke17: I think this could/should be `if nqp::eqaddr(self.WHAT,Rat) {`. Then you're actually checking against the type itself, not the string representation of its name.

src/core/Rational.pm6
```diff
@@ -94,35 +94,38 @@ my role Rational[::NuT = Int, ::DeT = ::("NuT")] does Real {
...
+                $whole += 1;
+                $fract = 0
```
MasterDuke17: This is kind of a nitpick, but I think there's a convention that last statements in blocks only leave off the semi-colon if the value is being returned.

src/core/Rational.pm6
```diff
@@ -94,35 +94,38 @@ my role Rational[::NuT = Int, ::DeT = ::("NuT")] does Real {
...
+            }
+            $reduce = True
```
MasterDuke17: Ibid.

src/core/Rational.pm6
```diff
@@ -94,35 +94,38 @@ my role Rational[::NuT = Int, ::DeT = ::("NuT")] does Real {
...
-                $fract -= $f;
-                $fract-result ~= $f < 10 ?? "0$f" !!
-                                 (!$fract and $f %% 10) ?? ($f / 10).floor !! $f;
+            my $c;
+            if $!denominator < 100000 {
+                $c = 6;
+                $fract *= 1000000;
+            } else {
+                $c = $!denominator.chars + 1;
+                $fract *= nqp::pow_I(10, nqp::decont($c), Num, Int);
```
MasterDuke17: The use of nqp ops seems a little inconsistent. Why use `nqp::pow_I` instead of just `**`? Or conversely, why use e.g., `$!denominator.chars` instead of `nqp::chars($!denominator.Str)` a line above?

src/core/Rational.pm6
```diff
@@ -94,35 +94,38 @@ my role Rational[::NuT = Int, ::DeT = ::("NuT")] does Real {
...
-                given $fract.floor {
-                    $fract-result ~= $_;
-                    $fract        -= $_;
-                }
+            my $f  = $fract.round;
+            my $fc = $f.chars;
+            # maintain precision (trailing zeros) of large denominator FatRats.
+            # There are situations where the extra zeros are significant, they
+            # can be easily removed in userspace but not easily restored
+            if ($c < 8) || (nqp::unbox_s(self.^name) eq 'Rat') || $reduce {
```
MasterDuke17: Same thing about `nqp::eqaddr(self.WHAT,Rat)` as above.

src/core/Rational.pm6
thundergnat: I am not arguing, but I did just copy that syntax from line 14 in the same file. I can certanly change that if it is a better way to go.

src/core/Rational.pm6
thundergnat: Fair enough.

src/core/Rational.pm6
thundergnat: I did some testing both ways and the nqp::pow_I seemed to be marginally but measurably faster. I could easily change that back though. I am not really familiar with the different nqp ops and which give enough of a benefit to outweigh the added complexity. 

Thanks for your review.

src/core/Rational.pm6
MasterDuke17: Yeah, but that line is intentionally creating a string. Line 12 does `nqp::eqaddr(self.WHAT,Rational),` because there it's checking for a type. If you grep through the Rakudo sources you'll see a whole lot more uses of `nqp::eqaddr(<something>.WHAT, <some type>)` than `<something>.^name eq <some type name>`. What you have would work, it's just slower.

src/core/Rational.pm6
MasterDuke17: What you have is definitely not wrong, so now we're getting into personal preference. But since you're changing the algorithm used (and the original code is almost completely pure Perl 6), maybe a first commit that converts to the new algorithm, and then a second commit that changes some of the new code's slow parts to nqp ops?

tbrowder:
A nit: the cuddled else (1) is not consistent with the rest of the code and (2) ignores the style guide.

thundergnat:
Obsolete. Rewrote following critiques and suggestions

PR #1812, opened 2018-05-10
Improve and consolidate distribution lookup

ugexe:
Adds method candidates to `CompUnit::Repository::Installation` ( CURI ) and `CompUnit::Repository::FileSystem` ( CURFS ). All distribution searching/resolving goes through this method. This gives developers insight into how a CURI/CURFS/CUR??? resolves a module request, which was not possible with `.resolve` -- it would chain to the next CUR and only returned a single result. The centralization of the recommendation-manager-like functionality should aid in implementing additional NYI s22 features.

https://github.com/perl6/roast/commit/cab3dd0571e0ceadaa5b60d2888e668b483b923c
https://design.perl6.org/S22.html#candidates

#### CURI and CURFS changes

`!matching-dist` is now backed by `.candidates`, but also caches a single 'best match' so it acts similar to before.

`.candidates` ( and by extension !matching-dist ) now returns a Distribution object and not ($dist-id, $dist-data). This is possible because of a role that allows our lazy distribution to avoid IO calls to obtain meta data. Previously this was done by passing around a hash internally, but now we dogfood our Distribution interface throughout CURI.

Removes `.script`, and reverts to using just `.files` - `.script` was just a fast path of `.files` under certain conditions and lacked the ability to filter by name. Now `.files` can provide the faster path for other files automatically. Additionally this fixes an issue where `.script` would invoke older version bin/ scripts from ~/.perl6

#### CURFS changes

Creates full Distribution for both -I. and -Ilib, where the later will use the directory recursion it was already doing ( to determine id ) to automatically populate meta data for provides and files ( bin/, resources/ ).

Gives bin scripts a way to more robustly know their own version ( although nyi `$*DISTRIBUTION` would be the more appropriate way, but would still be helped by this work ):
```
# bin/zef
use Zef::Client;

sub MAIN("version") {
    my $spec = CompUnit::DependencySpecification.new(:short-name<Zef::Client>);
    say $*REPO.need($spec).distribution.meta<ver version>.first
}
```
The above currently only works for `perl6 bin/zef` (See: https://github.com/ugexe/zef/issues/122#issuecomment-280796779), and would blow up on `perl6 -I. bin/zef` and `perl6 -Ilib bin/zef`. This PR works for all 3; -I. ( which we assume is pointing at a directory with a META file ) will pull its version from the meta file, and -Ilib (a directory without a META file) is version 0.

Note this does look at `../` to build the meta data for -Ilib, but this is already the case for CURFS.resources.


ugexe:
I’m going to merge shortly after release if no one has mentioned any reason not to.

zoffixznet:
*ping* is this mergeable or what is needed? It's blocking [auth tests](https://github.com/perl6/roast/blob/d2ac49334b049f0210680e461b99c84e4462ff1c/S11-repository/cur-candidates.t) that I believe are needed [to spec the `auth` key](https://github.com/perl6/roast/issues/450#issuecomment-404932032)

ugexe:
Progress is blocked by a consensus ( and possible associated code changes ) for https://github.com/rakudo/rakudo/issues/1853 since it adds the same auth/ver/api lookup/verification to CURFS / -I. -Ilib

Additionally the performance of the CURFS changes probably need to be improved, as it slowed down the roast a fair bit ( although this appeared to be because of how poorly the roast organizes the various it uses/tests, resulting in having to iterate and recurse over significantly more files ).

ugexe:
I was going to wait until after the release was cut, but I only have the next 5 days to keep a close watch on this which is already less than I wanted. So hopefully the release can still be handled without blocking master. @AlexDaniel

niner:
Yeah!

PR #1822, opened 2018-05-12
Rewrote Rat / FatRat stringification to be much faster.

thundergnat:
Rats stringify on average twice as fast, FatRats, depending on the
size of the denominator, range from about the same to 200+ times as
fast. Also fixes a few bugs: FatRat strings very close to 1 no
longer round off to 1. Rats and FatRats no longer ever stringify
with trailing zeros in the fractional part. FatRats are now capable
of stringifying to sixteen minimum significant digits or more if
necessary to accurately express the value, but holding off until v6.d
because it isn't spec. There are 8 tests in roast that expect only 6 digits
of precision from FatRats. They are all six years old or older.

E.G. Right now, Fatrat.new(2,3) stringifies to .666667 (same as Rat.new(2,3) ). With 16 minimum significant digits it would stringify to .6666666666666667. If you are trading off the lower speed of FatRat calculations for higher precision, it seems to me you should get more precise results back.

### DIFF
--- a/src/core/Rational.pm6
+++ b/src/core/Rational.pm6
@@ -95,34 +95,52 @@ my role Rational[::NuT = Int, ::DeT = ::("NuT")] does Real {
         my $fract  = self.abs - $whole;
 
         # fight floating point noise issues RT#126016
-        if $fract.Num == 1e0 { ++$whole; $fract = 0 }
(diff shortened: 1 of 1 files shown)


thundergnat:
See https://gist.github.com/thundergnat/075e98feab27434147b3aa455361282c for a demo script using augmented methods to compare before and after performance.

Issue #1821, opened 2018-05-12
Can't get Pod table formatted right


zoffixznet:
I have this table in my code:

```perl6
=begin table

    Type    | Comments
    ========+=================================================
    Complex |
    Num     |
    FatRat  |
    MidRat  | Special infectiousness. See prose that follows. 
    Rat     |
    Int     |

=end table
```

Which errors out with this:
```
$ perl6 -c doc/Language/numerics.pod6                                                                                                                                                                   
===FATAL: Table 2 has a mixture of visible and invisible column-separator types.
    Type    | Comments
    ========+=================================================
    Complex |
    Num     |
    FatRat  |
    MidRat  | Special infectiousness. See prose that follows. 
    Rat     |
    Int     |

===end FATAL table 2 input
===SORRY!===
```

I've no idea what it's talking about since as far as I can tell I've only used visible separators for all the things I wanted to use separators for.

I changed the code to this:

```perl6
=begin table

    Type    | Comments                                        |
    ========+=================================================+
    Complex |                                                 |
    Num     |                                                 |
    FatRat  |                                                 |
    MidRat  | Special infectiousness. See prose that follows. |
    Rat     |                                                 |
    Int     |                                                 |

=end table
```

And now the fatal error is gone, but I get a warning instead:

```
$ perl6 -c doc/Language/numerics.pod6 
===WARNING: One or more tables evidence bad practice.
==          Set environment variable 'RAKUDO_POD_TABLE_DEBUG' for more details.
Syntax OK

$ RAKUDO_POD_TABLE_DEBUG=1 perl6 -c doc/Language/numerics.pod6
[...]
===WARNING: Table 2 has unneeded border vis col separators.
    Type    | Comments                                        |
    ========+=================================================+
    Complex |                                                 |
    Num     |                                                 |
    FatRat  |                                                 |
    MidRat  | Special infectiousness. See prose that follows. |
    Rat     |                                                 |
    Int     |                                                 |

===end WARNING table 2 input rows
[...]
```

First of all, why is drawing the full table a "bad practice"? What's bad about it? And second, what's the right way to draw such a table?

zoffixznet:
> two adjacent embedded spaces in the filled second column?

You mean after `See prose that follows.`? No, there's just one space before the border and no spaces in the variant without the border.

tbrowder:
On Sat, May 12, 2018 at 07:36 Zoffix Znet <notifications@github.com> wrote:

[...]

PR #1826, opened 2018-05-14
Fix for GH #1821: unexpected table failure

tbrowder:
The failure was unexpected because the failure message
reported mixed visual and non-visual column separators
but such were not seen. The fix was to properly
detect trailing visual column separators.

Tests have been added for the fix.

Additional changes:

+ turn off unsolicited warnings for tables violating
  best practices (but code just commented out for now)
+ typo fix


lizmat:
Thanks!

PR #1840, opened 2018-05-18
Unify Rat and FatRat stringification. Fixes some TODO tests.

thundergnat:
Remove some branching logic and use same parameters for all
Rationals. Should be same or microscopically faster for small
denominator Rats/FatRats and probably a few percent slower for
some large denominator Rats. Rats no longer truncate fractional
portions that can be exactly expressed.


src/core/Rational.pm6
```diff
@@ -106,39 +106,17 @@ my role Rational[::NuT = Int, ::DeT = ::("NuT")] does Real {
...
-                #}
-                if $!denominator < 100000 {
-                    $precision = 6;
-                    $fract *= 1000000;
-                }
-                else {
-                    $precision = nqp::chars($!denominator.Str) + nqp::chars($whole.Str) + 1;
-                    $fract *= nqp::pow_I(10, nqp::decont($precision), Num, Int);
-                }
+                $precision = nqp::chars(~$!denominator) + nqp::chars(~$whole) +< 1;
```
zoffixznet: What's the reasoning for multiplication of the whole part by 2? Not following how the length of the whole part fits in into the equation of calculating precision even in the original code.

zoffixznet: And that means `(1000*2⁶²)/(2⁶²+1)` has 19 chars after the decimal, yet `(10²⁰⁰*2⁶²)/(2⁶²+1)` has 418 chars after the decimal (and is now 5.5x slower to generate). I'm not following at all the logic of this precision difference. The reverse would be more understandable, but not having extra precision for numbers large enough for it to matter **less**.

thundergnat: The more I think about this pull request, the less I think it is a good idea. The obvious deficiencies that Zoffix++ pointed out are a huge red flag. I've spent some time trying to come up with a better heuristic for determining a "correct" number of decimal places to return but they have all been wildly inaccurate or despairingly slow. 

At this point I think it  might be better to drop this for CORE. Extremely precise Rat/FatRat strings may be a niche better filed by a module.

PR #1842, opened 2018-05-18
Revert "JSONException -> Rakudo::Internals::JSONException"

ugexe:
Reverts rakudo/rakudo#1841

These can be changed in the ecosystem but that still leaves an already installed e.g. zef with a compile error if rakudo is upgraded.

### DIFF
--- a/src/core/CompUnit/Repository/FileSystem.pm6
+++ b/src/core/CompUnit/Repository/FileSystem.pm6
@@ -27,7 +27,7 @@ class CompUnit::Repository::FileSystem does CompUnit::Repository::Locally does C
                 try {
                     %!meta = Rakudo::Internals::JSON.from-json: $meta.slurp;
                     CATCH {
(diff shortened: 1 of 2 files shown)


AlexDaniel:
OK. Does it mean that we change zef first to work with JSONException in both namespaces, and a few releases later can safely switch?

ugexe:
The offending code has been modified in zef ( and Net::HTTP ) to not throw a compile error.

jnthn:
If this is something that's being used in the ecosystem, whatever it's renamed to - if it is - should not be something starting with `Rakudo::Internals::`.

ugexe:
@jnthn its only being used by myself, and its only specifically for very old code paths in rakudo from before the JSON classes were moved. The problem was I did not change `JSONException.new` to `::("JSONException").new`, so its a compile time error despite the fact it'll never be run.

PR #1844, opened 2018-05-19
Rewrite @a[1,3] to @a.AT-POS(1),@a.AT-POS(3) for literal indices

MasterDuke17:
Before, `my @a = ^10; my @d; @d = @a[3, 5] for ^100_000; say @d; say now - INIT now` took 2.8s, after it takes 0.5s.


MasterDuke17:
I wanted to get some other eyes on this. It works, but I feel it could be better. And any ways to make it more general would be great.

MasterDuke17:
@lizmat ping. Any further thoughts?

PR #1845, opened 2018-05-20
Issue 1843: Limit slice end index in splice implementation

jstuder-gh:
Splice allows the user to enter a number of elems greater than the elems
in the list (the extra is just ignored). However, nqp::slice throws an
exception if too large an index is submitted. Check that the end index
is not greater than the total elems.


zoffixznet:
👍 Thanks!

Issue #1203, opened 2017-10-24
Would be nice to have an easy way to work with SetHashes and its kin


zoffixznet:
Based on discussion [on RT#132352](https://rt.perl.org/Ticket/Display.html?id=132352), it would be nice if you could have a `%-` sigiled var to worth with SetHash and its kin and so you could use meta-assign form of set ops, like `∖=`

```
[...] make the following variation of it work:

    my %days is SetHash = Date.today … Date.new: '2014-04-02';
    %days ∖= %days.grep: *.key.day-of-week > 5;
    %days{Date.today}:delete;

...and then promote %-sigiled SetHash variables as the recommended way to store SetHash'es.

It should be possible to make this last example work by implementing `method STORE` for type SetHash, right?

(That it currently doesn't, may well be an oversight rather than a design choice.)
```

jnthn:
Yes, I think `STORE` should be implemented on `SetHash`, `MixHash`, and `BagHash`; and it should work very much like `STORE` does in `Hash`, except with the setty/baggy/mixy semantics. I think the only reason it's not been done yet is that nobody got around to it; there's no design reason I can think of for this to be missing.

lizmat:
So should .STORE use .new or .new-from-pairs semantics?

jnthn:
Neither, it seems. If we look at what assignment to an ordinary hash does:

```
> my %h = a => 1, 'b', 2
{a => 1, b => 2}
```

Then we see it knows what to do with both pairs and lone elements. Ditto for object hash:

```
> my %h{Any} = a => 1, 'b', 2
{a => 1, b => 2}
```

So I'd expect:

```
my %h is BagHash = a => 2, 'b', 2;
```

To result in a `BagHash` the same as:

```
(my %h = a => 1, 'b', 2).BagHash
```

Currently does.

lizmat:
That’s the new-from-pairs semantics  :-)  I’ll try to get that in after the release.

jnthn:
Ah, cool, the docs left me thinking it really only worked on pairs. :-)

AlexDaniel:
So what's the status of this now?
FWIW: [RT#132352](https://rt.perl.org/Ticket/Display.html?id=132352) [RT#132353](https://rt.perl.org/Ticket/Display.html?id=132353)

lizmat:
It appears to me these can both be closed as there appear to be tests for it that pass.

zoffixznet:
[...]

PR #1847, opened 2018-05-20
Add Set candidate for SetHash's STORE

MasterDuke17:
So set metaops like `∖=` will DWIM. A possible final fix for GH #1203. `perl6 -e 'my %days := SetHash.new: Date.today … Date.new: "2018-06-02"; say %days.elems; %days ∖= %days.grep: *.key.day-of-week > 5; say %days.elems'` now prints out 14 and 10, instead of 14 and 1.

Rakudo builds ok and passes `make m-test m-spectest` (will add some new test to roast if this is merged).

### DIFF
--- a/src/core/SetHash.pm6
+++ b/src/core/SetHash.pm6
@@ -176,9 +176,23 @@ my class SetHash does Setty {
     }
 
 #--- interface methods
-    method STORE(*@pairs --> SetHash:D) {
+    proto method STORE(|) {*}
+
+    multi method STORE(*@pairs --> SetHash:D) {
         nqp::if(
           (my $iterator := @pairs.iterator).is-lazy,
+          Failure.new(X::Cannot::Lazy.new(:action<initialize>,:what(self.^name))),
+          self.SET-SELF(
+            Rakudo::QuantHash.ADD-PAIRS-TO-SET(
+              nqp::create(Rakudo::Internals::IterationSet), $iterator
+            )
+          )
+        )
+    }
+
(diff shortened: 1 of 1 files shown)


lizmat:
FWIW, I'm -1 on this fix atm on the grounds that it doesn't feel right: will need some time to look into the problem deeper, which will be tomorrow.

MasterDuke17:
To add some additional data, the docs say that the set operators on SetHashes return immutable sets (https://docs.perl6.org/type/SetHash#Operators), which does seem to be specced in roast (e.g., https://github.com/perl6/roast/blob/master/S03-operators/set_symmetric_difference.t#L43).

MasterDuke17:
@lizmat ping. Any further thoughts?

lizmat:
Looking at this again now  :-)

MasterDuke17:
ISTR @jnthn had some comments on IRC, I'll see if I can find them and add them here.

MasterDuke17:
I haven't found the chat yet, but I think it was something about how it should be done as a new candidate for METAOP_ASSIGN (https://github.com/rakudo/rakudo/blob/master/src/core/metaops.pm6#L2-L4), but since that just takes an op and returns a sub, it would be a bit tricky.

MasterDuke17:
Ah, here it was, http://colabti.org/irclogger/irclogger_log/perl6-dev?date=2018-06-10#l180

lizmat:
Sorry, rejecting this PR on the same grounds as #3013

PR #1856, opened 2018-05-22
Execute code in elided duplicate argument pair values

skids:


skids:
Behavior before this patch:

```
$ perl6 -e 'sub f (|c) { c.say }; my $c = 2; f(:a($c--), :a($c*=5));'
\(:a(5))
```

Behavior with this patch applied:

```
$ perl6 -e 'sub f (|c) { c.say }; my $c = 2; f(:a($c--), :a($c*=5));'
\(:a(10))
```

...there may quite well be a more efficient way to express this in QAST
than as implemented in the patch.

Passes all spectests.  There's no bug filed for this behavior AFAICT.

lizmat:
I'm not sure what is being fixed here?  Isn't the behaviour of multiple identically named variables undefined?

Also, I already get the behaviour you describe now:

    $ 6 'sub f (|c) { dd c }; my $c = 2; f(:a($c--), :a($c*=5))'
    \(:a(10))

zoffixznet:
> Behavior with this patch applied:

If I'd expect any behaviour from it, I'd expect `a` to be 10. Going from left to right, it's being decreased to `1` and then multiplied by `5`. Why is `10` the right answer?

As lizmat mentioned this might be undefined behaviour. I know it is so in C, for example.

vendethiel:
In C, it is also UB to do `a[++i] = i` (also in C++ until C++17). I think it would be very surprising if this broke.
Just for my personal curiosity, why is `f(:1a, :2a)` allowed in the first place?

lizmat:
> On 22 May 2018, at 17:31, skids <notifications@github.com> wrote:
> 
> Also there is this never-implemented behavior from S06 to consider, for whatever it is worth:
> 
> Perl 6 allows multiple same-named arguments, and records the relative
> order of arguments with the same name.  When there are more than one
> argument, the C<@> sigil in the parameter list causes the arguments
> to be concatenated:

I found the phrase “concatenated” confusing here, so I’ve changed this to “appended”.


> 
>     sub fun (Int :@x) { ... }
>     fun( x => 1, x => 2 );              # @x := (1, 2)
>     fun( x => (1, 2), x => (3, 4) );    # @x := (1, 2, 3, 4)

And a little later it says:

Other sigils bind only to the I<last> argument with that name:

    sub fun (Int :$x) { ... }
    fun( x => 1, x => 2 );              # $x := 2
    fun( x => (1, 2), x => (3, 4) );    # $x := (3, 4)

So I would argue that 10 is the correct answer of the case:

    $ 6 'sub f (|c) { c.say }; my $c = 2; f(:a($c--), :a($c*5)); dd $c'
    \(:a(10))
    Int $c = 2

In other words, only the last specified parameter is taken into account, the code in the first is *not* executed.

Feels to me this is according to spec, and on that ground this PR should be rejected.


Or am I missing something?


Liz

skids:
Good call on "concatenated"

I disagree with that interpretation of the spec.  The spec says that sigils *bind* only to the last argument.  It doesn't say that sigils cause arguments to cease to exist.  If we were to implement the @a behavior, that would mean that the sigil on the a parameter would affect whether the $c got incremented or not with the same exact arglist... sub f(:@a) would increment it, and sub f(:$a) would not.

Furthermore, there is no compelling need to have inconsistent behavior between calling and hash assignment:

$ perl6 -e 'my $c = 2; {:a($c--), :a($c*=5)}.say'
5

zoffixznet:
> The spec says that sigils bind only to the last argument. It doesn't say that sigils cause arguments to cease to exist.

:+1: Not executing the code at all would be pretty weird, especially when the expression can be rather complex `-> :$a {}( :a{ multi-line-code ...; }, :42a )`

Also, using the last argument in the list with the same name makes sense and is also how it work in Perl 5. And how it'd have to remain working in variants where the name of the argument comes from a slipped hash or a slipped Pair:  `my $key = "a"; -> :$a { dd $a }( |("$key" => 42), :100a )`

masak:
My two eurocents: I like the fix, and I think it's undoubtedly better than the previous behavior.

I've never been 100% comfortable with the spec allowing duplicate _literal_ named arguments. I totally see the need for later named arguments to be able to override earlier ones when we're slurping entire hashes into the arguments, but I see little benefit in allowing actual duplicate named arguments in a call.

I know it's not necessarily "the way of the camel", but to me it would make sense to consider duplicate named (literal) arguments in a call to be a compile-time error. I might be swayed by learning about some use case that would prevent, but none come to mind.

jnthn:
I think the current behavior is a bit too aggressive in stripping away the side-effects of a dropped argument; we're generally not *that* lazy in Perl 6. The patch could be more efficient, however, by replacing:

```
                                QAST::Op.new(:op('callmethod'),
                                             :name('FLATTENABLE_LIST'),
                                             QAST::Op.new(:op('call'),
                                                          :name('&infix:<,>'))),
```

With just:

```
                                QAST::Op.new(:op('list'))
```

There's no need to make a Perl 6 level `List` object to just discard it immediately. With that change, I'm +1 to the patch.

On other issues raised here:

* I don't see the never-implemented bit of speculation in S05 ever being implemented. I don't think it fits well with the rest of the design around signatures and argument passing.
* For 6.d, I think I'd be OK with duplicate literal same-named arguments being a worry (compile time warning), and code-generators that somehow end up with that can shove a `no worries` at the top.

skids:
I'll try just `QAST::Op.new(:op('list'))` and push a commit with that assuming it does the trick.

masak:
:+1:

PR #1859, opened 2018-05-23
fix nbsp-fix problem by adding vertical and other missing breaking an…

tbrowder:
…d non-breaking space chars for a complete unicode solution


tbrowder:
@samcv this should fix the broken pod tests.

samcv:
@tbrowder what originally broke the pod tests? Can you link me to the commit (if you know it).

PR #1861, opened 2018-05-23
Workaround Nil value for &DEPRECATED in phaser from trait (R#1219)

skids:
This does not explain why the symbol is Nil, but makes it work
  by carrying the symbol in from the trait scope in a variable.


lizmat:
Is this intended as a temporary fix, or a more permanent one?

lizmat:
Fixed another way with 472f6e4844 .  Thanks for the nudge!

Issue #1862, opened 2018-05-24
Some objects don't .item'ize even though they probably should


zoffixznet:
Some objects that don't do `Iterable` are nevertheless things that have a bunch of items in them, such as `Capture` and `QuantHash`.

When trying to use `$()` contexualizer we call the `.item`. For `Iterable`s, this method stuff the object into a `Scalar` container, however for `Capture` and `QuantHash` we end up using `Mu.item`, which is a no-op.

As a result, trying to iterate over a `Capture` or a `QuantHash` while explicitly asking it to be a single item results in iterating over the things it contains instead:

```
m: for \(1,2,3,:a,:b).item { .say }
rakudo-moar 154d98551: OUTPUT: «1␤2␤3␤»

m: dd $_ for $(set <a b c>)
rakudo-moar 154d98551: OUTPUT: «:b␤:a␤:c␤»
```

The same would be the case for any userland objects that implement their own iterators. They'd need to implement their own `.item` method to achieve correct behaviour here.

Should `Mu.item` really be a no-op?

zoffixznet:
Looks like it's been this way [since the beginning of times](https://github.com/rakudo/rakudo/commit/7845bc146a33d10b7abc28a0315e6baa606a27bf) and I did not spot any explicit reasons mentioned for current no-op behaviour in all of the commits that touched Mu.item's code.

zoffixznet:
```
<jnthn> | If  something decides that it's going to implement .iterator without making  itself Iterable, then it's on that type to also implement method item,  I'd say
```

PR #1863, opened 2018-05-24
Always containerize stuff when asked to with .item

zoffixznet:
Potential solution to [R#1862](https://github.com/rakudo/rakudo/issues/1862).

Passes stresstest, but I did not think much about possible implications of this change.

### DIFF
--- a/src/core/Iterable.pm6
+++ b/src/core/Iterable.pm6
@@ -13,10 +13,6 @@ my class Rakudo::Internals::HyperIteratorBatcher { ... }
 my role Iterable {
     method iterator() { ... }
 
-    method item() {
-        nqp::p6bindattrinvres(nqp::create(Scalar), Scalar, '$!value', self)
-    }
-
     method flat(Iterable:D:) {
         Seq.new(class :: does Iterator {
             has Iterator $!source;
--- a/src/core/Mu.pm6
+++ b/src/core/Mu.pm6
@@ -571,7 +571,9 @@ Perhaps it can be found at https://docs.perl6.org/type/$name"
     }
     multi method Stringy(Mu:D $:) { self.Str }
 
-    method item(Mu \item:) is raw { item }
+    method item() is raw {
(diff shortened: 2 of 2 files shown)


niner:
But that would itemize things that are already itemized, wouldn't it? So scalars will be packed into other scalars.

Wouldn't it be better to have Capture and QuantHash do the Iterable role if they are in fact iterable?

zoffixznet:
> But that would itemize things that are already itemized, wouldn't it?

No, `self` is always deconted (unless it's a `Proxy`).

> Wouldn't it be better to have Capture and QuantHash do the Iterable role if they are in fact iterable?

Yeah, maybe.

zoffixznet:
> >    Wouldn't it be better to have Capture and QuantHash do the Iterable role if they are in fact iterable?

> Yeah, maybe.

But at the same time, maybe not. For example, making them `Iterable` would cause them to be dismantled by slurpies: 

```
$ perl6 -e 'my $c := \(1, 2, 3, :foo, :bar); my $o := (<a b c>, $c); my $i := (<a b c>, ($c but Iterable)); -> *@a { dd @a }($o)'
["a", "b", "c", \(1, 2, 3, :bar, :foo)]

$ perl6 -e 'my $c := \(1, 2, 3, :foo, :bar); my $o := (<a b c>, $c); my $i := (<a b c>, ($c but Iterable)); -> *@a { dd @a }($i)'
["a", "b", "c", 1, 2, 3]
```

Or descended into by hypered method calls:

```
$ perl6 -e 'my $c := \(1, 2, 3, :foo, :bar); my $o := (<a b c>, $c); my $i := (<a b c>, ($c but Iterable)); dd $o».uc'
No such method 'uc' for invocant of type 'Capture'
  in block <unit> at -e line 1

$ perl6 -e 'my $c := \(1, 2, 3, :foo, :bar); my $o := (<a b c>, $c); my $i := (<a b c>, ($c but Iterable)); dd $i».uc'
($("A", "B", "C"), $("1", "2", "3"))
```

The slurpy one is a particularly large semantic change.

PR #1868, opened 2018-05-25
Throws X::Multi::NoMatch if a proto routine is defined without multis.

jsimonet:


zoffixznet:
What does the error that's now thrown looks like? Because non-`where` version throws `X::TypeCheck::Argument` not `X::Multi::NoMatch`.

Should the exception type really change just because a parameter in a proto got a `where` clause attached to it?

```
<Zoffix_> m: proto z($, $) {*}; try say z "x"; say $!.^name
<camelia> rakudo-moar 79d7c4bec: OUTPUT: «===SORRY!=== Error while compiling <tmp>␤Calling z(Str) will never work with signature of the proto ($, $)␤at <tmp>:1␤------> proto z($, $) {*}; try say ⏏z "x"; say $!.^name␤»
<Zoffix_> m: proto z($, $) {*}; try 'say z "x"'.EVAL; say $!.^name
<camelia> rakudo-moar 79d7c4bec: OUTPUT: «X::TypeCheck::Argument+{X::Comp}␤»
```

jsimonet:
As I remember, the error is generated because the proto is a dispatcher (because of the presence of the whatever) and does not have multi candidates. In this case, the proto cannot be called (it can if it does not contains a whatever).
If we are using a where constraint without the proto being a dispatcher, we got the correct message: 
```
~/src/rakudo (d0a8a7af0 ✔) ᐅ perl6 -e 'proto z($ where rand) {}; say z "x"' 
Constraint type check failed in binding to parameter '<anon>'; expected anonymous constraint to be met but got Str ("x")
  in sub z at -e line 1
  in block <unit> at -e line 1
```

The `X::Multi::NoMatch` error tries to express that it's not possible to call a proto dispatcher with no candidates.

The error now look like this:
```
~/src/rakudo (d0a8a7af0 ✔) ᐅ ./install/bin/perl6 -e 'proto z($ where rand) {*}; say z "x"' 
Cannot resolve caller z(...); Routine does not have any candidates. Is only the proto defined?
at -e:1
```

zoffixznet:
:+1: Thanks!

timo:
There is a test in roast that is meant to check if the error message for `proto sub foo(Str) {*}; foo(42);` makes it easy to see what parameter was responsible for the error. See this RT ticket for a brief discussion: https://rt.perl.org/Public/Bug/Display.html?id=76368

> <masak> rakudo: sub foo(Str) {}; foo 42
> <p6eval> rakudo 7017d2: OUTPUT«Nominal type check failed for parameter
> ''; expected Str but got Int in 'foo' at line 11

I vote we fix the test by adding a multi candidate, what do the others think?

zoffixznet:
Is that too specific a test? We make compile-time "will never work" call evaluation in the optimizer stage. An implementation that doesn't or can't perform such evaluation won't be able to throw the same error as rakudo.

Perhaps it should be placed into [APPENDICES](https://github.com/perl6/roast/tree/master/APPENDICES) instead?

jsimonet:
I don't know if I can vote, but I would vote to fix the test by adding a multi candidate as timo said, and also to add a second one without the `*` dispatcher.
```perl6
proto sub foo(Str) {}; foo 42;
```

PR #1869, opened 2018-05-25
Turn BIND-KEY into multis

MasterDuke17:
The CORE.setting compilation no longer gets confused.


niner:
What's the goal with this? The commit message doesn't tell, so it's hard to say if the commit reaches the goal.

lizmat:
Thanks!

PR #1873, opened 2018-05-26
Hide parent's submethods from ^methods() result.

jsimonet:
Only "public" methods should be returned.


lizmat:
This PR now adds a single blank line?  If that's the intent, we can close it now  :-)

JJ:
No, that's just the commit that solved the conflict. The previous one is
the one that makes the change

vrurg:
@JJ Actually, it is just a blank line. This is what github reports in 'files changed' tab. Don't know what happened to the first commit, but it is not applicable anymore due to significantly changed code. 

Though anyway, the patch is simple and implementing it would only take removal of lines 72-77 from MethodContainer.nqp.

JJ:
> @JJ Actually, it is just a blank line. This is what github reports in 'files changed' tab. Don't know what happened to the first commit, but it is not applicable anymore due to significantly changed code.
> 

Right. The code was so different, that I didn't realize it was actually the evolved version of the code that was originally removed.

> Though anyway, the patch is simple and implementing it would only take removal of lines 72-77 from MethodContainer.nqp.

Doing that now.

PR #1884, opened 2018-05-30
Correct param name with signature constraint (fixes rt#132209)

skids:


skids:
Tests may be pulled from https://github.com/skids/roast/tree/rt132209

lizmat:
I was looking at resolving the conflict, but I am unsure how to solve it.  @JJ maybe you have a better idea?

JJ:
Will try to have a look...

lizmat:
This appears to have been effectively implemented by e9a0b18c6c, so closing

PR #1901, opened 2018-06-04
Simplify Kernel signals method using Signal values

jstuder-gh:
The Kernel signals method previously had a lot of logic for determining
signal values on the host system and backends. Now that that logic is
encapsulated into the getsignals ops, this method can be greatly
simplified.

### DIFF
--- a/src/core/Kernel.pm6
+++ b/src/core/Kernel.pm6
@@ -110,48 +110,30 @@ class Kernel does Systemic {
     }
 
     has @!signals;  # Signal
-#?if jvm
-    method signals (Kernel:D:) {
-        @!signals //= [2, 9]
-    }
-#?endif
-
     has $!signals-setup-lock = Lock.new;
-#?if !jvm
-    has $!signals-setup = False;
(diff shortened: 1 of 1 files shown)


lizmat:
I assume this should still be merged, correct?

jstuder-gh:
It's been a long time since I've looked at this, but I believe it should still be okay to merge in. All three backends have getsignals working (although I do have an [old PR on NQP repo](https://github.com/perl6/nqp/pull/465) to bring parity between the signals available on WIN32 between the backends).

I know that at least one module in the ecosystem was calling this method (I don't remember what that module was called though), but I don't see how this would break anything.

lizmat:
ok, will merge after the next compiler release, just to be on the safe side

jstuder-gh:
I force-pushed a new version of the branch that is exactly the same as the old version except:

* Rebased on top of current Rakudo master
* Replaced hard-coded ```32``` with value based off ```Signal.enums.values.max```

JJ:
There's a conflict after the change of name. Can you please check?

PR #1916, opened 2018-06-12
Ease maintenance of Actions.nqp, Grammar.nqp

tbrowder:
The monolithic files are divided into separate files by class or
grammar. Files are then concatenated into one file, in the required
order, at build time.

Actions.nqp was shortened to the Perl 6 class and the remains divided
into:

+ QActions.nqp
+ RegexGrammar.nqp
+ P5RegexActions.nqp

Grammar.nqp was shortened to the Perl 6 grammar and the remains
divided into:

+ QGrammar.nqp
+ RegexActions.nqp
+ P5RegexGrammar.nqp

The new files were added to the appropriate places in files:

+ tools/build/Makefile-JVM.in
+ tools/build/Makefile-Moar.in


zoffixznet:
> Ease maintenance of Actions.nqp, Grammar.nqp

I'm not following how this commit eases the maintenance. 

Right off the bat, you're making `git blame` more annoying to use as nearly everything past the point of merge will be in a different file. So that makes maintenance harder.

Also, instead of having all the Grammar in one, easily searchable file, the tokens are now split into 3 files. Parsing bugs currently require opening just two files, this PR bumps to 6.

For example: `Perl6::RegexGrammar.rxstopper` uses `<stopper>` token. What is its definition?

**New Way:**  I search the current file, it's not there. So now I have to look at inheritance and figure out that it might be in STD. Ok, let's find STD.nqp file... Oh, it's not there, OK, let's try Grammar.nqp file. OK, there's the STD and there's the token.

**Current Way:** Just search the file for `token stopper` and 90% of the time you find the token you seek, with the other 10% it being called `rule` or `method`.

Example 2: Look at the dump of `perl6 --target=parse -e '/{"$_"}/'` where's the token `noun` defined?

**New Way:**  I search `RegexGrammar.nqp` it's not there. `Grammar.nqp` nope not there. Well, I got quotes in my code, maybe it's in `QGrammar.nqp`, nope no there either. So it must be from NQP.

**Current Way:** I search `Grammar.nqp`. Nope no there, so it must be from NQP.

And since the braid gets switched from Main to Regex to Quote to Main and then back, tracing that parse now requires you juggling a whole bunch of files.

-------

How exactly does messing with git blame and adding a lot of extra steps to parser debugging make maintenance easier?

tbrowder:
How do you feel about just separate Pod*nqp files, then?

zoffixznet:
> How do you feel about just separate Pod*nqp files, then?

I think that's fine. Would be nice to have it as an entirely separate module even, to possibly support safe pod rendering that avoids executing arbitrary code.

jnthn:
Agree with @zoffixznet; yes, it's a big file, but it's also easy enough to navigate by simply searching for the desired token.

I'm OK with the Pod stuff going out to a new file; unlike the Q and Regex slangs it seems that it does not call back into the MAIN language's tokens, so it can be considered in isolation more naturally than the others.

tbrowder:
Okay, closing this PR.

Issue #1918, opened 2018-06-13
assuming a sub that has a **@param slurpy gives a compile error


kjkuan:
```
[20:48] <kjk> p6: sub x(**@a, :$b) { 123 }; &x.assuming(:!b)
[20:48] <+camelia> rakudo-moar f2b503785: OUTPUT: «===SORRY!=== Error while compiling /home/camelia/EVAL_2␤Redeclaration of symbol '@a'␤at /home/camelia/EVAL_2:1␤------> $res = (my proto __PRIMED_ANON (*@a, *@a⏏, Any :b($b) --> Mu) { {*} });␤    expecting any of:␤        shape…»
```
I should be able to `&x.assuming(:!b)` with the sub above, but it gives a compile error instead.

## Environment
* Operating system: MacOS 10.13.4
* Compiler version: `This is Rakudo version 2018.05 built on MoarVM version 2018.05
implementing Perl 6.c.`

jstuder-gh:
This is happening for ```+@``` params as well

    sub x(+@a, :$b) { say "{@a.perl} and $b" }; &x.assuming(:!b);

If you add slurpy_lol and slurpy_onearg to Parameter's internal NOT_POSITIONAL constant then they will work (commit doing so [here](https://github.com/jstuder-gh/rakudo/commit/93a8ec66553b78405052b71fa793f062e073b109)).

    sub x(**@a, :$b) { say "{@a.perl} and $b" }

    my &xx = &x.assuming(:!b);
    &xx(1, 2, (3, 4));

    # OUTPUT: «[1, 2, (3, 4)] and False␤»

Since these were testing positively for both ```positional``` and ```slurpy``` they were being added a new param list twice ([here](https://github.com/rakudo/rakudo/blob/master/src/core/Code.pm6#L196) and [here](https://github.com/rakudo/rakudo/blob/master/src/core/Code.pm6#L216)) resulting in redeclaration when EVALed.

I don't know a lot about the design of the ```Parameter``` types. Is there a reason these should be considered ```positional``` and ```slurpy``` or is it perhaps an oversight that they were not added to NOT_POSITIONAL? The spectest passes with these changes so presumably these properties are safe to modify? Assuming it is safe to do I will submit the above commit as a PR.

jstuder-gh:
[...]

PR #1921, opened 2018-06-14
Issue 1918: Fix for .assuming sub with **@ and +@ params

jstuder-gh:
Add slurpy_lol (**@) and slurpy_onearg (+@) to param types that are not
considered positional.

This caused an issue when using '.assuming'. The fact that they were
considered both positional and slurpy caused them
to be added multiple times to a new param list and miscompiled on EVAL
(The below example would result in a redeclaration error).

    sub x(+@a, :$b) { say "{@a.perl} and $b" }; &x.assuming(:!b);

See [GH Issue 1918](https://github.com/rakudo/rakudo/issues/1918).

### DIFF
--- a/src/core/Parameter.pm6
+++ b/src/core/Parameter.pm6
@@ -36,6 +36,8 @@ my class Parameter { # declared in BOOTSTRAP
 
     my constant $SIG_ELEM_IS_NOT_POSITIONAL = $SIG_ELEM_SLURPY_POS
                                            +| $SIG_ELEM_SLURPY_NAMED
+                                           +| $SIG_ELEM_SLURPY_LOL
+                                           +| $SIG_ELEM_SLURPY_ONEARG
                                            +| $SIG_ELEM_IS_CAPTURE;
     my constant $SIG_ELEM_IS_SLURPY = $SIG_ELEM_SLURPY_POS
                                    +| $SIG_ELEM_SLURPY_NAMED


jstuder-gh:
Is there a reason these should be considered ```.positional``` and ```.slurpy``` or is it perhaps an oversight that they were not added to NOT_POSITIONAL? The spectest passes with these changes so presumably these properties are safe to modify?

If these params do need to remain both ```.positional``` and ```.slurpy``` then this is not a suitable fix, but since you cannot have positionals after variadic params in signatures, I'm assuming that this change is correct.

zoffixznet:
> Is there a reason these should be considered .positional and .slurpy

Not familiar with this part of the codebase, but how come would these not be positional? There are slurpy named params `*%foo` and slurpy positional params `*@foo`, `+@foo`, `**@foo`

zoffixznet:
>  how come

Though in the code I see `$SIG_ELEM_SLURPY_POS` is already in that list, so I guess the meaning in that code is slightly different than what "positional" means in the language itself.

zoffixznet:
👍 Thanks!

jstuder-gh:
@zoffixznet, Thanks for the merge! 👍 

Yeah it threw me for a loop too.

I always thought of a "slurpy" as the final "positional" param that slurps up the remaining data, but in this case it appears that "positional" essentially means a parameter that binds to one object/primitive and preceeds the slurpy. Or at least from what I can gather.

They are definitely treated differently in the ```assuming``` method though. [Here](https://github.com/rakudo/rakudo/blob/master/src/core/Code.pm6#L196) and [here](https://github.com/rakudo/rakudo/blob/master/src/core/Code.pm6#L216) are two instances where it pushes the data multiple times to the new param list, first filtered from the ```.positional``` param then filtered from the ```.slurpy``` params.

Issue #1891, opened 2018-05-31
Signal(0) gives FreeBSD signals on Linux


AlexDaniel:
```
<AlexDaniel> m: say Signal(0)
<camelia> rakudo-moar c80ea2f28: OUTPUT: «SIGTHR␤»
<AlexDaniel> e: say Signal(0)
<evalable6> AlexDaniel, rakudo-moar c80ea2f28: OUTPUT: «SIGINFO␤»
```

Locally I also get SIGEMT. Seems to be somewhat random. Bisected to around this merge: https://github.com/rakudo/rakudo/pull/1601

Ping @jstuder-gh.

Related ticket: #1888.

AlexDaniel:
Ah, expected behavior is `(Signal)` type object. It used to do that previously and it also does this for other values (e.g. `-1`).

AlexDaniel:
FWIW please also test high values (like 32 and above).

jstuder-gh:
All the sigs that don't have a signum on the host system are being set to 0. My thought process behind this was that a symbol would be registered for every sig but the value would be 0 so that it wouldn't be dangerous to use in conjunction with the native C kill function if it was not available (passing a 0 signal to kill does not emit anything). And the ```signal``` sub (that registers signal handlers with the backend) filters out all the 0's it encounters.

We could make it so that no entry is included in the hash for signals that are not available on the platform, but that would create issues where user code using a particular signame could produce compile-time error depending on the platform it is run on.

Personally, I don't think it's a problem having ```Signal(0)``` return a value instead of the type object but I do think it's a problem that it is inconsistent and that if there are multiple enumerations with the same value that some are not included.

Regardless of what we decide to do with ```Signal```, what should we do if multiple enumerations for a given type have the same value? We could make it so that it returns a list of all the enums that match if multiple, singular enum if only one?

(In regards to more consistent ordering, I figure that we can modify the getsignals op to return a list instead of a hash and have ```.[idx] %% 2``` be key and ```!(.[idx] %% 2)``` be the value. Since feeding the ```Signal``` enum is the only thing using this op now it would be safe to change. This would make the results of using ```.pred``` and ```.succ``` on ```Signal``` more sane as well.)

AlexDaniel:
@jstuder-gh thanks! Merged your PRs.

Can't we just hardcode “nothing” for `0`?

jstuder-gh:
[...]

PR #1928, opened 2018-06-17
Issue 1891: Add role to Signal for custom zero logic

jstuder-gh:
This add custom logic for ```Signal(0)``` so that it will return the type object
like it did previously.

See [Issue 1891](https://github.com/rakudo/rakudo/issues/1891)

I set the target of the PR to the post-release branch, but if it makes more
sense to merge it into master before release, then I am okay with it being
changed.

### DIFF
--- a/src/core/signals.pm6
+++ b/src/core/signals.pm6
@@ -1,5 +1,5 @@
 my role Signally {
-    multi method CALL-ME(Int $signum) {
+    multi method CALL-ME(Int() $signum) {
         return self if $signum == 0;
         nextsame
     }
--- a/src/core/signals.pm6
+++ b/src/core/signals.pm6
@@ -1,4 +1,10 @@
-my enum Signal (
+my role Signally {
+    multi method CALL-ME(Int $signum) {
+        return self if $signum == 0;
+        nextsame
+    }
+}
+my enum Signal does Signally (
     |nqp::stmts(
         ( my $res  := nqp::list ),
         ( my $iter := nqp::iterator(nqp::getsignals) ),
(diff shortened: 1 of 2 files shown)


lizmat:
Not sure it is a good idea to make CALL-ME a multi-sub

src/core/signals.pm6
```diff
@@ -1,4 +1,10 @@
-my enum Signal (
+my role Signally {
+    multi method CALL-ME(Int $signum) {
```
AlexDaniel: But this won't cover other reasonable cases like `Signal(‘0’)`?

jstuder-gh: Thanks. I've pushed up a new commit that allows for scenarios like this.

AlexDaniel:
Let's try that.

@jstuder-gh There's an interesting idea [here](http://colabti.org/irclogger/irclogger_log/perl6-dev?date=2018-06-17#l135), perhaps we should consider reworking it later.

AlexDaniel:
I think there's a different issue (?) now:

> \<AlexDaniel\> c: 2018.05,HEAD say Signal(∞)
> \<committable6\> AlexDaniel, ¦2018.05: «(Signal)␤» ¦HEAD(5a6ff40): «Cannot convert Inf to Int: ␤  in block <unit> at /tmp/5jwSYHlfGp line 1␤␤ «exit code = 1»»

I don't know if it's important, but I still wonder if there's any easy way to keep previous behavior.

jstuder-gh:
@AlexDaniel, I pushed another commit to the same branch that adds an Inf candidate, which you can merge in if you'd like.

PR #1933, opened 2018-06-17
Issue 1891 Continued: Add special case for Inf in Signally CALL-ME

jstuder-gh:
Add an Inf candidate that takes care of that special case.

I suppose we may need to reconsider how Signal is implemented if
there are too many of these special cases. But in isolation, it this one
seems acceptable?

### DIFF
--- a/src/core/signals.pm6
+++ b/src/core/signals.pm6
@@ -1,4 +1,5 @@
 my role Signally {
+    multi method CALL-ME(Inf) { self }
     multi method CALL-ME(Int() $signum) {
         return self if $signum == 0;
         nextsame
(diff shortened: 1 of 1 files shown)


AlexDaniel:
What about
```perl6
return self unless $signum;
```
?

The problem in there is that `Inf.Int` returns a Failure, so it blows up when you try to compare it with 0. I guess that should cover all relevant cases.

AlexDaniel:
See http://colabti.org/irclogger/irclogger_log/perl6-dev?date=2018-06-17#l230 and [RT#132980](https://rt.perl.org/Ticket/Display.html?id=132980).

PR #1934, opened 2018-06-17
Prepend inst# to home when setting up repositories

hoelzro:
The keys to the `$repos` hash are in the form of `inst#/path/to/repo` (among
others), but $home was just the path before this change, causing all
`nqp::atkey($repos, $home)` calls to come back as null

### DIFF
--- a/src/core/CompUnit/RepositoryRegistry.pm6
+++ b/src/core/CompUnit/RepositoryRegistry.pm6
@@ -95,8 +95,7 @@ class CompUnit::RepositoryRegistry {
                    (nqp::existskey($ENV,'HOMEPATH')
                      ?? nqp::atkey($ENV,'HOMEPATH') !! '')
                  ) -> $home-path {
-                $home = "$home-path/.perl6";
-                my str $path = "inst#$home";
+                $home = "inst#$home-path/.perl6";
             }
         }
 
@@ -147,7 +146,7 @@ class CompUnit::RepositoryRegistry {
                 CompUnit::Repository::Installation.new(:prefix("$prefix/site"), :$next-repo)
             )) unless nqp::existskey($unique, $site);
             nqp::bindkey($custom-lib, 'home', $next-repo := self!register-repository(
-                "inst#$home/.perl6",
+                $home,
                 CompUnit::Repository::Installation.new(:prefix($home), :$next-repo)
             )) if $home and not nqp::existskey($unique, $home);
         }


AlexDaniel:
This looks fine to me.

Ping @ugexe just in case.

ugexe:
This PR is incomplete. See:
https://github.com/rakudo/rakudo/blob/5bd3366921ea4f0f0d7a4a14f144d027d105c737/src/core/CompUnit/RepositoryRegistry.pm#L157

ugexe:
https://github.com/rakudo/rakudo/commit/b6a7d7f64d08589750803de946e38c12cddcbf43 looks to remove the code that used to handle this.

hoelzro:
Thanks for that input @ugexe - I'll fix this up tomorrow!

hoelzro:
@AlexDaniel @ugexe I've made that fix - could I get a quick look before this is merged?

ugexe:
@hoelzro as far as I can tell a similar solution would be to revert https://github.com/rakudo/rakudo/commit/b6a7d7f64d08589750803de946e38c12cddcbf43 so there is still the question of what, if any, problem that fixed.

hoelzro:
@ugexe @AlexDaniel Ok, I tried doing that revert, managed to cause the `No such method 'prefix' for invocant of type 'Str'` message via just building Rakudo.  It looks like `$custom-lib` is supposed to have repository objects for its values - setting a value to `$path` triggers that error, which makes sense.  Looking at the code, I'm not familiar enough with that feature and its history to feel comfortable making the change, especially as @AlexDaniel is gearing up for a release; would my original change in this PR be good enough for now?  Or how else should we proceed?

ugexe:
+1 for merging — given you were able to recreate that issue.

Seems like $home should be left unset unless $home-path.chars > 0, but maybe out of scope ( “”. IO vs “inst#”.IO )

ugexe:
This needs to receive a path, not an e.g. `inst#foo/bar` ( repo spec ):
https://github.com/rakudo/rakudo/blob/91b211ac2f2a569f3112c695d60785deb4af95b3/src/core/CompUnit/RepositoryRegistry.pm6#L150

AlexDaniel:
Pretty sure that was fixed in https://github.com/rakudo/rakudo/commit/cf10780daa7f7d6f7cfc8e333c3a802a0d03e560.

Issue #1959, opened 2018-06-24
Empty set()/bag()/mix() do not return a unique object


zoffixznet:
This is a recent change in https://github.com/rakudo/rakudo/commit/8a4df16204eaddeeea5fb28364c8b5a88b3a9a91 and it results in program-wide action-at-a-distance if you modify the returned object by mixing a role. Any place that used a `set()` will now have a `Set` that returns `"foo"` for `.Str`:

```
<Zoffix__> m: my $a := set(); my $b := set(); $a does 'foo'; $b.print
<camelia> rakudo-moar e9351cbaa: OUTPUT: «foo»

<Zoffix__> m: my $a := set(1); my $b := set(1); $a does 'foo'; $b.print
<camelia> rakudo-moar e9351cbaa: OUTPUT: «1»
```

If the commit is reverted, there are likely more changes that will need to be removed as well, as for example `Set.perl` checks whether the object is the object from `set()`: https://github.com/rakudo/rakudo/issues/1958#issuecomment-399769123

lizmat:
The empty set/bag/mix are just that: immutable empty Set/Bag/Mix, and as such special (and we should only need one of each for the lifetime of a process).

Now, the behaviour with `does` is unfortunate, and a bit of DIHWIDT, but I think we can fix *that*

zoffixznet:
> The empty set/bag/mix are just that: immutable empty Set/Bag/Mix, and as such special

Out of curiosity, what's so special about them that we want them to always use the same object?

zoffixznet:
> Now, the behaviour with does is unfortunate, and a bit of DIHWIDT, but I think we can fix that

In fact, it's not just `does`. You can mutate with `.^mixin` too: 

```
set().^mixin: role { method Str { "meow\n" } }; print set()
```

The behaviour of `=:=` is also affected:

```
$ perl6 -e 'sub get-parts { set |("part" xx 3.rand) }; dd unique :with(&[=:=]), get-parts() xx 10'
(set(), ("part").Set, ("part").Set, ("part").Set, ("part").Set, ("part").Set, ("part").Set, ("part").Set, ("part").Set).Seq
```

zoffixznet:
> a bit of DIHWIDT

I also don't understand this judgment. What's DIHWIDT about mixing in out of band data? That's exactly what `does` is for. `App::GymScheduler` uses it liberally to earmark relevant `DateTime` objects; were it working with `Set`s instead, it'd be broken.

[...]

PR #1962, opened 2018-06-25
Keep .perl/.gist consistency among Set/Mix/Bag

### LINKED ISSUE #1958: Set.perl inconsistent with Bag/Mix.perl
The bag and mix get `.perl` in a form of `bag()`/`mix()`, but set gets dumped as either `().Set` or `set()` (kinda surprised `Set.new` and `set()` get dumped differently, given they compare `===` as True)

```
<Zoffix__> m: .perl.say for Set.new, Bag.new, Mix.new
<camelia> rakudo-moar e9351cbaa: OUTPUT: «().Set␤bag()␤mix()␤»
<Zoffix__> m: .perl.say for set(), bag(), mix()
<camelia> rakudo-moar e9351cbaa: OUTPUT: «set()␤bag()␤mix()␤»
<Zoffix__> m: dd set() === Set.new
<camelia> rakudo-moar e9351cbaa: OUTPUT: «Bool::True␤»
```

W4anD0eR96:
This will fix #1958

### DIFF
--- a/src/core/Setty.pm6
+++ b/src/core/Setty.pm6
@@ -143,33 +143,32 @@ my role Setty does QuantHash {
     multi method gist(Setty:D $ : --> Str:D) {
         nqp::concat(
           nqp::concat(
-            nqp::if(
-              nqp::istype(self,Set),
-              'set(',
-              nqp::concat(self.^name,'(')
-            ),
-            nqp::join(" ",
+            nqp::concat(self.^name,'('),
+            nqp::join(', ',
               Rakudo::Sorting.MERGESORT-str(
                 Rakudo::QuantHash.RAW-VALUES-MAP(self, *.gist)
               )
             )
           ),
-          ')'
+          ')',
         )
     }
     multi method perl(Setty:D $ : --> Str:D) {
         nqp::if(
-          nqp::eqaddr(self,set()),
-          'set()',
+          $!elems && nqp::elems($!elems),
           nqp::concat(
-            '(',
             nqp::concat(
-              nqp::join(",",Rakudo::QuantHash.RAW-VALUES-MAP(self, *.perl)),
-              nqp::concat(
-                ').',
-                self.^name
+              '(',
+              nqp::join(',',
+                Rakudo::QuantHash.RAW-VALUES-MAP(self, *.perl)
               )
-            )
+            ),
+            nqp::concat(').',self.^name)
+          ),
+          nqp::if(
+            nqp::istype(self,Set),
+            'set()',
+            nqp::concat('().',self.^name)
           )
         )
     }


lizmat:
FWIW, I seem to recall a reason to the inconsistency.  But don't have time to look at it in the next 48 hours.  Please refrain from merging until I've had time.

zoffixznet:
> FWIW, I seem to recall a reason to the inconsistency. 

There's a related ticket that could use your input as well: https://github.com/rakudo/rakudo/issues/1959

W4anD0eR96:
Let's say that `.gist` if for human readability, so consistency does no harm.

for `.perl`, `set()` is eqv to `().Set` and if it is not an empty set, nothing changed.

zoffixznet:
> Let's say that .gist if for human readability, so consistency does no harm.

Comma-consistency looks to be all over the place:

```
$ perl6 -e '.say for <a b c>, [<a b c>], Blob.new(1, 2, 3), <a b b>.SetHash, <a b b>.BagHash, <a b b>.MixHash, set(<a b b>), bag(<a b b>), mix(<a b b>)'
(a b c)
[a b c]
Blob:0x<01 02 03>
SetHash(a b)
BagHash(a, b(2))
MixHash(a, b(2))
set(a b)
Bag(a, b(2))
Mix(a, b(2))
```

W4anD0eR96:
@lizmat as a reminder.

lizmat:
Not sure, but on 12 April I did this: c0212af9b91bd7f916c7e2  so I wonder what has changed now to make it inconsistent again?

zoffixznet:
> so I wonder what has changed now to make it inconsistent again?

That commit doesn't affect the code path in the Issue.

Right now Set.perl checks `eqaddr` with `set()`, but Bag/Mix.perl doesn't do the same for `bag()`/`mix()`, which is where the inconsistency comes from.

However, that fact that `bag()`/`set()`/`mix()` always return the same object instead of creating a new one (which allows that `eqaddr` to work in the first place) IMO [is itself a bug](https://github.com/rakudo/rakudo/issues/1959).

So once [R#1959](https://github.com/rakudo/rakudo/issues/1959) is resolved, this inconsistency will also be resolved, and just the now-useless `.eqaddr` in `Set.perl` will need to be cleaned up.

zoffixznet:
> There's a related ticket that could use your input as well: #1959

That's now resolved as not-a-bug, so it's fine for `set()`/`bag()`/`mix()` to be `.perl`'ed differently since in Rakudo we intern those.

The actual inconsistency now is that `Set.new` returns a separate object, but `Mix.new`/`Bag.new` return an interned copy. So we can "legally" make `Set.new` return an interned copy as well. I'll update the ticket with this info.

lizmat:
AFAIK, Set.new / Bag.new / Mix.new return new instances.  Because this is what makes my %h is Set, my %h is Bag, my %h is Mix, work.

If we want go be a able to intern Set.new/Bag.new/Mix.new, we will need another way to fix my %h is … .

W4anD0eR96:
> Set.new / Bag.new / Mix.new return new instances

FWIW, they all call `nqp::create(self)`

lizmat:
Which is a new instance

zoffixznet:
> but Mix.new/Bag.new return an interned copy

correction: they just `.perl` like `bag()`, but they return a new object.

```
m: say Bag.new.perl, Bag.new =:= bag()
rakudo-moar 587cd4f9e: OUTPUT: «bag()False␤»
```

lizmat:
Ah that, will fix that.

W4anD0eR96:
for 84f60adad10f04b3ffa4a5747af9820f8244c2a2

PR #1995, opened 2018-06-30
Issue 1975: Fix WIN32 Proc::Async process termination issue 

jstuder-gh:
Signals are emulated via libuv on Windows, so the signal handler in the
test is not called. Rather upon receiving the emulated SIGINT, libuv will just call
WIN32's TerminateProcess() call. But the process isn't terminated immediately;
the process' kernel object is not destroyed until other processes with open
handles have released them.

The process is waiting for the handles to be released before it
terminates but Proc::Async is waiting on the process terminating before
it will close it's IO handles (if any open).

Fixing this by calling done on any available suppliers within the kill
method so that the promises are no longer blocked on (in this test
the promises are being awaited). The IO handles are closed and then the
process is terminated.

Note that this solves the issue of the process hanging indefinitely, but
it does not cause the test to pass, since for it to pass the process
would need to receive the signal, execute the signal handler, and send
the right data to its stdout.

### DIFF
--- a/src/core/Proc/Async.pm6
+++ b/src/core/Proc/Async.pm6
@@ -103,11 +103,11 @@ my class Proc::Async {
     has $.enc = 'utf8';
     has $.translate-nl = True;
     has Bool $.started = False;
-    has $!stdout_supply;
+    has $!stdout_supplier;
     has CharsOrBytes $!stdout_type;
-    has $!stderr_supply;
+    has $!stderr_supplier;
     has CharsOrBytes $!stderr_type;
-    has $!merge_supply;
+    has $!merge_supplier;
     has CharsOrBytes $!merge_type;
     has $!stdin-fd;
     has $!stdout-fd;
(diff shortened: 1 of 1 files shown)


src/core/Proc/Async.pm6
```diff
@@ -412,15 +424,24 @@ my class Proc::Async {
     proto method kill(|) {*}
     multi method kill(Proc::Async:D: Signal:D \signal = SIGHUP) {
         X::Proc::Async::MustBeStarted.new(:method<kill>, proc => self).throw if !$!started;
+        $!stdout_supplier.done() if $!stdout_supplier;
```
ugexe: Will users expect `&done` or `&quit` callbacks to be fired? I'm not really sure.

src/core/Proc/Async.pm6
```diff
@@ -140,7 +140,19 @@ my class Proc::Async {
           if the-supplier and type != value;
 
         type           = value;
-        the-supplier //= Supplier::Preserving.new;
+        the-supplier //= class :: is Supplier::Preserving {
```
ugexe: Anonymous classes were removed from the code base for performance reasons.

src/core/Proc/Async.pm6
```diff
@@ -260,11 +272,11 @@ my class Proc::Async {
 
     method !capture(\callbacks,\std,\the-supplier) {
         my $promise = Promise.new;
-        my $vow = $promise.vow;
+        the-supplier.vow = $promise.vow;
```
ugexe: This seems race-y ( although I haven't bothered to investigate that hunch )

src/core/Proc/Async.pm6
```diff
@@ -412,15 +424,24 @@ my class Proc::Async {
     proto method kill(|) {*}
```
ugexe: Might be worth consolidating all the duplicated logic (including some existing) into the proto:
```
proto method kill(|) {
    X::Proc::Async::MustBeStarted.new(:method<kill>, proc => self).throw if !$!started;
    $!stdout_supplier.done() if $!stdout_supplier;
    $!stderr_supplier.done() if $!stderr_supplier;
    $!merge_supplier.done()  if $!merge_supplier;
    {*}
}
```

jstuder-gh: I had left the duplicate logic because of the comment a few lines above stating that it significantly impacted performance. That comment is quite old though so things could be different now with the improvements made to the optimizer and spesh.

src/core/Proc/Async.pm6
jstuder-gh: Looking at it again, you're right. I figure this particular method is probably alright (since it's a private method called with 3 different suppliers each time), the 'start' method could definitely cause a data race if called multiple times from different threads. I've updated the branch to add a commit wrapping it in a Lock.

I've cherry-picked that commit and will submit it on it's own, so that it can be considered regardless of what we decide to do with this PR.

src/core/Proc/Async.pm6
jstuder-gh: Currently, the ```killprocasync``` op invoked in the method call always cancels the work on the event queue ([in MoarVM at least](https://github.com/MoarVM/MoarVM/blob/master/src/io/procops.c#L1007)). However, not knowing that, it isn't unreasonable that the user might expect that the process could recover from the signal.

I was originally approaching it from the standpoint that the user is going in expecting the termination of the process, but I'm not so sure now. Might be too late to change the behavior though?

JJ:
There are many outstanding issues, as well as a conflict now. It's probably better to close this PR and open it back again. Any other opinion?

lizmat:
Closing because of lack of comments and opinions.

Issue #1941, opened 2018-06-20
Document Rakudo-specific classes to support Perl 6 documentation


JJ:
## The Problem

Whenever a Rakudo-specific issue arises in [perl6/doc](https://docs.perl6.org), we are of two minds. On one hand, it's rakudo specific, so it's subject to change. On the other hand, we do need to document those Rakudo features so that people can use them. Also, we need to understand them to document some parts of Perl 6, which are public-facing and where, for lack of documentation, we have to go directly to the source to understand what a function actually does. 

In any case we might have to go to the source, but having some Rakudo-specific, searchable documentation, showing class hierarchy, or a brief description of the functionality, would really save us a lot of work. And probably help Rakudo implementors too.

## Expected Behavior

It would be nice if some class specific documentation was created in this area. For instance, [it would be useful to know the behavior of the CompUnit classes](https://github.com/rakudo/rakudo/blob/98e575de00db42101d4879e43fe740a34791c418/src/core/CompUnit/Repository/Spec.pm6), which are [very slightly documented](https://docs.perl6.org/language/modules-core#index-entry-CompUnit_%28Rakudo_classes%29) (with "documented" mainly meaning source is linked from a page). I am not sure whether this should actually go to this repo or to the docs repo (and if that's eventually the decision, feel free to open an issue there), but I'm positive we need this kind of thing...

coke:
As long as it's clear that a thing is compiler specific, I'm OK with it being in the docs.

zoffixznet:
FWIW, I don't think CompUnit types are meant to be rakudo specific. They just never made it into the spec because IIRC they were all built just a month or two before first lang release.

JJ:
This is not CompUnit specific. It's a more general concern. The problem is that when some feature is requested in the doc issues, you have to look up in a couple of places to really find out if it's Rakudo specific or not. And if they are, generally they get low priority over other things. Just a stub of documentation, maybe attached to the class itself, could really help, at least to know the specificity of the thing and maybe learn a little bit about the behavior instead of having to wade through the source to know what's going on. 
WRT to CompUnit itself, there's a lot of functionality that is just not being used due to lack of knowledge and thus documentation... Maybe we can return this issue back to perl6/doc (and close it) saying "Improve *that* documentation", but actually it's not the only case, just the most outstanding.

zoffixznet:
> if it's Rakudo specific or not. And if they are, generally they get low priority over other things.

[...]

PR #2004, opened 2018-07-01
Some typos and updating of variable names

JJ:
In fact, this would contribute towards solving #1941. It's really helping me, at least.


zoffixznet:
👍 Thanks.

Issue #1990, opened 2018-06-28
Opt opportunity in METAOP_ASSIGN with % and @ vars


zoffixznet:
***See [Newcomer Guide to Contributing to Core Perl 6](https://rakudo.party/post/Newcomer-Guide-to-Contributing-to-Core-Perl-6) for tips on fixing this Issue***

----------------

e.g. `my %h = :42foo; %h ,= :100bar` on `2018.05-54-g148d7c5` still got `Op (callstatic &METAOP_ASSIGN)` in it, which can be rewritten to p6store.

Very likely all that needs to be done is to add a `@` and `%` sigil branch here that sets `$assignop` to `"p6store"`: https://github.com/rakudo/rakudo/blob/2177ee234cda52d05763693fc2d54ed7aa333f75/src/Perl6/Optimizer.nqp#L1916-L1921

It's possible the tests covering the behaviour are sparse and it might be a good idea to write a few to cover this opt.

dogbert17:
Fixed with commit 55b2e32b95aa90f491d, tests needed.

lizmat:
Not sure how we can test this, suggestions anyone?

zoffixznet:
> Not sure how we can test this, suggestions anyone?

The lack of tests was just a guess on my part. I don't know if there are any for the metassign for `%` and `@` sigils. The tests should just test the metassign to such variables with some ops.

Something like this (needs to be different for array, probably, 'cause in those `,=` makes a self-referential one)

```perl6
my %h = :42foo, :70bar; %h ,= {:100foo, :50meow}; use Test; is-deeply %h, {:bar(70), :foo(100), :meow(50)}, ',= meta assign op with a %-sigilled var'
<camelia> rakudo-moar 08b449e1a: OUTPUT: «ok 1 - ,= meta assign op with a %-sigilled var␤»
```

lizmat:
Closing as we won't be working on the legacy optimizer anymore

PR #2012, opened 2018-07-02
Fix R#1990

dogbert17:
Zoffix++, in https://github.com/rakudo/rakudo/issues/1990, pointed
out that there's an optimization opportunity in optimize_nameless_call
where the, faster, p6store op can be used instead when assigning
@ and % sigiled variables

### DIFF
--- a/src/Perl6/Optimizer.nqp
+++ b/src/Perl6/Optimizer.nqp
@@ -1916,6 +1916,9 @@ class Perl6::Optimizer {
               elsif $sigil eq '$' {
                 $assignop := 'assign';
               }
+              elsif $sigil eq '@' || $sigil eq '%' {
+                $assignop := 'p6store';
+              }
               else {
                 # TODO support @ and % sigils and check what else we need
                 # to "copy" from assign_op in Actions


src/Perl6/Optimizer.nqp
```diff
@@ -1916,6 +1916,12 @@ class Perl6::Optimizer {
               elsif $sigil eq '$' {
                 $assignop := 'assign';
               }
+              elsif $sigil eq '@' {
+                $assignop := 'p6store';
+              }
+              elsif $sigil eq '%' {
+                $assignop := 'p6store';
+              }
```
zoffixznet: You can just write a single conditional as `elsif $sigil eq '@' || $sigil eq '%' {...}` and avoid duplication of the body.

src/Perl6/Optimizer.nqp
```diff
@@ -1916,6 +1916,9 @@ class Perl6::Optimizer {
               elsif $sigil eq '$' {
                 $assignop := 'assign';
               }
+              elsif $sigil eq '@' || $sigil eq '%' {
+                $assignop := 'p6store';
+              }
               else {
                 # TODO support @ and % sigils and check what else we need
```
vendethiel: can this comment now be removed?

lizmat: removed with 2a414aee18

PR #2015, opened 2018-07-03
Fix potential race condition in EVAL

ugexe:


src/Perl6/Grammar.nqp
```diff
@@ -472,7 +472,18 @@ role STD {
 }
 
 grammar Perl6::Grammar is HLL::Grammar does STD {
-    my $sc_id := 0;
+    my class SerializationContextId {
+        my class Lock is repr('ReentrantMutex') { };
```
jnthn: There's an [NQPLock class](https://github.com/perl6/nqp/blob/4fd4b48afb45c8b25ccf7cfc5e39cb4bd658901d/src/core/NQPLock.nqp#L1) in the NQP setting; please use that.

Issue #1906, opened 2018-06-05
[RESOLVED; NEED TEST] &prompt should pass stuff through `val`


zoffixznet:
`&prompt` should behave similar to how `sub MAIN` behaves. Namely, it should convert numeric data to allomorphs (we also do the enum thing in `MAIN`, but I think doing it in `&prompt` is overkill).

Since stuff typed at the terminal is always stringy, we convert it to allomorphs when dealing with
`sub MAIN` args:

```
$ perl6 -e 'sub MAIN(UInt:D $id) { dd [$id] }' 42
[IntStr.new(42, "42")]
```

`&prompt` similarly takes stringy-only stuff from the "terminal", so I'd expect it to offer a similar conventience. 

```
$ perl6 -e 'my UInt:D $id = prompt "Enter ID of song to play: "'                                                                                                                                  
Enter ID of song to play: 42
Type check failed in assignment to $id; expected UInt:D but got Str ("42")
  in block <unit> at -e line 1

$ perl6 -v
This is Rakudo version 2018.05-43-gc80ea2f28 built on MoarVM version 2018.05-34-gdb5029122
implementing Perl 6.c.
```

So instead of being just `print '...' and get`, `&prompt` would be equivalent to `print '...' and val get`

PR #2018, opened 2018-07-04
Issue 1906

jkramer:
Pass &prompt result through &val (quietly).

Resolves rakudo/rakudo#1906.


zoffixznet:
Instead of using `quietly` you can just use something cheaper, like `nqp::defined(my \res := $*IN.get) ?? al(res) !! res`

This would also avoid silencing other warnings, if the code in question should start emitting any in the future.

zoffixznet:
:+1: Thanks!

PR #2026, opened 2018-07-04
Display build/commit info on config/make

jstuder-gh:


JJ:
Well, this is totally salvageable, since it adds a new functionality, but it includes a lot of old code and would have to be tweaked to work. Is it worth the while?

vrurg:
@JJ Neither I see what problem does it solve. I'd close it.

JJ:
OK, closing them. Maybe open a new PR if someone think this is still necessary.

Issue #2022, opened 2018-07-04
LTA error: "not found module" message does not include file


zoffixznet:
The error says "at line 1", but line 1 of which file? Many projects have multiple files with modules

```
$ echo "use Blah" > Foo.pm6
$ perl6 -I. -MFoo -e ''
===SORRY!===
Could not find Blah at line 1 in:
    /tmp/tmp.9BKT1VcNMJ
    /home/zoffix/.perl6
    /home/zoffix/rakudo/install/share/perl6/site
    /home/zoffix/rakudo/install/share/perl6/vendor
    /home/zoffix/rakudo/install/share/perl6
    CompUnit::Repository::AbsolutePath<67564160>
    CompUnit::Repository::NQP<45884840>
    CompUnit::Repository::Perl5<45884880>
$ perl6 -v
This is Rakudo version 2018.05-54-g148d7c5 built on MoarVM version 2018.05-57-g89dee3b
implementing Perl 6.c.
```

zoffixznet:
(for records, there's been additional discussion on the PR for this Issue: https://github.com/rakudo/rakudo/pull/2038#issuecomment-403482926 )

niner:
I see two ways to get the source information into the error message:
* we either communicate the source location from the caller down to where the error is created or
* we catch the error and enhance the message at the point where we have the required information

I think the second option is the way to go as it works without adding unrelated attributes to `CompUnit::DependencySpecification` and without requiring each CompUnit::Repository implementation to support it. All that's required is catching the X::CompUnit::UnsatisfiedDependency, adding the source location info and then rethrowing it.

niner:
Implemented in https://github.com/rakudo/rakudo/commit/6f527cea9a

PR #2038, opened 2018-07-09
Include source file name in module-not-found error message.

jkramer:
Fixes rakudo/rakudo#2022.

I'm a bit unhappy with the attribute name `source-file-name` because as far as I understand the specification is about the thing that's to be loaded, not the place where it's loaded *from*, so the naming is a bit confusing IMO. But then again `source-line-number` already existed so I went along.

### DIFF
--- a/src/Perl6/World.nqp
+++ b/src/Perl6/World.nqp
@@ -1266,7 +1266,8 @@ class Perl6::World is HLL::World {
             :auth-matcher(%opts<auth> // $true),
             :api-matcher(%opts<api> // $true),
             :version-matcher(%opts<ver> // $true),
-            :source-line-number($line)
+            :source-line-number($line),
+            :source-file-name(self.current_file)
         );
         self.add_object($spec);
         my $registry := self.find_symbol(['CompUnit', 'RepositoryRegistry'], :setting-only);
--- a/src/core/CompUnit/DependencySpecification.pm6
+++ b/src/core/CompUnit/DependencySpecification.pm6
@@ -1,6 +1,7 @@
 class CompUnit::DependencySpecification {
     has str $.short-name is required;
     has int $.source-line-number = 0;
+    has str $.source-file-name = '';
     has str $.from = 'Perl6';
     has $.version-matcher = True;
     has $.auth-matcher = True;
--- a/src/core/Exception.pm6
+++ b/src/core/Exception.pm6
@@ -2958,9 +2958,10 @@ my class X::CompUnit::UnsatisfiedDependency is Exception {
     method message() {
         my $name = $.specification.short-name;
         my $line = $.specification.source-line-number;
+        my $file = $.specification.source-file-name;
         is-core($name)
             ?? "{$name} is a builtin type, not an external module"
-            !! "Could not find $.specification at line $line in:\n"
+            !! "Could not find $.specification at $file:$line in:\n"
                 ~ $*REPO.repo-chain.map(*.Str).join("\n").indent(4)
                 ~ ($.specification ~~ / $<name>=.+ '::from' $ /
                     ?? "\n\nIf you meant to use the :from adverb, use"


ugexe:
Why not just use —ll-exception?

jkramer:
> Why not just use —ll-exception?

I was not aware of that option, but I just tried it and even though it prints a stack trace with all the internal calls, the actual source file/line that contains the `use` does not seem to be included.

lizmat:
Will you also add tests for this?  If so, where should they be added?

ugexe:
Here are my issues with this PR:

1) `source-file-name` doesn't make sense as a name for the reason you already mentioned. However I would point out that `source-line-number` doesn't imply anything about coming from a file.

2) The error really shouldn't explicitly mention coming from a *file* -- it might not be from a file.

3) What `source-file-name` points at might not even exist on disk when shown to the user if all that remains is the precomp'd version of the file. This could be confusing.

4) It is probably a breaking change since I expect this to be in use: `perl6 -e 'try { require XXX; CATCH { default { say "Missing module: " ~ ($_ ~~ /^"Could not find "(.*?)" at"/ ?? $0 !! "") } } }'`

jkramer:
> Will you also add tests for this? If so, where should they be added?

I was looking into existing tests regarding this exception, but apparantly all that's tested currently is that it is thrown when `use`ing a non-existing module: https://github.com/perl6/roast/blob/master/S32-exceptions/misc.t#L149 (similar tests at two more places). The actual output text of the error isn't tested anywhere yet and I'm really sure whether it would make sense to test that and if so, how to go about it.

lizmat:
Thanks!

zoffixznet:
I think due to email-vs-github interface use of the tracker, there's been premature merge here.

There's been some missed feedback from ugexe above.

All valid points, except maybe the last one: IMO such code is fragile and shouldn't have been in use, given the exception `X::CompUnit::UnsatisfiedDependency` is thrown.

jkramer:
@ugexe @zoffixznet 

I agree on the first point. As I pointed out in the initial comment I think the naming is confusing. Maybe `source-name` would be a better name taking into account that it indeed might not be a file. But it's still confusing in a way because it's not about the source of the unit to be loaded (which the class is representing) but line of code where it's loaded from. Actually I'd even argue that these two attributes don't even belong into this class.

I don't see point 2 & 3 so critically as the message for non-files is not very misleading IMO:

```
# perl6 -e 'use Asd' 
===SORRY!===
Could not find Asd at -e:1 in:
[...]

# perl6
To exit type 'exit' or '^D'
> use ASD
Could not find ASD at <unknown file>:1 in:
[...]
```

If anything I would maybe change `<unknown file>` to `<unknown source>` or similar in the latter case.

zoffixznet:
.oO( `.requested-from-line`, `.requested-from-source` )

lizmat:
patches welcome  :-)

jkramer:
> .oO( .requested-from-line, .requested-from-source )

I like it, updated my branch accordingly.

niner:
I really wonder if source-line and source-file-name should be attributes of CompUnit::DependencySpecification at all. Feels like we're abusing this class for a different purpose to what it's intended for. According to the design "a dependency specification is how one module or script's need for another module is declared. In source code, it follows a `use` or `need` statement. An example is `Sereal:auth<cpan:*>:ver(1..*)`. The same syntax is used in the `depends` section of a `META6.json`."
The CompUnit::DependencySpecification class is the structured way to represent such a dependency specification. Seems weird to have it contain information about something else, too.

ugexe:
@niner I agree. I think `file` even has a place still, although it would be to serve the role of e.g.
`require Test:file</home/ugexe/XXX.pm6>;` / https://github.com/perl6/roast/blob/e0d71a2b64d7d25202fcdc7fb5f60731a563e542/S11-modules/require.t#L60 which might also be able to solve the issue at hand.

Issue #1453, opened 2018-01-29
Please improve the class IO::Path routines copy() and move() 


kolikov:
## Methods copy and move expect only a file as their $to argument 
my $src   = IO::Path.new('/home/zz/file.gz');
my $dest  = IO::Path.new('/home/zz/tmp');
$src.move($dest);

## Actual Behavior
Failed to move '/home/zz/zz.tar.gz' to '/home/zz/tmp': Failed to move file: illegal operation on a directory.

## Expected Behavior
As a the good old 'cp' or 'mv'  commands in Linux (move/copy the file into the target directory), as IO::Path has methods to identify files from directories, Make the 'move' and 'copy' a multi sub with a signature test for a file on $from and directory for $to.

That would be BTA 👍 
Except if it breaks too much things ... or @zoffixznet 's heart  :-)

## Environment
* Operating system: Linux 4.14.0-3-amd64 #1 SMP Debian 4.14.13-1 (2018-01-14) x86_64
* Compiler version (`perl6 -v`): Rakudo version 2017.10 MoarVM version 2017.10  Perl 6.c.

zoffixznet:
👍 Makes sense to me.

Would also mirror how Perl 5's core [File::Copy](https://metacpan.org/pod/File::Copy) behaves.

zoffixznet:
> +1 Makes sense to me.

After the discussions on https://github.com/rakudo/rakudo/pull/2043 , IMO this should be an ecosystem module first

doomvox:
This issue is still confusing people: 

https://stackoverflow.com/questions/71909580/move-function-does-not-work-in-raku-in-windows

PR #2043, opened 2018-07-10
Allow directories as destinations for IO::Path.copy/move.

jkramer:
Fixes rakudo/rakudo#1453.

### DIFF
--- a/src/core/IO/Path.pm6
+++ b/src/core/IO/Path.pm6
@@ -448,7 +448,12 @@ my class IO::Path is Cool does IO {
         }}
     }
 
-    method copy(IO::Path:D: IO() $to, :$createonly --> True) {
+    method copy(IO::Path:D: IO() $to is copy, :$createonly --> True) {
+        nqp::if(
+            $to.d,
+            ($to.=add(self.basename))
+        );
+
         $createonly and $to.e and fail X::IO::Copy.new:
             :from($.absolute),
             :to($to.absolute),


zoffixznet:
It would be good for this type of thing to come with thorough test coverage that's examined on multiple OSes before merging.

For example, normally, if you try to `.move` a file onto itself, you get an error:

```
$ perl6 -e '"foo".IO.move: "foo"'
Failed to move '/tmp/tmp.tMKFiJOrMN/foo' to '/tmp/tmp.tMKFiJOrMN/foo': source and target are the same
  in block <unit> at -e line 1
```

However, if you move it onto itself this way, you get a program clobbers the file and hangs on the first run, and then removes it entirely on the second run:

```
$ echo "test" > foo
$ mkdir cont
$ perl6 -e 'sub c (IO() \SELF, IO() $to is copy) { $to.=add: SELF.basename if $to.d; SELF.move: $to}; "foo".&c: "cont/../"'
^C
$ cat foo 
$ perl6 -e 'sub c (IO() \SELF, IO() $to is copy) { $to.=add: SELF.basename if $to.d; SELF.move: $to}; "foo".&c: "cont/../"'
$ cat foo
cat: foo: No such file or directory
```

Also, consider this setup with a broken symlink to a directory:

```
$ mkdir orig
$ mkdir cont
$ echo "test" > foo
$ ln -s orig cont/broken
$ tree
.
├── ./cont
│   └── ./cont/broken -> orig
├── ./foo
└── ./orig
```

If you try to copy a file into that symlink with `cp`, you get an error:

```
$ cp foo cont/broken
cp: not writing through dangling symlink `cont/broken'
```

But using the sub that I believe replicates the code in this PR, we copy the file into what appears to be the name of the dir the broken symlink points to:

```
$ perl6 -e 'sub c (IO() \SELF, IO() $to is copy) { $to.=add: SELF.basename if $to.d; SELF.copy: $to}; "foo".&c: "cont/broken"'
$ tree
.
├── ./cont
│   ├── ./cont/broken -> orig
│   └── ./cont/orig
├── ./foo
└── ./orig

$ cat cont/orig 
test
```

----

Perl 5's `File::Copy` module does the same, but can we do better?

```
$ perl -MFile::Copy -e 'copy foo => "cont/broken"'
$ tree
.
├── ./cont
│   ├── ./cont/broken -> orig
│   └── ./cont/orig
├── ./foo
└── ./orig

2 directories, 3 files
```

--------

Also, would making all these things work right totally kill performance for the old-style `.copy`? Maybe this should be trialed as a module first and/or added under a special named arg to enable the extra behaviour.

jkramer:
> However, if you move it onto itself this way, you get a program clobbers the file and hangs on the first run, and then removes it entirely on the second run

Your example lacks the sameness check that IO.copy/move already have builtin, so this problem (at least this specific case) is already taken care of and the following tests (for example) pass:

```
throws-like { $file.move($file.dirname) }, X::IO::Move, 'should throw if source/target are the same file';
throws-like { $file.move($file.dirname.IO.add("cont").add("..")) }, X::IO::Move, 'should throw if source/target are the same file';
```

But I agree that this should be thorougly tested on all supported operating systems.

zoffixznet:
> Your example lacks the sameness check

Ah, right, in the PR, the check is done first, but in my code the addition of a filename is done first.

ugexe:
Consider a dynamically generated `$dest-filename` never intended to be copied to a directory, even if a directory named the same thing as `$dest-filename` happens to get created at just the wrong time. This behavior might be ok for a command line utility, but I think a little more control at the language level would be better.

One way to serialize the difference between dirs and files as a string is for dirs to always have a trailing `/`, but alas we don't have such a thing ( yet? )... another way would be if we had `IO::Dir` and `IO::File` types so one could be explicit and let a multi candidate choose the proper logic.

zoffixznet:
> if we had IO::Dir and IO::File types

new-io branch had these, but the problem (at least with those implementations) is you have no idea what your path actually points to until you query the filesystem, and after you do you can't just continue assuming it'll point to a dir forever.

Now this makes me think even more that this should be an ecosystem module first.

zoffixznet:
> Now this makes me think even more that this should be an ecosystem module first

Especially because as someone on IRC pointed out, this behaviour is not just for `.copy` and `.move` but also applies to `.link`/`.symlink`. And probably some other methods too (like `.rename`)

ugexe:
I would lean towards `IO::File` and `IO::Dir` being an explicit way of telling what behavior you want to happen. I imagine a generic `IO::Path` candidate which does a redispatch based on the `.f / .d` checks ( that likely happen within most such methods already) . i.e.:

```
role IO::File {  };
role IO::Dir { };

multi sub foo(IO::File $f) { say "file" };
multi sub foo(IO::Dir $d) { say "dir" };
multi sub foo($p where IO::Path) { samewith($p.f ?? ($p but IO::File) !! ($p but IO::Dir)) };

my $file = "$*CWD/foo.txt".IO but IO::File;
my $dir  = $*CWD but IO::Dir;

foo($file);
foo($dir);
foo($*CWD);
```

I wouldn't necessarily expect for any methods to return these over a generic `IO::Path`. More as a ( probably slow due to extra candidates ) method of overriding the various `$path.f ?? this() !! that()` checks of things taking an `IO::Path`. Of course this ignores that there would likely be logic shared between various `IO::Dir` and `IO::File` candidates, such that the design of such methods would need an additional routine to accommodate that logic.

I think it could be figured out if someone thought and cranked on it for long enough, but I'm under no impression that it would be easy. Useful? Yes -- at least I think so.

zoffixznet:
> an explicit way of telling what behavior 

That looks like an overly complicated way of using named arguments:

```perl6
multi sub foo(IO $, :$force-file!) { say "file" }
multi sub foo(IO $, :$force-dir! ) { say "dir" }
multi sub foo(IO $_) { foo $_, |(.f ?? :force-file !! :force-dir) }

foo :force-file, $*CWD.add: 'foo.txt';
foo :force-dir,  $*CWD;
foo $*CWD;
```

because something needs to create the IO::File/IO::Dir objects, and once it does, you're faced with the original problem that they don't actually represent the reality of the filesystem.

JJ:
OK, reading through the discussion, I see this is a feature that would be kinda-OK, but no one actually requested, it's not thoroughly tested, and might cause optimization and other kind of problems. So at this point, I'm closing the PR and [creating this issue](https://github.com/Raku/problem-solving/issues/181) to discuss possible solutions.

tbrowder:
See my module File::Copy which is a stab at this problem.

JJ:
One of the suggestions was precisely that, to move it to the ecosystem. Makes sense. And thanks!

tbrowder:
I need help with my modue [File::Copy](https://github.com/tbrowder/File-Copy), or at least a reality check. So far, at the module level, I can't get past the failure on detecting the $to as a directory, even after using the method of @jkramer. I haven't yet tried more nqp changes,

PR #2051, opened 2018-07-11
Improve the error message for slicing a shaped array

### LINKED ISSUE #1943: LTA error when using `,` instead of `;` to index a shaped array
Here, a user accidentally used a `,` instead of `;` to index a cell in a two-dimentional shaped array. The error talks about partially-dimentioned views being NYI.

Took me some time to realize what the actual problem in the code was, but another user might think that indexing a shaped array is NYI. Perhaps, the error should suggest the user didn't use proper indexing.

```
$ perl6 -e 'my @a[2;2]; say @a[1,1]'
Partially dimensioned views of shaped arrays not yet implemented. Sorry.
  in block <unit> at -e line 1

$ perl6 -v
This is Rakudo version 2018.05-54-g148d7c5 built on MoarVM version 2018.05-57-g89dee3b
implementing Perl 6.c.
```

bduggan:
This generates a better NYI error message for the case when a user takes a slice of a shaped array, and it looks like they may have inadvertently used a comma instead of a semicolon to access an element of the shaped array.

Before:
```
$ perl6 -e 'my @x[2;2]; @x[0,1]'
Partially dimensioned views of shaped arrays not yet implemented. Sorry.
  in block <unit> at -e line 1
```
After:
```
$ ./perl6 -e 'my @x[2;2]; @x[0,1]'
Slices of shaped arrays not yet implemented. Sorry.
Did you mean: 0;1?
  in block <unit> at -e line 1
```
Resolves #1943

### DIFF
--- a/src/core/array_slice.pm6
+++ b/src/core/array_slice.pm6
@@ -86,14 +86,6 @@ multi sub POSITIONS(
         nqp::bindattr(pos-list, List, '$!todo', todo);
     }
     else {
-        if nqp::defined(SELF.shape) and !nqp::eqaddr(pos.WHAT,Range) {
-            if SELF.shape.elems > 1 {
-                X::NYI.new(
-                    feature => 'Slices of shaped arrays',
-                    did-you-mean => pos.join(';'),
-                ).throw;
-            }
-        }
         pos-iter.push-all: target;
     }
     pos-list
@@ -240,6 +232,12 @@ multi sub postcircumfix:<[ ]>( \SELF, Iterable:D \pos ) is raw {
     nqp::iscont(pos)
       ?? SELF.AT-POS(pos.Int)
       !! POSITIONS(SELF, pos).map({ SELF[$_] }).eager.list;
+}
+multi sub postcircumfix:<[ ]>( Array \SELF where SELF.shape.elems > 1, Iterable:D \pos ) is raw {
+   X::NYI.new(
+       feature => 'Slices of shaped arrays',
+       did-you-mean => (pos.join(';') unless nqp::eqaddr(pos.WHAT,Range)),
+   ).throw;
 }
 multi sub postcircumfix:<[ ]>(\SELF, Iterable:D \pos, Mu \val ) is raw {
     # MMD is not behaving itself so we do this by hand.
--- a/src/core/array_slice.pm6
+++ b/src/core/array_slice.pm6
@@ -86,6 +86,14 @@ multi sub POSITIONS(
         nqp::bindattr(pos-list, List, '$!todo', todo);
     }
     else {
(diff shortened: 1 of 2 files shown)


zoffixznet:
([R#1943](https://github.com/rakudo/rakudo/issues/1943) is a ticket for this PR)

bduggan:
> Isn't there a place for shaped arrays only where you could do this check and throw?

I'm not sure; I tried making changes to [ShapedArray.pm6](https://github.com/rakudo/rakudo/blob/master/src/core/ShapedArray.pm6) -- but it seemed to be too late at that point; `AT-POS` was already being called multiple times for a slice, e.g. 
```
$ cat pos.p6
use MONKEY;

augment class Str {
    method AT-POS($arg) {
        say "called with $arg";
    }
}

"abc"[1,2];
$ perl6 pos.p6
called with 1
called with 2
$
```
Any other ideas I might investigate?

bduggan:
I've [changed](https://github.com/rakudo/rakudo/pull/2051/commits/56e1111ce36eaa95e44c5b9bec93c80cb1698e7f) this to use multi-dispatch on the post-circumfix operator (and added a test, and ensured that the travis tests pass).  But I'm not sure if this addresses the potential performance hit for non-shaped arrays.

bduggan:
Yes, I guess there is a slight hit;
```
./perl6 --profile -e 'my @x = 1..10; my $y; for ^10000000 { $y = @x[$++ % 10]; }'
```
[before](https://rawgit.com/bduggan/7a6f875fca7fc47dd84565952edd31e0/raw/profile-1531410055.781473.html#/overview) (14824.59ms) vs [after](https://rawgit.com/bduggan/70455720b4e63e3cb88b7d69a7ecc42f/raw/profile-1531409847.1478279.html#/overview) (15389.48ms)

lizmat:
Ok, so we're losing 3% performance here just in order to get a better error message if it goes wrong?

Perhaps we should invest in trying to find out where an error occurred, and use that information to determine a better error message?  That way only *when* there is an error, will the performance penalty be paid.  And that's ok then, as the exception was already thrown then.

I vote we close this PR.

bduggan:
Sounds fine to me -- I'll close it.  I was thinking even better would be to just implement the NYI feature -- it seems to work if the array isn't declared as shaped, so maybe having that as a fallback would be a start (if I'm understanding correctly what the behavior should be).
```
 $ perl6 -e 'my @a[4;3] = <a b c>, <d e f>, <g h i>, <j k l>; say @a[0,1]'
Partially dimensioned views of shaped arrays not yet implemented. Sorry.
  in block <unit> at -e line 1

~ $ perl6 -e 'my @a = <a b c>, <d e f>, <g h i>, <j k l>; say @a[0,1]'
((a b c) (d e f))
```

PR #2065, opened 2018-07-13
Remove undocumented/unspecced gethostname()

ugexe:


ugexe:
>  um. what's your alternative to it? because it's kinda needed in network code

`$*KERNEL.hostname` or some such

T1S0N:
@ugexe there is not a `$*KERNEL.hostname`. So really how could we get hostname of this comp.

zoffixznet:
> @ugexe there is not a $*KERNEL.hostname

I think the point was is were something for getting the hostname specced, it should go into `$*KERNEL.foo` rather than be a standalone sub among hundreds of others.

> So really how could we get hostname of this comp.

NativeCall works:

```
<Zoffix__> m: use NativeCall; sub gethostname(CArray[int8], int8) returns int32 is native { * }; gethostname((my $h := CArray[int8].allocate(65)), 65); my $host := Blob.new: $h; dd $h[0..^ $h.first: 0, :k]».chr.join
<camelia> rakudo-moar 3905ac0bc: OUTPUT: «"camelia"␤»
```

scovit:
For this kind of C api where a caller allocated buffer is requested, I wrote the `NativeHelpers::CBuffer` module:
```perl6
use NativeCall;
use NativeHelpers::CBuffer;

sub gethostname(CBuffer, size_t) returns int32 is native {*};

my $buffer = CBuffer.new(100);
gethostname($buffer, 100) == 0 or die;
say $buffer'
```

another possibly wasteful alternative:
```perl6
say qqx/hostname/.chomp
```

jnthn:
We should provide this functionality in `CORE.setting`, even if not as the sub in question. I guess a `gethostname` sub has the benefit that folks who know the functionality under that name will easily find it. On the other hand, I'm at least as sympathetic to the argument that it doesn't really deserve to be a sub in the setting. Not entirely thrilled about `$*KERNEL`, but don't have a better suggestion. I'd approve of a PR that moves the functionality there; -1 to just toss it without deciding where it's moving to.

zoffixznet:
https://github.com/perl6/doc/commit/31c64c2e2bc11bdd299ea2af8659f858b2602a01
https://github.com/perl6/roast/commit/3cef1fdeba

PR #2066, opened 2018-07-13
Remove unused Universal role

ugexe:


ugexe:
I can't really tell from S28 what its purpose would be today... maybe it was going to be the underlying machinery to `class Foo:ver<1> { }`/`Foo.^ver`?

PR #2081, opened 2018-07-16
Rename IO::Handle's generic methods to .READ/.WRITE/.EOF

zoffixznet:
Per previously discussed 6.d TODO item.


zoffixznet:
This includes IO::CatHandle too, which just overrides these methods to throw X::NYI

Issue #1684, opened 2018-04-03
once doesn't work in C-style loop conditional


b2gills:
``` perl6
my $end = 9;

loop (
    my $i = 0
  ;
    $i < (once ++$end)  # <--
  ;
    ++$i
) {
  last if $i > 20
};

say $end;
```

This should print `10` but instead prints `31`

A workaround is to use `(my $ //= ++$end)`.

rakudo (2018.03.153.gc.53.fde.3)

jstuder-gh:
When running @lizmat's example with --target=ast, it returns this QAST subtree for the e2 (conditional expression) part of the c-style loop:

    ->    - QAST::Op(while) <sunk> loop (my $i = 0; $i < once { print $i; 10 }; ++$i)...
            - QAST::Op(chain &infix:«<») <wanted> <
              - QAST::Var(lexical $i) <wanted> $i
              - QAST::Op(if) <wanted>
    ->          - QAST::Op(p6stateinit)
                - QAST::Op(p6store) <wanted>
                  - QAST::Var(lexical once__1) <wanted>
                  - QAST::Op(call)
                    - QAST::Op(p6capturelex) <wanted> :past_block<?> :code_object<?>
                      - QAST::Op(callmethod clone)
                        - QAST::WVal(Block)  :uninstall_if_immediately_used<?> :past_block<?> :code_object<?>
                - QAST::Var(lexical once__1) <wanted>

```p6stateinit``` returns a truthy value if the current frame happens to be the first invocation of a closure (each clone of a block). In this instance it is the first invocation of the outer scope and each invocation of the ```p6stateinit``` in the QAST ```while``` is returning a truthy value, running the once block each time.

I have a series of PRs related to state initialization that could apply here as well, with a little tweaking. 
 [See here](https://github.com/rakudo/rakudo/compare/master...jstuder-gh:state_init_4_EXPERIMENT) for a modified version (WIP) that would address this issue. This coincides with [this MoarVM PR](https://github.com/MoarVM/MoarVM/pull/806) and [this NQP PR](https://github.com/perl6/nqp/pull/420), which create an ```is_hll_init``` value on a frame's coderef. With the modifications included in this linked branch it solves this issue and passes the existing spectest.

[...]

PR #2107, opened 2018-07-21
Fix Multiple Statevar Initialization Bugs

jstuder-gh:
This PR is one part of a series spanning the MoarVM, NQP, and Rakudo repositories.
[MoarVM](https://github.com/MoarVM/MoarVM/pull/916)
[NQP](https://github.com/perl6/nqp/pull/487)

This series of PRs modifies the state init check to instead keep a record of the statevars that have been 'HLL init-ed' and have that value determine whether to perform the assignment or not. If the statevar is not assigned to on the first invocation, it is shown as needing assignment the next time it is encountered despite what the MVM_FRAME_FLAG_STATE_INIT value is. If the statevar has been assigned to already and is encountered again within the same frame invocation (as the C-style loop example below), it is aware the that assignment has occurred and does not repeat it.

I had submitted a PR similar to this in the past, but this one has been rebased to a much newer HEAD and changed enough to the point that it feels appropriate to open a new PR.

### Specific modifications made are:

* Modified the ```p6stateinit``` extop to take a symbol as an argument. Within the function, lookup that symbol within the frame's lexical registry and determine whether the 'HLL assignment' has taken place. If so, do not assign again.

  * The value that persists whether an 'HLL assignment' has taken place is stored in the CodeRef object (alongside the statevar values themselves), which is unique to each closure. This 'isHllInit' value is a variable-length bit array, with each lexical represented by one bit.

* Added a new extop, ```p6stateinitbulk```, that will take a list of symbols and check multiple in one extop execution (for use in destructuring state assignments)

* Implemented for both the MoarVM and JVM backends.

### Issue being addressed:

In the past, we have determined whether to perform an assignment to a state variable by determining whether or not it was the first invocation of a closure (or clone of a frame). In MoarVM, an MVM_FRAME_FLAG_STATE_INIT value is set on the first frame of a given closure.

This works most of the time, but there are some instances where this is producing undesired results (see [RT #102994](https://rt.perl.org/Public/Bug/Display.html?id=102994))

```perl6
sub f($x) {
    return if $x;
    state $y = 5;
    $y;
} 

f(1);
say f(0);
# OUTPUT:   (Any)
# EXPECTED: 5
```

Since during the first invocation of this closure (when MVM_FRAME_FLAG_STATE_INIT is set) the statevar is not assigned to, on the next invocation the assignment is skipped and the unassigned value (Any) is returned.

In the case of 'once' with a C-style loop conditional (see [GH issue 1684](https://github.com/rakudo/rakudo/issues/1684)), this method also produces some unexpected results.

```perl6
loop (my $i = 0; $i < once { print $i; 10 }; ++$i ) { }
# OUTPUT:   012345678910
# EXPECTED: 0
```

In this case, the 'once' appears in the e2 expression in the C-style loop. The statevar init check occurs within the loop conditional and is executed multiple times within the same initial closure invocation. Because of this, MVM_FRAME_FLAG_STATE_INIT is set and the code within the block (containing print) is executed for each iteration.

### DIFF
--- a/docs/ops.markdown
+++ b/docs/ops.markdown
@@ -213,7 +213,18 @@ Sets the "pre" flag on the current frame.
 * p6sort(Mu @data, Mu &comparator)
 
 ## p6stateinit
-* p6stateinit()
+* p6stateinit(str $sym)
+
+Looks up the given symbol to determine if the associated variable (within this closure instance) has been assigned a value by the HLL. If so, we do not want to assign again.
+
+## p6stateinitbulk
+* p6stateinitbulk(list_s $symlist)
+
+Given a list of symbols, determine if the associated variables (within this closure instance) have been assigned a value by the HLL. Used to perform the state-initialization check on multiple variables at once (like in the case of a destructuring assignment).
+
+```perl6
+state ($a, $b, $c) = 1, 2, 3;
+```
 
 ## p6staticouter
 * p6staticouter(Mu $coderef)
--- a/src/Perl6/Actions.nqp
+++ b/src/Perl6/Actions.nqp
@@ -2671,7 +2671,7 @@ class Perl6::Actions is HLL::Actions does STDActions {
           QAST::Op.new(
             :op('if'),
             QAST::Op.new( :op('p6stateinit'),
-                QAST::Var.new( :name($sym), :scope('lexical') )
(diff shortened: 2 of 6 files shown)


lizmat:
Looks like this PR has bitrotted in light of new-disp.  So closing now.

PR #2116, opened 2018-07-22
Implement native-descriptor attribute for IO::Socket::Async

Kaiepi:
This allows one to get the raw file descriptor of clients and servers
created with IO::Socket::Async.


timo:
we'll probably have to have at least a stub in the JVM so that it pushes the right number of arguments for all the callbacks, find the code in NQP and perhaps take my last commit as inspiration :)

Kaiepi:
Done

Kaiepi:
Closing this since the way I worked out how to get the server's native descriptor as soon as it gets tapped needs to be two different pull requests

Issue #2110, opened 2018-07-22
Confusing error message for \1+ in regex


jmaslak:
## The Problem

A confusing error message is given when a particular invalid Perl 6 regex (that is something you would use in Perl 5) is used.

Example:

```
[0] red:jmaslak$ perl6 -e '"a" ~~ /(a)\1+$/'
===SORRY!===
Unrecognized backslash sequence (did you mean $0?)
at -e:1
------> "a" ~~ /(a)\1⏏+$/
Quantifier quantifies nothing
at -e:1
------> "a" ~~ /(a)\1+⏏$/
    expecting any of:
        infix stopper
        term
```

What is confusing is that there are two errors, when the problem isn't so much the +, but rather the \1.  I should have seen that this was two errors being returned, but I focused on the one near the end and started wondering "What the heck is an infix stopper?!"  The first error is awesome because it says exactly what you need to do to fix it, it's the second error that is confusing.

## Expected Behavior

Tell the user \1 is wrong, nothing about quantifiers following.

## Actual Behavior

See above.

## Steps to Reproduce

See above.

## Environment
* Operating system: Ubuntu 18.04
* Compiler version (`perl6 -v`): 

```
This is Rakudo version 2018.06 built on MoarVM version 2018.06
implementing Perl 6.c.

```

zoffixznet:
Agreed. I'd look at the last one as well and think why is it telling it follows nothing when I it does have stuff to follow.

> "What the heck is an infix stopper?!"

Filed as https://github.com/rakudo/rakudo/issues/2115

-------------------------------------------

Marking this as "good first issue". Tips for potential fixers for this:

The easiest fix is to just go into [src/Perl6/Grammar.nqp](https://github.com/rakudo/rakudo/blob/master/src/Perl6/Grammar.nqp), search for all occurences of "backslash sequence", and replace `<.sorry("…` with `<.panic("…`. There are several places where this message is thrown: in quote braid grammar as well as the regex one.

However, we do have a proper typed exception for this class of errors: `X::Backslash::UnrecognizedSequence`. So the more involved fix would be changing `<.sorry` to a `<.typed_panic` and giving it that exception to use as well as modifying that exception to take
suggestions as well.

Here's the diff of doing that for quote braid (`"\1"`). Needs to be done similar to this in two other places, (just search for "backslash sequence" to find them):


```diff
diff --git a/src/Perl6/Grammar.nqp b/src/Perl6/Grammar.nqp
[...]

PR #2117, opened 2018-07-22
Improve err for obs positionals w/ quantifiers

jmaslak:
This addresses issue #2110 .

Example: /\1+/ would previously give two messages, one about \1
being obs and "quantifier quantifies nothing".

Now the new behavior will just be to report the unrecognized backslash
sequence and offer a suggestion.

Thanks zoffixznet for determining what needed to be done and providing
some of the code.

Note that I could not figure out a way to test the backslash:misc token in "rule cc" - there's something strange going on their that is beyond my understanding for how to trigger that unrecognized backslash sequence token there, so it might be good for someone else to look at what I did there.

### DIFF
--- a/src/Perl6/Grammar.nqp
+++ b/src/Perl6/Grammar.nqp
@@ -5075,11 +5075,6 @@ if $*COMPILING_CORE_SETTING {
 }
 
 grammar Perl6::QGrammar is HLL::Grammar does STD {
-
-    method throw_unrecog_backslash_seq ($sequence) {
-        self.typed_sorry('X::Backslash::UnrecognizedSequence', :$sequence);
-    }
-
     proto token escape {*}
     proto token backslash {*}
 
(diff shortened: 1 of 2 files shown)


zoffixznet:
👍 Thanks!

The only minor thing is on [this line](https://github.com/rakudo/rakudo/pull/2117/files#diff-d3980092fdbde9a1dd4a30438f8dba19R5325) the `:sequence($<x>)` should be `:sequence(~$<x>)`. Because otherwise we're setting that attribute to an `NQPMatch` object instead of `Str`. I fixed in https://github.com/rakudo/rakudo/commit/03574546935501110e6913befbc38ec03faaf89c

> Note that I could not figure out a way to test the backslash:misc token in "rule cc" 

Me neither. Seems like the cases that would trigger the token override it and it doesn't get run. I filed that as https://github.com/rakudo/rakudo/issues/2119

PR #2141, opened 2018-07-28
Adds doc for option that apparently works

JJ:
Just use 

     perl6 -V

and it shows a bunch of variables and their values. This was also an option in perl5, which was apparently carried over.


lizmat:
Thank you!

PR #2142, opened 2018-07-28
Add Callable, Whatever, and zen indexing candidates for CArray

jstuder-gh:


lib/NativeCall.pm6
```diff
@@ -588,6 +588,21 @@ multi sub postcircumfix:<[ ]>(CArray:D \array, $pos) is export(:DEFAULT, :types)
...
     @pos.map: { array.AT-POS($_) };
 }
+multi sub postcircumfix:<[ ]>(CArray:D \array, Callable:D $block) is export(:DEFAULT, :types) {
+    nqp::stmts(
+      (my $*INDEX = 'Effective index'),
+      array[$block.pos(array)]
+    )
+}
+multi sub postcircumfix:<[ ]>(CArray:D \array) is export(:DEFAULT, :types) {
+    array[^array.elems]
```
zoffixznet: In regular lists, this one has [explict "just return `self`" meaning](https://github.com/rakudo/rakudo/blob/72c8cf68c/src/core/Any.pm6#L303-L310), which, for example, avoids reification.

```
<Zoffix_> m: my @a := <a b c>; dd @a =:= @a[]
<camelia> rakudo-moar 0ec4f5f28: OUTPUT: «Bool::True␤»
<Zoffix_> m: my @a := <a b c>; dd @a =:= @a[*]
<camelia> rakudo-moar 0ec4f5f28: OUTPUT: «Bool::False␤»
```

Does it make sense for `CArray` to simply return the `array` we get as-is without any indexing into it?

lizmat: Could it be that the slicing is necessary to create true HLL Perl 6 objects?

zoffixznet: Looking at candidates, seems [the candidate that calls `.ZEN-POS` is already called](https://github.com/rakudo/rakudo/blob/0ec4f5f283905f4e4ca79cac6a92ed12ec74310e/src/core/array_slice.pm6#L478-L480), but you don't right the right thing when you, say, interpolate it in a string:

```
<Zoffix_> m: use NativeCall; my @z  := CArray[int32].new( 1, 2, 3, 4, 5 ); say "foo @z[] bar"
<camelia> rakudo-moar 0ec4f5f28: OUTPUT: «foo  bar␤»
```

Makes me think we should be adding methods to [CArray class](https://github.com/rakudo/rakudo/blob/0ec4f5f283905f4e4ca79cac6a92ed12ec74310e/lib/NativeCall/Types.pm6#L92) rather than adding more postcircumfix operators, since even with the postcircumfix defined, calling `.ZEN-POS` directly would still give the wrong thing.

jstuder-gh:
Fixed the zen candidate and added a Str method on CArray so that it stringifies correctly.

zoffixznet:
> Fixed the zen candidate

But does `.ZEN-POS` work properly in CArrays?

If I call it this way, I get a type name in the string, but I was expecting the same behaviour as with normal lists:

```
<Zoffix_> m: use NativeCall; my @z  := CArray[int32].new( 1, 2, 3, 4, 5 ); say "foo @z.ZEN-POS() bar"
<camelia> rakudo-moar 0ec4f5f28: OUTPUT: «foo NativeCall::Types::CArray[int32]<62683472> bar␤»

<Zoffix_> m: my @z  := 1, 2, 3, 4, 5; say "foo @z.ZEN-POS() bar"
<camelia> rakudo-moar 0ec4f5f28: OUTPUT: «foo 1 2 3 4 5 bar␤»
```

jstuder-gh:
I added a ```Str``` method on ```CArray``` in the last commit that fixes that issue. Right now it is using the candidate from one of the base classes (probably ```Any```?) 

This is the result I am getting running the modification on my local machine: 

    $ perl6-m -e 'use NativeCall; my @z := CArray[int32].new(1, 2, 3, 4, 5); say "foo @z.ZEN-POS() bar"'
    > foo 1 2 3 4 5 bar

zoffixznet:
👍 Thanks!

You should [consider signing the CLA with the Perl Foundation to get a commit bit to this repo](https://github.com/rakudo/rakudo/blob/master/docs/obtaining-a-commit-bit.pod#how-to-obtain-a-commit-bit-for-the-rakudo-repository)

Issue #2143, opened 2018-07-29
Can't .gist/.Str/.perl Distribution::Resource:U


zoffixznet:
***See [Newcomer Guide to Contributing to Core Perl 6](https://rakudo.party/post/Newcomer-Guide-to-Contributing-to-Core-Perl-6) for tips on fixing this Issue***

----------------

```
<Zoffix_> m: say Distribution::Resource.^mro
<camelia> rakudo-moar b7379a804: OUTPUT: «Cannot look up attributes in a Distribution::Resource type object␤  in block <unit> at <tmp> line 1␤␤»
```

Fix would simply involve [changing the methods to `multi`](https://github.com/rakudo/rakudo/blob/72c8cf68c/src/core/Distribution.pm6#L184-L192) and type-constraining them on `:D` invocant. e.g.:

```perl6
    multi method perl(::?CLASS:D: |c) {
        self.IO.perl(|c)
    }
```

ribbon-otter:
Where would an appropriate place to write the tests be? Searching for either "Distribution" or "%?RESOURCES" on Perl6/Roast doesn't seem to give any hints for tests should be placed.

zoffixznet:
> Distribution::Resource.Str

It doesn't error. That's a stringification *warning* that usually is displayed when you stringify classes ain't meant to be stringified.

PR #2145, opened 2018-07-29
Fix RT #2143. Make Distribution::Resource.^mro not error.

ribbon-otter:
This commit fixes RT #2143. Besides just making `say
Distribution::Resource.^mro` not error, `.WHAT`, `.gist,`, `.perl`,
and `.say` now also work on the type object.

Credit to zoffixznet for figuring out how to solve the problem,
as I just followed their instructions. See this irc log for more
context, http://colabti.org/irclogger/irclogger_log/perl6?date=2018-07-29#l84

### DIFF
--- a/src/core/Distribution.pm6
+++ b/src/core/Distribution.pm6
@@ -181,13 +181,13 @@ class Distribution::Resource {
     }
 
     # delegate appropriate IO::Path methods to the resource IO::Path object
-    method Str(|c) {
+    multi method Str(::?CLASS:D: |c) {
         self.IO.Str(|c)
     }
-    method gist(|c) {
+    multi method gist(::?CLASS:D: |c) {
         self.IO.gist(|c)
     }
-    method perl(|c) {
+    multi method perl(::?CLASS:D: |c) {
         self.IO.perl(|c)
     }
     method absolute(|c) {


zoffixznet:
👍 Thanks!

zoffixznet:
> This commit fixes RT #2143. 

BTW, "RT" refers to our old tracker on https://rt.perl.org/  , not to GitHub's tracker.

zoffixznet:
Currently, you'll be listed in release credits as "ribbon-otter", if you'd like to be listed under a different name, please add yourself to the CREDITS file: https://github.com/rakudo/rakudo/blob/master/CREDITS

Gnouc:
@zoffixznet Sorry for intercept you here, but I don't see my name there, even I'm in commit list https://github.com/rakudo/rakudo/commits?author=Gnouc

zoffixznet:
> I don't see my name there

The contributors are listed in release announcements for the release into which the commits went in. For example, in [2017.10 announcement](https://github.com/rakudo/rakudo/blob/master/docs/announce/2017.10.md), you're listed as `Cuong Manh Le`. 

The [contributors generating script](https://github.com/rakudo/rakudo/blob/master/tools/contributors.p6) uses names/emails from commits and references them against the CREDITS file and attributes using the name in the CREDITS file if one is present, falling back to just the name in the commit if the user is not listed in CREDITS.

You can add yourself to CREDITS file, if you like, so that your future commits get attributed using some pre-defined name, rather than what's in the commit itself.

Gnouc:
@zoffixznet Ah ok, I thought that file is auto generated. So nevermind, it's ok.

PR #2163, opened 2018-08-01
Change to non-native type

JJ:
Native types in roles are still affected by this bug: https://rt.perl.org/Public/Bug/Display.html?id=126120. It may be changed back when that's solved.
On the other hand, this will most likely have some performance impact on Bools, so I don't know. I love fast Bools.

### DIFF
--- a/src/core/Enumeration.pm6
+++ b/src/core/Enumeration.pm6
@@ -2,7 +2,7 @@
 my role Enumeration {
     has $.key;
     has $.value;
-    has int $!index;
+    has Int $!index;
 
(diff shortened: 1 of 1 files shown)


lizmat:
Judging by f925c648 , this will only affect `.succ` and `.pred` on enums, so this looks good to me.

lizmat:
Reverted: :-(

Stage parse      : P6opaque: invalid native bind to attribute '$!index' in type Order for kind int64
  at src/Perl6/World.nqp:3747  (blib/Perl6/World.moarvm:create_enum_value)

PR #2171, opened 2018-08-02
implement S26 =defn block properly, add tests

tbrowder:
+ adds two new subs to Pod.npq
+ comments out one old sub in Pod.npq
+ add grammar and actions to properly handle =defn2
+ rename pod defn class to Pod::Defn
+ add reference to new spec tests for S26 =defn block (tests are already in roast)

### DIFF
--- a/src/Perl6/Pod.nqp
+++ b/src/Perl6/Pod.nqp
@@ -81,6 +81,11 @@ class Perl6::Pod {
     # literal normal space (U+0020)
     my $SPACE := "\x[0020]";
 
+    # defn block types
+    my $defn-para   := 'paragraph';
+    my $defn-delim  := 'delimited';
+    my $defn-abbrev := 'abbreviated';
+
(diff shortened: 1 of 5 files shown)


src/Perl6/Pod.nqp
```diff
@@ -516,7 +648,21 @@ class Perl6::Pod {
...
+    }
+
+    our sub lines($S, :$no-chomp) {
+        # Takes a string and separates it into an array of lines at
+        # each newline.
+        #
+        # uses nqp::split(str $delimiter, str $string --> Mu)
+
+        # not usual:
+        return nqp::split("\n", $S) if $no-chomp;
```
ugexe: Does any consideration need to be given to `\r\n` line endings?

tbrowder: Nick, I don’t know, and didn’t think of that.

PR #2172, opened 2018-08-03
Fix the error of using  on

FCO:
the error:
```
$ perl6 -e 'say 42.HOW.isa(Metamodel::ClassHOW)'
Too few positionals passed; expected 3 arguments but got 2
  in block <unit> at -e line 1
```

after the change:
```
$ ./perl6 -e 'say 42.HOW.isa(Metamodel::ClassHOW)'
True
```


lizmat:
Thanks!

zoffixznet:
I had to revert this in https://github.com/rakudo/rakudo/commit/e0a27ce371 as it did not pass `make test`  (and presumably `make spectest` either)

jnthn:
Good this was reverted; I think the change is wrong (makes it far too easy to confuse object level and meta-level otherwise)

PR #2175, opened 2018-08-04
Speed up subbuf by using the new nqp::slice op...

MasterDuke17:
instead of manually looping with nqp::bindpos+nqp::atpos.

`my $a = buf8.new(^1_000); my $b; for ^1_000_000 { $b = $a.subbuf(200, 500); Nil };` used to take ~17s, now it takes ~2s.


lizmat:
Way cool!

PR #2189, opened 2018-08-07
teach "assign" spesh plugin about Proxy

timo:
requires https://github.com/perl6/nqp/pull/495
also requires MoarVM/MoarVM#932

With this simple test code it gives only a 6.3% speedup:

    time perl6-m -e 'sub proxies is rw { Proxy.new( FETCH => -> | { say "fetched" }, STORE => -> | { }) }; for ^5_000_000 { my $prox := proxies(); $prox = 5 }'


lizmat:
I think has bitrotted beyond repair, especially now that new-disp has landed.  So closing.

Issue #2190, opened 2018-08-07
Substitute Rakudo::Internals::is-win to $*DISTRO.is-win for easier (or possible) testing of code when possible.


JJ:
## The Problem

Since it's impossible to change the value of  Rakudo::Internals::is-win, it's impossible to test that code unless you are in an actual machine that causes that value to be true.

## Expected Behavior

It would be ideal if we could test all code, or most of it, in a single machine. In all cases where there is no actual interaction with the OS, it should be possible to check from a single machine the behavior of Win and non-Win OSes.

## Actual Behavior

In most cases Windowness is detected using `$*DISTRO.is-win`, which can be modified in runtime to test that particular code, but in 5 cases (so far) it uses that other method to detect Windowness. Maybe that's also easy to change through the lexpad, but anyway it would be better to detect Windowness through a single method throughout all the code for testing purposes, if that's not a significant impact in overall performance.

ugexe:
Testing is only possible because it is the way it is. If it was changed as suggested then rakudo would barely function, since `BEGIN $*DISTRO does role :: { method is-win { True } }; say $*DISTRO.is-win` would then affect internal stuff like path handling, choosing HOME/TMP, etc that has nothing to do with an external test.

JJ:
$*DISTRO is used all over the place to check what OS Rakudo is on,
including some kernel functions:
https://github.com/rakudo/rakudo/blob/c71fa17d8da36b8cfcbd85584cee32566f25f059/src/core/Kernel.pm6
That's bound to be affected if you do stuff like that, but it also allows
to test that code even if you're not using Windows. Using
Rakudo::Internals::IS-WIN is, in fact, relatively rare.

raiph:
I think the following is sound logic and guidance. If not, I apologize.
> It would be ideal if we could test all code, or most of it, in a single machine.
Obviously not _all_ code, because some of it is OS specific, as you point out.
> In all cases where there is no actual interaction with the OS, it should be possible to check from a single machine the behavior of Win and non-Win OSes.
In academic theory, sure. But that implies indirections for everything, and indirections slow things down, so it's very likely there's code that will stymie some such testing in the name of performance or simplicity or both. Given that there's always going to be _some_ code that requires running on the actual OS, and that attaining the academic theoretical "ideal" has both performance and code complexity costs, I think you'll find you'll end up needing to very carefully justify each change you propose to core code to enable/ease increasing text coverage using a single OS.

JJ:
[...]

PR #2192, opened 2018-08-08
Uses DISTRO when available, refs #2190

JJ:


zoffixznet:
Looking at the [list of other occurances of `R::I`](https://github.com/rakudo/rakudo/issues/2190#issuecomment-411311721) that would still remain, is there a point in changing this one?

The new version is 11x slower (which, granted, in this case doesn't matter much) and adds shelling out to `lsb_release`/`uname` to fetch info we're not interested in.

If the original Issue remains unresolved, is there a point in increasing complexity here?

JJ:
First, thanks for taking the time to consider this. I'll try to address your concerns.
@lizmat I'm changing an undocumented function by a stable and documented one. See also #1941 , which was rejected, so no waiting until it's documented. #1941 also makes one just guess what the two functions are all about (or at least the one which is not documented). In absence of documentation, my hunch is that I should use the documented function when available, such as in this case. 
@zoffixznet see above for your first question. In other cases where R::I still remains, `$*DISTRO` has not been registered yet so it's unavailable. 
WRT speed, you are probable comparing `$*DISTRO.is-win` witn `R::I::IS-WIN`. That's another error that I intend to address, see #2190. Although it would make a priori sense to have two different sources for truth for Windowsness (I don't know, in a cross-compiling environment, for instance), it probably makes much more sense to use the same code for both. I was expecting this issue to be accepted (or rejected) to work on that. In this particular case, the code is being called only once when the class is created. I don't think that's going to have any significant impact, even if the call itself is 11x slower.

niner:
I still don't understand what problem this PR is supposed to fix. The change is to rakudo's core code. That's exactly the place that Rakudo::Internals is there for. IOW I don't see an issue with using Rakudo::Internals in rakudo's internals.

I've worked really hard over the years to make module loading as fast as possible. I would rather not give up any of the performance advantage when our user's don't gain anything in return. It's hard enough as it is to make this code faster while keeping it somewhat readable.

JJ:
[`$*DISTRO.is-win` is also used in the core](https://github.com/rakudo/rakudo/blob/7c08ba7443af98cc5def12780379cf076cbb6337/src/core/REPL.pm6#L47)
Are you implying that `$*DISTRO` should be used only in user-facing code such as tests?
If this PR is intended to fix something is the absence of a guideline of when to use `$*DISTRO` or `R::I` within Rakudo. My hunch by looking at the code (which, as per #1941, is the only documentation available) was that `$*DISTRO` should be used if available. I'm OK with "*never* use `$*DISTRO` in the core", as above.

ugexe:
No one is implying "$*DISTRO should be used only in user-facing code", or saying "never use $*DISTRO in the core". Instead lets flip this the other way -- are you stating that all rakudo code must be refactored such that there is no duplication despite any benefits the duplication may bring?

It seems like you are trying to codify how rakudo's internal code should be written, but I posit that it is more beneficial to give developers with specialized knowledge in specific areas the freedom to make an educated choice with a given context -- after all we don't state "you must write hot loops using nqp in core", "rakudo should be written in as much perl6 as possible despite performance", or similar.

JJ:
This probably should be closed.

PR #2212, opened 2018-08-14
Normalize the path portion of repo specs

ugexe:
repo-specs are used as lookup keys, so the path slashes need to be normalized. For instance -- on windows when trying to `use Linenoise;`  it turns out `CompUnit::RepositoryRegistry.repository-for-name("site")` is `Nil` inside of `Distribution::Resource`.

This is a redo of https://github.com/rakudo/rakudo/commit/c3e1ec0bb86fad16ecca6ca5997d81cd6afc2ba5

### DIFF
--- a/src/core/CompUnit/RepositoryRegistry.pm6
+++ b/src/core/CompUnit/RepositoryRegistry.pm6
@@ -49,6 +49,7 @@ class CompUnit::RepositoryRegistry {
         my $raw-specs;
         # only look up environment once
         my $ENV := nqp::getattr(%*ENV,Map,'$!storage');
+        my $sep := $*SPEC.dir-sep;
 
         # starting up for creating precomp
         my $precomp-specs = nqp::existskey($ENV,'RAKUDO_PRECOMP_WITH')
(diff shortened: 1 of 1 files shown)


ugexe:
@niner @hoelzro

ugexe:
This gets Linenoise a little further in windows, but alas it looks like it might not be happy with the dll filename being changed.

```
C:\Users\ugexe\.zef\store\p6-linenoise.git\480fd919b2c082e691c518fd50c19ad8719532b6>perl6
I ran into a problem while trying to set up Linenoise: Cannot locate symbol 'linenoiseSetCompletionCallback' in native library 'C:\Users\ugexe\.rakudobrew\moar-blead-master\install\share\perl6\site\resources\323767E89F2D0CBE48797015570799E56F6F4FE5.dll'
Continuing without tab completions or line editor
You may want to consider using rlwrap for simple line editor functionality

To exit type 'exit' or '^D'
>
```

hoelzro:
Thanks for taking that on, @ugexe!  I'll see if I can push Linenoise past the finish line.

hoelzro:
@ugexe I just installed Rakudo, zef, and Linenoise all from `HEAD` on a fresh VM: Linenoise works without a problem!  Could it be that you have an old DLL or something hanging around?

ugexe:
@hoelzro It's not an old dll... I've always had this problem with Linenoise ( it was I was referring to here https://github.com/hoelzro/p6-linenoise/commit/d0a2ec0ade6d9a4ecc1fd7d6b81ee305a5f9a3f1 ). Note that `perl6 -MLinenoise -e 1` works, its just when I do `perl6`

hoelzro:
@ugexe Interesting...plain `perl6` is what I was testing with.  What does your Windows environment look like?

JJ:
So I guess we still need the OpenSSL hack, right?

Issue #2094, opened 2018-07-18
"decimal point must be followed..." error is too eager


zoffixznet:
It should probably check we're not gonna call the prefix-as-posfix before giving up entirely and throwing:

```
19:54 | Zoffix | m: say 42.:<->
-- | -- | --
19:54 | camelia | rakudo-moar a847abe52: OUTPUT: «=== SORRY!===  Error while compiling <tmp>␤Decimal point must be followed by digit␤at <tmp>:1␤------> say 42. ⏏ :<-> ␤»

19:54 | Zoffix | m: say (42).:<->
19:54 | camelia | rakudo-moar a847abe52: OUTPUT: «-42␤»
```

PR #2225, opened 2018-08-18
Allow prefix calls with .:< ... > on integer literals

ribbon-otter:
Fix GitHub #2094 . Before this commit, `42.:<->` would error because the check which prevented integers ending with fullstops was overly aggressive: any time an integer was followed by `.:`, an error would be
generated. With this commit, an integer followed by `.:` will only error if the next thing after it is not `<` then a prefix operator then `>`

### DIFF
--- a/src/Perl6/Grammar.nqp
+++ b/src/Perl6/Grammar.nqp
@@ -3688,7 +3688,7 @@ grammar Perl6::Grammar is HLL::Grammar does STD {
             ]
         | <VALUE=decint>
         ]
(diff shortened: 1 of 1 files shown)


AlexDaniel:
Found more cases, here's the list of all I know:
* ```say 42.:<->```
* ```say 42.:<<->>```
* ```say 42.:«-»```
* ```say 42.:['-']```

ribbon-otter:
I commited an update with does the changes which you requested. I was unsure if it is correct policy to squash my commit into my already existing commit or not considering that I had already made a pull request off of it. I am willing to squash the commit together if that is preferred.

niner:
Looks good to me! Personally I prefer squashed commits.

PR #2232, opened 2018-08-20
Fix Regex.ACCEPTS: NDF does not set $/ #2176

### LINKED ISSUE #2176: Regex.ACCEPTS: NDF does not set $/
***See [Newcomer Guide to Contributing to Core Perl 6](https://rakudo.party/post/Newcomer-Guide-to-Contributing-to-Core-Perl-6) for tips on fixing this Issue***

----------------

There [is a fudge in the propspec](https://github.com/perl6/roast/commit/b8bae2967) for this bug, but it doesn't look like it's been filed as a bug, only [for the match success itself](https://rt.perl.org/Ticket/Display.html?id=130458), which has been resolved.

The bug:
When you match with NFD, the `$/` variable is not set (the fudge itself tests `$/.orig` is of type NFD)

```
<Zoffix_> m: say so "7\x[308]".NFD ~~ /^ \d+ $/; dd $/.orig
<camelia> rakudo-moar c4bb1b19d: OUTPUT: «True␤Nil␤»
```

I suspect the fix involves just sticking the following into the [`.ACCEPTS: NDF`  candidate](https://github.com/rakudo/rakudo/blob/c4bb1b19d/src/core/Regex.pm6#L37) before calling the `Str` candidate:

```
$/ := nqp::getlexcaller('$/');
```

ryn1x:
Fix #2176

Behavior before:
```
> say so "7\x[308]".NFD ~~ /^ \d+ $/; dd $/.orig;
True
Nil
```

Behavior after:
```
> say so "7\x[308]".NFD ~~ /^ \d+ $/; dd $/.orig;
True
"7̈"
```

@zoffixznet 's suggestion ```$/ := nqp::getlexcaller('$/');``` did not work, but using ```$/ := nqp::getlexrelcaller(nqp::ctxcallerskipthunks(nqp::ctx()),'$/');``` as found here: https://github.com/rakudo/rakudo/blob/c4bb1b19d/src/core/Regex.pm6#L109 is what seems to fix the issue.

This does not fix https://github.com/perl6/roast/commit/b8bae2967#diff-c90019d7badd770a0d9abeea8d59795aR91 as ```$/.WHAT``` is a ```Str``` and not ```NFD```.

### DIFF
--- a/src/core/Regex.pm6
+++ b/src/core/Regex.pm6
@@ -34,7 +34,8 @@ my class Regex { # declared in BOOTSTRAP
     }
 
 #?if !jvm
-    multi method ACCEPTS(Regex:D \SELF: Uni:D \uni) {  # RT #130458
+    multi method ACCEPTS(Regex:D \SELF: Uni:D \uni) {
+        $/ := nqp::getlexrelcaller(nqp::ctxcallerskipthunks(nqp::ctx()),'$/');
         self.ACCEPTS(uni.Str)
     }
 #?endif


zoffixznet:
:+1: Thanks!

PR #2237, opened 2018-08-23
Fix precision calculation in FatRat .Str

thundergnat:
Added as part of a prior pull request; #1822. In retrospect, and after review and commentary by Zoffix++ (See comments in rejected pull request #1840) I realized this was a bad idea: gives relatively useless precision, and slows down stringification for everyone. Spectested ok with this change.

### DIFF
--- a/src/core/Rational.pm6
+++ b/src/core/Rational.pm6
@@ -129,7 +129,7 @@ my role Rational[::NuT = Int, ::DeT = ::("NuT")] does Real {
                     $fract *= 1000000;
                 }
                 else {
-                    $precision = nqp::chars($!denominator.Str) + nqp::chars($whole.Str) + 1;
+                    $precision = nqp::chars($!denominator.Str) + 1;
                     $fract *= nqp::pow_I(10, nqp::decont($precision), Num, Int);
                 }
             }
--- a/src/core/Rational.pm6
+++ b/src/core/Rational.pm6
@@ -102,37 +102,21 @@ my role Rational[::NuT = Int, ::DeT = ::("NuT")] does Real {
 
         if $fract {
             my $precision;
-            # Stringify Rats to at least 6 significant digits. There does not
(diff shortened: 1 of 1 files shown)


src/core/Rational.pm6
```diff
@@ -129,7 +129,7 @@ my role Rational[::NuT = Int, ::DeT = ::("NuT")] does Real {
                     $fract *= 1000000;
                 }
                 else {
-                    $precision = nqp::chars($!denominator.Str) + nqp::chars($whole.Str) + 1;
+                    $precision = nqp::chars($!denominator.Str) + 1;
                     $fract *= nqp::pow_I(10, nqp::decont($precision), Num, Int);
                 }
             }
```
zoffixznet: What's the difference between this conditional branch and the truthy branch? The code in both looks the same.

thundergnat: After this pull request, they would be the same. There was some speculative code stringifying FatRats to a higher precision, so there are two code paths. If the higher precision FatRat stringification isn't going in core, this could be collapsed to a single code path. (and the longer I think about it, the more I lean that direction.)

I've been puttering with a module to do high precision stringification. It's slow, but only people who specifically want it have to pay the cost.

zoffixznet: > speculative code

It could just be a text comment for future programmers without any code and we'd be able to get rid of duplicate code and the now-useless typecheck in the conditional there as well.

zoffixznet:
:+1: looks OK to me though I've not thought about the impact of this change, so I'll leave it to someone else to merge.

lizmat:
Is this something that needs to be salvaged?

JJ:
Well, after conflicts are fixed, maybe...

lizmat:
Feels to me this PR has bitrotted, yet could be easily made anew for anybody willing to put in the effort.  So closing now.

PR #2240, opened 2018-08-28
Make native arrays throw when storing lazy Seq

jstuder-gh:
Currently a Failure is returned when attempting to store a lazy Seq but
since the result is not sunk, the Failure is not thrown. This leads to
odd results like below:

    my int @geo = 2, 4, 8 ... *;
    say @geo[2..10];

    # OUTPUT:   [0 0 0 0 0 0 0 0 0]
    # EXPECTED: FAILED

Resolve by explicitly throwing.

### DIFF
--- a/src/core/native_array.pm6
+++ b/src/core/native_array.pm6
@@ -89,7 +89,7 @@ my class array does Iterable {
 
     my role strarray[::T] does Positional[T] is array_type(T) {
 #- start of generated part of strarray role -----------------------------------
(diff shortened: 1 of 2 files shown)


lizmat:
If we throw, there is no point making the `Failure` object.  Just throw directly on the `Exception` object.

jstuder-gh:
@lizmat, I've updated the branch to throw the ```Exception``` directly, as you suggested.

I also added another commit that throws when attempting to ```splice``` a lazy ```Seq``` into the native array (that I didn't see that before).

PR #2277, opened 2018-09-15
Partially implement Pod '#' alias for %config :numbered

tbrowder:
Implemented only for abbreviated blocks, e.g.,

    =para # some
    text

or

    =para
    # some
    text

+ Added TODOs and notes for upcoming implementations.

+ Some code cleanup.

Files changed:

+ src/Perl6/Actions.nqp

+ src/Perl6/Grammar.nqp

+ src/Perl6/Pod.nqp

+ t/spectest.data

    - added tests in t/spec/S26*/15-numbered-alias.t

### DIFF
--- a/src/Perl6/Actions.nqp
+++ b/src/Perl6/Actions.nqp
@@ -7,6 +7,11 @@ use QAST;
 
 my $wantwant := Mu;
 
+# block types
+my $para-block   := 'paragraph';
+my $delim-block  := 'delimited';
+my $abbrev-block := 'abbreviated';
+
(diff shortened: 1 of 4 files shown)


tbrowder:
After closer reading of S26 I believe that only head, item, and defn blocks are to use the ‘#’ shorthand for :numbered. Thus I will rectify that in my next PR which will implement the ‘#’ for abbreviated, paragraph, and delimited blocks, but ONLY for head, list, and defn blocks. All other block types can still use the explicit :numbered %config per S26.

PR #2291, opened 2018-09-19
Check for definedness of callframe.code when creating error message

### LINKED ISSUE #2222: LTA error message with non-dwimmy hyper infix
## The Problem
I get a confusing error message when using the non-dwimmy hyper infix:
```
> 5 »*» (2..4)
No such method 'name' for invocant of type 'Mu'. Did you mean any of these?
    note
    take
```

## Expected Behavior
Exchanging the operands but keeping it non-dwimmy gives a helpful message:
```
> (2..4) «*« 5
Lists on either side of non-dwimmy hyperop of infix:<*> are not of the same length
left: 3 elements, right: 1 elements
```

## Environment
Tested on camelia `rakudo-moar 34468f11d` (HEAD)

taboege:
This fixes #2222. The "non-dwimmy hyperop" hint uses callframe(…).code.name
to detect if it recurses. In some cases callframe.code is Mu (unknown file)
which leads to an error looking up .name. Inserting `andthen` solves this.

-----

These three appear to be all such uses of callframe.

### DIFF
--- a/src/core/metaops.pm6
+++ b/src/core/metaops.pm6
@@ -566,7 +566,7 @@ multi sub HYPER(&operator, Positional:D \left, \right, :$dwim-left, :$dwim-right
     my @result;
     X::HyperOp::Infinite.new(:side<left>, :&operator).throw if left.is-lazy;
     my int $elems = left.elems;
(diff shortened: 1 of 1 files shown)


taboege:
Yeah, good catch.

zoffixznet:
:+1: Thanks!

PR #2330, opened 2018-09-28
Add a BOM on open for utf16 filehandles

samcv:
When we open a file for write we will write a BOM upon open. Since we
already truncate the file if it exists, this shouldn't cause any loss of
data aside from the previous functionality.

If we are appending, then only add a BOM if there is nothing yet written
to the file. If the file already has data in it, then don't add a byte
order mark.

### DIFF
--- a/src/core/IO/Handle.pm6
+++ b/src/core/IO/Handle.pm6
@@ -177,8 +177,17 @@ my class IO::Handle {
             $!decoder.set-line-separators(($!nl-in = $nl-in).list);
             $!encoder := $encoding.encoder(:translate-nl);
             $!encoding = $encoding.name;
+
+            # Add a byte order mark to the start of the file for utf16
+            nqp::if(nqp::iseq_s($!encoding, 'utf16'), (
+                if $create && !$exclusive && (!$append || $append && $!path.s == 0) {
+                  self.write: Buf[uint16].new(0xFEFF);
+                })
+            );
         }
         self!set-out-buffer-size($out-buffer);
+
+
         self;
     }
 


src/core/IO/Handle.pm6
```diff
@@ -177,8 +177,17 @@ my class IO::Handle {
             $!decoder.set-line-separators(($!nl-in = $nl-in).list);
             $!encoder := $encoding.encoder(:translate-nl);
             $!encoding = $encoding.name;
+            # Add a byte order mark to the start of the file for utf16
+            nqp::if(nqp::iseq_s($!encoding, 'utf16'), (
+                if $create && !$exclusive && (!$append || $append && $!path.s == 0) {
+                  state $bom = buf16.new(0xFEFF);
```
zoffixznet: Is it better if this were a constant instead? `my constant $bom = buf16.new(0xFEFF);` ?

samcv: I was mostly trying to make sure if you opened many files as utf16 it wouldn't have to create a new buf16 each time. If `my constant` will do that then that would be acceptable for me.

zoffixznet: > If my constant will do that then that would be acceptable for me.

Yeah, it'd create one at compile time and then re-use it for all the writes.

lizmat:
Looks ok to me.

src/core/IO/Handle.pm6
lizmat: Won't this get the `utf16` role get punned at compile time, and then create all sorts of extra startup overhead?

src/core/IO/Handle.pm6
```diff
@@ -177,8 +177,17 @@ my class IO::Handle {
             $!decoder.set-line-separators(($!nl-in = $nl-in).list);
             $!encoder := $encoding.encoder(:translate-nl);
             $!encoding = $encoding.name;
+            # Add a byte order mark to the start of the file for utf16
+            nqp::if(nqp::iseq_s($!encoding, 'utf16'), (
+                if $create && !$exclusive && (!$append || $append && $!path.s == 0) {
```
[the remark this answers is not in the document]
samcv: Yeah. I realized there could be a data race here, but the alternative of not putting BOM as the standard requires would be less desirable.

src/core/IO/Handle.pm6
samcv: @lizmat it only uses buf16 and not the utf16 role, though do you mean that would cause it to have to load anyway? I am thinking about using `blob16` instead of `buf16` since this is a constant and not designed to be modified.

lizmat: if it works, then it's ok with me.  I think we've abandoned reducing startup time at this point anyway.  I would use `buf16` as that is more likely to be used by users, so it would have already been punned for them.  See also src/core/core_epilogue about `buf8`

jnthn: @lizmat Deserialization works lazily, so in this case we won't trigger it unless the constant is used.

src/core/IO/Handle.pm6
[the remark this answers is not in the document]
jnthn: Yeah, I'm not sure there's a perfect option here; I'm fine with this one.

samcv:
It turns out setting a constant did not work and I get issues.

http://colabti.org/irclogger/irclogger_log/perl6-dev?date=2018-09-29#l360

```

19:35 | samcv | when i try and make my BOM a consntant then compilation fails
-- | -- | --
19:35 | samcv | "An exception occurred while evaluating a constant"
19:37 |   | masak .oO( somebody set up us the constant )
19:38 |   | emschwartz left
19:55 |   | bla8 joined
19:56 |   | p6bannerbot set mode: +v bla8
19:57 |   | bla8 left
20:12 | diakopter | samcv: sounds like a bug in the serializer/deserializer

```
Going to merge this using `Buf[uint16]` to keep it as simple as possible and so it won't have to deal with the utf16 type at all.

lizmat:
Yeah, trying to use classes / roles that early in the setting is fraught with dragons :-(

PR #2341, opened 2018-10-04
Perl6-architecture

ccworld1000:
Perl6-architecture

### DIFF
--- a/docs/Perl6-architecture.md
+++ b/docs/Perl6-architecture.md
@@ -1,0 +1,1 @@
+![Perl 6 architecture ](https://raw.github.com/ccworld1000/CCArchitecture/master/6.svg?sanitize=true)


zoffixznet:
Don't know if there was a discussion elsewhere about this, but what is this? Also, should the linked SVG be part of rakudo's repo (and link changed) rather than a third-party repo being used?

ugexe:
Looks like something that would need to be generated automatically with a git hook, similar to how p6doc has to generate the type graphs.

AlexDaniel:
As for diagramming, ping @JJ :)

JJ:
I just adjusted a bit the current diagram, with your help... I can help some if needed...

ccworld1000:
Q:"These are not classes, so I don't think they should be on the diagram:"
A:Sorry, note information is displayed, I will deal with it right away.

@AlexDaniel

ccworld1000:
that's OK!
[https://github.com/ccworld1000/CCArchitecture](https://github.com/ccworld1000/CCArchitecture)

AlexDaniel:
@ccworld1000 what about code?

PR #2342, opened 2018-10-05
Merge in the js branch

pmurias:
This will added support for the JS backend to master


lizmat:
A merge with this appears to make the CORE.setting.moarvm about 54K smaller.  I would have expected the MoarVM build to come out the same size.  Am I wrong?

lizmat:
Why does Test.pm need a `use MONKEY-SEE-NO-EVAL` ? while there are no further changes to that file?

lizmat:
What is the point of `PhasersList` as opposed to `IterationBuffer` ?

lizmat:
Why the changes in `Date` changing `int` -> `Int` ?

lizmat:
In `ShapedArray` all backends (including JVM) now use `nqp::multidimref_n($!from,$!indices)` .  Does that mean that also works on the JVM backend now??

lizmat:
In method `Str.ords` anything not on Moar now checks for CRLF specifically.  Is that correct?

lizmat:
in `Str.codes` now the JVM backend also uses `nqp::codes`, is that correct?

lizmat:
This concludes my review and my questions so far.

pmurias:
> In `ShapedArray` all backends (including JVM) now use `nqp::multidimref_n($!from,$!indices)` . Does that mean that also works on the JVM backend now?? Same for `native_array` and `tools/build/makeNATIVE_SHAPED_ARRAY`

Yes, I implemented nqp::multidimref_* ops on the JVM backends as the ifdefs where annoying

pmurias:
> What is the point of `PhasersList` as opposed to `IterationBuffer` ?

IterationBuffer is not composed at that point, it works on MoarVM somehow but I don't think nqp::create on uncomposed things is correct

pmurias:
> in `Str.codes` now the JVM backend also uses `nqp::codes`, is that correct?

I implemented nqp::codes on JVM

pmurias:
> Why does Test.pm need a `use MONKEY-SEE-NO-EVAL` ? while there are no further changes to that file?

Doesn't seem needed anymore. I got rid of it.

src/Perl6/Grammar.nqp
```diff
@@ -3739,6 +3743,7 @@ grammar Perl6::Grammar is HLL::Grammar does STD {
 
     token rad_number {
         ':' $<radix> = [\d+] <.unsp>?
+        {} # workaround bug
```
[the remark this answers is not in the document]
pmurias: I'll check this

lizmat:
pmurias++

src/vm/js/make-release.js
```diff
@@ -0,0 +1,66 @@
...
+  process.exit();
+}
+
+try {
+  fs.mkdirSync(releaseDir);
+} catch (e) {
+  if (e.code !== 'EEXIST') throw e;
+}
+
+const precompiledPerl6 = '/home/pmurias/rakudo/node_modules';
```
ugexe: There are a couple instances of hard coded absolute paths in this file with your home path

ugexe: There are a couple instances of hard coded absolute paths in this file with your home path

src/vm/js/perl6-runtime/package.json
```diff
@@ -0,0 +1,18 @@
...
+  "bugs": {
+    "email": "pawelmurias@gmail.com"
+  },
+  "licenses": [
+    {
+      "type": "Artistic 2",
+      "url": "http://opensource.org/licenses/Artistic-2.0"
+    }
+  ],
+  "repository": "http://github.com/rakudo/rakudo/",
```
ugexe: nitpick: this should be `https`

src/core/Iterable.pm6
```diff
@@ -110,19 +110,29 @@ my role Iterable {
...
     method race(Int(Cool) :$batch = 64, Int(Cool) :$degree = 4) {
+#?if !js
         RaceSeq.new:
           configuration =>
             HyperConfiguration.new(:$degree, :$batch, :method<race>),
           work-stage-head =>
             Rakudo::Internals::HyperIteratorBatcher.new(:$.iterator)
+#?endif
+#?if js
+        self
```
ugexe: This is probably technically ok since the return type isn't constrained, but I feel like it should be ( and thus the `#!js` logic would end up in `RaceSeq`/`HyperSeq` ) due to stuff like `isa-ok (^1000).hyper.map(*+1).race, RaceSeq, 'Can switch from hyper to race mode';` ( from the roast ).

src/vm/js/make-release.js
pmurias: The whole way rakudo.js npm releases are done currently sucks, and needs to be thought over and replaced :( I'll replace the paths with command line arguments tommorow

src/Perl6/Metamodel/BOOTSTRAP.nqp
```diff
@@ -2854,6 +2850,9 @@ BEGIN {
             sub add_to_cache($entry) {
 #?if jvm
                 return 0 if nqp::capturehasnameds($capture);
+#?endif
+#?if js
```
MasterDuke17: Are you intentionally adding an `#?if js` for some reason, or could the `#?if jvm` just be turned into `#?if !moar`?

src/core/JavaScriptScheduler.pm6
```diff
@@ -0,0 +1,53 @@
...
+        die "Cannot specify :every in {self.^name}"
+          if $every;
+
+        die "Cannot specify :at in {self.^name}"
+          if $at;
+        die "Cannot specify :in in {self.^name}"
+          if $in;
+
+        my $delay = $at ?? $at - now !! $in;
+        #sleep $delay if $delay;
```
MasterDuke17: Should this commented out line just be removed?

src/core/Lock/Async.pm6
```diff
@@ -79,9 +80,14 @@ my class Lock::Async {
                 }
             }
         }
+#?endif
+#?if js
+                    return KEPT-PROMISE;
```
MasterDuke17: This may be nit-picking, but I'd just indent this one level, not the same as the other `return`s.

src/vm/js/perl6-runtime/runtime.js
```diff
@@ -0,0 +1,361 @@
...
+
+  op.p6capturelex = function(ctx, codeObj) {
+    const closure = codeObj.$$getattr(Code, "$!do");
+    const wantedStaticInfo = closure.staticCode.outerCodeRef;
+
+    if (ctx.codeRef().staticCode === wantedStaticInfo) {
+      closure.outerCtx = ctx;
+      closure.capture(closure.staticCode.freshBlock());
+    } else {
+      /* HACK - workaround for rakudo bugs */
```
MasterDuke17: Is this still needed?

src/vm/js/perl6-runtime/runtime.js
```diff
@@ -0,0 +1,361 @@
...
+
+    while (find) {
+        if (find.codeRef().staticCode === wantedStaticInfo) {
+          closure.outerCtx = find;
+          closure.capture(closure.staticCode.freshBlock());
+          return codeObj;
+        }
+        find = find.$$caller;
+    }
+  //  console.log("HORRIBLE hack - p6capturelexwhere will do nothing");
```
MasterDuke17: And this?

src/core/Lock/Async.pm6
pmurias: Done

src/vm/js/make-release.js
pmurias: Stopped hard coding paths

src/vm/js/perl6-runtime/runtime.js
pmurias: I think so, rakudo still sometimes emits broken closures

src/Perl6/Metamodel/BOOTSTRAP.nqp
pmurias: I should just update the caching on the js backend and get rid of that workaround soon.

src/core/Iterable.pm6
pmurias: Done. I'm returning RaceSeq/HyperSeq which are Seq subclasses on the js backend now.

src/Perl6/Metamodel/BOOTSTRAP.nqp
pmurias: Merged the ?ifs in the meantime

niner:
Massive congratulations to pmurias++!

PR #2350, opened 2018-10-08
Use the correct ContainerDescriptor for Shaped3Array

SyrupThinker:
The previous version fails to create Arrays with a dimension of three.
EDITED `my @a[1;1;1]; say @a; # Fails`


AlexDaniel:
Thank you!

PR #2363, opened 2018-10-12
throws-like perform exact match for non-strings

rbt:
Matching boolean and other exception fields which change form during stringification is tricky without this fallthrough to an exact type match.

Without this change tests like the following fail with:
    # Failed test '.value-false matches False'
    # at SETTING::src/core/Any-iterable-methods.pm6 line 634
    # Expected: Bool::False
    # Got:      False

use Test;
plan 4;
class X::Test is Exception {
    has Bool $.value-false = False;
}

throws-like {
    die X::Test.new(:message('Thursday'));
}, X::Test, 'Exception match test',
  value-false => False;

### DIFF
--- a/lib/Test.pm6
+++ b/lib/Test.pm6
@@ -611,11 +611,21 @@ sub throws-like($code, $ex_type, $reason?, *%matcher) is export {
                 if $type_ok {
                     for %matcher.kv -> $k, $v {
                         my $got is default(Nil) = $_."$k"();
-                        my $ok = $got ~~ $v,;
-                        ok $ok, ".$k matches $v.gist()";
-                        unless $ok {
-                            _diag "Expected: " ~ ($v ~~ Str ?? $v !! $v.perl)
-                              ~ "\nGot:      $got";
+                        # Match strings and Regex as regex, everything else gets a direct comparison.
+                        if (nqp::eqaddr($v.WHAT, Str) or nqp::eqaddr($v.WHAT, Regex)) {
+                            my $ok = $got ~~ $v,;
+                            ok $ok, ".$k matches $v.gist()";
+                            unless $ok {
+                                _diag "Expected: " ~ ($v ~~ Str ?? $v !! $v.perl)
+                                  ~ "\nGot:      $got";
+                            }
+                        } else {
+                            my $test = nqp::eqaddr($v.WHAT, Mu)
+                                ?? nqp::eqaddr($got.WHAT, Mu)
+                                    !! nqp::eqaddr($got.WHAT, Mu)
+                                        ?? False
+                                        !! $got eq $v;
+                            my $ok = proclaim($test, ".$k matches $v.gist()");
                         }
                     }
                 } else {


lib/Test.pm6
```diff
@@ -611,11 +611,21 @@ sub throws-like($code, $ex_type, $reason?, *%matcher) is export {
                 if $type_ok {
                     for %matcher.kv -> $k, $v {
                         my $got is default(Nil) = $_."$k"();
-                        my $ok = $got ~~ $v,;
-                        ok $ok, ".$k matches $v.gist()";
-                        unless $ok {
-                            _diag "Expected: " ~ ($v ~~ Str ?? $v !! $v.perl)
-                              ~ "\nGot:      $got";
+                        # Match strings and Regex as regex, everything else gets a direct comparison.
+                        if (nqp::eqaddr($v.WHAT, Str) or nqp::eqaddr($v.WHAT, Regex)) {
```
MasterDuke17: These are only style comments, not about the change in logic.

Core code doesn't usually use parens about the conditional of an `if`.
Unless performance is a concern, this might be simpler as `$v ~~ Str|Regex`;

lib/Test.pm6
```diff
@@ -611,11 +611,21 @@ sub throws-like($code, $ex_type, $reason?, *%matcher) is export {
...
+                            my $ok = $got ~~ $v,;
```
MasterDuke17: The trailing comma was in the original code, but looks like it was just a typo there, maybe get rid of it while you're making changes?

lib/Test.pm6
```diff
@@ -611,11 +611,21 @@ sub throws-like($code, $ex_type, $reason?, *%matcher) is export {
...
+                                _diag "Expected: " ~ ($v ~~ Str ?? $v !! $v.perl)
+                                  ~ "\nGot:      $got";
+                            }
+                        } else {
+                            my $test = nqp::eqaddr($v.WHAT, Mu)
+                                ?? nqp::eqaddr($got.WHAT, Mu)
+                                    !! nqp::eqaddr($got.WHAT, Mu)
+                                        ?? False
+                                        !! $got eq $v;
+                            my $ok = proclaim($test, ".$k matches $v.gist()");
```
MasterDuke17: It looks like you aren't checking the result here? Is something like https://github.com/rakudo/rakudo/blame/master/lib/Test.pm6#L154-L161 missing?

zoffixznet:
Thanks for the PR, but I don't believe these changes to this routine are warranted. (Also, the changes likely don't [pass the Perl 6 specification](https://rakudo.party/post/Newcomer-Guide-to-Contributing-to-Core-Perl-6#testingstuff) as I believe some of the spec tests use `Callable`s as matchers, which your PR doesn't address).

The named matchers in `throws-like` are smartmatched against the received `Exception`'s methods and smartmatch against `False` is always `False`. To properly test for a Bool:D value you'd use (for example) `* === False` or `* === True` WhateverCode.

I think the change that should be made is to watch for `Bool:D` matchers and throw, the same as how we do with `.grep` for example:

```
$ perl6 -e '().grep: True'
Cannot use Bool as Matcher with '.grep'.  Did you mean to use $_ inside a block?
  in block <unit> at -e line 1
```

and perhaps improve the message of that exception (`X::Match::Bool`) to mention `* === True` / `* === False` WhateverCodes as well.

(N.B.: I don't really want it to do `===` test automatically for Bool:D matchers, as that breaks expectations of how smartmatch works elsewhere; best to consistently re-enforce that message)

PR #2393, opened 2018-10-19
Make .[] cache the Seq. Closes #2383

### LINKED ISSUE #2383: .[] doesn't cache a Seq
Positional access on a `Seq` caches it. However, there's an edgecase where `.[]` doesn't cache it.

On `List`, `.[]` is a no-op and doesn't reify the `List`. I feel that the latter part can still apply to `Seq`s, but `.[]` should `.cache`, same as all other positional access would.

```
$ perl6 -e 'Nil for my $s := "abc".comb[*]; say $s'
(a b c)

$ perl6 -e 'Nil for my $s := "abc".comb[]; say $s'
The iterator of this Seq is already in use/consumed by another Seq
(you might solve this by adding .cache on usages of the Seq, or
  in block <unit> at -e line 1

$ perl6 -v
This is Rakudo version 2018.06-146-g1b3dd35 built on MoarVM version 2018.06-141-gcae4731
implementing Perl 6.c.
```

FCO:


zoffixznet:
👍 Thanks!

Issue #2392, opened 2018-10-19
Mathy ops don't write native attributes operations to native ops


zoffixznet:
Normally, we rewrite core mathy ops on native variables to cheaper native ops:

```
$ perl6 --target=optimize -e 'my int $x, my int $y;  $x ≤ ++$y'
...
              - QAST::Stmt <sunk final> $x ≤ ++$y
                - QAST::Stmts  ≤
                  - QAST::Stmt <wanted final>
                    - QAST::Op(hllbool) 
                      - QAST::Op(isle_i) 
                        - QAST::Var(lexicalref $x) <wanted> $x
                        - QAST::Op(assign_i)  ++
                          - QAST::Var(lexicalref $y) <wanted> $y
                          - QAST::Op(add_i) 
                            - QAST::Var(lexicalref $y) <wanted> $y
                            - QAST::Want  ++
                              - QAST::WVal(Int) 
                              - Ii
                              - QAST::IVal(1) 
...
```

The same isn't done for native *attributes* however:
```
$ perl6 --target=optimize -e 'class Foo { has int $!x; has int $!y; method z { $!x ≤ ++$!y } }'
...
                            - QAST::Op(callstatic &infix:<≤>)  :statement_id<5> ≤
                              - QAST::Var(attributeref $!x) <wanted> $!x
                                - QAST::Var(local __lowered_lex_1) 
                                - QAST::WVal(Foo) 
                              - QAST::Op(callstatic &prefix:<++>) <wanted> ++
                                - QAST::Var(attributeref $!y) <wanted> $!y
                                  - QAST::Var(local __lowered_lex_1) 
                                  - QAST::WVal(Foo) 

...
```

------------------------

For the `++` thing, a wild guess is this conditional needs to account for `attributeref` in addition to `lexicalref` it's doing now: https://github.com/rakudo/rakudo/blob/136164427caf06d1c1f7ab971ab435e75683fc20/src/Perl6/Optimizer.nqp#L1916

PR #2394, opened 2018-10-19
Optimize more ops on native attributes

MasterDuke17:
By re-writing into native ops.

At least partially fixes https://github.com/rakudo/rakudo/issues/2392.

Before, `class Foo { has int $!x; has int $!y; method z { $!x ≤ ++$!y } }; my $f = Foo.new(:x((^10).pick), :y((^10).pick)); my $z; for ^1_000_000 { $z = $f.z; }; say $z; say now - INIT now` took 0.60s, after it takes 0.11s.


zoffixznet:
👍 Thanks!

Issue #2320, opened 2018-09-25
LTA error with $^foo/$:foo variables in methods


zoffixznet:
***See [Newcomer Guide to Contributing to Core Perl 6](https://rakudo.party/post/Newcomer-Guide-to-Contributing-to-Core-Perl-6) for tips on fixing this Issue***

----------------

```
$ perl6 -e 'class { method z { $^a } }'
===SORRY!=== Error while compiling -e
Placeholder variables cannot be used in a method
at -e:1
------> class { method⏏ z { $^a } }



16:12 | SmokeMachine | what this mean? `Placeholder variables cannot be used in a method`
-- | -- | --
16:12 | SmokeMachine | what's a Placeholder variable?
```

Should probably extend the error text to say `"please specify an explicit signature"` or something along the lines. Maybe explain the vars too:

```
Placeholder variables (e.g $^foo/$:foo) cannot be used in a method. 
Please specify an explicit signature
```

Bonus points for reporting actual variable name the user tried to use.

dogbert17:
Shouldn't this be closed? The error message has been improved and tests have been added. Perhaps I'm missing something here.

vrurg:
Closing.

@dogbert17 thanks for clearing up a bunch of tickets!

PR #2415, opened 2018-10-23
Fix placeholder method error

hoelzro:
Addresses GH #2320 

There are a few things I'd like some discussion on for this PR:

  * I noticed there's an `X::Signature::Placeholder` - should this error be using something like that for its exception?  I understand the documentation for that exception class indicates an explicit signature conflicting with a placeholder, but the class name itself implies a more broad category.
  * I haven't added tests yet - what's the best way to test error messages without them being too brittle?
  * I didn't know that there was such a thing as slurpy placeholders - I don't know what that would look like, so this change might not work with those!

### DIFF
--- a/src/Perl6/Actions.nqp
+++ b/src/Perl6/Actions.nqp
@@ -4269,7 +4269,26 @@ class Perl6::Actions is HLL::Actions does STDActions {
         else {
             $past := WANTED($<blockoid>.ast,'method_def');
             if $past.ann('placeholder_sig') {
-                $/.PRECURSOR.panic('Placeholder variables cannot be used in a method');
+                my $placeholders := nqp::iterator($past.ann('placeholder_sig'));
+                my @non-placeholder-names;
+                my $method-name := $<longname><name>;
+                while $placeholders {
+                    my $placeholder := nqp::shift($placeholders);
+                    my $name := $placeholder<placeholder>;
+                    my $non-placeholder-name;
+                    if $placeholder<named_names> {
+                        $non-placeholder-name := nqp::concat(':', nqp::concat(nqp::substr($name, 0, 1), nqp::substr($name, 2)));
+                    } else {
+                        $non-placeholder-name := nqp::concat(nqp::substr($name, 0, 1), nqp::substr($name, 2));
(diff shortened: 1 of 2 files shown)


zoffixznet:
> what's the best way to test error messages without them being too brittle?

Currently, that goes to https://github.com/rakudo/rakudo/blob/master/t/05-messages/03-errors.t

>  slurpy placeholders - I don't know what that would look 

I believe these are `*@`, `*%`, and `|`, aren't they?

src/Perl6/Actions.nqp
```diff
@@ -9688,6 +9709,7 @@ class Perl6::Actions is HLL::Actions does STDActions {
             pos_slurpy        => $pos_slurpy,
             named_slurpy      => $named_slurpy,
             placeholder       => $full_name,
+            ast               => $/,
```
zoffixznet: Is this a common name used for this purpose?

All I ever seen `ast` used for made stuff, like QAST trees, not for Perl6::Grammar objects. `node` is used for those usually in things like QAST nodes.

hoelzro: Thanks for the suggestion! I'll incorporate that.

hoelzro:
Ok, I've added the test and changed `ast` to `node`.

Regarding the slurpy placeholders, looking at the code, it looks like those are `@_` and `%_`.  I ask because my code currently makes the this suggestion:

```
===SORRY!=== Error while compiling /home/rob/projects/rakudo/test.pl
Placeholder variables (eg. $^a) cannot be used in a method.
Please specify an explicit signature, like method z ($a) { ... }
at /home/rob/projects/rakudo/test.pl:6
------>     method z { ⏏$^a }
```

I'm not sure what to suggest if `@_` is used; currently `method z (@) { ... }` is suggested, which is pretty LTA.

hoelzro:
I handle `@_` and `%_` now - although it seems that if I try `class { method z { %_ } }` it does _not_ fail to compile, as expected.

hoelzro:
Although now, looking at the docs, I can't really see the justification for _why_ `@_` isn't allowed!

PR #2425, opened 2018-10-25
Fix potential race condition in Proc::Async start

jstuder-gh:
There is potential for a race here if two threads call start on the same
Proc::Async instance. Add a lock to mitigate that risk.

### DIFF
--- a/src/core/Proc/Async.pm6
+++ b/src/core/Proc/Async.pm6
@@ -102,7 +102,6 @@ my class Proc::Async {
     has $.w;
     has $.enc = 'utf8';
     has $.translate-nl = True;
-    has atomicint $.started = 0;
     has $!stdout_supply;
     has CharsOrBytes $!stdout_type;
     has $!stderr_supply;
@@ -117,6 +116,13 @@ my class Proc::Async {
     has @!promises;
     has $!encoder;
     has @!close-after-exit;
+
+#?if moar
+    has atomicint $.started = 0;
+#?endif
+#?if !moar
+    has Bool $.started = False;
+#?endif
 
     proto method new(|) {*}
     multi method new(*@args where .so) {
@@ -271,6 +277,7 @@ my class Proc::Async {
         $promise;
     }
 
+#?if moar
     method start(Proc::Async:D: :$scheduler = $*SCHEDULER, :$ENV, :$cwd = $*CWD) {
         when cas($!started, 0, 1) == 0 {
             my @blockers;
@@ -283,6 +290,27 @@ my class Proc::Async {
         }
         X::Proc::Async::AlreadyStarted.new(proc => self).throw if $!started;
     }
+#?endif
+
+#?if !moar
+    has $!start-lock = Lock.new;
+    method start(Proc::Async:D: :$scheduler = $*SCHEDULER, :$ENV, :$cwd = $*CWD) {
+        X::Proc::Async::AlreadyStarted.new(proc => self).throw if $!started;
+        $!start-lock.protect: {
+            X::Proc::Async::AlreadyStarted.new(proc => self).throw if $!started;
+            $!started = True;
+
+            my @blockers;
+            if $!stdin-fd ~~ Promise {
+                @blockers.push($!stdin-fd.then({ $!stdin-fd := .result }));
+            }
+            @blockers
+                ?? start { await @blockers; await self!start-internal(:$scheduler, :$ENV, :$cwd) }
+                !! self!start-internal(:$scheduler, :$ENV, :$cwd)
+        }
+    }
+#?endif
+
 
     method !start-internal(:$scheduler, :$ENV, :$cwd) {
         my %ENV := $ENV ?? $ENV.hash !! %*ENV;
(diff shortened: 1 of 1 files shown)


src/core/Proc/Async.pm6
```diff
@@ -271,17 +271,21 @@ my class Proc::Async {
         $promise;
     }
 
+    has $!start-lock = Lock.new;
     method start(Proc::Async:D: :$scheduler = $*SCHEDULER, :$ENV, :$cwd = $*CWD) {
         X::Proc::Async::AlreadyStarted.new(proc => self).throw if $!started;
-        $!started = True;
+        $!start-lock.protect: {
```
ugexe: Does all of this logic need to be inside the lock, or could would the following suffice?
```
method start(Proc::Async:D: :$scheduler = $*SCHEDULER, :$ENV, :$cwd = $*CWD) {
    $!start-lock.protect: {
        X::Proc::Async::AlreadyStarted.new(proc => self).throw if $!started;
        $!started = True;
    }

    ...
}
```

jnthn: No need for a lock at all, just use `cas` on the `$!started`.

jstuder-gh: I updated the branch to use ```cas``` as suggested

src/core/Proc/Async.pm6
```diff
@@ -272,16 +272,16 @@ my class Proc::Async {
...
+                @blockers.push($!stdin-fd.then({ $!stdin-fd := .result }));
+            }
+            @blockers
+                ?? start { await @blockers; await self!start-internal(:$scheduler, :$ENV, :$cwd) }
+                !! self!start-internal(:$scheduler, :$ENV, :$cwd)
         }
-        @blockers
-            ?? start { await @blockers; await self!start-internal(:$scheduler, :$ENV, :$cwd) }
-            !! self!start-internal(:$scheduler, :$ENV, :$cwd)
+        X::Proc::Async::AlreadyStarted.new(proc => self).throw if $!started;
```
ugexe: Won't this always throw an exception? Additionally I think you'd want to do an atomic read of `$!started` ( although I'm not sure you need an `atomicint` at all for this )

jstuder-gh: Since the ```when``` block is used, it won't execute the following statements in the same outer block. It looks bit cleaner than an if/else chain IMO.

zoffixznet: > It looks bit cleaner

Note that it's also about an order of magnitude slower:

```
<Zoffix__> c: HEAD sub foo ($x) { when $x == 1 { }; Nil }; my $x = 1; foo $x for ^10_000_000; say now - ENTER now
<committable6> Zoffix__, ¦HEAD(1ea3ff2): «5.9091969␤»
<Zoffix__> c: HEAD sub foo ($x) { if $x == 1 { } else {} }; my $x = 1; foo $x for ^10_000_000; say now - ENTER now
<committable6> Zoffix__, ¦HEAD(1ea3ff2): «0.7404541␤»
```

zoffixznet: > if an atomicintis needed for a 0-or-1 cas switch

It isn't. `cas` can work with arbitrary objects: https://github.com/perl6/doc/issues/2393

src/core/Proc/Async.pm6
```diff
@@ -118,6 +117,13 @@ my class Proc::Async {
     has $!encoder;
     has @!close-after-exit;
 
+#?if moar
+    has atomicint $.started = 0;
+#?endif
+#?if !moar
+    has Bool $.started = False;
+#?endif
```
jnthn: We should have it as the `Bool` on all platforms. The `started` attribute is exposed API (due to the accessor); we should not be changing it to start returning an integer.

jstuder-gh: Changed it back with most recent commit

src/core/Proc/Async.pm6
jstuder-gh: I do like the ```when``` aesthetically, but I changed it to ```if```/```else``` in the most recent commit due to the performance difference you pointed out.

jstuder-gh:
I've added another commit that splits the code common to both moar and !moar ```start``` into it's own internal method for reuse. Not sure if this new internal method can be inlined, though. If this doesn't make sense from a performance standpoint, I can undo it.

Would it make more sense to have all backends use the lock for simplicity's sake? I was thinking that moar could benefit from the ```cas``` and avoid blocking on the locks, but that creates separate ```start``` methods and maybe it's better to just have one.

JJ:
Can you please solve conflict by applying the changes to the new file?

niner:
We favor speed over simplicity. So using `cas` on MoarVM and a lock on backends where that isn't available is fine.

AlexDaniel:
FWIW the code was changed a little bit so this is a good task for anybody who loves tackling merge conflicts.

lizmat:
PR re-imagined with https://github.com/rakudo/rakudo/commit/8321ce10da , closing now.

PR #2439, opened 2018-10-28
fix for missing defn features

tbrowder:


tbrowder:
I think I missed another code change requirement—will fix when I get home.

PR #2443, opened 2018-10-29
Add type constraints to port numbers in IO::Socket::Async

### LINKED ISSUE #2424: High port in IO::Socket::Async.listen appears to hang on Win32
On 2017.07 on Win7 this code appears to hang. On Linux, it gives an error about wrong argument:

```
C:\Users\z>perl6 -e "react whenever IO::Socket::Async.listen: 'localhost', 2**16+100 {say 42}"
```

jmaslak:
I believe this fixes #2424.  I'm not sure if this is an acceptable way to fix it.  I also don't have the ability to test this on Windows (but see no reason it wouldn't work - famous last words, I know)  I couple things:

I defined a subset (Port-Number) in I::S::A that includes undefined and ^65536.  I changed the class properties representing port numbers to use the subset.  I also matched the coerced port numbers in the various methods that take port numbers against this subset.

This does change the exceptions thrown - I'm not sure if this is unacceptable or not.  I do believe the exceptions make more sense now.

I also changed the test for an invalid high port number, restoring it's ability to execute again.  I removed the todo noting the less-than-helpful exception that was previously thrown.

If I should do something differently, I'm open to that.

### DIFF
--- a/src/core/IO/Socket/Async.pm6
+++ b/src/core/IO/Socket/Async.pm6
@@ -8,11 +8,13 @@ my class IO::Socket::Async {
     has $!close-promise;
     has $!close-vow;
 
+    subset Port-Number of Int where { !defined($_) or $_ ~~ ^65536 };
+
     has Str $.peer-host;
-    has Int $.peer-port;
+    has Port-Number $.peer-port;
 
     has Str $.socket-host;
-    has Int $.socket-port;
+    has Port-Number $.socket-port;
 
     method new() {
         die "Cannot create an asynchronous socket directly; please use\n" ~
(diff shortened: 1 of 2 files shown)


niner:
Wouldn't uint16 be kinda the perfect type for this?

jmaslak:
int16 doesn't actually do the type check, it just silently wraps when assigned into.  I'm thinking throwing an exception is what we want.

Example:
```
my int16 $b = 70_000; say $b;
```

Gives output (no error indication):
```
4464
```

I'm thinking we want the exception if someone tries to listen() on port 70_000.

zoffixznet:
Is this good to merge, considering native type is currently buggy and cannot be used instead?

jmaslak:
I'd like to see this get merged.  :)  Can someone do that for me?

lizmat:
Thank you!

src/core/IO/Socket/Async.pm6
```diff
@@ -290,8 +292,8 @@ my class IO::Socket::Async {
         method serial(--> True) { }
     }
 
-    method listen(IO::Socket::Async:U: Str() $host, Int() $port, Int() $backlog = 128,
-                  :$enc = 'utf-8', :$scheduler = $*SCHEDULER) {
+    method listen(IO::Socket::Async:U: Str() $host, Int() $port where * ~~ Port-Number,
```
MasterDuke17: FWIW, `where Port-Number` is a bit faster. You don't need the `* ~~`.

PR #2460, opened 2018-11-02
move pod-only code to end of their enclosing blocks

tbrowder:


tbrowder:
This implements the first method in my proposed reorganization per gist <https://gist.github.com/tbrowder/e855755022483a2cd829acb59f3a020e>.

ugexe:
I would suggest waiting more than a few hours to see how others feel about large format-only changes in the future.

tbrowder:
You’re probably right Nick, but, as @AlexDaniel says, “we can always revert.”

AlexDaniel:
@tbrowder yes, except for that particular case… Very often people have concerns regarding `git blame` and other git features that rely on the latest change to a particular line. Basically, now you touched a bunch of lines, and if we revert it, we will touch them again… meaning that there's no way to get back to the state where these lines were untouched :) Either way its not the end of the world, but @ugexe++ is right that these kind of changes should sit for a little bit before being merged.

AlexDaniel:
**For anyone trying to make sense of these changes**: this is exactly the situation where you need `patience` strategy to review the actual changes instead of looking at the default diff behavior where brackets from functions are matched to brackets in other functions. See this:

```bash
git diff --patience 71ba96683^..71ba96683
```

Looking at it, the change is much more simple than it looks on github :)

tbrowder:
*Mea culpa*!  My only defense is my large change waiting in the wings and it was getting harder and harder to merge with all the other grammar and actions changes.  And there is always the log and history to lay the *blame*, so that is not lost AFAIK.

AlexDaniel:
Yes, nothing is lost. In fact, for me there's no problem .

This:
```
git blame -w -M src/Perl6/Grammar.nqp
```
Will show the blame correctly (ignoring whitespace and moved lines). However, it is true that `-M` flag can't be used on github (though `-w` should work fine).

PR #2475, opened 2018-11-04
Reimplement .native-descriptor for async sockets

### LINKED ISSUE #2437: .native-descriptor for async sockets
Even though @Kaiepi++ did some great job to get .native-descriptor stuff working, we had to revert it before the release.

Reverted in:
* Moar https://github.com/MoarVM/MoarVM/commit/b2e870c6ff
* NQP https://github.com/perl6/nqp/commit/16f0f458fa
* Rakudo https://github.com/rakudo/rakudo/commit/005f66670e

These commits can be rereverted after the release if https://github.com/MoarVM/MoarVM/issues/987 is resolved.

Ping @Kaiepi.

Kaiepi:
Fixes #2437 
Passes `make test`, `make spectest`

### DIFF
--- a/src/core/IO/Socket/Async.pm6
+++ b/src/core/IO/Socket/Async.pm6
@@ -196,11 +196,18 @@ my class IO::Socket::Async {
     }
 
     class ListenSocket is Tap {
+        has Promise $!VMIO-tobe;
         has Promise $.socket-host;
         has Promise $.socket-port;
 
-        method new(&on-close, Promise :$socket-host, Promise :$socket-port) {
-            self.bless(:&on-close, :$socket-host, :$socket-port);
+        submethod TWEAK(Promise :$!VMIO-tobe, Promise :$!socket-host, Promise :$!socket-port) { }
+
+        method new(&on-close, Promise :$VMIO-tobe, Promise :$socket-host, Promise :$socket-port) {
+            self.bless(:&on-close, :$VMIO-tobe, :$socket-host, :$socket-port);
+        }
+
+        method native-descriptor(--> Int) {
+            nqp::filenofh(await $!VMIO-tobe)
         }
     }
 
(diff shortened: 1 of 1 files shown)


Kaiepi:
Related pullreqs: https://github.com/MoarVM/MoarVM/pull/991, https://github.com/perl6/nqp/pull/513

Kaiepi:
Can I get a review on this? I need some of the features from this for another pullreq I'm making

niner:
I've tried to review this, but it's not easy. The commit message just tells me that something is re-implemented, but not why it needs re-implementing in the first place. Also it doesn't look like .native-descriptor was available before this commit, so it's more like "implement .native-descriptor" anyway.

FWIW I don't see anything obviously wrong in the commit either. The only thing I wonder is why ListenSocket needs a TWEAK submethod now? With the commit message expanded and the TWEAK question answered (by comment or removal), I'll be happy to merge this!

Kaiepi:
The TWEAK's needed since `$!VMIO` is a private property. `.native-descriptor` was implemented at one point, but was reverted before it could make it into 2018.10, which is why I called it reimplement instead of implement

lizmat:
Shouldn't https://github.com/rakudo/rakudo/pull/2475/files#diff-e597449b4134c8062ae28659c3f3a208R311 not also need an `await`?  As it is written now, it will return a `Promise` rather than an `Int`?

Kaiepi:
No, since that's a native object and not a Promise. In the future it may become one though

lizmat:
Well, maybe the `Promise` one should be given another name then?  Having the same name in 2 classes in the same file for something that is inherently different, is a source of a future maintenance burden.

Kaiepi:
What name for it would you recommend?

lizmat:
$!VMIO-promise
$!VMIO-tobe
$!VMIO-eventually
?

Kaiepi:
I like `$!VMIO-tobe`

PR #2478, opened 2018-11-05
[IP6NS Grant] Implement .set-option/.get-option for async and INET sockets

Kaiepi:
Done:
- expose socket option types as constants
- write `.get-option`/`.set-option` methods for `IO::Socket` and `IO::Socket::Async
- wait until `.native-descriptor` for async sockets is merged so `.get-option`/`.set-option` can be used with `IO::Socket::Async::ListenSocket`

Todo:
- nothing

### DIFF
--- a/src/core/sockopts.pm6
+++ b/src/core/sockopts.pm6
@@ -1,22 +1,19 @@
 my role SockOptLike {
-    multi method CALL-ME(Int() $sonum) {
-        return self unless $sonum;
-        nextsame
+    multi method CALL-ME(Int() $option) {
+        $option ?? (callsame) !! self
     }
 }
 
-my enum SockOpt does SockOptLike (
-    |nqp::stmts(
-        ( my \res  := nqp::list ),
-        ( my \iter := nqp::iterator(nqp::getsockopts()) ),
-        nqp::while(
-            iter,
-            nqp::stmts(
-                ( my \p := nqp::p6bindattrinvres(nqp::create(Pair), Pair, '$!key', nqp::shift(iter)) ),
-                nqp::bindattr(p, Pair, '$!value', nqp::abs_i(nqp::shift(iter)) ),
-                nqp::push(res, p)
-            )
-        ),
-        res
-    )
-);
+enum SockOpt does SockOptLike ( |do {
+    my $res  := nqp::list();
+    my $iter := nqp::iterator(nqp::getsockopts());
+    nqp::while(
+      $iter,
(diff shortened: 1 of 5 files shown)


jmaslak:
This is awesome!  I'd love a way to also set options before the listen() or connect() call with IO::Socket::Async - the option I'm needing is TCP_MD5.  I've written a look-similar module to I::S::A that lets give me these hooks, but I'd love to have something portable (although this particular TCP option isn't portable between OSes).

Kaiepi:
>This is awesome! I'd love a way to also set options before the listen() or connect() call with IO::Socket::Async

I was actually planning on writing that after this, since `SO_REUSEADDR` is impossible to use without being able to do that. I'll need to discuss with others how best to implement it though since it almost certainly would end up becoming a breaking change the way `.listen` handles creating sockets

Kaiepi:
Now that this is pretty well finished, can I get a review?

src/core/sockopts.pm6
```diff
@@ -0,0 +1,22 @@
...
+            iter,
+            nqp::stmts(
+                ( my \p := nqp::p6bindattrinvres(nqp::create(Pair), Pair, '$!key', nqp::shift(iter)) ),
+                nqp::bindattr(p, Pair, '$!value', nqp::abs_i(nqp::shift(iter)) ),
+                nqp::push(res, p)
+            )
+        ),
+        res
+    )
+);
```
jstuder-gh: One thing to note is that in the enum definition, a separate lexical scope isn't being created, so the variables introduced there leak into the [CORE lexical environment](https://github.com/rakudo/rakudo/issues/2773). A mistake I'm guilty of as well 😝.

Check out  @lizmat's [commit here](https://github.com/rakudo/rakudo/commit/76b6416d00dca8dc2f669b438581e34fefdf49bb) which adds a new lexical scope and avoids leakage.

src/core/sockopts.pm6
```diff
@@ -0,0 +1,19 @@
+my role SockOptLike {
+    multi method CALL-ME(Int() $option) {
+        $option ?? (callsame) !! self
+    }
+}
```
jnthn: Why this? I think the `enum` should just have its standard coercion semantics?

JJ:
Due to changing file names, it's got conflicts now too. Can you please check?

Kaiepi:
See the MoarVM PR for the reasoning behind closing this.

PR #2481, opened 2018-11-10
Remove redundant smartmatch usage in IO::Socket::Async

jmaslak:
Based on comments in #2443 by @b2gills, this changes "where * ~~ Port-Number" to simply "where Port-Number".


jnthn:
Thanks for tidying that up :+1:

jmaslak:
Thanks for merging.  At some point I'll actually know how to write Perl 6 code.  ;)

Issue #2122, opened 2018-07-23
Apparent laziness propagation bug in triangle reduce


zoffixznet:
Spotted here: https://stackoverflow.com/questions/51484963/why-does-perl-6-try-to-evaluate-an-infinite-list-only-in-one-of-two-similar-situ

Looks like triangle reduce fails to notice lazy stuff in some cases. Non-triangle reduce works fine:

```
<Zoffix_> m: say (0, |(1, 2 ... *)).is-lazy
<camelia> rakudo-moar 85495890d: OUTPUT: «True␤»
<Zoffix_> m: my @a = [+] (0, |(1, 2 ... *)); say "fine"
<camelia> rakudo-moar 85495890d: OUTPUT: «fine␤»
<Zoffix_> m: my @a = [\+] (0, |(1, 2 ... *)); say "hangs"
<camelia> rakudo-moar 85495890d: OUTPUT: «(timeout)»
```

jstuder-gh:
It looks like the sum reduce handling laziness is a [special case](https://github.com/rakudo/rakudo/blob/master/src/core/metaops.pm6#L185). Reduce with an arbitrary operator will still hang unfortunately.

```perl6
my @a = [*] (1, |(1, 2 ... *)); say 'fine'
# RESULT: hangs indefinitely
```

In the case of triangle reduce, you are right, it's absolutely an issue with an incorrect laziness value being propagated up the line (starting in ```List::Reifier```). I've submitted a [PR that addresses this issue](https://github.com/rakudo/rakudo/pull/2493).

lizmat:
Fixed the `[\+]` case with 8f424c9d8e

Altai-man:
1)The PR the commit was re-imagening is still open, why not close it?
2)As for 2021.05, the problem is not resolved?

```
<sena_kun> m: say (0, |(1, 2 ... *)).is-lazy
<camelia> rakudo-moar 1bd0411c0: OUTPUT: «True␤»
<sena_kun> m: my @a = [+] (0, |(1, 2 ... *)); say "fine"
<camelia> rakudo-moar 1bd0411c0: OUTPUT: «fine␤»
<sena_kun> m: my @a = [\+] (0, |(1, 2 ... *)); say "hangs"
<camelia> rakudo-moar 1bd0411c0: OUTPUT: «(timeout)»
```

melezhik:
still fails on the latest commit?

```
[root@fa619e557332 r3tool]# tomty 2122
[...]

PR #2493, opened 2018-11-17
Fix is-lazy propagation bug in List::Reifier

jstuder-gh:
Triangle reduce was not recognizing Lists containing lazy Slips as lazy.
This can be tracked back to the is-lazy method on List::Reifier (and was
ultimately being propagated up to the triangle reduce routine). is-lazy
in List::Reifier was not considering the possibility of lazy Slips. Fix
this by having the method check any $!future elements it may have for
laziness.

See [Github Issue #2122](https://github.com/rakudo/rakudo/issues/2122)

### DIFF
--- a/src/core/List.pm6
+++ b/src/core/List.pm6
@@ -187,7 +187,7 @@ my class List does Iterable does Positional { # declared in BOOTSTRAP
                   nqp::while(
                     $iter && nqp::eqaddr($res, False),
                     nqp::if(
-                      nqp::eqaddr((my $cur := nqp::shift($iter)), Slip),
+                      nqp::istype((my $cur := nqp::shift($iter)), Slip),
                       ($res := $cur.is-lazy),
                     ),
                   ),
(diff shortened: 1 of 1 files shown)


jstuder-gh:
Pushed up one more commit. The previous one was a bit too overzealous in declaring the resulting ```List``` lazy. From the commit message:

> Make is-lazy only check the is-lazy values of Slips in the $!future.
> The way it was before, any lazy element following a Slip would cause the
> result to be lazy, which is not quite right.
> 
> ```perl6
> my @a = [\+] (0, |(1, 2), (3, 4 ... *));
> # RESULT: @a would be lazy, but shouldn't be
> ```

lizmat:
Thank you for your work.  Since there was no reply to my comment and we're 3 months later now, I decided to take the idea and re-imagine it.

lizmat:
Reverted my version of the PR as it caused a regression: https://github.com/rakudo/rakudo/issues/2920 .

JJ:
Can you please fix the conflict and continue here?

lizmat:
Closing this PR as my re-imagining didn't work out, and this PR didn't get its conflicts solved.

PR #2508, opened 2018-11-26
Fix shaped dynamic var (my %*a{Int})

FCO:


AlexDaniel:
Not important, but still interesting, it used to work on 2015.10 and some releases before: https://gist.github.com/246c36cb06148ffaebe2c4ae8e017fd8
Then it kinda stopped after (2015-11-05) https://github.com/rakudo/rakudo/commit/1458f0ea3ea2943252c4436fd66a22861845f8ad

@FCO what about tests?

FCO:
@AlexDaniel Im entering into a meeting now... Ill try to write it at night...

FCO:
Tests are merged

lizmat:
Thanks!

PR #2511, opened 2018-12-02
Ignore library versions on OpenBSD.

mornfall:
See the commit for details. This seems reasonable to me, since absolute ABI
version numbers don't really carry the information that is desired. In an ideal
world, rakudo would perhaps follow the way libraries work in C programs: at
build time, you say -lcairo and the system linker picks a version. The compiled
program then includes the major version (SONAME) that was used, since this
corresponds to the ABI it needs. On the other hand, what I think the intention
really was is to specify the API version, since that is what is presumably
relevant for NativeCall-based modules. Unfortunately, the shared object does
not carry that information (and the ABI version works better on Linuxes as a
proxy than it does on OpenBSD; I don't know about other Unices).

### DIFF
--- a/src/core/IO/Notification.pm6
+++ b/src/core/IO/Notification.pm6
@@ -26,7 +26,7 @@ my class IO::Notification {
                 }
                 else {
                     my $event = rename ?? FileRenamed !! FileChanged;
-                    my $full-path = $is-dir ?? $*SPEC.catdir($path, path) !! $path;
+                    my $full-path = ( $is-dir and path ) ?? $*SPEC.catdir($path, path) !! $path;
                     $s.emit(Change.new(:path($full-path), :$event));
                 }
             },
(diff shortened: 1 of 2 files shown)


lizmat:
Error checking on nqp::ops is very minimal:  this does **not** what you expect:

    nqp::add_i($is-darwin,$is-win,$is-openbsd)

`nqp::add_i` is supposed to take only 2 args, not N.  The error is that there is no error for that.  Observe:

    use nqp; my int $a = 42; dd nqp::add_i($a,$a,$a)  # 84, not 126  :-(

mornfall:
Oh. And it is the *last* two arguments that are used, which explains why the patch works on openbsd (but presumably breaks darwin).

    > use nqp; nqp::add_i(1, 2, 3)
    5

mornfall:
I don't know how to move the PR to a different branch, so I am closing this one and creating a new one, hopefully with the correct patch this time.

lizmat:
Ah, indeed, `nqp::add_i` takes the last two arguments, also on macOS.  My example was ambiguous.

PR #2527, opened 2018-12-08
Fix use of uninitialized value in IO::Notification

mornfall:
When libuv returns a NULL path (which it does for directory
notifications on kqueue/kevent systems, at minimum), the path returned
from nqp::watchfile turns into an undef Str. Hence do not try to concat
this undef with the directory path.


lizmat:
Thanks!

PR #2530, opened 2018-12-08
Ignore shared library versions on OpenBSD.

mornfall:
Shared library version numbers in OpenBSD rarely match those provided by
upstream, breaking many packages which specify them. Replaces PR#2511.


Kaiepi:
This fixes the issues I was having getting OpenSSL to work on OpenBSD. A couple of its tests still fail, but those are just checking versions. If they're different on LibreSSL, that would explain why

AlexDaniel:
Thanks! Consider also adding yourself to https://github.com/rakudo/rakudo/blob/master/CREDITS so that your name shows up (if you prefer that) in the release announcement.

PR #2559, opened 2018-12-22
Add naive $*DISTRIBUTION implementation

ugexe:
Everywhere we load a module where we also know the distribution,
set $*DISTRIBUTION.

This is naive in that:
* ~~perl6 -e 'use Zef; say $*DISTRIBUTION' will show zef's
distribution object, when it should probably be contained to
within the owning distribution.~~
* it requires all CURs (even non-core ones) to set $*DISTRIBUTION
making this a line of required boilerplate.


ugexe:
This is mostly to get the ball rolling. If someone else can improve this or do it better, than please go crazy.

jnthn:
Anything in `PROCESS` is *extremely* global, so I suspect this will just tell us the `Distribution` object of the last module that was loaded. I can probably help work out something better, but:

* Is this a compile-time or runtime property?
* What should its scope ideally be?

ugexe:
It is a run-time property.

Scoping should be similar to `%?RESOURCES` such that the module being loaded may access its own `$*DISTRIBUTION`.

Specifically this is to allow:
```
package MyApp::CLI {
    multi sub MAIN('--version') {
        $*DISTRIBUTION.meta<version>;
    }
}
```

ugexe:
Scoping is now similar to `%?RESOURCES`. I'm not sure if I should remove the following though:

https://github.com/rakudo/rakudo/blob/752db84ac3d30a28a6aedfd4705f79c1d1d689fb/src/core/Distribution.pm6#L224-L226

If removed:
```
$ perl6 -I. -e 'use Zef; say $*DISTRIBUTION'
Dynamic variable $*DISTRIBUTION not found
  in block <unit> at -e line 1
```
If kept:
```
$ perl6 -I. -e 'use Zef; say $*DISTRIBUTION'
(Any)
```

ugexe:
As-is this works for
```
unit module MyModule;
note $*DISTRIBUTION;
```
but not
```
unit module MyModule;
sub distribution is export { $*DISTRIBUTION }
```

ugexe:
Closing in favor of https://github.com/rakudo/rakudo/pull/2574

PR #2590, opened 2019-01-07
Adds comment to avoid deletion

JJ:
Because one could imply move the .gitignore thing to the general file,
and this directory would be deleted. Also eliminated *.moarvm from it,
since it's already in the general file.


jnthn:
I think using a `.gitignore` file - even an empty one - to retain an otherwise empty directory is a quite common practice. If really wanting an additional comment to say why it's still there, then it could go in the `.gitignore` file.

PR #2597, opened 2019-01-10
Don't add concreteness guard if return value = Nil

### LINKED ISSUE #2593: 'index' routine shows severe performance decline in a loop
## The Problem
Performance for 'index' routine in a loop grows increasingly worse as memory use grows without bounds:

`index $_, 0 for 1 .. 10000`

True of 'rindex' as well. Difference in valgrind output between good/bad versions for the above snippet:
```
999,200,592 bytes in 2,018 blocks are still reachable in loss record 2,181 of 2,181
   at 0x10000A261: malloc (vg_replace_malloc.c:302)
   by 0x10004CD16: MVM_fixed_size_alloc (in /Users/dhoekman/.rakudobrew/moar-66f8ee0fb/install/lib/libmoar.dylib)
   by 0x1000D1141: add_resolution_to_guard_set (in /Users/dhoekman/.rakudobrew/moar-66f8ee0fb/install/lib/libmoar.dylib)
   by 0x10003A048: remove_one_frame (in /Users/dhoekman/.rakudobrew/moar-66f8ee0fb/install/lib/libmoar.dylib)
   by 0x100039AC7: MVM_frame_try_return (in /Users/dhoekman/.rakudobrew/moar-66f8ee0fb/install/lib/libmoar.dylib)
   by 0x10001FB81: MVM_interp_run (in /Users/dhoekman/.rakudobrew/moar-66f8ee0fb/install/lib/libmoar.dylib)
   by 0x1001592B5: MVM_vm_run_file (in /Users/dhoekman/.rakudobrew/moar-66f8ee0fb/install/lib/libmoar.dylib)
   by 0x1000012B6: main (in /Users/dhoekman/.rakudobrew/moar-66f8ee0fb/install/bin/moar)

LEAK SUMMARY:
   definitely lost: 888 bytes in 2 blocks
   indirectly lost: 0 bytes in 0 blocks
     possibly lost: 242,192 bytes in 4,328 blocks
   still reachable: 1,058,103,527 bytes in 222,498 blocks
        suppressed: 22,284 bytes in 193 blocks

```

## Environment
The problem began with commit [66f8ee0](https://github.com/rakudo/rakudo/commit/66f8ee0fbbf9235884439048b04c39a017bae876) 

This is Rakudo version 2018.12-64-gef565b2ba built on MoarVM version 2018.12-4-g1404d5168
implementing Perl 6.d.

MasterDuke17:
The guard set doesn't end up matching the values being passed, so the
plugin is repeatedly called when it shouldn't be.

Fixes #2593. Before this change the code in the issue took 4.5s on my machine, /usr/bin/time reported 2.4gb maxresident, and some added logging showed the return type checking spesh plugin being called ~29.5k times. After the change, those values decreased to 0.5s, 89mb, and 30.

This does cause a single spectest fail, in t/spec/S32-temporal/DateTime.t. When run directly, the error is `Type check failed for return value; expected Int:D but got Int (Int) in block <unit> at t/spec/S32-temporal/DateTime.t line 682 # Looks like you planned 299 tests, but ran 248`. However, according to @jnthn, "Hm, that almost looks legit...I wonder if you've fixed a bug too?".

### DIFF
--- a/src/vm/moar/spesh-plugins.nqp
+++ b/src/vm/moar/spesh-plugins.nqp
@@ -257,7 +257,7 @@ sub identity($obj) { $obj }
                 if $definite_check == 0 {
                     nqp::speshguardtypeobj($rv);
                 }
-                elsif $definite_check == 1 {
+                elsif $definite_check == 1 && !nqp::istype($rv, Nil) {
                     nqp::speshguardconcrete($rv);
                 }
 


lizmat:
t/spec/S32-temporal/DateTime.t doesn't fail for me.  So merging!

lizmat:
Thank you!

PR #2600, opened 2019-01-10
Change to actual variable name

JJ:
$? would refer to a compile-time variable which does not actually exist.


lizmat:
Thank you!

PR #2648, opened 2019-01-26
Allow CStruct pass-by-value in Nativecall

scovit:
This is the initial implementation of the pass-by-value
functionality for nativecall, it works on Parameters by recognizing
an 'is copy' trait, and as a new 'is copy' trait on Routine for the
return value. The 'copy' key is then added to the parameter/return
hash. The patch includes also tests of the basic functionality.

### DIFF
--- a/lib/NativeCall.pm6
+++ b/lib/NativeCall.pm6
@@ -66,6 +66,7 @@ sub param_hash_for(Parameter $p) {
     my $type := $p.type();
     nqp::bindkey($result, 'typeobj', nqp::decont($type));
     nqp::bindkey($result, 'rw', nqp::unbox_i(1)) if $p.rw;
+    nqp::bindkey($result, 'copy', nqp::unbox_i(1)) if $p.copy;
     if $type ~~ Str {
         my $enc := $p.?native_call_encoded() || 'utf8';
         nqp::bindkey($result, 'type', nqp::unbox_s(string_encoding_to_nci_type($enc)));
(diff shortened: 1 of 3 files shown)


lizmat:
Suggest we merge this *after* the 2019.01 release

scovit:
Surely there is no hurry as I will be the only one using this functionality for quite some time. I would like to get feedback on the use of the "is copy" trait to indicate pass-by-value. The advantage of this choice is that it is not already used, and it is already implemented and optimized as part of the Parameter class.

The drawback is that, after analysing the Nativecall MoarVM source code, I start to suspect that 'copying' might be a desirable semantics Perl-side in general, while pass by value is very C-side and specific of CStruct? For this reason, another trait name might be preferable. Of course this is not a problem right now, since all of this is not implemented nor planned, but might be an useful discussion if there will ever be a plan to design a NativeCall 2.0 module.

jonathanstowe:
And I think it makes the IBM MQ one which I have been planning for a while but didn't want write another ugly wrapper possible.

lib/NativeCall.pm6
```diff
@@ -203,6 +205,10 @@ multi sub map_return_type($type) {
                             !! nqp::istype($type, Num) ?? Num !! $type;
 }
 
+my role CopyReturnStruct {
+    method ret-struct-copy(--> 1) { }
```
b2gills: `CopyReturnStrct` doesn't need a method as far as I can tell.

AlexDaniel:
@scovit fwiw, maybe this change needs to be reflected on https://docs.perl6.org/language/nativecall, after we merge it.

Also please consider mentoring a GSoC project for improving nativecall: https://github.com/perl-gsoc-2019/ideas (submit it as a new idea).

scovit:
If anybody is wondering why I closed this,
I am working on MoarVM and rakudo with those patches applied locally, so I am keeping them updated, and upon a rebase, github decided to close the pull request.

@jonathanstowe, all, if you want to try this out, or to use it, it is easy, you can find the patch here:

MoarVM: https://github.com/scovit/MoarVM/tree/cstruct-by-value
Rakudo: https://github.com/scovit/rakudo/tree/cstruct-by-value

MoarVM needs to be compiled with --has-libffi

Because of this, I wouldn't use this for code that needs to be publicly shared, just yet.

Issue #2659, opened 2019-01-30
Accidental use of 'is' with a role in class declaration causes 'No concretization' error


vrurg:
## The Problem

If by a class is inheriting from a role instead of doing it then class-qualified calls to role methods fail with 'No concretization found' exception.

## Steps to Reproduce

The following code demosntrated the problem:

     role R2 {
         method foo {
             "R2::foo"
         }
     }
     class Foo is R2 {
         method foo {
             say "Foo::foo" ~ callsame;
         }
     }
     Foo.new.foo;

## Environment
* Operating system: macOS, Linux
* Compiler version (`perl6 -v`): This is Rakudo version 2018.12 built on MoarVM version 2018.12
_Also tested on 2018.08 and 2017.10_

jonathanstowe:
Neither for me, I was  expecting  'Did you mean to use "does" instead?'' but I guess it is punned somehow before adding the new parent.  If that is the case is there a way of having the role to indicate that it shouldn't be inherited from as I'm pretty sure the difference between the two may be significant in some cases.

lizmat:
To prevent a role from being punned, `::?ROLE.HOW.archetypes..inheritalizable` should return a false value at compile time.  It will then die with a `X::Inheritance::Unsupported` exception.  Leaving this as an exercise to the reader on how to make that happen, and what syntactic sugar we will devise for that.

jonathanstowe:
Need a new trait word : ```isn't``` so we can have ```role Foo isn't inheritalizable { }``` ;-)

jnthn:
Use of a role with `is` is a deliberate feature, and actually the point of `inheritalizable` ("am I able to turn this into something I can inherit from?") even existing. The result is to pun the role, and then inherit from that.

The `callsame` behavior looks a bit broken, however.

vrurg:
I'm very sorry for a mistake. While playing with different variants of the code I chose a wrong one to post. The problematic version is following:

     role R2 {
         method foo {
             "R2::foo"
         }
     }
     class Foo is R2 {
         method foo {
             say "Foo::foo", self.R2::foo;
         }
     }
     Foo.new.foo;

And it fails on blead as well as on 2018.12.

lizmat:
FWIW, this gives the same error:

    role R2 {
        method foo {
            "R2::foo"
        }
    }
    my constant R3 = R2.^pun;
    class Foo is R3 {
        method foo {
            say "Foo::foo", self.R2::foo;
        }
    }
[...]

PR #2675, opened 2019-02-06
Fixed #2282, #2657, #2659

### LINKED ISSUE #2282: Applying additional role with 'does' breaks role-qualified method call
## The Problem

When a role is dynamically applied to instantiated object using `does` it breaks role-qualified calls to any previously applied roles.

## Expected Behavior

```
self.Role1::method;
self does Role2;
self.Role1::method;
```

must pass without errors

## Actual Behavior

In the above example second call to `method` would fail with _No concretization found for Role1_ error.

## Steps to Reproduce

The following code demonstrates the bug:

     role Foo1 { method foo { say "Foo1::foo"; } }
     role Foo2 { method foo { say "Foo2::foo"; } }
     role Foo3 { method foo { say "Foo3::foo"; } }

     class Foo { method foo { say "Foo::foo"; } }

     class Bar is Foo does Foo1 {
         method foo {
             say "Bar::foo";
             self.Foo::foo;
             self.Foo1::foo;

             say "-- Do Foo2";
             self does Foo2;
             self.Foo::foo;
             try {
                 self.Foo1::foo;
                 CATCH { default { say $_.message } }
             }
             self.Foo2::foo;

             say "-- Do Foo3";
             self does Foo3;
             try {
                 self.Foo2::foo;
                 CATCH { default { say $_.message } }
             }
             self.Foo3::foo;
         }
     }

     Bar.new.foo;

Being run it produces the output:

> ```
> Bar::foo
> Foo::foo
> Foo1::foo
> -- Do Foo2
> Foo::foo
> No concretization found for Foo1
> Foo2::foo
> -- Do Foo3
> No concretization found for Foo2
> Foo3::foo
> ```

Note that problem occurs with role-qualified calls only while class-qualified remain unaffected.

## Environment
* Operating system: macOS 10.13, Linux Mint 19
* Compiler version (`perl6 -v`): This is Rakudo version 2018.08 built on MoarVM version 2018.08
2018.06 and 2017.10 are affected too.

### LINKED ISSUE #2657: "No concretization found" on role-qualified method call with inheritance
## The Problem

The problem applies to a case when a class is inheriting from a parent with a role applied. Lets assume that the parent defines a method which does a role-qualified call to a method in the role. In this case calling this method on an instance of the child class causes "No concretization found" exception.

## Steps to Reproduce

Here is a code snippet which demonstrated the problem:

```
     role R1 {
         method foo { "R1::foo" }
     }

     class Foo does R1 {
         method foo {
             note "Foo::foo";
             say "direct:   ", self.R1::foo;
         }
     }

     class Bar is Foo {
     }

     say "--- Foo:";
     Foo.new.foo;
     say "--- Bar:";
     Bar.new.foo;
```

Executing it would result in a output like the following:

```
--- Foo:
Foo::foo
direct:   R1::foo
--- Bar:
Foo::foo
No concretization found for R1
  in method foo at ./role.p6 line 8
```

## Notes

This is likely to be related to #2496 and #2282.

## Environment
* Operating system: macOS, Linux
* Compiler version (`perl6 -v`): This is Rakudo version 2018.12 built on MoarVM version 2018.12 _but also observed on 2018.06, 2017.10_

vrurg:
This is a proposed fix for issues #2282, #2657, and #2659. It implements deep resolution for concretization lookups allowing for direct and indirect qualification on a object. In other words, it doesn't matter how deep in the MRO a role is applied – it's concretization will be found and used.

The fix required introduction of additional `Perl6::Metamodel::Concretization` role consumed by `ClassHOW` and `ConcreteRoleHOW`.

The changes was done in a way to minimize any possible impact for the backward compatibility and preserve the spesh-plugin optimizations.

A pull request for the roast is following.

### DIFF
--- a/src/Perl6/Metamodel/ClassHOW.nqp
+++ b/src/Perl6/Metamodel/ClassHOW.nqp
@@ -22,10 +22,10 @@ class Perl6::Metamodel::ClassHOW
     does Perl6::Metamodel::REPRComposeProtocol
     does Perl6::Metamodel::InvocationProtocol
     does Perl6::Metamodel::Finalization
+    does Perl6::Metamodel::Concretization
 {
     has @!roles;
     has @!role_typecheck_list;
-    has @!concretizations;
     has @!fallbacks;
     has $!composed;
 
@@ -98,7 +98,7 @@ class Perl6::Metamodel::ClassHOW
                 @!role_typecheck_list[+@!role_typecheck_list] := $r;
                 my $ins := $r.HOW.specialize($r, $obj);
                 @ins_roles.push($ins);
-                nqp::push(@!concretizations, [$r, $ins]);
+                self.add_concretization($obj, $r, $ins);
             }
             self.compute_mro($obj); # to the best of our knowledge, because the role applier wants it.
             @stubs := RoleToClassApplier.apply($obj, @ins_roles);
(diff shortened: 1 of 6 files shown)


lizmat:
@vrurg: feels to me that you should provide a Pull Request for inclusion of your name in `CREDITS` as well  :-)

vrurg:
@lizmat Thanks! May be later.

PR #2680, opened 2019-02-08
Fix Pointer parameterization with uncomposed types

### LINKED ISSUE #2673: Cannot parameterize NativeCall Pointer argument of native sub using the class within the class
## The Problem
See title.

## Expected Behavior
`Pointer[::?CLASS]` should parameterize when used as an argument to a native sub in a class using the `CStruct` REPR.

## Actual Behavior
This gets thrown:
```
===SORRY!=== Error while compiling /home/morfent/Documents/p6-Crypt-Mode/lib/Crypt/Cipher/Context.pm6 (Crypt::Cipher::Context)
An exception occurred while parameterizing Pointer
at /home/morfent/Documents/p6-Crypt-Mode/lib/Crypt/Cipher/Context.pm6 (Crypt::Cipher::Context):7
Exception details:
  Cannot resolve caller ACCEPTS(Bool:U: Crypt::Cipher::Context); none of these signatures match:
      (Bool:D: Mu \topic, *%_)
      (Bool:U: \topic, *%_)
    in method parameterize at /home/morfent/.perl6/share/perl6/sources/8660F65A7B3492675BB3B2058DB30E411A4C4E54 (NativeCall::Types) line 80
```

## Steps to Reproduce

Compile this as `libfoo.so`:
```c
#include <stddef.h>
#include <stdlib.h>

#define ERR_OK   0
#define ERR_OOM -1

typedef struct Foo {
    int bar;
} Foo;

int
foo_init(Foo **fooptr, int bar)
{
    *fooptr = calloc(1, sizeof(Foo));
    if (*fooptr == NULL) return ERR_OOM;

    (*fooptr)->bar = bar;
    return ERR_OK;
}

void
foo_free(Foo *foo)
{
    if (foo != NULL) free(foo);
}

const uint8_t *
foo_strerror(int err)
{
    switch (err) {
        case ERR_OK:
            return "success";
        case ERR_OOM:
            return "out of memory";
        default:
            return "unknown";
    }
}
```

Run this in the same path:
```perl6
use v6.d;
use NativeCall;
unit class Foo is repr('CStruct');

my Str constant LIB = './foo'.IO.absolute;

sub foo_init(Pointer[::?CLASS] is rw, int32 --> int32) is native(LIB) {*}
sub foo_free(::?CLASS) is native(LIB) {*}
sub foo_strerror(int32 --> Str) is native(LIB) {*}

method new(Int $bar = 0) {
    my Pointer[::?CLASS] $ptr .= new;
    my Int $err = foo_init($ptr, $bar);
    die "Failed to initialize Foo: {foo_strerror($err)}" if $err;
    $ptr.deref
}

submethod DESTROY() {
    foo_free(self)
}
```

## Environment
* Operating system: 
OpenBSD bastille.kennel.qt 6.4 GENERIC.MP#364 amd64
This is Rakudo version 2018.12-279-g92ebc333a built on MoarVM version 2018.12-45-ga9d02578a
implementing Perl 6.d.

Kaiepi:
Fixes #2673

### DIFF
--- a/lib/NativeCall/Types.pm6
+++ b/lib/NativeCall/Types.pm6
@@ -6,16 +6,17 @@ sub nativecast($target-type, $source) {
         nqp::decont(map_return_type($target-type)), nqp::decont($source));
 }
 
-our native long     is Int is ctype("long")     is repr("P6int") { };
-our native longlong is Int is ctype("longlong") is repr("P6int") { };
+our native long      is Int is ctype("long")                 is repr("P6int") { };
+our native longlong  is Int is ctype("longlong")             is repr("P6int") { };
 our native ulong     is Int is ctype("long")     is unsigned is repr("P6int") { };
 our native ulonglong is Int is ctype("longlong") is unsigned is repr("P6int") { };
 our native size_t    is Int is ctype("size_t")   is unsigned is repr("P6int") { };
 our native ssize_t   is Int is ctype("size_t")               is repr("P6int") { };
 our native bool      is Int is ctype("bool")                 is repr("P6int") { };
-our class void                                  is repr('Uninstantiable') { };
+our class void                                               is repr('Uninstantiable') { };
+
 # Expose a Pointer class for working with raw pointers.
-our class Pointer                               is repr('CPointer') {
+our class Pointer is repr('CPointer') {
     method of() { void }
 
     multi method new() {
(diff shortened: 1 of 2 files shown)


vendethiel:
Does this break `$date after any(@dates)`?

jnthn:
We use `Any` in place of `Mu` in order to trigger auto-threading of `Junction`; I suspect this will cause regressions in that regard.

Kaiepi:
It did cause a regression. Updated `before` and `after` to take `Any` again

lizmat:
Also: R#2676

raiph:
Hi @Kaiepi 

> It did cause a regression. Updated before and after to take Any again

But what about the rest?

As jnthn said, the distinction between Mu and Any is what allows junctions to automatically do their thing to operations that don't know about junctions (which is almost all of them).

Conversely, the way to make operations *aware* of junctional arguments is precisely to have them take Junctions (or Mu).

So changing an operation to take a Mu argument instead of Any is akin to changing it to claim it's aware of Junctions and wants to stop them doing their usual thing.

That *might* be OK during the bootstrap for some or all operations (I don't know; does any of the code during the bootstrap use `any`, all`, etc. or `|` etc.?) but not once the bootstrap is over if those operations are supposed to be usable with junctions (which almost all are -- by taking Any or narrower arguments) after the bootstrap.

Kaiepi:
I see. `Bool.ACCEPTS` still works with the changes, but `&infix:<===>` does not. Since all the other equality operators work with `Any` instead of `Mu`, I'll change `Pointer.^parameterize` to use `~~` instead of `===` to fix the second bug from the issue

JJ:
Again, conflicts.

PR #2691, opened 2019-02-10
Fix segfaults in Blob.new and Buf.STORE

Kaiepi:
The composed Blob didn't have any elements to assign to and also took
the wrong number of elements from the passed argument in certain cases.


Kaiepi:
I missed `Buf.new`, give me a minute to fix that as well

Kaiepi:
Updated

Kaiepi:
This ended up being the wrong fix after further testing. I'll make another pullreq later

PR #2693, opened 2019-02-13
Relocatable Rakudo + C based perl6 runner

patzim:
This PR makes rakudo relocatable by default. The old behavior can be reactivated with the --no-relocatable Configure flag.
I also replaced the perl6 and perl6-debug shell runners with a C based runner that links libmoar directly.
I introduced two environment variables PERL6_HOME and NQP_HOME that can be used to set the path perl6 searches for its NQP/Perl6 files. There was some discussion of whether these variables are a good idea the way they are. [See here](https://gist.github.com/patzim/710baab43da6e8f127df563b5b81c923).
I tested this manually on Linux/Moar, Linux/JVM and Windows10.
I still have to do a full spectest (I didn't manage to get it to run yet) and  and some performance tests.
This PR is mostly about getting feedback from other devs. I do not expect this PR (and the respective PRs in NQP and Rakudo) to be merged right away.


ugexe:
Does PERL6_HOME make more sense than RAKUDO_HOME? i.e. does it make sense for anything calling itself perl6 to implement this, or is it specific to rakudo?

AlexDaniel:
Holy butterflies. This is amazing.

patzim:
> Does PERL6_HOME make more sense than RAKUDO_HOME? i.e. does it make sense for anything calling itself perl6 to implement this, or is it specific to rakudo?

My reasoning for the PERL6_HOME variable name was, that the folder itself is called `perl6` and is located in a most obvious place (/usr/share/perl6) and thus already looks like the one place all perl6es install to. Also it might (I'm not sure) make sense for different perl6 implementations to share their CompRepos.
On the other hand this might just end up being a mess once another implementation arises and the two step on their toes in that folder.
Also see [this gist](https://gist.github.com/patzim/710baab43da6e8f127df563b5b81c923#file-second_try-md) for some more ramblings on those environment variables.

ugexe:
Ah I see. I was thinking it was being used as a prefix for the executable — not as a type of user definable default @INC

MasterDuke17:
https://github.com/rakudo/rakudo/blob/e94b49eae4200aa594af26c5f2381f1b8c6a0a6a/Configure.pl#L342 could this TODO be implemented now?

lizmat:
After this, MacOS fails to build:

     clang  -install_name @rpath/libmoar.dylib   -fno-omit-frame-pointer -fno-optimize-sibling-calls -O3 -DNDEBUG -Wno-logical-op-parentheses -D_DARWIN_USE_64_BIT_INODE=1  -fno-omit-frame-pointer -fno-optimize-sibling-calls -O3 -DNDEBUG -Wno-logical-op-parentheses -O3 -DNDEBUG  -O3 -DNDEBUG  -lpthread -Wl,-rpath,@executable_path/../lib -I/Users/liz/Github/rakudo.moar/install/include -I/Users/liz/Github/rakudo.moar/install/include/moar \
	    -I/Users/liz/Github/rakudo.moar/install/include/libatomic_ops -I/Users/liz/Github/rakudo.moar/install/include/dyncall -I/Users/liz/Github/rakudo.moar/install/include/sha1 -I/Users/liz/Github/rakudo.moar/install/include/tinymt -I/Users/liz/Github/rakudo.moar/install/include/libtommath -I/Users/liz/Github/rakudo.moar/install/include/libuv \
	    -L/Users/liz/Github/rakudo.moar/install/lib  -o perl6-mc src/vm/moar/runner/main.c -lmoar 
clang: error: invalid argument '-install_name @rpath/libmoar.dylib' only allowed with '-dynamiclib'

lizmat:
Adding `-dynamiclib` seems to fix the build, but still anything that uses "make" completely fails, such as "make test" and "make spectest"

patzim:
I suspect the -dynamiclib thing is a version incompatibility in clang. Hopefully adding it won't break the build on other clang versions. Can you post the error you receive when running `make test` or so? Also did moar actually build successfully? The compilation of the moar executable and perl6-mc executable should be basically identical.

lizmat:
The "make" failures disappeared after having run "make install" with the `-dynamiclib` added.

PR #2701, opened 2019-02-17
Cleanup Configure.pl, etc and add some JS support

MasterDuke17:


MasterDuke17:
@ugexe thanks. They didn't seem like very controversial changes, but I'd never touched this part of the codebase before.

Issue #2714, opened 2019-02-22
Role inheriting from a class fails type checking against that class


vrurg:
## The Problem

When a role defines inheritance from a class type checking against that class doesn't pass.

## Steps to Reproduce

```
class C { }
role R is C { }
say R ~~ C;
```

outputs

>```
>False
>```

## Notes

The fix should be a simple patch of role metaclasses `type_check` methods.

## Environment
* Operating system: 
* Compiler version (`perl6 -v`):

jnthn:
A role doesn't really "inherit from a class", it just stashes away the parent and then adds it as a parent to a class that it is eventually composed in to. It's *probably* harmless to patch the `type_check` method if we want to try and pretend a bit harder. Note that it will have an impact on multi dispatch candidate sorting, although I suspect this construct is used rarely enough (I can think of zero uses of it in anything I've written...) that it won't have many consequences.

vrurg:
As to the terminology – that's easier to say 'it inherits'. From user point of view – it actually does. So, why not? Also, for a user it looks reasonable if role would pass the check.

I'll try and see if it passes spectest.

vrurg:
Closed via #2717

PR #2717, opened 2019-02-22
Fix for #2714

vrurg:
This fix would make a role pass typechecking against its parent classes.

It is also changes the default logic of `Metamodel::ParametricRoleGroupHOW` with regard to its assumed default role for introspection. Where previously it was the first candidate it would now use a non-signatured role if such exists and only fall back to the first candidate otherwise.

Another side-effect of this PR is a fix for matching roles with `;;` in their signatures. For example, in excerpt from `S14/roles/parameterized-mixin.t`:

```
role InitialAttribBoth[Str $type;; Str $name] {
    has $.type = $type;
    has $.name = $name;
 }

ok $d.HOW.does($d, InitialAttribBoth["type1"]),
  ".HOW.does gives correct information (4-2)";
```

rakudo fails with:

```
===SORRY!===
No appropriate parametric role variant available for 'InitialAttribBoth'
```

(_TODO_ should also provide file/line of the error source).

Ticket: rakudo/rakudo#2714 
Suporting tests: perl6/roast#514


vrurg:
Requesting for review due to non-backward compat behavior of role parameterization.

Issue #2724, opened 2019-03-01
(s)printf misleading error message should report values, not types


Altreus:
`printf` should report the value when it could not be applied to a directive; currently, it reports the *type* of the value.

```
> printf("%d",Inf)   
Directive d not applicable for type Num
```

This is misleading when the type is expected to work, but can have unworkable values; in the above example, it is surprising to hear that `Num` does not apply to `d`, whereas in fact it is `Inf` or `NaN` that don't apply to `d`.

Another misleading case is `Str`, since again it is the value that is unusable, not the type:

```
> printf("%d", "1")
1

> printf("%d", "raku")
Directive d not applicable for type Str
```

lizmat:
Thanks for the report.  This is now fixed!

PR #2726, opened 2019-03-02
Add given value to (s)printf bad-directive-type error

MasterDuke17:
which is `X::Str::Sprintf::Directives::BadType`. Requires https://github.com/perl6/nqp/pull/521. Rakudo builds ok and passes `make m-test m-spectest`. Fixes https://github.com/rakudo/rakudo/issues/2724.


lizmat:
Thank you!

PR #2728, opened 2019-03-02
Fix for SHORT-GIST dying with a non-Any inheriting parameter type

vrurg:
If `SHORT-GIST` method of `Rakudo::Internals` gets a non-Any inheriting object as a parameter it dies. As the method is used by exceptions, sometimes it results in obscuring the original exception/error.


lizmat:
Thank you!

PR #2732, opened 2019-03-04
Add basic `emulates` support to CURs

ugexe:
This adds basic support for the META6 field `emulates` as described by https://design.perl6.org/S22.html#emulates

### DIFF
--- a/src/core/CompUnit/Repository/Installation.pm6
+++ b/src/core/CompUnit/Repository/Installation.pm6
@@ -62,7 +62,7 @@ sub MAIN(:$name, :$auth, :$ver, *@, *%) {
     method !bin-dir       { with $.prefix.add('bin')       { once { .mkdir unless .e }; $_ } }
     method !short-dir     { with $.prefix.add('short')     { once { .mkdir unless .e }; $_ } }
 
-    method !add-short-name($name, $dist, $source?, $checksum?) {
+    method !add-short-name($name, $dist, $source?, $checksum?, $lookup-name?) {
         my $id = nqp::sha1($name);
         my $lookup = self!short-dir.add($id) andthen { .mkdir unless .e }
         $lookup.add($dist.id).spurt(
@@ -71,6 +71,7 @@ sub MAIN(:$name, :$auth, :$ver, *@, *%) {
             ~   "{$dist.meta<api>  // ''}\n"
             ~   "{$source // ''}\n"
             ~   "{$checksum // ''}\n"
+            ~   "{$lookup-name // $name // ''}\n"
         );
     }
 
@@ -243,6 +244,15 @@ sub MAIN(:$name, :$auth, :$ver, *@, *%) {
             $handle.close;
         }
 
+        # add `emulates` short-name lookups
+        for $dist.meta<emulates>.kv -> $provide-name, $emulate-name is copy {
+            my $id       = self!file-id($provide-name, $dist-id);
+            my $handle   = $dist.content($dist.meta<provides>{$provide-name});
+            my $content  = $handle.open(:bin).slurp(:close);
+            my $checksum = nqp::sha1(nqp::join("\n", nqp::split("\r\n",$content.decode('iso-8859-1'))));
+            self!add-short-name($emulate-name, $dist, $id, $checksum, $provide-name);
+        }
+
         my %meta = %($dist.meta);
         %meta<files>    = %links;    # add our new name-path => content-id mapping
         %meta<provides> = %provides; # new meta data added to provides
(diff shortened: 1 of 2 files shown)


ugexe:
This still lacks the ability to parse and understand the e.g. `:auth<cpan:MLEHMANN>` of `JSON::XS:auth<cpan:MLEHMANN>`

Tests: https://github.com/perl6/roast/commit/2ae093b893d54a594e46575395e4b56a2e81d711

src/core/CompUnit/Repository/Installation.pm6
```diff
@@ -430,10 +441,10 @@ sub MAIN(:$name, :$auth, :$ver, *@, *%) {
         }
 
         # Sort from highest to lowest by version and api
-        my $sorted-metas := $matching-metas.sort(*.value<api>).sort(*.value<ver>).reverse;
+        my $sorted-metas := $matching-metas.sort({ .value<name> eq $spec.short-name }).sort(*.value<api>).sort(*.value<ver>).reverse;
 
         # There is nothing left to do with the subset of meta data, so initialize a lazy distribution with it
-        my $distributions := $sorted-metas.map(*.kv).map: -> ($dist-id, $meta) { self!lazy-distribution($dist-id, :$meta) }
+        my $distributions = $sorted-metas.map(*.kv).map: -> ($dist-id, $meta) { self!lazy-distribution($dist-id, :$meta) }
```
niner: Was this intentional?

src/core/CompUnit/Repository/Installation.pm6
```diff
@@ -430,10 +441,10 @@ sub MAIN(:$name, :$auth, :$ver, *@, *%) {
         }
 
         # Sort from highest to lowest by version and api
-        my $sorted-metas := $matching-metas.sort(*.value<api>).sort(*.value<ver>).reverse;
+        my $sorted-metas := $matching-metas.sort({ .value<name> eq $spec.short-name }).sort(*.value<api>).sort(*.value<ver>).reverse;
```
niner: Won't this give warnings about uninitialized `.value<name>`?

src/core/CompUnit/Repository/Installation.pm6
```diff
@@ -243,6 +244,15 @@ sub MAIN(:$name, :$auth, :$ver, *@, *%) {
             $handle.close;
         }
 
+        # add `emulates` short-name lookups
+        for $dist.meta<emulates>.kv -> $provide-name, $emulate-name is copy {
+            my $id       = self!file-id($provide-name, $dist-id);
+            my $handle   = $dist.content($dist.meta<provides>{$provide-name});
+            my $content  = $handle.open(:bin).slurp(:close);
+            my $checksum = nqp::sha1(nqp::join("\n", nqp::split("\r\n",$content.decode('iso-8859-1'))));
```
niner: Can we get rid of the duplicated code here?

ugexe: The only thing I can think you mean is slurping the handle as `iso-8859-1`, but the file could be in any encoding (not necessarily the single encoding used throughout the code base to generate a checksum).

src/core/CompUnit/Repository/Installation.pm6
ugexe: I suppose for upgraded (instead of freshly installed) rakudo yes. `name` is a new field added to the short lookup:
https://github.com/rakudo/rakudo/blob/02c46a96b403c73d9fef68f99db5fd8a1bfde6ab/src/core/CompUnit/Repository/Installation.pm6#L424

I'll need to add some logic to `upgrade-repository` to add the extra field in the short-name-lookup file. Since https://github.com/rakudo/rakudo/pull/2778 also need to upgrade the repository version it would probably be wise for these two PRs to be rolled into one once the details are ironed out.

src/core/CompUnit/Repository/Installation.pm6
ugexe: I'm not sure what you mean -- `.candidates()` return an `Array of Distribution`, not an `Array of Hash`

niner:
Yes, I meant the slurp file, decode and hash it dance. It doean't actually matter what encoding the file is in. I used latin-1 because it will keep the bytes as they are. That wouldn't be necessary if nqp::sha1 could take a blob instead of a string. I hope we can get rid of the workaround at some point which is why it'd be good to not spread it around too much

src/core/CompUnit/Repository/Installation.pm6
JJ: Did you?

ugexe: Yes. Did you have a concern about this change?

JJ: Not really, I just saw this stuck conversation, thought about pinging in the only issue that seemed unsolved.

ugexe:
Why was this closed? I see other PRs with conflicts are left open, so I assume it cant be that...

lizmat:
@ugexe I incorrectly assumed the PR was dead.  I will look at resolving the conflicts soon, now that it's clear the PR is not dead.

lizmat:
Argh, it appears that my commitment to solve the conflicts, has fallen through the cracks.  @ugexe do you still want me to do that?

lizmat:
Is this PR still salvageable / worth saving?

PR #2734, opened 2019-03-04
Speed up Mu.say

scovit:
By using the IOHandler implementation of say, instead of print

### DIFF
--- a/src/core/io_operators.pm6
+++ b/src/core/io_operators.pm6
@@ -11,7 +11,7 @@ proto sub say(|) {*}
 multi sub say() { $*OUT.print-nl }
 multi sub say(\x) {
     my $out := $*OUT;
-    $out.print(nqp::concat(nqp::unbox_s(x.gist),$out.nl-out));
+    $out.say(x);
 }
 multi sub say(**@args is raw) {
     my str $str;


lizmat:
Good idea.  However, I'm afraid that we only force a `print` method to be implemented by any class to be compatible with `IO::Handle`.  So we can only do this if the actual object is actually an `IO::Handle`.

MasterDuke17:
Would something like `nqp::if(nqp::eqaddr($out, IO::Handle), $out.say(x), $out.print(nqp::concat(nqp::unbox_s(x.gist),$out.nl-out)))` introduce too much overhead?

lizmat:
testing something like that now

lizmat:
A check for with `nqp::can(...,'say')` makes things about 4% faster even with the additional overhead.

lizmat:
Ah, but alas that breaks spectest :-(

MasterDuke17:
Huh, what breaks?

scovit:
what about `nqp::if($out ~~ IO::Handle, $out.say(x), $out.print(nqp::concat(nqp::unbox_s(x.gist),$out.nl-out)))` ?

lizmat:
yeah, that's the one I'm working on...  Except it will have to be `$out.say(x.gist)` as we have spectests that check whether `.say` calls the `.gist` method.

lizmat:
Approach with checking for `IO::Handle` makes it about 8% faster than the original.  Checking for spectest failures now.  Meanwhile, the most CPU intensive part is still the looking up of `$*OUT`.  You can get almost 2x as fast with:

    yes | head -n1000000 | time perl6 --profile -e 'my $o = $*OUT; for lines() { $o.say($_) }' > /dev/null

Which makes it only lookup $*OUT once, rather than 1M times.  This makes it about 1.9x as fast as the original benchmark.

scovit:
Thank you @lizmat, moving the dynamic look-up out of the loop seems like a quite complex optimization for the compiler. Regarding the `IO::Handle` optimization it seems to me that the win is in doing `$out.nl-out` inside $out vs outside, do you agree? Is it something to expect?

lizmat:
Implemented with 9770c1aa15 .  As to $out.nl-out, in the `IO::Handle` case, it is a private attribute, so that saves one method call.

Closing this PR.  Thanks for the spot!

PR #2737, opened 2019-03-04
Speedup ARGFILES read when there is only one input

scovit:
It is common that ARGFILES consists of only one file, such as when working with pipes, this introduces a fast path for that specific condition.

### DIFF
--- a/src/core/IO/CatHandle.pm6
+++ b/src/core/IO/CatHandle.pm6
@@ -134,10 +134,12 @@ my class IO::CatHandle is IO::Handle {
     method !WORDS {
       nqp::if(
         nqp::defined($!active-handle),
-        (flat $!active-handle.words, gather {
-          nqp::while(
-            nqp::defined(self.next-handle),
-            take $!active-handle.words)}),
+          nqp::if(nqp::isfalse(nqp::elems($!handles)), # Only 1 handle?
+            $!active-handle.words,
+            (flat $!active-handle.words, gather {
+              nqp::while(
+                nqp::defined(self.next-handle),
+                take $!active-handle.words)})),
         Seq.new(Rakudo::Iterator.Empty)
       )
     }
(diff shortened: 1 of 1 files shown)


scovit:
This is the benchmark:
```
$ yes | head -n2000000 | time ./perl6 -e 'my $o = $*OUT; for lines() { $o.say($_) }' > /dev/null
```

lizmat:
What would be against removing the `nqp::isfalse` and reversing the alternatives?

scovit:
I wonder if the fact that this patch is needed does not point to a general slowness and an opportunity for optimisation of `flat`, or if it is not possible to envision an alternative Iterator that concatenate a list of N iterators and optimise in case of N=1..

scovit:
The fix is wrong if we consider why (as far as I understand) there is the need of a `CatHandle` at all, see
#2796

JJ:
First, there are conflicts; second, there does not seem to be a clear need for this, or it's not clear what its effect could be. In the second case, it's probably best to take it over to the problem solving repo. What do you think?

scovit:
Hi JJ. it has been quite awhile. I really don't know if there are still issues.

Could you please run the following benchmarks on Rakudo head?

yes | head -n2000000 | time raku -e 'for lines() {  }' > /dev/null
        2.08 real         2.02 user         0.19 sys
yes | head -n2000000 | time raku -ne ' ' > /dev/null
        3.84 real         4.33 user         0.07 sys
yes | head -n2000000 | time raku -e 'for $*IN.lines() {  }' > /dev/null
        0.72 real         0.67 user         0.09 sys
raku -v
This is Rakudo version 2019.11-506-g06ae7d614 built on MoarVM version 2020.01-17-gefb014f19
implementing Perl 6.d.

MasterDuke17:
FWIW, I just ran those on my system and get identical times for all three.

jnthn:
>

scovit:
This means that the most common usage case has been optimized, good news!

scovit:
@JJ maybe one could open a problem solving issue for the optimization of flat, if it is still as slow as it was before

scovit:
Should the following: 
https://github.com/rakudo/rakudo/issues/2796

also be moved to problem solving?

JJ:
El lun., 4 may. 2020 a las 12:04, Vittore F. Scolari (<
notifications@github.com>) escribió:

> @JJ <https://github.com/JJ> maybe one could open a problem solving issue
> for the optimization of flat, if it is still as slow as it was before
>
Please do that. I'm not aware of all the context.

jnthn:
Almost certainly was handled by the changes I did in response to #3545.

Issue #2178, opened 2018-08-04
.wrap of a method breaks callsame/nextsame/etc.


vrurg:
## The Problem

Wrapping a method in a child class breaks MRO chaining for callsame/nextsame. Consider the following code:

```
     class Foo {
         has Int $.attr is rw;

         method m {
             say "<this is Foo:m>";
             callsame;
         }
     }

     class Bar is Foo {
         method m {
             say "<this is Bar::m>";
             callsame;
         }
     }

     class Baz is Bar {
         method m {
             say "<this is Baz::m>";
             callsame;
         }
     }

     Baz.^lookup( 'm' ).wrap( method { say "<this is WRAPPER:m>"; nextsame } );

     my $inst = Baz.new;

     $inst.m;
``` 

It produces the following output:

```
<this is WRAPPER:m>
<this is Baz::m>
```

Method `m` of class `Baz` cannot call the next method in MRO chain.

There is even worse case scenario if wrap is applied to `Bar::m` method. Replacing the wrapping line in the source with the following:

```
Bar.^lookup( 'm' ).wrap( method { say "<this is WRAPPER:m>"; nextsame } );
```

gives us:

```
<this is Baz::m>
Cannot invoke object with invocation handler in this context
```

## Environment
* Operating system: macOS 10.13
* Compiler version (`perl6 -v`): This is Rakudo version 2018.06 built on MoarVM version 2018.06

vrurg:
There is interesting development for the situation when the method in `Bar` is wrapped. Adding a call to `nextcallee` into `Baz::m`:

```
     class Baz is Bar {
         method m {
             say "<this is Baz::m>";
             my &nc = nextcallee;
             #self.&nc;
             callsame;
         }
     }
```

avoid the exception about invocation handler. Instead the following output is produced:

```
<this is Baz::m>
<this is Foo:m>
```

It looks like this time `callsame` just got around `Bar::m` by simply ignoring it! By uncommenting the `self.&nc` call I get the expected output:

```
<this is Baz::m>
<this is WRAPPER:m>
<this is Bar::m>
<this is Foo:m>
```

but the point that it is produced by two separate calls. I.e., `callsame` from `Bar::m` did not pass the control to `Foo`, but from `Baz::m` did!

Now, the final trick: leave the `self.&nc` but comment out `callsame` in `Baz::m`. It raises another exception:

```
<this is Baz::m>
Cannot invoke this object (REPR: Uninstantiable; Callable)
[...]

PR #2749, opened 2019-03-07
Fix for rakudo/rakudo#2178

vrurg:
The fix is done by implementing cooperation between dispatchers. If a
dispatcher finds out that next call in the queue is a invocation handler
with attribute `$!dispatcher` containing a object of descendant type of
`Metamodel::BaseDispatcher` it passes control over to that dispatcher by
calling either `enter_with_args` or `enter_with_capture` method. The
subsequent dispatcher passes control back to the calling dispatcher upon
exhausting its own queue.

The known problems of this fix:

- there is no clear way to determine if next call is a invocation
handler. The approach used by this fix is to wrap the attempt to read
from `$!dispatcher` attribute into a try block is hacky and would have
performance penalty. Must be changed as soon as more straightforward
approach is developed or found.

#2178


vrurg:
A problem has been discovered which needs fixing before this PR is ready.

vrurg:
The problem has been resolved, PR is ready for review.

src/Perl6/Metamodel/Dispatchers.nqp
```diff
@@ -1,32 +1,64 @@
...
+    has $!next_dispatcher; # The dispatcher we must pass control to when own queue exhausts
 
     method candidates() { @!candidates }
 
-    method exhausted() { $!idx >= +@!candidates }
+    method exhausted() { $!idx >= +@!candidates && (!nqp::isconcrete($!next_dispatcher) || $!next_dispatcher.exhausted()) }
 
     method last()      { @!candidates := [] }
 
+    method set_next_dispatcher ($next_dispatcher) { $!next_dispatcher := $next_dispatcher }
```
MasterDuke17: Completely unrelated to what this PR actually does, but you have a couple methods with a space between the method name and the signature.

vrurg: My Perl6 habbit. Will fix now. Thanks!

PR #2754, opened 2019-03-09
Allow :at/:in/:every to be Inf/-Inf/NaN in ThreadPoolScheduler.cue

Kaiepi:
Before, this would throw with a somewhat cryptic error.
Now, this uses the smallest time interval allowed for the parameter.


lizmat:
Thanks!

PR #2770, opened 2019-03-17
Refactor $*KERNEL internal use of uname

ugexe:
Uses `nqp::uname()` on MoarVM instead of shelling out, while shelling out on other backends.

Changes .release() and .version() to return more appropriate values -- uname version instead of uname release for .version(), and uname release instead of uname version for .release() 😕 By making these consistent with their uname counterpart people can better understand the values it may contain and by what name.

### DIFF
--- a/src/core/Kernel.pm6
+++ b/src/core/Kernel.pm6
@@ -8,11 +8,53 @@ class Kernel does Systemic {
     has Str $!hardware;
     has Str $!arch;
     has Int $!bits;
-    has Bool $!has_uname;
 
-    method !uname($opt) {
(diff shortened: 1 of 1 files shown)


ugexe:
It would be nice if we could make `.version()` return a Str instead of a Version. If anything `.release()` should be a Version, but even then it doesn't entirely make sense to create the version object since it doesn't make sense to use the Version comparison operators on strings that may not order themselves as Version expects.

lizmat:
Thank you for your work and persistence  :-)

lizmat:
Perhaps it makes sense to create a StrVersion allomorph?

PR #2778, opened 2019-03-21
Retain original provides metadata structure

ugexe:
Installing a distribution would change `provides => { a => "a.pm6" }`
to something like `provides => { a => { "a.pm6" => { ... } } }`,
with the purpose of saving some extra meta data about the file.

This removes the need for the additional meta data, and makes
installation no longer modify the provides structure. In the
future we may find the meta data useful again, but it should
be put into its own top level field.

### DIFF
--- a/src/core/CompUnit/Repository/Installation.pm6
+++ b/src/core/CompUnit/Repository/Installation.pm6
@@ -393,6 +393,9 @@ sub MAIN(:$name, :$auth, :$ver, *@, *%) {
         ));
     }
     multi method candidates(CompUnit::DependencySpecification $spec) {
+        my $repo-version = self!repository-version;
+        self.upgrade-repository unless $repo-version == 3;
+
         return Empty unless $spec.from eq 'Perl6';
 
         # $lookup is a file system resource that acts as a fast meta data lookup for a given module short name.
--- a/src/core/CompUnit/Repository/Installation.pm6
+++ b/src/core/CompUnit/Repository/Installation.pm6
@@ -11,13 +11,18 @@ class CompUnit::Repository::Installation does CompUnit::Repository::Locally does
 
     my $verbose := nqp::getenvhash<RAKUDO_LOG_PRECOMP>;
 
+    my sub file-id(Str $name, Str $dist-id) {
+        nqp::sha1($name ~ $dist-id)
+    }
+
     submethod BUILD(:$!prefix, :$!lock, :$!WHICH, :$!next-repo --> Nil) { }
 
     my class InstalledDistribution is Distribution::Hash {
+        has $.dist-id;
         method content($address) {
-            my $entry = $.meta<provides>.values.first: { $_{$address}:exists };
+            my $entry = $.meta<provides>.antipairs.hash{$address};
             my $file = $entry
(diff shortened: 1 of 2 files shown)


src/core/CompUnit/Repository/Installation.pm6
```diff
@@ -11,13 +11,18 @@ class CompUnit::Repository::Installation does CompUnit::Repository::Locally does
...
+        nqp::sha1($name ~ $dist-id)
+    }
+
     submethod BUILD(:$!prefix, :$!lock, :$!WHICH, :$!next-repo --> Nil) { }
 
     my class InstalledDistribution is Distribution::Hash {
+        has $.dist-id;
         method content($address) {
-            my $entry = $.meta<provides>.values.first: { $_{$address}:exists };
+            my $entry = $.meta<provides>.antipairs.hash{$address};
```
niner: Does this actually work? We only upgrade repositories on write. So at this point we can still get the old meta data format.

ugexe: We upgrade repositories during `make install` too.

https://github.com/rakudo/rakudo/blob/68c2c9fea187fcdae33e841ae8a882d722e4804d/tools/build/Makefile-Moar.in#L330

niner:
Upgrading during make install won't affect repos from a packaged rakudo and home or custom repos

ugexe:
I would expect packaged rakudos to follow the upgrade repository procedure, even if they don't currently do so. existing `home` and `custom` repos are problematic. But at this point in time I expect the actual problems caused to people to be negligible, especially compared to the alternative (adding runtime checks whens searching to possible trigger repo upgrades).

ugexe:
I suppose the various `Distribution` could have their constructor or `.meta` modified to return what "looks" like the original. But I'm not exactly a fan of this.

niner:
On Saturday, 23 March 2019 18:59:29 CET Nick Logan wrote:
> I would expect packaged rakudos to follow the upgrade repository procedure,
> even if they don't currently do so.

On openSUSE it certainly does not. Actually in the case of the site repo, 
upgrading to a rakudo with version 3 repositories will break right now. Need 
to find a good solution for that. The vendor and perl repositories are not a 
problem, since the module packages will be rebuilt anyway.

> existing `home` and `custom` repos are
> problematic. But at this point in time I expect the actual problems caused
> to people to be negligible

I disagree. The code would after all return false data when run on a version 2 
repository, wouldn't it? I know that at work we have a couple of cases where 
modules are loaded out of the (service) user's home directory.

ugexe:
I'm not sure that in reality this really breaks anything for existing repositories of modules since the json should not be getting read period expect in some very specific circumstances (that didn't really exist until the last release and the unification of Distribution handling between CURs).

niner:
Ah ok, then it's much less of an issue indeed. Still need to figure out how to deal with repo version changes in a packaged rakudo though. Post install scripts are very much frowned upon and for good reason. I guess the rakudo package should not contain anything but the "site" directory itself but no contents at all.

lizmat:
I feel this PR is in need of some love, or needs to be abandoned.

Issue #2772, opened 2019-03-19
Multi-dispatch chooses wrong proto during compile-time execution


vrurg:
## The Problem

When the following conditions are met:

- class 'Foo' defines a `proto method foo`
- class `Bar` is declared in the same compunit and inherits from `Foo`
- `Bar` defines a `multi method foo`

In this case  if the `multi` method `foo` would be called on a instance of `Bar` during compilation time rakudo will chose the `proto` `method` from `Foo` resulting in use of wrong `MultiDispatcher` instance and, correspondingly, ignoring the `multi` variant from `Bar`.

## Steps to Reproduce

The problem is best demonstrated when a module file is used. For example, the following archive contains _multi-method.pm6_ and _multi-method.p6` files. Both are instantiating `Bar` class and calling its `foo` method in `BEGIN` blocks while _.p6_ also does it during normal run time.

[multi-method.zip](https://github.com/rakudo/rakudo/files/2985713/multi-method.zip)

To ease describing the problem, I'll duplicate their code here:

_multi-method.pm6_
```
     unit module multi-method;
     no precompilation;

     class Foo is export {
         proto method foo (|) {*}
         multi method foo (Mu \o) { say "foo(Mu)" }
     }

     class Bar is Foo is export {
         multi method foo (Any \o) { say "foo(Any)" }
     }

     BEGIN {
         say "precomp.";
         Bar.new.foo(Str);
     }
```

_multi-method.p6_
```
     #!/usr/bin/env perl6
     use lib '.';
     use multi-method;

     BEGIN {
         say ".p6 BEGIN";
         Bar.new.foo(Str);
     };

     Foo.new.foo(Str);
     Bar.new.foo(Str);
```

_multi-method.p6_ would generate the following output:

>```
>precomp.
>foo(Mu)
>.p6 BEGIN
>foo(Mu)
>foo(Mu)
>foo(Mu)
>```

Note the `no precompilation;` line. Removing it results in completely different picture. The first run when there is no precompiled code for the module results in the following output:

>```
>precomp.
>foo(Mu)
>.p6 BEGIN
>foo(Any)
>foo(Mu)
>foo(Any)
>```

Obviously, any following execution would only have the part of the output which starts with '.p6 BEGIN` line.

# Environment
* Operating system: macOS, Linux
* Compiler version (`perl6 -v`): This is Rakudo version 2019.03 built on MoarVM version 2019.03

vrurg:
[...]

PR #2780, opened 2019-03-22
Fixed method cloning

vrurg:
Fix for rakudo/rakudo#2772

Two changes:

- Stub is now passing the call to precompiled code only if current code object is unavailable. 
- Method code objects were not added to the list of clones previously.


lizmat:
Thanks!

Issue #2786, opened 2019-03-23
Incorrect error message produced when numeric (Real) operators are applied to a numeric type object


vrurg:
## The Problem

The problem is demonstrated best with the following code snippet:

```
my Int $i;
say $i > 0;
```

Rakudo outputs:

>```
>Invocant of method 'Bridge' must be an object instance of type 'Int', not a type object of type 'Int'.  Did you forget a '.new'?
>```

## Expected Behavior

The message should be:

>```
>Use of uninitialized value of type Int in numeric context
>```

## Environment
* Operating system: any
* Compiler version (`perl6 -v`): This is Rakudo version 2019.03 built on MoarVM version 2019.03
also tested on `master`

vrurg:
This is caused by signatures of operators in _Real.pm6_ using `Real` type without `:D`. I'll produce a PR for this.

AlexDaniel:
Reopened, see GH-3051.

JJ:
Not sure about this one. It's reopened because an issue that has actually been closed. Where does that leave this one?

PR #2790, opened 2019-03-23
Make all operators in Real require definite objects

vrurg:
Fix for rakudo/rakudo#2786. As a side effect, this changes the way
rakudo complains about the use of type objects. Where previously it was
throwing an exception it is now warns about use of undefined value:

>```
>Use of uninitialized value of type Int in numeric context
>```

This is to conform with similar behavior of other types like `Str`, `Date`,
`Version`. Also, I would refer to this statement:

https://colabti.org/irclogger/irclogger_log/perl6?date=2011-07-03#l554

### DIFF
--- a/src/core/Real.pm6
+++ b/src/core/Real.pm6
@@ -132,42 +132,42 @@ my role Real does Numeric {
 }
 
 proto sub cis($, *%) {*}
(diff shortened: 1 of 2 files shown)


lizmat:
Thanks!

PR #2800, opened 2019-03-27
Fix for FatRat.base failure for large denominators 

thundergnat:
FatRat.base conversion with denominators larger than 2 ** 1023 (~10 ** 308) falls into an infinite sequence calculation. The problem is in the `$!denominator.log($base).ceiling + 1` code on line 213.  https://github.com/rakudo/rakudo/blob/master/src/core/Rational.pm6#L213 -  log of integers > 10 ** 308 or so returns Inf, and it tries to calculate the fraction to infinite precision. This pull request works around that and has a few other minor speed-up tweaks. Spectest has some failures but they aren't related to FatRats, base conversion or stringification so I think this is ok to merge. See https://www.nntp.perl.org/group/perl.perl6.users/2019/03/msg6640.html for bug report.


src/core/Rational.pm6
```diff
@@ -236,16 +245,17 @@ my role Rational[::NuT = Int, ::DeT = ::("NuT")] does Real {
 
         # Round the final number, based on the remaining fractional part
         if 2*$fract >= 1 {
-            for @fract-digits-1 ... 0 -> $n {
+            for @fract-digits - 1 ... 0 -> $n {
```
MasterDuke17: I don't know how big `@fract-digits` gets, but `.reverse` (e.g., `(^@fract-digits).reverse`) is much faster than a decreasing `...`.

src/core/Rational.pm6
```diff
@@ -210,16 +217,18 @@ my role Rational[::NuT = Int, ::DeT = ::("NuT")] does Real {
             }
         }
         else {
-            $prec = ($!denominator < $base**6 ?? 6 !! $!denominator.log($base).ceiling + 1);
+            my $lim = 10**307;
```
MasterDuke17: Could you add a comment about where `10**307` comes from?

src/core/Rational.pm6
thundergnat: In general, yes. In this specific case, it is exceedingly rare that it iterates more than 3-4 times before short circuiting and exiting so the iterator only needs to produce 3-4 elements. '(^@fract-digits).reverse)' will require that all the elements be generated every time. In testing, '@fract-digits - 1 ... 0' is slightly faster most of the time.

src/core/Rational.pm6
thundergnat: Comments added

lizmat:
Thanks!

PR #2819, opened 2019-04-06
Improve ThreadPoolScheduler.cue/CurrentThreadScheduler.cue behaviour when Inf/-Inf/NaN is passed

Kaiepi:
- If `Inf` is passed to `:at`/`:in`/`:every`, the method will never invoke the
block passed, but will return a `Cancellation` anyway
- If `-Inf` is passed, the method will invoke the block as soon as
possible and return a `Cancellation`
- If `NaN` is passed, the method will throw an exception

This also ensures a Cancellation is always returned from the dispatchers
for `.cue` that are supposed to be returning them, since calling it with
`:every` as an argument would sometimes return a `List` instead.

This needs unit tests and an update to the docs before merging

### DIFF
--- a/src/core/CurrentThreadScheduler.pm6
+++ b/src/core/CurrentThreadScheduler.pm6
@@ -25,7 +25,7 @@ my class CurrentThreadScheduler does Scheduler {
                     ($delay := 0),
                     nqp::if(
                         nqp::isnanorinf($delay),
-                        (die "Cannot set NaN as a number of seconds")
+                        X::Scheduler::CueInNaNSeconds.new().throw()
                     )
                 )
             )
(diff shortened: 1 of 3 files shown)


Kaiepi:
I forgot to handle the edge cases for `CurrentThreadScheduler` as well. I'll deal with that tomorrow

timo:
I would expect `every => Inf` to execute the code once in total, especially since the docs say `every` is about `re-executing` at a given interval

Kaiepi:
That seems reasonable. I'll update it in a minute

Kaiepi:
I should probably make a `X::Scheduler::NaNSeconds` exception as well so others implementing `Scheduler` don't need to know what message to die with

Kaiepi:
Should be pretty well done now

lizmat:
Thank you!

PR #2821, opened 2019-04-06
Use sh instead of bash

patzim:
Replaces all uses of bash with sh to be more portable. This relates to https://github.com/perl6/nqp/issues/528

### DIFF
--- a/README.md
+++ b/README.md
@@ -121,10 +121,10 @@ run the spectests.  For example:
 
     $ make t/spec/S12-traits/parameterized.t
 
-Run all tests in one S* directory with a bash script. One example:
+Run all tests in one S* directory with a sh script. One example:
 
     $ cat run-tests.sh
-    #!/bin/bash
+    #!/bin/sh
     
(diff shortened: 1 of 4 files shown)


AlexDaniel:
```bash
target=$(command ls -l "$fname")
```
Even though ls itself is fragile, what about adding `--` before `"$fname"`?

patzim:
@AlexDaniel What does '--' do with ls? It's not in the man page and seems to produce output identical to `ls -l` without the '--'.

AlexDaniel:
@patzim same reason `--` is used everywhere else in the code. If your fname starts with `-`, then ls will do the wrong thing. Compare `ls -l -- -foo` and `ls -l -foo`.

I know that in general most kinda assume that filenames and paths are relatively sane, but the code here is consistently right with the use of quoting and `--` so that line looks kinda weird.

AlexDaniel:
Ah, also here after `cd`
```
command cd "$(command dirname -- "$target")"
```

PR #2822, opened 2019-04-07
fix error when set RAKUDO_EXCEPTIONS_HANDLER

wenjie1991:
The instance variable `what` of `Exception` class is native string type `str`. The list input `:what<RAKUDO_EXCEPTIONS_HANDLER env var>` will cause an error:
```
> Rakudo::Deprecations.DEPRECATED: "PERL6_EXCEPTIONS_HANDLER", Nil,'2019.07', :file("N/A"), :line("N/A"),:what<RAKUDO_EXCEPTIONS_HANDLER env var>;
This type cannot unbox to a native string: P6opaque, List
```

### DIFF
--- a/src/core/Exception.pm6
+++ b/src/core/Exception.pm6
@@ -380,7 +380,7 @@ do {
             # REMOVE DEPRECATED CODE ON 201907
             Rakudo::Deprecations.DEPRECATED: "PERL6_EXCEPTIONS_HANDLER", Nil,
                 '2019.07', :file("N/A"), :line("N/A"),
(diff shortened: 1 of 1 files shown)


wenjie1991:
The error causes a backward compatibility problem of vim Perl 6 Syntastic checker: <https://github.com/vim-syntastic/syntastic/pull/2253>.

lizmat:
This should probably need to be `:what("RAKUDO_EXCEPTIONS_HANDLER env var")`

wenjie1991:
Thanks, @lizmat, I will test your idea. Although the problem of Syntastic Perl 6 checker was solved, I will deal with it for potential affected users.

Issue #2826, opened 2019-04-08
Rounding behaviors


alabamenhu:
## The Problem

`.round` currently works on ~half-up~ away-from-zero principle for equidistant values.  Some applications of rounding need to use half-even (IEEE specifies this method, for instance)

## Expected Behavior

1.5.round == 2
2.5.round == 2
3.5.round == 4
4.5.round == 4

## Actual Behavior

1.5.round == 2
2.5.round == 3
3.5.round == 4
4.5.round == 5

## Steps to Reproduce

See above.

I'm not sure the best way to introduce it, as the current method is certainly the "default" interpretation in most people's mind and changing the default would be breaking.  Perhaps adding an adverb for rounding modes?

AlexDaniel:
May I ask why you need this? It'll help us understand why this needs to be part of the language. Somewhat related: https://github.com/perl6/problem-solving/issues/4

alabamenhu:
I'm currently working on an implementation of CLDR / ICU stuff, and the rounding is specified as being the default for binary floats.  This is also the default rounding method for Java, Python, Haskell, among others, since it's the default under IEEE 754.  For any implementations or ports, there's likely going to be an unstated assumption that the rounding mode is this way and to maintain equivalent results between languages there should be an easy way to maintain similar math results.  (I only noticed it because the Unicode's TR35 specifically mentioned that half-even rounding was the expected default, and upon testing realized that — even when using Num or native floats rounding is always up)

For most applications, not doing it won't cause a problem and of course a separate module can be written (and I'll probably do that in the meantime), but it seems like it'd be better handled at the core level — and it was suggested by tony-o on #perl6 to open a ticket for it.

tony-o:
If there isn't any objection, it wouldn't be difficult to add adverbs
half-[even|odd|down] or [away-from|toward]-zero to round while maintaining
the default half-up

alabamenhu:
> If there isn't any objection, it wouldn't be difficult to add adverbs half-[even|odd|down] or [away-from|toward]-zero to round while maintaining the default half-up

I should actually probably correct myself in that P6 is defaulting to an away-from-zero.  The other IEEE defined methods are toward-zero, toward-positive-infinity, toward-negative-infinity and half-even (but curiously, not half odd). 

toward-negative-infinity is a long adverb though.  Maybe use half-up/half-down for the infinities, and then have toward-zero / away-from-zero (longer, but also the default), and half-even / half-odd to gravitate to those numbers?

[...]

PR #2831, opened 2019-04-09
see:rakudo/rakudo#2826; add HALF-[UP|DOWN|EVEN|ODD] | [TO|FROM]-ZERO …

tony-o:
for the issue see:rakudo/rakudo#2826
for the related tests see:perl6/roast#527

this commit adds HALF-[UP|DOWN|EVEN|ODD] | [TO|FROM]-ZERO rounding options as adverbs on `.round` for `Real` and `Rat` and maintains the default `HALF-UP` when no adverb is applied.

### DIFF
--- a/src/core/Rational.pm6
+++ b/src/core/Rational.pm6
@@ -298,8 +298,81 @@ my role Rational[::NuT = Int, ::DeT = ::("NuT")] does Real {
             ?? $!numerator
             !! self;
     }
-
-    multi method round(::?CLASS:D: --> Int:D) {
+    multi method round(::?CLASS:D: :$HALF-DOWN! --> Int:D) {
+        return Failure.new(
+          X::Numeric::DivideByZero.new(
+            :details('when calling .round on Rational')
+          )
+        ) unless $!denominator;
+        my ($numerator, $denominator) = (
+          nqp::sub_I(nqp::mul_I($!numerator, 2, Int), $!denominator, Int),
+          nqp::mul_I($!denominator, 2, Int),
+        );
+        ceiling($numerator/$denominator);
+    }
+    multi method round(::?CLASS:D: :$HALF-EVEN! --> Int:D) {
+        return Failure.new(
+          X::Numeric::DivideByZero.new(
+            :details('when calling .round on Rational')
+          )
+        ) unless $!denominator;
+        if $!numerator mod $!denominator == $!denominator / 2 {
+          return floor(self) %% 2 ?? floor(self) !! ceiling(self);
+        }
+        self.round(:HALF-UP);
+    }
+    multi method round(::?CLASS:D: :$HALF-ODD! --> Int:D) {
+        return Failure.new(
+          X::Numeric::DivideByZero.new(
+            :details('when calling .round on Rational')
+          )
+        ) unless $!denominator;
+        if $!numerator mod $!denominator == $!denominator / 2 {
+          return floor(self) %% 2 ?? ceiling(self) !! floor(self);
+        }
+        self.round(:HALF-UP);
+    }
+    multi method round(::?CLASS:D: :$TO-ZERO! --> Int:D) {
+        return Failure.new(
+          X::Numeric::DivideByZero.new(
+            :details('when calling .round on Rational')
+          )
+        ) unless $!denominator;
+        my ($numerator, $denominator);
+        if $!numerator < 0 ^^ $!denominator < 0 { #it's a negative
+            ($numerator, $denominator) = (
+              nqp::add_I(nqp::mul_I($!numerator, 2, Int), $!denominator, Int),
+              nqp::mul_I($!denominator, 2, Int),
+            );
+            return floor($numerator/$denominator);
+        }
+        ($numerator, $denominator) = (
+          nqp::sub_I(nqp::mul_I($!numerator, 2, Int), $!denominator, Int),
+          nqp::mul_I($!denominator, 2, Int),
+        );
+        return ceiling($numerator/$denominator);
+    }
+    multi method round(::?CLASS:D: :$FROM-ZERO! --> Int:D) {
+        return Failure.new(
+          X::Numeric::DivideByZero.new(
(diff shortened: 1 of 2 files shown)


coke:
Why uppercase?

src/core/Rational.pm6
```diff
@@ -298,8 +298,81 @@ my role Rational[::NuT = Int, ::DeT = ::("NuT")] does Real {
             ?? $!numerator
             !! self;
     }
-
-    multi method round(::?CLASS:D: --> Int:D) {
+    multi method round(::?CLASS:D: :$HALF-DOWN! --> Int:D) {
```
ugexe: Required truth-only booleans need to be declared `Bool:D :$name! where .so`, not just `:$name!`

src/core/Rational.pm6
```diff
@@ -298,8 +298,81 @@ my role Rational[::NuT = Int, ::DeT = ::("NuT")] does Real {
...
+        ) unless $!denominator;
+        my ($numerator, $denominator);
+        if ($!numerator < 0 && $!denominator < 0) || ($!numerator > 0 && $!denominator > 0) { #it's a negative
+            ($numerator, $denominator) = (
+              nqp::add_I(nqp::mul_I($!numerator, 2, Int), $!denominator, Int),
+              nqp::mul_I($!denominator, 2, Int),
+            );
+            return floor($numerator/$denominator);
+        }
+        ($numerator, $denominator) = (
```
ugexe: These should be assigned to directly to each (or skip assignment altogether and just return the values from the following nqp ops) -- there is no need to do this array assignment thing.

Same with the other occurrences.

src/core/Real.pm6
```diff
@@ -38,9 +38,31 @@ my role Real does Numeric {
...
+                  ?? (self.floor %% 2 ?? self.floor !! self.ceiling)
+                  !! self.round(:HALF-UP));
+    }
+    multi method round(Real:D: :$HALF-ODD!) {
+        (self - self.floor == 1/2
+                  ?? (self.floor %% 2 ?? self.ceiling !! self.floor)
+                  !! self.round(:HALF-UP));
+    }
+    multi method round(Real:D: :$TO-ZERO!) {
+        (self + (self < 0 ?? 1/2 !! -1/2))."{self < 0 ?? 'floor' !! 'ceiling'}"();
```
ugexe: String interpolating methods is going to be slow and arguably avoided for math (or even core) stuff.

src/core/Rational.pm6
tony-o: Corrected in b4d3723

src/core/Rational.pm6
tony-o: Changed/resolved in b4d3723

src/core/Real.pm6
tony-o: Removed from `Real` in b4d3723

tony-o:
> Why uppercase?

This is changed to lc in b4d3723

AlexDaniel:
What happens if you provide more than one?

tony-o:
> What happens if you provide more than one?

@AlexDaniel It ignores everything after the first (similar behavior to today):

perl6 without this commit:
```
~/projects/rak $ perl6 -e 'say (1.5).round(:half-down, :half-up)'
2
```

perl6 with this commit:
```
./install/bin/perl6 -e 'say (1.5).round(:half-down, :half-up)'
1
```

AlexDaniel:
Hm. I'm asking because another way to do it is to have one `:$mode` named param and an enum with values like HALF-UP. Maybe another advantage of that is that we'll be able to reuse that enum elsewhere… but then, are there any other subroutines where rounding mode matters? I still feel like this should go through https://github.com/perl6/problem-solving/

tony-o:
> Hm. I'm asking because another way to do it is to have one `:$mode` named param and an enum with values like HALF-UP. Maybe another advantage of that is that we'll be able to reuse that enum elsewhere… but then, are there any other subroutines where rounding mode matters? I still feel like this should go through https://github.com/perl6/problem-solving/

Even with `:$mode` you'd be able to pass multiple arguments with the signature `round(::?Class:D: ...`.  I don't have an opinion about the suggestion, it reduces the number of signatures in the dispatcher having one one with `:$mode` but you have one rounding method with a long `if/then/else` chain.

I do agree this could use some discussion prior to merge.

src/core/Rational.pm6
```diff
@@ -298,8 +298,75 @@ my role Rational[::NuT = Int, ::DeT = ::("NuT")] does Real {
...
+          nqp::mul_I($!denominator, 2, Int).Int
+        );
+    }
+    multi method round(::?CLASS:D: Bool:D :$from-zero! where .so --> Int:D) {
+        return Failure.new(
+          X::Numeric::DivideByZero.new(
+            :details('when calling .round on Rational')
+          )
+        ) unless $!denominator;
+        my ($numerator, $denominator);
```
ugexe: These are not used anymore

tony-o: Ty fixed with 3787883

AlexDaniel:
I'm going through the list of PRs trying to resolve as many as possible. This PR is adding a new feature as requested in https://github.com/rakudo/rakudo/issues/2826. However, there was no decision to change the language yet, so this PR is just illustrative rather than something mergeable. Nowadays language changes should be requested in https://github.com/perl6/problem-solving/, so if anybody wants this issue to go faster they should move (recreate) the rakudo ticket there.

However, I don't feel comfortable with the proposed change at all.

> If there isn't any objection, it wouldn't be difficult to add adverbs

It's never difficult to add more features. Maintaining them, however, is sometimes extremely difficult. I'm not comfortable with the proposed change at all.

Anyway, I'll close this PR for now but leave the ticket open for more discussion. If we ever need it again it's possible to reopen this PR.

AlexDaniel:
Also, as far as I can see, this PR is far from being complete. It is only introducing changes for Rats but we also have Num.round, Complex.round and possibly more.

Issue #2825, opened 2019-04-08
Different behavior between Unicode and ASCII operators


frithnanth:
## The Problem

Using ASCII operators this works fine:
```
$ perl6 -e'my $a = BagHash.new: <a a a a b b b c c c c c d>;for $a.keys -> $k { say "$k => $a<<$k>>" }'

d => 1
b => 3
c => 5
a => 4
```
while using Unicode operators works funny:
```
$ perl6 -e'my $a = BagHash.new: <a a a a b b b c c c c c d>;for $a.keys -> $k { say "$k => $a«$k»" }'

c => c(5) a(4) b(3) d«c»
a => c(5) a(4) b(3) d«a»
b => c(5) a(4) b(3) d«b»
d => c(5) a(4) b(3) d«d»
```
But this works even using Unicode operators:
```
$ perl6 -e'my $a = BagHash.new: <a a a a b b b c c c c c d>;for $a.keys -> $k { say "$k => {$a«$k»}" }'
d => 1
b => 3
a => 4
c => 5
```
(See [https://stackoverflow.com/questions/55569744/a-difference-between-unicode-and-ascii-operators](https://stackoverflow.com/questions/55569744/a-difference-between-unicode-and-ascii-operators)
## Expected Behavior
I think that both type of operators should work the same way.
## Environment
I tested this under Debian sid using Rakudo version 2019.03.1 built on MoarVM version 2019.03

jnthn:
It's probably just a missing case of postcircumfix bracket handling [around here](https://github.com/rakudo/rakudo/blob/master/src/Perl6/Grammar.nqp#L3645), so it considers it's not valid for interpolation.

lizmat:
Tests added with https://github.com/perl6/roast/commit/b0ab879996

PR #2835, opened 2019-04-11
Fix parsing of Unicode brackets in interpolation

MasterDuke17:
Add them as a possibility in postcircumfix bracket ending check.

Will be a fix for #2825 when tests are added.


lizmat:
Thanks!

PR #2842, opened 2019-04-16
Fix build runners

patzim:
Change the runners in the build directory so they work irrespective of CWD. Also change the libpath include order, so the files in the build dir take precedence over the ones in the install location. (We can't just leave the install location out, because the NQP that's installed there is needed.)


patzim:
@jnthn The reason I added you as reviewer is because you observed the problems. I'm primarily interested if my diagnosis (dependence on CWD) matches the failures you observed.
You can of course look over the changes, but I'm quite confident they're good.

patzim:
The Windows failure is a timing test. It succeeded in two of three runs. I'm confident the one failure is not related to these changes.

Issue #2824, opened 2019-04-07
[OpenBSD] Cannot run much of anything without MVM_JIT_DISABLE


Kaiepi:
## The Problem
This is really two separate issues, but I didn't want to be too spammy with them. See title.

Everything worked fine in 2019.03.

## Expected Behavior
zef should be possible to install and it should be possible to run scripts without `MVM_JIT_DISABLE=1` in the env.

## Actual Behavior
See "Steps to Reproduce" since I don't want to repeat myself.

## Steps to Reproduce

Let's try installing zef normally:
```
bastille% perl6 -Ilib bin/zef install .
zsh: trace trap (core dumped)  perl6 -Ilib bin/zef install .
```
Oops. This also happens when trying to install nqp and Rakudo, for the record. The backtrace from the core dump doesn't give much info:
```
(gdb) bt
#0  0x0000009fc176ecb0 in MVM_exception_throwpayload (tc=0x0, mode=0 '\0', cat=301, payload=0x9fdb268fad, 
    resume_result=0x7f7ffffc4160) at src/core/exceptions.c:785
#1  0x000000a094a2cc16 in ?? ()
#2  0x0000000000000004 in ?? ()
#3  0x0000000000000000 in ?? ()
```
When I had moar compiled without `--optimize=0 --debug=3`, one of the JIT ops popped up in the backtrace, so the JIT appears to be to blame. The output of `MVM_SPESH_LOG` shows that it can't find the templates for `getattr_o` and `sp_fastbox_i_ic` in particular, with these also failing, but for different reasons (such as being legojit and thus having no expr template, for instance):
- `bindlex`
- `coerce_in`
- `coerce_ni`
- `eq_n`
- `gt_n`
- `lt_n`
- `prepargs`
- `sp_bind_s_nowb`
- `sp_getlex_ins`
- `sp_getlex_o`
- `sp_guardsf`
- `throwpayloadlex`

Anyway, let's try again with `MVM_JIT_DISABLE=1`:
```
bastille% MVM_JIT_DISABLE=1 perl6 -Ilib bin/zef install .
===SORRY!===
illegal operation on a directory
bastille% MVM_JIT_DISABLE=1 perl6 --ll-exception -Ilib bin/zef install . 
no such file or directory
   at SETTING::src/core/Exception.pm6:71  (/home/morfent/.perl6/share/perl6/runtime/CORE.setting.moarvm:rethrow)
[...]

PR #2845, opened 2019-04-16
Don't use nqp::execname on OpenBSD

Kaiepi:
Fixess one of two parts to #2824

### DIFF
--- a/src/main.nqp
+++ b/src/main.nqp
@@ -2,14 +2,22 @@ use Perl6::Grammar;
 use Perl6::Actions;
 use Perl6::Compiler;
 
+my $config := nqp::backendconfig();
 # Determine Perl6 and NQP dirs.
-my $sep := nqp::backendconfig()<osname> eq 'MSWin32' ?? '\\' !! '/';
+my $sep := $config<osname> eq 'MSWin32' ?? '\\' !! '/';
 #?if jvm
 my $execname := nqp::atkey(nqp::jvmgetproperties,'perl6.execname');
 my $exec-dir := nqp::substr($execname, 0, nqp::rindex($execname, $sep));
 #?endif
-#?if !jvm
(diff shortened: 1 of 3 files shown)


lizmat:
Thanks!

src/main.nqp
```diff
@@ -2,14 +2,22 @@ use Perl6::Grammar;
...
+my $sep := $config<osname> eq 'MSWin32' ?? '\\' !! '/';
 #?if jvm
 my $execname := nqp::atkey(nqp::jvmgetproperties,'perl6.execname');
 my $exec-dir := nqp::substr($execname, 0, nqp::rindex($execname, $sep));
 #?endif
-#?if !jvm
-my $exec-dir := nqp::substr(nqp::execname(), 0, nqp::rindex(nqp::execname(), $sep));
+#?if moar
+my $exec-dir := $config<osname> eq 'openbsd'
+    ?? $config<prefix> ~ '/bin/perl6-m'
```
patzim: Isn't the `/perl6-m` part wrong? `$exec-dir` should be the directory the executable resides in, not the executable itself.

lizmat: /me has no idea

timo:
@Kaiepi can you check if the issue raised in the line comments above is legitimate? it does look like the other parts of the code chop the last bit of the path off with that substr/rindex bit.

timo:
never mind, looks like https://github.com/rakudo/rakudo/commit/e11d0d54e471447efd8d41cc56da0fa130b500ac did this already!

patzim:
@timo The comment is still valid. But since [this](https://github.com/rakudo/rakudo/commit/d3d4b2e8b4fd71cb48e9c83ef5a8b86f486304ce) commit rakudo is not built relocatable on OpenBSD anyways. So one can just revert this change. I'll create a PR this evening.

PR #2852, opened 2019-04-18
Add revision 6.e and improve multi-revision support

### LINKED ISSUE #2783: “Cannot call method 'typed_panic' on a null object” when trying to use v6.e or v6.e.PREVIEW
```perl6
use v6.e.PREVIEW
```
or
```perl6
use v6.e
```

Output:
```
===SORRY!===
Cannot call method 'typed_panic' on a null object
```

vrurg:
This PR is based upon PR #2816.

Also fixes #2783 and adds a warning when `.PREVIEW` is used with a released specification or not used with unreleased.


AlexDaniel:
See also: https://github.com/perl6/roast/issues/506

vrurg:
> See also: [perl6/roast#506](https://github.com/perl6/roast/issues/506)

Like this: perl6/roast#530?

vrurg:
Ah, version dependant spectest.data... Not sure if this is really a good idea as it would complicate things a lot. I think a spectest must skip depending on the current compiler version. Could be fudged, or something like:

```
skip() unless v6.e ∈ $*PERL.compiler.versions
```

src/Perl6/World.nqp
```diff
@@ -530,44 +530,72 @@ class Perl6::World is HLL::World {
           $want
         ) == -1
     }
+
+    method !check-for-PREVIEW ($ver-match, $rev, $default_rev, $rev_mod) {
+        $ver-match.PRECURSOR.worry('PREVIEW modificator is used with a production specification 6.' ~ $rev)
+            if nqp::isle_s($rev, $default_rev) && nqp::iseq_s($rev_mod, 'PREVIEW');
+
+        $ver-match.PRECURSOR.worry(
+            'Language specification 6.' ~ $rev ~ ' is not released yet, PREVIEW modifiecor is expected'
```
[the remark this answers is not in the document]
vrurg: I sure did. This is what happens when touchpad is occasionally clicked. Thank you!

src/Perl6/World.nqp
```diff
@@ -530,44 +530,72 @@ class Perl6::World is HLL::World {
           $want
         ) == -1
     }
+
+    method !check-for-PREVIEW ($ver-match, $rev, $default_rev, $rev_mod) {
+        $ver-match.PRECURSOR.worry('PREVIEW modificator is used with released specification 6.' ~ $rev)
+            if nqp::isle_s($rev, $default_rev) && nqp::iseq_s($rev_mod, 'PREVIEW');
+
+        $ver-match.PRECURSOR.worry(
```
jnthn: I think this one should probably be an error.

src/Perl6/World.nqp
```diff
@@ -530,44 +530,72 @@ class Perl6::World is HLL::World {
...
-            $comp.set_language_version:       '6.' ~ $lang;
-            self.load_setting: $ver-match, 'CORE.' ~ $lang;
-            return;
+        # Do we have dot-splitted version string?
+        if (@vparts > 1) || ($version eq 'v6') {
+            my $revision := @vparts[1] || $default_rev;
+            my $lang_ver := '6.' ~ $revision;
+
+            self."!check-for-PREVIEW"($ver-match, $revision, $default_rev, @vparts[2] || '')
+                if ($revision eq 'c') || ($lang_ver eq $comp.config<language-version>);
```
jnthn: Hm, why the explicit check for `'c'` here? Should we not be getting the language version (`c`, `d`, etc.) and comparing it with `le` against the version we're currently at?

src/Perl6/World.nqp
```diff
@@ -530,44 +530,72 @@ class Perl6::World is HLL::World {
...
+            $comp.set_language_version: $lang_ver;
+            # fast-path the common cases
+            if $revision eq 'c' {
+                $*CAN_LOWER_TOPIC := 0;
+                # CORE.c is currently our lowest core, which we don't "load"
+                return;
+            }
+
+            # Does this makes any sense besides of speeding up loading assuming that the default language version would be
+            # the most used one?
```
jnthn: I think that's what it's doing.

src/Perl6/World.nqp
vrurg: There is revision definition file `PERL6_SPECS` in `tools/templates` now. Looks like:

```
# [*]version [suffix]
# * defines the current spec
c
*d PREVIEW # TODO: It's about time to drop PREVIEW
#e PREVIEW
```

If there is no `PREVIEW` for a revision then this is a error. I'll re-check if this is the case. Otherwise the warning is for the transition period to allow graceful migration.

src/Perl6/World.nqp
vrurg: There is no `CORE.c.setting`, thus – special case for `c`.

src/Perl6/World.nqp
vrurg: What I think must be done differently here is don't hardcode PREVIEW but rely on what is defined in `PERL6_SPECS`.

vrurg:
Everything is ready for merge. Would await for perl6/problem-solving#31 to decide how to handle spectests.

PR #2854, opened 2019-04-22
Temporary: Fix PERL6/NQP_HOME broken without --execpath

### LINKED ISSUE #2850: Invocation used to run Perl 6 code under remote debugger broken
## The Problem

The only current way to launch remote debugging is through invoking MoarVM with `perl6.moarvm` followed by arguments. A typical invocation line looks like this:

```
/home/jnthn/dev/MoarVM/install/bin/moar --debug-port=9999 --debug-suspend --libpath=/home/jnthn/dev/MoarVM/install/share/nqp/lib --libpath=/home/jnthn/dev/MoarVM/install/share/perl6/lib --libpath=/home/jnthn/dev/MoarVM/install/share/perl6/runtime /home/jnthn/dev/MoarVM/install/share/perl6/runtime/perl6.moarvm -Ilib /home/jnthn/foo/foo.p6
```

## Expected Behavior

This invocation works, as it did in 2019.3 and many prior releases.

## Actual Behavior

It doesn't find installed modules. Output:

```
===SORRY!===
Could not find Cro::HTTP::Log::File at line 1 in:
    file#/home/jnthn/op/easii/easii/lib
    inst#/home/jnthn/.perl6
    inst#/share/perl6/site
    inst#/share/perl6/vendor
    inst#/share/perl6
    ap#
    nqp#
    perl5#
```

Those paths starting with `/share` look broken. I suspect this regressed as part of the work on relocation. That work is very welcome, but we should support existing workflows too.

## Steps to Reproduce

Construct an invocation line like the one above. Or, if you have [Comma](https://commaide.com/download), just try to debug anything that uses an installed module. The 1-line program `use NativeCall;` will do it.

## Environment
* Operating system: Linux
* Compiler version (`perl6 -v`): This is Rakudo version 2019.03.1-228-g30e97a5 built on MoarVM version 2019.03-82-g7b112da
implementing Perl 6.d.

patzim:
The previous behavior broke comma debugging, because comma directly calls moar but did not pass an --execpath parameter, thus breaking PERL6_HOME and NQP_HOME detection.
This temporary fix adds a fallback when exec-path is empty to then use the hard coded install path of the configure config. This breaks relocatability in this specific case, but better than not working at all.
Should fix #2850.
I have not tested this with comma yet. Will try to install it now and see how it goes.

### DIFF
--- a/src/main.nqp
+++ b/src/main.nqp
@@ -2,32 +2,47 @@ use Perl6::Grammar;
 use Perl6::Actions;
 use Perl6::Compiler;
 
-my $config := nqp::backendconfig();
-# Determine Perl6 and NQP dirs.
-my $sep := $config<osname> eq 'MSWin32' ?? '\\' !! '/';
-#?if jvm
-my $execname := nqp::atkey(nqp::jvmgetproperties,'perl6.execname');
-my $install-dir := nqp::substr($execname, 0, nqp::rindex($execname, $sep, nqp::rindex($execname, $sep) - 1));
-#?endif
-#?if moar
-my $execname := nqp::execname();
-my $install-dir := $config<osname> eq 'openbsd'
-    ?? $config<prefix> ~ '/bin/perl6-m'
-    !! nqp::substr($execname, 0, nqp::rindex($execname, $sep, nqp::rindex($execname, $sep) - 1));
-#?endif
-#?if js
-my $execname := nqp::execname();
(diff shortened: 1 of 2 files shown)


patzim:
The failure on appveyor is a timing related test. I'm quite sure this is unrelated to this PR, given it succeeded in the two other appveyor runs.
I also tested this with comma and debugging does work.

jnthn:
Looks sensible to me; thanks!

PR #2857, opened 2019-04-24
Don't special case OpenBSD in relocatability handling

patzim:
Relocatability is disabled entirely on OpenBSD, because of technical
difficulties. So no need to put hacks in to make it work in relocatable
mode.
Also simplify logic a bit more. Should still do the same as before.


Kaiepi:
Regardless of whether or not the executable is relocatable, `nqp::execname` is still being used when it returns a random flag passed to moar on OpenBSD. I don't think what you're trying to do here will work on OpenBSD without changing `MVM_execname` in MoarVM to return an empty string, but that behaviour would be coupling.

Cleaning up the code here is good, but I don't think you can entirely get rid of the hack. I'd write something like this to get this to work:
```perl6
my $execname := $*VM.osname eq 'openbsd' ?? '' !! nqp::execname();
```

PR #2858, opened 2019-04-24
Document MoarVM specific commandline options

patzim:
Several perl6 commandline options are specific to MoarVM. Only show
those when running on MoarVM. Add the missing ones. Also add documentation
for PERL6_HOME and NQP_HOME.

### DIFF
--- a/docs/running.pod
+++ b/docs/running.pod
@@ -30,6 +30,12 @@ compiled code.
   --profile            write profile information as HTML file (MoarVM)
   --profile-filename   provide a different filename (also allows .json)
   --doc=[module]       Use Pod::To::[module] to render inline documentation.
+  --dump               dump the bytecode to stdout instead of executing (MoarVM)
+  --full-cleanup       try to free all memory and exit cleanly (MoarVM)
+  --debug-port=port    listen for incoming debugger connections (MoarVM)
+  --debug-suspend      pause execution at the entry point (MoarVM)
+  --tracing            output a line to stderr on every interpreter instr (only
+                       if enabled in MoarVM)
 
 Note that only boolean single-letter options may be bundled.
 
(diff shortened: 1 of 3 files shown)


docs/running.pod
```diff
@@ -30,6 +30,12 @@ compiled code.
   --profile            write profile information as HTML file (MoarVM)
   --profile-filename   provide a different filename (also allows .json)
   --doc=[module]       Use Pod::To::[module] to render inline documentation.
+  --dump               dump the bytecode to stdout instead of executing (MoarVM)
```
jnthn: Hmm, not sure how useful this is, in that it doesn't work with Perl 6 input at all, but only a MoarVM bytecode file (assuming it just forwards to `moar --dump`)?

patzim: `./perl6 --dump -e 'say "Hi"'` does seem to do stuff.

jnthn: Wow, it does. It appears to dump the bytecode of the compiler entrypoint program, though, not the script given to `-e`!

patzim: It'll make sense to remove the argument altogether then. I'll see what I can do.

jnthn: (To add: it's not obvious how to wire the pieces together to actually use that dumper with the input given to the compiler.)

patzim:
This PR now also fixes --full-cleanup, which missed its implementation.
The --dump flag is now removed as it made little sense in the perl6 executable. When dealing with moarvm files
one should better call moar instead.

src/vm/moar/runner/main.c
```diff
@@ -349,7 +347,13 @@ int wmain(int argc, wchar_t *wargv[])
...
 #endif
 
-    MVM_vm_exit(instance);
+    if (full_cleanup) {
+        MVM_vm_destroy_instance(instance);
+        return EXIT_SUCCESS;
+    }
+    else {
+        MVM_vm_exit(instance);
+    }
```
jnthn: While it doesn't block merging this at all, note that the `free` lines below are unreachable, due to the fact that the first branch `return`s and the second calls `exit`. This isn't a change from before, but is maybe worth looking at.

patzim: I fixed this in 34162e967161f34755fb .

PR #2874, opened 2019-05-03
Add support for bundling arguments to the default argument parser

Leont:
This adds a new entry to `%*SUB-MAIN-OPTS`, `bundling`. If it is enabled, any single dash argument will be interpreted as a bundle of single character arguments: `-ab` will parse to `:a, :b` instead of `:ab`. If it is not enabled, or if the argument starts with a double dash, it will still parse as `:ab` as before.

This is meant as a solution for perl6/problem-solving#18

### DIFF
--- a/src/core.c/Main.pm6
+++ b/src/core.c/Main.pm6
@@ -46,6 +46,7 @@ my sub RUN-MAIN(&main, $mainline, :$in-as-argsfiles) {
     # Convert raw command line args into positional and named args for MAIN
     sub default-args-to-capture($, @args is copy --> Capture:D) {
         my $no-named-after = nqp::isfalse(%sub-main-opts<named-anywhere>);
+        my $bundling = nqp::istrue(%sub-main-opts<bundling>);
 
         my $positional := nqp::create(IterationBuffer);
         my %named;
@@ -72,21 +73,31 @@ my sub RUN-MAIN(&main, $mainline, :$in-as-argsfiles) {
 
             # named
             elsif $passed-value
-              ~~ /^ [ '--' | '-' | ':' ] ('/'?) (<-[0..9\.]> .*) $/ {  # 'hlfix
-                my str $arg = $1.Str;
+              ~~ /^ ( '--' | '-' | ':' ) ('/'?) (<-[0..9\.]> .*) $/ {  # 'hlfix
+                my str $arg = $2.Str;
                 my $split  := nqp::split("=",$arg);
 
                 # explicit value
                 if nqp::isgt_i(nqp::elems($split),1) {
                     my str $name = nqp::shift($split);
-                    %named.push: $name => $0.chars
+                    die "Can't combine bundling with explicit arguments" if $bundling && nqp::iseq_s($0.Str, '-');
+                    %named.push: $name => $1.chars
                       ?? thevalue(nqp::join("=",$split)) but False
                       !! thevalue(nqp::join("=",$split));
                 }
 
                 # implicit value
                 else {
-                    %named.push: $arg => !($0.chars);
+                    if $bundling && nqp::iseq_s($0.Str, '-') {
+                        die "Can't combine bundling with explicit negation" if $1.chars;
+                        my @chars = nqp::split('',$arg);
+                        for @chars -> $char {
+                            %named.push: $char => True;
+                        }
+                    }
+                    else {
+                        %named.push: $arg => !($1.chars);
+                    }
                 }
             }
 


ugexe:
Does this consistently handle negation or parameters? e.g. `-/abc`, `-abc=1`, `-/abc=1`

Leont:
> Does this consistently handle negation or parameters? e.g. -/abc, -abc=1, -/abc=1

It handles negation but not parameters. I'm not quite sure what it should do with parameters to be honest.

ugexe:
It feels like it should be the same as `--a=1 --b=1 --c=1`, although I don't entirely disagree with the alternative.

Another edge case could be types: `-abc=42` with `sub MAIN(Bool :$a, Int :$b, :$c) { }`.

chee:
i think the closest thing to POSIX getopt style behaviour (and, i’m guessing here that the closest to that would be what people using shells would expect) is that `-abc=lol` would set only `c` to `lol`.

(like, when something has a boolean “a” and “b” arg and “c” takes an argument)

but maybe if its just too hard to guess what people would expect, could setting a parameter in that way print a helpful error message? and then wait for people to report what it was they expected to happen?

tony-o:
Looking at `-abc=1` I'd expect `a = True, b = True, c = 1`, `-/abc=1` to be `a = False, b = True, c = 1`

Leont:
I think that the behavior that @ugexe describes is the easiest to describe and implement.

The behavior @chee describes is indeed most in line with getopt implementations (and thus least surprising to outsiders), though it feels rather awkward when combined with negation.

That said I hadn't even realized negation and parameters could be combined until I had read the code; and to be honest it doesn't sound like something anyone should be doing anyway.

ugexe:
> and to be honest it doesn't sound like something anyone should be doing anyway.

Its not immediately intuitive, but I'm not sure I'd say it shouldn't be used.

I like how simple it makes handling both broad and granular options instead of special casing `"*"` everywhere (which might be a valid non-all-encompassing value).

```
# Update all module indexes
--update

# Do not update any module indexes
--/update

# Only update the cpan module index
--update=cpan

# Do not update the cpan module index
--/update=cpan

# Do not update the cpan or p6c module index
--/update=cpan
--/update=p6c
```

lizmat:
Looks ok to me, but won't merge until the 2019.05 release is out.

JJ:
Maybe look at this with fresh eyes? (And solve conflicts)

Leont:
> Maybe look at this with fresh eyes? (And solve conflicts)

I rebased it, but didn't encounter any conflicts

patrickbkr:
Reading through the above discussion, what is the current implementation doing for `-/abc=1` ? I think this is a change in Raku that is very difficult to change later because of backwards compatibility. So I'd like this to be very well thought through. (I do like argument bundling. So I am in favour of getting this in one way or the other.)

I personally find `-/abc=1` meaning the same as `-/a -b -c=1` most intuitive. Also I think aligning as closely as possible with the original posix utilities a very rewarding feature.

ugexe:
Currently to Raku -abc and -/abc are the same as —abc and —/abc. So it is possible people have been using (probably by accident) the single dash variants in their e.g. CI steps for zef.

ugexe:
Ah but again, this is an option to be enabled so it shouldn’t regress anything

Leont:
> Reading through the above discussion, what is the current implementation doing for -/abc=1 ?

Currently, nothing. or better said it's equivalent to `--/abc=1`

> I think this is a change in Raku that is very difficult to change later because of backwards compatibility.

We could also disallow these combinations when `:bundling` is enabled until we figured out what would be best.

> I personally find -/abc=1 meaning the same as -/a -b -c=1 most intuitive.

I think that only makes sense to me if you can prepend that `/` before other letters as well? It also feels like this is getting a bit complicated.

patrickbkr:
I like the idea of disallowing these combinations. I think they'll be used rarely. So disallowing them won't be much pain. This will also give us a feel for how important it actually is to provide a solution for argument bunding with negation or values, depending on how fast a request for supporting this will pop up.

Leont:
> I like the idea of disallowing these combinations.

I have implemented just that

lizmat:
While CI is doing its thing, is this description still accurate?

> This adds a new entry to %*SUB-MAIN-OPTS, bundling. If it is enabled, any single dash argument will be interpreted as a bundle of single character arguments: -ab will parse to :a, :b instead of :ab. If it is not enabled, or if the argument starts with a double dash, it will still parse as :ab as before.

src/core.c/Main.pm6
```diff
@@ -72,21 +73,31 @@ my sub RUN-MAIN(&main, $mainline, :$in-as-argsfiles) {
...
                       !! thevalue(nqp::join("=",$split));
                 }
 
                 # implicit value
                 else {
-                    %named.push: $arg => !($0.chars);
+                    if $bundling && nqp::iseq_s($0.Str, '-') {
+                        die "Can't combine bundling with explicit negation" if $1.chars;
+                        my @chars = nqp::split('',$arg);
+                        for @chars -> $char {
```
lizmat, line 94: Couldn't we just do:

    for nqp::split('',$arg) -> $char {

??

Leont: Sure

patrickbkr:
Looking at the changeset i realize that this PR prohibits negation and explicit values with single-letter-single-dash arguments as well. As far as I can see, it's unproblematic to allow that. Is there any reason for not allowing that?

lizmat:
@patrickbkr perhaps something for a followup PR ?

lizmat (approved):
Am about to merge this.  Please let me know in the next few hours if this should **not** be done.

src/core.c/Main.pm6
Leont: Actually we can't, that dies with «Cannot find method 'iterator' on 'BOOTArray': no method cache and no .^find_method»

Leont:
> Looking at the changeset i realize that this PR prohibits negation and explicit values with single-letter-single-dash arguments as well. As far as I can see, it's unproblematic to allow that. Is there any reason for not allowing that?

It's currently not allowed because we didn't seem to agree on how those should behave. I'm fine with us adding behavior at a later time.

lizmat:
@Leont congratulations and thanks for your perseverance!

PR #2880, opened 2019-05-06
Fix some double frees in the moar runner

Kaiepi:


src/vm/moar/runner/main.c
```diff
@@ -123,13 +125,11 @@ int retrieve_home(char *out_home, char *rel_home, char *env_var, char *exec_dir_
 
     check_file_path = (char*)malloc(home_size + 50);
     memcpy(check_file_path, out_home, home_size);
-    strcpy(check_file_path + home_size, check_file);
-    if (!file_exists(check_file_path)) {
-        free(check_file_path);
-        return 0;
-    }
+    strncpy(check_file_path + home_size, check_file, 50);
```
niner: srncpy does not add a trailing \0 when it runs into the limit. We should do that manually

Kaiepi: Oh, right. I was getting it confused with `strlcpy`, which does as long as the destination buffer has room, not the source

Kaiepi:
This should pass tests now and ensures the paths `strncpy` are used with are null terminated, along with a couple other minor fixes

PR #2884, opened 2019-05-09
Improve GNU C++ name mangling

Kaiepi:
This makes it so the symbol is not assumed to be a constructor when it
includes new unless it's the last part of the symbol and infers whether
or not the symbol is a constructor by checking if the last two parts of
it are identical.

This also mangles Str, Pointer, and CArray attributes with the rw or
cpp-const traits properly.

### DIFF
--- a/lib/NativeCall/Compiler/GNU.pm6
+++ b/lib/NativeCall/Compiler/GNU.pm6
@@ -6,21 +6,30 @@ our sub mangle_cpp_symbol(Routine $r, $symbol) {
     $r.signature.set_returns($r.package)
         if $r.name eq 'new' && !$r.signature.has_returns && $r.package !~~ GLOBAL;
 
-    my $mangled = '_Z'
-                ~ ($r.package.REPR eq 'CPPStruct' ?? 'N' !! '')
-                ~ ($r.?cpp-const ?? 'K' !! '')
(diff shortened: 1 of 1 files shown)


patzim:
Travis failed with

```
t/04-nativecall/11-cpp.t ........................................ 
1..21
Don't know how to mangle symbol 'SizeofDerived1' for library '/home/travis/build/rakudo/rakudo/lib11-cpp.so'
  in sub guess-name-mangler at /home/travis/build/rakudo/rakudo/lib/NativeCall.pm6 (NativeCall) line 250
  in method setup at /home/travis/build/rakudo/rakudo/lib/NativeCall.pm6 (NativeCall) line 282
  in method CALL-ME at /home/travis/build/rakudo/rakudo/lib/NativeCall.pm6 (NativeCall) line 576
  in block <unit> at t/04-nativecall/11-cpp.t line 50
# Looks like you planned 21 tests, but ran 0
Dubious, test returned 255 (wstat 65280, 0xff00)
Failed 21/21 subtests
```

Kaiepi:
My bad, I was only running `t/04-nativecall/14-cpp-mangling.t`. Should be fixed now

AlexDaniel:
So that whole piece of code looks like a total hack both before and after this PR, but I guess that's OK? Anyway, this seems like an improvement, we'll see how it goes (`make test` and `make spectest` are clean).

Thanks!

PR #2890, opened 2019-05-13
Don't assign to $*IN/$*OUT/$*ERR, bind instead

### LINKED ISSUE #2888: Regression in get/prompt subs when using $*IN
## The Problem
`get` using $*IN and `prompt` both print `^M` when enter is pressed instead of returning.

## Expected Behavior
The sub should return the text input.

## Actual Behavior
The sub never returns.

## Steps to Reproduce

Run `RAKUDO_LINE_EDITOR=none perl6` and try to hit enter after typing in some code.

## Environment
* Operating system: 
OpenBSD bastille.kennel.qt 6.5 GENERIC.MP#8 amd64
This is Rakudo version 2019.03.1-297-g62c486b410 built on MoarVM version 2019.03-111-g24b3ef1ad
implementing Perl 6.d.

Kaiepi:
Fixes #2888

### DIFF
--- a/src/core/io_operators.pm6
+++ b/src/core/io_operators.pm6
@@ -197,9 +197,9 @@ multi sub indir(IO() $path, &what, :$d = True, :$r, :$w, :$x) {
     }
 
     # Activate the skeletons at runtime
-    PROCESS::<$IN>  = activate-handle(STDIN,  nqp::getstdin);
-    PROCESS::<$OUT> = activate-handle(STDOUT, nqp::getstdout);
-    PROCESS::<$ERR> = activate-handle(STDERR, nqp::getstderr);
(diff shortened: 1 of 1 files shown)


lizmat:
Won't this prevent $*IN and friends from being re-assigned ???

b2gills:
```perl6
$*ERR = $*OUT;

note 'foo'; # should go to $*OUT
```

Kaiepi:
...right. It's still possible to reassign to them like this, but I don't think it's ideal:
```perl6
PROCESS::{'$ERR'} := $*OUT;
```

I'll look for a better solution.

Kaiepi:
Tested again on HEAD and it works fine...? I have no clue what went wrong before but it seems more like a terminal issue, not a Rakudo issue

PR #2891, opened 2019-05-13
Require Java 9 for JVM backend

usev6:


usev6:
This should only be merged after the next nqp bump.

usev6:
I've commited this directly.

PR #2896, opened 2019-05-14
Make feed operators more compliant with the spec

Kaiepi:
This speeds up use of feed operators by roughly 650%. Using feed
operators is now even faster than directly calling methods on lists.

This passes `make test`, but fails one test on `make spectest` that depends on using feed operators on lazy lists throwing.

A side effect of this is the output generated by feed operators is now `List`, not `Array`. This may break packages that depend on the output being mutable.

### DIFF
--- a/src/Perl6/Actions.nqp
+++ b/src/Perl6/Actions.nqp
@@ -7144,16 +7144,20 @@ class Perl6::Actions is HLL::Actions does STDActions {
 
     sub make_feed($/) {
         # Assemble into list of AST of each step in the pipeline.
-        my @stages;
-        if $/<infix><sym> eq '==>' {
-            for @($/) { @stages.push($_); }
+        my $stages := nqp::list();
+        if $<infix><sym> eq '==>' {
+            for $/.list {
+                nqp::push($stages, $_);
+            }
         }
-        elsif $/<infix><sym> eq '<==' {
-            for @($/) { @stages.unshift($_); }
+        elsif $<infix><sym> eq '<==' {
+            for $/.list {
+                nqp::unshift($stages, $_);
+            }
         }
         else {
             $*W.throw($/, 'X::Comp::NYI',
-                feature => $/<infix> ~ " feed operator"
+                feature => $<infix> ~ " feed operator"
             );
         }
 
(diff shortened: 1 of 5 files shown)


lizmat:
Looking at the diff, I'm surprised this makes that much of a difference.  Could you paste the benchmark that you've used?

Kaiepi:
```perl6
my Instant $b = now; my @ <== map { $_ ** 2 } <== 1..100 for 1..10000; say now - $b;
```

Before the changes, this took 7.4531064 seconds. After the changes, it takes 1.1717105 seconds. The equivalent code using normal method calls takes 5.9365526 seconds:

```perl6
my Instant $b = now; my @ = (1..100).map({ $_ ** 2 }) for 1..10000; say now - $b;
```

Kaiepi:
Since the benchmark I posted before was bugged I reran it using this code:

```perl6
use v6.d;

sub run-with-feed {
    my @ <== map { $_ ** 2 } <== 1..100
}

sub run-with-method {
    my @ = (1..100).map({ $_ ** 2 })
}

sub run(&runner --> Duration) {
    my Instant $b = now;
    runner for 1..100_000;
    now - $b
}

say "Feed:   " ~ run &run-with-feed;
say "Method: " ~ run &run-with-method;
```

Output on master:
```
bastille% perl6 test.p6
Feed:   79.81106198
Method: 61.1389592
```

Output on this branch:
```
bastille% perl6 test.p6
Feed:   8.48146285
Method: 62.3589368
```

These changes make actually make feed operators run even faster than I originally stated??? This is a 950% change.

lizmat:
I'm pretty sure the perceived improvement is from creating a List (with bound values) rather than an Array (bound containers with values in them):

    $ perl6 =e 'my @a := ^100 .List for ^100_000'     # real	0m0.306s
    $ perl6 =e 'my @a := ^100 .Array for ^100_000'   # real	0m2.062s

So the question becomes really: do we want to change the semantics of the feed operator so that it returns List rather than Arrays?

Kaiepi:
I'll take a look through the ecosystem to see if these changes would break any modules later. If not, I don't see why it would be a bad idea

lizmat:
Specs about feeds:

>Note that all such feeds (and indeed all lazy argument lists) supply
an implicit promise that the code producing the lists may execute
in parallel with the code receiving the lists.  (Feeds, hyperops,
and junctions all have this promise of parallelizability in common,
but differ in interface.  Code which violates these promises is
erroneous, and will produce undefined results when parallelized.)

However, feeds go a bit further than ordinary lazy lists in enforcing
the parallel discipline: they explicitly treat the blunt end as a
cloned closure that starts a subthread (presumably cooperative).

Kaiepi:
While looking through the ecosystem I found that this breaks using feed operators to assign to hashes. The fix for this pretty well removes any optimization this originally had.

I'm making this a pullreq to fully implement spec-compliant feed operators in part because of #2899

lizmat:
I think in light of recent developments, this PR can be closed ?

Kaiepi:
> I think in light of recent developments, this PR can be closed ?
I was originally planning on making this a pullreq to add *all* missing features when it comes to feed operators, but that's too much work. I'll close this and make individual pullreqs for each feature instead

PR #2903, opened 2019-05-17
Parallelize feed operators and support lazy lists

Kaiepi:
This brings feed operators (`==>` and `<==` at least) more in line with the spec.

Passes `make test`, fails `make spectest` since one of its tests expects trying to use lazy lists with feed operators to throw.

### DIFF
--- a/src/Perl6/Actions.nqp
+++ b/src/Perl6/Actions.nqp
@@ -7142,27 +7142,34 @@ class Perl6::Actions is HLL::Actions does STDActions {
         make $past;
     }
 
-    sub make_feed_result($source, $target, @stages) {
-        my $result := nqp::if(
-          nqp::elems(@stages),
-          # There are routines to be called; invoke
-          # Rakudo::Internals.EVALUATE-FEED to get the result.
-          QAST::Op.new(
-              :op('callmethod'), :name('EVALUATE-FEED'),
-              QAST::WVal.new( :value($*W.find_symbol(['Rakudo', 'Internals'], :setting-only)) ),
-              $source, |@stages
-          ),
-          # There are no routines to be run; the result is just the source.
-          $source
-        );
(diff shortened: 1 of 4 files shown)


Kaiepi:
There's a problem with this... this benches slower than the current implementation of feed operators, even when there's blocking I/O going on at the same time. I think more discussion needs to be made about whether or not this should be implemented.

vrurg:
Have you tried to simulate same process in a test? I mean without operator implementation, just chain a bunch of workers together and see what happens?

Kaiepi:
Do you mean something like this?
```perl6
sub run-no-operators {
    my Channel $c .= new;
    $c.send: (1..100);
    await (
        { map { $_ ** 2 }, $_ },
        { grep { $_ %% 2 }, $_ },
        { reverse $_ },
        { sort $_ }
    ).map({
        start $c.send: $_($c.receive)
    });
    my @ = $c.receive;
}
```
If so, it runs slightly faster than with feed operators, but is still slower than before. If not, what do you mean?

Kaiepi:
Turns out getting this to run faster was just a matter of handling QAST better. Feed operators now run roughly 500% faster than the current implementation.

Kaiepi:
So I tinkered around with the optimizer trying to get the values of variables that can't be accessed from actions. `$*W.compile_time_evaluate` panics when I try to use it from there, which was what I was depending on for my last attempt at optimizing this to work. I can't think of any other ways to optimize this so it's faster than the current implementation.

Kaiepi:
I changed the way I parallelize feed operators to my first attempt at implementing them. It may be slightly slower than the current implementation, but hopefully once more work is done on `Channel` they will be more performant.

Kaiepi:
Now's probably not the best time for this with upcoming RakuAST stuff. Besides, this isn't a very good implementation.

Issue #2904, opened 2019-05-17
Rakudo no longer builds properly on OpenBSD


Kaiepi:
## The Problem
The new build scripts try to build a relocatable executable by default on OpenBSD when that doesn't work.

## Expected Behavior
They should build a non-relocatable executable.

## Actual Behavior
They don't. Passing `--no-relocatable` also doesn't work without manually editing the makefile since it gets passed to it as `--no-relocatable="1"` when it doesn't take a value, making one of the scripts called by `make` throw. Also, `Configure.pl` incorrectly states that you should use `gmake` instead of `make`.

## Steps to Reproduce

Run `./Configure.pl && make install`, then try to run `perl6`.

## Environment
* Operating system: 
OpenBSD bastille.kennel.qt 6.5 GENERIC.MP#8 amd64
* Compiler version (`perl6 -v`): 
This is Rakudo version 2019.03.1-405-g7a7a23df6a built on MoarVM version 2019.03-111-g24b3ef1ad
implementing Perl 6.d.

vrurg:
With regard to gmake – do you have it in your path? If so, then this is the recommended make to use. But if you don't then this is a problem.

vrurg:
I'm installing openBSD in a VM. Meanwhile may I ask for the output of the failing script?

Kaiepi:
> With regard to gmake – do you have it in your path? If so, then this is the recommended make to use. But if you don't then this is a problem.

I do. I don't think it should be the recommended make on *BSD though since it's not installed by default.

> I'm installing openBSD in a VM. Meanwhile may I ask for the output of the failing script?

```
bastille% ./Configure.pl --backends=moar --prefix=$HOME/.perl6 --no-relocatable
Using /home/morfent/.perl6/bin/nqp-m (version 2019.03-83-g0adfd6ada / MoarVM 2019.03-111-g24b3ef1ad).
Cleaning up ...

You can now use '/usr/local/bin/gmake' to build Rakudo.
After that, '/usr/local/bin/gmake test' will run some tests and
'/usr/local/bin/gmake install' will install Rakudo.
bastille% make                                                                 
'perl' -I'/home/morfent/Documents/perl6/rakudo/tools/lib' -I'/home/morfent/Documents/perl6/rakudo/3rdparty/nqp-configure/lib' '/home/morfent/Documents/perl6/rakudo/tools/build/check-nqp-version.pl' '/home/morfent/.perl6/bin/nqp-m'
[...]

PR #2906, opened 2019-05-17
Support for latest changes in nqp-configure

vrurg:
OpenBSD and command line options improvements.

rakudo/rakudo#2904


Benabik:
This PR fixes building on macOS with `--no-relocatable`.

Issue #2919, opened 2019-05-24
Configure.pl fails with NQP in a different prefix


niner:
```
nine@sphinx:~/test/build/rakudo (master=)> perl Configure.pl --prefix=/home/nine/test/build/install/rakudo/ --with-nqp=/home/nine/rakudo/install/bin/nqp

===SORRY!===
No active backends found. Please, report this bug!

 at /home/nine/test/build/rakudo/3rdparty/nqp-configure/lib/NQP/Config.pm line 34.
        NQP::Config::__ANON__("\x{a}===SORRY!===\x{a}No active backends found. Please, report this b"...) called at /home/nine/test/build/rakudo/3rdparty/nqp-configure/lib/NQP/Config.pm line 162
        NQP::Config::sorry(NQP::Config::Rakudo=HASH(0x56552c722f68), "No active backends found. Please, report this bug!") called at /home/nine/test/build/rakudo/3rdparty/nqp-configure/lib/NQP/Config.pm line 312
        NQP::Config::active_backends(NQP::Config::Rakudo=HASH(0x56552c722f68)) called at /home/nine/test/build/rakudo/tools/lib/NQP/Config/Rakudo.pm line 64
        NQP::Config::Rakudo::configure_backends(NQP::Config::Rakudo=HASH(0x56552c722f68)) called at Configure.pl line 92
```

vrurg:
Question similar  to perl6/nqp#544: would it make sense to pick up `nqp` from `PATH`?

kazcw:
The next problem I'm encountering is that current nqp still reports its version as "2019.03" (https://github.com/perl6/nqp/blob/master/VERSION), while rakudo's nqp_want compares as newer: "2019.03-177-g6372077". I presume the solution is to bump nqp's VERSION, although I don't know the conventions around doing that.

Looking at the source, I think it works when gen-nqp is enabled because that code path has a different way of checking the version: it tries to fetch $nqp_want from the git repo, and then assumes what is obtained is the correct version. Maybe it would be prudent to check nqp's reported version even if it has been built from git, so that nqp must pass the same version test no matter where it came from?

vrurg:
@kazcw Could you please provide me with more details about the `VERSION` situation? I don't see what's wrong with it. Are you trying to build against the last release?

Otherwise if you could give a try to #2927. It's very likely not ready yet but at least it builds with `./Configure --with-nqp ~/perl6/bin/nqp --prefix ./install`.

[...]

PR #2922, opened 2019-05-24
fix excessive quoting in nqp_bin test

kazcw:
With extra quotes, expression evaluates to nothing and with-nqp option
never works. Part of fixing #2919.


vrurg:
Looks reasonable. Thanks!

PR #2938, opened 2019-05-30
fix install-dist

kazcw:
It was deleting the `version` file from the installed dist, rendering it
impossible to import (because CompUnitRepo::Installation.candidates
assumes a dist without a version file is version 0 and tries to
.lines its short/ directory, which does not go well).

### DIFF
--- a/tools/install-dist.p6
+++ b/tools/install-dist.p6
@@ -52,7 +52,7 @@ multi sub MAIN(:from(:$dist-prefix) = '.', :to(:$repo-prefix)!, :for(:$repo-name
         :name($repo-name),
     ).install($dist);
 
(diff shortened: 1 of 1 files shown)


niner:
It only deletes those files when installing the module to a Staging repository. The Staging repo is meant for packaging, i.e. you package all the generated files into an .rpm, .deb, .tar or whatever and install later by just extracting that archive. That's why install-dist.p6 deletes all files from the repository that are not directly related to the installed distribution. It would be a bit cleaner though to have the Staging repo itself care about such things.

If you intend to actually use the repository you install to directly, do not use install-dist.p6's `--for` argument

PR #2940, opened 2019-05-31
Move the core comp unit repo to a separate folder

patzim:
As proposed [in this issue comment](https://github.com/rakudo/rakudo/issues/2919#issuecomment-496417006), move the core compilation unit repository previously installed directly into `<Perl6 home>/` to `<Perl6 home>/core`. This should continue to work as before, but does not attempt to remove any old stale comp unit repository a previous install might have put there. So some cruft might be left behind.

### DIFF
--- a/tools/build/install-core-dist.p6
+++ b/tools/build/install-core-dist.p6
@@ -22,16 +22,16 @@ PROCESS::<$REPO> := CompUnit::Repository::Staging.new(
         # Make CompUnit::Repository::Staging available to precomp processes
         CompUnit::Repository::Installation.new(
             :prefix(@*ARGS[0]),
-            :next-repo(CompUnit::RepositoryRegistry.repository-for-name('perl')),
+            :next-repo(CompUnit::RepositoryRegistry.repository-for-name('core')),
         )
     ),
-    :name('perl'),
+    :name('core'),
 );
 $*REPO.install(
     Distribution::Hash.new(
         {
-            name     => "CORE",
-            auth     => "perl",
+            name     => 'CORE',
+            auth     => 'perl',
             ver      => $*PERL.version.Str,
             provides => %provides,
         },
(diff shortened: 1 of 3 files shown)


patzim:
I just noticed, that this branch is based on my reloc-fix branch and thus contains these changes as well. So please do not merge this branch prior to PR#2939.

ugexe:
I wonder if we could get away with switching the short name from `perl` to `core`. Its already poorly named since its parent folder is `perl6`...
```
$ perl6 -e 'say CompUnit::RepositoryRegistry.repository-for-name("perl")'
inst#/Users/ugexe/.rakudobrew/moar-blead-master/install/share/perl6
```

patzim:
I'm not too familiar with the ins and outs of the repo stuff, but my guess would be the rename would probably work out without any fallout. @nine Do you have an opinion?

ugexe:
Only notable thing I can think of is:
https://github.com/rakudo/rakudo/blob/2566db7b9dca3e07d5b9fa889a3e70210b430b88/tools/build/install-core-dist.p6#L23-L26

patzim:
@ugexe, @niner I tried to also rename the repo from `perl` to `core` as per the suggestion of ugexe. I'm absolutely unsure about the changes I did for the rename (I know little about comp repos). It's possible I changed a `'perl'` to `'core'` that's unrelated to the repo name. Please do review.
Because the release is nearing it's probably a good idea to merge this as soon as possible, so we at least have some time left before the release.

tools/build/install-core-dist.p6
```diff
@@ -22,16 +22,16 @@ PROCESS::<$REPO> := CompUnit::Repository::Staging.new(
...
-    :name('perl'),
+    :name('core'),
 );
 $*REPO.install(
     Distribution::Hash.new(
         {
-            name     => "CORE",
-            auth     => "perl",
+            name     => 'CORE',
+            auth     => 'core',
```
ugexe: The auth here should not be `core`

patzim: Done!

patzim:
@ugexe, @niner I'd really like to have some kind of approval of this PR before merging, as I don't have much of an opinion on this change. Shall I merge?

patzim:
I'm merging this now to at least have a little time before the release. If someone has objections, just revert.

PR #2942, opened 2019-06-01
Give SetHash a set/unset method

lizmat:
Inspired by seeing things like:

    %sethash{$_} = True for @to-be-added;

in the wild.  Instead, one can now do:

    %sethash.set(@to-be-added);
    %sethash.unset(@to-be-removed);

which is about 4x as fast.

### DIFF
--- a/src/core/SetHash.pm6
+++ b/src/core/SetHash.pm6
@@ -4,6 +4,23 @@ my class SetHash does Setty {
         Rakudo::Internals.PARAMETERIZE-KEYOF(base,type)
     }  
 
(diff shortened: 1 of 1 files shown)


Altai-man:
Isn't it work for the optimizer?

lizmat:
Yes,  Eventually.  I think  it will be quite some years before the optimizer is fast and smart enough to optimize this.

lizmat:
Implemented with https://github.com/rakudo/rakudo/commit/6df78ca3ed

PR #2946, opened 2019-06-02
Make subsets validate against their constraints same way as definites do.

vrurg:
Work on perl6/problem-solving#3


vrurg:
The PR is for demo purposes only. Not ready for merging.

Issue #2952, opened 2019-06-05
"use v6.*" no longer works


timo:
## The Problem

A recent change made "use v6.*" give an exception.

## Expected Behavior

Probably want it to select the highest language version that is not `.PREVIEW`.

## Actual Behavior

===SORRY!=== Error while compiling -e
No compiler available for Perl v6.*
at -e:1
------> use v6.*⏏<EOL>

timo:
impacted modules

| File | Code |
|--|--|
| [finanalyst/pod-cached<br>``…/PrecompilationDoc.pm6`` :*1*:](https://github.com/finanalyst/pod-cached/blob/5ec09b9320e4b98edcf068def75c2218c82692c3/lib/PrecompilationDoc.pm6#L1) | <code><b>use v6.\*</b>;</code> |
| [finanalyst/pod-cached<br>``…/100-heavy-duty.t`` :*2*:](https://github.com/finanalyst/pod-cached/blob/5ec09b9320e4b98edcf068def75c2218c82692c3/t/100-heavy-duty.t#L2) | <code><b>use v6.\*</b>;</code> |
| [jonathanstowe/Audio-Liquidsoap<br>``…/010-use.t`` :*3*:](https://github.com/jonathanstowe/Audio-Liquidsoap/blob/4c2f108deb1be9ce08f0b040d5d2b007b7f2a77b/t/010-use.t#L3) | <code><b>use v6.\*</b>;</code> |
| [jonathanstowe/Audio-Liquidsoap<br>``…/030-builtins.t`` :*3*:](https://github.com/jonathanstowe/Audio-Liquidsoap/blob/4c2f108deb1be9ce08f0b040d5d2b007b7f2a77b/t/030-builtins.t#L3) | <code><b>use v6.\*</b>;</code> |
| [jonathanstowe/Audio-Liquidsoap<br>``…/050-list.t`` :*3*:](https://github.com/jonathanstowe/Audio-Liquidsoap/blob/4c2f108deb1be9ce08f0b040d5d2b007b7f2a77b/t/050-list.t#L3) | <code><b>use v6.\*</b>;</code> |
| [tadzik/JSON-Unmarshal<br>``…/null.t`` :*3*:](https://github.com/tadzik/JSON-Unmarshal/blob/e1b6288c5f3165058f36c0f4e171cdf2dfd640da/t/null.t#L3) | <code><b>use v6.\*</b>;</code> |

ugexe:
[...]

PR #2953, opened 2019-06-06
Make v6.* work again.

vrurg:
- Also modifications like `v6.*.*`, or `v6.a+`, or `v6.e.*` or anything else
what `Version` accepts would work as expected. Contrary to incorrect
behavior of older releases now the highest version available will be
used. So that

  ```
  use v6.*;
  ```

  loads v6.e.PREVIEW now.

- Added `whatever` attribute to `Version` object. Similarly to `plus` it
indicates that version has a `Whatever` part.

- Added `language_modifier` attribute to `Perl6::Compiler`

rakudo/rakudo#2952

### DIFF
--- a/src/Perl6/Compiler.nqp
+++ b/src/Perl6/Compiler.nqp
@@ -4,6 +4,7 @@ use Perl6::Optimizer;
 
 class Perl6::Compiler is HLL::Compiler {
     has $!language_version;  # Default language version in form 6.c
+    has $!language_modifier; # Active language modifier; PREVIEW mostly.
     has $!language_revisions; # Hash of language revision letters. See gen/<vm>/main-version.nqp
     has $!can_language_versions; # List of valid language version
 
@@ -17,9 +18,13 @@ class Perl6::Compiler is HLL::Compiler {
     method language_name()    { 'Perl' }
     method reset_language_version() {
         $!language_version := NQPMu;
+        $!language_modifier := NQPMu;
     }
     method set_language_version($version) {
         $!language_version := $version;
+    }
+    method set_language_modifier($modifier) {
+        $!language_modifier := $modifier;
     }
     method language_version() {
         if nqp::defined($!language_version) {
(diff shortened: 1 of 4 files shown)


ugexe:
`6.*.*` shouldn't act different from how versions act elsewhere.

If someone says `use v6.*` let them get the latest version even if its `.PREVIEW`. Well written code will declare an exact version, just like well formed modules should use pinned versions.

vrurg:
This implementation is a result of today's IRC discussion. I would rather agree to [timo's note](https://colabti.org/irclogger/irclogger_log/perl6?date=2019-06-05#l625) and better not do what's not been asked for explicitly. `v6.*.*` is different because:

- it is unlikely to be used
- it is asking for a modifier and easing a restriction.

BTW `v6.*` won't load `6.d.PREVIEW` but will go for plain `6.d` for the same reason: two-part version is requested.

ugexe:
But that isn't how Versions work. You are creating specialized behavior for versions in a specific scenario. `6.*` is not a request for 1 extra version part -- it is a request for 0 or more version parts and is thus the same as `6.*.*`

vrurg:
I know. But if I blindly follow `Version`'s match then `6.d.TESTDEPR` would be loaded which is intended for test purposes only. Ok, I can skip this particular one (it's for deprecation testing), but then there is absolutely normal and valid `TEST` modifier. And `PREVIEW`. I'm not sure somebody would expect those to be loaded with `use v6.*;`

My point here is that this would follow DWIM rule for a decent user.

ugexe:
People using `use v6.*` are opting into the same behavior they would get from `use Some::Dependency:ver<1.*>` -- this includes `.PREVIEW`s being included. A more disciplined code base would simply use `use v6.d` or whatever they have tested on. So its not important to give special behavior to `6.*` because it should be discouraged (but not banned) in the first place.

vrurg:
Ok, let it match with full correspondence to `Version` rules. Gonna merge if travis passes.

PR #2958, opened 2019-06-07
Fix another use after free in the moar runner

Kaiepi:
This happens when NQP_HOME or PERL6_HOME are passed to the moar runner.


patzim:
After staring at this patch for some time and trying it out I don't understand the change.
You removed the logic to strip the excess `\\` or `/` of the home path. Why?
Can you clarify what this change intends to fix?

Kaiepi:
I can't replicate the use after free I was originally trying to fix. I'm guessing I was trying to build on an old branch or something.

Issue #1202, opened 2017-10-23
Async qqx sometimes hangs or dies ( await (^5).map({start { say qqx{… …} } }) )


AlexDaniel:
Code:
```perl6
await (^5).map({start { say qqx{echo -n foo $_} } })
```

But it does not always crash, so chances are you will need to run it like this in your shell:
```bash
while perl6 -e 'await (^5).map({start { say qqx{echo -n foo $_} } })'; do :; done
```

There are different ways you can observe the bug. Sometimes it hangs, sometimes it prints this:
```
foo 2
foo 4
foo 1
foo 0
Tried to get the result of a broken Promise
  in block <unit> at -e line 1

Original exception:
    This type (Nil+{Mu::Suggestion[Str]}) does not support elems
      in block  at -e line 1
```

The error message can be different. I've also seen ``This representation (Uninstantiable) does not support elems (for type (null))``.

The problem is reproducible on 2017.09 and HEAD(3f595acfb), so maybe it's not a regression (at least not a fresh one).

See also [RT #132287](https://rt.perl.org/Ticket/Display.html?id=132287) and especially [this comment](https://rt.perl.org/Ticket/Display.html?id=132287#txn-1501863).

lizmat:
Current state of affairs:

- Sample program runs indefinitely on an unloaded system
- As soon as there is some load, e.g. building, or running `make spectest`, it will usually crash with:
  - MoarVM panic: Heap corruption detected: pointer 0x10362bf48 to past fromspace (usually)
  - Cannot invoke null object  (very occasionally, see tracktrace below)

So I think there is still at least one, but maybe two issues in MoarVM related to async.

stack trace:

   at src/Perl6/World.nqp:3775  (/Users/liz/Github/rakudo.moar/install/share/nqp/lib/Perl6/World.moarvm:resolve)
 from SETTING::src/core/Mu.pm:76  (/Users/liz/Github/rakudo.moar/install/share/perl6/runtime/CORE.setting.moarvm:WHY)
 from gen/moar/Metamodel.nqp:1746  (/Users/liz/Github/rakudo.moar/install/share/nqp/lib/Perl6/Metamodel.moarvm:make_pun)
 from gen/moar/Metamodel.nqp:1773  (/Users/liz/Github/rakudo.moar/install/share/nqp/lib/Perl6/Metamodel.moarvm:find_method)
[...]

PR #2966, opened 2019-06-08
Added test for async bug

vrurg:
rakudo/rakudo#1202

### DIFF
--- a/t/02-rakudo/15-gh_1202.t
+++ b/t/02-rakudo/15-gh_1202.t
@@ -11,16 +11,16 @@ await Promise.anyof(
     start {
         my $out = "";
         # This was a bug with unpredictable output. Hopefully 50 repetiotions is enough to make sure everything works.
-        for 1..50 -> $rep {
-            my $proc = run $*EXECUTABLE, '-MTest', '-e', q<await (^5).map({start { print qqx{echo $_} } })>, :out;
-            $out ~= $proc.out.slurp;
+        for ^50 -> $rep {
+            my $proc = run $*EXECUTABLE.absolute, '-e', q<await (^5).map({start { print qqx{echo $_} } })>, :out;
+            $out ~= $proc.out.slurp(:close);
         }
         like $out, /[ <[0..4]> \n ] ** 250/, "shell output";
         $complete = True;
     },
     Promise.in(600), # 10mins timeout
 );
-    
+
 ok $complete, "all runs completed";
 
 # vim: ft=perl6 sw=4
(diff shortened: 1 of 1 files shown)


t/02-rakudo/15-gh_1202.t
```diff
@@ -0,0 +1,26 @@
...
+plan 2;
+
+my Bool $complete = False;
+
+await Promise.anyof(
+    start {
+        my $out = "";
+        # This was a bug with unpredictable output. Hopefully 50 repetiotions is enough to make sure everything works.
+        for 1..50 -> $rep {
+            my $proc = run $*EXECUTABLE, '-MTest', '-e', q<await (^5).map({start { print qqx{echo $_} } })>, :out;
```
ugexe: Always use `$*EXECUTABLE.absolute` with `run` or `shell`. Why are you using `-MTest`? Does that one-liner work on windows with its weird quoting rules?

t/02-rakudo/15-gh_1202.t
```diff
@@ -0,0 +1,26 @@
...
+            $out ~= $proc.out.slurp;
```
ugexe: Change to `$proc.out.slurp(:close)` or add a`$proc.out.close`

t/02-rakudo/15-gh_1202.t
vrurg: Ok. Test is just overlooked here, I was playing with `is` in the subprocess. Will remove.

t/02-rakudo/15-gh_1202.t
vrurg: Ok. Thanks!

t/02-rakudo/15-gh_1202.t
vrurg: As to Windows - it seems to be working under appveyor. I'll have a chance testing on my win 7 VM tomorrow.

ugexe: I tested on windows and it does what is expected. However the very first time I got the follow error (which I haven't replicated since):
```
C:\Users\ugexe\source>perl6 -e "my $proc = run $*EXECUTABLE, '-MTest', '-e', q<await (^5).map({start { print qqx{echo $_} } })>, :out;"

C:\Users\ugexe\source>An operation first awaited:
  in block <unit> at -e line 1

Died with the exception:
    Failed to write bytes to filehandle: Invalid argument
      in block  at -e line 1
```

vrurg: On my VM it never failed. I merge and close #1202.

Issue #2965, opened 2019-06-07
I can't HAS stub?


kazcw:
```perl6
use NativeCall;
class MyStub is repr(｢CStruct｣) { ... }
class SpecialDelivery:«💣» is repr(｢CStruct｣) {
    HAS MyStub $.stubby;
}
```
```Segmentation fault```

**Comment by kazcw** [comment]
I could understand the compiler not being happy with what I was trying to do 🐟, but segfaulting seems a bit much

**Comment by kazcw** [comment]
This is probably for the same reason:

```perl6
use NativeCall;
class Oops is repr(｢CStruct｣) { HAS Oops $.more; }
```

```Segmentation fault```

**Comment by MasterDuke17** [comment]
```
dan@hermes:~/p6$ pg6 -e 'use NativeCall;
> 
> class Oops is repr(｢CStruct｣) { HAS Oops $.more; }'
================================================================================================
This is Rakudo Perl 6 running in the  debugger, which often allows the user to generate useful back-
traces to debug or report issues in Rakudo, the MoarVM backend or the currently running code.

This Rakudo version is 2019.03.1.574.gee.2.f.2.d.2.df built on MoarVM version 2019.05.36.ge.865.f.43.df,
running on ubuntu (19.04.Disco.Dingo) / linux (17.Ubuntu.SMP.Wed.May.15.10.52.21.UTC.2019)

Type `bt full` to generate a backtrace if applicable, type `q` to quit or `help` for help.
------------------------------------------------------------------------------------------------
Reading symbols from /home/dan/Source/perl6/install/bin/perl6-m...
Starting program: /home/dan/Source/perl6/install/bin/perl6-m -e use\ NativeCall\;'
''
'class\ Oops\ is\ repr\(｢CStruct｣\)\ \{\ HAS\ Oops\ \$.more\;\ \}
[Thread debugging using libthread_db enabled]
Using host libthread_db library "/lib/x86_64-linux-gnu/libthread_db.so.1".
[New Thread 0x7ffff6ff6700 (LWP 21302)]

Thread 1 "perl6-m" received signal SIGSEGV, Segmentation fault.
compute_allocation_strategy (st=<optimized out>, repr_data=<optimized out>, repr_info=<optimized out>, tc=0x55555555ae90) at src/6model/reprs/CStruct.c:263
263                             bits                                      = cppstruct_repr_data->struct_size * 8;
(gdb) bt
[...]

PR #2967, opened 2019-06-08
add tests for #2965

kazcw:
This will segfault until it's built on MoarVM/MoarVM#1116

### DIFF
--- a/t/04-nativecall/23-incomplete-types.t
+++ b/t/04-nativecall/23-incomplete-types.t
@@ -1,0 +1,22 @@
+use v6;
+
+use MONKEY-SEE-NO-EVAL;
+
+use lib <lib>;
+use NativeCall :TEST;
+use Test;
+
+plan 6;
+
+for (<CUnion CStruct CPPStruct>) -> $cls {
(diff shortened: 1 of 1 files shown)


kazcw:
Rebased to rerun CI now that the fix is in Moar.

kazcw:
The tests with `--gen-moar=master` are now passing, except for [a problem with nmake in AppVeyor?](https://ci.appveyor.com/project/rakudo/rakudo/builds/25152301/job/b13vxaykxil985i3)

ugexe:
I reran the appveyor test, and master is now failing on `t\04-nativecall\04-pointers.t`. Not sure if its related or not.

vrurg:
`04-pointers.t` fails for quite some time now. It is not related to nmake issue, that's for sure.

As to nmake, look weird. As if `nmake` printed something different from it's version information. I would guess it failed in some way and ended with an error message.

ugexe:
@vrurg I get failures every single time I run that test on windows. This does not happen on `2019.03.1` so something is definitely broken.

PR #2997, opened 2019-06-18
Fix for line() not using @*ARGS within MAIN

vrurg:
This makes an example from Learning Perl6 book by brian d foy work as
expected. Before the fix lines() within MAIN was always reading from
$*IN.


vrurg:
Closing because this is the behavior of v6.d+.

PR #3013, opened 2019-06-21
Correct assignment to SetHash

vrurg:
rakudo/rakudo#1203

### DIFF
--- a/src/core/SetHash.pm6
+++ b/src/core/SetHash.pm6
@@ -1,8 +1,8 @@
 my class SetHash does Setty {
 
-    method ^parameterize(Mu \base, Mu \type) { 
+    method ^parameterize(Mu \base, Mu \type) {
         Rakudo::Internals.PARAMETERIZE-KEYOF(base,type)
-    }  
+    }
 
 #--- selector methods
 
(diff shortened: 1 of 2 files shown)


src/core/SetHash.pm6
```diff
@@ -203,6 +203,14 @@ my class SetHash does Setty {
     multi method Mixy (SetHash:D:) { self.MixHash }
 
 #--- interface methods
+    multi method STORE(SetHash:D: Setty:D \set --> SetHash:D) {
+        nqp::say("SetHash STORE from Setty") if nqp::getenvhash<RAKUDO_DEBUG>;
```
MasterDuke17: This looks like a debugging leftover.

vrurg: It surely does. Thanks!

MasterDuke17:
It also looks like this is pulling in some commits related to #2178?

src/core/SetHash.pm6
```diff
@@ -203,6 +203,14 @@ my class SetHash does Setty {
...
+        self.SET-SELF(
```
MasterDuke17: Also, this new candidate doesn't need the `.is-lazy` check that the existing candidate has?

vrurg: I think a `Setty` cannot be lazy.

vrurg:
Looks like I need to redo this PR and rebase it on master. Thank for noticing!

Issue #3012, opened 2019-06-21
Some issue after “Decontainerize non-scalar symbols upon import” commit


AlexDaniel:
Pastebin::Shadowcat no longer passes its test.

Bisected: https://github.com/rakudo/rakudo/commit/3af3bde70eaed13b235bc55c54cf0ca604be6fcf

Error:
```
Cannot invoke this object (REPR: Null; VMNull)
  in sub decode at /root/Blin/installed/HTML::Entity_*/sources/E6858CA39AB81D28F90ED1A7D6D1DC66D0E9FAA5 (HTML::Entity) line 15
  in method fetch at /root/Blin/data/zef-data/store/perl6-Pastebin-Shadowcat.git/7e8769e51fd27618681b2d6be623c6c64c996a9b/lib/Pastebin/Shadowcat.pm6 (Pastebin::Shadowcat) line 30
  in block <unit> at t/01-paste.t line 12
```

Related pieces of code:
* https://github.com/perl6-community-modules/perl6-Pastebin-Shadowcat/blob/7e8769e51fd27618681b2d6be623c6c64c996a9b/lib/Pastebin/Shadowcat.pm6#L31
* https://github.com/Mouq/HTML-Entity/blob/57931c74f08344fe06459368d80f2b3a16cc5dee/lib/HTML/Entity.pm6#L25

Seems like something is not right in rakudo and the module is not at fault.

Ping @vrurg.

vrurg:
I have golfed it down to [this](https://gist.github.com/vrurg/6f7d4940a62648e76b28a429c39d1113). The point is that my commit is unrelated in this case. I have reverted it on a local branch and the outcome is still the same – `Cannot invoke ...`. My commit could have resulted in more predictable behavior of the inital problem.

Debug printing of `&test-test.WHAT` reveals that within `Baz` module it's `Sub`; whereas in `cannot_invoke.p6` it turns into `Callable`.

What could possibly be the good outcome of this is that I probably know where to look for the source of containerization problem which I was trying to fix with `3af3bde`.

vrurg:
Final for today. The problem is in binding. `is export` trait is applied before binding takes place. Correspondingly, the `Scalar` container with `Callable` type object gets exported. Afterwards `&test-test` is bound to `&test` and by this moment it's no more linked to the exported `Scalar`.

[...]

Issue #2979, opened 2019-06-13
Re-import from a 3rd module results in installation of containerized symbols.


vrurg:
## The Problem

Exporting symbols using `EXPORT` routine results in them being installed in containerized form; i.e. wrapped into a `Scalar`). Under certain circumstances this may result in failing mutli-dispatch due to `derive_dispatch` method trying to act on a `Scalar` instance.

Occasionally similar behavior was observed with `is export` trait.

## Steps to Reproduce

_Foo.pm6_:

```
unit module Foo;

our @foo is export;
sub foo is export { }
```

_Bar.pm6_:

```
use Foo;
our @b;
sub b { }

multi EXPORT {
     return %(
         Foo::EXPORT::ALL::,
         '&bar' => &b,
         '@bar' => @b,
     );
}
```

Script:
```
use Bar;
say @foo.VAR.^name;
say &foo.VAR.^name;
say @bar.VAR.^name;
say &bar.VAR.^name;
```

This would output:

>```
>Scalar
>Scalar
>Scalar
>Scalar
>```

Where expected output would be:

>```
>Array
>Sub
>Array
>Sub
>```

## Environment
* Operating system: macOS
* Compiler version (`perl6 -v`): This is Rakudo version 2019.03.1 built on MoarVM version 2019.03

vrurg:
I have a fix, the ticket is just to have an anchor for tests and PR.

vrurg:
Closed via #2980

PR #3016, opened 2019-06-22
Implement more fine-grained deconting of exported symbols

vrurg:
The problem with containerized exports has been narrowed down to EXPORT sub using hash for declaring exports where all values are getting wrapped into Scalars. To fix that method import on World got a named parameter :need-decont which signals that imported symbols need to be checked for their sigils. Unless a sigil is $ or & the symbol value is nqp::decont'ed.

This commit works for both rakudo/rakudo#2979 and rakudo/rakudo#3012.


niner:
So the error is actually on the exporting side? Why not fix it there instead? What if someone actually wants to export a container?

vrurg:
> So the error is actually on the exporting side? Why not fix it there instead? What if someone actually wants to export a container?

How can I fix it on the exporting side when `EXPORT` must return a hash where _everything_ is containerized? Moreover, if you re-export using `Foo::EXPORT::ALL::` or anything like that – you also get containerized symbols no matter what `is export` was applied to.

The only ultimate solution I see here is a `Map` type which would always bind its values. And then we would have to demand `EXPORT` to use it.

niner:
EXPORt must return a Map, not necessarily a Hash. That's both documented: https://doc.perl6.org/language/modules#index-entry-sub_EXPORT and done so in the wild: https://github.com/niner/Inline-Perl5/blob/master/lib/Inline/Perl5.pm6#L941

lizmat:
my %m is Map = a => 42; %m<a> = 666;  # Cannot change key 'a' in an immutable Map

vrurg:
Ok, unless I have done something wrong, `Map` is really the solution for correct exporting. Moreover, it also correctly preserves re-exported symbols. I was worried if `FLATTENABLE_HASH` don't mangle with values, but it seemingly doesn't.

But this doesn't help with cases where people are returning `Hash`es. Perhaps it would make sense to make an additional check and do deconting only if a `Hash` is returned? To have DWIM in action because many seem to not understand the clear difference between the two.

vrurg:
**UPDATE** Exporting via `Map` has a funny side effect. What is exported this way becomes immutable too. Basically, `Map` strips off containers on its own!

```
module Bar {
    our $bar = 42;
}

multi EXPORT {
    my %ex is Map = (
        '$bar' => $Bar::bar,
    );
    %ex;
}
```

```
use Bar;

say $bar.VAR.^name; # Int
$bar = 1; # Cannot assign to an immutable value
```

I don't see any alternative for now. And it's time to move to my gate anyway. ;)

src/Perl6/World.nqp
```diff
@@ -1068,7 +1068,7 @@ class Perl6::World is HLL::World {
                 my $Map := self.find_symbol(['Map'], :setting-only);
                 if nqp::istype($result, $Map) {
                     my $storage := $result.hash.FLATTENABLE_HASH();
-                    self.import($/, $storage, $package_source_name, :need-decont);
+                    self.import($/, $storage, $package_source_name, :need-decont(!($result =:= $Map)));
```
niner: Shouldn't that be $result.WHAT?

vrurg: Sure it is. Thanks!

PR #3020, opened 2019-06-24
Fix PseudoStash.WHICH's signature

### LINKED ISSUE #3018: Regression in PseudoStash.WHICH
## The Problem
Calling `.WHICH` on a `PseudoStash` in certain cases throws `Cannot call method 'WHICH' on a null object`.

## Expected Behavior
It should always return the `PseudoStash`'s object ID.

## Actual Behavior
It throws in certain cases.

## Steps to Reproduce

```
bastille% perl6 -e 'say MY.WHO.WHICH'
PseudoStash|FDE22CC964C44A1649BA67F0A1D501B1616A4049
bastille% perl6 -e '{ say MY.WHO.WHICH }'
Cannot call method 'WHICH' on a null object
  in block <unit> at -e line 1
```

## Environment
* Operating system: 
OpenBSD bastille.kennel.qt 6.5 GENERIC.MP#11 amd64
This is Rakudo version 2019.03.1-643-g0c52bfed0d built on MoarVM version 2019.05-91-g81e5cbf2a
implementing Perl 6.d.

Kaiepi:
It included PseudoStash:D as an argument when it's supposed to be the
type of the object itself.

Fixes #3018

### DIFF
--- a/src/core/PseudoStash.pm6
+++ b/src/core/PseudoStash.pm6
@@ -20,7 +20,7 @@ my class PseudoStash is Map {
         $obj
     }
 
-    multi method WHICH(PseudoStash:D --> ObjAt:D) { self.Mu::WHICH }
+    multi method WHICH(PseudoStash:D: --> ObjAt:D) { self.Mu::WHICH }
 
     my $pseudoers := nqp::hash(
         'MY', sub ($cur) {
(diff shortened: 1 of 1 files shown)


lizmat:
Good catch!

PR #3021, opened 2019-06-25
Warn when x..^* range is used as subscript

### LINKED ISSUE #3010: 1..^* needs to give a warning
See https://github.com/rakudo/rakudo/commit/35b69f074b9db3ffeb1638fc8a548fc2c0643d5a and related tickets.

Basically, `1..^*` is identical to `1..*` meaning that in the most cases it is a user error (it does not DWIM). Therefore, it needs an error message that suggests to use `*-2`.

taboege:
Excluding the last element of an infinite range subscript has no effect
and is likely a thinko. Print a warning and suggest to use *-2 as upper
endpoint to omit the last item.

Fixes #3010

### DIFF
--- a/src/core/array_slice.pm6
+++ b/src/core/array_slice.pm6
@@ -55,10 +55,15 @@ multi sub POSITIONS(
     }
 
     # we can optimize `42..*` Ranges; as long as they're from core, unmodified
-    my \is-pos-lazy = pos.is-lazy;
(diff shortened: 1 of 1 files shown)


lizmat:
Please refrain from merging until we have a decision: https://github.com/perl6/problem-solving/issues/50

AlexDaniel:
Yeah, I'll just close it for now. Once https://github.com/Raku/problem-solving/issues/50 is resolved we can come back to this.

Issue #3028, opened 2019-07-03
No new symbols could be defined in CORE.<letter>.setting


vrurg:
## The Problem

A new symbol defined in either `CORE.d.setting` or `CORE.e.setting` is not available to the user code.

This is an anchor issue for a fix in progress.

vrurg:
Related issue: #2432

vrurg:
Closed via #3040

PR #3040, opened 2019-07-10
Implement support for defining new symbols in CORE.e.setting

vrurg:
Because loading of all CORE settings caused some tests to fail due to the way `SETTING` and `CORE` pseudo-packages behaved, this fix also includes improvements to `PseudoStash` class. Most notable changes are:

- `PSEUDO::<missing-symbol>` now returns `Failure` where previously it was `Nil`. Done so to conform to the way `::()` works.
- `LEXICAL::` now includes all symbols from scope's lexical chains and dynamic symbols from its caller chain. This conforms to the rule "all symbols visible at location"
- `SETTING` now includes all symbols from all `CORE.setting`s.
- `CORE` now includes few additional symbols installed by `World.loading_and_symbol_setup`. Not sure about them, but seem to be harmless and doesn't affect roast though was breaking `rakudo` test.
- `UNIT` now includes all lexicals visible at the unit level.

- Also binding via chained pseudos is now supported:

  ```
  my $*foo;
  sub do-bind { CALLERS::<$*foo> := 42 }
  do-bind;
  say $*foo; # 42
  ```

- `EVAL` doesn't load `CORE.setting` anymore. Effectively, `use v6.X` inside `EVAL`s is just ingnored. If this approach is not getting re-considered I would suggest making it a `X::Language::TooLate` exception.

   Also `EVAL` context now gets !EVAL_MARK` symbol alongside with `!UNIT_MARK`. It is necessary to allow `SETTING` pseudo find the real unit.

- `die` and `fail` now work with `$!` as default. The symbol is been searched in `CALLER::LEXICAL::` as `S29-context/die.t` spectest suggests.

- new sub in `CORE`: `CORE-SETTING-REV`. Returns currently active language revision. It's purpose is to support rakudo tests.

- Fixed a hidden bug where `request` was always installing submodules into its `CALLER::MY::` context. The bug was hidden until pseudo-packages started returning `Failure` for missing symbols.


vrurg:
Fix for #3028

src/Perl6/Actions.nqp
```diff
@@ -6613,35 +6615,26 @@ class Perl6::Actions is HLL::Actions does STDActions {
...
+            my $numsemis := +$<semilist><statement>;
             if $numsemis > 1 {
                 $past := QAST::Stmts.new( :node($/) );
                 $past.push(QAST::Op.new( :op('call'), :name('&infix:<,>'), :node($/)));
             }
-
-            my $semi := -1;
-            while ++$semi < $numsemis {
-                my $EXPR := @EXPR[$semi];
+            my $semi := 0;
```
lizmat: Feels like you missed a commit (a26e95bab41e7b787c) that changed this.  Is there a reason you want to go back to the old behaviour here?

src/Perl6/Actions.nqp
```diff
@@ -6882,6 +6863,59 @@ class Perl6::Actions is HLL::Actions does STDActions {
...
+        # Look for any chained adverb pairs and relocate them.
+        # Try to reuse existing QAST where possible.
+        elsif $*FAKE_INFIX_FOUND {
+            my $numsemis := +$<semilist><statement>;
+            if $numsemis > 1 {
+                $past := QAST::Stmts.new( :node($/) );
+                $past.push(QAST::Op.new( :op('call'), :name('&infix:<,>'), :node($/)));
+            }
+            my $semi := 0;
+            repeat until $semi >= $numsemis {
```
lizmat: Same comment here, why revert a26e95bab41e7b787c8 ?

src/Perl6/Optimizer.nqp
```diff
@@ -20,7 +20,7 @@ my class Symbols {
     # Some interesting scopes.
     has $!GLOBALish;
     has $!UNIT;
-    has $!SETTING;
+    # has $!SETTING;
```
lizmat: Why comment out when it is not needed anymore?

src/Perl6/Optimizer.nqp
```diff
@@ -260,28 +260,47 @@ my class Symbols {
...
 
+    # method find_in_setting($symbol) {
+    #     if !nqp::defined($!SETTING) {
+    #         my int $i := +@!block_stack;
+    #         while $i > 0 && !nqp::defined($!SETTING) {
+    #             $i := $i - 1;
+    #             my $block := @!block_stack[$i];
+    #             my %sym := $block.symbol("!CORE_MARKER");
+    #             if +%sym {
+    #                 $!SETTING := $block;
```
lizmat: NQP does support `last`.  Put it here, then you could remove the `!defined($!SETTING)` from the while condition for more simple code.

src/Perl6/Optimizer.nqp
```diff
@@ -260,28 +260,47 @@ my class Symbols {
...
-                }
-            }
-            if !nqp::defined($!SETTING) {
-                nqp::die("Optimizer couldn't find CORE while looking for $symbol.");
-            }
-        } else {
-            if nqp::existskey(%!SETTING_CACHE, $symbol) {
-                return %!SETTING_CACHE{$symbol};
-            }
+        if nqp::existskey(%!SETTING_CACHE, $symbol) {
```
lizmat: Am I seeing the same method twice ??

src/Perl6/Actions.nqp
vrurg: That's confusing. The commit is actually there vrurg/rakudo@a26e95bab41e7b787c8283f7ae24ae3859727cfe. Perhaps a result of chained migration macbook -> mac mini -> macbook with help of backups and rsync.

src/Perl6/Optimizer.nqp
vrurg: Temporary state of things. Explanation is below.

src/Perl6/Optimizer.nqp
vrurg: The previous version is commented out. Mine is still active. I wasn't sure if my version is not flawed and temporarily left the old one in the comments.

Thanks, Liz!

BTW, this all is coming from attempt to make `QuantHash`es iterables in 6.e.

src/Perl6/Actions.nqp
vrurg: I have re-applied related commits by cherry-picking them.

jnthn:
> MY:: now includes all symbols from its lexical chains and dynamic symbols from its caller chain

Hm, I'd expect `MY::` to only include lexical things (with the proviso that `my $*FOO` is actually a lexical)?

jnthn:
> SETTING is now includes all symbols from all CORE.settings. Not sure if this is right. 

Good question. I guess if it doesn't, then we'd want a `SETTINGS` that did. What does `CORE` do with regard to this question? Include everything from all the CORE.setting.*?

src/core/PseudoStash.pm6
```diff
@@ -25,10 +27,8 @@ my class PseudoStash is Map {
     my $pseudoers := nqp::hash(
         'MY', sub ($cur) {
             my $stash := nqp::clone($cur);
-            nqp::bindattr_i($stash, PseudoStash, '$!mode', PRECISE_SCOPE);
-            nqp::setwho(
-                Metamodel::ModuleHOW.new_type(:name('MY')),
-                $stash);
+            nqp::bindattr_i($stash, PseudoStash, '$!mode', STATIC_CHAIN +| DYNAMIC_CHAIN);
```
jnthn: I think `PRECISE_SCOPE` is right for `MY`.

src/core/PseudoStash.pm6
```diff
@@ -74,17 +74,13 @@ my class PseudoStash is Map {
                 nqp::bindattr($stash, Map, '$!storage', nqp::ctxlexpad($ctx));
                 nqp::bindattr($stash, PseudoStash, '$!ctx', $ctx);
                 nqp::bindattr_i($stash, PseudoStash, '$!mode', PRECISE_SCOPE);
```
jnthn: Since I think really `MY::` is meant to just be like `OUTER::`, but without walking a step of the chain.

vrurg:
> > MY:: now includes all symbols from its lexical chains and dynamic symbols from its caller chain
> 
> Hm, I'd expect `MY::` to only include lexical things (with the proviso that `my $*FOO` is actually a lexical in terms of its storage)?

That'd be my guess too. But according to `S02-names/pseudo.t`, lines 100+, `MY` is expected to include dynamics from its outer. In a way it makes sense if we interpret `MY` as what is visible in the current scope.

Making `MY` PRECISE_SCOPE breaks the test. So, we either shall remove it from the roast or leave things as they're in this PR.

vrurg:
> > SETTING is now includes all symbols from all CORE.settings. Not sure if this is right.
> 
> Good question. I guess if it doesn't, then we'd want a `SETTINGS` that did. What does `CORE` do with regard to this question? Include everything from all the CORE.setting.*?

This is my weak point here as many aspects are not clear to me yet. `CORE`, I suppose, must do exactly what you say. But what would make it different from `SETTINGS` then?

vrurg:
> Hm, I'd expect `MY::` to only include lexical things (with the proviso that `my $*FOO` is actually a lexical in terms of its storage)?

Another uncertainty:

```
sub foo {
    my $foo;

    {
        say MY::<$foo>;
    }
}
```

Shall this fail? It will if `MY` is `PRECISE_SCOPE`.

vrurg:
The top comment has been updated to reflect latest changes suggested by @jnthn and @lizmat

vrurg:
Related topic: perl6/problem-solving#58

vrurg:
At this point any all known bugs revealed by this PR were fixed, all spectests are passing with perl6/roast#557 applied. I consider this PR ready for merge.

Ping @jnthn

PR #3042, opened 2019-07-11
Issue 3041: Propagate laziness in KeyValue/Pair Iterators

jstuder-gh:
Propagate the laziness value from the base iterator used.

Otherwise using .kv, .pairs, or .antipairs on a lazy list would cause it
to be eagerly evaluated.


lizmat:
Good catch!  Thanks!

PR #3044, opened 2019-07-11
Implement NativeCall wide string support

Kaiepi:
This does two things, mainly:
- implements support for `wchar_t` and `wint_t`, as well as C++'s `char16_t` and `char32_t` types. These are accurate on MoarVM, but may be unreliable on the JVM and JS backends.
- implements wide string support for the `P6str` and `CStr` REPRs on MoarVM.

This needs testing on Windows.

Passes `make test`, `make spectest` on OpenBSD.

See https://github.com/MoarVM/MoarVM/pull/1137 and https://github.com/perl6/nqp/pull/564

### DIFF
--- a/lib/NativeCall.pm6
+++ b/lib/NativeCall.pm6
@@ -59,7 +59,13 @@ sub string_encoding_to_nci_type(\encoding) {
         ?? "asciistr"
         !! nqp::iseq_s($enc,"utf16")
           ?? "utf16str"
-          !! die "Unknown string encoding for native call: $enc"
+          !! nqp::iseq_s($enc, 'wide')
+            ?? "widestr"
+            !! nqp::iseq_s($enc, 'u16')
+              ?? "u16str"
+              !! nqp::iseq_s($enc, 'u32')
+                ?? "u32str"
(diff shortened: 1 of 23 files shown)


raiph:
Note https://stackoverflow.com/questions/57584155/how-to-properly-use-char-and-wchar-t-in-nativecall-perl6#comment101984459_57584155 response to your comment

Kaiepi:
Update on this: I do plan on completing this PR at some point, since I want editline support for the REPL eventually, but I've been caught up with the work for the IP6NS grant, among other things.

lizmat:
Is this PR still salvageable?

lizmat:
It feels like all of Kaiepi's work will be lost unless someone reimagines this work.

@patrickbkr @timo @niner

zhuomingliang:
If it doesn't break test, I would suggest merging into main branch after resolving conflicts, and improving it later, so these work won't be lost.

zhuomingliang:
If it doesn't break test, I would suggest merging into main branch after resolving conflicts, and improving it later, so these work won't be lost.

PR #3055, opened 2019-07-16
Re-implement CORE loading sequence

vrurg:
PR #3040 wasn't a complete fix to the new symbols issue. Compile-time
symbols were still fetched from CORE.setting by the compiler and
installed as if no other settins existed.

This fix postponed CORE binding to the $*UNIT_OUTER until lang-version
token is parsed. For symbols required to pass this stage (like Version
class) a new method on World find_symbol_in_setting tries to load
appropriate CORE and fetch the requested symbol directly.

Until CORE is bound to $*UNIT_OUTER the unit is not considered
ready.


vrurg:
Note: roast is not yet ready for this PR. Will follow soon.

t/08-performance/04-settingkeys.t
```diff
@@ -5,7 +5,7 @@ plan 1;
 my %allowed = (
   Q{!CORE_MARKER},
   Q{Int},
-  Q{&CORE-SETTING-REV},
+  Q{CORE-SETTING-REV},
```
lizmat: Just wondering, but do we need a `!CORE_MARKER` still?  Or vice-verse, couldn't we just put the revision info in `!CORE_MARKER` ?

vrurg: Probably not any more. I wouldn't put version into `!CORE_MARKER` because it would not be directly accessible. Besides, CORE-SETTING-REV allows for testing of constant declaration in COREs.

If no other modules somehow rely on `!CORE_MARKER` I can pull it out later.

Issue #3036, opened 2019-07-08
Diagnostic or definition problem?


alanrocker:
## The Problem
It's not clear what bounding characters are legitimate for multi-line comments
## Expected Behavior
Ambiguous: Either accept character following ` as valid, or complain about it at the start

## Actual Behavior 
For #`! ......
===SORRY!=== Error while compiling /home/guru/bin/./trivial
Prefix !
 requires an argument, but no valid term found
at /home/guru/bin/./trivial:21
------> !⏏<EOL>
    expecting any of:
        prefix

For #`@
===SORRY!=== Error while compiling /home/guru/bin/./trivial
Undeclared name:
    Doesn't used at line 32
Undeclared routines:
    either used at line 32. Did you mean 'ENTER', 'GATHER'?
    work used at line 32

## Steps to Reproduce
Run the following program, (with correct #!) then remove the comments around the examples.
Note the line numbers may not match the ones in the examples, depending on editing

#! /home/guru/bin/perl6
# program trivial 

#`(
  multi-line {1} works
  )

#`[
  another {2} works
  ]

#`{
  another (3) works
}

=begin comment

#`!
  Doesn't work
!

#=end comment
#=begin comment
#`@
  Doesn't work either
@

=end comment

# end trivial
                         
## Environment
* Operating system: Linux localhost 4.14.131-desktop-1.mga6 #1 SMP Thu Jun 27 11:19:36 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux

* Compiler version (`perl6 -v`): 
Rakudo Star version 2019.03.1 built on MoarVM version 2019.03

tbrowder:
What you refer to as "multi-line comments" are in fact "declarator blocks." The docs clearly state that a declarator block uses "matched pairs of parenthesis constructs such as ()..."  But I agree a completely delineated list would be more appropriate. 

Also declarator blocks are not defined inside other pod blocks, even "=begin comment" blocks.

alanrocker:
> The docs clearly state that a declarator block uses "matched
> pairs of parenthesis constructs such as ()..."  But I agree a completely
> delineated list would be more appropriate.

In that case, perhaps the compiler should test for an invalid character
after the `, and flag it with "Invalid character for delimiting block, or
something similar, rather than exploding at random later?

> Also declarator blocks are not defined inside other pod blocks, even
> "=begin comment" blocks.

I provided the code to be uncommented to generate the error messages, in
contrast to the 3 working examples.

[...]

PR #3056, opened 2019-07-16
Improve error messages for invalid open brackets

tbrowder:
Fixes GH #3036

### DIFF
--- a/src/Perl6/Grammar.nqp
+++ b/src/Perl6/Grammar.nqp
@@ -4670,13 +4670,24 @@ if $*COMPILING_CORE_SETTING {
        '#' {} \N*
     }
 
+    # an in-line or multi-line comment (aka 'embedded comment')
+    # we panic when a non-opening bracket char follows the sym
+    token comment:sym<#`> {
+        '#`' 
+        <.typed_panic: 'X::Syntax::Comment::Embedded'> 
+    }
     token comment:sym<#`(...)> {
         '#`' <?opener> {}
         [ <.quibble(self.slang_grammar('Quote'))> || <.typed_panic: 'X::Syntax::Comment::Embedded'> ]
     }
 
+    # leading declarator blocks 
+    # we panic when a non-opening bracket char follows the sym
     token comment:sym<#|(...)> {
-        '#|' <?opener> <attachment=.quibble(self.slang_grammar('Quote'))>
+        '#|' 
+        [ <?opener> | <.typed_panic('X::Syntax::Pod::Declarator::Leading')> ] 
+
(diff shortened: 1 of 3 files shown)


lizmat:
Could you give some examples of before / after ?

tbrowder:
Yes, I'll copy the gist text here.

src/Perl6/Grammar.nqp
```diff
@@ -4670,13 +4670,24 @@ if $*COMPILING_CORE_SETTING {
...
         '#`' <?opener> {}
         [ <.quibble(self.slang_grammar('Quote'))> || <.typed_panic: 'X::Syntax::Comment::Embedded'> ]
     }
 
+    # leading declarator blocks 
+    # we panic when a non-opening bracket char follows the sym
     token comment:sym<#|(...)> {
-        '#|' <?opener> <attachment=.quibble(self.slang_grammar('Quote'))>
+        '#|' 
+        [ <?opener> | <.typed_panic('X::Syntax::Pod::Declarator::Leading')> ]
```
timo: this should perhaps use `||` instead of `|`?
```suggestion
        [ <?opener> || <.typed_panic('X::Syntax::Pod::Declarator::Leading')> ] 
```

src/core/Exception.pm6
```diff
@@ -1528,6 +1528,14 @@ my class X::Syntax::Comment::Embedded does X::Syntax {
     method message() { "Opening bracket required for #` comment" }
 }
 
+my class X::Syntax::Pod::Declarator::Leading does X::Syntax does X::Pod {
+    method message() { "Opening bracket required for #| declarator block" }
+}
+
+my class X::Syntax::Pod::Declarator::Trailing does X::Syntax does X::Pod {
```
timo: I wonder if perhaps these two exception classes would work better as a single one with an attribute to declare what kind of comment it is; probably doesn't want to be merged with the one above since that's not Pod-related

tbrowder: I'll do that.

tbrowder: Hm, I have to learn more about Exception classes to do that.

timo: should be enough to add a named argument to the `<.typed_panic ...>` and have an attribute in the exception class, then use that attribute in the message method to change how it is displayed

tbrowder: Okay.

tbrowder:
I'm having trouble trying to make an exception to handle both types of declarator blocks. The main problem is not finding the correct way to handle the trailing declarator type. I may have to give up that effort.

tbrowder:
Closing this PR and will resubmit as a new one after I clean up code

Issue #3070, opened 2019-07-19
String interpolation with block evaluates to declared return value of enclosing routine, if any


chloekek:
## Expected Behavior

```perl6
sub f(--> Nil) { say "{1}"; }
sub g(--> 2) { say "{1}"; }
sub h { say "{1}"; }
f; # OUTPUT: «1␤»
g; # OUTPUT: «1␤»
h; # OUTPUT: «1␤»
```

## Actual Behavior

```perl6
sub f(--> Nil) { say "{1}"; }
sub g(--> 2) { say "{1}"; }
sub h { say "{1}"; }
f; # OUTPUT: «Use of Nil in string context␤␤  in sub f at <tmp> line 1␤»
g; # OUTPUT: «2␤»
h; # OUTPUT: «1␤»
```

## Environment

* Operating system: Ubuntu GNU/Linux 16.04.
* Compiler version (`perl6 -v`): This is Rakudo version 2019.03 built on MoarVM version 2019.03 implementing Perl 6.d.

## P.S.

If confirmed a bug I'd like to add this to Roast to get familiar with the workflow and maybe even fix it in Rakudo.

AlexDaniel:
Yes, it's clearly a bug. If you're on IRC (or Matrix, or Discord), then you can ask `committable6` to run that code snippet on `all` releases ([log](https://colabti.org/irclogger/irclogger_log/perl6-dev?date=2019-07-19#l144)):

> \<AlexDaniel\> c: all `sub f(--> Nil) { say "{1}"; }; sub g(--> 2) { say "{1}"; }; sub h { say "{1}"; }; f; g; h;`
> \<committable6\> AlexDaniel, https://gist.github.com/5d984e23482807fb04214c05e10e3d6f

So we can see it's not a regression, and it has always been like that. Weird. So yes, a fix is welcome :) I'll assign you to this issue.

JJ:
I'm not so sure this is a bug, or at least all of them, or at least that type of bug. 
For instance, let's check this out:
```perl6
sub f( --> Nil ) { say "hey" }; say f
# hey
# Nil
```
That's OK. It works in the same way if we just use `f`. It will return Nil in a void context, we don't care about it, that's fine. But we are using a block inside a string. Well.
```perl6
sub f() { say "hey" }; say f
hey
True
```
This is what happens. It returns True, because that is what say does, runs the body, that's fine too. We turn that into a block:
```perl6
sub f() { say "{'hey'}" }; say f
hey
True
[...]

PR #3072, opened 2019-07-20
Clear signature before parsing block

chloekek:
Fixes https://github.com/rakudo/rakudo/issues/3070.
Fudged test at https://github.com/perl6/roast/pull/562/files.

If we do not do this then the block takes on the signature of the enclosing routine, leaking %*SIG_INFO<returns> into the block and causing the block to return any declared return value of the enclosing sub.

The token pblock also resets $*POD_BLOCK, $*DOC, and $*LINE_NO, but the token block does not. I don't know if that's necessary.

I'm also not entirely sure why the issue only seems to occur for return values and only inside string interpolation (see the linked issue). Do all other places use pblock instead?

### DIFF
--- a/src/Perl6/Grammar.nqp
+++ b/src/Perl6/Grammar.nqp
@@ -1038,6 +1038,8 @@ grammar Perl6::Grammar is HLL::Grammar does STD {
     token block($*IMPLICIT = 0) {
         :my $*DECLARAND := $*W.stub_code_object('Block');
         :my $*CODE_OBJECT := $*DECLARAND;
+        :my $*SIG_OBJ;
+        :my %*SIG_INFO;
         :dba('scoped block')
         :my $borg := $*BORG;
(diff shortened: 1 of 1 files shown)


jnthn:
> The token pblock also resets $*POD_BLOCK, $*DOC, and $*LINE_NO, but the token block does not. I don't know if that's necessary.

It's consistent with `pblock` having logic in place to set docs on the pointy block, and `block` having that possibility. Since `block` isn't doing that, I don't *think* it needs to reset those - it will be done by the next documentable entity.

> I'm also not entirely sure why the issue only seems to occur for return values and only inside string interpolation (see the linked issue). Do all other places use pblock instead?

Well, certainly `circumfix:<{ }>` - which is the common case for anonymous blocks - calls `pblock`. The quote language, by contrast, does call `block`, and so do a number of other places, such as the statement prefixes.

Issue #3073, opened 2019-07-21
Cannot do 'say' on a handle in binary mode when using it on closed handle


voldenet:
## The Problem
```perl6
m: my $a = "/tmp/bla".IO.open(:w); $a.close; $a.say: "bla"
```
```output
«Cannot do 'say' on a handle in binary mode␤  in block <unit> at <tmp> line 1␤␤»
```
## Expected Behavior
```output
«Cannot do 'say' on a closed handle␤  in block <unit> at <tmp> line 1␤␤»
```

## Environment
camelia bot on freenode, rakudo-moar ecf2b1e4f

lizmat:
Test added with https://github.com/Raku/roast/commit/0e25ee24b8

PR #3074, opened 2019-07-21
Try fixing closed handle operations error

voldenet:
This tries solving the problem described in bug #3073.

It has not been thoroughly tested, there are a few problems with this:
- new method naming
- race condition when closing/destroying handle: `$!decoder` is `Encoding::Decoder`, `$!PIO` is not null
- wrong exception because of race condition: `$!decoder` is not null (set by seek), `$!PIO` is null
- missing "closed handle" errors for a few methods

### DIFF
--- a/src/core.c/Exception.pm6
+++ b/src/core.c/Exception.pm6
@@ -745,6 +745,11 @@ my class X::IO::BinaryMode does X::IO {
     method message { "Cannot do '$.trying' on a handle in binary mode" }
 }
 
+my class X::IO::ClosedHandle does X::IO {
+    has $.trying;
+    method message { "Cannot do '$.trying' on a closed handle" }
+}
+
 my role X::Comp is Exception {
     has $.filename;
     has $.pos;
--- a/src/core.c/IO/Handle.pm6
+++ b/src/core.c/IO/Handle.pm6
@@ -225,6 +225,12 @@ my class IO::Handle {
         );
     }
 
+    method !no-decoder-exception(:$trying) {
+        nqp::if(nqp::isnull($!PIO),
+          X::IO::ClosedHandle,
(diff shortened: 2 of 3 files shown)


voldenet:
There were changes in error message, I've changed relevant test for now, but in this case similar languages do this:
- python/ruby - high level object is closed: `I/O operation on closed file`, `'read': closed stream (IOError)`
- nodejs/perl/golang/C - low level object is invalid: `EBADF`, `Bad file descriptor`

voldenet:
As written before, there are still some lingering errors, but solving them would introduce additional overhead into basic IO methods, impacting overall performance. It might be better to turn that message into EBADF inside MoarVM.

- `encoding` should fail when there are bytes available in the decoder, because `nqp::seekfh` would be used
- seek: `seek requires an object with REPR MVMOSHandle (got VMNull with REPR Null)`
- tell: `tell requires an object with REPR MVMOSHandle (got VMNull with REPR Null)`
- slurp: `read bytes requires an object with REPR MVMOSHandle (got VMNull with REPR Null)`
- WRITE: `write bytes requires an object with REPR MVMOSHandle (got VMNull with REPR Null)`
- READ: `read bytes requires an object with REPR MVMOSHandle (got VMNull with REPR Null)`
- lock: `get native descriptor requires an object with REPR MVMOSHandle (got VMNull with REPR Null)`
- unlock: `get native descriptor requires an object with REPR MVMOSHandle (got VMNull with REPR Null)`

JJ:
Can you please try and fix the conflicts?

voldenet:
I've fixed the conflicts

lizmat:
Looks like this was effectively implemented with 3865f1caee so closing this now

PR #3076, opened 2019-07-22
Better support for EXPORT with Stash

vrurg:
Turn off containerization for Stash values

Also added operator (<EXPR>):: which is a shortcut for
Stash.new(<EXPR>)

Now during import values from Stash are used as-is without deconting.

perl6/problem-solving#74


lizmat:
Perhaps https://github.com/rakudo/rakudo/pull/3081 is a better solution?

vrurg:
Closed due to #3081 merge and acceptance.

PR #3079, opened 2019-07-23
Add gb2312 to the encoding lists

ZhongnianTao:


src/core/Rakudo/Internals.pm6
```diff
@@ -321,6 +321,8 @@ my class Rakudo::Internals {
       # ShiftJIS
       'windows-932',     'windows-932',
       'windows932',      'windows-932',
+      # GB2312
+	  'gb2312',          'gb2312',
```
AlexDaniel: This line is misindented (tabs vs spaces?).

src/core/Encoding/Registry.pm6
```diff
@@ -21,7 +21,8 @@ my class Encoding::Registry {
             'latin1','latin-1','csisolatin1','l1','ibm819','cp819'),
           nqp::list('windows-1251', 'windows1251'),
           nqp::list('windows-1252', 'windows1252'),
-          nqp::list('windows-932',  'windows932')
+          nqp::list('windows-932',  'windows932'),
+          nqp::list('gb2312',  'gb2312')
```
AlexDaniel: Are trailing commas supported here? If yes, then I'd recommend to write it as `nqp::list('gb2312',  'gb2312'),` so that the next PR for another encoding doesn't have to touch that line.

ZhongnianTao: Yes, it's changed in the new commit.

src/core/Rakudo/Internals.pm6
samcv: Agreed. Just fix the alignment of this one.

PR #3081, opened 2019-07-24
Keep containers in Maps

lizmat:
In reference to https://github.com/perl6/problem-solving/issues/74

### DIFF
--- a/src/core/Map.pm6
+++ b/src/core/Map.pm6
@@ -4,45 +4,6 @@ my class Map does Iterable does Associative { # declared in BOOTSTRAP
     # my class Map is Iterable is Cool
     #   has Mu $!storage;
 
-    multi method WHICH(Map:D: --> ValueObjAt:D) {
-        nqp::box_s(
-          nqp::concat(
-            nqp::if(
-              nqp::eqaddr(self.WHAT,Map),
-              'Map|',
-              nqp::concat(self.^name,'|')
-            ),
-            nqp::sha1(
-              nqp::join(
-                '|',
-                nqp::stmts(  # cannot use native str arrays early in setting
-                  (my \keys := nqp::list_s),
-                  (my \iter := nqp::iterator($!storage)),
-                  nqp::while(
-                    iter,
-                    nqp::push_s(keys,nqp::iterkey_s(nqp::shift(iter)))
-                  ),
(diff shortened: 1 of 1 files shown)


lizmat:
Please note that **IF** this PR gets merged, I will bring out a `TupleMap` functionality to the ecosystem, so that people wanting to use the old `Map` functionality can do so with only minor changes.

vrurg:
Spectest `S32-hash/map.t` fails at 

```
    my class C {
        has Int @.a;
        has Int @.b;
    }
    lives-ok { C.new(|(a => (1, 2, 3), b => (4, 5, 6)).Map) },
        'Map does not introduce bogus Scalar containers';

```

While it just checks for survival, here is what I get as an object of `C`:

```
C.new(a => [(1, 2, 3),], b => [(4, 5, 6),])
```

What actually happens is:

```
(a => (1, 2, 3), b => (4, 5, 6)).Map<a>.VAR.^name.say; # Scalar
```

There is an extra containerization takes place somewhere in the middle because in any of the following cases:

```
my $m := (1,2,3); say $m.VAR.^name; # List
my $m := a=>(1,2,3); say $m.VAR.^name, " ", $m.value.VAR.^name; # Pair List
my $m := (a=>(1,2,3), b=>(1,2,3)); say $m[0].VAR.^name, " ", $m[0].value.VAR.^name; # Pair List
```

everything is correct.

vrurg:
Ok, tracked this one down. The fix is to replace metho `Map` on `Any` with:

```
    multi method Map() { Map.new(self) }
```

where it previously was:

```
    multi method Map() { self.hash.Map }
```

With this fix all spectests are happy,

lizmat:
What is the diff on that ??

vrurg:
```
diff --git a/src/core/Any.pm6 b/src/core/Any.pm6
index 51a5e96ed..855282bdd 100644
--- a/src/core/Any.pm6
+++ b/src/core/Any.pm6
@@ -84,7 +84,7 @@ my class Any { # declared in BOOTSTRAP
     multi method Hash() { self.hash.Hash }

     proto method Map(|) is nodal {*}
-    multi method Map() { self.hash.Map }
+    multi method Map() { Map.new(self) }

     proto method elems(|) is nodal {*}
     multi method elems(Any:U: --> 1) { }
```

PR #3085, opened 2019-07-28
[WIP] Formalize what the IO role actually does and introduce IO::NativeDescriptor

Kaiepi:
Originally, this was only supposed to add support for using file descriptors/`HANDLE`s with `IO::Handle`. This alone wasn't quite good enough since `IO::Socket::INET` and `IO::Socket::Async` will also need to be able to support creating sockets given a file descriptor or path once I implement sending file descriptors over sockets as part of my networking grant work, and I want to be able to just pass an `IO` or `IO::Handle` instance to `IO::Socket::INET.new`/`IO::Socket::Async.new` and have it work without hardcoding behaviour specific to file descriptors or paths in v6.e. This requires three different changes in particular.

This starts out by formalizing what `IO` actually is: a role done by classes describing files in some way. Methods that were originally specific to `IO::Path` and `IO::Special` but are generic file operations are now stubbed in `IO`.

Second, this introduces `IO::NativeDescriptor`, an abstraction over file descriptors (`HANDLE`s on Windows). It can be used in the same way `IO::Path` and `IO::Special` are.

Finally, this makes whatever uses `IO` as a type more generic in order to make it possible to create custom file classes. `IO::Handle` no longer hardcodes logic for how `IO`-typed variables should be handled depending on their type in a way that makes that impossible; that is now implemented in the types themselves.

All of this is done while aiming to preserve compatibility with code that already uses `IO` as a type. I'll need to grep the ecosystem for cases where people specify `IO` as a type when they really mean `IO::Path` before this can be merged, as that's the only code that I know of that this may break.

This is unfinished and hasn't been fully tested past the first commit. There are still several nqp ops for dealing with native descriptors that haven't been implemented yet, as well as changes to RI necessary for `IO::NativeDescriptor`'s stat methods to work. Roast will need tests for `IO::NativeDescriptor` and the docs will need to be updated. This will need testing on Windows once it's finished.

### DIFF
--- a/src/core/IO/Handle.pm6
+++ b/src/core/IO/Handle.pm6
@@ -124,50 +124,15 @@ my class IO::Handle {
 
         fail X::IO::Directory.new(:$.path, :trying<open>) if $!file.d;
 
-        my Bool $bom;
-        if nqp::istype($!file, IO::Path) {
+        {
+            $!PIO := $!file.open-native: :$mode, :$create, :$append, :$truncate, :$exclusive;
             CATCH { .fail }
-            $!PIO := nqp::open(
-              $!file.absolute,
-              nqp::concat(
-                nqp::if(nqp::iseq_s($mode, 'ro'), 'r',
-                nqp::if(nqp::iseq_s($mode, 'wo'), '-',
-                nqp::if(nqp::iseq_s($mode, 'rw'), '+',
-                  die "Unknown mode '$mode'"))),
-                nqp::concat(nqp::if($create,      'c', ''),
-                nqp::concat(nqp::if($append,      'a', ''),
-                nqp::concat(nqp::if($truncate,    't', ''),
-                            nqp::if($exclusive,   'x', ''))))));
-            $bom = True;
+        }
+
+        my Bool $file-not-special = $!file.f || $!file.d || $!file.l;
 #?if moar
-            self!remember-to-close;
+        self!remember-to-close if $file-not-special;
 #?endif
-        }
-        elsif nqp::istype($!file, IO::Special) {
-            my $what := $!file.what;
-            if $what eq '<STDIN>' {
-                $!PIO := nqp::getstdin();
-            }
-            elsif $what eq '<STDOUT>' {
-                $!PIO := nqp::getstdout();
-            }
-            elsif $what eq '<STDERR>' {
-                $!PIO := nqp::getstderr();
-            }
-            else {
-                die "Don't know how to open '$what' especially";
-            }
-            $bom = False;
-        }
-        elsif nqp::istype($!file, IO::NativeDescriptor) {
-            $!PIO := nqp::fdopen(nqp::unbox_i($!file));
-            $bom   = $!file.f;
-        }
-        else {
-            # XXX TODO: abstract this entire if out of IO::Handle.open and
-            # implement it in IO so custom IO classes can be used.
-            die "The type of file given to '.open' is unsupported: '$!file'."
-        }
 
         $!chomp = $chomp;
         $!nl-out = $nl-out;
@@ -178,8 +143,9 @@ my class IO::Handle {
             $!encoder := $encoding.encoder(:translate-nl);
             $!encoding = $encoding.name;
 
-            # Add a byte order mark to the start of the file for utf16.
-            if $bom && nqp::iseq_s($!encoding, 'utf16') {
+            # Add a byte order mark to the start of the file for utf16 if the
+            # file given isn't special.
+            if $file-not-special && nqp::iseq_s($!encoding, 'utf16') {
                 if $create && !$exclusive && (!$append || $append && $!file.s == 0) {
                     self.WRITE: Blob[uint16].new: 0xFEFF;
                 }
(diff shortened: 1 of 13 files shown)


src/core/IO/Handle.pm6
```diff
@@ -1,8 +1,7 @@
-my class IO::Path { ... }
 my class Proc { ... }
 
 my class IO::Handle {
-    has $.path;
```
ugexe: I disagree with this rename. I use `IO::Handle` based abstractions for http, pipes, etc which all have a path/uri representation but are not files. It also seems to add some needless complexity to the constructor which now can seemingly take both `path` and `file`.

src/core/IO/NativeDescriptor.pm6
```diff
@@ -0,0 +1,147 @@
...
+    multi method perl(IO::NativeDescriptor:D: --> Str:D) {
+        nqp::p6box_s(
+          nqp::concat(
+            nqp::concat(self.^name, '.new('),
+            nqp::concat(nqp::coerce_is(nqp::unbox_i(self)), ')')
+          )
+        )
+    }
+
+    multi method e       (IO::NativeDescriptor:D: --> Bool)    {
```
ugexe: This indentation style isn't used anywhere else. Adding any new method that has a longer name will require breaking the `blame` for all methods when they need to be reindented.

src/core/IO/Socket/Async.pm6
```diff
@@ -208,8 +208,8 @@ my class IO::Socket::Async {
             self.bless(:&on-close, :$VMIO-tobe, :$socket-host, :$socket-port);
         }
 
-        method native-descriptor(--> Int) {
-            nqp::filenofh(await $!VMIO-tobe)
+        method native-descriptor(--> IO::NativeDescriptor) {
```
ugexe: This seems like a fairly big breaking change

src/core/IO/Special.pm6
```diff
@@ -1,45 +1,70 @@
...
+          )
+        )
+    }
+
+    multi method e       (IO::Special:D: --> True)    {         }
+    multi method d       (IO::Special:D: --> False)   {         }
+    multi method f       (IO::Special:D: --> False)   {         }
+    multi method s       (IO::Special:D: --> 0)       {         }
+    multi method l       (IO::Special:D: --> False)   {         }
+    multi method r       (IO::Special:D: --> Bool)    {
```
ugexe: This is supposed to return `Bool:D`, not `Bool`

src/core/IO/Special.pm6
```diff
@@ -1,45 +1,70 @@
...
-    method accessed(IO::Special:D: --> Instant) { Instant }
-    method changed( IO::Special:D: --> Instant) { Instant }
-    method mode(IO::Special:D: --> Nil) { }
+    multi method x       (IO::Special:D: --> False)   {         }
+    multi method rw      (IO::Special:D: --> False)   {         }
+    multi method rwx     (IO::Special:D: --> False)   {         }
+    multi method modified(IO::Special:D: --> Instant) { Instant }
+    multi method accessed(IO::Special:D: --> Instant) { Instant }
+    multi method changed (IO::Special:D: --> Instant) { Instant }
+    multi method mode    (IO::Special:D: --> IntStr)  { IntStr  }
```
ugexe: This returned `Nil` before. Not its returning `IntStr`?

src/core/Int.pm6
```diff
@@ -46,6 +47,10 @@ my class Int does Real { # declared in BOOTSTRAP
 
     method Capture() { die X::Cannot::Capture.new: :what(self) }
 
+    method IO(Int:D: --> IO::NativeDescriptor:D) {
```
ugexe: This feels pretty wrong

src/core/io_operators.pm6
```diff
@@ -81,7 +81,7 @@ multi sub dir(IO::Path:D $path, |c) { $path.dir(|c) }
 multi sub dir(IO()       $path, |c) { $path.dir(|c) }
 
 proto sub open($, |) {*}
-multi sub open(IO() $path, |c) { IO::Handle.new(:$path).open(|c) }
+multi sub open(IO() $file, |c) { IO::Handle.new(:$file).open(|c) }
```
ugexe: Again, I disagree with the renaming to `$file` from `$path` here.

ugexe:
There are a *lot* of breaking changes in here.

Kaiepi:
Most of them are mistakes from while I was refactoring. `Int.IO` exists was because I was forgetting `Cool` already had an `IO` method, and I wasn't considering stuff like URLs when I renamed `IO::Handle.path` to `IO::Handle.file`. I was hoping I could keep returning `IO::NativeDescriptor` from `.native-descriptor` since it's a subclass of `Int`, but after thinking for a while you're right, that would break stuff that depends on it being an `Int` specifically.

src/core/IO/NativeDescriptor.pm6
AlexDaniel: I love vertical alignment but not as much :) Looks weird indeed.

lizmat:
Is this PR still worthwhile salvaging with the conflicts?  Or can it be closed?

Kaiepi:
It'll still be a while yet before I can come back to this.. this can be closed for now.

PR #3103, opened 2019-08-06
Return Failure from failed P6-level .parse on 6.e

vrurg:
Cherry-pick of @9501edae4f73a970e3270e3b0336a7b3045d3329:

The `.parse` and `.parsefile` methods will now return a `Failure` object that points
to the high-water state of the parser, if available.  (The `.subparse` method
does not, since it is presumed to be for lower-level piecemeal matching that
is likelier to be success-oriented and have its own failure recovery modes.)

The behavior isn't changing for `6.c` and `6.d`.

Resolving perl6/6.d-prep#9

### DIFF
--- a/src/core/Grammar.pm6
+++ b/src/core/Grammar.pm6
@@ -1,5 +1,62 @@
 my class Grammar is Match {
 
+    # (oughta be down in Match or even nqp really)
+    method locprepost() {
+        my $orig = self.orig;
+        my $marked = self.?MARKED('ws');
+        my $pos = $marked && index(" }])>»", substr($orig, self.pos, 1)) < 0 ?? $marked.from !! self.pos;
+
+        my $prestart = $pos - 40;
+        $prestart = 0 if $prestart < 0;
+        my $pre = substr($orig, $prestart, $pos - $prestart);
+        $pre = $pre.subst(/.*\n/, "");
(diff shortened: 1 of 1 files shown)


vrurg:
Note to myself: these two to be reverted:

@b53616f8e67f9b19366008b3abf55400a3d6cd2b
@1fb68c4b7a7c975f26fc81ad79f000958d1b4afd

AlexDaniel:
It should also resolve this ticket: https://rt.perl.org/Ticket/Display.html?id=131919

vrurg:
> I believe they definitely do need to be reverted in 6.c-errata and 6.d-errata branches, it's my mistake that I didn't do it back then. Not sure what exactly should be done with them for master and upcoming 6.e.

I'll have a better look at what's going on there. But typical approach for now is to have separate tests for releases. Most likely I'll split the corresponding test files in two: for pre-`6.e` and `6.e`.

vrurg:
I have certain doubts now about the correctness of this change. 

Change of `parse` return value is actually a change of class behavior. By taking into account opinion stated by @jnthn in perl6/problem-solving#3 and perl6/problem-solving#71 it makes sense for `6.e`-specific `Grammar` class which only depends on what code created it, not the code which is calling it.

This approach has its implications which I noted in perl6/problem-solving#71.

PR #3108, opened 2019-08-06
Pod: allow empty config value

### LINKED ISSUE #3098: Config attributes cannot be empty strings in pods
## The Problem

Empty configuration attributes cannot be defined:

~~~perl6
=begin pod :config("")
=end pod
~~~

## Expected Behavior

That code should compile.

## Actual Behavior

This errors is thrown:

~~~
===SORRY!===
No such method 'Str' for invocant of type 'NQPMu'
~~~

## Steps to Reproduce

Execute:

~~~
perl6 -e "=begin pod :kind('') \n =end pod" 
~~~

## Environment
* Operating system: Ubuntu 16.04

~~~
This is Rakudo version 2019.03.1 built on MoarVM version 2019.03
implementing Perl 6.d.
~~~

### LINKED ISSUE #3101: LTA error msg with empty pod config value
See issue #3098 for details.

taboege:
Previously a Pod block such as

``` perl6
=begin pod :config('')
=end pod
# or
=begin pod :config[]
=end pod
```

would throw a compiler error (coercing NQPMu to Str). This is because,
first, an empty quoted string was treated like nothing when compiling
the list of values of the `:config` key above, and second, an empty value
list was unexpected on the caller site.

This commit makes empty quoted strings recognized and empty value lists
supported, so that

``` perl6
=begin pod :x('') :y( )
=end pod
```

parses successfully and gives

``` perl6
dd $=pod.head.config
#= Hash %!config = {:x(""), :y($[])}
```

Fixes #3098 and closes #3101

### DIFF
--- a/src/Perl6/Pod.nqp
+++ b/src/Perl6/Pod.nqp
@@ -486,7 +486,7 @@ class Perl6::Pod {
                 $quoted   := nqp::null();
             }
 
-            if nqp::chars($quoted) {
+            if nqp::chars($quote) || nqp::chars($quoted) {
                 $word := nqp::concat($word, $quoted);
             }
             elsif nqp::chars($unquoted) {
@@ -498,7 +498,7 @@ class Perl6::Pod {
                 @pieces.push($delim) if ($keep eq 'delimiters');
                 $word := '';
             }
-            if !nqp::chars($line) && nqp::chars($word) {
+            if !nqp::chars($line) && (nqp::chars($quote) || nqp::chars($word)) {
                 @pieces.push($word);
                 $word := '';
             }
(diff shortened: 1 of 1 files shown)


taboege:
I'm not sure how to roast this because it involves a compiler error, and I have not seen any such Pod tests (`find -name *pod*` in roast). If this gets merged, maybe the maintainer can point me in the right direction.

taboege:
This is the response from D. Conway in question: https://gist.github.com/tbrowder/dd9bf5cdf217b1168205bbef6c4a51b0. I didn't find it on the documentation.

> A comprehensive solution with his suggestions would have to start in the grammar and maybe actions where the config chunks are parsed.

I don't see this. The grammar uses the usual Perl 6 `colonpair`. As per Conway's explanation this is fine -- there should be no difference between a Perl 6 colonpair literal and a Pod configuration pair literal.

https://github.com/rakudo/rakudo/blob/636888c69455c3a2e6cc8f6ea72ee372ebfdaa35/src/Perl6/Grammar.nqp#L4903-L4905

The grammar side is fine in my opinion. And the actions is just what this commit touches. Or are you saying that this PR should not be merged because the code it touches needs to be changed fundamentally anyway?

However, I'm not sure if this PR covers all the instances of empty values. It does fix the named issue(s) though. Possibly we should have tests in roast first of all featuring all the desired constructs and develop against those.

taboege:
> I appreciate your interest in pod6 and hope my comments aren't too negative
No worries, it should be about making the best rakudo possible.

PR #3112, opened 2019-08-07
Implement CORE::v6<lang-rev> pseudo-packages

vrurg:
Extend `CORE` with second-level packages representing individual cores
for specific language revisions. I.e, `CORE::v6c` is for `CORE.setting`
or, correspondingly, for `6.c` language. `CORE::v6d`, `CORE::v6e`, and so on.
New letters become available as soon as a `CORE` context with
`CORE-SETTING-REV` symbol becomes available.

Each subpackage has `PRECISE_SCOPE`. Thus, `CORE::v6d::Any` will fail
because `Any` is not defined in `CORE.d.setting`.

perl6/problem-solving#80


jnthn:
> Each subpackage has PRECISE_SCOPE. Thus, CORE::D::Any will fail because Any is not defined in CORE.d.setting.

What's the argument for having it this way? I can more easily see the argument for having it do the lookup in outer COREs too, so it'd mean "give me the symbol as it was in 6.d". At first I thought there was an argument that the current way is useful if asking if that symbol was available in 6.d, but even then I think I'd want the "flat" view. Perhaps being able to introspect what's new benefits from this?

Also, what are the semantics of saying `CORE::E` if I never actually used that version so far - or perhaps if I'm in code declaring `use v6.d`? Is it restricted to just those versions below the current language version of the current compilation unit?

I also worry a bit about what happens when we get to version X, and then `CORE::X` is nicely ambiguous. :) I wonder if `CORE:ver<D>` or some such would be better.

jnthn:
Ah, and what I should have said in the first comment: I'm good with the functionality, just questioning a couple of the details.

AlexDaniel:
> I also worry a bit about what happens when we get to version X, and then `CORE::X` is nicely ambiguous. :) I wonder if `CORE:ver<D>` or some such would be better.

I dislike that the first half of the version disappears. Is there any way to preserve the full string? Like `CORE::v6d` ?

vrurg:
> What's the argument for having it this way? I can more easily see the argument for having it do the lookup in outer COREs too, so it'd mean "give me the symbol as it was in 6.d". At first I thought there was an argument that the current way is useful if asking if that symbol was available in 6.d, but even then I think I'd want the "flat" view. Perhaps being able to introspect what's new benefits from this?

Sorry, not sure what you mean by "flat" view.

My logic was: there is `CORE` which represents the final core, i.e. list of symbols available for the current language revision. It's a result of flattening down cores starting with `c`. How do we access symbols of a specific core? `CORE::<rev>` was floating atop. And it really gives nothing else but simple direct access for introspection and direct access to symbols from a specific core.

The direct outcome I have from it is:

```
multi foo(CORE::C::PseudoStash $p) { ... }
multi foo(CORE::E::PseudoStash $p) { ... }
```

> Also, what are the semantics of saying `CORE::E` if I never actually used that version so far - or perhaps if I'm in code declaring `use v6.d`? Is it restricted to just those versions below the current language version of the current compilation unit?

There is the restriction. Second level exists only if corresponding core is loaded. `CORE::E` is a `Failure` under the scope of `use v6.d`.

> I also worry a bit about what happens when we get to version X, and then `CORE::X` is nicely ambiguous. :) I wonder if `CORE:ver<D>` or some such would be better.

`CORE::<rev>` has highest precedence, as any other pseudo-package. I.e. as `MY::OUTER` would override any possibly defined `OUTER` symbol in the current scope, `CORE::X` would override `X`. While checking for possible clashes, I overlooked one of the obvious candidates.

A solution would be to switch to lowercase, `CORE::x`. It's not that neat, but conforms to the current revision letters policy.

PR #3122, opened 2019-08-14
Introduce the $*PRECISION variable

lizmat:
This dynamic variable indicates what should happen if a `Rat` is running
out of range for its denominator.

- when not explicitely set, downgrade to Num
- when explicitely set to the `Num` type object, downgrade to Num
- when explicitely set to the `FatRat` type object, upgrade to FatRat
- when explicitely set to the `Exception` type object, die saying so
- when explicitely set to something else, die with a huh?

This also applies to literals if this dynamic variable has been set at
compile time.  Some examples:

    my $a = 1 / 2**63; dd $a.^name; $a /= 2; dd $a.^name  # Rat, Num

    my $*PRECISION = FatRat;
    my $a = 1 / 2**63; dd $a.^name; $a /= 2; dd $a.^name  # Rat, FatRat

    my $*PRECISION = Exception;
    my $a = 1 / 2**63; dd $a.^name; $a /= 2; dd $a.^name  # Rat, dies

    my $a = 1 / 2**64; dd $a.^name   # Num

    BEGIN $*PRECISION = FatRat;
    my $a = 1 / 2**64; dd $a.^name   # FatRat

    BEGIN $*PRECISION = Exception;
    my $a = 1 / 2**64; dd $a.^name   # dies

Inspired by discussions with Martin Becker at PerlCon 2019 in Riga.

### DIFF
--- a/src/core/Rat.pm6
+++ b/src/core/Rat.pm6
@@ -70,20 +70,35 @@ sub DIVIDE_NUMBERS(Int:D $nu, Int:D $de, \t1, \t2) is raw {
     )
 }
 
+# Initialize the $*PRECISION dynamic var so that it can be used
+{
+    PROCESS::<$PRECISION> = Num;
+}
+
 # ALL RATIONALS MUST BE NORMALIZED, however in some operations we cannot
 # ever get a non-normalized Rational, if we start with a normalized Rational.
 # For such cases, we can use this routine, to bypass normalization step,
 # which would be useless.  Also used when normalization *was* needed.
 proto sub CREATE_RATIONAL_FROM_INTS(|) {*}
 multi sub CREATE_RATIONAL_FROM_INTS(Int:D \nu, Int:D \de, \t1, \t2) is raw {
-    nqp::if(
-      nqp::islt_I(de,UINT64_UPPER),         # do we need to downgrade to float?
-      nqp::p6bindattrinvres(                # no, we need to keep a Rat
-        nqp::p6bindattrinvres(nqp::create(Rat),Rat,'$!numerator',nu),
-        Rat,'$!denominator',de
-      ),
-      nqp::p6box_n(nqp::div_In(nu,de))      # downgrade to float
-    )
+    nqp::islt_I(de,UINT64_UPPER)                          # downgrade to float?
+      ?? nqp::p6bindattrinvres(                            # no, keep as Rat
+           nqp::p6bindattrinvres(nqp::create(Rat),Rat,'$!numerator',nu),
+           Rat,'$!denominator',de
+         )
+      !! nqp::eqaddr(                                      # downgrade to Num?
+           (my $precision := nqp::decont($*PRECISION)),Num
+         )
+        ?? nqp::p6box_n(nqp::div_In(nu,de))                 # yes, downgrade
+        !! nqp::eqaddr($precision,FatRat)                   # keep precision?
+          ?? CREATE_RATIONAL_FROM_INTS(nu,de,FatRat,FatRat)  # yes, make FatRat
+          !! nqp::eqaddr($precision,Exception)               # no downgrade?
+            ?? X::AdHoc.new(                                  # not allowed to
+                 payload => 'Not allowed to downgrade to Num'
+               ).throw
+            !! X::AdHoc.new(                                  # huh?
+                 payload => "Unexpected value $*PRECISION.perl() for \$*PRECISION"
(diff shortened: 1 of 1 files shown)


src/core/Rat.pm6
```diff
@@ -70,20 +70,35 @@ sub DIVIDE_NUMBERS(Int:D $nu, Int:D $de, \t1, \t2) is raw {
...
+           (my $precision := nqp::decont($*PRECISION)),Num
+         )
+        ?? nqp::p6box_n(nqp::div_In(nu,de))                 # yes, downgrade
+        !! nqp::eqaddr($precision,FatRat)                   # keep precision?
+          ?? CREATE_RATIONAL_FROM_INTS(nu,de,FatRat,FatRat)  # yes, make FatRat
+          !! nqp::eqaddr($precision,Exception)               # no downgrade?
+            ?? X::AdHoc.new(                                  # not allowed to
+                 payload => 'Not allowed to downgrade to Num'
+               ).throw
+            !! X::AdHoc.new(                                  # huh?
```
vrurg: As @AlexDaniel mentioned on IRC, I hope we'll take care of those `AdHoc`s. Meanwhile, I have a suggestion of typechecking against `Exception` and throwing the type in `$*PRECISION`. I.e. allow one to:

```
class X::MyOverflow does X::Overflow {
 ...
}

$*PRECISION = X::MyOverflow
```

`Rat` could then pass `nu` and `de` to the exception consturctor (pseudo code follows):

```
$*PRECISION.new(numerator => nu, denominator => de).throw
```

lizmat: Hmmm.... perhaps if it is not `Num` of `FatRat`, assume it is a class to be called with `:$nu, :$de` and just instantiate that.  And add a `X::Overflow` class, which one could specify to get the behaviour previously described for `Exception` ?

jnthn:
While I agree with the principle of providing some user control over this, I don't feel that a dynamic variable is the way to do it, since:

* From a language design point of view, it's affecting the behavior of an operator, which is part of the current language, and so it should be a lexical property.
* From a performance point of view, math code tends to be hot-path-y. Dynamic variables don't go so well with that. Furthermore, making the decision about what to do at runtime is also going to cost something.

A pragma (really, module doing the right thing) like `use rat-overflow :fat`, `use rat-overflow :die`, and the (default) `use rat-overflow :num` would seem preferable both from a language design point of view, but also from a performance one, since these would just need to export operators that behave in the appropriate way.

vrurg:
I accidentally started a review. Just ignore it.

lizmat:
> From a language design point of view, it's affecting the behavior of an operator, which is part of the current language

I don't think it's affecting any specific operator, it's affecting behaviour if any operation on a `Rat` results in an overflow.  This applies to many `Rat` operations, or am I missing something?

> , and so it should be a lexical property.

I disagree.  I can see value in a library that does all sorts of operations on `Rat`s and a script that can decide whether to upgrade to `FatRat`s or to downgrade to `Num`s when using that library.  I don't see how you can do this with a lexical property.

> From a performance point of view, math code tends to be hot-path-y. Dynamic variables don't go so well with that. Furthermore, making the decision about what to do at runtime is also going to cost something.

The dynamic variable is **only** tested when an overflow is about to occur.  So we're in a special case then already, aka a path that probably won't be speshed or jitted anyway.

src/core/Rat.pm6
vrurg: _I have cancelled the review and re-posting this comment_

Hope I got you correctly. My suggestion was for the "no downgrade?" branch only where you anyway throw. Why thrown implicit `AdHoc` if we can allow a user to define what exception he wants to be thrown?

Besides, my example was incomplete. It should either be:

```
role X::Overflow is Exception {
    has $!numerator;
    has $!denominator;
}
```

or 

```
class X::MyOverflow is Exception does X::Overflow {
...
}
```

Something like this.

vrurg:
> I disagree. I can see value in a library that does all sorts of operations on `Rat`s and a script that can decide whether to upgrade to `FatRat`s or to downgrade to `Num`s when using that library. I don't see how you can do this with a lexical property.

Perhaps it's an overcomplication, but it would be a class variable and a lexically scoped pragma which would set it do a desired value?

AlexDaniel:
This is nice but we need a decision first. Further discussion in https://github.com/Raku/problem-solving/issues/185, please.

PR #3123, opened 2019-08-16
implement Nativecall::DLL

cygx:
see https://github.com/MoarVM/MoarVM/pull/1158 and https://github.com/perl6/nqp/pull/573


JJ:
There's a conflict now. Also, I understand the other two PRs need to be merged, right?

lizmat:
Is this still a thing after new-disp has landed and @niner has done extensive work on the NativeCall interface?

Issue #3089, opened 2019-07-30
bad link in README: to features page on perl6.org


GregDavidson:
## The Problem
the features page on perl6.org, linked to by the README does not link to a valid page
# Expected Behavior
A working link
## Actual Behavior
A broken link
## Environment
* Operating system: 
* Compiler version (`perl6 -v`):

tqdv:
This was fixed by [commit 302bc8e](https://github.com/rakudo/rakudo/commit/302bc8ed8b22adce05f1b84277ca471e0300a3a4) in August 2019. We can probably close this issue now.

Altai-man:
Indeed. Thank you very much for noticing! Closing now.

PR #3127, opened 2019-08-18
Fix misleading paragraph in readme.

ribbon-otter:
The old paragraph contained a dead link as was noted [Github issue
#3089](rakudo#3089); Moreover the paragraph reflected when there
were multiple partial implementations of Perl 6. As Rokudo is now the
most complete implementation, it is better to reflect that. Plus, the
page comparing the compilers seems to be gone from the website, or at
least I couldn't find it. Thus, I substituted for a link to the FAQ
explaining the difference Perl 6 vs Roku vs Rokudo.


coke:
It's "Rakudo" not "Rokudo"

ribbon-otter:
spelling mistake corrected

README.md
```diff
@@ -12,9 +12,9 @@ that would normally come with a full Perl 6 distribution.  If you're
 after more than just the bare compiler, please download [the latest
 Rakudo Star package](http://rakudo.org/downloads/star).
 
-Note that different backends implement slightly different sets of
```
vrurg: Please return the part about backends implementing slightly different sets of features. This is actually true statement and a user must not expect all them to be running any kind of code equally well. Since there is no documentation page on this a plain text with no link would be ok.

Thank you!

ribbon-otter: I have reinserted that note. I also found the [compilers page](https://perl6.org/compilers/) and now link to it instead of the FAQ.

PR #3138, opened 2019-08-24
Improve multi-revision support

vrurg:
This is implementation for perl6/problem-solving#94. At the current
stage Rakudo successfully compiles and passes most tests. There is a bug
related to setting lower `CORE` version as outer for the currently
compiled `CORE`.

- `core/` is now `core.c/`
- each core now has own BOOTSTRAP and EXPORTHOW.
- bootstrap is now compiled into Perl6/BOOTSTRAP/v6<rev>.moarvm
- whereas words `CORE` and `NULL` were used for `6.c`, they now have
  `.c` suffix as other revisions do.


vrurg:
At the current stage Rakudo successfully compiles and passes most tests. Those failing are either unrelated, or it's `t/08-performance/04-settingkeys.t` which is yet to be considered. It currently fails with: 

```
# Missing symbols: $=finish $?PACKAGE ::?PACKAGE EXPORT GLOBALish
```

but the real need for this symbols in a `CORE` is yet to be considered. In either case, it wouldn't be a problem to get them back any time.

blib/Perl6/BOOTSTRAP/.gitignore
```diff
@@ -0,0 +1,5 @@
+# Don't remove this file and dir.
+# They are needed to host the bytecode files during the build process
+for any VM.
```
jnthn: Missing `#` at the start of this line?

blib/Perl6/BOOTSTRAP/.gitignore
```diff
@@ -0,0 +1,5 @@
...
+*.pbc
```
jnthn: This was "parrot byte code", which is no more. But maybe we want to ignore `*.maorvm`?

src/Perl6/Actions.nqp
```diff
@@ -1886,7 +1886,7 @@ class Perl6::Actions is HLL::Actions does STDActions {
         # isn't already declared.
         my $BLOCK := $*W.cur_lexpad();
         my $type := $BLOCK.ann('IN_DECL');
-        if $type eq 'mainline' && %*COMPILING<%?OPTIONS><setting> eq 'NULL' {
+        if $type eq 'mainline' && %*COMPILING<%?OPTIONS><setting> eq 'NULL.c' {
```
jnthn: I don't see why `NULL.c`. `NULL` is simply the absence of a setting; I don't see how nothing can be versioned.

blib/Perl6/BOOTSTRAP/.gitignore
vrurg: Actually, same we have in `blib/Perl6`. I copied it over without actually checking its content.

blib/Perl6/BOOTSTRAP/.gitignore
vrurg: Makes sense, will fix both.

src/Perl6/Actions.nqp
vrurg: Currently `NULL.d` and `NULL.e` are used for building corresponding cores. `ModuleLoader` uses this revisions to find and pre-load previous `CORE`s. `NULL.c` is to unify building of all three cores in the build subsystem. Otherwise it requires a separate rule in `Makefile.in` template.

jnthn: > Currently NULL.d and NULL.e are used for building corresponding cores.

Hmm, that's odd, in that I thought CORE.d would just use CORE.c as its setting, and CORE.e would use CORE.d as its setting.

PR #3142, opened 2019-08-24
add better err msgs for invalid open brackets for inline and leading …

tbrowder:
…declarator blocks


tbrowder:
I am filing an issue to demonstrate the current errors this fixes (I thought I had but I can't find it).

tbrowder:
The missing used module, after being added back in, causes some test failures--not sure why yet.

PR #3154, opened 2019-08-30
Fix for failing declaration of grammar named Grammar

vrurg:
- Re-implemented `v6.e` `Grammar` use by `grammar` using `EXPORTHOW`.
- `p6getlexclient` got second optional parameter: `$setting-only` and was
fixed for correct handling of situations where our client is in `SETTING`
or `CORE` scope.
- Fixed core classes not having their language version set.

### DIFF
--- a/src/Perl6/Metamodel/DefaultParent.nqp
+++ b/src/Perl6/Metamodel/DefaultParent.nqp
@@ -1,10 +1,8 @@
 role Perl6::Metamodel::DefaultParent {
     my @default_parent_type;
-    my $setting_only := 0;
 
-    method set_default_parent_type($type, :$setting-only = 0) {
-        @default_parent_type[0] := nqp::isconcrete($type) ?? $type !! $type.HOW.name($type);
-        $setting_only := $setting-only;
+    method set_default_parent_type($type) {
+        @default_parent_type[0] := $type;
     }
 
     method has_default_parent_type() {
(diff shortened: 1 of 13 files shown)


src/Perl6/Metamodel/DefaultParent.nqp
```diff
@@ -1,15 +1,17 @@
 role Perl6::Metamodel::DefaultParent {
     my @default_parent_type;
+    my $setting_only := 0;
 
-    method set_default_parent_type($type) {
+    method set_default_parent_type($type, :$setting-only = 0) {
         @default_parent_type[0] := nqp::isconcrete($type) ?? $type !! $type.HOW.name($type);
+        $setting_only := $setting-only;
```
jnthn: I really dislike this. Types are not represented by their string names, they're represented by type objects. The metamodel of all places should be really clear about this. And all of this is a workaround for not doing things the way I suggested in the first place.

vrurg: I ask to have it for a while as a workaround. Making versionized metamodel would require a bit of time to design it and adapt the build subsystem. I would make a problem-solving for meta and going to finalize rebuild of the makefile templates in meanwhile. Replacing this with a new HOW would then be a matter of 10 minutes.

vrurg: In addition, the point of me asking for this: this could serve as a base for testing the new metamodel structure.

vrurg:
perl6/roast#573

Issue #3093, opened 2019-08-02
Installer is broken with --libdir=/usr/lib


atweiden:
rakudo-2019.07.1 with `Configure.pl --libdir=/usr/lib` does not work post-installation:

```
Unhandled exception: While looking for '/usr/share/perl6/runtime/perl6.moarvm': no such file or directory
```

The 2019.03 release did not have this problem.

It can be fixed with `ln -rs /usr/lib/perl6 /usr/share/perl6`.

patzim:
@atweiden The is a plan to remove the `--libdir` option, because it's conceptually conflicting with a recent change. In what context do you need to put the core libraries in a different folder? Would a solution that allows to install the entire `share/perl6` and the entire `share/nqp` folder into a different place also solve your need?

atweiden:
> In what context do you need to put the core libraries in a different folder?

Placing ELF binaries in `/usr/share` [violates](https://github.com/void-linux/void-packages/blob/master/common/hooks/post-install/11-pkglint-elf-in-usrshare.sh) Void Linux packaging standards.

> Would a solution that allows to install the entire share/perl6 and the entire share/nqp folder into a different place also solve your need?

So long as it prevents ELF binaries from being installed under `/usr/share`, yes.

AlexDaniel:
@patzim any updates on this? :)

patzim:
I have a branch [home-config-option](https://github.com/patzim/rakudo/tree/home-config-option) (and related nqp and nqp-configure branches) lying around that fix this. They was waiting for vrurgs build refactorings so the merge wouldn't interfere. And now I have two more weeks of No Time At All™ ahead of me. I might find some time in between, but no guarantees.

vrurg:
@patzim BTW, I'm done with rakudo. Hope that NQP rework doesn't affect you.

patzim:
@atweiden I have now implemented `--perl6-home` and `--nqp-home` `Configure.pl` options that allow putting the respective folders in a different place. I also removed the now broken `--libdir` option. I'd be grateful for feedback whether these changes address your packaging needs.

PR #3155, opened 2019-08-30
use libdir instead of rebuilding path from prefix

the-eater:
Currently when building with a custom `--libdir` it will place all runtime files in there, but the rest of rakudo rebuilds the path from only the prefix so starts searching in `$PREFIX/share/perl6` which results in the following error if you run `perl6`

```
[docker:f33e5790e49e /]$ perl6
Unhandled exception: While looking for '/usr/share/perl6/runtime/perl6.moarvm': no such file or directory
```

With this fix it should respect libdir and run as expected.

Current patch originates from https://github.com/void-linux/void-packages/pull/14073

### DIFF
--- a/src/main.nqp
+++ b/src/main.nqp
@@ -33,7 +33,7 @@ my $install-dir := $execname eq ''
 
 my $perl6-home := $comp.config<static_perl6_home>
     // nqp::getenvhash()<PERL6_HOME>
-    // $install-dir ~ '/share/perl6';
+    // $comp.config<libdir> ~ '/perl6';
 if nqp::substr($perl6-home, nqp::chars($perl6-home) - 1) eq $sep {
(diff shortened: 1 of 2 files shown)


patzim:
The concept of `--libdir` was superseded by the new concept of a perl6-home / nqp-home folder. Those folders are not yet configurable via a `Configure.pl` parameter but I am working on changing that.
The plan is to add two `Configure.pl` parameters: `--perl6-home` and `--nqp-home` and remove the `--libdir` parameter. I should have that finished soon. Also see [rakudo#3093](https://github.com/rakudo/rakudo/issues/3093) for a related issue.
@the-eater Would this solution also solve your problem?

the-eater:
Oh I didn't see that issue! yes that would fix it :) I'll close this MR in favor of that issue (we're talking about the exact same issue for the exact same reason :p)

PR #3164, opened 2019-09-02
Use ProtocolFamily to represent socket protocol families

Kaiepi:
This also removes PF_MAX, which ended up making it harder to verify if a
socket family is actually a supported socket family rather than easier,
and adds PF_UNSPEC, which was missing.

Pulling this since this could be a breaking change if I screwed anything up.

Passes `make test`, `make spectest`.


Kaiepi:
It'd make more sense to include this with a set of changes I'll be making soon rather than making it its own thing.

PR #3166, opened 2019-09-03
[WIP] [HEADSUP] Extract common parts of backend makefiles into a common template

vrurg:
Unify the build across different backends as much as possible.


patzim:
@vrurg: I can't make any promises until 1st of October. I have a *very* tight schedule this month...

Issue #3168, opened 2019-09-05
Unpacking a |%h.Map has a different behaviour from 2019.07.1


destroycomputers:
## The Problem
This bit of code executes differently in 2019.07.1 and HEAD. 
```
class A { has @.xs }
my %args = xs => [1, 2, 3];
A.new(|%args.Map).say
```
Relevant discussion: https://colabti.org/irclogger/irclogger_log/perl6?date=2019-09-05#l54
Commit that introduced the change in behaviour (taken from the irc log): https://github.com/rakudo/rakudo/commit/2ee82afe6e6713c26b929e2855ea05d8eb3b9fee
## Expected Behavior
The output to be `A.new(xs => [1, 2, 3])`
## Actual Behavior
Output is `A.new(xs => [[1, 2, 3],])`
## Steps to Reproduce
Run the program above.

## Environment
* Operating system: Whatever evalable6 runs on
* Compiler version (`perl6 -v`): version used by evalable6, apparently HEAD

AlexDaniel:
Output on all releases: https://gist.github.com/f530b87b947ae95217adb0c9db6a36ac

vrurg:
This was intentional change fixing a bug in previous releases. The fix is about synchronizing `List` and `Map` behaviors. Here is an example:

```
my $foo = 42;
my @a is List = 1, $foo;
say @a.map: { $_.VAR.^name };
my %m is Map = a => 1, b => $foo;
say %m<a b>.map: { $_.VAR.^name }
```

2019.08 output:

```
(Int Scalar)
(Int Int)
```

`HEAD` output:

```
(Int Scalar)
(Int Scalar)
```

What happens with `%h.Map` is about the fact that hashes are implicitly containerizing their values:

```
my @a = <1 2>; 
say @a.VAR.^name; # Array
my %h = a => @a; 
say %h<a>.VAR.^name; # Scalar
```

`Map` was implicitly decontainerizing its values, as the code above displays. But it doesn't anymore. Instead it preserves everything as it gets it. In our case, it preserves `Scalar`s from the hash. What actually happens with the clas initialization is something like:

```
my @a = <1 2>;
my $a = @a;
my @b = $a;
say @b; # [[1, 2]]
```

This is nothing else but a downside of using containers. The following code:

```
class A { has @.a; }; 
my %h = a => [1,2]; 
A.new(|%h.map( { $_.key => $_.value<> } ).Map).say
```

works as expected.

[...]

PR #3174, opened 2019-09-09
Provide means of preserving the decontainerization of values on Hash -> Map coercion

vrurg:
A solution for perl6/problem-solving#99 and rakudo/rakudo#3168.

This solution provides a way to control decontainerization of values my `Map` externally by supplying `:DECONT` named argument to `STORE` method when called with `:INITALIZE`. _This would allow any other `Map`-deriving class to provide functionality similar to that of `Hash` with minimum amount of code._

### DIFF
--- a/src/core.c/Hash.pm6
+++ b/src/core.c/Hash.pm6
@@ -17,7 +17,7 @@ my class Hash { # declared in BOOTSTRAP
             nqp::create(Map), Map, '$!storage',
             nqp::getattr(self,Map,'$!storage')
           ),
-          nqp::create(Map).STORE(self, :INITIALIZE)
+          nqp::create(Map).STORE(self, :INITIALIZE, :DECONT)
         )
     }
     method clone(Hash:D:) is raw {
--- a/src/core.c/Map.pm6
+++ b/src/core.c/Map.pm6
@@ -295,20 +295,30 @@ my class Map does Iterable does Associative { # declared in BOOTSTRAP
             !! "Cannot change key '{key}' in an immutable {self.^name}"
     }
 
-    # Directly copy from the other Map's internals: the only thing we need
-    # to do, is to decontainerize the values.
-    method !STORE_MAP_FROM_MAP(\map --> Nil) {
+    # Directly copy from the other Map's internals.
+    method !STORE_MAP_FROM_MAP(\map, :$DECONT? --> Nil) {
         nqp::if(
           nqp::elems(my \other := nqp::getattr(map,Map,'$!storage')),
           nqp::stmts(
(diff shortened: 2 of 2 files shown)


vrurg:
Tests provided by perl6/roast#575

vrurg:
A note to simplify reviewing. As @AlexDaniel stated on IRC, the code look bloated and I agree to him. It would be much simpler to just decont values in `Hash::Map` method. The only point of this approach is emphasized in the first comment: allowing inheriting classes to decont values with just a single parameter. If value of this feature is considered overestimated I will close this PR and re-work in other way.

lizmat:
I think the whole matter of Map / Hash semantics needs a problem solving ticket.  I've tried in the past to make this somewhat consistent and performant, which was recently reverted I believe.  It's the reason I wrote a [Tuple](https://modules.perl6.org/dist/Tuple).  I'm considering writing a `Dict` module with similar semantics with the `Associative` role.

vrurg:
@lizmat I agree in general to your opinion. I even consider a situation where Tuple and Dict become  part of the CORE so instead of discussing how exactly our base types behave we cover different possible scenarios and user needs.

But meanwhile, in the light of coming release date, I'd like to get the above mentioned blocker be resolved.

vrurg:
@lizmat something went wrong while pulling from upstream into my repo. I had to re-create my branch and then force-push it back to sort things out. This stripped off some commits and related comments.

Anyway, I have fulfilled your request by moving `nqp::if($DECONT)` out of the loops and duplicating them in each branch of `nqp::if`. Not that pretty as it'd be with multi-dispatch, but we don't have it for private methods yet.

BTW, I think this could be of interest for you. I tried a reference implementation of `Dict` with this PR. Here is what I got:

```
class Dict is Map {
    multi method new(Dict: *@args --> Dict:D) {
        self.Map::new.STORE(@args, :INITIALIZE, :DECONT);
    }

    method STORE (|c) {
        self.Map::STORE(|c, :DECONT);
    }
}

my $foo = 'foo';
my $d = Dict.new( (foo => $foo, int => 42) );

note $d.perl; # $(Dict.new((:foo("foo"),:int(42))))
note $d<foo>.VAR.^name; # Str

my $m = Map.new( (foo => $foo, int => 42) );
note $m.perl; # $(Map.new((:foo("foo"),:int(42))))
note $m<foo>.VAR.^name; # Scalar
```

lizmat:
Cool to see `Dict` easily implementable like that!

vrurg:
Ok, inlining... Sure, will redo. Thank you!

vrurg:
I think it is now ready. Simple benchmarking displays no visible performance degradation.

PR #3189, opened 2019-09-16
Build: use ?= to conditionally assign to variables, not ifndef

Kaiepi:
`ifndef` is not compatible with BSD make.

### DIFF
--- a/tools/templates/Makefile-backend-common.in
+++ b/tools/templates/Makefile-backend-common.in
@@ -4,22 +4,12 @@
 @bpv(GEN_CAT)@ = @shquot(@script(gen-cat.nqp)@)@ @backend@
 
 @bpv(BUILD_DIR)@ = @nfp(gen/@backend@)@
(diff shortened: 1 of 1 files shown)


Kaiepi:
Oop, now this makes nmake unhappy. How is stuff dependent on which type of make's used handled?

vrurg:
I have switched to ifndef for the purpose that they all are supported by all makes except for the prefix used (`.` for BSD, `!` for NMAKE, and nothing for GNU). This is why `@make_pp_pfx@` is there, it's set by Configure depending on what make is used.

Most likely, what's happened in your case, Configure found `gmake` and wants you to that instead of `make`.

Kaiepi:
That is the case. I just ignored when it said to use gmake and used BSD make instead since that worked up until recently. The build system seems to be able to find and make a Makefile for BSD make when gmake is installed, so I guess there's no issue to be fixed here.

PR #3196, opened 2019-09-24
Allow sub-classing of Proxy objects

lizmat:
Inspired by:

  https://stackoverflow.com/questions/58055309/a-scalar-with-memory-or-how-to-correctly-tie

This allows one to say:

    class Foo is Proxy {
    }

and implement additional functionality to the Proxy object, like keeping
track which values a variable has had (from the SO question).

With much help from Jonathan Worthington.

### DIFF
--- a/src/Perl6/Metamodel/ContainerSpecProtocol.nqp
+++ b/src/Perl6/Metamodel/ContainerSpecProtocol.nqp
@@ -1,0 +1,34 @@
+role Perl6::Metamodel::ContainerSpecProtocol {
+    has $!code_pair;
+
+    method get_container_spec($obj) {
+        $!code_pair
+    }
+
+    method set_container_spec($obj, $code_pair) {
+        $!code_pair := $code_pair;
+    }
+
+    method publish_container_spec($obj) {
+        # If we have a container specification here, install that
+        if $!code_pair {
+            nqp::setcontspec($obj, 'code_pair', $!code_pair);
+            $obj.HOW.compose_repr($obj);
+        }
+
+        # look in the parents for any if none here
+        else {
+            for self.mro($obj) -> $class {
+                if nqp::can($class.HOW, 'get_container_spec') {
+                    my $code_pair := $class.HOW.get_container_spec($class);
+                    if $code_pair {
+                        $!code_pair := $code_pair;
+                        nqp::setcontspec($obj, 'code_pair', $!code_pair);
+                        $obj.HOW.compose_repr($obj);
+                        last;
+                    }
+                }
+            }
+        }
+    }
+}
--- a/src/Perl6/Metamodel/ContainerSpecProtocol.nqp
+++ b/src/Perl6/Metamodel/ContainerSpecProtocol.nqp
@@ -10,22 +10,13 @@ role Perl6::Metamodel::ContainerSpecProtocol {
     }
 
     method publish_container_spec($obj) {
-        # If we have a container specification here, install that
-        if $!code_pair {
-            nqp::setcontspec($obj, 'code_pair', $!code_pair);
-            $obj.HOW.compose_repr($obj);
-        }
-
-        # look in the parents for any if none here
-        else {
-            for self.mro($obj) -> $class {
-                if nqp::can($class.HOW, 'get_container_spec') {
-                    my $code_pair := $class.HOW.get_container_spec($class);
-                    if $code_pair {
-                        nqp::setcontspec($obj, 'code_pair', $code_pair);
-                        $obj.HOW.compose_repr($obj);
-                        last;
-                    }
+        for self.mro($obj) -> $class {
+            if nqp::can($class.HOW, 'get_container_spec') {
+                my $code_pair := $class.HOW.get_container_spec($class);
+                if $code_pair {
+                    nqp::setcontspec($obj, 'code_pair', $code_pair);
+                    $obj.HOW.compose_repr($obj);
+                    last;
                 }
             }
         }
--- a/src/Perl6/Metamodel/ContainerSpecProtocol.nqp
+++ b/src/Perl6/Metamodel/ContainerSpecProtocol.nqp
@@ -15,7 +15,6 @@ role Perl6::Metamodel::ContainerSpecProtocol {
                 my $code_pair := $class.HOW.get_container_spec($class);
                 if $code_pair {
                     nqp::setcontspec($obj, 'code_pair', $code_pair);
-                    $obj.HOW.compose_repr($obj);
                     last;
                 }
             }
--- a/src/Perl6/Metamodel/ContainerSpecProtocol.nqp
+++ b/src/Perl6/Metamodel/ContainerSpecProtocol.nqp
@@ -22,8 +22,7 @@ role Perl6::Metamodel::ContainerSpecProtocol {
                 if nqp::can($class.HOW, 'get_container_spec') {
                     my $code_pair := $class.HOW.get_container_spec($class);
                     if $code_pair {
-                        $!code_pair := $code_pair;
-                        nqp::setcontspec($obj, 'code_pair', $!code_pair);
+                        nqp::setcontspec($obj, 'code_pair', $code_pair);
                         $obj.HOW.compose_repr($obj);
                         last;
                     }
--- a/src/Perl6/bootstrap.c/BOOTSTRAP.nqp
+++ b/src/Perl6/bootstrap.c/BOOTSTRAP.nqp
@@ -1759,6 +1759,7 @@ BEGIN {
         'store', $PROXY_STORE
     ));
     Proxy.HOW.compose(Proxy);
+    Proxy.HOW.compose_repr(Proxy);
 
     # Helper for creating a scalar attribute. Sets it up as a real Perl 6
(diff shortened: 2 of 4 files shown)


vrurg:
Well done! I felt like I need to implement a container for my `AttrX::Mooish`, and here is your work in the same area!

lizmat:
FWIW, this also scratched a long time personal itch  :-)

src/Perl6/Metamodel/ContainerSpecProtocol.nqp
```diff
@@ -0,0 +1,34 @@
...
+
+    method set_container_spec($obj, $code_pair) {
+        $!code_pair := $code_pair;
+    }
+
+    method publish_container_spec($obj) {
+        # If we have a container specification here, install that
+        if $!code_pair {
+            nqp::setcontspec($obj, 'code_pair', $!code_pair);
+            $obj.HOW.compose_repr($obj);
```
jnthn: Hm, I'm surprised to find this line here (and the similar one below); is it needed for something?

src/Perl6/bootstrap.c/BOOTSTRAP.nqp
```diff
@@ -1749,17 +1749,16 @@ BEGIN {
         nqp::getattr($cont, Proxy, '&!STORE')($var, $val)
     })));
     Proxy.HOW.add_method(Proxy, 'new', nqp::getstaticcode(sub ($type, :$FETCH!, :$STORE!) {
-        my $cont := nqp::create(Proxy);
+        my $cont := nqp::create($type);
```
jnthn: I'd `nqp::decont($type)` here.

src/Perl6/Metamodel/ContainerSpecProtocol.nqp
```diff
@@ -0,0 +1,34 @@
...
+        }
+
+        # look in the parents for any if none here
+        else {
+            for self.mro($obj) -> $class {
+                if nqp::can($class.HOW, 'get_container_spec') {
+                    my $code_pair := $class.HOW.get_container_spec($class);
+                    if $code_pair {
+                        $!code_pair := $code_pair;
```
jnthn: I don't think it should be assigned here, otherwise we lose track of which class it was actually set on (not terribly important, but still, feels odd to do this).

src/Perl6/Metamodel/ContainerSpecProtocol.nqp
lizmat: That's for when the class.HOW itself had `set_container_spec` called on it.  Specifically for the `Proxy` class itself?

src/Perl6/bootstrap.c/BOOTSTRAP.nqp
lizmat: Addressed with ead4fcc3c3

src/Perl6/Metamodel/ContainerSpecProtocol.nqp
lizmat: Addressed with 963646644f

BTW, for some reason, I couldn't get:

    if $class.HOW.get_container_spec($class) -> $code_pair {

to work: it would fail in compilation with:

    Too few positionals passed; expected 3 arguments but got 2
       at gen/moar/Metamodel.nqp:298  (blib/Perl6/Metamodel.moarvm:publish_container_spec)

src/Perl6/Metamodel/ContainerSpecProtocol.nqp
lizmat: Or are you implying that it should *always* do an nqp::setcontspec, even if `$!code_pair` hasn't been set?

lizmat: Aaah, I see.  The `.mro` has its own class as the first element, so it will find and handle the `container_spec` that way.

So addressed in fe1dcadaec

lizmat:
On merging, I will make sure we have tests  :-)

src/Perl6/Metamodel/ContainerSpecProtocol.nqp
jnthn: Ah, I see. The `Proxy.HOW.compose_repr(Proxy);` line is specific to the bootstrap case, and so should have been left behind in `BOOTSTRAP`; it's not needed on the typical path.

lizmat: Addressed with 5eea5f4eb6

src/Perl6/Metamodel/ContainerSpecProtocol.nqp
```diff
@@ -0,0 +1,23 @@
...
+
+    method set_container_spec($obj, $code_pair) {
+        $!code_pair := $code_pair;
+    }
+
+    method publish_container_spec($obj) {
+        for self.mro($obj) -> $class {
+            if nqp::can($class.HOW, 'get_container_spec') {
+                my $code_pair := $class.HOW.get_container_spec($class);
+                if $code_pair {
```
MasterDuke17: This is just idle curiosity, but could `if $class.HOW.get_container_spec($class) -> $code_pair {` be done? Does the `if ... ->` form work for/with binding?

lizmat: I get compilation errors if I do it like that.  I have no idea why and have no intent to look into NQP's guts to find out why that is not working.  I assume the `.HOW` in the expression is somehow throwing things off.

holli-holzer:
I was playing around with this tonight. `Proxy` is still not a regular class as it seems. I added an attribute to my subclass `has $.register-duplicates`, but when I try to set that in the constructor I'm getting `Unexpected named argument 'register-duplicates' passed`

lizmat:
@holli-holzer can you supply a code example that I can work with ?

holli-holzer:
@lizmat Yes, I posted it to the very SO thread you mention above just now

holli-holzer:
FWIW and your convenience, here are tests that should pass once you find the source of the recursion issues

```
class TestProxy is Proxy { has $.foo = 11; }

subtest 'default proxy attributes' =>
{
    my $proxy;
    lives-ok { $proxy := TestProxy.new(
        FETCH => method ( \SELF: ) { SELF.foo },
        STORE => method ( \SELF: $value ) { SELF.foo = $value }
    )}, "can instantiate Proxy subclass and default attributes";

    ok( $proxy.defined && $proxy == 11,  "default attribute value is correct" );
}

subtest 'proxy attributes' =>
{
    my $proxy;
    lives-ok { $proxy := TestProxy.new(
        foo => 99,
        FETCH => method ( \SELF: ) { SELF.foo },
        STORE => method ( \SELF: $value ) { SELF.foo = $value }
    )}, "can instantiate Proxy subclass and set attributes";

    ok( $proxy.defined && $proxy.foo == 99,  "attribute value is correct" );
}
```

jnthn:
> STORE => method ( \SELF: $value ) { SELF.foo = $value }

This line requires that `foo` is declared `is rw`, e.g.

```
class TestProxy is Proxy { has $.foo is rw = 11; }
```

lizmat:
@jnthn that doesn't solve the issue:

    class TestProxy is Proxy { has $.foo is rw = 11; };
    TestProxy.new( FETCH => { ... }, STORE => { ... }, foo => 42 );
    # Unexpected named argument 'foo' passed

The thing I've come up with so far, is a custom `new` and using a `TWEAK` to set additional parameters:

    class A is Proxy {
        has $.position;

        method new(:&FETCH!, :&STORE!, *%_) is raw {
            my $self := self.Proxy::new(:&FETCH,:&STORE);
            $self.VAR.TWEAK(|%_);
            $self
        }

        method TWEAK(:$!position = 42 --> Nil) { }
    }

lizmat:
I guess we could consider in the MOP to put such a method `new` in the class if it is subclassing `Proxy` ?

lizmat:
@jnthn BTW, I also tried removing the `new` method from the Proxy bootstrap, hoping that `Mu.new` would pick it up.   But that results in `Cannot invoke this object (REPR: Null; VMNull)` at installation.

lizmat:
Alternate solution:

    role ProxyTweak is Proxy {
        method new(:&FETCH!, :&STORE!, *%_) is raw {
            my $self := self.Proxy::new(:&FETCH,:&STORE);
            $self.VAR.TWEAK(|%_);
            $self
        }
    }

    class A does ProxyTweak {
        has $.position;

        method TWEAK(:$!position = 42 --> Nil) { }
    }

Perhaps we should add the `ProxyTweak` role to the core?

holli-holzer:
I would say the above custom constructor is good enough. One can then put `Proxy::Inheritable` that implements that into the ecosystem and call it a day. I don't really like the idea of a role only for a custom constructor. That feels weird.

jnthn:
> that doesn't solve the issue

Ah, I didn't expect it to, though wasn't clear on that (was rushing out somewhere). It's just needed in addition to whatever else.

holli-holzer:
I have pushed a preliminary version of [Scalar::History](https://github.com/holli-holzer/perl6-Scalar-History)

I will wait a few days to see how this turns out and hope not too many changes are required before putting it on cpan =)

PR #3199, opened 2019-09-28
Re-consider roles application, submethods and phasers semnatics.

vrurg:
- perl6/problem-solving#103
- Related tests are in perl6/roast#581

### DIFF
--- a/src/Perl6/Metamodel/Concretization.nqp
+++ b/src/Perl6/Metamodel/Concretization.nqp
@@ -35,10 +35,11 @@ role Perl6::Metamodel::Concretization {
     }
 
     method concretization_lookup($obj, $ptype, :$local = 0, :$transitive = 1, :$relaxed = 0) {
-        $ptype := nqp::decont($ptype);
-        my $id := ~nqp::objectid($ptype);
         self.'!rebuild_table'() if nqp::elems(%!conc_table) < nqp::elems(@!concretizations);
-        return [0] unless $local || $transitive || nqp::elems(%!conc_table);
+        return [0] unless !$local || $transitive || nqp::elems(%!conc_table);
+        $ptype := nqp::decont($ptype);
+        return [0] if $obj.HOW.is_composed($obj) && !nqp::istype($obj, $ptype);
+        my $id := ~nqp::objectid($ptype);
         my @result;
         if nqp::existskey(%!conc_table, $id) {
             return [1, %!conc_table{$id}];
--- a/src/Perl6/Metamodel/Concretization.nqp
+++ b/src/Perl6/Metamodel/Concretization.nqp
@@ -1,36 +1,76 @@
 # Support for mapping of non-specialized roles into their concretized state.
 role Perl6::Metamodel::Concretization {
     has @!concretizations;
+    has %!conc_table;
 
     method add_concretization($obj, $role, $concrete) {
         @!concretizations[+@!concretizations] := [$role, $concrete];
+        %!conc_table := nqp::hash(); # reset the cache
     }
 
-    method concretization_lookup($obj, $ptype, :$local, :$transitive) {
+    method concretizations($obj, :$local = 0, :$transitive = 1) {
+        my @conc;
+        for @!concretizations {
+            nqp::push(@conc, $_[1]);
+            if $transitive && nqp::can($_[1].HOW, 'concretizations') {
+                for $_[1].HOW.concretizations($_[1], :$local) {
+                    nqp::push(@conc, $_);
+                }
+            }
+        }
+        unless $local {
+            for self.parents($obj, :local) {
+                for $_.HOW.concretizations($_, :$local) {
+                    nqp::push(@conc, $_)
+                }
+            }
+        }
+        @conc
+    }
+
+    method !rebuild_table() {
+        for @!concretizations {
+            %!conc_table{~nqp::objectid(nqp::decont($_[0]))} := nqp::decont($_[1]);
+        }
+    }
+
+    method concretization_lookup($obj, $ptype, :$local = 0, :$transitive = 1, :$relaxed = 0) {
+        $ptype := nqp::decont($ptype);
+        my $id := ~nqp::objectid($ptype);
+        self.'!rebuild_table'() if nqp::elems(%!conc_table) < nqp::elems(@!concretizations);
+        return [0] unless $local || $transitive || nqp::elems(%!conc_table);
         my @result;
-        if nqp::istype($obj, $ptype) {
+        if nqp::existskey(%!conc_table, $id) {
+            return [1, %!conc_table{$id}];
+        }
+        if $relaxed {
+            # Try search by role group for curryings. The first match is ok. Used by FQN method calls.
             for @!concretizations {
-                if nqp::decont($_[0]) =:= nqp::decont($ptype) {
+                next unless $_[0].HOW ~~ Perl6::Metamodel::CurriedRoleHOW;
+                if nqp::decont($_[0].HOW.curried_role($_[0])) =:= $ptype {
                     return [1, $_[1]];
                 }
-                # Do preliminary type check for concrete role to avoid extra calls.
-                if $transitive && nqp::istype($_[1], $ptype) {
+            }
+        }
+        if $transitive {
+            for @!concretizations {
+                if nqp::istype($_[1], $ptype) {
                     @result := $_[1].HOW.concretization_lookup($_[1], $ptype, :$local, :transitive(1));
                     return @result if @result[0];
                 }
             }
-            unless $local {
-                for self.parents($obj, :local(1)) {
-                    @result := $_.HOW.concretization_lookup($_, $ptype, :local(0), :$transitive);
-                    return @result if @result[0];
-                }
+        }
+        unless $local {
+            for self.parents($obj, :local) {
+                @result := $_.HOW.concretization_lookup($_, $ptype, :local(0), :$transitive);
+                return @result if @result[0];
             }
         }
         [0]
     }
 
-    method concretization($obj, $ptype, :$local = 0, :$transitive = 1) {
-        my @result := self.concretization_lookup($obj, $ptype, :$local, :$transitive);
+    method concretization($obj, $ptype, :$local = 0, :$transitive = 1, :$relaxed = 0) {
+        my @result := self.concretization_lookup($obj, $ptype, :$local, :$transitive, :$relaxed);
         nqp::die("No concretization found for " ~ $ptype.HOW.name($ptype)) unless @result[0];
         @result[1]
     }
--- a/src/Perl6/Metamodel/Concretization.nqp
+++ b/src/Perl6/Metamodel/Concretization.nqp
@@ -5,7 +5,9 @@ role Perl6::Metamodel::Concretization {
 
     method add_concretization($obj, $role, $concrete) {
         @!concretizations[+@!concretizations] := [$role, $concrete];
+        nqp::scwbdisable();
         %!conc_table := nqp::hash(); # reset the cache
+        nqp::scwbenable();
     }
 
     method concretizations($obj, :$local = 0, :$transitive = 1) {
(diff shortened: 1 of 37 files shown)


vrurg:
# Stage 1

Provide MRO with roles included. This is necessary to traverse submethods when they're no more copied over from roles into consuming classes. Proper implementation requried:

- added support for language revision recording on roles. Not yet used but will be required to prevent consumption of `6.e` roles by `6.d-` classes.
- support for `hides` and `is hidden` traits on roles

coke:
perl6/problem-solving#103 (to get a clickable link)

vrurg:
# Stage 2

Implementation of `BUILD`/`TWEAK`/`DESTROY` support in roles.

vrurg:
A note `mro` implementation when called with `:roles` argument.

I expect the return of `.^mro` to be directly used by any kind of MRO tree-walkers. I.e. as long as it is currently possible for anyone to call a method directly on a class object returned by `mro`, I expected same be possible for roles. For this reason the method returns specialized roles. This is especially handy in for cases of parameterized roles:

```
class C does R[Int] { ... }
C.^new.mro[1].role_method;
```

vrurg:
Where exactly? Or do you mean  decrement is slow? Sorry, I miss your point.

But basically, except for bug fixes, the whole thing is yet raw and will need polishing.

ugexe:
I mean all $i++ were changed to —$i for a reason at one point.

ugexe:
s/$i++/$i—/

vendethiel:
wasn't that before the WANTED work?

vrurg:
@ugexe I now vaguely remember being told once about this issue. Will change this just to be on the safe side.

vrurg:
I tested nqp on a simple loop like:

```
while $i-- > 0 {
    $sum := $sum + $i;
}
```

With three different decrements. Two others, obiously: `--$i`, and `$i - 1` in the loop body. Here is what I got:

```
$i--:   0.6671936511993408
--$i:   0.5243737697601318
$i - 1: 0.5107505321502686
```

While `$i--` loop is 26% slower than the alternatives, I don't thinks it's significant overhead for loops with comparatively heavy bodies, contrary to the lightweight summing used in the test. So, perhaps it doesn't worth worrying a lot about it.

Though `--$i` is also aesthetic enough but performs better. ;)

vrurg:
# Stage 3

Role->class composition. See my [comment](https://github.com/perl6/problem-solving/issues/103#issuecomment-537711003) in perl6/problem-solving#103.

vrurg:
# Stage 4 and the last

- Implemented `.-`, `.?+`, `.?-` operators.
- Implemented _relaxed_ lookup of parameterized roles in class-qualified method resolution. I.e.:

   ```
   class Foo does Bar::R[Str] {
       method foo { self.Bar::R::bar }
   }
   ```

   works now as expected.

With a couple of spectests covering the changes this is ready for merge.

lizmat:
Question: would this work also fix https://github.com/lizmat/Method-Also/issues/1 ?

vrurg:
@lizmat I expected this. This PR mostly works on the things which happen after  roles are composed.

jnthn:
> Implemented .-, .?+, .?- operators.

These don't seem to fit so well with the wider language design. The `.?`, `.+`, and `.*` all have somewhat guessable semantics through the analogy with the regex quantifiers. And it turns out there is already a precedent established for how quantifiers are given semantic tweaks, which is via a suffix (e.g. to do minimal matching we have `??`, `+?`, and `*?`). *If* we're going to add syntax for this top-down walking order, I'd rather follow this existing precedent than to create a new one.

Before we go *there*, though, I'd like to ask: what do we actually need this for? Construction time is handled by the generated `BUILDALL`, and I find it hard to think of a use case where we'd want to go top-down rather than bottom-up once constructed. Of note, destruction is bottom up.

jnthn:
I'd also like to point out that we already have `Mu.WALK`, which is a means to get all methods when we walk the tree in a certain way. It has `ascendent` and `descendent` options, and more options besides. Thus there already *is* a means to query for such an order; if we were to extend `WALK` to also have the new `submethod` semantics - presumably as an option for back-compat - then that would perhaps be a better way to provide for this than new syntax for a rare case. I think `.WALK(:descendant, :role-submethods).map({ $_($args) })` or some such is probably more self-explanatory and, for how rare I expect to be, appropriately huffmanized.

vrurg:
> Before we go _there_, though, I'd like to ask: what do we actually need this for? Construction time is handled by the generated `BUILDALL`, and I find it hard to think of a use case where we'd want to go top-down rather than bottom-up once constructed. Of note, destruction is bottom up.

A real life example of a use case for the semantics (syntax is irrelevant here) was an application which primary `App` class was built from a handful of classes/roles. Each was defining own command line options which the main class need not to know about. Having submethods just returning arrays of option definitions turned out to be the most effective way to collect them. Reverse walk order allowed to override options if necessary.

To generalize, submethods are great for exporting class/role specific data and iterating over them is the best was to collect it.

vrurg:
> I'd also like to point out that we already have `Mu.WALK`

Enough is said in this quote. :) I was in doubt if it makes sense for a new method on `Mu`, but since one exists already – it's a big relief! As I said earlier, I'm not a big fan of new operators too.

Unfortunately, `WALK` is not documented and it's not easy to find it unless one knows what to look for. This would have taken care of too.

So, if you don't mind, I remove ops, teach WALK a new trick – and merge.

jnthn:
> So, if you don't mind, I remove ops, teach WALK a new trick

:+1: on this; if the pattern then in the wild turns out to be used enough to justify shorter syntax, we can consider that then.

> – and merge.

I'm still looking through the looser parametric calls thing, and must admit a little discomfort there; will be back with a comment soon.

jnthn:
Doing the same role with two parameterizations is entirely possible; for example:

```
role R[::T] { multi method m(T $x) { say T.^name } }
class C does R[Int] does R[Str] { };
C.m(42);
C.m("wat");
```

Would output:

```
Int
Str
```

And there's probably various patterns that could be used with this. Using the qualified call syntax is not so useful given the `multi` means you can still reach either from the class, but if one does have a situation where wishing to disambiguate then it still needs the `constant` trick:

```
role R[::T] { method m(T $x) { say T.^name } }
class C does R[Int] does R[Str] {
    method m($x) {
        my constant RI = R[Int];
        my constant RS = R[Str];
        self.RI::m($x.Int);
        self.RS::m($x.Str);
    }
}
```

So there is a way, as before, though part of me wonders if we can't (not in this PR!) make `self.R[Int]::m($x.Int)` Just Work, given it's actually a syntax error today.

Turning back to this PR: I do wonder if we should only allow the shortcut when it's unambiguous. Today we just pick one of the concretizations, and that feels like asking for accidents. It'd be unfortunate if a mechanism that exists for helping resolve composition ambiguities were itself to embrace them!

lizmat:
> On 7 Nov 2019, at 17:10, Jonathan Worthington <notifications@github.com> wrote:
> 
> Doing the same role with two parameterizations is entirely possible; for example:
> 
> role R[::T] { multi method m(T $x) { say T.^name } }
> class C does R[Int] does R[Str] { };
> C.m(42);
> C.m("wat");

Which is actually a spectest, I found out recentlyn  :-)

vrurg:
BTW, parameterized referencing was discussed in perl6/problem-solving#110. 

I didn't want extra checking logic for this rather rare case. It's steel feels ok to me to explicitly define that _the first role in the list is used_. This effectively resolves any ambiguity. But if you consider it not sufficient, I'll make it fail.

src/Perl6/Metamodel/Concretization.nqp
```diff
@@ -1,36 +1,84 @@
...
-        if nqp::istype($obj, $ptype) {
+        if nqp::existskey(%!conc_table, $id) {
+            return [1, %!conc_table{$id}];
+        }
+        if $relaxed {
+            # Try search by role group for curryings. The first match is ok. Used by FQN method calls.
             for @!concretizations {
-                if nqp::decont($_[0]) =:= nqp::decont($ptype) {
+                next unless $_[0].HOW ~~ Perl6::Metamodel::CurriedRoleHOW;
+                if nqp::decont($_[0].HOW.curried_role($_[0])) =:= $ptype {
```
Kaiepi: Can this check be written in a way that would allow people to use this meta-role with their own HOWs for parameterized types like it did before?

vrurg: Any suggestions? Code which would be broken with this change?

Kaiepi: > Code which would be broken with this change?

This would make it so I can't use this meta-role with one of the HOWs in a typeclass library I've been working on for a while now (which used to be a monad library). While it's not something I think will be in a releasable state for quite a while yet, if at all, it's useful for testing how you'd go about implementing some behaviours related to parameterization whose syntax is supported, but none of the parametric types in Rakudo know how to deal with. Should they be behaviours that should be supported, there would be less risk involved in implementing them for Rakudo's parametric types this way since the solutions will have already been tested through this library.

> Any suggestions?

What I'd do is narrow down the types that were concretized to types that are both parametric and inheritalizable, then check if the type parameterized is the parametric type passed:

```perl6
next unless $_[0].HOW.archetypes.parametric && $_[0].HOW.archetypes.inheritalizable;
if nqp::typeparameterized($_[0]) =:= $ptype { ... }
```

While types that aren't parameterized but fit these archetypes (like `Perl6::Metamodel::ParametricRoleHOW`) would pass the first check if they ended up here, `nqp::typeparameterized` is not one of the parametric type ops that throws if it doesn't receive a type that can actually be used with it.

Kaiepi: Actually, now that I think about it, there may be ways I can handle typeclass instances without needing to make the typeclass a parent of the types they mark as instances, so I may not need this meta-role in the end. Still, someone may need to make a punnable parametric HOW some day.

vrurg: @Kaiepi I wasn't able to commit my yesterday's change, but now it looks like this:

```
                next unless $_[0].HOW.archetypes.parametric;
                my $conc := nqp::can($_[0].HOW, 'curried_role')
                                ?? $_[0].HOW.curried_role($_[0])
                                !! $_[0];
                if nqp::decont($conc) =:= $ptype {
```

So, eventually, it should be fine with custom `RoleHOW`s and one could even simulate curryings if desire.

vrurg: @Kaiepi I think my code would fulfill your requirements.

vrurg:
@jnthn I'm finally done with this. `.-`, `.?+`, and `.?-` were removed completely. Instead, `WALK` method was taught to:

- respect roles (by request)
- consider not only methods but submethods too. This is by default but with `:!method` or `:!submethods` corresponding classes could be excluded from the result.

It also now returns a `WalkList` instance – a new List type which can batch-invoke methods it carries:

```
@results = $obj.WALK("method").invoke(1,2, :a, :b);
@results = $obj.WALK("another")($arg1, :named);
```

With `WalkList` it is also possible to wrap any method-thrown exception into a `Failure` and use it as if returned by methods:

```
$obj.WALK("method").quiet.invoke.[1] ~~ Failure; # True if second method thrown
```

Invocation also itemizes `Slip`s into a `List` so that `@results` always contains as many elements as many methods were invoked. If anybody needs another behavior then:

```
@results = $obj.WALK("method")».($obj);
```

is always at his disposal.

With regard to relaxed FQN, I succeed in getting it work the right way. Best it demoed by https://github.com/perl6/roast/blob/problem-solving-103/S12-methods/qualified.t

What actually happens now is that qualification is first attempted on the immediate owner of the code would it be class or a role concretization. This way we always now precisely what parameterization must be used to find the destination method. If this qualification fails then we fall back to the invocator and try it from there. 

Note that in this behavior ambiguities are only take place when two roles are applied directly to a class. This allows to worry less about cases when a class consumes a role, and another third-party role which also does the former role and then FQN-dispatch to it. The developer of the third-party role must not be worried about the class it will be applied to. As well as the class developer not necessarily must know all the implementation details of a module he's using.

So far, I'm done with it.

Issue #2092, opened 2018-07-18
"only alpha methods may be detached" message could be a bit clearer/accurater


zoffixznet:
***See [Newcomer Guide to Contributing to Core Perl 6](https://rakudo.party/post/Newcomer-Guide-to-Contributing-to-Core-Perl-6) for tips on fixing this Issue***

----------------

If you try to detach a wrong method, you get this error:

```
<Zoffix_> m: 42 .&abs
<camelia> rakudo-moar b30800c8e: OUTPUT: «===SORRY!=== Error while compiling
        <tmp>␤Malformed postfix call (only alphabetic methods may be detached)␤at
        <tmp>:1␤------> 42 .⏏&abs␤»
```

Though technically correct (the best kind of correct!), there are two pieces of information it lacks that could make it more accurate:

1. It's not just methods, but entire method *chains* that are forbidden. The error suggests writing `42 .self.&abs` is allowed, but it isn't because the `.&` call is present in the detached method chain, even though it's not the method where detachment is at.
2. It's not just alphabetic methods that can be detached. `.I'love-detachin42` can also be detached. Come to think of it, all regular identifier-like method names can be detached and it's the type of methodop used that's problematic, and not the method name itself. So it probably should be reworded to something like `only basic method calls that only use a dot can be detached` or something.

AlexDaniel:
This is now resolved! Thanks @RayMPerry!

PR #3204, opened 2019-10-01
Issue #2092 | Clarify detached method error messaging.

RayMPerry:
Originally, this error said "only alphabetic methods may be detached"
but those aren't the only methods that can be detached. The new
messaging is a bit clearer on valid detachment candidates.

To trigger this, eval: `42 .&abs`.

Resulting error should be: `Malformed postfix call (only basic method calls that exclusively use a dot can be detached)`.


RayMPerry:
Oh, hey, this actually got in. Neat!

AlexDaniel:
Looks good to me! Thanks! Sorry it took us so long :)

PR #3218, opened 2019-10-06
Implement ¿? and ¡! as alternative to ""

lizmat:
This is only semi-serious.  Yet, there's a lot of Spanish speaking people
in the world, that may find this feature attractive / useful.  Mostly done
to see whether the approach would work, since it did *not* work as an
external module.

The external module `Slang::SpanishQuotes` that does not work:

    module Slang::SpanishQuotes:ver<0.0.1>:auth<cpan:ELIZABETH> { }
  
    sub EXPORT(|) {
        my role SpanishQuotesGrammar {
            token quote:sym<question> {
                :dba('question mark quotes')
                '¿' ~ '?'
                <nibble(self.quote_lang(self.slang_grammar('Quote'), '¿', '?', ['qq']))>
            }
            token quote:sym<exclamation> {
                :dba('exclamation mark quotes')
                '¡' ~ '!'
                <nibble(self.quote_lang(self.slang_grammar('Quote'), '¡', '!', ['qq']))>
            }

            token metachar:sym<'> {
                <?[ ' " ‘ ‚ ’ “ „ ” ｢ ¿ ¡ ]> <quote=.LANG('MAIN','quote')> <.SIGOK>
            } 
            token escape:sym<'> {
                <?[ ' " ‘ ‚ ’ “ „ ” ｢ ¿ ¡ ]> <quote=.LANG('MAIN','quote')>
            }
        }
    
        my role SpanishQuotesActions {
            method quote:sym<question>($/){ make $<nibble>.ast; }
            method quote:sym<exclamation>($/){ make $<nibble>.ast; }
        }
    
        $*LANG.define_slang:
          'MAIN',
          $*LANG.slang_grammar('MAIN').^mixin(SpanishQuotesGrammar),
          $*LANG.actions.^mixin(SpanishQuotesActions);
    
        {}
    }

### DIFF
--- a/src/Perl6/Actions.nqp
+++ b/src/Perl6/Actions.nqp
@@ -8507,6 +8507,7 @@ class Perl6::Actions is HLL::Actions does STDActions {
     method quote:sym<crnr>($/) { make $<nibble>.ast; }
     method quote:sym<question>($/){ make $<nibble>.ast; }
     method quote:sym<exclamation>($/){ make $<nibble>.ast; }
+    method quote:sym<dash>($/){ make $<nibble>.ast; }
     method quote:sym<qq>($/)   { make $<quibble>.ast; }
     method quote:sym<q>($/)    { make $<quibble>.ast; }
     method quote:sym<Q>($/)    { make $<quibble>.ast; }
--- a/src/Perl6/Actions.nqp
+++ b/src/Perl6/Actions.nqp
@@ -8505,6 +8505,8 @@ class Perl6::Actions is HLL::Actions does STDActions {
     method quote:sym<ldblq>($/){ make $<nibble>.ast; }
     method quote:sym<hdblq>($/){ make $<nibble>.ast; }
     method quote:sym<crnr>($/) { make $<nibble>.ast; }
+    method quote:sym<question>($/){ make $<nibble>.ast; }
+    method quote:sym<exclamation>($/){ make $<nibble>.ast; }
     method quote:sym<qq>($/)   { make $<quibble>.ast; }
     method quote:sym<q>($/)    { make $<quibble>.ast; }
     method quote:sym<Q>($/)    { make $<quibble>.ast; }
--- a/src/Perl6/Grammar.nqp
+++ b/src/Perl6/Grammar.nqp
@@ -3406,6 +3406,8 @@ grammar Perl6::Grammar is HLL::Grammar does STD {
     token quote:sym<ldblq> { :dba('low curly double quotes') '„' ~ <[”“]> <nibble(self.quote_lang(self.slang_grammar('Quote'), '„', ['”','“'], ['qq']))> }
     token quote:sym<hdblq> { :dba('high curly double quotes') '”' ~ <[”“]> <nibble(self.quote_lang(self.slang_grammar('Quote'), '”', ['”','“'], ['qq']))> }
     token quote:sym<crnr>  { :dba('corner quotes') '｢' ~ '｣' <nibble(self.quote_lang(self.slang_grammar('Quote'), '｢', '｣'))> }
+    token quote:sym<question> { :dba('question mark quotes') '¿' ~ '?' <nibble(self.quote_lang(self.slang_grammar('Quote'), '¿', '?', ['qq']))> }
+    token quote:sym<exclamation> { :dba('exclamation mark quotes') '¡' ~ '!' <nibble(self.quote_lang(self.slang_grammar('Quote'), '¡', '!', ['qq']))> }
     token quote:sym<q> {
         :my $qm;
         'q'
@@ -5298,7 +5300,7 @@ grammar Perl6::QGrammar is HLL::Grammar does STD {
 
     role ww {
         token escape:sym<'> {
-            <?[ ' " ‘ ‚ ’ “ „ ” ｢ ]> <quote=.LANG('MAIN','quote')>
+            <?[ ' " ‘ ‚ ’ “ „ ” ｢ ¿ ¡ ]> <quote=.LANG('MAIN','quote')>
         }
         token escape:sym<colonpair> {
             <?[:]> <!RESTRICTED> <colonpair=.LANG('MAIN','colonpair')>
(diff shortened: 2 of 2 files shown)


lizmat:
Not sure this would warrant a problem-solving ticket.  Would gladly accept a way to make this work as a `slang` as well.  Or just closing this PR as "not appropriate" / "too featureful" :-)

AlexDaniel:
You can add it to this page: https://github.com/rakudo/rakudo/wiki/save-me-from-ASCII

So the idea is that the final ? or ! is part of the string? Or how should it work?

AlexDaniel:
Oh:
> Keep in mind that it is common when printing dialogue in Spanish to dispense with quote marks entirely and use a long dash ("—"), sometimes known as an em dash or "raya" in Spanish, to indicate the beginning and end of the quotation or a change in speaker.

I'm really not aware how it works in Spanish, but if that's the case, then why ¿? and ¡! but not — —?

lizmat:
> So the idea is that the final ? or ! is part of the string? Or how should it work?

No, they would *not* be part of the string.  At least not the way I just implemented it.  Including them would be, eh... interesting and quite a lot more work, I'd think/

lizmat:
> why ¿? and ¡! but not — —?

Good question.  Added with e35881a673

lizmat:
> You can add it to this page: https://github.com/rakudo/rakudo/wiki/save-me-from-ASCII

done.

llelf:
> I'm really not aware how it works in Spanish, but if that's the case, then why ¿? and ¡! but not — —?

Is it really a good idea?  Primary uses of em-dash are different.

AlexDaniel:
@llelf I don't know. Can you explain how these quotes (¿?, ¡!, — —) are used in real life?

llelf:
¿? and ¡! are sentence/closure-only.  ¿Por qué no?  ¡Qué sorpresa!  You can mix them, e. g. ¿¡ + !?, or even ¡ + ? (but I don’t know how popular is this).

Em-dashes are for
1) breaking the flow, — like a colon
2) quotation:
  a) — Question.
     — Answer.
  b) $Proverb — someone.
3) as kind of parenthesis—like this—in the sentence. That is the only Spanish use.
4) some more typographic things, e. g. interrup— she was interrupted.

FCO:
Em-dashes has the same use in Portuguese.

tbrowder:
Should the grapheme be in the opener token if intended as a bracketing character?

tbrowder:
> Em-dashes has the same use in Portuguese.

And same uses in English--I use them frequently.

vrurg:
I don't think em-dashes must ever be considered as quotes. This semantics makes purely symbolical sense, they don't really quote but just separate one part of a text from another. This is how I read @llelf note and this is the same as in Ukrainian and Russian.

alabamenhu:
> ¿? and ¡! are sentence/closure-only. ¿Por qué no? ¡Qué sorpresa! You can mix them, e. g. ¿¡ + !?, or even ¡ + ? (but I don’t know how popular is this).
> 
> Em-dashes are for
> 
> 1. breaking the flow, — like a colon
> 2. quotation:
>    a) — Question.
>    — Answer.
>    b) $Proverb — someone.
> 3. as kind of parenthesis—like this—in the sentence. That is the only Spanish use.
> 4. some more typographic things, e. g. interrup— she was interrupted.

Em, not really.  Em dashes in Spanish have quite a few uses.  Dialog is a huge use:

```
    — ¿Cómo estás? —me exclamó— ¡Cuánto tiempo 
sin verte!
    — Es verdad, hará cinco años en septiembre.
    — ¡No jodas! ¿Tanto tiempo? Bueno, lo siento.
Sabes que llevo un tiempo viviendo fuera.
```
They don't close when in dialog, but they do everywhere else (unlike English, which allows them left open if the sentence ends).  They are also used formally speaking instead of hyphens for lists item demarcation when full sentences are used.  

Also, the `¿?` and `¡!` can be used midsentence, which is most common for tag questions: *La respuesta es cinco, ¿verdad?*  Honestly, for me as a Spanish-speaker, I would have found it more interesting to have the following be equivalent:

```
!( $foo || $bar) === ¡ $foo || $bar !
?( $foo || $bar) === ¿ $foo || $bar ?
```

But when I played around with it as a slang, IIRC, I found some problems with the terminating ! and ? that were getting interpreted as prefix/postfix operators.

JJ:
Cool as this is, and much as I like new Unicode symbols, I don't think these make sense _as quotes_, even if they are a pair. If we see ¿something like this? in Spanish, quotation will never come to mind. It will be see as a question. And ¡This will be an interjection!. We won't think "this is a string", but "Why are they shouting?" Still, I like the em-dash, and it would really make sense.

JJ:
Still, it's an option, and people are not obliged to use them. If this does not hamper or make anything else slower, why not give it a go?

llelf:
> Still, I like the em-dash, and it would really make sense.

Why? It's primary use is absolutely different.

(Actually the more I think I think about it, the more it doesn't make any sense, ¿? ¡! included).

JJ:
> > Still, I like the em-dash, and it would really make sense.
> 
> Why? Its primary use is absolutely different.

But it's still a way to enclose a sentence, so it makes sense as a quote.
> 
> (Actually the more I think I think about it, the more it doesn't make any sense, ¿? ¡! included).

But on the other hand, they're not mandatory. And if it does not cost much to add it to the code base...

Altai-man:
>But on the other hand, they're not mandatory. And if it does not cost much to add it to the code base...

If they are not mandatory, what's the benefit? It just increases overall complexity of the machine without making the code easier to write or to read. There is absolutely no need to pack everything that can be packed into the language because we can.

llelf:
> if it does not cost much to add it to the code base...

then it probably not worth adding (I'm joking).

On a serious note, there will be problems with using all of them as (user-defined) operators.

AlexDaniel:
It's always nice to have a PR to see how the compiler will be changed. However, there's now a problem-solving ticket, so please discuss it there: https://github.com/Raku/problem-solving/issues/184

We'll reopen this PR if we decide to implement it.

PR #3219, opened 2019-10-06
Allow for subscripts to be used in identifiers

lizmat:
Just as digits.  Allows for:

    my $H₂SO₄ = 42; say $H₂SO₄;  # 42

### DIFF
--- a/src/Perl6/Grammar.nqp
+++ b/src/Perl6/Grammar.nqp
@@ -566,7 +566,7 @@ grammar Perl6::Grammar is HLL::Grammar does STD {
     }
 
     token identifier {
-        <.ident> [ <.apostrophe> <.ident> ]*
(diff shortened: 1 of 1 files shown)


lizmat:
Updated https://github.com/rakudo/rakudo/wiki/save-me-from-ASCII

jnthn:
Also, given how common identifier parsing is, we should at least ask if we want to pay for this. If we do, it's cheaper to do it as a character range rather than the way it's done here (I'm assuming Unicode is sensible enough to have them in order...)

AlexDaniel:
It's nice to have a PR but without a consensus there's no need to keep this open. We can reopen and merge it if we decide to go this way. Further discussion here: https://github.com/Raku/problem-solving/issues/183

PR #3225, opened 2019-10-11
[IP6NS Grant] Prepare for upcoming sync UDP/UNIX/raw socket support

Kaiepi:
Some constants need tweaking.

See https://github.com/MoarVM/MoarVM/pull/1192 and https://github.com/perl6/nqp/pull/583.


vrurg:
I didn't find a ticket on this, so writing here. Is it possible to have changes to `IO` be `6.e` specific? As far as you say they're breaking backcompat that'd be reasonable.

Kaiepi:
It would make the changes coming directly after this a lot easier, so possibly. This stage of the grant already needed a problem solving issue, but on second thought that needs to be dealt with before I can even prepare for it.

Issue #1969, opened 2018-06-26
^D doesn't work in REPL on Windows Server 2008


zoffixznet:
Despite working fine on Win7, pressing ^D on Win2008 server doesn't exit the REPL. N.B.: it's possible the problem is due to my using Remote Desktop, rather than this being a problem for a specific windows version.

To be more exact: pressing Ctrl+D just types "^D" in both Windows versions and after pressing ENTER, the Win7 version exits cleanly, while Win2008S version shows the bogus statement error.

```
C:\Users\Administrator>perl6 -v
This is Rakudo Star version 2018.04.1 built on MoarVM version 2018.04.1
implementing Perl 6.c.

C:\Users\Administrator>ver

Microsoft Windows [Version 6.1.7601]

C:\Users\Administrator>perl6 -MFile::Find
I ran into a problem while trying to set up Linenoise: Must specify something as
 a path: did you mean '.' for the current directory?
Continuing without tab completions or line editor
You may want to consider using rlwrap for simple line editor functionality

To exit type 'exit' or '^D'
> ^D
===SORRY!=== Error while compiling:
Bogus statement
------> <BOL><HERE>♦
    expecting any of:
        prefix
        term
> ^D
===SORRY!=== Error while compiling:
Bogus statement
------> <BOL><HERE>♦
    expecting any of:
        prefix
        term
>
```

vrurg:
Closing via #3230

PR #3230, opened 2019-10-13
Fix REPL exit instructions on Windows

### LINKED ISSUE #1677: REPL exit instructions don't work on Windows
It says to press `^D` to exit but that don't do nothing on `cmd.exe` on Win10: 
```
C:\rakudo>perl6

To exit type 'exit' or '^D'
> ^D^D^D^D^D^D^D^D^D^D
===SORRY!=== Error while compiling:
Bogus statement
------> <BOL><HERE>
    expecting any of:
        prefix
        term
> ^D
===SORRY!=== Error while compiling:
Bogus statement
------> <BOL><HERE>
    expecting any of:
        prefix
        term
> ^CTerminate batch job (Y/N)? y

C:\rakudo>perl6 -v
This is Rakudo version 2018.03-107-ge31dd96 built on MoarVM version 2018.03-31-gfdb5e4d
implementing Perl 6.c.
```

(partly related https://github.com/rakudo/rakudo/pull/1674 )

hankache:
Fixes #1677 and #1969

### DIFF
--- a/src/core.c/REPL.pm6
+++ b/src/core.c/REPL.pm6
@@ -304,7 +304,11 @@ do {
 
         method repl-loop(*%adverbs) {
 
-            say "To exit type 'exit' or '^D'";
+            if $*DISTRO.is-win {
+                say "To exit type 'exit' or '^C'";
+            } else {
+                say "To exit type 'exit' or '^D'";
+            }
 
             my $prompt;
             my $code;
--- a/src/core.c/REPL.pm6
+++ b/src/core.c/REPL.pm6
@@ -305,7 +305,7 @@ do {
         method repl-loop(*%adverbs) {
 
             if $*DISTRO.is-win {
-                say "To exit type 'exit' or '^C'";
+                say "To exit type 'exit' or '^Z'";
             } else {
                 say "To exit type 'exit' or '^D'";
             }
--- a/t/02-rakudo/repl.t
+++ b/t/02-rakudo/repl.t
@@ -9,7 +9,7 @@ my $*REPL-SCRUBBER = -> $_ is copy {
     if $*DISTRO.is-win {
     s/^^ "You may want to `zef install Readline` or `zef install Linenoise`"
         " or use rlwrap for a line editor\n\n"//;
-    s/^^ "To exit type 'exit' or '^C'\n"//;
+    s/^^ "To exit type 'exit' or '^Z'\n"//;
     s:g/ ^^ "> "  //; # Strip out the prompts
     s:g/    ">" $ //; # Strip out the final prompt
     s:g/ ^^ "* "+ //; # Strip out the continuation-prompts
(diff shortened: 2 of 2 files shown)


src/core.c/REPL.pm6
```diff
@@ -304,7 +304,11 @@ do {
 
         method repl-loop(*%adverbs) {
 
-            say "To exit type 'exit' or '^D'";
+            if $*DISTRO.is-win {
+                say "To exit type 'exit' or '^C'";
```
vrurg: Unix `^D` (EOF) is `^Z` on DOS-derivatives, isn't it?

t/02-rakudo/repl.t
```diff
@@ -6,13 +6,23 @@ use Test::Helpers;
...
 my $*REPL-SCRUBBER = -> $_ is copy {
+    if $*DISTRO.is-win {
+    s/^^ "You may want to `zef install Readline` or `zef install Linenoise`"
+        " or use rlwrap for a line editor\n\n"//;
+    s/^^ "To exit type 'exit' or '^C'\n"//;
+    s:g/ ^^ "> "  //; # Strip out the prompts
+    s:g/    ">" $ //; # Strip out the final prompt
+    s:g/ ^^ "* "+ //; # Strip out the continuation-prompts
+    $_
+    } else {
```
vrurg: Same `^D` -> `^Z` note. But more important: the whole block is duplicate for the only purpose of changing one symbol in a message.

src/core.c/REPL.pm6
hankache: Indeed, but from a practical point of view, `^Z` requires the ENTER key to be hit afterwards on Windows, while `^C` will exit the REPL directly.
What do you propose I should move forward with? `^Z` or `^C`

vrurg: Hm... Luckily I'm not a Win user. All those inconsistencies are so confusing.

`^Z` would still be preferable because it's a request for graceful shutdown. Perhaps the message must reflect that nuance and contain `^Z Enter`?

hankache: Done updated to ^Z

t/02-rakudo/repl.t
hankache: Updated to `^Z`.
I tried using `if else` and `?? !!` within the block in order not to duplicate but tests kept failing.

vrurg: Try wrapping just that s/// line which contains the message. Or better use [`$(code)` construct](https://docs.perl6.org/language/regexes#Regex_interpolation)

hankache: I just made the change. Is it ok now?

hankache:
@vrurg is it merge-able?

vrurg:
@hankache I would like to get rid of redundant duplication of printing 'To exit ...' message. Otherwise no problem.

hankache:
@vrurg updated.

vrurg:
Thanks!

PR #3231, opened 2019-10-14
Detect non-native literal passed in when rw container is expected

usev6:


lizmat:
Thanks!  Are there any tests planned?

usev6:
@lizmat , sorry, I've missed the merge (thanks) and your question. I've unfudged some now passing tests for the JVM backend (https://github.com/perl6/roast/commit/3f2cb17c5b) and I've opened a PR for adding two tests for Rakudo: https://github.com/rakudo/rakudo/pull/3304

Apart from that, I don't have a good idea for additional tests.

PR #3236, opened 2019-10-15
Don't put CompUnitRepo folders outside the build dir during build

patzim:
CompUnitRepo generates a folder structure in the PERL6_HOME folder. That
folder does not exist yet during the build. So the normal logic to
auto determine the PERL6_HOME folder thinks it's `build_dir/../share/perl6`.
Thus CompUnitRepo puts some folders there. This is not wanted.
As a temporary solution to prevent pollution of the users directories, we
just explicitly set PERL6_HOME to `build_dir/gen/dummy_p6_home` then
CompUnitRepo can litter it's folders there and all is well.
_edit: Corrected the dir it currently creates the folders in._


vrurg:
Would be slightly more reliable if `PERL6_HOME` is only set if not preset by the user.

patzim:
I'm not sure. The build-perl6 binary is not making use of the PERL6_HOME directory. All the files that would be located there are somewhere else and hard coded in the build-perl6 binary. I'd say if we'd allow a PERL6_HOME env var to affect the build perl6 executable it'd be undefined behavior. So better just ignore it during build.

patzim:
Just noticed this PR ignores Windows. That's still to be done.

patzim:
That last force push 9 days ago added the respective Windows implementation.

AlexDaniel:
So, TL;DR build process writes outside the build dir, which it shouldn't, and this PR fixes the issue.

IMO it's important to fix it, though I don't like how it's fixed by writing into environ. That said:

> \<vrurg\> AlexDaniel: it's not a good way. But currently it's the only way to do it without moving the whole build destination into blib.

So I guess it works for now!

Thank you!

Issue #2545, opened 2018-12-14
LTA error message when declaring a PrecompStore with default values


JJ:
## The Problem

This program

```perl6
use nqp;

my $io=$*PROGRAM-NAME.IO;
my $precomp-store = CompUnit::PrecompilationStore::File.new();
my $precomp = CompUnit::PrecompilationRepository::Default.new(store => $precomp-store);

my $id = nqp::sha1(~$*PROGRAM-NAME);
my $handle = $precomp.load($id)[0];

without $handle {
    $precomp.precompile($io, $id, :force);
    $handle = $precomp.load($id)[0] // fail("Could not precompile");
}

say nqp::atkey($handle.unit,'$=pod')[0];
```

Fails with error: 
```
Invocant of method 'add' must be an object instance of type 'IO::Path', not a type object of type 'IO::Path'.  Did you forget a '.new'?
```

## Expected Behavior

Since this is due to the fact that we are not initializing correctly the named argument `prefix`, which is getting the default value `IO::Path`, it would be more awesome if either a reasonable default would be assigned to that argument or an error was thrown at that particular point.

## Actual Behavior

No object of that type is actually declared in the script, and it's obscure why that error is being thrown in line 

```
my $handle = $precomp.load($id)[0];
```

## Steps to Reproduce

Run [the script above](https://github.com/JJ/my-perl6-examples/blob/master/precomp.p6). 

## Environment
* Operating system: 
* Compiler version (`perl6 -v`): 

Perl6 2018.11

PR #3245, opened 2019-10-18
Test rakudo-specific warning for shape specifier

Altai-man:
A simple test that runs code and checks if a warning is issued.
Closes https://github.com/rakudo/rakudo/issues/2545


lizmat:
Thanks!

PR #3246, opened 2019-10-18
Add support for .rakumod extension

AlexDaniel:


AlexDaniel:
Do we want this to go in early or not? People seem to be very impatient, maybe we should start supporting the new extensions unofficially?

AlexDaniel:
Also, what would be the best way to expose these and should we do it? For example, [they appear in zef](https://github.com/ugexe/zef/blob/e73897378b36cbf53f2ed8d98363d94bac70767d/lib/Zef/CLI.pm6#L406).

jnthn:
> Do we want this to go in early or not? 

I'd say so. I think the general approach is:

* Start the work on making things prepared for them
* Discourage their use for now

So that at the point we start to recommend their use, then there's:

* A decent chance of most folks running Rakudo versions that supports them 
* Some chance of one's editor supporting them

AlexDaniel:
My biggest worry is that people will start changing extensions in their *modules* which will make them unusable for a lot of people who rely on distro packages and such. :S

jnthn:
Indeed. Alas, the distro packages are precisely what I have in mind as the reason for getting it into Rakudo sooner rather than later. We can't exactly control what Raku users will do, but we can control what's in the docs, what we suggest folks use, etc. Whereas when distros package stuff is much less out of our control.

If we're interesting in "creative" solutions - though I'm a bit uneasy about this - we could ship support for extension, but make it do a date check. So it'd be in, say, a Rakudo 2019.10 release, but won't actually start working until sometime in 2020.

vrurg:
A time bomb is somewhat over-creative. Why not simply block it with a conditional which can be removed when we're ready? For testing purposes the condition could be overruled with an env variable like `RAKUDO_RAKU_EXTENSIONS`.

jnthn:
> A time bomb is somewhat over-creative.

Well, it's meant to enable something rather than disable, but yeah, I'm not keen on it. Have a hard time saying *why*, just an uneasy feeling. Probably mostly fear of screwing it up and being the laughing stock of the day. :-P

> Why not simply block it with a conditional which can be removed when we're ready?

Because that doesn't help with getting it into distributions while avoiding it being used until that's happened.

vrurg:
Distributions tend to lag behind anyway. That's couldn't be the reason why we can't just have a release with full extension support.

But what we can think about is to teach zef rejecting installation of non-Perl6 modules. The point is  META6.json has a key `perl` which is gonna be replaced with `raku`. I don't remember at the moment if the key is mandatory, but it must be required. If a module is switching to new extensions it means full switch to Raku and introduction of `raku` key in META6. This would make it possible to prevent installation of modules with unsupported extensions.

Same should work the other way around by preventing installation of old modules when support for Perl6 extensions is dropped.

What I worry about, due to lack of knowledge about how CPAN works, wether it'd be possible to prefetch module's META6 prior to module itself in order to perform compatibility check with minimal overhead.

Hopefully, ecosystem would not be that hard to implement support for zef. Looking at META.list I would say it most definitely needs improvements, but that'd be different subject anyway. For now introduction of RAKU.list or RAKU-META.list would be sufficient as the starting point.

tony-o:
Deprecating the use of _old_ extensions with a warning over several version seems prudent and to have worked out well with other languages.

niner:
Let's not overthink this. Supporting older rakudo versions is already an active act for module owners as they already need to decide how much lost compatibility that new feature is worth to them.

That said, as this patch shows, the only place in rakudo that even cares about the module file extension is the FileSystem repository. This is only relevant for module development. Installation (and thus zef) is completely agnostic as it will just use the files that are listed in the META6.json's `provides` section. And even the FileSystem repository will load whatever's in META6.json if there is one:

```
nine@sphinx:~/Inline-Perl5 (master=)> perl -p -i -e 's!lib/Inline/Perl5.pm6!lib/Inline/Perl5.rakumod!' META6.json 
nine@sphinx:~/Inline-Perl5 (master *=)> mv lib/Inline/Perl5.pm6 lib/Inline/Perl5.rakumod
nine@sphinx:~/Inline-Perl5 (master *=)> prove -e 'perl6 -I.'
t/argv.t ...................... ok   
t/autoload.t .................. ok
```

This is on an unpatched rakudo.

jonathanstowe:
I'm with @tony-o . I have a **lot** of modules in the ecosystem and on a practical level I'd rather have to do fewer releases of each module to effect the consequences of the rename.  A transitional period of supporting both silently ⇒ deprecation warning on _old_ names ⇒ remove support for _old_ names would probably help but at the cost of more work in core and tooling.

AlexDaniel:
I was aware of this, but having it written down really helped me make up my mind regarding this PR:

> This is only relevant for module development. Installation (and thus zef) is completely agnostic as it will just use the files that are listed in the META6.json's provides section. And even the FileSystem repository will load whatever's in META6.json if there is one

@niner++

I think this can go in early then!

Issue #2385, opened 2018-10-17
LTA role multi method stubs enforce arguments but not return value


zoffixznet:
```
19:09 | ryn1x | m: role Foo { multi method bar(Str --> Str) { … } }; class Bar does Foo { multi method bar(Str $a --> Int) { return 1} }
19:09 | camelia | rakudo-moar 4bf55b1ef: ( no output )
```

(related ticket: https://github.com/rakudo/rakudo/issues/2384 )

FCO:
It looks that it happens because it uses `eqv` to compare both signatures, but `eqv` do not compare the return.

http://colabti.org/irclogger/irclogger_log/perl6?date=2018-10-20#l332

FCO:
Here is were the signatures are being tested: https://github.com/rakudo/rakudo/blob/de4e82259be36ac9c97b56d594591ea796a1b131/src/core/Signature.pm6#L207

FCO:
Should `eqv` use the return to compare signatures or shouldn’t the comparison of multi stub with multi method use `eqv`?

zoffixznet:
> Should eqv use the return to compare signatures or shouldn’t the comparison of multi stub with multi method use eqv?

Not an expert on this, but to me it'd make sense for `eqv` to check return as well.

FCO:
Fixed, test needed

PR #3249, opened 2019-10-19
Test return constraints for role implementors

Altai-man:
Closes https://github.com/rakudo/rakudo/issues/2385

The failed tests are all different and appear to be flappers to me...


AlexDaniel:
After a discussion on IRC, I think testing the error message is fine in rakudo tests but we also need a roast test too, somewhere around here: https://github.com/perl6/roast/blob/93cc2e4274df21e9bb7ff485acfc5279305ba87c/S14-roles/composition.t#L179 (we already have similar tests that don't consider return types, and that should be fixed).

Altai-man:
Was writing a test and now I am confused:
```
> role R5 { method m-ret(--> Int) { ... } }
(R5)
> class C11 does R5 { method m-ret() { 99 } }
(C11)
> C11.m-ret;
99
> 
```
should not we throw an error here?

FCO:
We should throw an error if 2 roles implement the same method, but the way to fix it is to write it on the class if I remember it well

JJ:
> Was writing a test and now I am confused:
> 
> ```
> > role R5 { method m-ret(--> Int) { ... } }
> (R5)
> > class C11 does R5 { method m-ret() { 99 } }
> (C11)
> > C11.m-ret;
> 99
> > 
> ```
> 
> should not we throw an error here?

But this is a different issue, right?

Also, can you check the CI errors?

lizmat:
Awaiting CI results now

PR #3251, opened 2019-10-20
.samemark: Support emoji, flags and Prepend

samcv:
`.samemark` is currently broken. it assumes the first codepoint of a grapheme is the "base", and *anything* occuring afterward is a "mark".
(below the `*` is used in the same meaning as regex, and things in brackets are single codepoint tokens)

**How samemark assumes the world exists**
`[base] [mark]*`

**The real world it is actually more like**
`[mark]*[base][mark]*`

**But** then you have emoji, flags etc.
`[Regional Indicator][Regional Indicator]`
**or**
`[Base Emoji]([ZWJ][Emoji])*`

What I think is the most reasonable solution is to treat `Grapheme_Cluster_Break=Extend` and `Grapheme_Cluster_Break=Prepend` as “mark”’s. This would cause it to treat flags or emoji sequences as the "base", since it's not really separable the same way accent marks are separable.

This does not address the fact that "\c[Canada]".samemark("é") would give you a Canadian flag with an accent mark on it, and why anybody would desire this. As this is how samemark has always worked it doesn't make sense to change this.

Fixes perl6/problem-solving#61 and tests are ready in https://github.com/perl6/roast/pull/589

### DIFF
--- a/src/core.c/Str.pm6
+++ b/src/core.c/Str.pm6
@@ -2054,6 +2054,44 @@ my class Str does Stringy { # declared in BOOTSTRAP
     }
 
 #?if !jvm
+    sub get-parts ($nfd is raw) {
+      my int $p = 0;
+      my int $last-prepend = -1;
+      my int $first-extend = -1;
+      while $p < nqp::elems($nfd) &&
+        nqp::iseq_s('Prepend', nqp::getuniprop_str(
+          nqp::atpos_i($nfd, $p),
+          nqp::unipropcode('gcb'))) {
+        $last-prepend = $p;
+        $p = nqp::add_i($p, 1)
+      }
+      $p = nqp::elems($nfd) - 1;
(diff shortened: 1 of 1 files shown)


AlexDaniel:
@samcv samemark no longer works.
```
<AlexDaniel> c: 2019.07.1 say "hõllo world".samemark("aaaaäa")
<committable6> AlexDaniel, ¦2019.07.1: «hollö world␤»
<AlexDaniel> c: HEAD say "hõllo world".samemark("aaaaäa")
<committable6> AlexDaniel, ¦HEAD(4b2a45f): «This type (List) does not support positional operations␤  in block <unit> at /tmp/_qk5VRWnq9 line 1␤␤ «exit code = 1»»
```

vrurg:
And spectests failing. Can we revert this for the time being?

samcv:
It has been reverted.

PR #3252, opened 2019-10-20
Fix test count failure on windows

ugexe:


AlexDaniel:
Hm, why does that work? Looking at `compile_test_lib`, it dies on unknown platforms. I assume the patch helps, but I don't understand how switching `done-testing` can do it.

AlexDaniel:
Ah, I see it. It's just the `unless $*DISTRO.is-win` bit. I think we should do `skip` instead of switching to done-testing (at least that's what I've seen people do before).

AlexDaniel:
That said, `grep` disagrees with me :)
```
01-sanity/98-test-deprecated.t
18:done-testing;

01-sanity/99-test-basic.t
145:done-testing;

02-rakudo/13-exceptions.t
12:done-testing;

02-rakudo/reproducible-builds.t
44:done-testing;

02-rakudo/14-revisions.t
72:done-testing;

03-jvm/01-interop.t
4:BEGIN unless $*VM ~~ "jvm" { plan 0; skip-rest "jvm only test"; done-testing; exit 0; };

04-nativecall/18-routine-sig-sanity.t
127:done-testing

09-moar/General_Category__UnicodeData__2.t
12:    done-testing;

09-moar/General_Category__extracted-DerivedGeneralCategory.t
12:    done-testing;

09-moar/Line_Break__LineBreak.t
12:    done-testing;

09-moar/NAME__UnicodeData.t
12:    done-testing;
```

PR #3254, opened 2019-10-21
Fix a wrong reference

crnlskn:
otherwise we end up with an empty pattern in a split call later which breaks stuff.

To expand on that commit message, `$(NQP_HOME)` doesn't seem to be defined in that spot in the Makefile, which results in an empty string being passed to `create-jvm-runner.pl`, which in turn results in an empty string being used as the pattern in a perl `split` call, which resulted in the classpath in *installed* `perl6-j` to be a lot of letters separated by `:`, which doesn't work.


vrurg:
One more typo... Thank you!

PR #3261, opened 2019-10-23
[WIP] Die early on a duplicate role

lizmat:
When you apply the same role to a class more than once, you wind up with
very confusing error about needing to resolve your methods because they
are being provided by more than one role.  This patch adds a check (by
name) to see whether the same role has been specified more than once.

This changes the error for:

    role A { method foo() { } }
    class B does A does A { }

from:

    Method 'foo' must be resolved by class B because it exists in multiple roles (A, A)

to:

    Can only apply role 'A' to class 'B' once

FWIW, tried to do the uniqueness check using objectid, but for some reason
identical roles still arrive in the method with different object id's, even
when deconted.

### DIFF
--- a/src/Perl6/Metamodel/RoleToClassApplier.nqp
+++ b/src/Perl6/Metamodel/RoleToClassApplier.nqp
@@ -42,6 +42,23 @@ my class RoleToClassApplier {
     }
 
     method apply($target, @roles) {
+
+        # Make sure the same role isn't added more than once, so that we don't
+        # get disambiguation messages on identical methods later, which can be
+        # really confusing.
+        my %seen;
+        for @roles {
+            my $name := $_.HOW.name($_);
+            if nqp::existskey(%seen,$name) {
+                nqp::die(
+                  "Can only apply role '$name' to class '"
+                  ~ $target.HOW.name($target)
+                  ~ "' once"
+                );
+            }
+            nqp::bindkey(%seen,$name,1)
+        }
+
         # If we have many things to compose, then get them into a single helper
         # role first.
         my $to_compose;
--- a/src/Perl6/Metamodel/RoleToClassApplier.nqp
+++ b/src/Perl6/Metamodel/RoleToClassApplier.nqp
@@ -43,20 +43,26 @@ my class RoleToClassApplier {
 
     method apply($target, @roles) {
 
-        # Make sure the same role isn't added more than once, so that we don't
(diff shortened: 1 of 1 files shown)


vrurg:
May I ask you to merge this after #3199? I'm only awaiting for release and for @jnthn approval for `.?+`, `.?-` operators.

lizmat:
Hmmm... this appears to break the case of:

    my role R[::T] { multi method foo(T $t) { T.gist } }
    my class A does R[Str] does R[Int] { }

Which I guess implies we need to check if the role is parameterizable?

lizmat:
I have no intention of merging this before the 2019.10 release.

lizmat:
Perhaps I should move the check to the actual merging, where the error used to be found.

vrurg:
Just to summarize: the only correct way is to check for concretizations. Prior to this you have curryings which holds non-concretized parameters. I.e. `R[::T]` is a curry, R[Int] is its concretization.

It just crossed my mind that with #3199 re-application of the role might be just ignored because I took care of multi-path  role consumption when `R0` is consumed by `R1` and `R2` and then class consumes `R1` and `R2`. Let me check it out...

vrurg:
Good, as I expected:

```
role R { method foo { say "foo" } }
class C does R does R { }; 
C.new.foo # foo
```

The point is that whatever is the real structure of "inheritance", role application is suppose to flatten down the list. Because, as I stated above, it is possible that same role could in this case be applied "twice" the final behavior should be as if it only appears once:

```
role R0 { }
role R1 does R0 { }
role R2 does R0 { }
class C does R1 does R2 { }
```

is equivalent to:

```
class C does R1 does R0 does R2 does R0 { }
```

but has to be interpreted as:

```
class C does R1 does R0 does R2 { }
```

This actually makes the bug in question a reduced case of the above. Therefore, double application of a role must not be a bug. Though it could be a warning.

lizmat:
Alas, this doesn't fix this case:

    role A[::T] { method foo() { } }; class B does A[Int] does A[Str] { }

vrurg:
In the last case A[Int] and A[Str] are different roles. Correspondingly, the error message about duplicate method is correct. The message itself could be made more informative by providing full role names with parameters. I.e. instead of:

```
Method 'foo' must be resolved by class C because it exists in multiple roles (A, A)
```

I'd rather expect to see:

```
Method 'foo' must be resolved by class C because it exists in multiple roles (A[Int], A[Str])
```

PR #3267, opened 2019-10-24
Update CREDITS

peschwa:
due to personal reasons I'd like this changed


AlexDaniel:
Thanks for updating the information!

PR #3269, opened 2019-10-25
[WIP] Ensure CArray and Pointer can be parameterized with non-Any types

### LINKED ISSUE #3264: Self-referential struct does not work with CArray[Pointer[]] declaration.
## The Problem
When attempting to define a self-referential structure from C, I ran into the following error:

```
    ===SORRY!===
    Cannot resolve caller ACCEPTS(Bool:U: GIO::DBus::Raw::Types::GDBusAnnotationInfo); none of these signatures match:
        (Bool:D: Mu \topic, *%_)
        (Bool:U: \topic, *%_)
Stage start      :   0.000
```

This stems from the fact that it appears that CArray[] and Pointer[] do not see the self-referential nature of the declaration. Here is a small sample that exposes this issue:

```
use NativeCall; 
class A is repr<CStruct> { 
  has uint32 $.a; 
  has CArray[Pointer[A]] $.aa; 
};
```

Note that this _does_ work, possibly due to special case handling:

```
class A is repr<CStruct> { 
  has uint32 $.a; 
  has A $.aa; 
};
```

## Environment
* Operating system: 

Kaiepi:
Fixes #3264

This hasn't been fully tested yet. @Xliff

### DIFF
--- a/lib/NativeCall/Types.pm6
+++ b/lib/NativeCall/Types.pm6
@@ -6,16 +6,18 @@ sub nativecast($target-type, $source) {
         nqp::decont(map_return_type($target-type)), nqp::decont($source));
 }
 
-our native long     is Int is ctype("long")     is repr("P6int") { };
-our native longlong is Int is ctype("longlong") is repr("P6int") { };
+our native long      is Int is ctype("long")                 is repr("P6int") { };
+our native longlong  is Int is ctype("longlong")             is repr("P6int") { };
 our native ulong     is Int is ctype("long")     is unsigned is repr("P6int") { };
 our native ulonglong is Int is ctype("longlong") is unsigned is repr("P6int") { };
 our native size_t    is Int is ctype("size_t")   is unsigned is repr("P6int") { };
 our native ssize_t   is Int is ctype("size_t")               is repr("P6int") { };
 our native bool      is Int is ctype("bool")                 is repr("P6int") { };
-our class void                                  is repr('Uninstantiable') { };
+
+our class void is repr('Uninstantiable') { };
+
 # Expose a Pointer class for working with raw pointers.
-our class Pointer                               is repr('CPointer') {
+our class Pointer is repr('CPointer') {
     method of() { void }
 
     multi method new() {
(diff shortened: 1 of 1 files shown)


lib/NativeCall/Types.pm6
```diff
@@ -76,22 +78,24 @@ our class Pointer                               is repr('CPointer') {
...
-        die "A typed pointer can only hold:\n" ~
-            "  (u)int8, (u)int16, (u)int32, (u)int64, (u)long, (u)longlong, num16, num32, (s)size_t, bool, Str\n" ~
-            "  and types with representation: CArray, CPointer, CStruct, CPPStruct and CUnion" ~
-            "not: {t.^name}"
-            unless t ~~ Int|Num|Bool || t === Str|void || t.REPR eq any <CStruct CUnion CPPStruct CPointer CArray>;
-        my $w := p.^mixin: TypedPointer[t.WHAT];
-        $w.^set_name: "{p.^name}[{t.^name}]";
-        $w;
+    method ^parameterize(Mu:U $this is raw, Mu:U $type is raw) {
+        when $type ~~ any(Int, Num, Str, void)
```
MasterDuke17: This isn't related to the problem you're fixing, but why the `when` with only one condition and then a `default`? That construct is a good bit slower than just an `if` and then an `else`.

Kaiepi: I suspect the type check here and in `CArray.^parameterize` might make this fail some tests

Kaiepi: This used to be inside a `given` with a few different `when`s, but I forgot to change it back after removing it.

Xliff: I think @MasterDuke17 's comments should be well-received. I'm already experiencing significant slow downs with the respect of compilation. I think I can wait a bit until @Kaiepi can rewrite this using if/elsif/else rather than when to improve performance.

Thanks much @Kaiepi and @MasterDuke17 !

lizmat:
Is this still valid after new-disp has landed and @niner has done a lot of work on the NativeCall interface?

Xliff:
I'll look into this, today.

PR #3272, opened 2019-10-30
Unify dynamic lookups performed by pseudo-packages and &DYNAMIC

vrurg:
So far, pseudo-package did not take into account `GLOBAL` and `PROCESS` packages. This PR teaches them this and a couple of other tricks.

- `&DYNAMIC` is now capable of searching relative to an arbitrary context, not only it's current one.
- Because not only `CALLER` pseudo-package can now throw `NotDynamic` kind of exception, more generic `X::SymbolNotDynamic` one is introduced which takes a package name parameter. `X::Caller::NotDynamic` is now its child and preserved for backward compatibility with `6.c`.
- `PROCESS` and `GLOBAL` now raise `dynamic` flag on containers created via key-assignment. I.e.:

   ```
   PROCESS::<$foo> = 1;
   $*foo.VAR.dynamic; # True
   ```

- [6.e] Assignment to a non-existing symbol via a dynamic pseudo-package fallbacks to `PROCESS` assignment:

    ```
    DYNAMIC::<$*foo>:exists; # False
    DYNAMIC::<$foo> = 42;
    PROCESS::<$foo>:exists; # True
    ```

- [6.e] Support for `Promise`-wrapped code, i.e. make symbols declared outside of `start`ed code available via a dynamic pseudo.

   _NOTE 1_ `6.c` provides only partial support due to backward compatibility matters.


vrurg:
@jnthn I wouldn't be a big deal to remove it. But if I have a chance to convince you – I'd like to use it. Though to use the chance I would like to know why do you object it?

jnthn:
From a language design perspective, it requires making `GLOBAL` and `PROCESS` special cases, and introduces an entire new and user-visible `DynamicStash` type, which is two new things the language user might need to know about (and so more things to document). From an implementation perspective, it looks like as much extra complexity as I suspected, and doesn't seem to have much benefit. If the overall point here is to unify dynamic lookup, then surely handling the `GLOBAL` and `PROCESS` fallback mechanism in one place is simpler than all the additions this approach requires?

Further, the way it's done here seems to introduce a discrepancy between assigning and binding into those packages, and resolving that creates action at a distance; binding into a stash modifying the value being found feels very wrong.

vrurg:
My points:

1. `DynamicStash` is an attempt to prevent the overhead (though minor) of addition of extra method on Stash. I.e. I just wanted only `PROCESS` and `GLOBAL` stashes to have `ASSIGN-KEY`. Alternative approach of adding the method on `Stash` is simply raising an additional `$!is-dynamic` flag on it works well for my purpose. I can revert. (Heh, it was fun playing with `DynamicStash` anyway... :) )

2. It is not only about unifying lookups but also about unifying user experience. As a user I would expect `dynamic` flag set on a `*`-twigilled container. Otherwise what's the purpose of the attribute if it's not consistent? I mean, wether `dynamic` is _True_ or _False_ – it tells me almost nothing about the container itself.

   This is perhaps of my most concerns about this matter.

3. Assigning and binding are different. Considering that an assignment is ok to re-containerize a value, it is also ok that it would implicitly set `dynamic`. On the contrary, binding as 100% _as-is_ operation. Thus

    ```
    PROCESS::<$*foo> = $bar;
    PROCESS::<$*foo> := $bar;
    ```

   are different operations and it's clear that `$bar` container would be preserved and remain unmodified. Similarly, binding of a plain value is a synonym of creating a constant:

    ```
    PROCESS::<$*foo> := 42;
    $*foo.VAR.^name; # Int
    ```

   So, just basic understanding of difference between binding and assigning must be enough to resolve any possible confusions.

Moreover, due to the role of `GLOBAL` in handling of `our`-scoped things, the following:

```
our $foo = 42;
say $*foo.VAR.dynamic; # False
```

also makes sense to me as in this case `dynamic` clearly reflects the initial intention of a variable's purpose.

So, if the complexity of implementation is your most concern – I can easily fall back to much simpler path. Though the current implementation has been done to prevent any performance impact on stashes at all, the other way introduces very little impact anyway.

If nothing of the above convincing you, I will pull out this feature. But then it would raise a question of overall usability of `dynamic` because, AFAIK, `is dynamic` trait isn't much used and not really supported.

vrurg:
@jnthn How about this implementation?

PR #3277, opened 2019-11-04
Create FUNDING.yml

ugexe:
Enable github sponsorship users to be listed.


AlexDaniel:
Let's see how this goes! I have enabled the Sponsorships feature in the settings.

For those who don't know, it's a new github feature that was previously in beta. I see no reason not to try it out.

PR #3282, opened 2019-11-06
Hint about spectest with Inline::Perl5

patzim:
This PR adds a note to the harness5/6 scripts adding a hint on how to install Inline::Perl5 to also run the respective tests. Also does a little bit of minor slightly related cleanup.


niner:
Looks good!

PR #3284, opened 2019-11-07
Fix Stash'es clashing after deserialization

vrurg:
When two modules are precompiled independently but then loaded by the
same process it is possible that if they share common name prefix like
in, say, `A::B::C::D1` and `A::B::C::D2` then the module loaded second
would get its Stash on `A` overriden by the module loaded first. This
would effectively result in some symbols in `A::B::...` hierarchy to be
lost to the code importing the modules. This behavior is caused by the
fact that when second module is being loaded by MoarVM from its
precompilation, symbols from parent stashes are being resolved to the
existing type objects from the first module. I.e., if `D1` is loaded
first, then upon loading `D2` it's `A` package would be resolved to the
`A` implicitly created earlier for `D1`. And that package would bear a
`Stash` object which contains `D1` but knows nothing about `D2`.

The problem is being fixed by preserving a `Stash` structure in a form
which wouldn't lost child objects after MoarVM resolves package symbols.
Currently it is done by creating a nested list of items containing
symbol name, symbol value and, optionally, subtable built from symbol's
`Stash` if it's a package. The structure is then stored in `Stash`
attribute `@!snapshot` before serialization. Then symbols from it are
merged back into the `Stash` object after deserialization.

It should also be noted that because `Stash` is Raku's level entity, it
makes sense to take care of it in Raku. Thus the solution of `Stash`
taking care of itself.

Fixes rakudo/rakudo#3075

### DIFF
--- a/src/Perl6/Actions.nqp
+++ b/src/Perl6/Actions.nqp
@@ -1292,7 +1292,7 @@ class Perl6::Actions is HLL::Actions does STDActions {
         $*W.assert_stubs_defined($/);
         $*W.sort_protos();
 
-        if $*W.is_precompilation_mode {
+        if !$*COMPILING_CORE_SETTING && $*W.is_precompilation_mode {
             # Perform GLOBALish stash fixup as early as possible but after repossession conflicts are resolved. The
             # purpose is to restore correct stash in cases when two package objects with compound names share common
             # prefix. Like, for example, A::B::C::D1 and A::B::C::D2 share A::B::C. In this case sometimes when defined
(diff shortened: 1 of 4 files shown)


vrurg:
This PR won't be accompanied by tests because I didn't manage to golf it down to any compact form. Even though I know what happens I failed to determine how exactly it happens.

lizmat:
Perhaps this needs another review?  Or should be closed?

vrurg:
Closing. This PR is a workaround, not a real fix, after all. A real one would require some changes in precompilation logic.

PR #3286, opened 2019-11-07
Match subst

lizmat:
Add Match.subst($replacement) for easier replacing
    
Basically replace the part that matched by the replacement string and return that.  This basically allows the result of a match to be used in a condition, and then do a replacement and possibly other stuff without having to do the match again:
    
        if "foo".match( / o / ) -> $match {
            say $match.subst("x");   # fxo
            # do other stuff if we found a match
        }

### DIFF
--- a/src/core.c/Cool.pm6
+++ b/src/core.c/Cool.pm6
@@ -322,7 +322,7 @@ my class Cool { # declared in BOOTSTRAP
         $/ := nqp::getlexcaller('$/');
         {*}
     }
-    multi method subst(Cool:D: $original, $replacement, *%options) {
+    multi method subst(Cool:D: $original, $replacement = "", *%options) {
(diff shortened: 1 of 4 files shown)


vrurg:
I think this breaks URI module. Haven't bisected but it works on the release but fails on master. Considering the error it reports:

```
Cannot modify an immutable Perl6::Grammar (Perl6::Grammar)
at /Users/vrurg/src/Raku/uri/lib/URI.pm (URI):33
```

this PR is the likely candidate.

lizmat:
Will check

lizmat:
No, it actually is today's `BUILDPLAN` work that is causing this issue.

Altai-man:
Can we delete the branch?

lizmat:
Yes.  :-)

PR #3287, opened 2019-11-07
Update instructions and fix md layout

nxadm:


AlexDaniel:
Why `$` symbols?

nxadm:
So it looks like a shell prompt.

vrurg:
Sometimes docs are used by people to copy-paste commands. Even few commands at once.

nxadm:
> Sometimes docs are used by people to copy-paste commands. Even few commands at once.

Fair enough. Remove "$" and also this command because the directory is no longer present in recent Rakudos:

"cp -r tools/build/binary-release/Linux/* install"

patzim:
@nxadm The folder is right here: [https://github.com/rakudo/rakudo/tree/master/tools/build/binary-release/Linux](https://github.com/rakudo/rakudo/tree/master/tools/build/binary-release/Linux)
I suspect you think these instructions and the referenced folders have been there a long time and bitrotted by now. That's not the case. I added them very recently (that's why they don't show up in any releases).

nxadm:
@patzim, ok, the the line needs to be added. I guess these tools are of no interest for regular users?

patzim:
@nxadm: They are of interest. The respective [README.md](https://github.com/rakudo/rakudo/blob/master/tools/build/binary-release/Linux/README.md) explains their usage. The `set-env.sh` currently just sets up the $PATH variable to include the `bin` and `share/perl6/site/bin` directories. So users can just open up a console, call that script and have a working raku environment in that console.

On Windows there is an additional script ([vs_build_tools_install_assistant.ps1](https://github.com/rakudo/rakudo/blob/master/tools/build/binary-release/Windows/bin/vs_build_tools_install_assistant.ps1)). That's an assistant that downloads and guides the user through the installation and usage of the build environment that is necessary for installing modules that compile C code.

patrickbkr:
I have created and merged [PR#3848](https://github.com/rakudo/rakudo/pull/3848) that pulls in the changes of this PR adapted to latest master. So this PR can be closed now.
@nxadm Thanks for your work with this!

PR #3288, opened 2019-11-07
Some more s/Perl 6/Raku changes

lizmat:
- $*RAKU now says "Raku" instead of "Perl", but $*PERL still says "Perl"
- Change various text references from "Perl 6" to "Raku"
- Change verious text references from "Perl 5" to "Perl"


lizmat:
This has basically been done.

PR #3289, opened 2019-11-07
Allow UInt to return a failure from Int

mienaikage:
Int can give a failure for e.g. `X::Str::Numeric`, which results in an exception when used with `< 0`. This is to check that `$got` actually contains an Int before checking that it is not negative.

### DIFF
--- a/src/core.c/Cool.pm6
+++ b/src/core.c/Cool.pm6
@@ -370,7 +370,7 @@ my class Cool { # declared in BOOTSTRAP
     proto method UInt(|) {*}
     multi method UInt()  {
         my $got := self.Int;
-        $got < 0
+        $got ~~ Int && $got < 0
           ?? Failure.new(X::OutOfRange.new(
                :what('Coercion to UInt'),
                :$got,


lizmat:
I'm still not sure I understand what problem this is fixing?  Could you provide some example code, that could be used in a test?

vrurg:
@lizmat I think it's the following case:

```
say "a".Int.WHAT; # Failure
say "a".UInt.WHAT; # Cannot convert string to number: base-10 number must begin with valid digits or '.' in '⏏aaa' (indicated by ⏏)
```

The patch bypasses the `Failure` to the caller transparently, as I'd expect it.

mienaikage:
Thank you @vrurg, that is indeed the issue being addressed.

lizmat:
@mienaikage: thank you for this Pull Request.  I've taken the liberty to address the problem in another way in https://github.com/rakudo/rakudo/commit/8f30cc76a5c8d204472b31b2447018c7179d7c04 .  Please do keep those PR's coming!

Related: it would be very nice if you could create a PR to test the fix!

PR #3291, opened 2019-11-08
Add test to check UInt passes on Failure from Int

mienaikage:
Following up on https://github.com/rakudo/rakudo/pull/3289#issuecomment-551980475

### DIFF
--- a/t/05-messages/03-errors.t
+++ b/t/05-messages/03-errors.t
@@ -2,7 +2,7 @@ use lib <t/packages/>;
 use Test;
 use Test::Helpers;
 
-plan 25;
+plan 26;
 
 subtest '.map does not explode in optimizer' => {
     plan 3;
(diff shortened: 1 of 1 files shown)


t/05-messages/03-errors.t
```diff
@@ -175,4 +175,8 @@ cmp-ok X::OutOfRange.new(
 is-run 'class { method z { $^a } }', :err{ my @lines = $^msg.lines; @lines.grep({ !/'⏏'/ && .contains: '$^a' }) }, :exitcode{.so},
 'Use placeholder variables in a method should yield a useful error message';
 
+# https://github.com/rakudo/rakudo/pull/3289
+fails-like { "a".UInt }, X::Str::Numeric,
+'UInt passes on failure given by Int';
```
vrurg: It's better for a spectest description not to rely on internal implementation. Something like _"UInt on a string produces a Failure"_, with an adjustment for my imprefect English perhaps, would be a bit more correct.

vrurg:
Also, this kind of tests belongs to (roast)[https://github.com/perl6/roast]. In particular, this one belongs to `S32-str/numeric.t`, alongside with a similar test for Int.

mienaikage:
I'm not too familiar with roast, but there seems to be a similar test for `Int` in [S32-num/int.t](https://github.com/perl6/roast/blob/48bd86159a3e1c8f7ca3f4cbf79e89aa1363cb3d/S32-num/int.t#L82). Would it be better to be alongside that one, or is `S32-str/numeric.t` more appropriate?

vrurg:
Though my opinion is it belongs to `S32-num`, but `S32-str/numeric.t` covers exactly our case here. For this reason, until roast is getting reorganized, lets put it into the latter.

PR #3300, opened 2019-11-14
Add any type information to positional in usage

lizmat:
Given a sub MAIN(Int $dial) in a script "frobnicate.p6", and calling it without
a single integer value, would give:

    Usage:
      frobnicator.p6 <dial>

It now gives:

    Usage:
      frobnicator.p6 <dial=Int>

which I hope better conveys that an Int was expected to be given.


ugexe:
I think thats going to confuse some people into doing `script dial=1`

MasterDuke17:
> I think thats going to confuse some people into doing `script dial=1`

I agree that seems like it might confuse people, but I do also really want the type there somehow.

Here are a couple other options:
```
Usage:
1. frobnicator.p6 <Int dial>
2. frobnicator.p6 <(Int) dial>
3. frobnicator.p6 <dial(=Int)>
4. frobnicator.p6 <dial=(Int)>
```
Of these, I'm sort of leaning toward 2 and 3. Anybody else have thoughts/suggestions?

lizmat:
Options 1, 2 and 4 make sense from a Raku point of view.  But the user of the script may well have no knowledge of Raku at all.

Option 3 optically looks to me like making it optional, which it isn't.  But I could very well live with that option.

FWIW, getting this sort of discussion was the reason I made it a PR.  Glad that is working out.

vrurg:
Not sure how this would look to a user, but this at least is unlikely to confuse with command line format:

```
frobnicator.p6 <dial, Int>
```

PR #3304, opened 2019-11-17
Add tests for catching error at compile time

usev6:
The second new test refers to a change introduced with
https://github.com/rakudo/rakudo/pull/3231 (to fix RT #129812
on the JVM backend). The first new test only serves to
show that the native and the non-native case behaves identically.


lizmat:
So, any objections against merging this?

vrurg:
@lizmat I would merge it.

PR #3310, opened 2019-11-21
Implement log2 function

tbrowder:


tbrowder:
Working on tests...

ugexe:
Is log(x) / log(2) deserving of its own function? What about every other number than 2?

vrurg:
@ugexe log2 does deserve to be explicitly included as a shortcut [because it a common one](https://en.wikipedia.org/wiki/Logarithm#Particular_bases).

tbrowder:
And, if I could get away with, I would also like to see 'ln', 'lb', and 'lg' added as built-ins.

tbrowder:
Note the build from this PR passed (Travis build #12333).

vrurg:
> And, if I could get away with, I would also like to see 'ln', 'lb', and 'lg' added as built-ins.

@tbrowder I totally forgot to reply to this... I don't think we shall go that far due to the risk of polluting the namespace. Say, `ln` could cause many to think it's about file system symlinking, etc. I'd rather welcome those provided by a 3rd party package where one knows what he does when imports it.

tbrowder:
@vrug, you are correct, I was not thinking of the big picture. Thanks for the dash of cold water!

Issue #3303, opened 2019-11-15
`raku` executable


AlexDaniel:
We need a `raku` executable so that people can use `raku` in their shebangs and stuff. `perl6` executable should be preserved so the two have to be linked together somehow. It's likely there will be different solutions for different platforms.

> \<nine\> AlexDaniel: on Windows just copy the executable. It's just a couple of kilobytes anyway

vrurg:
I think we can close this with 41ca7570b796e46cc1c6b77b2254b447a2867d36

AlexDaniel:
Thank you, @vrurg! Any reports from windows users? Can somebody verify that it works?

AlexDaniel:
To clarify, this issue is not blocking in any way, I'd just love to know that it does work on other platforms.

vrurg:
My Windows VM came back to its sense and managed to compile `moar`. So, I can now confirm successful builds on:

- FreeBSD 11.3
- OpenBSD 6.5
- Windows 7, MSVC toolchain

Spectests are passing except for Windows. But this is on `master` and most definitely unrelated to the build anyway.

MasterDuke17:
Can this be closed now?

vrurg:
It can.

PR #3312, opened 2019-11-22
Generate `raku` executable

vrurg:
Have both `perl6` and `raku` executables to allow smoother trasnition period.

Implementation of #3303


lizmat:
I'm afraid I didn't follow this as much as I should have, but it seems contrary to what was in the Raku PR:

    ### Executables

    As some packagers have already done, the executable should be called `rakudo`,
    since `rakudo` is the name of the implementation.  `raku` and `perl6` should
    be symlinks.  If at all technically possible, running a script using the
    `perl6` as the executor should provide a DEPRECATED warning at some point.

AlexDaniel:
Actually, I think Path to Raku PR is right, it should be `rakudo` and both `perl6` and `raku` should link to it. Which way does this PR implement it right now? Is it simply copying the runners?

vrurg:
Ah, crap... Of course it has to be `rakudo`! Thanks @lizmat! Not sure how much time will I have today, perhaps getting it right will take a bit longer unless it would considered ok to have only `raku` and `raku-debug`. The rest of `perl6-` runners will be available as `rakudo-`. Or otherwise we end up with floodwaters of executables in _bin/_.

@AlexDaniel textual runners are re-generated. This would allow possibly have different content on them if we ever need it. For example, I have simply removed mentioning of Perl 6 from their greeting leaving only Rakudo in place. But we could have the language name back in it if desirable. (I'm just not sure which is better and that wasn't high on the priority list).

Binaries are copied. I'm somewhat reluctant to symlinking due to its absense in Windows. With `rakudo` -> `raku` its use becomes inevitable anyway...

vrurg:
Unless I missed something, here we go with `raku` alias for `moar` backend. 

No JS, no JVM yet. And, please-please-please!, don't ask me to handle those! Let it be somebody who understands building of those backends better than me.

vrurg:
And one more thing. Windows is untested. No idea what's going on with my VM but moar build freezes at the linking stage. If there is anything wrong about Windows I hope appveyor would report it.

AlexDaniel:
@vrurg can you squash some of your commits for some cleaner history and also so that it's then easier to cherry-pick this into the release branch?

Issue #3306, opened 2019-11-19
"colon syntax" on $. (e.g. self) in a method does not work


chromis:
## The Problem

I asked a question about "colon syntax" on the perl6-users mailing list. See https://www.nntp.perl.org/group/perl.perl6.users/2019/11/msg7170.html

It seems that I've stumbled over a bug.

Consider the following snippet of code:

```
#!/usr/bin/env perl6

use v6.d;

class Foo
{
    has $.attribute;

    method say-times($times)
    {
        say $.attribute * $times
    }

    method say-default-works
    {
        $.say-times(2)
    }

    method say-default-works-too
    {
        self.say-times: 2
    }

    method say-default-does-not-parse
    {
        $.say-times: 2
    }
}

my $foo = Foo.new: attribute => 42;

$foo.say-times: 2;

$foo.say-default-works;

$foo.say-default-works-too;
```

## Expected Behavior

I would expect that this program parses and that the `say-default-does-not-parse` method should "just work" (and emit "84" on `stdout`)

## Actual Behavior

I get a parse error:

```
===SORRY!=== Error while compiling /home/chromisx/raku/./oo.pl6
Confused
at /home/chromisx/raku/./oo.pl6:26
------>         $.say-times:⏏ 2
    expecting any of:
        colon pair
```

## Environment

```
This is Rakudo version 2019.07.1 built on MoarVM version 2019.07.1
implementing Perl 6.d.
```

Thanks!

Regards,

Raymond.

vrurg:
Fixed with #3317

vrurg:
Test commited with perl6/roast@03a18d29c69d3caf8cbd744c23d4871a7cb8b815. Closing.

chromis:
Thanks!

PR #3317, opened 2019-11-24
Allow colon syntax on $.method

vrurg:
I.e. alongside to `self.foo: 1,2` make it possible to `$.bar: 1,2`.

Resolves rakudo/rakudo#3306

### DIFF
--- a/src/Perl6/Grammar.nqp
+++ b/src/Perl6/Grammar.nqp
@@ -1822,7 +1822,7 @@ grammar Perl6::Grammar is HLL::Grammar does STD {
             [ <.unsp> | '\\' | <?> ] <?[(:]> <!RESTRICTED> #<arglist=.postcircumfix>
             :dba('method arguments')
             [
-                | ':' <!{ $*QSIGIL }> <arglist>
+                | ':' <?before \s | '{'> <!{ $*QSIGIL }> <arglist>
                 | '(' <arglist> ')'
             ]
         ]?
--- a/src/Perl6/Grammar.nqp
+++ b/src/Perl6/Grammar.nqp
@@ -1746,7 +1746,7 @@ grammar Perl6::Grammar is HLL::Grammar does STD {
     }
 
     token special_variable:sym<$.> {
-        <sym> {} <!before \w | '(' | '^' >
+        <sym> {} <!before \w | '(' | ':' | '^' >
         <.obsvar('$.')>
     }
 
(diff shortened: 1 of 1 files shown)


vrurg:
Spectests are ok.

src/Perl6/Grammar.nqp
```diff
@@ -1819,7 +1819,12 @@ grammar Perl6::Grammar is HLL::Grammar does STD {
         | {} <sigil> <!{ $*QSIGIL }> <?MARKER('baresigil')>   # try last, to allow sublanguages to redefine sigils (like & in regex)
         ]
         [ <?{ $<twigil> && ( $<twigil> eq '.' || $<twigil> eq '.^' ) }>
-            [ <.unsp> | '\\' | <?> ] <?[(]> <!RESTRICTED> <arglist=.postcircumfix>
+            [ <.unsp> | '\\' | <?> ] <?[(:]> <!RESTRICTED> #<arglist=.postcircumfix>
```
jnthn: Please don't leave commented code behind.

jnthn:
I think this is a reasonable addition, and indeed one that makes the language more consistent. I can't think of any downsides or potential ambiguities.

vendethiel:
Doesn't this change the meaning of `foo $.a:;`?

i.e.:

```
class A { method foo { 1 } }
class B { has A $.a .= new; method b { foo $.a:; } }
B.new.b
```

src/Perl6/Grammar.nqp
vrurg: Thanks, it was unintentional.

vrurg:
@vendethiel In your example it doesn't due to no whitespace after the colon. But:

```
class A { method foo { 1 } }
class B { has A $.a .= new; method b { foo $.a: 1,2; } }
B.new.b
```

It does. But it seems to be the right thing. Wish I could run Blin against this PR and see if there is any breakage in the code. But this kind of syntax is pretty exoctic and I barely expect it being used in the wild.

jnthn:
@vendethiel makes a good point. I guess today if we write `foo $!x: 42;` and then refactor it into `foo $.x: 42` then things keep working, and I think we'd be breaking that. On the other hand, the change makes us differently consistent.

> But this kind of syntax is pretty exoctic and I barely expect it being used in the wild.

It's true that the indirect object syntax seems to be used very little, and so is maybe not the thing to bias towards in this case. Still, we should see what Blin makes of it. If we find that almost nobody uses indirect object syntax at all, it may be worth discussing why we have it.

vrurg:
Just for the record: if I'm wrong and it's being used rather extensively, then it would make sense to make this fix specific to `6.e`. For now I'm avoiding this as to not introduce extra slow down into the grammar.

PR #3330, opened 2019-11-29
Make Parameter's BUILD use named args

vendethiel:
Between 3d3578 and d31fb2, the by-name marker has been lost.
This commit restores it, allowing a Parameter to be constructed
manually.

### DIFF
--- a/src/core.c/Parameter.pm6
+++ b/src/core.c/Parameter.pm6
@@ -90,15 +90,15 @@ my class Parameter { # declared in BOOTSTRAP
     }
 
     submethod BUILD(
-       Str:D $name      is copy = "",
-       Int:D $flags     is copy = 0,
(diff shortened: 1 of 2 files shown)


lizmat:
Are you sure this is not in conflict with d31fb27064 ?

vendethiel:
> Are you sure this is not in conflict with [d31fb27](https://github.com/rakudo/rakudo/commit/d31fb27064f078b8cf2f962b9ebc154f650af4b8) ?

Yes, I just asked jnthn++ about that :). Turns out the way BUILD is currently is - after that commit - is not reachable.

jnthn:
Yeah, I screwed up; I wanted to turn `SET-SELF` into `BUILD` and drop the `new`, and did so, but forgot to update the `SET-SELF` signature to take nameds.

PR #3331, opened 2019-11-29
Add support for Dict

lizmat:
- add the Dict class, inheriting from Map
- add the necessary logic to make e.g. my %d is Dict work
- make sure a .Dict coercer is available

This to make the discussion about

  https://github.com/perl6/problem-solving/issues/135

easier.


lizmat:
Even though new-disp has landed, I think we need to revisit these issues from scratch, so closing this now.

Issue #3326, opened 2019-11-27
Async socket performance regression in 2019.11


bbkr:
server.pl
```perl6
react {
    whenever IO::Socket::Async.listen('localhost', 3333) -> $connection {
        whenever $connection.Supply.lines -> $line {
            die unless $line eq 'ping'; # just checking...
            $connection.print( "pong\n" );
        }
    }
}
```

client.pl
```perl6
my $connection = IO::Socket::INET.new( host => 'localhost',  port => 3333 );

my $start = now;
for ^100000 {
    $connection.put( 'ping' );
    my $status = $connection.get( );
    die unless $status eq 'pong'; # just checking...
}
say now - $start;
```

When server.pl is running on Rakudo 2019.07.1 client reports 13.1s on average, but for server running on Rakudo 2019.11 it takes 15.5s. Thats **13% performance drop**.

Only Async sockets are affected, it does not matter which Rakudo client.pl is running on.

Kaiepi:
This doesn't seem to be related to any of the changes made to how sockets are handled since 2019.07.1. All that's changed with sockets is how they can now specify a family of address to use (which isn't exposed for use with `IO::Socket::Async` yet) and how GC is handled when listening on an async socket. The sockets created by `IO::Socket::INET` and `IO::Socket::Async` use the same addresses as before in this case, and how socket reads and writes are handled hasn't changed. Maybe it has to do with how the data received is decoded?

vrurg:
@bbkr Can you by chance try bisecting it?

AlexDaniel:
Pretty sure [I bisected it](https://colabti.org/irclogger/irclogger_log/raku-dev?date=2019-11-28#l79).

(2019-08-06) https://github.com/rakudo/rakudo/commit/ae4ba74262b1155ddf57cd2bac1f3aaccfc1e47f

I can't believe that the commit affects performance that much but it does. See https://gist.github.com/44c459e804c8a2d0231b56eb669243cf.

Can somebody try confirming it locally? It can also be a change in gcc and it just happens that whateverable got an update between these commits.

AlexDaniel:
Confirming that the GCC version didn't change between these builds:

```
ae4ba74262b1155ddf57cd2bac1f3aaccfc1e47f/bin$ objdump --full-contents --section=.comment moar

moar:     file format elf64-x86-64

Contents of section .comment:
 0000 4743433a 20284465 6269616e 20362e33  GCC: (Debian 6.3
[...]

PR #3332, opened 2019-11-30
Change the way compiler version is chosen for COREs

vrurg:
Compiling a CORE with version lower than default can cause performance
degradation because not all current optimizations would be used. Thus,
the previous decision of compiling each core with corresponding compiler
considered erroneous.

On the other hand, the next revision core is better be compiled with
its respective compiler version to have additional testing.

For these reasons, compiler version is now set using core revision if
it's higher than the default language version; and set to the default
otherwise.

Additionally, `World` now remembers the revision of the currently
compiled `CORE` to provide correct versioning for core classes.

Should fix rakudo/rakudo#3326


vrurg:
Appveyor checks are broken for now. My own test in a Win VM passed, so I consider this safe to merge.

PR #3335, opened 2019-12-01
Make the raku symlinks relative

patzim:
This makes them work when relocated.


vrurg:
Thanks! I always forget about relocability...

PR #3340, opened 2019-12-05
Prevent silently swallowing errors on submodule update

patrickbkr:
I'm currently tracking down [this rakudobrew bug](https://github.com/tadzik/rakudobrew/issues/151) and noticed that Configure.pl seems to silently ignore nqp-configure submodule update failures. This PR should hopefully make this more vocal.

@vrurg: I'm not entirely sure the ignoring of errors wasn't on purpose. (Maybe it's totally fine for this to fail in some cases?) So can you give a quick look?


vrurg:
I'd say I rather forgot about the `else` branch. No other reason comes into my mind.

PR #3355, opened 2019-12-15
Adapt release message template

patrickbkr:
I'm not sure all of these changes are actually a good idea. Thus the review.

- Move non-changing "boilerplate" text to the bottom
- Don't claim that Rakudo is on MoarVM only
- Note that pre-compiled archives might not be uploaded yet
- Don't mention Rakudo Star

### DIFF
--- a/tools/create-release-announcement.p6
+++ b/tools/create-release-announcement.p6
@@ -6,7 +6,18 @@ my $template = q:to/END_TEMPLATE/;
 
 On behalf of the Rakudo development team, I’m very happy to announce the
 «month» «year» release of Rakudo #«release-num». Rakudo is an implementation of
-Raku on the Moar Virtual Machine[^1].
+the Raku[^1] language.
+
+The source tarball for this release is available from
+<https://rakudo.org/files/rakudo>.
+Pre-compiled archives will be available shortly.
+
+New in «release-name»:
+«changes»
+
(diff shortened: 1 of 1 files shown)


AlexDaniel:
It looks fine to me!

Altai-man:
No issues with this, though:
>Don't mention Rakudo Star

Can anyone specify what's the status of this? As in, AFAIK Rakudo star was the only way to install the thing on Windows, was it changed? How "official" (related to rakudo project, not raku) is Rakudo Star now?

patrickbkr:
@Altai-man Currently it's true, that Rakudo Star is the only way to install Rakudo on Windows (except compiling). This is about to change when we start to release precompiled archives and hopefully also installers. Currently Rakudo Star is an official Rakudo release.

I plan to give Rakudo Star less prominence though, as I do believe Rakudo Star serves a purpose as a nice bundle of modules, but should not be promoted as the primary way to install Rakudo. I have changes reflecting this on rakudo.org in the works (see <https://stage-rakudo.rakulang.site/>). Disclamer: Obviously there will be a review period before any such changes are happening.

Altai-man:
Roger, thanks for the info.
I am assuming the failed build is a flapper? Others pass on gnu/linux.

patrickbkr:
Yeah, looks like a flapper. :-(

PR #3356, opened 2019-12-16
Correct eqv between non-Positional Iterables

dumarchie:
This resolves rakudo/issues/3346.

Note that I moved the `eqv` candidate to `Iterable` because `constant IterationEnd` is not defined inside `Mu`. It's also a more appropriate location from an architectural perspective.

Furthermore I did not modify the existing implementation for `Positional` iterables becomes some iterators appeared to be buggy.

### DIFF
--- a/src/core.c/Iterable.pm6
+++ b/src/core.c/Iterable.pm6
@@ -196,6 +196,60 @@ my role Iterable {
     multi method SetHash(Iterable:D:) { SETIFY(self,SetHash) }
 }
 
+multi sub infix:<eqv>(Iterable:D \a, Iterable:D \b) {
+    nqp::hllbool(
+      nqp::unless(
+        nqp::eqaddr(nqp::decont(a),nqp::decont(b)),
+        nqp::if(                                 # not same object
+          nqp::eqaddr(a.WHAT,b.WHAT),
+          nqp::if(                               # same type
+            a.is-lazy,
+            nqp::if(                             # a lazy
+              b.is-lazy,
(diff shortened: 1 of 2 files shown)


lizmat:
Could you elaborate on which iterators appear to be buggy?

dumarchie:
`Empty[^2]` produces `(Nil, Nil)`, but the `.iterator` of this list returns `IterationEnd` on the first `pull-one`.

lizmat:
dd ()[^10]
    (Nil, Nil, Nil, Nil, Nil, Nil, Nil, Nil, Nil, Nil)

So, this is nothing special about `Empty`: slicing any list with insufficient elements will produce `Nil` at the locations where there is no value.

dumarchie:
The list of `Nil`s did not worry me. What worried me was that it _looked_ as if the list `.iterator` _just_ produced `IterationEnd`. But this was because I jumped to a wrong conclusion. I just ran the spectest without the special case for `Positional` and there were no failing tests apart from those that fail on Windows anyway. So there appear to be _no_ issues with iterators.

If you like I can benchmark the iterative and the positional approaches on a large array, but I think there's a big chance that the positional approach will out-perform the iterative approach. So I suggest keeping the positional approach for iterables that do `Positional`.

dumarchie:
This approach is flawed. Comparing `Iterable`s by iterating over them is like comparing `Callable`s by calling them. It may produce unwanted side-effects like consuming a `Seq`. And it compares _behavior_, not _structure_.

Issue #3352, opened 2019-12-14
Configure.pl should detect compilers that won’t work


ugexe:
MoarVM is no longer restricted to C89. However, this means certain windows configurations using something older than VS2017 will no longer compile Rakudo. Since many appveyor setups use VS2012 it would be a good idea for the build system to give an error noting the incompatible compiler instead of trying to derive that knowledge from the appveyor output ( https://ci.appveyor.com/project/ugexe/zef/builds/29531256 )

vrurg:
@ugexe can you provide a sample of version report from the compiler?

ugexe:
On VS2017 a VM gives the following result as the first line when running `cl /?`

```
C:\Users\ugexe\.rakudobrew>cl /?
Microsoft (R) C/C++ Optimizing Compiler Version 19.12.25831 for x64
...
```

vrurg:
This one I also have. I wonder what 2012 prints to decide how to handle it correctly.

vrurg:
I don't think we want to limit ourselves to 2017 exclusively. :)

ugexe:
VS2012 on appveyor:
```
cmd /c "cl.exe"
Microsoft (R) C/C++ Optimizing Compiler Version 16.00.40219.01 for x64
Copyright (C) Microsoft Corporation.  All rights reserved.
```

vrurg:
BTW, I'm adding the check to nqp-configure, so both NQP and Rakudo will be bailing out on anything less than 19.0. But MoarVM needs separate handling which is out of my hands.

vrurg:
Closed with merge of #3359

PR #3359, opened 2019-12-17
[WIP] Test for Microsoft Compiler version

vrurg:
The report (#3352) has it that Rakudo is currently can only be compiled with VC2017. Adding a check which would make Configure.pl to bail out if compiler version is less than 19.0 (my VC2017 reports 19.20).


vrurg:
NB. This PR is currently only for testing the changes. The final version might use different submodule version.

vrurg:
> I would add a check whether the regex to match the output of `cl` actually matched. If it doesn't match, continue and maybe add a notice. Otherwise the build might break needlessly when the output changes some time in the future.

The check has to be done, that's for sure. But we can't assume that if it fails then it's some newer version. It could as well be something else.

niner:
If the check fails because a newer version of VS changed the version output, we prevent a build needlessly. If the check fails because of an ancient version with output we didn't anticipate, or because something else, the build would fail anyway and the user would hopefully report it.

I think the check should be conservative and only prevent build where we really know that they would fail.

PR #3360, opened 2019-12-17
Pod::To::Text display enums

kalkin:
This was part of the stuck PR https://github.com/rakudo/rakudo/pull/1037
I think this commit has merit on its own.

### DIFF
--- a/lib/Pod/To/Text.rakumod
+++ b/lib/Pod/To/Text.rakumod
@@ -93,6 +93,9 @@ sub declarator2text($pod) {
         when Sub {
             'sub ' ~ $_.name ~ signature2text($_.signature.params)
         }
+        when .HOW ~~ Metamodel::EnumHOW {
+            "enum $_.raku() { signature2text $_.enums.pairs } \n"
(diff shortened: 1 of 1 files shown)


lib/Pod/To/Text.pm6
```diff
@@ -93,6 +93,9 @@ sub declarator2text($pod) {
         when Sub {
             'sub ' ~ $_.name ~ signature2text($_.signature.params)
         }
+        when .HOW ~~ Metamodel::EnumHOW {
+            "enum $_.perl() { signature2text $_.enums.pairs } \n"
```
Xliff: Are you sure you don't mean the following, here?

```
enum ' ~ $_.perl ~ ' ' ~ signature2text($_.enums.pairs) ~ " \n"
```

kalkin: `"$_.perl()"` is fine. Method calls on interpolated variables are supported by Raku

kalkin:
Solved the conflict

PR #3389, opened 2019-12-31
Attribute $!package fix

vrurg:
This fix guarantees that traits applied to attributes can now safely use `$attr.package` if applied to an attribute in a role. This code:

```perl6
multi trait_mod:<is> (Attribute:D $a, :$foo!) {
    note $a.package.^name;
}

role R {
    has $.a is foo;
}
```

without this PR outputs `$?CLASS`. With the PR it would output `R`, as actually expected.

This PR implements the fix by using `$*PACKAGE` instead of `$?CLASS` symbol to initialize attribute `package` at compilation time. The attribute is later changed to the actual class the attribute is being installed into.

Alternative path of resolving the problem would involve introduction of an additional symbol which would resolve to either `$?ROLE` or `$?CLASS` depending on the context. The `package` would be set to that symbol then. But this path is somewhat cumbersome and would introduce additional symbol into the user namespace.

### DIFF
--- a/src/Perl6/Metamodel/ParametricRoleHOW.nqp
+++ b/src/Perl6/Metamodel/ParametricRoleHOW.nqp
@@ -186,7 +186,7 @@ class Perl6::Metamodel::ParametricRoleHOW
         # the concrete role.
         for self.attributes($obj, :local(1)) {
             $conc.HOW.add_attribute($conc,
-                $_.is_generic ?? $_.instantiate_generic($type_env) !! $_);
+                $_.is_generic ?? $_.instantiate_generic($type_env) !! nqp::clone($_));
         }
 
         # Go through methods and instantiate them; we always do this
--- a/src/Perl6/Actions.nqp
+++ b/src/Perl6/Actions.nqp
@@ -3750,7 +3750,7 @@ class Perl6::Actions is HLL::Actions does STDActions {
                 has_accessor => $twigil eq '.',
                 container_descriptor => $descriptor,
                 type => %cont_info<bind_constraint>,
-                package => $*W.find_symbol(['$?CLASS']));
+                package => $*PACKAGE);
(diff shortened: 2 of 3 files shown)


src/Perl6/Metamodel/ParametricRoleHOW.nqp
```diff
@@ -186,7 +186,7 @@ class Perl6::Metamodel::ParametricRoleHOW
         # the concrete role.
         for self.attributes($obj, :local(1)) {
             $conc.HOW.add_attribute($conc,
-                $_.is_generic ?? $_.instantiate_generic($type_env) !! $_);
+                $_.is_generic ?? $_.instantiate_generic($type_env) !! nqp::clone($_));
```
jnthn: Hm, I think far cleaner would be for `is_generic` to return `True` when the package of the attribute is generic, and then generic instantiation to do the substitution, rather than having it happen in an ad-hoc way in `Attribute.compose`.

jnthn: Of course, that means one cannot do the other change. But attributes are never truly installed on a role, but rather exist on the classes that a role is composed in to. So it'd seem appropriate that the `package` of an attribute in a role is generic.

vrurg: @jnthn The point is that traits applying to attributes often have to consider if `$attribute.package` is valid or it is not. It's been advised a few times that `$*PACKAGE` is to be used instead of `$attribute.package` within a trait code.

The point of this PR is to reduce confusion and to have `$!package` set to what people usually expected it to be set to. I target at improving overall user experience.

With regard to `is_generic` – by what you say it would then be true for all attributes in roles until applied to a class because `$?CLASS` is generic. But that's certainly not correct.

Issue #3382, opened 2019-12-25
A method in a role masks a public attribute in a class doing that role


lizmat:
role A {
        method a() { 666 }
    }
    class B does A {
        has $.a = 42;
    }
    dd B.new.a     # 666 rather than 42

I think this is due to the accessor creation logic looking at whether the class is already providing a method, but disregarding the fact that that method is provided by the role.

Or is this a case of DIHWIDT?

vrurg:
It looks like it's by design. Incorporated methods are the same first-class citizens and class' own. A role is being applied as early as possible during class composition (is it actually documented somewhere?) whereas accessors are children of a `BUILD` constructor which is invoked much later.

I think a warning could be issued here if `BUILD` finds out that a pre-existing method is actually originating from a role. But even in this case I wouldn't report this without an additional pragma like `use warnings :extra;`.

lizmat:
So we're leaning towards DIHWIDT?

Thing is that the role could be imported from an external module, so one could be using that and not notice that the accessor is never made.

In any case, this is a compilation error:

    role A {
        method a() { 666 }
    }
    class B does A {
        has $.a = 42;
    }

So maybe we shouldn't allow a public attribute like that in a class either when there is a method in the role that would mask the attribute's accessor ?

vrurg:
Is it a real compilation error or imaginary one? For me the code always compiles.

Anyway, I would rather agree to the statement that the situation has to be reported somehow. Either as a warning, or error, but we must not leave it unnoticed. But this looks like a @jnthn word on this is needed.

BTW, can't quickly locate how exactly accessors are generated. Must be a part of `BUILDPLAN` but brief overlook didn't come up with precise location.

jonathanstowe:
It's in the `compose` method of `Attribute`

vrurg:
@lizmat You could try the initial solution with #3390. It currently has some limits:

- A warning is only issued at compile time, when `$*LANG` is available. There're two reasons for this: I don't know how to issue a standard warning from MOP code at run time; I don't know if there is a way to have access to pragmas at run-time to `die` if `fatal` is set.

- I only warn if attribute originates from the current object (i.e. most likely – a class). Otherwise core fails compiling `DateTime` because `Datish` defines `$.daycount` and has `method daycount`. `fatal` is on for the `CORE`, so `$*LANG.worry` just dies. It could be solved with replacing `$.daycount` with `$!daycount` which is anyway the right way to have it, perhaps.

[...]

PR #3390, opened 2019-12-31
[WIP] Warn if attribute accessor cannot be installed

vrurg:
Only compile time for now.

A solution for #3382


jnthn:
This is wrong; it's a feature that you can declare something `has $.foo` and then provide a custom accessor. The use of the `.` sigil isn't just accessor generation; it's also gives a different initialization and destructuring semantics.

lizmat:
@jnthn but this is not about that?  This is about a *method* in a role preventing the creation of an accessor method for a public attribute in a class consuming that role.  See https://github.com/rakudo/rakudo/issues/3382

jnthn:
@lizmat Yeah; I've proposed a cleaner solution to that on the issue. I worry this one will have false positives in other cases, and isn't really solving the actual issue. :-)

vrurg:
Agree to @jnthn. I had a spare hour to try this one out, but it's not the right approach.

Dropping away...

PR #3393, opened 2019-12-31
Fix compiler warnings in main runner

MasterDuke17:
Now builds clean with `--relocatable` and without.

### DIFF
--- a/src/vm/moar/runner/main.c
+++ b/src/vm/moar/runner/main.c
@@ -148,7 +148,9 @@ int wmain(int argc, wchar_t *wargv[])
 
     char   *exec_path;
     size_t  exec_path_size;
+#ifndef STATIC_EXEC_PATH
     int     res;
+#endif
 
     char   *exec_dir_path;
     char   *exec_dir_path_temp;
--- a/src/vm/moar/runner/main.c
+++ b/src/vm/moar/runner/main.c
@@ -148,9 +148,6 @@ int wmain(int argc, wchar_t *wargv[])
 
     char   *exec_path;
     size_t  exec_path_size;
-#ifndef STATIC_EXEC_PATH
-    int     res;
-#endif
 
     char   *exec_dir_path_temp;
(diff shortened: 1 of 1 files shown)


src/vm/moar/runner/main.c
```diff
@@ -186,8 +186,10 @@ int wmain(int argc, wchar_t *wargv[])
     int flag;
     int new_argc     = 0;
 
+#ifdef HAVE_TELEMEH
```
niner: Have you considered moving these variable declarations down into the existing ifdef in line 262? Would save some clutter.

src/vm/moar/runner/main.c
```diff
@@ -148,7 +148,9 @@ int wmain(int argc, wchar_t *wargv[])
 
     char   *exec_path;
     size_t  exec_path_size;
+#ifndef STATIC_EXEC_PATH
```
niner: You could just move the int res down into line 197

src/vm/moar/runner/main.c
MasterDuke17: Inside the `if (getenv("MVM_TELEMETRY_LOG")) {`? But then `interval_id` wouldn't be available at https://github.com/rakudo/rakudo/pull/3393/files/bebd268f6a4ccb8e5cdd96b2d74da015065f34d3..293a0b631fe93fbe2ab3858d3b7455e62a71baad#diff-61d548d860ffcb20f03c63354f4132c0L386

src/vm/moar/runner/main.c
MasterDuke17: Done in fd59414.

src/vm/moar/runner/main.c
niner: No, just before the if:
```
#ifdef HAVE_TELEMEH
    unsigned int interval_id = 0;
￼    char telemeh_inited = 0;
    if (getenv("MVM_TELEMETRY_LOG") && telemeh_inited) {
```

PR #3394, opened 2020-01-01
Check type before looking for qualified method

usev6:
This fixes a failing test in S12-methods/qualified.t on the JVM
backend. The code in src/vm/moar/spesh-plugins.nqp does a similiar
check before calling 'find_method_qualified'.

A somewhat golfed version of the failing test would be:

  role R  { method x { 'R::x'  } }
  class C { method x { self does R; self.R::x } }
  say C.new.x;

Running this code fails with

  No concretization found for R

instead of printing 'R::x' on the JVM backend.

### DIFF
--- a/src/core.c/Mu.pm6
+++ b/src/core.c/Mu.pm6
@@ -810,7 +810,8 @@ Perhaps it can be found at https://docs.perl6.org/type/$name"
                 }
                 $ctx := nqp::ctxouterskipthunks($ctx);
             } while $ctx && !$sym-found;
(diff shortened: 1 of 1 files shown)


vrurg:
As far as I understand, the reason for this is because spesh in moar is using the fallback purely for reporting a problem. But JVM doesn't do this kind of optimization and the sub just falls short on replicating speh plugin behavior. Am I right?

usev6:
That's how I understand the problem, yes.

vrurg:
Appveyor failure looks weird but seems to be unrelated to the PR itself.

PR #3396, opened 2020-01-02
Allow DateTime to nanoseconds.

rbt:
This was restricted via regex and can be enlarged with minimal change. Other entities such as Google Cloud commonly use nanosecond times in their APIs.

Simple nanosecond roundtrip test:
https://github.com/perl6/roast/pull/609

### DIFF
--- a/src/core.c/DateTime.pm6
+++ b/src/core.c/DateTime.pm6
@@ -224,7 +224,7 @@ my class DateTime does Dateish {
           ':'
           (\d\d)                                         # minute
           ':'
-          (\d\d[<[\.,]>\d ** 1..6]?)                     # second
+          (\d\d[<[\.,]>\d ** 1..12]?)                    # second
           (<[Zz]> || (<[\-\+]>) (\d\d) (':'? (\d\d))? )? # timezone
         $/;
 


vrurg:
Not sure if nano-precision is always good in the default output. I'd rather have it optional.

rbt:
Any recommendations on how to make it configurable?

Something like this, where the default is 6?
`DateTime.new(formatter-second-precision => 9);`

Or did you mean the user should provide their own formatter function? It's a surprisingly complicated function to get all the details.

On my machine I can get 12 digits, so it does seem like someone would eventually want picosecond precision.

```
DateTime.now(formatter => { sprintf "%09.12f", .second });
59.742140769958
```

vrurg:
Frankly saying, I'm not really sure which way it would be better. Basically, just `precision` property would be sufficient simply because there is nothing else but seconds it can be applied to. But it implies that both parsing and output must be affected. What if somebody needs different precision for output? Though I tend to think that in the latter case it'd be better to provide one's own formatter.

rbt:
Currently, the input for broken down values ignores the second precision entirely. It's only the ISO8601 input which enforces a maximum precision.

Changing the ISO8601 input to accept additional digits (up to 12?) while output only shows 6 would be consistent with the other methods of new().

I've updated my patch to do this which does not require any changes to the tests (but I'll add a couple examples anyway).
```
> DateTime.new(:2010year, :10month, :10day, :5hour, :5minute, second => '59.987654321').say;
2010-10-10T05:05:59.987654Z

# 2019.11
> DateTime.new('2010-10-10T05:05:59.987654321Z');
Invalid DateTime string '2010-10-10T05:05:59.987654321Z'; use an ISO 8601 timestamp (yyyy-mm-ddThh:mm:ssZ or yyyy-mm-ddThh:mm:ss+01:00) instead

# With patch
> DateTime.new('2010-10-10T05:05:59.987654321Z').say
2010-10-10T05:05:59.987654Z
```

For what it's worth, PostgreSQL (my goto for time related stuff) also truncates the precision silently.

vrurg:
I'll wait for travis to pass the tests and if it's all green will merge this then.

vrurg:
Thanks @rbt!

Now, it's only the minor fix for roast to be done.

PR #3403, opened 2020-01-03
Actually return found method

usev6:
I have to admit that I don't know what code would trigger this code.
But the missing 'return' was obviously an oversight (cmp. also
https://colabti.org/irclogger/irclogger_log/raku-dev?date=2020-01-02#l302).


usev6:
Maybe it's cleaner to remove the trailing ```nqp::null()```. But since I've stumbled about this line yesterday, I want to make sure it doesn't pass out of mind -- hence this PR.

lizmat:
looking at this now

lizmat:
Implemented slightly differently with abcf37f2e3 .   Thanks for the spot and the fix!

usev6:
Thanks @lizmat. Your fix looks better -- esp. since the other day I've stumbled upon a problem with the result of nqp::findmethod being messed up by an explicit return.

lizmat:
Yeah, I had that in mind when making that change  :-)

PR #3404, opened 2020-01-04
Add an "is accessible" attribute to Attribute

lizmat:
This allows you to define a private attribute in a class, yet have
an accessor automatically created for it (without it being settable
with .new):

    class A {
        has $!a is accessible = 42;
    }
    dd A.new(a => 666).a;   # 42


lizmat:
Related to this problem-solving ticket: https://github.com/perl6/problem-solving/issues/145

PR #3413, opened 2020-01-10
Run add_phasers_handling_code once for methods

### LINKED ISSUE #3411: ENTER fires more often than expected
## The Problem
The below code snipped spits out "IN f" twice.
```
class A {
	method f {
		ENTER {say "IN f"}
		LEAVE {say "OUT f"}
		say "EXECUTE f";
	}
}

A.new().f;

```
Output:
```
IN f
IN f
EXECUTE f
OUT f
```

## Expected Behavior
ENTER block will only be fired a single time.

Output:
```
IN f
EXECUTE f
OUT f
```

## Environment
Linux rbt-inspiron-123 5.3.0-24-generic #26-Ubuntu SMP Thu Nov 14 01:33:18 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux

This is Rakudo version 2019.11 built on MoarVM version 2019.11
implementing Perl 6.d.

rbt:
Trouble code:
class A {
    method f() {
        ENTER say "foo";
    }
}
A.f;   # foo foo

A simple skip flag is used to prevent the call duplication.

Thanks to Elizabeth Mattijsen for pointing me at the problem code.

Closes #3411

### DIFF
--- a/src/Perl6/Actions.nqp
+++ b/src/Perl6/Actions.nqp
@@ -3913,6 +3913,8 @@ class Perl6::Actions is HLL::Actions does STDActions {
             :method, :$invocant_type);
         my $code := methodize_block($/, $*W.stub_code_object('Method'),
             $a_past, $signature, %sig_info);
+        $*W.add_phasers_handling_code($code, $a_past);
+
         install_method($/, $meth_name, 'has', $code, $install_in, :gen-accessor);
     }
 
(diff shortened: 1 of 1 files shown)


lizmat:
hmmm...  I was hoping we'd not be needing an additional flag.  @jnthn suggestions?

rbt:
There are only 3 callers of methodize_block. A better alternative is to have the other 2 callers do add_phasers_handling_code themselves.

Making add_phasers_handling_code idempotent looks like a headache.

lizmat:
I think I would prefer that approach, actually.  It makes it clearer where what is being done, rather than sort of hiding that behind a named parameter.

rbt:
Okay. I changed it to work that way instead.

The failure looks like a perl server issue.

lizmat:
Thanks for the PR and the changes!

PR #3418, opened 2020-01-12
Add Str.(r)index(@needles)

lizmat:
Sometimes you want to find the first (or last in the case of rindex)
of a number of characters in a string.  This is now easily possible:

    "foobarbaz".index(<z a o>);   # 1
    "foobarbaz".rindex(<o a z>);  # 8

Of course, if none of the needles are found, it returns Nil


lizmat:
With the implementation of `:i(ignorecase)` and `:m(ignoremark)` support, this needs to be expanded.  Will do so in separate commits.

PR #3422, opened 2020-01-13
Implement `--git-reference` Configure.pl option

patrickbkr:
The option was documented but not implemented anymore. I suspect it was
lost during the great build system refactor.
This option makes the configure process create / update git repositories
for all of its submodules and NQP in the given folders and use those
as reference repos. The option is forwarded to a potential NQP
sub-build.


vrurg:
I never used referencing, can tell nothing on the functionality itself. But what worries me is bloating of the `BEGIN` block. It is never a good idea to overexploit it. Can we do it some other way around? Event a separate script spawned from `BEGIN` code would do better to me.

PR #3423, opened 2020-01-14
Implementation of the "is built" attribute trait

lizmat:


lizmat:
Related to this problem-solving ticket: https://github.com/perl6/problem-solving/issues/145

PR #3424, opened 2020-01-14
Make default degree for hyper/batch make more sense

lizmat:
Currently, the default :degree for .hyper / .race is set to 4.  This may be too high for small single core CPU's, and to small for bigger beasts (such as my MBP with 2x8 cores).  Set the default to the number of CPU cores available, and introduce the $*DEGREE dynamic variable for setting the default for a given scope.

### DIFF
--- a/src/core.c/Iterable.pm6
+++ b/src/core.c/Iterable.pm6
@@ -28,7 +28,7 @@ my role Iterable {
         Seq.new(Rakudo::Iterator.Lazy(self))
     }
 
-    method hyper(Int(Cool) :$batch = 64, Int(Cool) :$degree = 4) {
+    method hyper(Int(Cool) :$batch = 64, Int(Cool) :$degree = $*DEGREE // nqp::cpucores) {
 #?if !js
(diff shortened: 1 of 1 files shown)


vrurg:
Perhaps it would make sense to leave one core free for other resources in case hyper/race would be too much of a hog? Not really sure as I have arguments for both pro and cons.

jnthn:
> Perhaps it would make sense to leave one core free for other resources in case hyper/race would be too much of a hog? Not really sure as I have arguments for both pro and cons.

One fairly solid argument for the `- 1` is that the degree is the number of workers, but the original thread is still responsible from pulling from the source iterator and downstream processing of the results beyond the parallel phase. Often that is some amount of work.

lizmat:
So it's now nqp::cpucores - 1, with a minimum of 1.  @jnthn any comments about `$*DEGREE` ?

jnthn:
@lizmat Quite comfortable with the new default. Not immediately so sure about `$*DEGREE`.

PR #3434, opened 2020-01-19
NativeCall: add CArray[…].new(Str:D, …) candidate

taboege:
This constructor creates a CArray by first encoding the given string
and then appending a NUL byte. Its primary use would be to make a copy
of a string being passed to a native function and keeping a reference
to that copy, which the `is encoded` trait does not permit.

This came up in Raku/doc#3171 and refs Raku/doc#3169.

### DIFF
--- a/lib/NativeCall/Types.rakumod
+++ b/lib/NativeCall/Types.rakumod
@@ -244,6 +244,10 @@ our class CArray is repr('CArray') is array_type(Pointer) {
             nqp::create(self)
         }
     }
+    multi method new(Str:D $str, :enc(:$encoding) = 'utf8', *%encode-opts) {
+        # Add NUL terminator to the encoded C string:
+        self.new($str.encode($encoding, |%encode-opts).list, 0)
+    }
 }
 
 # duplicated code from NativeCall.pm to support Pointer.deref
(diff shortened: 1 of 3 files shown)


niner:
A little over a year ago I had a look at memory management issues surrounding strings in NativeCall:
https://colabti.org/irclogger/irclogger_log/perl6-dev?date=2018-09-01#l38

This resulted in the following API proposal: https://gist.github.com/niner/53c4da1355802e042139ee05a1d7895b

As mentioned in the log, the memory management issues do not affect only strings. That's probably just the most common use case.

lizmat:
I had completely forgotten about that proposal.  @niner++  @taboege perhaps a good thing to implement instead?

taboege:
Thank you. Definitely something to look at!

lib/NativeCall/Types.rakumod
```diff
@@ -244,6 +244,10 @@ our class CArray is repr('CArray') is array_type(Pointer) {
             nqp::create(self)
         }
     }
+    multi method new(Str:D $str, :enc(:$encoding) = 'utf8', *%encode-opts) {
+        # Add NUL terminator to the encoded C string:
+        self.new($str.encode($encoding, |%encode-opts).list, 0)
```
fecundf: The problem with this
self.new($data.encode($encoding).list, 0)

is that at least one encoding, UTF-16, needs two 0 bytes to terminate the string.

Would this work?

self.new(($data~"\0").encode($encoding).list)

(also posted this same comment in the issue discussion)

lizmat:
This PR appears to have stalled.  Maybe it should be closed?

lizmat:
Closing as my remark of >1 year ago, has not been replied on.

PR #3446, opened 2020-01-27
Decontainerize arguments in eqv between Seqs

dumarchie:
Different containers holding the same Seq have equal value.


lizmat:
Thanks!

PR #3449, opened 2020-01-29
Typecheck Attribute's build attribute against Code

vrurg:
Attrbiute `build` of `Attrbiute` class can not only be set to a method
but to any kind of generic code. This fixes a breakage in `Red` ORM.

Related to FCO/Red#422


vrurg:
@lizmat I'm not merging instantly as you might have another idea about the fix. After all, it was your change. :)

lizmat:
Thanks!

PR #3451, opened 2020-01-30
[IP6NS Grant] Implement support for bind before connect with IO::Socket::INET

Kaiepi:
This also helps with upcoming support for UDP sockets with `IO::Socket::INET`.

See https://github.com/MoarVM/MoarVM/pull/1231 and https://github.com/perl6/nqp/pull/597

### DIFF
--- a/src/core.c/IO/Socket/INET.pm6
+++ b/src/core.c/IO/Socket/INET.pm6
@@ -122,8 +122,8 @@ my class IO::Socket::INET does IO::Socket {
         # which is derived from the protocol, is SOCK_STREAM then connect() is called.
         if $!listening || $!localhost || $!localport {
             nqp::bindsock($PIO, nqp::unbox_s($!localhost || "0.0.0.0"),
-                                 nqp::unbox_i($!localport || 0), nqp::unbox_i($!family),
-                                 nqp::unbox_i($!backlog || 128));
+                                 nqp::unbox_i($!localport || 0), nqp::unbox_i($!family));
+            nqp::listen($PIO, nqp::unbox_i($!backlog || 128));
         }
 
         if $!listening {
--- a/src/core.c/IO/Socket/INET.pm6
+++ b/src/core.c/IO/Socket/INET.pm6
@@ -54,69 +54,73 @@ my class IO::Socket::INET does IO::Socket {
 
     # Create new socket that listens on $localhost:$localport
     multi method new(
-        Bool   :$listen! where .so,
-        Str    :$localhost is copy,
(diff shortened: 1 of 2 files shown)


src/core.c/IO/Socket/INET.pm6
```diff
@@ -99,10 +100,19 @@ my class IO::Socket::INET does IO::Socket {
             :$family,
         ) unless $family == nqp::const::SOCKET_FAMILY_UNIX;
 
-        # TODO: Learn what protocols map to which socket types and then determine which is needed.
+        with $localhost // $localport {
+            die 'Cannot set a source host or port for UNIX sockets'
```
jnthn: Would be better (and better testable) as a typed exception.

src/core.c/IO/Socket/INET.pm6
```diff
@@ -99,10 +100,19 @@ my class IO::Socket::INET does IO::Socket {
...
+                if $family == nqp::const::SOCKET_FAMILY_UNIX;
+            ($localhost, $localport) = split-host-port(
+                :host($localhost),
+                :port($localport),
+                :$family
+            ) orelse fail $_;
```
jnthn: This is probably wrong, 'cus `orelse` is looser than `=`. So it's testing the `List`, which is never undefined, and `$localhost` ends up with the exception. Probably also indicates a missing test.

```
$ perl6 -e 'sub foo() { fail "bar" }; sub baz() { my ($x, $y); ($x, $y) = foo() orelse fail $_ }; dd baz()'
(Failure.new(exception => X::AdHoc.new(payload => "bar"), backtrace => Backtrace.new), Any)
$ perl6 -e 'sub foo() { fail "bar" }; sub baz() { my ($x, $y); ($x, $y) = (foo() orelse fail $_) }; dd baz()'
Failure.new(exception => X::AdHoc.new(payload => "bar"), backtrace => Backtrace.new)

```

Kaiepi: Another issue with this type of failure is the candidate that handles connecting doesn't do it at all. I think that belongs in a separate commit, though.

Kaiepi:
@jnthn the issues you mentioned should be resolved now.

Kaiepi:
This solution doesn't work well for backends that don't allow separate binding and listening of streaming sockets, such as the JVM, where `bindsock` would need to keep addresses as state, which is less efficient to handle in a thread-safe way as it would be if they didn't need to to begin with. I'll rewrite this to keep binding/listening combined in the `bindsock` op at some point.

PR #3452, opened 2020-01-30
Fix regression with Attribute::build typecheck

vrurg:
This PR fixes a regression when `Attribute::build` can be set to a code object different from `Method`.

It's a follow-up to #3449 which caused breakage to some spectests. The problem was caused by the fact that `create_BUILDPLAN` can be called not only at compile-time (or, more precisely, at comp unit parse time) but at other stages too like optimization or even run-time, as a result of dynamic typeobject creation (think mixins or custom type objects). In this case `$*W` might not be available; or even if available, `find_symbol` method might not function properly as it expect parse-time structures to be at its hands. The fix currently skips type checking if we're not at unit parse stage.

If it makes sense to perform the check at run-time (and I think it actually does make sense) then the complete fix would require making `Code`, `Associative`, `Positional`, and `X::TypeCheck::Attribute::Default` available to the check. It should be possible either via using `nqp::gethllsym` or with `nqp::p6clientcorectx`. The latter is slower as it traverses contexts. So, it makes more sense to use the former.


vrurg:
And yes, it passes spectests.

If I get enough time, I'm gonna try to implement the run-time support too.

lizmat:
Thanks!  Not sure what you mean "implement the run-time support" too, as that is already implemented, is it not?  :-)

lizmat:
Ah, if you mean the bit that needs codegen in the BUILDPLAN executor, I'm already working on that, just got a bit distracted  :-)

vrurg:
@lizmat I meant that when a typeobject is being created at run-time same rules must apply. With the current version I strongly suspect that using `set_build` it could be possible to create an attribute with wrong value type.

PR #3453, opened 2020-01-30
Make attribute defaults typecheck work at run-time too.

vrurg:
Follow-up to #3452.

This PR turns on `create_BUILDPLAN` typechecking on attribute defaults for run-time. If a type is created by user code dynamically, any type mismatch could be detected as early as `compose` is called.

### DIFF
--- a/src/Perl6/Metamodel/BUILDPLAN.nqp
+++ b/src/Perl6/Metamodel/BUILDPLAN.nqp
@@ -133,29 +133,36 @@ role Perl6::Metamodel::BUILDPLAN {
 
                 # compile check constants for correct type
                 if nqp::isconcrete($default) {
+                    my $Code;
+                    my $Positional;
+                    my $Associative;
                     if !nqp::isnull(nqp::getlexdyn('$*W')) && $*W.in_unit_parse {
-                        # We're not currently compiling, skip typechecking for now.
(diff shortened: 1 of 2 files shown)


vrurg:
Not really sure how much sense does it really make though. It looks like a safeguard code to me anyway.

I thought mixins would be affected as they're run-time things, but quick tests didn't come up with any situation where the code would throw at run time. On the other hand, while testing I clearly seen `t/spec/S14-roles/parameterized-basic.t` failing because `nqp::gethllsym` was returning a wrong symbol due to a bug. So, I conclude that there're situations in the wild when typechecking happens after compunit parsing is done and `$*W` cannot be used. So I think this PR should make sense.

lizmat:
will look at this later tonight or tomorrow

lizmat:
Ok, I've looked at this PR several times now, and I fail to see how this would actually add runtime attribute typechecks.  Since  none of this affects the BUILDPLAN unrolling in Perl6/World.nqp.  So am I missing something?  Could you come up with a piece of code that fails now, and would not fail with this PR???

vrurg:
Closing because I also haven't found any case where this PR would work. At some point I was getting weird rakudo compilation errors while receiving incorrect typeobjects from `gethllsym`, but those was probably caused by something else.

Thanks, @lizmat!

PR #3454, opened 2020-01-31
Removes full directory lock on .precomp, allowing for faster parallel compilation

Xliff:
This is by -no means- intended to be a complete PR. I would like comments and assistance before I would even consider this ready for merging into master.

What I would like are thoughts on the changes herein.

Some numbers:

[p6-GLib](https://github.com/Xliff/p6-GLib)
- Non-parallel compile:
> real 342.19
> user 425.71
> sys 18.44
- Parallel compile: 
> Total compile time: 78.40569049s

[p6-GIO](https://github.com/Xliff/p6-GIO)
- Non-parallel compile:
> real 886.80
> user 1097.50
> sys 52.87
- Parallel compile:
> Total compile time: 179.515837s

[p6-Pango](https://github.com/Xliff/p6-Pango)
- Non-parallel compile:
> real 167.37
> user 207.07
> sys 9.63
- Parallel compile:
> Total compile time: 55.9401686s

[p6-GDK](https://github.com/Xliff/p6-GDK)
- Non-parallel compile:
> real 268.98
> user 323.06
> sys 13.92
- Parallel compile:
> Total compile time: 68.391482s

...and finally...

[o6-GtkPlus](https://github.com/Xliff/p6-GDK)
- Non parallel compile:
> real 1918.10
> user 2270.30
> sys 107.15
- Parallel compile:
> Total compile time: 432.639475s

The code has been commented, and if anyone has questions, I'll be more than happy to answer.

This code would not have been possible without the help of niner++ and the folks in ##raku-dev. Thank you so much for your help!

### DIFF
--- a/src/core.c/CompUnit/PrecompilationRepository.pm6
+++ b/src/core.c/CompUnit/PrecompilationRepository.pm6
@@ -161,10 +161,7 @@ class CompUnit::PrecompilationRepository::Default does CompUnit::PrecompilationR
         }
 
         if $resolve {
-            if self.store.destination($compiler-id, $precomp-unit.id, :extension<.repo-id>) {
-                self.store.store-repo-id($compiler-id, $precomp-unit.id, :repo-id($repo.id));
-                self.store.unlock;
-            }
+            self.store.store-repo-id($compiler-id, $precomp-unit.id, :repo-id($repo.id));
         }
         True
     }
@@ -259,6 +256,8 @@ class CompUnit::PrecompilationRepository::Default does CompUnit::PrecompilationR
         }
         my $source-checksum = nqp::sha1($path.slurp(:enc<iso-8859-1>));
         my $bc = "$io.bc".IO;
+
+        #LEAVE { self.store.unlock }
 
         $lle     //= Rakudo::Internals.LL-EXCEPTION;
         $profile //= Rakudo::Internals.PROFILE;
--- a/src/core.c/CompUnit/PrecompilationStore/File.pm6
+++ b/src/core.c/CompUnit/PrecompilationStore/File.pm6
@@ -114,6 +114,7 @@ class CompUnit::PrecompilationStore::File does CompUnit::PrecompilationStore {
     has %!compiler-cache;
     has %!dir-cache;
     has Lock $!update-lock = Lock.new;
+    has $!lock-file;
 
     submethod BUILD(IO::Path :$!prefix --> Nil) {
     }
@@ -148,17 +149,20 @@ class CompUnit::PrecompilationStore::File does CompUnit::PrecompilationStore {
 
     method unlock() {
         return if $*W && $*W.is_precompilation_mode();
-        {
-            LEAVE $!update-lock.unlock;
-            die "unlock when we're not locked!" if $!lock-count == 0;
-            $!lock-count-- if $!lock-count > 0;
-            if $!lock && $!lock-count == 0 {
-                $!lock.unlock;
-                $!lock.close;
-                $!lock = Nil;
-            }
-            True
-        }
+        # {
+        #     LEAVE $!update-lock.unlock;
+        #     die "unlock when we're not locked!" if $!lock-count == 0;
+        #     $!lock-count-- if $!lock-count > 0;
+        #     if $!lock && $!lock-count == 0 {
+        #         $!lock.unlock;
+        #         $!lock.close;
+        #         $!lock = Nil;
+        #     }
+        #     True
+        # }
+        return unless $!lock-file;
+        $!lock-file.unlock;
+        $!lock-file.close;
     }
 
     method load-unit(CompUnit::PrecompilationId $compiler-id,
@@ -199,7 +203,16 @@ class CompUnit::PrecompilationStore::File does CompUnit::PrecompilationStore {
             $!prefix.mkdir or return;
         }
         return unless $!prefix.w;
-        self!lock();
+
+        # The file-lock is actually an opened file-handle that is then locked.
+        $!lock-file = self!file(
+          $compiler-id,
+          $precomp-id,
+          :extension( $extension ~ '.lock' )
+        ).open(:w);
+        die 'Cannot open precompiler lock file!' unless $!lock-file;
+        $!lock-file.lock;
+
         self!file($compiler-id, $precomp-id, :$extension);
     }
 
@@ -237,9 +250,7 @@ class CompUnit::PrecompilationStore::File does CompUnit::PrecompilationStore {
                  CompUnit::PrecompilationId $precomp-id,
                  :$repo-id!)
     {
-        my $repo-id-file = self!file($compiler-id, $precomp-id, :extension<.repo-id.tmp>);
-        $repo-id-file.spurt($repo-id);
-        $repo-id-file.rename(self!file($compiler-id, $precomp-id, :extension<.repo-id>));
+        try self!file($compiler-id, $precomp-id, :extension<.repo-id>).spurt($repo-id);
     }
 
     method delete(


Altai-man:
I am not in a position for a proper review, but I assume the commented out code will be removed eventually, right?

Xliff:
> I am not in a position for a proper review, but I assume the commented out code will be removed eventually, right?

@Altai-man :Yes. As I said before, I am opening this PR primarily for discussion. If the discussion is favorable, then the code can be made ready for merge.

src/core.c/CompUnit/PrecompilationRepository.pm6
```diff
@@ -233,14 +233,22 @@ class CompUnit::PrecompilationRepository::Default does CompUnit::PrecompilationR
         my $compiler-id = CompUnit::PrecompilationId.new-without-check($*RAKU.compiler.id);
         my $io = self.store.destination($compiler-id, $id);
         return False unless $io;
+
+        my $bc = "$io.bc".IO;
+        my $bc-lock = "$io.bc.lock".IO;
+
+        $bcl = $bc-lock.open(:w);
+        self.store.unlock;
```
niner: Shouldn't the lock on the whole store be obsoleted by the lock on the individual precomp file?

We shouldn't add more assumptions that precomp files are stored in a directory structure as files. Any change should get us away from that instead. So we should either have the precomp store manage the locking of the individual precomp file, or put the lock in a temp directory. Since the lock will only be useful if it's name is predictable (so other rakudo processes will use the same file), having the precomp store manage locking seems preferable. That would also fit well with the existing API (the lock getting taken implicitly by `self.store.destination` which already gets all the information it needs to lock the individual precomp file.

src/core.c/CompUnit/PrecompilationRepository.pm6
```diff
@@ -233,14 +233,22 @@ class CompUnit::PrecompilationRepository::Default does CompUnit::PrecompilationR
...
+
         my $RMD = $*RAKUDO_MODULE_DEBUG;
         if $force
             ?? (
                 $precomp-stores
                 and my $unit = self!load-file($precomp-stores, $id, :refresh)
                 and do {
                     LEAVE $unit.close;
-                    nqp::sha1($path.slurp(:enc<iso-8859-1>)) eq $unit.source-checksum
+                     nqp::sha1($path.slurp(:enc<iso-8859-1>)) eq $unit.source-checksum
```
niner: Looks like an unintended white space change?

src/core.c/CompUnit/PrecompilationRepository.pm6
```diff
@@ -251,11 +259,10 @@ class CompUnit::PrecompilationRepository::Default does CompUnit::PrecompilationR
                 note "\n    load    $path.relative()";
                 $*ERR.flush;
             }
-            self.store.unlock;
+            #self.store.unlock;
```
niner: commented code
Shouldn't `$bcl` be unlocked and closed here?

src/core.c/CompUnit/PrecompilationRepository.pm6
```diff
@@ -321,7 +330,7 @@ class CompUnit::PrecompilationRepository::Default does CompUnit::PrecompilationR
 
         my @result is List = $out.lines.unique;
         if $status {  # something wrong
-            self.store.unlock;
+            #self.store.unlock;
```
niner: commented code
Shouldn't `$bcl` be unlocked and closed here?

src/core.c/CompUnit/PrecompilationRepository.pm6
```diff
@@ -332,7 +341,7 @@ class CompUnit::PrecompilationRepository::Default does CompUnit::PrecompilationR
         }
         unless $bc.e {
             $RMD("$path aborted precompilation without failure") if $RMD;
-            self.store.unlock;
+            #self.store.unlock;
```
niner: commented code
Shouldn't `$bcl` be unlocked and closed here?

src/core.c/CompUnit/PrecompilationRepository.pm6
```diff
@@ -281,6 +288,8 @@ class CompUnit::PrecompilationRepository::Default does CompUnit::PrecompilationR
         my $out = '';
         my $err = '';
         my $status;
+        my $bcl;
```
niner: Will need moving up, too

src/core.c/CompUnit/PrecompilationRepository.pm6
```diff
@@ -357,7 +366,8 @@ class CompUnit::PrecompilationRepository::Default does CompUnit::PrecompilationR
         );
         $bc.unlink;
         self.store.store-repo-id($compiler-id, $id, :repo-id($*REPO.id));
-        self.store.unlock;
+        $bcl.close;
+        #self.store.unlock;
```
niner: I guess closing the file will unlock it, too, but it's probably cleaner to explicitly unlock it.

src/core.c/CompUnit/PrecompilationRepository.pm6
Xliff: This is true, but that lock file might not yet exist. So locking the store, to create the file just made sense. I do agree that this is something that the store might be better at, however.

src/core.c/CompUnit/PrecompilationRepository.pm6
Xliff: Yes.

src/core.c/CompUnit/PrecompilationRepository.pm6
Xliff: Would that be better left to a LEAVE block?

src/core.c/CompUnit/PrecompilationRepository.pm6
Xliff: Oops! :)

src/core.c/CompUnit/PrecompilationRepository.pm6
Xliff: Sure thing.

src/core.c/CompUnit/PrecompilationRepository.pm6
Xliff: Done as a LEAVE block. See most recent commits.

niner:
> This is true, but that lock file might not yet exist. So locking the store,
> to create the file just made sense. I do agree that this is something that
> the store might be better at, however.

What happens if two processes try to open that same file? My guess would be 
that it'd just work anyway. The kernel will create the file for the first 
process and just open it for the second. Whichever one that will be. Taking 
the lock will decide who gets to precompile.

Xliff:
OK, so how about a CompUnit::PrecompilationUnit::File.get-file-lock?

Xliff:
So yeah... I went ahead and did that, since I was bored. Anything else?

ugexe:
Getting `make test` to work seems like a good idea

Xliff:
Yeah. That would be a good place to start, wouldn't it ? /o\

Xliff:
OK, this is odd. This is the output I get this when I run `make test`

```
t/01-sanity/55-use-trace.t ...................................... Dubious, test returned 1 (wstat 256, 0x100)
No subtests run
```

However, when I run the tests directly, they work fine! Here's the output from that same test run via `perl6 t/01-sanity/55-use-trace.t

```
1..3
ok 1 - did we get a Proc?
ok 2 - is the program output ok?
ok 3 - is the trace ok?
```

So what's the best way to track down the error?

vrurg:
> So what's the best way to track down the error?

First, I'd run it in a loop of 20-100 repetitions because a flapping is the primary suspect in such cases.

Xliff:
Same thing with t/53-transpose.t:

```
t/01-sanity/53-transpose.t ...................................... Dubious, test returned 1 (wstat 256, 0x100)
No subtests run
```

However it runs fine via `perl6`:

```
1..26
ok 1 - TRANSPOSE foof,f,a  -> aooa
ok 2 - TRANSPOSE foof,f,ab  -> abooab
ok 3 - TRANSPOSE foof,f,ff  -> ffooff
ok 4 - TRANSPOSE foof,f,  -> oo
ok 5 - TRANSPOSE offo,f,a  -> oaao
ok 6 - TRANSPOSE offo,f,ab  -> oababo
ok 7 - TRANSPOSE offo,f,ff  -> offffo
ok 8 - TRANSPOSE offo,f,  -> oo
ok 9 - TRANSPOSE foof,fo,a  -> aof
ok 10 - TRANSPOSE foof,fo,ab  -> abof
ok 11 - TRANSPOSE foof,fo,ff  -> ffof
ok 12 - TRANSPOSE foof,fo,  -> of
ok 13 - TRANSPOSE offo,fo,a  -> ofa
ok 14 - TRANSPOSE offo,fo,ab  -> ofab
ok 15 - TRANSPOSE offo,fo,ff  -> offf
ok 16 - TRANSPOSE offo,fo,  -> of
ok 17 - TRANSPOSE ofof,fo,a  -> oaf
ok 18 - TRANSPOSE ofof,fo,ab  -> oabf
ok 19 - TRANSPOSE ofof,fo,ff  -> offf
ok 20 - TRANSPOSE ofof,fo,  -> of
ok 21 - TRANSPOSE oooo,o,  -> 
ok 22 - TRANSPOSE oooo,o,x  -> xxxx
ok 23 - TRANSPOSE oooo,o,xx  -> xxxxxxxx
ok 24 - TRANSPOSE oooo,x,  -> oooo
ok 25 - TRANSPOSE oooo,x,y  -> oooo
ok 26 - TRANSPOSE oooo,x,yy  -> oooo
```

Xliff:
> > So what's the best way to track down the error?
> 
> First, I'd run it in a loop of 20-100 repetitions because a flapping is the primary suspect in such cases.

Done for both, and no error. It's not a flapper. Is it something dealing with how make is running it?

Xliff:
OK. Refreshed up to latest changes from master and (_hopefully_) fixes `make test`

niner:
`CompUnit::PrecompilationRepository::precompile` must not assume that the precomp store is a `CompUnit::PrecompilationStore::File` and not that the precomp unit is a `CompUnit::PrecompilationUnit::File`. It may only use what `CompUnit::PrecompilationStore`and `CompUnit::PrecompilationUnit` offer in terms of API.

When making changes or additions, try to imagine how to implement a CompUnit::PrecompilationStore::Memcached for example. That would be a useful thing to have and we should not prevent someone from implementing it.

Xliff:
OK, so let me propose this. Would it make sense to go back to the original way the code _looked_ ? I'm talking about how store.destination was used to obtain the IO::Path, and provided store.unlock, and this was the only thing PrecompilationUnit needed to deal with?

Xliff:
Odd... recent changes have thrown this error...

```
The iterator of this Seq is already in use/consumed by another Seq
(you might solve this by adding .cache on usages of the Seq, or
```

And I'm not doing anything with Seq objects...

Xliff:
Closed in favor of #3785

PR #3458, opened 2020-02-01
Releaseable tweaks

Altai-man:
SSH is not configurable via envvars. We have plenty of them already and it won't hurt to add some more, but makes the script more robust.
Also update tarball distributors to use a working one.

### DIFF
--- a/tools/releasable/Sakefile
+++ b/tools/releasable/Sakefile
@@ -23,8 +23,8 @@ use File::Directory::Tree;
 my $THREADS  = 7;
 
 # These are used so that you can run tests on a faster machine.
-my $SSH      = ‘alex@localhost’; # TODO make it configurable
-my $SSH-PATH = ‘/home/alex/git/Releasable/’; # TODO make it configurable
+my $SSH      = %*ENV<RELEASE_SSH> // "{qx{whoami}.trim}@localhost";
+my $SSH-PATH = %*ENV<RELEASE_HOME>;
 my $RSYNC    = “$SSH:$SSH-PATH”;
 
 my $WIKIPEDIA-NAME = %*ENV<WIKIPEDIA-NAME>; # currently not used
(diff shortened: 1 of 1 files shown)


tools/releasable/Sakefile
```diff
@@ -210,14 +210,14 @@ task ‘rakudo-push’, { # Step ??? (rakudo release guide)
...
-        run ‘scp’, $nqp-archive, “{$nqp-archive}.asc”, “$_/nqp/”
+        run ‘scp -p 2222’, $nqp-archive, “{$nqp-archive}.asc”, “$_/nqp/”
     }
     True
 }
 
 task ‘rakudo-upload’, { # Step ⓲ (rakudo release guide)
     for @tarball-distributors {
-        run ‘scp’, $rakudo-archive, “{$rakudo-archive}.asc”, “$_/rakudo/”
+        run ‘scp -p 2222’, $rakudo-archive, “{$rakudo-archive}.asc”, “$_/rakudo/”
```
AlexDaniel: That's not how `run` works. This won't chooch. It needs to be `run <scp -p 2222>`.

tools/releasable/Sakefile
```diff
@@ -23,8 +23,8 @@ use File::Directory::Tree;
 my $THREADS  = 7;
 
 # These are used so that you can run tests on a faster machine.
-my $SSH      = ‘alex@localhost’; # TODO make it configurable
-my $SSH-PATH = ‘/home/alex/git/Releasable/’; # TODO make it configurable
+my $SSH      = %*ENV<RELEASE_SSH> // "{qx{whoami}.trim}@localhost";
+my $SSH-PATH = %*ENV<RELEASE_HOME>;
```
AlexDaniel: This might work, but I have a feeling that I already had it like this and it didn't. Once it ssh's into a different machine, the ENV won't be propagated. Can you test it that it works when using two different machines?

Altai-man: Sadly, I can't as I have only a single laptop available around.
Also, if this is bad, I'd rather looked into making the whole ssh layer optional because it, in my opinion, ought to be so, but decided to remove some hardcoded values firstly.

PR #3463, opened 2020-02-07
Box result of nqp::time_n() in DateTime.now

dumarchie:
This is a workaround for /issues/3459.

On Windows, `DateTime.new` multi dispatch may break if the first argument is a native `num`. I have no idea why this appears to be an issue on Windows only.


dumarchie:
It's probably better for someone to fix the root issue with `num`.

Issue #2364, opened 2018-10-13
throws-like should die on Bool:D matchers


zoffixznet:
This is following https://github.com/rakudo/rakudo/pull/2363 that got closed...

```
<Zoffix_> m: use Test; class X::Foo is Exception { has $.meow = 42 }; throws-like { die X::Foo.new }, X::Foo, :meow
<camelia> rakudo-moar 2c43d2bf5: OUTPUT: «    1..3␤    ok 1 - code dies␤    ok 2 - right exception type (X::Foo)␤    ok 3 - .meow matches True␤ok 1 - did we throws-like X::Foo?␤»
```

Above, the test writer specified they expect `$.meow` to contain Bool:D value `True`. However, since we do smartmatching on these matchers, regardless of the value, the check will always succeed (or always fail, if we're testing for `False`).

----------------------------------

*Quoted from https://github.com/rakudo/rakudo/pull/2363:*

The named matchers in throws-like are smartmatched against the received Exception's methods and smartmatch against False is always False. To properly test for a Bool:D value you'd use (for example) * === False or * === True WhateverCode.

I think the change that should be made is to watch for Bool:D matchers and throw, the same as how we do with .grep for example:

$ perl6 -e '().grep: True'
Cannot use Bool as Matcher with '.grep'.  Did you mean to use $_ inside a block?
  in block <unit> at -e line 1
and perhaps improve the message of that exception (X::Match::Bool) to mention * === True / * === False WhateverCodes as well.

(N.B.: I don't really want it to do === test automatically for Bool:D matchers, as that breaks expectations of how smartmatch works elsewhere; best to consistently re-enforce that message)

--------------------

- [*Newcomer Guide to Contributing to Core Perl 6*](https://rakudo.party/post/Newcomer-Guide-to-Contributing-to-Core-Perl-6)
- Location of where the fix would go [lib/Test.pm6#L612-L620](https://github.com/rakudo/rakudo/blob/2c43d2bf5f3e2da5488eb9299ac9c6bbc454da55/lib/Test.pm6#L612-L620)

vrurg:
Implemented with #3464 and #3473

PR #3464, opened 2020-02-07
Make throws-like throw an exception on Bool:D matchers

atroxaper:
Now, throws-like and fails-like Test subs throw an Exception in case the matcher uses Bool:D as matcher value instead of explicit Bool:D check. The behavior is similar to .grep(True).

Changes for issue: #2364
New tests: https://github.com/perl6/roast/pull/616

### DIFF
--- a/lib/Test.rakumod
+++ b/lib/Test.rakumod
@@ -603,7 +603,7 @@ multi sub is-deeply(Mu $got, Mu $expected, $reason = '') is export {
 
 sub throws-like($code, $ex_type, $reason?, *%matcher) is export {
     for %matcher.kv -> $k, $v {
-        if $v === True|False {
+        if $v ~~ Bool and $v.DEFINITE {
             X::Match::Bool.new(:type(".$k")).throw;
         }
     }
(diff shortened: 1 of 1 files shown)


lib/Test.rakumod
```diff
@@ -602,6 +602,11 @@ multi sub is-deeply(Mu $got, Mu $expected, $reason = '') is export {
 }
 
 sub throws-like($code, $ex_type, $reason?, *%matcher) is export {
+    for %matcher.kv -> $k, $v {
+        if $v === True|False {
```
vrurg: Considering that `grep` is utilizing multi-dispatch for detecting a `Bool:D` matcher, I would rather suggest using smartmatching for consistency. For example, for readability one defines something like:

```
enum Answer (No => False, Yes => True);
say Yes ~~ Bool:D; # True
say Yes === True;  # False
```

vrurg:
I also don't like the error message, but that same complaint applies to `grep` too, so not really relevant. Just a note to think about a better one to myself or anybody else interested.

lib/Test.rakumod
atroxaper: I agreed. But I need to check like `if $v ~~ Bool and $v.DEFINITE` because `say Bool ~~ Bool:D; # True`

atroxaper:
@vrurg Vadim, please take a look my last changes.

vrurg:
@atroxaper Looks ok, but I want to try it alongside with the tests and it'd take time which I don't have much as of lately. Hopefully, I'd just merge it tomorrow or tonight by EST time.

vrurg:
@atroxaper Thank you!

atroxaper:
@vrurg Thank you!

PR #3481, opened 2020-02-14
Renaming of perl6 to rakudo

vrurg:
- `rakudo` binaries become the default, `perl6` are aliases.

- Changed config and makefile variable names where possible from perl6
(case considered) to Raku or rakudo, depending on context. I.e.  where
it goes about binaries – we say `rakudo`; `Raku` where it's about the
language.

- Some variables in scripts were renamed too.

- `harness6` `--perlpath` changed to `--rakupath`.

- `perl6.jar` became `rakudo.jar` though Java class is still `perl6`

- Switching over from `PERL6LIB` to `RAKUDOLIB`


vrurg:
Not tested yet. In either case, won't merge without proper reviews.

@lizmat Pls, don't merge accidentally! Just would like to remind that it's still *WIP*!

vrurg:
@patrickbkr can you consider renaming `perl6_ops_moar`? Not sure if it's not used outside of `rakudo` installation thus hesitate about renaming it.

vrurg:
@pmurias Can we do something about `org.perl6`, `org/perl6` and everything related? I'm not speaking Java and definitely have no idea what implications could be when the class is renamed.

Also, JS doesn't build for me, thus any changes for that are purely theoretical and most likely won't work.

Altai-man:
1)Can you make a list of externally visible changes caused by this?
2)Can you note measures done to preserve backward compatibility, e.g. for existing scripts with shebangs, executable names etc?

ugexe:
Rakudolib is not an alternative to perl6lib

patrickbkr:
If I remember a recent discussion right, there was some consensus, that we should introduce the RAKULIB environment variable as the successor to PERL6LIB and use that. Reasoning: The functionality to add a folder to the library search paths is not Rakudo specific, but a thing all Raku implementations should support. It's also (maybe only implicitly) part of the spectests. Thus it should be called RAKULIB.
The RAKUDOLIB variable predates the rename by a long time. It's existence does not mean it's the successor of PERL6LIB.

vrurg:
@patrickbkr where was `RAKULIB` discussed? My memory betrays me on this, neither has search came up with any results.

Anyway, this change is just two lines in in scripts where they export environment variable for a subprocess. This PR is not supposed to change any internal logic.

vrurg:
@Altai-man No worries about the backward compat as `perl6` is preserved. Same applies  to various `perl6-*` executables which are currently available alongside with `rakudo-*` defaults. Sure, `raku` and `raku-debug` symlinks are in place as they were since last November.

On JVM side `perl6-*-server` were renamed to `rakudo-*-server` and this is where I need somebody's voice as I've no idea how where these are used.

Nothing else I can think of what would be visible for an outside observer. Most other changes are about internal names like makefile and config variables.

What's been tested so far:

- [ ] MacOS:
  - [ ] test, spectest
  - [ ] local install
  - [ ] zef run
- [ ] Linux Mint VM
   - [ ] test, spectest
   - [ ] local install 
   - [ ] zef run
- [ ] Windows
   - [ ] test,spectest - failing, but same tests failing on master
   - [ ] local install
   - [ ] zef run

vrurg:
All four platforms at my disposal passing basic tests. `RAKUDOLIB` has been taken care of and reverted back to `PERL6LIB` for now, until we get RAKULIB in place.

My only doubt left is about JVM's `raku-eval-server`. @pmurias?

Issue #3483, opened 2020-02-14
Compiler error when using ::?CLASS probably a bit too much


Altreus:
## The Problem

An unlikely combination of four things causes an unintelligible error.

* A Role with a multi method
* One part of which is unimplemented (`...`)
* Using  `::?CLASS` to define the invocant
* Using `returns ::?CLASS` as well

Causes the compilation error:

`No such method 'raku' for invocant of type '::?CLASS'`

Any of these not being the case allows compilation with no problem.

The rationale behind these four aspects of the role is to allow a method with a class-level implementation as well as an object-level one. Any class consuming the role gets a default behaviour for the class-level implementation, which is to instantiate an object and run the other one.

Strictly, it is not necessary to use `returns ::?CLASS` but I wanted to be specific that they should both return an object of the consuming class's type.

[Example](https://github.com/shuppet/p6-api-discord/blob/8cc714c8ae9d905d35d2c76005312a17a7563874/lib/API/Discord/HTTPResource.pm6#L33):

```raku
role JSONy {
    multi method from-json (::?CLASS:U: $json) returns ::?CLASS {
        self.new.from-json($json);
    }

    multi method from-json (::?CLASS:D: $json) returns ::?CLASS { ... }
}
```

```
> class A does JSONy {}
===SORRY!=== Error while compiling:
No such method 'raku' for invocant of type '::?CLASS'
```

## Environment
```
$ raku -v
This is Rakudo version 2020.01 built on MoarVM version 2020.01.1
implementing Perl 6.d.

$ uname -a
Linux adouglas 4.15.0-74-generic #84-Ubuntu SMP Thu Dec 19 08:06:28 UTC 2019 x86_64 x86_64 x86_64 GNU/Linux

$ lsb_release -a
No LSB modules are available.
Distributor ID:	LinuxMint
Description:	Linux Mint 19.1 Tessa
Release:	19.1
Codename:	tessa
```

Altreus:
I also tried `returns ::?CLASS:D` since that's more specifically what I wanted; same error.

vrurg:
Actually, use of `::?CLASS` in a role not such a good idea in general because when a role is being defined the symbol is generic any any kind of issues are possible when the role is being introspected. Such as in this case, the actual error comes from an attempt to report another error where exception text is being formed using stringification of method's signature.

The immediate workaround would be to use `::?ROLE` instead.

Altreus:
[...]

PR #3484, opened 2020-02-16
Properly support generic types in Signature 'returns' attribute.

vrurg:
First, stringification must not use `raku` method on the type in `$!returns` if the type doesn't have the method. Fixes LTA error message as reported by #3483.

Also, instantiate a return type in `Signature`'s `instantiate_generic` method.


vrurg:
I have mentioned this in the ticket itself, but would like this info to stay here too.

A prototype like `method foo(--> ::?CLASS:D) {...}` can be fulfilled with `method foo(--> ::?CLASS) { }` because definedness of a generic return type is getting lost somewhere if it was there ever in first place. Unfortunately, I don't have time to hunt for this one for now.

PR #3487, opened 2020-02-16
Implement support for using the `is cached` trait with methods

Kaiepi:
By replacing the routine/method's `$!do` attribute with its wrapper's
instead of using `Routine.wrap` (think NativeCall), multi methods are now
possible to make cached, making method support possible to add to the
`is cached` trait.

### DIFF
--- a/lib/experimental.rakumod
+++ b/lib/experimental.rakumod
@@ -1,18 +1,54 @@
 use nqp;
 
 package EXPORT::cached {
-    multi sub trait_mod:<is>(Routine $r, :$cached!) {
-        my %cache;
-        $r.wrap(-> |c {
-            my $key := c.gist;
-            %cache.EXISTS-KEY($key)
-              ?? %cache{$key}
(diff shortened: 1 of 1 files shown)


lizmat:
Thank you for this PR.  Please note that before Christmas 2015, it was decided to make the `is cached` trait experimental, because the implementation didn't really do what @Larry thought it should do.  Since code was depending on that then, it was decided to not remove it.  Which is why I haven't done any other work on this.

Even though this would be easy to merge, I would advise against this, at least until we've had a problem-solving discussion on whether we want to continue having the `is cached` trait in core: if we do, we should probably remove the "experimental` status from it for 6.e.

Kaiepi:
Besides what @lizmat raised, this doesn't handle the outer context of the wrapped routine properly, so don't merge until that's fixed as well.

Kaiepi:
Closing this. Will reopen once routine dispatch is sorted out.

PR #3489, opened 2020-02-17
Allow debuggers to be specified by environment variables

Kaiepi:
This makes it so gdb/lldb/valgrind are used by default, but may be
overridden by RAKUDO_GDB/RAKUDO_LLDB/RAKUDO_VALGRIND environment
variables.

### DIFF
--- a/tools/templates/moar/rakudo-gdb-m.in
+++ b/tools/templates/moar/rakudo-gdb-m.in
@@ -5,5 +5,13 @@
 GDB="gdb"
 if [ -n "$RAKUDO_GDB" ]; then
     GDB="$RAKUDO_GDB"
+else
+    case "$(uname)" in
+        OpenBSD)
+            if which egdb >/dev/null; then
+                GDB="egdb"
+            fi
+    esac
 fi
+
 "$GDB" --quiet --ex=run --args "@expand(@MOAR@)@" @expand(@runner_opts@)@ "$@"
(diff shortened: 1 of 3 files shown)


vrurg:
What's use of it? What problem does it resolve?

Kaiepi:
On OpenBSD, the version of `gdb` it comes with is ancient, and the version you install from packages/ports is called `egdb` instead, so `rakudo-gdb-m` sometimes doesn't work like it should.

vrurg:
Then I'd propose to combine this with a more user-friendly solution where the default value would depend on `uname` output. For OpenBSD it would be find to test for the presence of egdb then in first place and only set to `gdb` as a fallback. Something like (roughly):

```
case $OSNAME in
    OpenBSD)
        if which egdb >/dev/null; then GDB=egdb; else GDB=gdb; fi
    *)
        GDB=gdb
esac
```

Alongside with `RAKUDO_GDB` this would provide both a comfortable default and flexibility.

vrurg:
My own comment terrifies me... What was I actually writing???

Anyway, if you plan no more changes then may I merge it?

JJ:
> My own comment terrifies me... What was I actually writing???
> 
> Anyway, if you plan no more changes then may I merge it?

You probably should...

niner:
Is rakudo-gdb-m actually still being used? These days I just use `gdb --args raku` as raku is a proper executable and things just work. And usually I have a command to debug in my history anyway, so I'm faster by jumping to the start of the line and typing `gdb --args `

jnthn:
> Is rakudo-gdb-m actually still being used? 

I am, at least, perhaps from muscle memory as much as anything.

PR #3491, opened 2020-02-17
Add a $.path attribute and PF_UNIX multis to IO::Socket::INET

Kaiepi:
The methods and attributes of `IO::Socket::INET` are currently geared more towards IPv4/IPv6 addresses. This makes it so UNIX sockets have more appropriate methods and attributes WRT their addresses.

### DIFF
--- a/src/core.c/IO/Socket/INET.pm6
+++ b/src/core.c/IO/Socket/INET.pm6
@@ -13,15 +13,16 @@ my class IO::Socket::INET does IO::Socket {
         constant MAX_PORT       = 65_535; # RFC 793: TCP/UDP port limit
     }
 
-    has Str  $.host;
-    has Int  $.port;
-    has Str  $.localhost;
-    has Int  $.localport;
-    has Int  $.backlog;
-    has Bool $.listening;
-    has      $.family     = nqp::const::SOCKET_FAMILY_UNSPEC;
-    has      $.proto      = PIO::PROTO_TCP;
-    has      $.type       = PIO::SOCK_STREAM;
+    has Str      $.host;
+    has Int      $.port;
+    has Str      $.localhost;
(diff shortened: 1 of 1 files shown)


ugexe:
Do other language implementations mix INET and UNIX? For example: perl has IO::Socket::INET and IO::Socket::UNIX

Kaiepi:
Python's `socket`, Haskell's `Network.Socket`, .NET's `System.Net.Sockets` are some examples of modules similar to `IO::Socket::INET` that mix UNIX and IP socket support.

I see `IO::Socket::INET` as being a lower level API for working with sockets than `IO::Socket::Async`, rather than just being a synchronous API for them. Outside of how their address is handled and the socket is initialized, there isn't much difference between how UNIX and IP sockets are used when used like sockets. Combined with a solution to https://github.com/Raku/problem-solving/issues/111 (which is something I want to deal with after this, before implementing sync UDP socket support), the differing state between different families of sockets would be handled by their address, not `IO::Socket::INET` itself. The name becomes a misnomer for what it's doing though...

Kaiepi:
With the solution for https://github.com/Raku/problem-solving/issues/111, it becomes clearer that support for UNIX sockets doesn't belong in `IO::Socket::INET`. I was hoping to eventually make this class handle all synchronous sockets as `IO::Socket::Sync`, but due to it being designed with IPv4 sockets in mind, this complicates its `listen` and `connect` methods significantly. `IO::Socket::UNIX` would be a more fitting place to handle UNIX socket support, but that needs to wait until the problem solving issue is solved.

lizmat:
Perhaps we should be more thinking about roles in this context?

Kaiepi:
How so? There are multiple ways they could get involved here.

PR #3496, opened 2020-02-18
Set attribute's rw status when added to an AttributeContainer

### LINKED ISSUE #3495: `is rw` trait applied to roles does not work
## The Problem
`is rw` trait applied to roles does not work

## Expected Behavior
```
> role X is rw { has $.x; }; (X.new.x = 4).say
4
```

## Actual Behavior
```
> role X is rw { has $.x; }; (X.new.x = 4).say
Cannot modify an immutable 'Any' type object
```

## Environment
rakudo-moar 8c995febd
camelia on freenode

vrurg:
It was previously set in `compose_attributes` which is called by
`ClassHOW` only. This was causing roles always ignoring the default `rw`
status and thus `is rw` trait not having effect on them. This fix
presumes that the trait is always applied before attributes are thrown
in.

Fixes #3495

### DIFF
--- a/src/Perl6/Metamodel/AttributeContainer.nqp
+++ b/src/Perl6/Metamodel/AttributeContainer.nqp
@@ -50,8 +50,13 @@ role Perl6::Metamodel::AttributeContainer {
     }
 
     # Makes setting the type represented by the meta-object rw mean that its
-    # attributes are rw by default.
+    # attributes are rw by default. For cases when status is late set, like
+    # with 'also is rw', fixup the previously added attributes. Note that we
+    # can safely use 'default_to_rw' because it would pay respect to `is readonly`
     method set_rw($obj) {
+        for @!attributes {
+            $_.default_to_rw();
+        }
         $!attr_rw_by_default := 1;
     }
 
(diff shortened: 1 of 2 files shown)


jnthn:
I think this will cause a change of behavior in `class C { has $.a; also is rw; }`.

jnthn:
(Without judgement on whether the change is an improvement or not... :-))

vrurg:
`also`, yes... There must be a spectest for this case.

vrurg:
Ok, it is solved by setting the `default_to_rw` for any attributes added prior to `also is rw`. I like it this way because:

- it keeps all logic in the same place, `AttributeContainer` role
- it's more performant than doing separate explicit run over all the attributes elsewhere; not sure if adding `if +@!attributes` around the loop would make it faster than the `for` alone for cases when no attributes were added yet
- it should be done outside of `compose_attributes` because role composition doesn't and shouldn't use the method. Introduction of a new method specifically for role composition or implementing the logic in the role is not reasonable
- the solution should easily extend to new or custom type objects would they be using `AttributeContainer`

PR #3500, opened 2020-02-21
Fix dispatchers and make them interact and chain properly, reconsider wrapping

### LINKED ISSUE #3499: multi method dispatch doesn't work with inheritance.
## The Problem

When both parent and child classes define `proto` calling `nextsame` in a child `multi` method doesn't dispatch to parent even if child's dispatcher exhausted.

## Expected Behavior

Must dispatch.

## Steps to Reproduce

```perl6
class Foo {
    proto method foo(|) {*}
    multi method foo(Str:D $s) { note "Foo Str:D"; nextsame }
    multi method foo(Any:D) { note "Foo Any:D"; nextsame }
}
class Bar is Foo {
    proto method foo(|) {*}
    multi method foo(Str $s) { note "Bar Str"; nextsame }
    multi method foo(Any) { note "Bar Any"; nextsame }
}
Bar.new.foo("foo");
```

This code outputs:

```
Bar Str
Bar Any
```

instead of expected:

```
Bar Str
Bar Any
Foo Str:D
Foo Any:D
```

vrurg:
A fix turned into a next revamp. This time it's about dispatchers and wrapping.

- Dispatchers now chain to each other. I.e. it is possible to wrap a
  candidate of a multi-dispatch method in the middle of inheritance chain.
  In turn, the wrapping routine can be a multi on its own and its
  candidates can be wrapped too, etc., etc.
- Wrapping is done using manipulations with `$!do` attribute. Whan a
  routine is completely unwrapped it's no different from its original
  state. Avoidance of use of `CALL-ME` invocation also fixes a few cases
  of 'Cannot invoke object with invocation handler in this context' error.
- `Routine` got attribute and accessor `wrappers` which is initialized
  with a list of routine wrappers, including a clone of the original
  routine itself.
- `Routine` got method `is_wrapped`.

Fixes #3499

### DIFF
--- a/src/Perl6/Metamodel/Dispatchers.nqp
+++ b/src/Perl6/Metamodel/Dispatchers.nqp
@@ -3,68 +3,115 @@ class Perl6::Metamodel::BaseDispatcher {
     has $!idx;
     has $!next_dispatcher; # The dispatcher we must pass control to when own queue exhausts
 
-    method candidates() { @!candidates }
+    method candidates()     { @!candidates }
 
-    method exhausted() { $!idx >= +@!candidates && (!nqp::isconcrete($!next_dispatcher) || $!next_dispatcher.exhausted()) }
+    method exhausted()      { $!idx >= +@!candidates && (!nqp::isconcrete($!next_dispatcher) || $!next_dispatcher.exhausted()) }
 
-    method last()      { @!candidates := [] }
+    method last_candidate() { $!idx >= +@!candidates }
 
-    method set_next_dispatcher($next_dispatcher) { $!next_dispatcher := $next_dispatcher }
+    method last()           { @!candidates := [] }
+
+    method set_next_dispatcher($next_dispatcher)
+                            { $!next_dispatcher := $next_dispatcher }
 
     # Wrapper-like dispatchers don't set dispatcher for the last candidate.
     method is_wrapper_like() { 0 }
 
     method get_call() { # Returns [$call, $is_dispatcher]
-        my $call := @!candidates[$!idx++];
+        my $call := @!candidates[$!idx];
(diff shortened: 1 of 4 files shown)


vrurg:
@jnthn Does it make sense to you? This PR makes the following work as I would expect it:


```perl6
use v6;

class Foo {
    proto method foo(|) {*}
    multi method foo(Str:D $s) { note "Foo Str:D"; nextsame }
    multi method foo(Int:D $i) { note "Foo Int:D"; nextsame }
    multi method foo(Any:D) { note "Foo Any:D"; nextsame }
}

class Bar {
    proto method foo(|) {*}
    multi method foo(Str $s) { note "Bar Str"; nextsame }
    multi method foo(Any) { note "Bar Any"; nextsame }
}

class Fubar {
    proto method foo(|) {*}
    multi method foo(|) { note "Fubar?"; nextsame }
}

class FooBar {
    proto method foo(|) {*}
    multi method foo(|) { note "FooBar!"; nextsame }
}

class C { method foo(|) { note "C foo"; nextsame } }

class Baz is C is Foo is Bar is Fubar is FooBar
{
    proto method foo(|) {*}
    multi method foo(Any) { note "Baz Any"; nextsame }
    multi method foo(Bool:D) { note "Baz Bool"; nextsame }
}

note "STR : ", Baz.new.foo("aa");
note "42  : ", Baz.new.foo(42);
note "True: ", Baz.new.foo(True);

note Baz.^method_table<foo>.candidates.map(*.raku).join("\n");
```

vrurg:
As I was thinking of it, the fix may not work correctly for when the child class overrides the method as plain. Looks like the whole method dispatch needs reconsidering...

PR #3502, opened 2020-02-21
Make listing and coercion of sequences more efficient

dumarchie:
Let `.list` return the cached list if it exists and base the `.List` and `.Slip` coercers on `.list`.

### DIFF
--- a/src/core.c/Sequence.pm6
+++ b/src/core.c/Sequence.pm6
@@ -44,8 +44,8 @@ Routine.'!configure_positional_bind_failover'(Positional, PositionalBindFailover
 
 my role Sequence does PositionalBindFailover {
     multi method Array(::?CLASS:D:) { Array.from-iterator(self.iterator) }
-    multi method List(::?CLASS:D:)  { List.from-iterator(self.iterator) }
-    multi method Slip(::?CLASS:D:)  { Slip.from-iterator(self.iterator) }
+    multi method List(::?CLASS:D:)  { self.list.List }
+    multi method Slip(::?CLASS:D:)  { self.list.Slip }
 
     multi method Str(::?CLASS:D:) {
         self.cache.Str
--- a/src/core.c/Sequence.pm6
+++ b/src/core.c/Sequence.pm6
@@ -28,8 +28,13 @@ my role PositionalBindFailover {
           ($!list := List.from-iterator(self.iterator))
         )
     }
+
     multi method list(::?CLASS:D:) {
-        List.from-iterator(self.iterator)
+        nqp::if(
+          nqp::isconcrete($!list),
+          $!list,
+          List.from-iterator(self.iterator)
+        )
     }
 
     method iterator() { ... }


src/core.c/Sequence.pm6
```diff
@@ -28,8 +28,13 @@ my role PositionalBindFailover {
           ($!list := List.from-iterator(self.iterator))
         )
     }
+
     multi method list(::?CLASS:D:) {
-        List.from-iterator(self.iterator)
+        nqp::if(
+          nqp::isconcrete($!list),
+          $!list,
+          List.from-iterator(self.iterator)
```
lizmat: Shouldn't that need to set `$!list` ?

dumarchie: No, that's what `.cache` does.

src/core.c/Sequence.pm6
```diff
@@ -38,9 +43,9 @@ nqp::p6configposbindfailover(Positional, PositionalBindFailover); # Binder
 Routine.'!configure_positional_bind_failover'(Positional, PositionalBindFailover); # Multi-dispatch
 
 my role Sequence does PositionalBindFailover {
-    multi method Array(::?CLASS:D:) { Array.from-iterator(self.iterator) }
-    multi method List(::?CLASS:D:)  { List.from-iterator(self.iterator) }
-    multi method Slip(::?CLASS:D:)  { Slip.from-iterator(self.iterator) }
+    multi method Array(::?CLASS:D:) { self.list.Array }
```
jnthn: I think in the case there is no cached list already, this would double roughly memory use, since it will need to reify the elements in a `List` as well as put them into `Scalar` containers in the `Array`?

jnthn: (Probably this should be done conditional on there being a cached list, and otherwise just do what was done before.)

dumarchie: You may have a point there. Let me review both the lazy and non-lazy cases.

src/core.c/Sequence.pm6
jnthn: Correct, only `.cache` should set the `list`, this is just using the cached list if it exists.

jnthn: (Which I think we can safely get away with.)

src/core.c/Sequence.pm6
dumarchie: I think you're right. I'll revert that change as `self.iterator` will be based on the cached list if it exists.

src/core.c/Sequence.pm6
```diff
@@ -39,8 +44,8 @@ Routine.'!configure_positional_bind_failover'(Positional, PositionalBindFailover
 
 my role Sequence does PositionalBindFailover {
     multi method Array(::?CLASS:D:) { Array.from-iterator(self.iterator) }
-    multi method List(::?CLASS:D:)  { List.from-iterator(self.iterator) }
-    multi method Slip(::?CLASS:D:)  { Slip.from-iterator(self.iterator) }
+    multi method List(::?CLASS:D:)  { self.list.List }
```
lizmat: Feels to me the `.List` is a noop at that stage?

lizmat: I would actually propose to make the `List` method leading, and `list` be a frontend for `List`.  Then, copy the contents of the `List` method, and change `$!list` to `p6bindattrvinres(nqp::create(Slip),List,'$!list`)` and `List.from-iterator` to `Slip.from-iterator`.

dumarchie: @lizmat, `.list.List` is indeed equivalent to `.list.self` _if_ `.list` returns a `List`. However, a `PositionalBindFailover` implementation may define `.list` to return a different type of list in which case `.List` is required to actually coerce the return value to `List`.

Regarding your suggestions, I consider them to be outside the scope of this PR as I just want to avoid unnecessary copying of `List`s. But I do have an opinion about them.

As I understand it, `PositionalBindFailover` is intended to be a minimalistic role that can be used by classes outside the core. The `.list` method is part of it because it's the method called by many built-ins that require a list-like representation of an object. Examples of such built-ins are `Any`'s `.List`, `.Slip` and `.Array` coercers. In fact, the `.Slip` implementation provided by `Sequence` is now exactly the same as the one provided by `Any`. I don't think that means the implementation can be removed from `Sequence`, because role `Sequence` does not carry the `Any` class.

Regarding the copying of implementations, I think such a violation of DRY is permissible only if it has noticeable performance benefits in real-world use cases. I've decided to keep identifying use cases and measuring performance outside the scope of this PR.

PR #3511, opened 2020-02-24
RFC: Change PositionalBindFailover MRO

dumarchie:
To me it makes sense for `PositionalBindFailover` to extend `Iterable` because both require an implementation of `method iterator()`.

This PR is a proof-of-concept to invite discussion. If there are good arguments against this change I will document them.

### DIFF
--- a/src/core.c/HyperSeq.pm6
+++ b/src/core.c/HyperSeq.pm6
@@ -1,7 +1,7 @@
 # A HyperSeq performs batches of work in parallel, but retains order of output
 # values relative to input values.
 #?if !js
-my class HyperSeq does Iterable does Sequence {
+my class HyperSeq does Sequence {
     has HyperConfiguration $.configuration;
     has Rakudo::Internals::HyperWorkStage $!work-stage-head;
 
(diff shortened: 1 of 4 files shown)


JJ:
Is there any impact on performance?

niner:
The comment above the PositionalBindFailover role reads:
```
# The actual memoization functionality is factored out into a role,
# PositionalBindFailover, which is used by the binder to identify types that,
# on failure to bind to an @-sigilled thing, can have .cache called on them
# and get memoization semantics. This decouples this functionality from the
# Sequence role, so other user-defined types can get access to this
# functionality.
```

Having PositionalBindFailover do Iterable already, narrows down its use to all types that are compatible (in spirit) with Iterable, i.e. if you want the positional bind failover behaviour you'd always get all of Iterable's (like itemization and flattening) as well, whether you want to or not. This reduces PositionalBindFailover's utility which makes me inclined to refuse this change.

On the other hand though the comment in Iterable claims "Iterable is done by anything that we should be able to get an iterator from" which actually does apply to PositionalBindFailover, since it requires an iterator method. This would suggest that this change is doing the right thing.

However in summary, I still lean towards leaving things as they are as in general, having smaller, independent building blocks that can be combined easily usually turns out to be more useful in the end. After all it's much easier to have a type do two roles than have it only do a part of one.

lizmat:
Given the lack of discussion, I'm closing this PR

Issue #1882, opened 2018-05-29
run(..., :$out, :merge) does not redirect stderr to the $out file handle


kjkuan:
***See [Newcomer Guide to Contributing to Core Perl 6](https://rakudo.party/post/Newcomer-Guide-to-Contributing-to-Core-Perl-6) for tips on fixing this Issue***

----------------

## The Problem
When using `run()` with stdout redirected to a file handle object and with the `:merge` option, stderr doesn't seem to be redirected to the file handle.

## Steps to Reproduce
```
my $out = './x.out'.IO.open(:w);
my $p = run <<ls -la adfadsf>>, :out($out), :merge; 
$out.close; 
```
The file `./x.out` should, but does not contain,  the error message from the `ls` command.

## Environment
* Operating system: MacOS 10.13.4
* Compiler version (`perl6 -v`): 
This is Rakudo version 2018.05 built on MoarVM version 2018.05
implementing Perl 6.c.

zoffixznet:
A bit of [chat log](https://irclog.perlgeek.de/perl6/2018-05-29#i_16221870):

> 18:27 	Zoffix 	Briefly scanning the code, looks like the pipe-to-IO::Pipe option is handled in the same conditional tree as merge, and merge comes first. Should be a simple fix.
> 18:27 		The merge being handled here: https://github.com/rakudo/rakudo/blob/master/src/core/Proc.pm6#L51-L62 while piping is up here, in the `else` branch from merge: https://github.com/rakudo/rakudo/blob/master/src/core/Proc.pm6#L90-L92

lizmat:
@atroxaper thanks again, also for the nudge  :-)

PR #3513, opened 2020-02-26
Fix run() with :out(Handle:D) and :merge

atroxaper:
Previously, the run routine with :merge and :out(Handle:D) arguments did not
write output to the :out argument. Now both stderr and stdout will be written
to the :out argument in such case.

Changes for: #1882
New tests: https://github.com/perl6/roast/pull/622

### DIFF
--- a/src/core.c/Proc.pm6
+++ b/src/core.c/Proc.pm6
@@ -49,16 +49,22 @@ my class Proc {
         }
 
         if $merge {
-            my $chan = Channel.new;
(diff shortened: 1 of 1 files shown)


lizmat:
Was this close intentional?  The patch seems ok?

atroxaper:
@lizmat Hi. Yes, I forget to change user.email :(
I made another pull request https://github.com/rakudo/rakudo/pull/3514

PR #3516, opened 2020-02-26
Enable env var RAKULIB

tbrowder:
Enable the compiler to respect the new environment variable RAKULIB for module paths to supersede the existing environment variable PERL6LIB.

Change is required for the name change from Perl 6 to Raku.

The existing test for the change is ready to be comited to roast and it was used to successfully test this change. See roast PR #625.

### DIFF
--- a/src/Perl6/Compiler.nqp
+++ b/src/Perl6/Compiler.nqp
@@ -174,18 +174,20 @@ Note that only boolean single-letter options may be bundled.
 The following environment variables are respected:
 
   RAKUDOLIB   Modify the module search path
-  PERL6LIB    Modify the module search path
+  RAKULIB     Modify the module search path
+  PERL6LIB    Modify the module search path # to be deprecated
   RAKUDO_HOME Override the path of the Rakudo runtime files
-  PERL6_HOME  Override the path of the Rakudo runtime files
+  RAKU_HOME   Override the path of the Rakudo runtime files
+  PERL6_HOME  Override the path of the Rakudo runtime files # to be deprecated
   NQP_HOME    Override the path of the NQP runtime files
 
 ♥); # end of usage statement
 
         nqp::exit(0);
 
-        # TODO: create and install a man page for Perl 6; then add the following
+        # TODO: create and install a man page for Raku; then add the following
         #       line to the end of the usage text above:
         #
-        #  For more information, see the perl6(1) man page.\n");
+        #  For more information, see the raku(1) man page.\n");
     }
 }
--- a/src/Perl6/Compiler.nqp
+++ b/src/Perl6/Compiler.nqp
@@ -177,8 +177,6 @@ The following environment variables are respected:
   RAKULIB     Modify the module search path
   PERL6LIB    Modify the module search path # to be deprecated
   RAKUDO_HOME Override the path of the Rakudo runtime files
-  RAKU_HOME   Override the path of the Rakudo runtime files
-  PERL6_HOME  Override the path of the Rakudo runtime files # to be deprecated
   NQP_HOME    Override the path of the NQP runtime files
 
 ♥); # end of usage statement
--- a/src/Perl6/Compiler.nqp
+++ b/src/Perl6/Compiler.nqp
@@ -173,7 +173,6 @@ Note that only boolean single-letter options may be bundled.
 
 The following environment variables are respected:
 
-  RAKUDOLIB   Modify the module search path
   RAKULIB     Modify the module search path
   PERL6LIB    Modify the module search path # to be deprecated
   RAKUDO_HOME Override the path of the Rakudo runtime files
--- a/src/core.c/CompUnit/RepositoryRegistry.pm6
+++ b/src/core.c/CompUnit/RepositoryRegistry.pm6
@@ -76,15 +76,15 @@ class CompUnit::RepositoryRegistry {
                  for parse-include-specS($specs);
             }
 
+            if nqp::existskey($ENV,'RAKUDOLIB') {
+                nqp::push($raw-specs,$_)
+                  for parse-include-specS(nqp::atkey($ENV,'RAKUDOLIB'));
+            }
             if nqp::existskey($ENV,'RAKULIB') {
                 nqp::push($raw-specs,$_)
                   for parse-include-specS(nqp::atkey($ENV,'RAKULIB'));
             }
-            elsif nqp::existskey($ENV,'RAKUDOLIB') {
-                nqp::push($raw-specs,$_)
-                  for parse-include-specS(nqp::atkey($ENV,'RAKUDOLIB'));
-            }
-            elsif nqp::existskey($ENV,'PERL6LIB') {
+            if nqp::existskey($ENV,'PERL6LIB') {
                 nqp::push($raw-specs,$_)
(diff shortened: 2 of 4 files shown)


src/Perl6/Compiler.nqp
```diff
@@ -174,18 +174,20 @@ Note that only boolean single-letter options may be bundled.
 The following environment variables are respected:
 
   RAKUDOLIB   Modify the module search path
-  PERL6LIB    Modify the module search path
+  RAKULIB     Modify the module search path
+  PERL6LIB    Modify the module search path # to be deprecated
   RAKUDO_HOME Override the path of the Rakudo runtime files
-  PERL6_HOME  Override the path of the Rakudo runtime files
+  RAKU_HOME   Override the path of the Rakudo runtime files
```
vrurg: `RAKU_HOME` should be removed. This is where `RAKUDO_HOME` is the right one as it is compiler specific, not language specific.

src/core.c/CompUnit/RepositoryRegistry.pm6
```diff
@@ -80,6 +80,10 @@ class CompUnit::RepositoryRegistry {
                 nqp::push($raw-specs,$_)
                   for parse-include-specS(nqp::atkey($ENV,'RAKUDOLIB'));
             }
+            if nqp::existskey($ENV,'RAKULIB') {
+                nqp::push($raw-specs,$_)
+                  for parse-include-specS(nqp::atkey($ENV,'RAKULIB'));
+            }
```
vrurg: It should be 'either-either', not both. I.e. if `RAKULIB` is defined it takes over `PERL6LIB`. This would allow older and newer versions of rakudo to co-exists peacefully whenever such functionality is needed.

src/Perl6/Compiler.nqp
tbrowder: Okay, but by that correct reasoning, we should also remove PERL6_HOME, correct?

src/core.c/CompUnit/RepositoryRegistry.pm6
tbrowder: I'll put an 'elsif' here testing for RAKULIB first.

tbrowder:
The latest changes reflect the reviewer's comments. The changes tested okay with the full spectest which includes the new roast test for RAKULIB.

src/core.c/CompUnit/RepositoryRegistry.pm6
niner: I think RAKUDOLIB and RAKULIB are supposed to work cumulatively.

patrickbkr: If I know right, both PERL6LIB and RAKUDOLIB are to be deprecated. This logic is there for backward compatibility only. I fear that cumulating might encourage ways of use that we don't want to support in the long run.

tbrowder:
We now have RAKULIB as first priority, and only one of the three are used. Changes pass all spectests including the new test for RAKUDOLIB which seemed to be absent in roast.

vrurg:
I have opened a problem-solving issue raku/problem-solving#167 to finally decide what to do about `RAKUDOLIB`. As a matter of fact, it's never been considered before and only was mentioned once in a while on IRC or some tickets.

tbrowder:
I added a comment to the problem-solving entry.

src/Perl6/Compiler.nqp
```diff
@@ -174,18 +174,18 @@ Note that only boolean single-letter options may be bundled.
 The following environment variables are respected:
 
   RAKUDOLIB   Modify the module search path
```
vrurg: We probably better remove this line until we're more certain what to do about `RAKUDOLIB`.

src/core.c/CompUnit/RepositoryRegistry.pm6
```diff
@@ -76,11 +76,15 @@ class CompUnit::RepositoryRegistry {
                  for parse-include-specS($specs);
             }
 
-            if nqp::existskey($ENV,'RAKUDOLIB') {
+            if nqp::existskey($ENV,'RAKULIB') {
+                nqp::push($raw-specs,$_)
+                  for parse-include-specS(nqp::atkey($ENV,'RAKULIB'));
+            }
+            elsif nqp::existskey($ENV,'RAKUDOLIB') {
```
vrurg: Whatever we decide about the variable, but if it stays – it must be the first, following the principle of most specific case. Wether it should be mutually exclusive with `RAKULIB` or not is still doubtful to me. But considering reasons given by @ugexe, it's better have it non-exclusive for now.

vrurg:
With these two adjustments we're ready to go.

src/core.c/CompUnit/RepositoryRegistry.pm6
tbrowder: Okay, I will make the changes as I understand the comments.

src/main.nqp
```diff
@@ -32,7 +32,6 @@ my $install-dir := $execname eq ''
     !! nqp::substr($execname, 0, nqp::rindex($execname, $sep, nqp::rindex($execname, $sep) - 1));
 
 my $rakudo-home := $comp.config<static_rakudo_home>
-    // nqp::getenvhash()<PERL6_HOME>
```
patrickbkr: Does this need a deprecation cycle?
I'd suspect the *_HOME variables have been used rarely, if at all, since their introduction. But it's impossible to tell for sure.

vrurg: Sure it does.

tbrowder:
So I will:

1. add the env var PERL6_HOME back into the code for using it
2. NOT add it back into the list of env vars shown to the user

Do I understand correctly?

vrurg:
> So I will:
> 
>     1. add the env var PERL6_HOME back into the code for using it
> 
>     2. NOT add it back into the list of env vars shown to the user
> 
> 
> Do I understand correctly?

I think it's a good plan. The Changelog has a line on renaming `PERL6_HOME` to `RAKUDO_HOME`, so no need for additional message in the help output. In a few of months we should add a deprecation note if `PERL6_HOME` is found in the environment. The best time for adding it is about the `6.e` release date because it would have to be removed in `6.f`.

src/rakudo-debug.nqp
```diff
@@ -480,7 +480,6 @@ sub MAIN(*@ARGS) {
         !! nqp::substr($execname, 0, nqp::rindex($execname, $sep, nqp::rindex($execname, $sep) - 1));
 
     my $rakudo-home := $comp.config<static_rakudo_home>
-        // nqp::getenvhash()<PERL6_HOME>
```
vrurg: @tbrowder you forgot this line! :)

tbrowder: arg! on it...

vrurg:
Almost there, but not yet. Nobody noticed that we're prioritizing `RAKU` over `PERL6` and the last commit respects this rule. But  e8f7eb5 restores _main.nqp_ as it was with `PERL6_HOME` having preference over `RAKUDO_HOME`. I think with this sorted out we're ok.

tbrowder:
Sorry, my bad. Will fix shortly....

tbrowder:
So changed back.

vrurg:
Thanks @tbrowder!

PR #3524, opened 2020-03-01
Add Cool.contains(Regex) candidates

taboege:
Commit c3c5dae123 added Str.contains candidates for a Regex $needle.
The Cool class also has a contains method which implicitly converts the
Cool to a Str and calls Str.contains on it. But the new candidates were
not mirrored in Cool, so the OO loop wasn't as convenient. This commit
fixes this oversight.

[uzl++](https://colabti.org/irclogger/irclogger_log/raku?date=2020-03-01#l152) for discovery.


lizmat:
Good catch, good fix!

PR #3525, opened 2020-03-02
Implement dispatchers using nextdispatcherfor/takenextdispatcher

vrurg:
It is the correct way to do chaining of dispatchers.

This PR fixes a problem which is covered by a new test in perl6/roast#627.

This depends on MoarVM/MoarVM#1252 and perl6/nqp#601.

### DIFF
--- a/src/Perl6/Actions.nqp
+++ b/src/Perl6/Actions.nqp
@@ -2706,6 +2706,11 @@ class Perl6::Actions is HLL::Actions does STDActions {
             :op('takedispatcher'),
             QAST::SVal.new( :value('$*DISPATCHER') )
         ));
+        $block[0].push(QAST::Var.new( :name('$*NEXT-DISPATCHER'), :scope('lexical'), :decl('var') ));
+        $block[0].push(QAST::Op.new(
+            :op('takenextdispatcher'),
+            QAST::SVal.new( :value('$*NEXT-DISPATCHER') )
+        ));
         make block_closure($ast);
     }
     method term:sym<unquote>($/) {
@@ -3987,6 +3992,11 @@ class Perl6::Actions is HLL::Actions does STDActions {
             :op('takedispatcher'),
             QAST::SVal.new( :value('$*DISPATCHER') )
         ));
+        $block[0].push(QAST::Var.new( :name('$*NEXT-DISPATCHER'), :scope('lexical'), :decl('var') ));
+        $block[0].push(QAST::Op.new(
+            :op('takenextdispatcher'),
+            QAST::SVal.new( :value('$*NEXT-DISPATCHER') )
+        ));
 
         # If it's a proto but not an onlystar, need some variables for the
         # {*} implementation to use.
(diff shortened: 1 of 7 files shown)


vrurg:
This PR currently breaks S32-exceptions/misc.t. The cause is `Backtrace` has extra frame in the stack. The problem traces down to line 94 of `Backtrace.pm6`:

```perl6
        try X::AdHoc.new(:payload("Died")).throw;
```

It looks like the additional frame is coming from the statement under the `try` is treated as a block or somehow else a block appears here. I suspect that optimization is broken in a way, but can't prove it for now.

My biggest worry about this is that performance of `try` is degraded due to extra frame involved.

@jnthn do you have an idea of what might be wrong here?

lizmat:
The other day I tried to replace the `try` by a `CATCH` block, but this had all sorts of fallout because now there was an extra frame.  I gave up on it, but maybe that should be revisited.  The only reason the `try` is there, is to set `$!` to a valid `Exception` object.  If you have a better way of doing that, then by all means!  For one, it will make `warn` a lot faster.

vrurg:
I was wondering too why so weird way of obtaining a backtrace has been chosen? Is it fast that collecting `nqp::ctxcallers`? Or does it provide more information (doubtful)? After all, the exception code in the core apparently builds the trace – is it so hard to make this a dedicated op?

lizmat:
There's an `nqp::backtrace` op, but by itself it always returns an empty list (if I remember correctly).

But yeah... I'm not sure why creating a backtrace from where one is, is so difficult.

vrurg:
As far as I remember, it's even documented that `nqp::backtrace` works with an exception object only. Basically, it shouldn't be that hard to re-implement backtracing using `nqp::callerctx`. But would it result in any speedup? I doubt so. But only test will show.

Meanwhile, the extra frame at `try` worries me. A quick benchmarking displays approx. 3-8% speed degradation per `try`...

lizmat:
Indeed.

What I think we need is a customizable backtrace capability: now we always create a full backtrace, and then start to filter that.  For warnings (by default), only one level is needed.  So not having to do a lot of work and then throwing it away, would be a good start.

Could you expand on the idea of using `nqp::callerctx`?  Perhaps a backtrace should be implemented like a lazy `nqp::iterator`, rather than as a `nqp::list` ?

vrurg:
Actually, I meant `nqp::ctxcaller` and though about possibly manually backtracing caller contexts. At a first glance, it should be slower than what moar does internally simply because high-level tracer would have to additionally use `nqp::getcodeobj` whereas moar has immediate access to all structures.

But I didn't consider making it lazy. And this is probably what will make a lot of code around faster because often backtraces are not used at all. 

What worries me in this scenario is memory footprint. A lazy iterator would cause a stack of closures to stick around for a while. Not a problem for deadly exceptions or occasionally thrown ones. But could pose some risk of bloating if an application is built around passing exceptions around.

With regard to possible customization of backtrace. Does it really make sense? I mean, a backtrace is usually used once and barely gets passed around.

PS. It starts looking like a branch of problem-solving. Perhaps makes sense to move over there?

lizmat:
Yes  :-)

vrurg:
Back on the track with this PR. I would appreciate any help with two problems I encountered.

- The extra frame. Still no idea where it comes from. What I know is that _t/02-rakudo/08-inline-native-arith.t_ is broken too. Perhaps these two are linked and the extra frame in `Backtrace` is a result of not inlined implicit block.

   I tracked the problem down to the additional `$*NEXT-DISPATCHER` lexical I added where `$*DISPATCHER` is installed. Also, wherever it makes sense, I patched `Optimizer` to respect the lexical same way it does with `$*DISPATCHER`. But no visible result.

- Another problem comes from `Regex::FuzzyToken`. The module wraps a `token` at compile time. Then for some reason the token can't be invoked using `<fuzzy: args>` notation in a grammar, but there is no problem to call it using `{ &fuzzy(args) }`.

   It currently looks to me that token object can't be invoked in normal way and moar falls back to the invocation protocol which ends up looking for `CALL-ME`. Previously wrapping was handled this way it everything worked nicely.

   It is not clear to me what is the difference in handling tokens/rules and subs/methods. To handle compile time properly, `wrap` do nothing if sees active `$*W` and instead installs `fixup_ast`/`deserialize_ast` hooks where it actually does substitution of `$!do`. Apparently compilation of `QAST::Regex` results in something what ignores the attribute and... *This is where I'm stuck now*.

@jnthn could you look into your pockets and see if there are any good advises in there? :)

vrurg:
**UPD** Interesting development: `<&fuzzy: args>` works as expected.

vrurg:
The problem with `Regex::FuzzyToken` is hunted down and is not related to wrapping except tangentially. It's a result of the old issue with `EXPORT` returning a hash and conterization/deconterization. The only reason it escaped the earlier Blin runs is because the old wrapper code was doing it via `CALL-ME`, so when invocation were failing to find a code object it was falling back to the invocation protocol and `CALL-ME` was used.

Not anymore. The fix is simple.

The only stopper for now is the extra frame in Backtrace.

vrurg:
More on extra frame.

The patches for MoarVM and for `Optimize.nqp` to support `takenextdispatcher` had their positive impact on the performance, a benchmark script for `try` shows no visible speed difference between 2020.02 and my HEAD. 5-min long runs fall withing +/- 4% delta.

The "extra" frame is a weird artifact which I can't trace down. The biggest confusion with it is that same code in a user script returns all the same frames on both 2020.02 and the HEAD. And, obviously, there is no difference wether I take the frames from `nqp::backtrace` or via `Backtrace.new(-10)`.

vrurg:
I think this PR is now ready.

Last and important note on `Backtrace`.  In d9ae8fa585a3009ff291d033dc35b0acf68da3c9 I explain why that kind of fix is necessary. Bad thing: the fix makes creation of `Backtrace` instance a bit slower. Not much, but yet. The only way to avoid this is to have `nqp::backtrace(nqp::ctx)` implemented which would replace backtracing via exceptions. I looked briefly into `MVM_exception_backtrace` implementation and it doesn't look pretty much feasible to extend it for contexts. Unfortunately, I lack knowledge of MoarVM internals to do it on my own in some reasonable time frame.

Issue #3527, opened 2020-03-02
"Use of Nil in String context" warning when adding declarator blocks to protos


JJ:
## The Problem

This piece of code:

```
#| Utility class for the Raku Recipes book
unit class Raku::Recipes::Classy;

#| Uses the knapsack algorithms to compute the ingredients that maximize protein content.
proto method optimal-ingredients (|) {*}

```

produces this warning

```
Use of Nil in string context
  in block  at /home/jmerelo/.rakudobrew/versions/moar-2020.02/install/share/perl6/core/sources/D47B7F04B51CBB19CDF1E266B32AF87ACAE3A08B (Pod::To::Text) line 90
Use of Nil in string context
  in block  at /home/jmerelo/.rakudobrew/versions/moar-2020.02/install/share/perl6/core/sources/D47B7F04B51CBB19CDF1E266B32AF87ACAE3A08B (Pod::To::Text) line 90

 ```

## Steps to Reproduce

Run with

    raku --doc=Text Classy.pm6

## Environment

Tested with Raku version:
* 2020.02
* 2019.03.1

PR #3528, opened 2020-03-02
Fix warning in Pod::To::Text signature processing

Altai-man:
Closes https://github.com/rakudo/rakudo/issues/3527

As for testing: do we even have an exhaustive test suite for Pod::To::Text outputs? I hope we do, but gut feeling tells me we don't.

### DIFF
--- a/lib/Pod/To/Text.rakumod
+++ b/lib/Pod/To/Text.rakumod
@@ -86,9 +86,11 @@ sub declarator2text($pod) {
     next unless $pod.WHEREFORE.WHY;
     my $what = do given $pod.WHEREFORE {
         when Method {
-            my @params=$_.signature.params[1..*];
-              @params.pop if @params.tail.name eq '%_';
-              'method ' ~ $_.name ~ signature2text(@params, $_.returns)
+            my @params = $_.signature.params[1..*];
+            with @params.tail.name {
+                @params.pop when '%_';
+            }
+            'method ' ~ $_.name ~ signature2text(@params, $_.returns)
         }
         when Sub {
             'sub ' ~ $_.name ~ signature2text($_.signature.params, $_.returns)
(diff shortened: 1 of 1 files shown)


JJ:
I don't think we do... There are several issues in roast which are still open.

JJ:
Can you please check what's happened with AppVeyor?

Altai-man:
No idea. Can it be re-triggered?

Altai-man:
>src/vm/moar/ops/perl6_ops.c:196:24: error: ‘MVMCollectable’ has no member named ‘flags’

How is that possible, is MoarVM broken right now?

lib/Pod/To/Text.rakumod
```diff
@@ -86,9 +86,11 @@ sub declarator2text($pod) {
     next unless $pod.WHEREFORE.WHY;
     my $what = do given $pod.WHEREFORE {
         when Method {
-            my @params=$_.signature.params[1..*];
-              @params.pop if @params.tail.name eq '%_';
-              'method ' ~ $_.name ~ signature2text(@params, $_.returns)
+            my @params = $_.signature.params[1..*];
+            with @params.tail.name {
+                @params.pop when '%_';
+            }
```
lizmat, line 92: This is probably relatively hot code.  This alternative is about 20x as fast:

    my @params = $_.signature.params.skip;
    @params.pop if @params.tail.name eq '%_';

Altai-man: Thanks!

Altai-man: Wait, nope. It is not equivalent.
The root of this issue is that `@params.tail.name` *can* be Nil and so it says "Compare with Nil is BAD", which is bogus. With this `with ...` construct we avoid working with Nil and your patch actually looks almost the same as it was, faster but having a bogus warning, no?

lizmat: is it not so that `@params.tail.name` can only be `Nil` if `@params == 0`?  if so, then:

    @params.pop if @params && @params.tail.name eq '%_';

lizmat:
Does this PR needs some TLC?  Or can it be merged?

Altai-man:
I think it fixes the warning and we have no tests at all for Pod::To::Text, so blocking this saying "but we don't have test" is odd, so can be merged.

PR #3531, opened 2020-03-03
Make it harder for typechecking exceptions themselves to throw

Kaiepi:
Typechecking exceptions were using `$.` syntax to get their `$!expected` and `$!got` attributes' values. This expects them to have an `item` method, which is not always the case. On top of that, in some places it expected methods like `raku` to exist on them or `name` to exist on their HOW, which isn't always the case either, and would containerize values with `Scalar`.

This makes it so typechecking exceptions don't containerize values, nqp calls are instead of method calls wherever possible, and whether or not a method exists on a value is confirmed before calling it wherever that isn't possible. Now, the only times they should throw are when `.raku` or `.^name` throw, which are errors the user should know about anyway.

Typechecking errors aren't the only cases where you can wind up with `Exception handler not loaded for catch` being thrown, but are by far the most annoying ones.

Passes `make spectest`. Not sure if it passes `make test` or not since `t/06-telemetry/02-usage.t` fails on master as well as this, but all other tests pass.


lizmat:
I thought the `$.foo` and `self.foo` syntax were completely equivalent??

Kaiepi:
`$.`, `@.`, `%.` will call the `item`, `list`, and `hash` methods respectively on their method's return value:

```perl6
class Foo {
    has $.foo;
    method bar() { $.foo }
    method baz() { @.foo }
    method qux() { %.foo }
}

class Bar {
    multi method item(|) {
        say 'called item';
        callsame
    }
    multi method list(|) {
        say 'called list';
        ()
    }
    multi method hash(|) {
        say 'called hash';
        %()
    }
}

my Foo:D $foo .= new: foo => Bar.new;
$foo.bar; # OUTPUT: called item
$foo.baz; # OUTPUT: called list
$foo.qux; # OUTPUT: called hash
```

Normally, this wouldn't matter, but it does here since uninstantiated generics that can pop up from time to time don't have that method, and types made with custom HOWs may not either.

vrurg:
BTW, not really related to this PR, but `$.` family also don't support class-qualification.

lizmat:
Thanks!

PR #3536, opened 2020-03-07
Move RUN_CLEAN_TARGET_FILES to separate script

patrickbkr:
And clean up the logic a lot. The way it looked before it was probably
broken in multiple ways:
- PERL6 and PERL6_DEBUG were never deleted, the line was missing a folder.
- The 'verbose' logic didn't work. There is no 'verbose' variable specified
  anywhere. I just removed the respective logic as I didn't find an obvious
  way to turn the NOECHO variable into a '-v' argument.

### DIFF
--- a/tools/build/clean-target-files.raku
+++ b/tools/build/clean-target-files.raku
@@ -1,0 +1,11 @@
+#!/usr/bin/env raku
+
+unit sub MAIN(Str $folder, *@files where @files.so, Bool :$v);
+
+for @files {
+    given ( $folder.IO.add($_) ) {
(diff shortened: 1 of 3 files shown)


vrurg:
With regard to the verbosity, look at the beginning of *Makefile-common-macros.in*. There're two conditional macros `@if(silent_build...`. I think you need the second one, `@if(silent_build!=on` in which you can insert a variable with `-v`.

patrickbkr:
@vrurg I think it's done. Merge at your convenience.

vrurg:
@patrickbkr Thanks!

Issue #3521, opened 2020-02-28
Declarator blocks not working for unit-declared roles.


JJ:
## The Problem

If you add an declarator block to a Role: 

```perl6
#| Why
unit role NoWhy;
```
It does not seem to work:

```
No documentation available for type 'NoWhy'.
Perhaps it can be found at https://docs.raku.org/type/NoWhy
```

## Expected Behavior

As in the case of classes or modules, it should work.

## Environment
* Operating system: 
* Compiler version (`perl6 -v`): 

Ubuntu 16.04, tested with 2020.02 and 2020.01 and down to 2019.03.1

raiph:
Fwiw, in a 2018.04.1 rakudo it doesn't work for *any* form of `role`: https://glot.io/snippets/fl3ri2wfrc

jnthn:
The reason is that the name installed for a role is actually a role *group*, not the role itself. That lives within the group. Often, however, we try to blur the difference a bit, so the role group behaves as if it was the unsignatured role in the group. We do that for quite a few things, but not `.WHY`. Making us delegate `.WHY` that way is easy enough, and I've implemented it. However it turns out there are three spectests that explicitly expect that we do *not* do that.

I'll make a PR with the changes I did so we can discuss a bit if there's a good reason to not do this change.

JJ:
Thanks!

PR #3549, opened 2020-03-13
Make .WHY on role group delegate to default role

jnthn:
Roles have a two-level structure: the role name is a parametric role
group, which in turns contains the parametric role. In the case of there
being many roles of the same short name, they are within the one group.
Declarator docs correctly attach to the parametric role. However, that
means .WHY on the group doesn't actually result in those docs. You have
to extract the candidate from the group.

In many cases, however, we people ain't using parametric roles, and we
make many meta-operations delegate to the single or unsignatured role
candidate, if any. We do not currently do this for .WHY. This commit
makes us do so, BUT it turns out there are 3 current spectests that will
fail because they explicitly expect that not to happen. Thus, we need to
make a decision over whether we want this change.

Relates to #3521.

### DIFF
--- a/src/Perl6/Grammar.nqp
+++ b/src/Perl6/Grammar.nqp
@@ -2644,6 +2644,8 @@ grammar Perl6::Grammar is HLL::Grammar does STD {
         :my $*zone := 'posreq';
         :my $*multi_invocant := 1;
         :my @*seps := nqp::list();
+        :my $*PRECEDING_DECL; # for #= comments
+        :my $*PRECEDING_DECL_LINE := -1; # XXX update this when I see another comment like it?
         <.ws>
         [
         | <?before '-->' | ')' | ']' | '{' | ':'\s | ';;' >
--- a/src/Perl6/Metamodel/ParametricRoleGroupHOW.nqp
+++ b/src/Perl6/Metamodel/ParametricRoleGroupHOW.nqp
@@ -146,7 +146,7 @@ class Perl6::Metamodel::ParametricRoleGroupHOW
     }
 
     method update_role_typecheck_list($obj) {
-        my $ns := self.'!get_nonsignatured_candidate'($obj);
+        my $ns := self.'!get_nonsignatured_candidate'();
         @!role_typecheck_list := $ns.HOW.role_typecheck_list($ns) unless nqp::isnull($ns);
     }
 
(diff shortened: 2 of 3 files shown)


vrurg:
My sense of DWIM likes the change.

With regard to spectest: this can be `6.e+` specific change. Considering that candidates on a group might be of different language revisions, it would make sense to test for revision of the default candidate.

niner:
The 3 failing spec tests all have the same description "Role group's WHY should not be defined" and were committed by the @hoelzro around the same time. They seem to be testing the same thing in different settings (leading and trailing docs). I'm pretty sure they result from this single example in the speculations:
```
 #| Attaches to the specific parameterized role, rather than the role group itself
    role R[::T] {}
    role R {}
```

This is what the tests were meant to establish. Note that this one sentence does not actually prohibit the role group from delegating to the default role. It merely requires that individual roles can be documented. This commit does not violate that requirement. It just extends availability of the default role's docs to the role group which as the linked issue makes pretty clear is what people expect (because they usually don't think in terms of role groups).

I think the failing spec tests are simply overly specific and adapting them so this can be merged will do much more good than harm. I only see potential for a little bit of confusion of tools, but since @jnthn is directly involved with the current major consumer (Comma-IDE), he'll know best about this anyway. Even if there is any confusion at all, tools would have to be adapted anyway to deal with the next language version and then maybe even differentiate between versions. So it's not just us who'd have to carry code to hide this feature in old versions, but those other tools as well. I'm pretty sure that's not worth the hassle and it's better to merge this as-is.

niner (approved):

lizmat:
Fixed the conflict, is this ready to be merged?

lizmat:
@vrurg @jnthn This is **NOT** ready to be merged, as there appears to be two interpretations of the `!get_default_candidate` method, and there are calls in the code to both interpretations.

vrurg:
@lizmat There is no contradiction, just an overlook. :) The first one is apparently a leftover. Lines 220 and 227 are to be fixed. I can get it done if @jnthn doesn't mind.

jnthn:
@vrurg Sure, go for it. :pray:

vrurg:
I couldn't stop on just fixing the methods...

Since a role group is not documentable, I removed the corresponding role from it. `set_why` will now throw because I consider delegating it potentially unsafe. This breaks none of Rakudo or spectests.

Spectests were adjusted to the new behavior in Raku/roast#783 by removing the requirement of `RoleGroup.WHY` to return `Nil`, according to @niner comment above. Instead I made why-both.t making sure that group's `WHY` is the same, as its default candidate. This required me to be explicit in the order of declaration of parametric roles and, basically, rely on the internal implementation. Doesn't feel good to me. Perhaps we should make `!get_default_candidate` public and rename it to `default-candidate`?

Otherwise PR is ready for merge. The failure of Lin_MVM_spec CI can be ignored, it is the only one ran against the specs, which are not ready yet.

Issue #1746, opened 2018-04-18
Error missing declared signature if `where` clause is used in proto


zoffixznet:
***See [Newcomer Guide to Contributing to Core Perl 6](https://rakudo.party/post/Newcomer-Guide-to-Contributing-to-Core-Perl-6) for tips on fixing this Issue***

----------------

Normally, this message tells you what the declared signature is (in this case, `(Int)`):
```
$ perl6 -e 'sub z(Int) {}; z "x"'
===SORRY!=== Error while compiling -e
Calling z(Str) will never work with declared signature (Int)
```

But if you use a `where` clause and stick it into `proto`, the `with declared signature` part doesn't say what the declared signature is:
```
19:39 	Zoffix 	m: proto z($ where rand) {*}; say z "x"
19:39 	camelia 	rakudo-moar 9e6ae276a: OUTPUT: «5===SORRY!5=== Error while compiling <tmp>␤Calling z(Str) will never work with declared signature ␤at <tmp>:1␤------> 3proto z($ where rand) {*}; say 7⏏5z "x"␤»
```

jsimonet:
In addition, using a where clause causes the signatures to be evaluated at runtime. If a multi is specified, the error is a TypeCheck Exception.
```
perl6 --stagestats -e 'proto z( $ where rand ) {*}; multi z($ where rand){}; say z "x"'
Stage start      :   0.000
Stage parse      :   0.147
Stage syntaxcheck:   0.000
Stage ast        :   0.000
Stage optimize   :   0.004
Stage mast       :   0.009
Stage mbc        :   0.000
Stage moar       :   0.000
Constraint type check failed in binding to parameter '<anon>'; expected anonymous constraint to be met but got Str ("x")
  in block <unit> at -e line 1
```

jsimonet:
Tried to implement a fix in PR #1868

dolmen:
This looks fixed in rakudo star 2018.06:
```
$ perl6 -e 'proto z($ where rand) {*}; say z "x"'
===SORRY!=== Error while compiling -e
Calling z(Str) will never work with declared signature
at -e:1
------> proto z($ where rand) {*}; say ⏏z "x"
```

zoffixznet:
> Tried to implement a fix in PR #1868

Commented on the PR.

> This looks fixed in rakudo star 2018.06:

That's still missing what the declared signature is.

tinmarino:
[...]

Issue #2227, opened 2018-08-19
Rerevert “Throws X::Multi::NoMatch if a proto routine is defined without multis”


AlexDaniel:
Discussion: https://github.com/rakudo/rakudo/pull/1868#issuecomment-413386084

AlexDaniel:
I reverted the commit for now because I don't know what would be the rightest solution. Please feel to rerevert once a fix for roast is ready.

Ping @jsimonet @timo @zoffixznet.

zoffixznet:
@AlexDaniel which test was failing? 

The PR was not merged by mistake, but it did reveal a previously hidden bug: https://github.com/rakudo/rakudo/issues/2216

AlexDaniel:
@zoffixznet This one: https://github.com/perl6/roast/blob/894a56c87b208d8e5017150eeac4bf142c03cf5e/S32-exceptions/misc.t#L48

zoffixznet:
Weird. It exists on master too, but I thought the spectest was passing for the PR.

I'm not sure, but the optimizer might be already too late a place to use lang-version-specific behaviours with [our current means](https://github.com/rakudo/rakudo/blob/5af3c5011946db73a43afea2dfe8d11d11413dcf/src/Perl6/World.nqp#L485-L493).

But if we can leave the old behaviour in 6.c, then for 6.d, I'd propose moving this test to spec [APPENDICES](https://github.com/perl6/roast/tree/master/APPENDICES), because I'd guess not all implementations would want/be able to examine all the candidates and check if we only got the proto.

jsimonet:
If you change the `Str` type to an `Int`, the generated exception is the same but without the signature proposition.
```
perl6 -e 'proto sub foo(Int) {*}; foo 42;'
===SORRY!=== Error while compiling -e
Calling foo(Int) will never work with declared signature 
at -e:1
------> proto sub foo(Int) {*}; ⏏foo 42;
```

I changed the exception because to me, it's not a `X::TypeCheck::Argument` exception.

PR #3558, opened 2020-03-20
Better error message when calling a star proto without declared multi 

tinmarino:
Fix: #1746

__Problem:__ Error message whencalling a star proto linked to an bad signature but it was impossible to call with any signature.
__Solution:__ New Error message `X::Multi::Undefined`

In comment the before then after messages

```bash
perl6 -e 'proto z(Int $) {*}; say z 3'
# Calling z(Int) will never work with declared signature 
# Multi dispatch has not been declared for star proto z(Int)

perl6 -e 'proto z($ where rand) {*}; say z "x"'
# Multi dispatch has not been declared for star proto z($ where { ... })
# Calling z(Str) will never work with declared signature 

perl6 -e 'sub z(Int) {}; z "x"'
# Calling z(Str) will never work with declared signature (Int)
# Same

perl6 -e 'proto z(Int $) {*}; say z "x"'
# Calling z(Str) will never work with signature of the proto (Int)
# Same

perl6 -e 'proto z(Int $a) {say "proto"}; say z "eaz"'
# Type check failed in binding to parameter '$a'; expected Int but got Str ("eaz")
# Same

perl6 -e 'proto z(Int $) {*}; multi z(Int $int where {$int > 0} ) {42}; say z "Int"'
# Calling z(Str) will never work with signature of the proto (Int)
# Same

perl6 -e 'proto z(Int $) {*}; multi z(Int $int where {$int > 0} ) {42}; say z -1'
# Cannot resolve caller z(-1); none of these signatures match:
#    (Int $int where { ... })
# Same
```

### DIFF
--- a/src/Perl6/Optimizer.nqp
+++ b/src/Perl6/Optimizer.nqp
@@ -2797,6 +2797,16 @@ class Perl6::Optimizer {
     }
 
     method report_inevitable_dispatch_failure($op, @types, @flags, $obj, :$protoguilt) {
+        my %opts := nqp::hash();
+        my $is_dispatcher := nqp::can($obj, 'is_dispatcher') && $obj.is_dispatcher;
+
+        # Clause: no multi declared
+        if $is_dispatcher && !$protoguilt && !$obj.dispatchees {
+            %opts<objname> := "star proto " ~ $obj.name ~ $obj.signature.gist;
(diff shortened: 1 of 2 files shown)


tinmarino:
Noticed latter: #1868 Handles better that issue but got Reverted #2227

tinmarino:
It seems that I forgot this pending PR, sorry.

PR #3565, opened 2020-03-23
Update config.yml

DieOfCode:


JJ:
@DieOfCode can you please clarify why the version was bumped?

AlexDaniel:
I don't get it. I tried figuring this out but I couldn't find any information on the specifics of the 2.1.1 version. If it was a bump to 2.1 it would've been more obvious, but we're already on 2.1. I'll close this until there's an explanation.

PR #3575, opened 2020-03-27
Improve introspection of anonymous parameters

Kaiepi:
`+` currently gets stringified as `+$ is raw`, likewise `\` as `$ is raw` and `Callable` as `Callable $`. The first two are bugs, the third is behaviour that Roast tests for, but is behaviour I disagree with.

Since `$ is raw` and `\` are impossible to differentiate otherwise, this adds the `SIG_ELEM_SCALAR_SIGIL` parameter flag. This also adds the `SIG_ELEM_NO_VARIABLE` flag to make it possible to differentiate between `Callable` and `Callable $`, which should also have the bonus of making compilation slightly faster, though I don't know exactly how to benchmark this.

Passes `make test`, passes all tests in `make spectest` apart from a handful in `S06-currying/positional.t` and `S06-signature/introspection.t` that expect the old behaviour with `Callable $`.

### DIFF
--- a/src/Perl6/Actions.nqp
+++ b/src/Perl6/Actions.nqp
@@ -1724,7 +1724,7 @@ class Perl6::Actions is HLL::Actions does STDActions {
                 if $*IMPLICIT {
                     my $optional := $*IMPLICIT == 1;
                     @params.push(hash(
-                        :variable_name('$_'), :$optional,
+                        :variable_name('$_'), :sigil('$'), :$optional,
                         :nominal_type($*W.find_symbol(['Mu'])),
                         :default_from_outer($optional), :is_raw(1),
                     ));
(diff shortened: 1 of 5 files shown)


lizmat:
> making compilation slightly faster

OOC, Why would improved introspection make things faster?

niner:
I guess because of https://github.com/rakudo/rakudo/pull/3575/commits/28cef0d4df5eeb9b3f7fa0109c5f2dcbf65b7c78

Kaiepi:
The `SIG_ELEM_NO_VARIABLE` flag is used in `Perl6::Actions` to determine if a bunch of flags need to be set or not, since a bunch of them should only apply if there are variables in the parameter. This shouldn't make compilation a whole lot faster, though.

lizmat:
Is this PR salvageable ?

lizmat:
Apparently **NOT** salvageable, so closing.

PR #3583, opened 2020-03-31
Ensure Routine subtypes are composed before performing mixins with their instances

Kaiepi:
The only reason that pesky branch in `Metamodel::Mixins.mixin` can't be removed at the moment is because mixins of routine types get made by traits before they get composed. By applying `Callable` to `Code` and composing `Sub`/`Method`/`Submethod` in the prologue, there are no longer any mixins of uncomposed metaobjects made during the setting, which should make it possible to remove this branch.


lizmat:
Good idea!  Looks good to me.

lizmat:
Thanks!

PR #3585, opened 2020-03-31
Improve Capture positional methods

Kaiepi:
The commits should speak for themselves (apart from the second, but the reasoning is explained in its commit message).


lizmat:
Thanks!

PR #3590, opened 2020-04-02
Fix for wrapping of parent's first multi-candidate

vrurg:
Resolves the problem described by @jnthn in Raku/problem-solving#170. The point of this fix is to make wrapper dispatcher transparent for the wrapped routine. Whatever combination of dispatcher/nextdispatcher the wrapped routine was invoked with – the same combination must be used when invoking the last candidate which is the actual wrapped routine code.

Temporary patch until we get the unified dispatcher, proposed in Raku/problem-solving#170.

### DIFF
--- a/src/Perl6/Metamodel/Dispatchers.nqp
+++ b/src/Perl6/Metamodel/Dispatchers.nqp
@@ -20,14 +20,12 @@ class Perl6::Metamodel::BaseDispatcher {
     method get_call() {
         my $call := @!candidates[$!idx];
         ++$!idx;
-        if (nqp::can($call, 'is_dispatcher') && $call.is_dispatcher)
-            || (nqp::can($call, 'is_wrapped') && $call.is_wrapped)
-        {
-            nqp::nextdispatcherfor(self, $call);
+        if self.is_wrapper_like && self.last_candidate {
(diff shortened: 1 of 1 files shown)


src/Perl6/Metamodel/Dispatchers.nqp
```diff
@@ -191,10 +192,19 @@ class Perl6::Metamodel::WrapDispatcher is Perl6::Metamodel::BaseDispatcher {
         my $next_dispatcher := nqp::existskey($lexpad, '$*NEXT-DISPATCHER')
                                 ?? nqp::atkey($lexpad, '$*NEXT-DISPATCHER')
                                 !! nqp::null();
-        self.new(:@candidates, :idx(1), :$next_dispatcher)
+        my $stashed_dispatcher := nqp::existskey($lexpad, '$*DISPATCHER')
+                                    ?? nqp::atkey($lexpad, '$*DISPATCHER')
+                                    !! nqp::null();
```
niner, line 197: I thought nqp::atkey already returns nqp::null for a non-existent key?

lizmat: It does for normal NQP hashes, afaik.  Not for stashes, for some reason.

PR #3602, opened 2020-04-05
Fix issue 3581

patrickbkr:
Fix `--nqp-home` Configure.pl parameter. Also be more explicit about whether the nqp_home and rakudo_home variables. Now the `nqp_home` and `rakudo_home` variables are only used to reference the folder where we install files to. `static_rakudo_home` and `static_nqp_home` are used to refer to the static compile time override to set a different than default home dir.
Can't be merged before [#nqp/611](https://github.com/Raku/nqp/pull/611) is merged and the nqp version bumped.


patrickbkr:
@vrurg Can you have another look? I'm especially unsure about `configure_moar_backend()` where I changed where the variables are saved.

PR #3604, opened 2020-04-05
Fix building non-moar backends on Windows

patrickbkr:
On non-moar backends the nqp executable extension is still '.bat' and not '.exe'.

### DIFF
--- a/tools/lib/NQP/Config/Rakudo.pm
+++ b/tools/lib/NQP/Config/Rakudo.pm
@@ -594,7 +594,6 @@ sub gen_nqp {
     my $prefix        = $config->{prefix};
     my $sdkroot       = $config->{'sdkroot'};
     my $startdir      = $config->{'base_dir'};
-    my $exe           = $config->{exe};
     my $nqp_want      = $config->{nqp_want};
     my $git_protocol  = $options->{'git-protocol'} // 'https';
     my @moar_options  = @{ $options->{'moar-option'} // [] };
@@ -612,6 +611,7 @@ sub gen_nqp {
     for my $b ( $self->active_backends ) {
         my $bconfig = $self->backend_config($b);
         my $postfix = $self->backend_abbr($b);
+        my $exe     = $b eq 'moar' ? $config->{exe} : $config->{bat};
         my $tpath   = File::Spec->catfile( $prefix, 'bin', "nqp-$postfix$exe" );
         my $bin     = $bconfig->{nqp_bin}
           || (


tools/lib/NQP/Config/Rakudo.pm
```diff
@@ -612,6 +611,7 @@ sub gen_nqp {
     for my $b ( $self->active_backends ) {
         my $bconfig = $self->backend_config($b);
         my $postfix = $self->backend_abbr($b);
+        my $exe     = $b eq 'moar' ? $config->{exe} : $config->{bat};
```
vrurg, line 614: I don't like us being too specific here. Consider rakudo getting more backends with different runner extensions (a voice inside me screams: "Damn you, Bill Gates!").

Perhaps, the most correct solution would be to set `nqp_runner` config variable in `configure_<backend>_backend` methods. To make life a little bit easier we might set it to some kind of common default – say, to `File::Spec->catfile( $prefix, 'bin', "nqp-$postfix$bat" );` so that backends with exe would just have to override it with `File::Spec->catfile( $prefix, 'bin', "nqp-$postfix$exe" );`. In other words, in the current situation it would mean that only `configure_moar_backend` would have to override it. Then at this location we would just use `$bin = $config->{nqp_runner}`.

Alternatively, we could iterate over exe and bat extensions and test for executable existence. But this method is slower and prone to errors if a runtime switches from script to binary format.

BTW, I just have noticed that `$tpath` from the next line is never used. Perhaps it's a leftover of some older or even debug code.

patrickbkr:
@vrurg Can you take this issue? I just stared at the build system code for over an hour and still didn't manage to understand where default variables for backends should be set. :-(

Also can you add some documentation of how the different config layers stack? Some diagram or similar that shows the relationship of the different config stores might help a lot.

I would expect a generic config layer (`$self->{config}`) on which either a js, moar or jvm layer is stacked (`$self->backend_config()`). So setting a variable in the generic layer would set a default that could be overwritten by a same named variable in a backend config. All configuration options can be accessed via the backend config.
The above seems not to be how it works. Can you clear up the confusion?

vrurg:
@patrickbkr I agree that some more docs would be useful. 

I'll see what I can do. Unfortunately, I'm too overloaded lately and not sure when would get time for this. But would try get my hands on this ASAP.

patrickbkr:
Merging this after consultation with @vrurg to get this bug fixed. I will create a separate ticket to find a cleaner solution.

PR #3609, opened 2020-04-08
Expose nqp parameterization ops through Metamodel::Primitives

Kaiepi:
This allows users to take advantage of parameterization caching.

Passes `make test` and `make spectest`. There are new tests to be added to Roast for these methods.


lizmat:
Thanks!

Issue #3535, opened 2020-03-07
Comparing Buf with the is routine from Test does not behave well


Skarsnik:
## The Problem

When comparing Buf with the routine `is` from the test module, it will work if data are the same, but if the 2 Buf is not the same it will make the module dies because it can't call Str to show the regular error of 'Expected : blablab, Got : claclalc'

## Expected Behavior

Fail like a regular test failing to pass when data are not the same and not like a crash

## Actual Behavior

Failed test at -e line 1
Cannot use a Buf as a string, but you called the Stringy method on it
  in block  at /home/skarsnik/raku/share/perl6/core/sources/2D4A7CA10695CD2B374573413D261A660E282E2A (Test) line 188

## Steps to Reproduce

raku -e 'use Test; is Buf.new([0 xx 40]), Buf.new([0 xx 40]); is Buf.new([0 xx 40]), Buf.new([1 xx 40]);'
ok 1 -
not ok 2 -
# Failed test at -e line 1
Cannot use a Buf as a string, but you called the Stringy method on it
  in block  at /home/skarsnik/raku/share/perl6/core/sources/2D4A7CA10695CD2B374573413D261A660E282E2A (Test) line 188
  in sub is at /home/skarsnik/raku/share/perl6/core/sources/2D4A7CA10695CD2B374573413D261A660E282E2A (Test) line 171
  in block <unit> at -e line 1

## Environment

skarsnik@DESKTOP-UIA12T1:/mnt/f/Project/QUsb2snes/tests/usb2snes-tests/raku-tests$ raku --version
This is Rakudo version 2020.01 built on MoarVM version 2020.01.1
implementing Perl 6.d.

Altai-man:
See conversation at https://colabti.org/irclogger/irclogger_log/raku?date=2020-03-07#l204

PR #3613, opened 2020-04-10
Add a fix for issue #3535

Skarsnik:
I hope it's simple enought

### DIFF
--- a/lib/Test.rakumod
+++ b/lib/Test.rakumod
@@ -188,8 +188,15 @@ multi sub is(Mu $got, Mu:D $expected, $desc = '') is export {
                     ~ "     got: $got.raku()";
             }
             else {
-                _diag "expected: '$expected'\n"
+		try { # if the type support Stringification
+                      # note: we can't use ^can('Str') as Buf error in its Str method itself
+                    _diag "expected: '$expected'\n"
                     ~ "     got: '$got'";
+                    True;
+                } or {
+                    _diag "expected: $expected.raku()\n"
+                    ~ "     got: $got.raku()";
+                }
             }
         }
     }
--- a/t/01-sanity/99-test-basic.t
+++ b/t/01-sanity/99-test-basic.t
@@ -142,6 +142,14 @@ todo( 'failing is-deeply returns False' );
 my $is-deeply = is-deeply {a => 1}, {}, 'is-deeply with exta key fails';
 nok $is-deeply, 'failing is-deeply returns False';
 
+# Testing issue #3535
+is Buf.new(42), Buf.new(42), "Comparing eq Buf";
+lives-ok {
+    # NOT_TODO
+    todo("Comparing 2 not eq Buf, should not pass");
+    is Buf.new(42), Buf.new(43);
+}, "Comparing neq Buf";
+
 done-testing;
 
 # vim: ft=perl6


lizmat:
Wouldn't it make more sense to just *always* `_diag` with the `.raku` method?

Skarsnik:
> Wouldn't it make more sense to just _always_ `_diag` with the `.raku` method?

For people that wrote Str method for their own class it's probably better. Maybe having another routine of a named argument to `is` to force the usage of .raku?

vrurg:
@lizmat `.raku` is over-verbose sometimes. Extra noise in a test output barely help finding a problem.

lib/Test.rakumod
```diff
@@ -188,8 +188,14 @@ multi sub is(Mu $got, Mu:D $expected, $desc = '') is export {
                     ~ "     got: $got.raku()";
             }
             else {
+		if try { # if the type support Stringification
+		        # note: we can't use ^can('Str') as Buf error in its Str method itself
                 _diag "expected: '$expected'\n"
                     ~ "     got: '$got'";
+		} {} else {
```
vrurg, line 195: This empty block here is too confusing. Wouldn't it be better to pre-stringify `$expect` and `$got` and then use these values for `_diag`? Besides, `_diag` below already duplicates the one above.

Skarsnik: I was not sure if using a try {} CATCH {} structure made more sense. I don't want to refactor too much the code.

vrurg: It's not that much of a refactor. But taking care of code duplication is what must be done anyway.

Besides, instead of `try` you can test for `Buf` type object.

Skarsnik: I had another look at the code and I am not sure I want to pre build the string for diag (can be somehow costly for some object) and combining the 2 conditions will remove some clarity. I can't test for just Buf, you never know what class the user will have that can have the same behaviour. Duplicated code is not the best, but clarity is always better for me.
I could maybe change the first message to highlight the fact that it's only whitespaces that were differents in the 2 compared object.

vrurg: Good point about `Buf` not being the only possible cause of the problem. Still, as I looked closer at the code,  there is a suggestion and a fix to be made.

First, to the potential bug. It is not specced what `_diag` returns. Thus, it is possible that `try { }` construct would return a false value despite no exception was thrown. So, it'd be more correct with `try { _diag; True }`. 

The suggestion should improve the clarity by removing extra constructs: 

```perl6
try {
    _diag(...); 
    True 
} or _diag(...)
```

would do the same what `if ... else ...` currently does but has better overall readability.

lib/Test.rakumod
```diff
@@ -188,8 +188,15 @@ multi sub is(Mu $got, Mu:D $expected, $desc = '') is export {
                     ~ "     got: $got.raku()";
             }
             else {
-                _diag "expected: '$expected'\n"
+		try { # if the type support Stringification
+		        # note: we can't use ^can('Str') as Buf error in its Str method itself
+                    _diag "expected: '$expected'\n"
                     ~ "     got: '$got'";
+                True;
```
vrurg, line 195: Weird indentation

t/01-sanity/99-test-basic.t
```diff
@@ -142,6 +142,14 @@ todo( 'failing is-deeply returns False' );
 my $is-deeply = is-deeply {a => 1}, {}, 'is-deeply with exta key fails';
 nok $is-deeply, 'failing is-deeply returns False';
 
+# Testing issue #3535
+is Buf.new(42), Buf.new(42), "Comparing eq Buf";
+lives-ok {
+	# NOT_TODO
+        todo("Comparing 2 not eq Buf, should not pass");
+	is Buf.new(42), Buf.new(43);
```
vrurg, line 150: And here indentation too

Skarsnik: 4 spaces is the good indent?

vrurg: Yes, this is the implicit standard.

Skarsnik:
This should be alright?

vrurg:
Didn't see your commit... Almost, but somehow 'True' at line 195 is still not indented properly.

Skarsnik:
Sorry for flood, but I hope it's good this time :)

vrurg:
It's not flood but a normal review process. But now everything is in good shape.

Thank you for your work!

vrurg:
@Skarsnik  And a little advice/request for the future PRs: don't use subjects referring to issue numbers, it's hard  to follow them after merging. Subject is good if it's self explanatory.

PR #3614, opened 2020-04-10
Improve smartmatching against Mu/Any/Junction

Kaiepi:
This allows any instance to typecheck against any type object by giving `Mu` an `ACCEPTS` candidate for `Mu`. This alone breaks smartmatching with junctions, though. To fix this, rather than implicitly autothreading junctions with an `Any:U` candidate, threading of junctions is now explicitly handled from a `Mu:U` candidate. While it's not strictly necessary for this to reside in `Mu`, this allows `any(Mu, Any) ~~ Mu` to return `True` instead of `False`, which I feel is more correct. A bonus to doing all this is smartmatching with junctions is now around twice as fast, and there should be a speedup of some form for *all* smartmatches since some `ACCEPTS` candidates in `Mu`, `Any`, and `Junction` are no longer necessary!

### DIFF
--- a/src/core.c/Any.pm6
+++ b/src/core.c/Any.pm6
@@ -15,13 +15,10 @@ my role  Numeric { ... }
 my class Any { # declared in BOOTSTRAP
     # my class Any is Mu
 
-    multi method ACCEPTS(Any:D: Mu:D \a) { self === a }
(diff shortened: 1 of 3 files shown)


lizmat:
Thanks!

Issue #3622, opened 2020-04-13
Some loop constructs crash when used with labels


usev6:
## The Problem
I've found some code examples with labels usage within nested loops that crashed unexpectedly. The shortest example is:
```
$ raku -e '@ = (FOO: loop { loop { last FOO } })'
labeled last without loop construct
  in block <unit> at -e line 1
```

As far as I understand these crashes happen because the following four classes call ```nqp::handle``` without any information about a used label:
* ```CStyleLoop```
* ```Loop```
* ```RepeatLoop```
* ```WhileLoop```

I'll add a PR that seems to fix the problem. Even if I'm not sure if the fix is correct, the PR should show what I suspect to be wrong.

## Compiler version
```
$ raku -v
This is Rakudo version 2020.01 built on MoarVM version 2020.01.1
implementing Perl 6.d.
```
(Happens on current HEAD, too.)

usev6:
Was fixed with https://github.com/rakudo/rakudo/pull/3623.

PR #3623, opened 2020-04-13
Add missing label support for some loop constructs

usev6:
A couple of loop constructs were missing support for labels. Probably
this went unnoticed, because 'for' loops were not affected and a lot
of iterating is handled by optimized code in
src/core.c/Any-iterable-methods.pm6 (IterateOneWithPhasers and friends).

This should fix the problem described in #3622.

### DIFF
--- a/src/Perl6/Actions.nqp
+++ b/src/Perl6/Actions.nqp
@@ -157,10 +157,18 @@ sub wanted($ast,$by) {
                 $block := Perl6::Actions::make_thunk_ref($body, $body.node);
                 $block-closure := block_closure($block);
             }
+
+            # make sure from-loop knows about existing label
+            my $label := QAST::WVal.new( :value($*W.find_symbol(['Any'])), :named('label') );
+            for @($ast) {
+                $label := $_ if nqp::istype($_, QAST::WVal) && nqp::istype($_.value, $*W.find_symbol(['Label']));
+            }
+
             my $past := QAST::Op.new: :node($body.node),
                 :op<callmethod>, :name<from-loop>,
                 QAST::WVal.new(:value($*W.find_symbol(['Seq']))),
-                $block-closure;
+                $block-closure,
+                $label;
 
             # Elevate statevars to enclosing thunk
             if $body.has_ann('has_statevar') && $block.has_ann('past_block') {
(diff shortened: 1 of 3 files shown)


usev6:
I'm still looking were to put some tests. Probably only some of them belong to roast, so I plan to add another commit to this PR.

lizmat:
Cool!  But I don't think we're going to need named variable `:$label` internally: just make that an optional positional.

usev6:
Ok, I have to think about that. I don't have a clear picture yet how to implement this with an option positional. I tried to mimic the code from flatmap and sequential-map from ```src/core.c/Any-iterable-methods.pm6```.

@lizmat If you'd like to fix this differently, just go ahead ;). It took me some time to arrive at this solution and I doubt I'll have much time during the weekdays.

usev6:
I guess, all these tests could go into roast: https://github.com/Raku/roast/pull/635

So I won't add a commit for tests for Rakudo to this PR.

usev6:
> But I don't think we're going to need named variable :$label internally: just make that an optional positional.

@lizmat Could you maybe elaborate where :$label should be changed to an optional positional?
I tried to do that for the ```from-loop``` candidates but failed. (Maybe because I don't understand how to do it.)

lizmat:
`$label?` instead of `:$label`

usev6:
Well, yes. But I still don't understand where that should be changed.

In one of my tests with changing the signature of the ```from-loop``` candidates to use ```$label = Any``` instead of ```:$label``` I got the following error:
```
Ambiguous call to 'from-loop(Seq: Block, Block, :repeat(Int), :label(Label))'; these signatures all match:
:(Seq: &body, $label = Any, *%_)
:(Seq: &body, &cond, $label = Any, :$repeat!, *%_)
```

I think I'll leave this alone. The bug report (https://github.com/rakudo/rakudo/issues/3622) seems still valid to me. Also, I hope the tests (https://github.com/Raku/roast/pull/635) could be useful. Maybe I'll add them as skipped tests for now.

lizmat:
no, it's fine, it was just nitpicking by me.

lizmat:
Thanks!

PR #3625, opened 2020-04-14
Add an iterator multi candidate to Supply

Kaiepi:
A problem with `Seq` and `List` supply coercions was that `take` could be used
inside of them, which would mix values taken with those that are
actually emitted by the supply in the coercion. Instead of using `gather`
to handle coercions in the first place, this adds an `iterator` method to
`Supply` to handle that for these two methods. Because of this,
`Supply.list` no longer needs to create a throwaway `Seq` to coerce to a
`List`.

### DIFF
--- a/src/core.c/Supply-coercers.pm6
+++ b/src/core.c/Supply-coercers.pm6
@@ -41,12 +41,22 @@
             nqp::stmts(
               (my $got := nqp::shift($!queue)),
               nqp::if(
-                nqp::eqaddr($got, ConcQueue),
+                nqp::eqaddr($got,ConcQueue),
                 nqp::if(
                   nqp::isconcrete($!exception),
                   $!exception.throw,
                   IterationEnd),
                 $got))
+        }
+
+        method push-all(\target --> IterationEnd) {
+            nqp::stmts(
+              nqp::while(
+                nqp::not_i(nqp::eqaddr((my $got := nqp::shift($!queue)),ConcQueue)),
+                target.push($got)),
+              nqp::if(
+                nqp::isconcrete($!exception),
+                $!exception.throw))
         }
 
         # method is-lazy(--> Bool:D) { ... }
--- a/src/core.c/Supply-coercers.pm6
+++ b/src/core.c/Supply-coercers.pm6
@@ -51,8 +51,8 @@
 
         method push-all(\target --> IterationEnd) {
             nqp::stmts(
-              nqp::while(
-                nqp::not_i(nqp::eqaddr((my $got := nqp::shift($!queue)),ConcQueue)),
+              nqp::until(
+                nqp::eqaddr((my $got := nqp::shift($!queue)),ConcQueue),
                 target.push($got)),
               nqp::if(
                 nqp::isconcrete($!exception),
--- a/src/core.c/Supply-coercers.pm6
+++ b/src/core.c/Supply-coercers.pm6
@@ -15,30 +15,51 @@
         $c
     }
 
-    my class ConcQueue is repr('ConcBlockingQueue') { }
+    my class SupplyIterator does Iterator {
+        my class ConcQueue is repr('ConcBlockingQueue') { }
+        has $!queue;
+        has $!exception;
+
+        method new($supply) {
+            nqp::create(self)!SET-SELF($supply)
+        }
+        method !SET-SELF($supply) {
+            $!queue     := nqp::create(ConcQueue);
+            $!exception := Nil;
+            $supply.tap: {
+                nqp::push($!queue, $_)
+            }, done => -> {
+                nqp::push($!queue, ConcQueue); # Sentinel value.
+            }, quit => {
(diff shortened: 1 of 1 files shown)


lizmat:
In the core, we write attributes with twigils, to at least improve readability.  So no `has $queue;`, but `has $!queue;`

lizmat:
Also, since there's a good change people will use this as `my @a = Supply.list`, perhaps a `push-all` candidate would make things even better?

Kaiepi:
> In the core, we write attributes with twigils, to at least improve readability. So no has `$queue;`, but `has $!queue;`

The reason I don't use twigiled attributes here is because attribute lookups are more expensive than lexical ones in this case. Normally, I don't think the difference is significant enough to matter, but here there's the potential for any number of lookups to be done all at once. This is actually slower than the original code when calling `.elems` on supplies emitting large numbers of values when written with twigiled attributes, and is something I think would be worth doing for other iterators in Rakudo as well if reasonable.

> Also, since there's a good change people will use this as `my @a = Supply.list`, perhaps a `push-all` candidate would make things even better?

I'll test out how `push-all` (maybe `push-exactly`, `skip-at-least`, and `sink-all` as well) improve things.

Kaiepi:
It's probably worth mentioning that I think supply iterators should be lazy (at least until the supply is done) since it's not necessarily known how many values will wind up being emitted, which is why that comment's there. I'm not sure if this is something that would be too breaking for v6.c or not though.

lizmat:
It is my understanding that removing the twigil from the attribute is only cosmetic, and does *not* result in a lexical lookup rather than an attribute lookup.  Looking at the output of `--target=optimize` of `class A { has $a; method a() { $a } }`, versus the output of `class A { has $!a; method a() { $!a } }`, I see **no** difference except where the names differ (and then that's the only difference).

So, in my opinion, *not* using twigils only makes for more obfuscated code and does *not* make things faster.

But of course, I'd like to be proven wrong: if so, we could win a lot of efficiency in the core!

jnthn:
> The reason I don't use twigiled attributes here is because attribute lookups are more expensive than lexical ones in this case. 

A declaration `has $foo` actually declares an attribute `$!foo`, and sticks a note somewhere that if it sees a `$foo` then it should emit an attribute lookup for `$!foo`. Thus they produce exactly the same code, except the compiler has to do a bit more bookkeeping along the way with the non-sigil'd form. It's completely compiled away otherwise. (I was pondering the other day whether by default RakuAST will even preserve this distinction, or if we just keep doing that symbol rewrite prior to AST construction...)

jnthn:
> So, in my opinion, not using twigils only makes for more obfuscated code and does not make things faster.

Ah, and I agree with this. Please use the twigils consistently. The `has $foo` form was a sop to folks who didn't like the idea of twigils, but most folks don't seem to care much in the end, so in the end this feature was probably a waste of time.

lizmat:
> in the end this feature was probably a waste of time

And therefore I would like to deprecate this form for 6.e.

jnthn:
There's a semantic change in here that we maybe should think about: before, the supply wasn't tapped until you started iterating, whereas now it's tapped at the point you obtain the iterator. Since iterators are one-shot that's probably OK (I don't immediately think of a thing this would break).

jnthn:
> And therefore I would like to deprecate this form for 6.e.

Presuming it's as rarely used as I think (or hope :-)) then we could consider that. And knowing it is going that way would save me a bunch of time figuring out how to model this oddity in RakuAST...

lizmat:
Related to RakuAST: I feel that we should be bringing out a 6.e *before* we really start on the RakuAST endeavour.  Waiting at least another year for a new language version feels too long to me.

jnthn:
This issue is probably not the place, but since you asked:

* I think major versions are primarily marketing driven, and I don't know that we've got a headline for one right now.
* Also I think we probably will want to move off to a new versioning scheme for the next major version. However, I don't want to discuss/debate that now. After the amount of energy the rename took, I want to focus on technical work for at least the next 6 months.

If there's a desire for a `6.d.1` to tide us over, I'm fine with that.

Kaiepi:
*facepalm*

Every benchmark I've been doing recently is dubious at best because I wasn't running enough iterations. This doesn't affect performance to a significant degree. I changed this back to use the proper style for attributes.

src/core.c/Supply-coercers.pm6
```diff
@@ -15,30 +15,61 @@
...
+                nqp::if(
+                  nqp::isconcrete($!exception),
+                  $!exception.throw,
+                  IterationEnd),
+                $got))
+        }
+
+        method push-all(\target --> IterationEnd) {
+            nqp::stmts(
+              nqp::while(
```
lizmat, line 54: Why not `nqp::until` so you can drop the `nqp::not_i` in the condition?

lizmat:
Thanks!

PR #3628, opened 2020-04-15
Add :$chomp to Str.lines()

softmoth:
Behaves like :$chomp in IO::Handle.new

### DIFF
--- a/src/core.c/Str.pm6
+++ b/src/core.c/Str.pm6
@@ -1837,6 +1837,11 @@ my class Str does Stringy { # declared in BOOTSTRAP
           ?? self.lines
           !! self.lines.head($limit)
     }
+    multi method lines(Str:D: $limit, Bool :$chomp! --> Seq:D) {
+        nqp::istype($limit,Whatever) || $limit == Inf
+          ?? self.lines(:$chomp)
+          !! self.lines(:$chomp).head($limit)
+    }
 
     my class Lines does PredictiveIterator {
         has str $!str;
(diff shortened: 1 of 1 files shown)


lizmat:
Thank you for this PR!

I agree that we should probably want to have that capability in `Str.lines`.  However, I really like to see that implemented as a separate iterator, as `lines` is really hot code and we really don't want to make that slower by adding an extra test for each line.  And having each call to `lines` having to process another named parameter, is also something I would like to avoid.  So a candidate without the `:$chomp` and a candidate with a required `:$chomp!` would allow the static optimizer already to select the correct candidate.

Also, if this is about making `indent`, `de-indent` faster, then we should probably be focusing on that, as the current implementation with `given` is screaming for re-implementation *and* there is actually a ticket for that with my name on it, but I'll gladly pass that on to someone else!

softmoth:
I'll update this with a separate :$chomp! candidate.

My focus was not on indent/de-indent, but if it's low-hanging fruit and you want to hand it off to me I can try a fix. I'm just dipping my toe into the pool of rakudo hacking itself.

lizmat:
If you're going to look at indent / de-indent, some pointers to make things faster:

- avoid having to use regexes
- use `nqp::list_s` and `nqp::join` to make things faster, e.g. to indent look first for end-of-line, then within there for the first non-whitespace: if there is any, push that to the `nqp::list_s`, then push the indent string, then push the remainder of the line.  Then at the and, use `nqp::join('', list_s)` to assemble the final string.

lizmat:
Thanks!

PR #3651, opened 2020-05-01
Dispatcher revert

Altai-man:
An attempt to revert dispatcher patches causing a performance loss.

1)There were new ops added to moar, but I don't know if we have to revert them as well and how that affects raw performance.
2)It does pass stresstest with https://github.com/Raku/roast/pull/638

However, when running stresstest numerous times I saw failures in random unrelated files, so the main question is: can dispatcher revert not done properly cause races?

### DIFF
--- a/src/Perl6/Actions.nqp
+++ b/src/Perl6/Actions.nqp
@@ -2720,11 +2720,6 @@ class Perl6::Actions is HLL::Actions does STDActions {
             :op('takedispatcher'),
             QAST::SVal.new( :value('$*DISPATCHER') )
         ));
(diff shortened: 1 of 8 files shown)


vrurg:
Ah, forgot to mention! Moar is likely to be ok without reverting. The primary cause of performance loss – additional variable and non-inlinable op – will be removed with this PR. The extra ops can be removed later.  BTW, this would require removing them from NQP too.

And I would have a personal request too. Could you, pls., notify me prior to merging this PR? I would make some preparations. Thanks!

Altai-man:
@vrurg pinging as promised.
I am going to merge this once you'll green-light it here, so please make the preparations wanted.

vrurg:
@Altai-man I'm ready to go underground! We can merge this PR.

PR #368, opened 2015-02-21
Add version of ok() which handles named $desc and $todo arguments

paultcochrane:
This makes S24-testing/6-done_testing.t pass its tests as well as all ok()
tests in S24-testing/1-basic.t now pass.

PR #3665, opened 2020-05-05
Incorporates @paultcochrane patch for ok with named args.

JJ:
This would close #368.

This passes tests 2 to 5 in this file:

https://github.com/Raku/roast/blob/master/S24-testing/1-basic.t#L27-L31


JJ:
Closing until Raku/problem-solving#188 is cleared.

PR #3668, opened 2020-05-06
Fix .comb(Str) to ensure non-overlapping results

softmoth:
It is expected that `.comb()` returns non-overlapping results, and `.comb('aa')` should yield the same results as `.comb(/aa/)`.


softmoth:
Tests are at https://github.com/Raku/roast/pull/639

lizmat:
Thanks!

PR #3675, opened 2020-05-08
Azure pipelines

patrickbkr:
Switch over to using Microsoft Azure Pipelines instead of CircleCI for creating the binary release builds.
The reason that motivated this move was CircleCI having a bug in their billing which prevented us from doing any Windows builds at all even though their documentation stated this should be possible. Azure Pipelines have unlimited builds on all three platforms. So with this change it should finally be possible to automatically build the binary release builds automatically on all platforms we currently provide builds for.
I have developed and tested this in my clone of the rakudo repository.
A bug in libtommath prevents successful compilation on Windows at the moment though. But that's a separate issue.

This PR also serves as a test whether the changes in the CircleCI config broke anything.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -3,20 +3,6 @@ version: 2.1
 orbs:
     win: circleci/windows@2.2.0
 
-parameters:
-  BUILD_PRECOMP_RELEASE:
-    type: boolean
-    default: false
-  RELEASE_URL:
-    type: string
-    default: ""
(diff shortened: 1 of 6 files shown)


JJ:
Looks like a great idea. Are they totally free?

patrickbkr:
@JJ Yes, they are for open source projects.

JJ:
If no one's against this, I will merge it.

PR #3677, opened 2020-05-09
Symbol lookup optimization

MasterDuke17:
Since a large number of the calls to `find_symbol` are single-element lists, create a new method just for that case and call it where we can. There doesn't seem to be a noticeable change when compiling Rakudo, but I did some micro-benchmarking after the first commit where I only added the new method. Running `use nqp; BEGIN { my $w := $*W;  my $a; $a := $w.find_symbol(nqp::list("Bool")) for ^1_000_000; say $a; say now - ENTER now }; say "hi"` took ~1.1s and ~2.51b instructions. Running `use nqp; BEGIN { my $w := $*W;  my $a; $a := $w.find_single_symbol("Bool") for ^1_000_000; say $a; say now - ENTER now }; say "hi"` took ~1.0s and ~2.41b instructions.

### DIFF
--- a/src/Perl6/World.nqp
+++ b/src/Perl6/World.nqp
@@ -132,7 +132,7 @@ sub levenshtein($a, $b) {
 }
 
 sub make_levenshtein_evaluator($orig_name, @candidates) {
-    my $Str-obj := $*W.find_symbol(["Str"], :setting-only);
+    my $Str-obj := $*W.find_single_symbol("Str", :setting-only);
     my $find-count := 0;
     my $try-count := 0;
     sub inner($name, $hash) {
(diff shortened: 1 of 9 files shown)


lizmat:
This PR does not seem to contain the definition of the new "find_single_symbol" ?

src/Perl6/World.nqp
```diff
@@ -4882,6 +4882,21 @@ class Perl6::World is HLL::World {
         }
     }
 
+    method find_single_symbol_in_setting($name) {
+        my $no-outers := 0; # If 'true' then don't look in the outer contexts
```
lizmat, line 4886: Is this some kind of debugging aid?  Shouldn't this need to go?

src/Perl6/World.nqp
```diff
@@ -4882,6 +4882,21 @@ class Perl6::World is HLL::World {
...
+
+        my $setting_name := Perl6::ModuleLoader.transform_setting_name($!setting_name);
+        my $ctx := Perl6::ModuleLoader.load_setting($setting_name);
+
+        while $ctx {
```
lizmat, line 4891: if `$no-outers` is gone, wouldn't this loop be completely unnecessary?

MasterDuke17: Whoops, `$no-outers` was missed when I copied out the single-element logic from `find_symbol_in_setting`. The loop is still necessary, since it checks outer contexts until there aren't any left.

PR #3690, opened 2020-05-13
add IO::Handle.do-not-close-at-exit() method

dwarring:
Which is currently a private method. For when file-handles are closed outside of Raku

```
     my UInt $fd = $ioh.native-descriptor();
     $ioh.forget-about-closing($fd);
```
Easiest solution to solving failing t/12html.t tests in latest LibXML https://github.com/libxml-raku/LibXML-raku/commit/cf4ea76fd3adb45962bc49fc062972e5069ce9e5

After some bug-fixes `LibXML.parse(:$fd)` and `LibXML::parse(:$io)` are being closed by libxml2. Causing errors because Rakudo is also trying to close the same file-handles.

Hoping this is a suitable fix and others might find this method handy.

### DIFF
--- a/src/core.c/IO/Handle.pm6
+++ b/src/core.c/IO/Handle.pm6
@@ -37,7 +37,7 @@ my class IO::Handle {
             )
         }
     }
-    method forget-about-closing(int $fileno --> Nil) {
+    method !forget-about-closing(int $fileno --> Nil) {
         $opened-locker.protect: {
             nqp::bindpos($opened,$fileno,nqp::null)
         }
@@ -232,7 +232,7 @@ my class IO::Handle {
             (my int $fileno = nqp::filenofh($!PIO)),
             nqp::closefh($!PIO), # TODO: catch errors
             ($!PIO := nqp::null),
-            self.forget-about-closing($fileno)
+            self!forget-about-closing($fileno)
 #?endif
           )
         )
@@ -631,7 +631,7 @@ my class IO::Handle {
         Bool:D :$non-blocking = False, Bool:D :$shared = False --> True
     ) {
 #?if moar
-        self.forget-about-closing(nqp::filenofh($!PIO));
+        self!forget-about-closing(nqp::filenofh($!PIO));
 #?endif
         nqp::lockfh($!PIO, 0x10*$non-blocking + $shared);
         CATCH { default {
@@ -885,7 +885,7 @@ my class IO::Handle {
 #?endif
 #?if moar
             ($!PIO := nqp::null),
-            self.forget-about-closing($fileno)
+            self!forget-about-closing($fileno)
 #?endif
           )
         )
--- a/src/core.c/IO/Handle.pm6
+++ b/src/core.c/IO/Handle.pm6
@@ -871,6 +871,10 @@ my class IO::Handle {
         }
     }
 
+    method do-not-close-at-exit(IO::Handle:D:) {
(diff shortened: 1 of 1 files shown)


niner:
An alternative (and backwards compatible solution) for LibXML would be to 
dup(2) the descriptor and pass that to LibXML.open

lizmat:
Do we have a Raku way of doing that?

lizmat:
Agree that duping would be a backwards compatible solution.

But, if LibXML wants to close the PIO when its done, wouldn't the unduped PIO hang around when it is really not needed anymore?  And thus take up valuable resources?

niner:
There's always NativeCall to get at libc functions ;)

The IO::Handle destructor will close the original PIO at some point which is 
as I understand what's causing the issue in LibXML

lizmat:
Right, but if we pass on the PIO to LibXML to close, and mark it not to be closed by Raku, wouldn't that be a better use of resources in the end?

niner:
Yes, of course. I'm not sure this PR is the right way to do this though. The nice things about private methods is that you don't have to think about whether it's a good API, as you are your only user. So to be safe, we have to assume that jnthn++ didn't waste much time thinking about other users.

It currently takes an int as that is conveniently available in `IO::Handle` code and lends itself to an efficient implementation of a list of file descriptors to close. What a user wants to express on the other hand is "I got this file handle and take over responsibility for closing it". This suggests creating an object method on the file handle, not a class method taking a very low level value.

The current private `!forget-about-closing` method lets you delete some arbitrary entry from the `$opened` list, regardless of who opened it and how. Do we really want that? The strangeness is even visible in the proposed change to LibXML-raku. It's got a `parse` multi candidate that takes an `UInt :$fd` and would pass that to `IO::Handle.forget-about-closing` without actually knowing where it comes from. Maybe it originated in some native function and has never even been processed by IO::Handle?

A curious thing is that libXML's `xmlReadFd` function doesn't actually close the file: http://www.xmlsoft.org/html/libxml-parser.html#xmlReadFd
It's libXML-raku's code that does.

dwarring:
The test which is giving failures calls `htmlNewParserCtxt`, which does install callbacks to close the file from within libxml.

dwarring:
Just to be a be a bit clearer. You can add callbacks via the Raku bindings, such as test https://github.com/libxml-raku/LibXML-raku/blob/master/t/17callbacks.t.

But the test in question https://github.com/libxml-raku/LibXML-raku/blob/master/t/12html.t isn't.
In this case, libxml provides open, read and close callbacks for you.

jnthn:
I've leave the method private and expose it as something like `method do-not-close-at-exit(IO::Handle:D:) { self!forget-about-closing(self.native-descriptor) }`, which is long, but this needn't be huffmanized at all, since it's very rarely needed.

dwarring:
^^ Have reworked it as do-not-close-at-exit() method. Also tweaked DESTROY method to avoid closing these.

lizmat:
Working on implementing that, while doing some other minor cleanup

lizmat:
https://github.com/rakudo/rakudo/commit/61046d7695 adds a `do-not-close-automatically`.  Calling that method on an `IO::Handle` will ensure that **only** an explicit call to `close` will close the low-level OS handle.  The `DESTROY` method, when called, will not do that, nor will this be done on exit of the program for this `IO::Handle`.

@dwarring: thanks for the PR, hope this solution is usable for your projects.

Issue #3693, opened 2020-05-14
INSTALL.txt out of date


coke:
Joseph Brenner noted on perl6-users that INSTALL.txt is out of date; it says you can use ./perl6 to test once you've completed the "make" step - pretty sure you need to do a 'make install' and then use ./install/bin/perl6 - someone should run through the directions and validate that they work, updating as needed.

vrurg:
It is outdated indeed.

vrurg:
Closed via #3696

PR #3696, opened 2020-05-16
Update INSTALL.txt for the latest changes

vrurg:
Reflect some changes in the build process and, eventually, replace all
Perl6 references.

### DIFF
--- a/INSTALL.txt
+++ b/INSTALL.txt
@@ -1,9 +1,9 @@
   Build requirements (Installing from source)
 
     For building Rakudo you need at least a C compiler, a "make" utility, and
-    Perl 5.10.1 or newer. To automatically obtain and build MoarVM as well as
-    NQP, you may also need a git client, which is also needed for fetching the
-    test suite.
+    Perl 5.10.1 or newer on all platforms except Windows which needs Perl 5.22.
+    To automatically obtain and build MoarVM as well as NQP, you may also need a
+    git client, which is also needed for fetching the test suite.
 
     Building rakudo can sometimes take >1.4G of memory when compiling for the 
     MoarVM runtime. The requirements are higher for the JVM backend.
--- a/INSTALL.txt
+++ b/INSTALL.txt
@@ -1,11 +1,12 @@
   Build requirements (Installing from source)
-    For building Rakudo you need at least a C compiler, a "make" utility, and
-    Perl 5.10 or newer. To automatically obtain and build MoarVM as well as NQP,
-    you may also need a git client, which is also needed for fetching the test
-    suite.
 
-    Building rakudo can take up to 1G of memory when compiling for the MoarVM
-    runtime. The requirements are higher for the JVM backend.
+    For building Rakudo you need at least a C compiler, a "make" utility, and
+    Perl 5.10.1 or newer. To automatically obtain and build MoarVM as well as
+    NQP, you may also need a git client, which is also needed for fetching the
+    test suite.
+
+    Building rakudo can sometimes take >1.4G of memory when compiling for the 
+    MoarVM runtime. The requirements are higher for the JVM backend.
 
     (Perl is installed by default already). To enable parallel testing you also
     need the CPAN module Test::Harness in version 3.16 or newer; you can control
@@ -13,7 +14,7 @@
     TEST_JOBS is not specified, 6 jobs will be used.
 
   Building and invoking Rakudo
-    If you're wanting the bleeding-edge version of the Rakudo Perl 6 compiler,
+    If you're wanting the bleeding-edge version of the Rakudo Raku compiler,
     we recommend downloading Rakudo directly from Github and building it from
     there.
 
@@ -123,9 +124,8 @@
     "roast" repository <http://github.com/Raku/roast/> and run all of these
     tests that are currently known to pass.
 
-    At present we do not have any plans to directly store the official test
-    suite as part of the Rakudo repository, but will continue to fetch it from
-    the roast repository.  Releases of Rakudo get a snapshot of the roast
+    Roast is not planned and unlikely to ever be included into Rakudo 
+    distribution. Instead releases of Rakudo will fetch a snapshot of the roast 
     repository as of the time of the release.
 
     You can also use "make" to run an individual test from the command line:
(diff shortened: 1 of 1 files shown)


vrurg:
Resolves #3693

INSTALL.txt
```diff
@@ -35,92 +34,79 @@
     or:
         $ perl Configure.pl --gen-nqp --backends=jvm  # needs JDK 1.8 installed
     then:
-        $ make
+        $ make install
```
Altai-man, line 37: Not `make; make install`, but just `make install`? I am not sure what works for rakudo, just seeing the former often.

vrurg: I use `make install` always when installation is the final target.

INSTALL.txt
```diff
@@ -1,30 +1,29 @@
...
+    (Perl is installed by default already). To enable parallel testing you also
+    need the CPAN module Test::Harness in version 3.16 or newer; you can control
+    the number of parallel jobs with the "TEST_JOBS" environment variable. If
+    TEST_JOBS is not specified, 6 jobs will be used.
 
   Building and invoking Rakudo
-    If you're wanting the bleeding-edge version of the Rakudo Perl 6
-    compiler, we recommend downloading Rakudo directly from Github
-    and building it from there.
+    If you're wanting the bleeding-edge version of the Rakudo Perl 6 compiler,
```
patrickbkr, line 16: There's another _Perl 6_ lurking here.

INSTALL.txt
```diff
@@ -133,28 +119,31 @@
...
+    Running "make spectest" will import the official Raku test suite from the
+    "roast" repository <http://github.com/Raku/roast/> and run all of these
+    tests that are currently known to pass.
 
     At present we do not have any plans to directly store the official test
-    suite as part of the Rakudo repository, but will continue to fetch it
-    from the roast repository.  Releases of Rakudo get a snapshot of
-    the roast repository as of the time of the release.
+    suite as part of the Rakudo repository, but will continue to fetch it from
+    the roast repository.  Releases of Rakudo get a snapshot of the roast
```
patrickbkr, line 128: This sentence reads as if roast would be bundled with releases. AFAIK this is not the case.
What's probably meant is: "Releases of Rakudo will fetch a snapshot of the roast repository as of the time of the release."

INSTALL.txt
```diff
@@ -1,30 +1,29 @@
...
-    as well as NQP, you may also need a git client, which is also needed
-    for fetching the test suite.
+    For building Rakudo you need at least a C compiler, a "make" utility, and
+    Perl 5.10 or newer. To automatically obtain and build MoarVM as well as NQP,
+    you may also need a git client, which is also needed for fetching the test
+    suite.
 
-    Building rakudo can take up to 1G of memory when compiling for the
-    MoarVM runtime. The requirements are higher for the JVM backend.
+    Building rakudo can take up to 1G of memory when compiling for the MoarVM
```
patrickbkr, line 7: The requirements have increased to >1.4GB for the MoarVM backend.

INSTALL.txt
```diff
@@ -1,30 +1,29 @@
   Build requirements (Installing from source)
-    For building Rakudo you need at least a C compiler, a "make" utility,
-    and Perl 5.10 or newer. To automatically obtain and build MoarVM
-    as well as NQP, you may also need a git client, which is also needed
-    for fetching the test suite.
+    For building Rakudo you need at least a C compiler, a "make" utility, and
+    Perl 5.10 or newer. To automatically obtain and build MoarVM as well as NQP,
```
patrickbkr, line 3: On Windows we now require >= Perl 5.22 ([370956](https://github.com/rakudo/rakudo/commit/370956b913ce5eac1d89a17626ebbdf99cc0170a)).

INSTALL.txt
vrurg: This was confusing me from the beginning. Actually, the whole paragraph needs to be changed.

INSTALL.txt
vrurg: Did you see [my comment](https://github.com/Raku/nqp-configure/issues/19#issuecomment-627672367)? With simplified nmake detection this could now be less of a problem, but the build sample provided by @Altai-man doesn't contain any nmake output at all! If anybody can confirm that 5.18 doesn't have the problem I'd consider sticking to that version for a while.

Besides, I don't see a reason why older versions of Perl won't work for Windows.

What certainly makes sense is changing to 5.10.1 as this is what stated in Config.pm.

INSTALL.txt
vrurg: @patrickbkr Thanks! Took your comments into account with e1db95e. Except for the Perl version at least for now.

INSTALL.txt
patrickbkr: The reason the dependency was bumped to 5.22 for Windows is the use of the [pipe form of open](https://github.com/rakudo/rakudo/blob/master/tools/build/update-submodules.pl#L71) which is only supported in Perl 5.22 and later on Windows. I think that's a reason unrelated to the one you had in mind.

vrurg: I'm not really happy about bumping to 5.22 only for the sake of a single `open` call. Can we get around it?

niner: Bumping to 5.22 for that sounds harsh, but I also wonder if this would ever cause a problem. 5.22 is 5 years old already. On Windows there's no such thing as "system perl" that is stuck on some ancient version and still gets support for another 5 years. Every perl on Windows has been installed manually and will probably have been the most recent version at that time. So everything **installed** during the last 5 years will be OK.

vrurg: The problem is that the script is used by every build, not only Windows. But as far as it is Windows only requirement a specific note can be added.

@patrickbkr what would you say?

patrickbkr: It is a Windows only requirement. `open` with pipe is a feature that's been in Perl for a long time. Only on Windows the support was added in 5.22.

There was quite a lot of discussion and exploration of different solutions for this very line of code, all having their tradeoffs (if there is interest I can dig up the discussion). In the end we settled on this solution with the tradeoff of having to bump the Perl version on Windows. So from what I remember there is no much nicer solution.

vrurg: Ok, updated with Perl 5.22 for Windows.

Heh, I was missing named capture call in Perl regexes while implementing macros...

vrurg:
Ok, if nobody else request for changes, I'm going to eventually merge it tomorrow.

Issue #3701, opened 2020-05-18
Handle expanded values for run option "--doc" 


tbrowder:
Currently the only way to affect pod leading declarator block text handling is with an environment variable.

It would be much more convenient for the user to select that behavior by adding an extra run option "--doc-fmt=" so one could run Raku something like this:

    raku --doc --doc-fmt=preserve ...

or 

    raku --doc=HTML --doc-fmt=preserve ...

coke:
Is it worth abbreviating  as "auth-fmt" instead of using "author-format" or perhaps "original" ?

tbrowder:
@lizmat and @vrurg have already weighed in on the PR I'm working on. The new PR will use a new option with one argument for the current situation. Example use:

    $ raku --doc --doc-fmt=preserve ...

where 'preserve' means "keep original text format in declarator blocks.

tbrowder:
A new PR is the way and will cover most of this discussion.

PR #3708, opened 2020-05-21
Add 'auth-fmt' compiler arg for option '--doc'

tbrowder:
Addresses Rakudo issue GH #3701.

This commit adds the capability for the user to affect parsing of Pod
in leading declarator blocks.

The standard behavior is to extract such Pod as a single line of
normalized text. A previous commit added the capability to change such
parsing to respect the original formatting by saving the author's
newlines and other spaces, but it had to be triggered by the existence
of a special Rakudo environment variable:

  RAKUDO_POD_DECL_BLOCK_USER_FORMAT

With this commit, the user can now select the same behavior by adding
the word 'auth-fmt' after the command line option '--doc='.  The new
word can be added alone or added to the existing argument choices of
'Text' or 'HTML'. The two-word version must have the words separated
by a comma (with no spaces) and in any order.

The compiler has been modified so the '--help' option explains the new
capability.

Files modified:

+ src/Perl6/Compiler.nqp
+ src/Perl6/Grammar.nqp
+ src/main.nqp

### DIFF
--- a/src/Perl6/Compiler.nqp
+++ b/src/Perl6/Compiler.nqp
@@ -49,9 +49,65 @@ class Perl6::Compiler is HLL::Compiler {
     }
 
     method command_eval(*@args, *%options) {
-        if nqp::existskey(%options, 'doc') && !%options<doc> {
-            %options<doc> := 'Text';
-        }
+        # Implementing 'fmt' option with '--doc='.
+        # Check for additional args tacked on to it.
+        if nqp::existskey(%options, 'doc') {
+            if !%options<doc> {
+                # currently this the default handler
+                %options<doc> := 'Text';
+            }
+            else {
+                # check actual values
+                my $f := nqp::atkey(%options,'doc');
+                # could have a combined string joined by a comma
+                unless nqp::isstr($f) {
+                    nqp::die("FATAL: non-string value type for option '--doc='");
+                }
+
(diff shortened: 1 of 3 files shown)


lizmat:
Nice, but the name `auth-fmt` feels LTA, mnemonic wise.  I assume you mean: "format as intended by the author" ?

tbrowder:
Yes, originally I was going to use 'fmt' but I have no qualms about changing. In fact, I really wanted to discuss the whole thing a bit more first, but bigger fish were being fried. I consider this a "strawman" that's easy to close and redo.

tbrowder:
Per @lizmat's comment, I'm adding other choices for the argument in place of 'auth-fmt'. Give a thumbs up for your choice or add a new suggestion

    fmt

tbrowder:
ufmt

tbrowder:
**IMPORTANT**

I don't want to merge until there is some consensus on an acceptable name. It is easy to close this PR and resubmit with a more acceptable name. Merging prematurely would be a mistake in my opinion.

vrurg:
Why `--doc`? Are we sparing options for the future? Wouldn't `--doc-fmt=original` or something like this be more up to the task? Besides, it'd give more freedom for future extensions.

tbrowder:
I'm closing this PR and will submit its replacement shortly.

PR #3709, opened 2020-05-22
Add '--doc-fmt=preserve' compiler option

tbrowder:
Addresses Rakudo issue GH #3701.

This commit adds the capability for the user to affect parsing of text
in leading declarator blocks.

The standard behavior is to extract such text as a single line of
normalized text. A previous commit added the capability to change such
parsing to respect the original formatting by saving the author's
newlines and other spaces, but it had to be triggered by the existence
of a special Rakudo environment variable:

  RAKUDO_POD_DECL_BLOCK_USER_FORMAT

With this commit, the user can now select the same behavior by adding
the new option '--doc-fmt=preserve' option along with other
'--doc' options.

The compiler has been modified so the '--help' option explains the new
option.

As an example consider this short Raku program in file 't.raku':

    #| line 1: foo
    #| line 2: bar
    sub baz() {}
    #= line 3: baz
    #= line 4: boo

Run using current usage:

    $ raku --doc t.raku
    sub baz()
    line 1: foo line 2: bar
    line 3: baz line 4: boo

Run with the new option:

    $ raku --doc --doc-fmt=preserve t.raku
    sub baz()
    line 1: foo
    line 2: bar
    line 3: baz line 4: boo

Notice the trailing declarator block hasn't been fixed yet. That is a
separate issue.

Files modified:

+ src/Perl6/Compiler.nqp
+ src/Perl6/Grammar.nqp
+ src/main.nqp

### DIFF
--- a/src/Perl6/Compiler.nqp
+++ b/src/Perl6/Compiler.nqp
@@ -52,6 +52,14 @@ class Perl6::Compiler is HLL::Compiler {
         if nqp::existskey(%options, 'doc') && !%options<doc> {
             %options<doc> := 'Text';
         }
+        if nqp::existskey(%options, 'doc-fmt') {
+            # check for valid arg
+            my $ok-arg := 'preserve';
+            my $arg    := %options<doc-fmt>;
+            unless nqp::iseq_s($arg,$ok-arg) {
+                self.panic("Unknown '--doc-fmt' arg '$arg', did you mean '$ok-arg'?");
+            }
+        }
 
         my $argiter := nqp::iterator(@args);
         nqp::shift($argiter) if $argiter && !nqp::defined(%options<e>);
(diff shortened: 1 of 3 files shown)


vrurg:
I would suggest having one more value alongside with `preserve` for the case if preserving becomes the default in the future and one wants to continue using current default. Could be something like `legacy`, or `flatten`, or something else meaningful.

tbrowder:
Yes, I have been working on that in another branch. Now that we have Damian's input I may close this PR and wrap all into another PR since this one does have not quite the correct behavior.

But I do like your choices for the opposite effect. I was think of 'Christmas' or 'v6c' or somthing like that.

tbrowder:
I will close this and combine with other work in a new PR.

Issue #3635, opened 2020-04-19
Hardwired /usr/bin/env breaks in some environments


lembark:
## The Problem

Line 53:
    my $perl_wrapper = '#!/usr/bin/env #perl#

/usr/bin/env does not exist on all *NIX platforms; the absolute path may also be '/bin/env'.

## Expected Behavior

Generating #! code that references an extant "env" executable.

e.g., checking if the absolute path exists prior to inserting it w/ a terniary operator prior to writing the file.

## Actual Behavior

Generating #! code that blows up due to a non-existant /usr/bin/env.

## Steps to Reproduce

Install, for example, zef and notice that ./bin/zef begins with "#!/usr/bin/env"

## Environment
* Operating system: Linux
* Distribution: Gentoo 

* Compiler version (`perl6 -v`): 

This is Rakudo version 2020.02.1 built on MoarVM version 2020.02.1
implementing Raku 6.d.
Raku::source-digest=004deb157fb8c3c0506108049265d0f68ab84f15

lembark:
e.g., 

state $env_path =  |</usr/bin/env /bin/env>.first( { "$_".IO.e } );

niner:
Can you please submit a PR with that change?

lembark:
Working on it. Basic iteration is to take the first of /bin/env or /usr/bin/env, lacking that find an 
absolute path to rakudo or give up.

lembark:
If you want to believe that all reasonable systems will have env on them this is trivial; the messy part is finding a solution using $.prefix to locate bin/rakudo where env doesn't exist.

PR #3716, opened 2020-05-24
Perl wrapper shebang

lembark:
Per issue #3635 <https://github.com/rakudo/rakudo/issues/3635>: If /usr/bin/env does not exist then the perl wrappers fail. Fix is to test for both /bin/env and  /usr/bin/env, using the first existing one.

As added boilerplate, if neither is found -- old enough System7?? -- then the install prefix path is walked to locate any 'bin/rakudo' executable which is then used with an absolute path as "#!/path/to/bin/rakudo". Lacking both env and rakudo executables this dies.

This replaces the literal "#perl#" in the following loop with "#ext#" so that instead of inserting "perl6-j", etc, it simply adds the extension via  "#!/usr/bin/ext rakudo#ext#" or "#!/path/to/bin/rakudo#ext#". This avoids hardwiring "perl6" as the basename and gracefully handles both /bin/env with a basename and /path/to/bin/rakudo with an absolute path.

Note: I'm unable to test this on Windows here. 

One thing this does *not* do is deal with missing rakudo-X basenames (e.g., producing the wrapper zef-js where no rakudo-js exists on the filesystem). This seems like a useful later addition.

### DIFF
--- a/src/core.c/CompUnit/Repository/Installation.pm6
+++ b/src/core.c/CompUnit/Repository/Installation.pm6
@@ -245,7 +245,6 @@ sub MAIN(:$name, :$auth, :$ver, *@, *%)
     CompUnit::RepositoryRegistry.run-script( "#name#", :$name, :$auth, :$ver );
 }
 SHELL
-
        };
 
         for %files.kv -> $name-path, $file is copy {
--- a/src/core.c/CompUnit/Repository/Installation.pm6
+++ b/src/core.c/CompUnit/Repository/Installation.pm6
@@ -201,7 +201,7 @@ __END__
 
         # bin/ scripts
 
-        my $perl-wrapper
+        state $perl-wrapper
         = do {
             # code below replaces #ext# with '-j', '-m', etc.
             #
--- a/src/core.c/CompUnit/Repository/Installation.pm6
+++ b/src/core.c/CompUnit/Repository/Installation.pm6
@@ -259,7 +259,7 @@ SHELL
                     $perl-wrapper.subst('#name#', $name, :g).subst('#ext#', $be );
                 if $is-win {
                     $.prefix.add("$withoutext$be.bat").IO.spurt:
-                        $windows_wrapper.subst('#perl#', "perl6$be", :g);
+                        $windows_wrapper.subst('#ext#', $be, :g);
                 }
                 else {
                     $.prefix.add("$withoutext$be").IO.chmod(0o755);
--- a/src/core.c/CompUnit/Repository/Installation.pm6
+++ b/src/core.c/CompUnit/Repository/Installation.pm6
@@ -50,10 +50,6 @@ goto endofperl
 __END__
 :endofperl
 ';
-    my $perl_wrapper = '#!/usr/bin/env #perl#
-sub MAIN(:$name, :$auth, :$ver, *@, *%) {
-    CompUnit::RepositoryRegistry.run-script("#name#", :$name, :$auth, :$ver);
-}';
 
     method !sources-dir   { with $.prefix.add('sources')   { once { .mkdir unless .e }; $_ } }
     method !resources-dir { with $.prefix.add('resources') { once { .mkdir unless .e }; $_ } }
@@ -204,6 +200,66 @@ sub MAIN(:$name, :$auth, :$ver, *@, *%) {
         }
 
         # bin/ scripts
+
+        my $perl-wrapper
+        = do 
+        {
+            # basename being dispatched is generated below, replaces '#perl#'
+            # in the generated script.
+            # question here is how to dispatch it: via env or the path to
+            # a rakudo executable.
+
+            # for whatever reason heredoc (:to/X/) fails.
+            # use the literal text for now.
+            # someone should clean this up in the future.
+
+            constant FIXED  = q|
+sub MAIN(:$name, :$auth, :$ver, *@, *%)
+{
+    CompUnit::RepositoryRegistry.run-script("#name#", :$name, :$auth, :$ver);
+}
+|;
+            my $exec = gather 
+            {
+                # change 'perl6' to 'rakudo' or 'raku' at some point.
+                constant SEARCH = 'bin/perl6';
+
+                with < /bin/env /usr/bin/env >.first: { .IO.e } -> $env 
+                {
+                    take "$env #perl#";
+                }
+                else
+                {
+                    warn q{System lacks 'env', using } ~ SEARCH;
+
+                    my $curr    = $.prefix.absolute;
+
+                    loop:
+                    {
+                        if $curr.add( SEARCH ).IO.ie
+                        {
+                            take "$curr/bin/#perl#";
+                            last
+                        }
+                        elsif $curr eq ( my $next = $curr.parent )
+                        {
+                            # one last option would be "$SHELL -e $0", but that
+                            # requirs shell-speicfic arguments. rahter than open
+                            # that can of worms here, it seems better to give up.
+
+                            die "No '{SEARCH}' above {$.prefix}";
+                        }
+                        else
+                        {
+                            $curr   = $next;
+                        }
+                    }
+                }
+            };
+
+            ( qq{#!$exec}, FIXED ).join( "\n" )
+        };
+
         for %files.kv -> $name-path, $file is copy {
             next unless $name-path.starts-with('bin/');
             my $name        = $name-path.subst(/^bin\//, '');
@@ -212,7 +268,7 @@ sub MAIN(:$name, :$auth, :$ver, *@, *%) {
             my $withoutext  = $name-path.subst(/\.[exe|bat]$/, '');
             for '', '-j', '-m', '-js' -> $be {
                 $.prefix.add("$withoutext$be").IO.spurt:
-                    $perl_wrapper.subst('#name#', $name, :g).subst('#perl#', "perl6$be");
+                    $perl-wrapper.subst('#name#', $name, :g).subst('#perl#', "perl6$be");
                 if $is-win {
                     $.prefix.add("$withoutext$be.bat").IO.spurt:
                         $windows_wrapper.subst('#perl#', "perl6$be", :g);
--- a/src/core.c/CompUnit/Repository/Installation.pm6
+++ b/src/core.c/CompUnit/Repository/Installation.pm6
@@ -202,8 +202,7 @@ __END__
         # bin/ scripts
 
         my $perl-wrapper
-        = do 
-        {
+        = do {
             # basename being dispatched is generated below, replaces '#perl#'
             # in the generated script.
             # question here is how to dispatch it: via env or the path to
@@ -213,12 +212,12 @@ __END__
             # use the literal text for now.
             # someone should clean this up in the future.
 
-            constant FIXED  = q|
+            constant FIXED  = q:to/SHELL/;
 sub MAIN(:$name, :$auth, :$ver, *@, *%)
 {
     CompUnit::RepositoryRegistry.run-script("#name#", :$name, :$auth, :$ver);
 }
-|;
+SHELL
             my $exec = gather 
             {
                 # change 'perl6' to 'rakudo' or 'raku' at some point.
@@ -226,7 +225,7 @@ sub MAIN(:$name, :$auth, :$ver, *@, *%)
 
                 with < /bin/env /usr/bin/env >.first: { .IO.e } -> $env 
                 {
-                    take "$env #perl#";
+                    take "$env rakudo";
                 }
                 else
                 {
@@ -234,21 +233,20 @@ sub MAIN(:$name, :$auth, :$ver, *@, *%)
 
                     my $curr    = $.prefix.absolute;
 
-                    loop:
+                    loop: 
                     {
-                        if $curr.add( SEARCH ).IO.ie
+                        if $curr.add( SEARCH ).IO.ie 
                         {
-                            take "$curr/bin/#perl#";
+                            take "$curr/bin/rakudo";
                             last
-                        }
-                        elsif $curr eq ( my $next = $curr.parent )
+                        } elsif $curr eq ( my $next = $curr.parent ) 
                         {
                             # one last option would be "$SHELL -e $0", but that
                             # requirs shell-speicfic arguments. rahter than open
                             # that can of worms here, it seems better to give up.
 
                             die "No '{SEARCH}' above {$.prefix}";
-                        }
+                        } 
                         else
                         {
                             $curr   = $next;


lizmat:
Good to see your first PR!   Hope it will work out with @patrickbkr and @vrurg

lembark:
Any suggestions on how to actually *test* this thing would be appreciated. Is there any stage of building that installing a fake wrapper would be acceptable?

src/core.c/CompUnit/Repository/Installation.pm6
```diff
@@ -204,6 +200,53 @@ sub MAIN(:$name, :$auth, :$ver, *@, *%) {
...
+                        } else {
+                            $curr   = $next;
+                        }
+                    }
+                }
+            };
+
+            # Q: why does this fail with the here-doc indented? 
+
+            ( qq{#!$exec#ext#}, q:to/SHELL/ ).join( "\n" )
```
ugexe, line 241: Shelling out to do this is way to fragile. Just remove it.

ugexe: Additionally we don't want to pay the penalty of calculating this every time an installed script is run.

src/core.c/CompUnit/Repository/Installation.pm6
```diff
@@ -204,6 +200,53 @@ sub MAIN(:$name, :$auth, :$ver, *@, *%) {
...
+            # #!/bin/sh -e but that seems to be going too far: give up.
+
+            my $exec = gather {
+                with < /bin/env /usr/bin/env >.first: { .IO.e } -> $env {
+                    take "$env rakudo";
+                } else {
+                    constant SEARCH = 'bin/rakudo';
+                    my $curr    = $.prefix.absolute.IO;
+
+                    note "# Perl-wrapper lacks 'env', using {SEARCH}";
```
ugexe, line 222: I'm not sure debug output is appropriate for this

src/core.c/CompUnit/Repository/Installation.pm6
```diff
@@ -204,6 +200,53 @@ sub MAIN(:$name, :$auth, :$ver, *@, *%) {
...
+
+                    loop {
+                        my $raku    = $curr.add( SEARCH );
+
+                        if $raku.f {
+                            take $raku;
+                            last
+                        } elsif $curr eq ( my $next = $curr.parent ) {
+                            die "No '{SEARCH}' above {$.prefix}";
```
ugexe, line 231: I'm not sure dying is appropriate

src/core.c/CompUnit/Repository/Installation.pm6
```diff
@@ -204,6 +200,53 @@ sub MAIN(:$name, :$auth, :$ver, *@, *%) {
...
+                with < /bin/env /usr/bin/env >.first: { .IO.e } -> $env {
+                    take "$env rakudo";
+                } else {
+                    constant SEARCH = 'bin/rakudo';
+                    my $curr    = $.prefix.absolute.IO;
+
+                    note "# Perl-wrapper lacks 'env', using {SEARCH}";
+
+                    loop {
+                        my $raku    = $curr.add( SEARCH );
```
ugexe, line 225: A CUR should not assume it knows where the perl executable is. If anything it should look at `$*EXECUTABLE`

ugexe:
It would seem to me a simpler solution is sufficient:
`my $shebang = do with </bin/env /usr/bin/env>.first(*.IO.e) { "#!$_" } else { "" };`

lembark:
On Sun, 24 May 2020 13:17:40 -0700
Nick Logan <notifications@github.com> wrote:

> 
> > +  
> +                        if $raku.f {
> +                            take $raku;
> +                            last
> +                        } elsif $curr eq ( my $next = $curr.parent )
> {
> +                            die "No '{SEARCH}' above {$.prefix}";
> +                        } else {
> +                            $curr   = $next;
> +                        }
> +                    }
> +                }
> +            };
> +
> +            # Q: why does this fail with the here-doc indented? 
> +
> +            ( qq{#!$exec#ext#}, q:to/SHELL/ ).join( "\n" )
> 
> Shelling out to do this is way to fragile. Just remove it.
> 

Removing the #! line from the wrapper will leave it un-exeutable.

The basic wrapper has to have the format:

    #!/absolute/path/to/executable args

in this case it'll be one of:

    #!//bin/env rakudo
    #!/usr/bin/env rakudo
    #!/path/to/bin/rakudo

-- 
Steven Lembark
Workhorse Computing
lembark@wrkhors.com
+1 888 359 3508

lembark:
On Sun, 24 May 2020 13:18:23 -0700
Nick Logan <notifications@github.com> wrote:

> @ugexe commented on this pull request.
> 
> 
> 
> > +  
> +                        if $raku.f {
> +                            take $raku;
> +                            last
> +                        } elsif $curr eq ( my $next = $curr.parent )
> {
> +                            die "No '{SEARCH}' above {$.prefix}";
> +                        } else {
> +                            $curr   = $next;
> +                        }
> +                    }
> +                }
> +            };
> +
> +            # Q: why does this fail with the here-doc indented? 
> +
> +            ( qq{#!$exec#ext#}, q:to/SHELL/ ).join( "\n" )
> 
> Additionally we don't want to pay the penalty of calculating this
> every time an installed script is run.
> 

OK, I've replaced the my with a state.
Other than that the only way to avoid computing it is to save the 
wrapper content in a separate module or config file at install
time.


-- 
Steven Lembark
Workhorse Computing
lembark@wrkhors.com
+1 888 359 3508

lembark:
On Sun, 24 May 2020 13:19:39 -0700
Nick Logan <notifications@github.com> wrote:

> @ugexe commented on this pull request.
> 
> 
> 
> > +            #  
> +            # question here is how to dispatch it: via env or
> absolute
> +            # path to a rakudo executable. this will use env in every
> +            # case of *NIX I know of; the bin/rakudo loop is
> boilerplate.
> +            #
> +            # lacking both env and raukdo we could try something with
> +            # #!/bin/sh -e but that seems to be going too far: give
> up. +
> +            my $exec = gather {
> +                with < /bin/env /usr/bin/env >.first: { .IO.e } ->
> $env {
> +                    take "$env rakudo";
> +                } else {
> +                    constant SEARCH = 'bin/rakudo';
> +                    my $curr    = $.prefix.absolute.IO;
> +
> +                    note "# Perl-wrapper lacks 'env', using
> {SEARCH}";
> 
> I'm not sure debug output is appropriate for this

Every *NIX system I know of has 'env' at this point. Unless someone
is installing rakudo in an embedded system or something else really
odd the branch for finding bin/rakudo will not be exercised. I 
originally had this as a warning: If env is not found on the local
system it should usually be reported.

Q: Is there any verbose/debug switch available in the object 
(e.g., sibling to $.prefix) that could use to switch the output?


-- 
Steven Lembark
Workhorse Computing
lembark@wrkhors.com
+1 888 359 3508

lembark:
On Sun, 24 May 2020 13:20:23 -0700
Nick Logan <notifications@github.com> wrote:

> > +                with < /bin/env /usr/bin/env >.first: { .IO.e } ->
> > $env {  
> +                    take "$env rakudo";
> +                } else {
> +                    constant SEARCH = 'bin/rakudo';
> +                    my $curr    = $.prefix.absolute.IO;
> +
> +                    note "# Perl-wrapper lacks 'env', using
> {SEARCH}"; +
> +                    loop {
> +                        my $raku    = $curr.add( SEARCH );
> +
> +                        if $raku.f {
> +                            take $raku;
> +                            last
> +                        } elsif $curr eq ( my $next = $curr.parent )
> {
> +                            die "No '{SEARCH}' above {$.prefix}";
> 
> I'm not sure dying is appropriate

If the wrapper requested cannot be installed at all what would
you suggest?

Lacking an absoltue path to both env and rakudo there can be no
#! line in the wrapper.



-- 
Steven Lembark
Workhorse Computing
lembark@wrkhors.com
+1 888 359 3508

ugexe:
> Lacking an absoltue path to both env and rakudo there can be no #! line in the wrapper.

I think no shebang in that scenario is reasonable. Considering that, would the following not be sufficient?

```
my $shebang = do with </bin/env /usr/bin/env>.first(*.IO.e) { "#!$_ raku" } else { "" };
```

niner:
Why don't we use "#!$*EXECUTABLE.absolute()"?

ugexe:
That would break relocatability

lembark:
> > Lacking an absoltue path to both env and rakudo there can be no #!
> > line in the wrapper.  
> 
> I think no shebang in that scenario is reasonable. Considering that,
> would the following not be sufficient?
> 
> ```
> my $shebang = do with </bin/env /usr/bin/env>.first(*.IO.e) { "#!$_
> raku" } else { "" }; ```


A wrapper with no executable following the shebang will fail.
What you describe above leaves us with:

    #!

    sub MAIN(:$name, :$auth, :$ver, *@, *%)
    {
        CompUnit::RepositoryRegistry.run-script( "zef", :$name, :$auth, :$ver );
    }

which is guaranteed to fail at runtime. 

Q: Why install a wrapper that is guaranteed to fail with a 
   red-herring error (malformed #!) rather than a meaningful
   error message about being unable to successfully install 
   the wrapper?

-- 
Steven Lembark
Workhorse Computing
lembark@wrkhors.com
+1 888 359 3508

ugexe:
My solution does not leave `#!`, it leaves no shebang at all. The wrapper only purpose is to allow multiple versions of distributions. Scripts will always work when invoked with e.g. `raku` so its not like its installing entirely broken stuff. And indeed it seems preferable to fail in a predictable way for edge cases than to create "magic" to guess and make getting at the root of the issue much more difficult (and still likely generally unsolvable due to relocatability, multiple rakudo installs, and other stuff).

niner:
> That would break relocatability

So we can use `$*EXECUTABLE.absolute()` for non-relocatable builds and only 
need a solution for the relocatable ones.

lembark:
On Sun, 24 May 2020 13:30:21 -0700
Nick Logan <notifications@github.com> wrote:

> It would seem to me a simpler solution is sufficient:
> `my $shebang = do with </bin/env /usr/bin/env>.first(*.IO.e) { "#!$_"
> } else { "" };`

Sanity check: Do you realize why:

    #!

    sub MAIN(:$name, :$auth, :$ver, *@, *%)
    {
        CompUnit::RepositoryRegistry.run-script( "#name#", :$name, :$auth, :$ver );
    }

is not an effective shell script?

-- 
Steven Lembark
Workhorse Computing
lembark@wrkhors.com
+1 888 359 3508

lembark:
On Mon, 25 May 2020 11:50:58 -0700
Nick Logan <notifications@github.com> wrote:

> > Lacking an absoltue path to both env and rakudo there can be no #!
> > line in the wrapper.  
> 
> I think no shebang in that scenario is reasonable. Considering that,
> would the following not be sufficient?
> 
> ```
> my $shebang = do with </bin/env /usr/bin/env>.first(*.IO.e) { "#!$_
> raku" } else { "" }; ```
> 

Please try saving the following to your disk and attempt to execute it.
What is the result:
<snip here>
#!

sub MAIN(:$name, :$auth, :$ver, *@, *%)
{
    CompUnit::RepositoryRegistry.run-script( "#name#", :$name, :$auth, :$ver );
}

lembark:
On Mon, 25 May 2020 12:26:20 -0700
niner <notifications@github.com> wrote:

> Why don't we use "#!$*EXECUTABLE.absolute()"?

Q: Is this guaranteed to be well-defined during the first call to
   generate a wrapper during rakduo installation?

If so then, yes, it works.

-- 
Steven Lembark
Workhorse Computing
lembark@wrkhors.com
+1 888 359 3508

ugexe:
The code you posted isn’t equivalent to the output of what my solution outputs

ugexe:
> So we can use `$*EXECUTABLE.absolute()` for non-relocatable builds and only need a solution for the relocatable ones.

If we are supporting the logic to do the magic lookup hoopla then why not always use it?

lembark:
On Mon, 25 May 2020 17:10:45 -0700
Nick Logan <notifications@github.com> wrote:

> My solution does not leave `#!`, it leaves no shebang at all. The
> wrapper only purpose is to allow multiple versions of distributions.
> Scripts will always work when invoked with e.g. `raku` so its not
> like its installing entirely broken stuff. And indeed it seems
> preferable to fail in a predictable way for edge cases than to create
> "magic" to guess and make getting at the root of the issue much more
> difficult (and still likely generally unsolvable due to
> relocatability, multiple rakudo installs, and other stuff).
> 

Existing zef looks like:

    #!/usr/bin/env perl6
    sub MAIN(:$name, :$auth, :$ver, *@, *%) {
    CompUnit::RepositoryRegistry.run-script("zef", :$name, :$auth, :$ver);
    }

this appears to require some variety of raku in the #! line to 
be executable. There is no way that "sub MAIN..." is going to 
be properly interpreted by the shell. This seems to require a "#!"
line in order to be executable on *NIX.

This means that the bare content:

    sub MAIN(:$name, :$auth, :$ver, *@, *%) {
    CompUnit::RepositoryRegistry.run-script("zef", :$name, :$auth, :$ver);
    }

will not be executable using a shell.

Leaving off the #! line and not setting the file as executable 
would work. That requires invoking, say, zef as:

    rakudo zef ... ;

but I don't think that's what people are expecting from a shell
wrapper.

If you don't want to trust $*EXECUTABLE, then whatever copy of 
rakudo is in whatever bin that lives above the installation 
directory is going to be the likeliest best for the #! argument
in a shell wrapper. 

As a last fallback we could use $*EXECUTABLE rather than die,
leaving an executable runable. 

Anyone who tries to install the rakudo distro in such a way that the 
wrappers are located in a directory completely divorced from the path 
to bin/rakudo  and unrelated to $*EXECUTABLE can write their own 
wrappers if need be -- using "/their/path/to/rakudo zef ... " and
override the default where necessary. The other 99% of cases will 
probably be sufficient ine of env, bin/rakdudo, or $*EXECUTABLE.

-- 
Steven Lembark
Workhorse Computing
lembark@wrkhors.com
+1 888 359 3508

ugexe:
A repository doesn’t even have to belong to a specific rakudo installation.

ugexe:
and fwiw I prefer dying to running scripts with the wrong version of rakudo sometimes and causing very confusing bugs

lembark:
> So we can use `$*EXECUTABLE.absolute()` for non-relocatable builds
> and only need a solution for the relocatable ones.

The fix for cases where someone wants to run, say, zef using some
rakudo in a completely separate tree is writing their own wrapper:

    #!/path/to/their/rakudo

    /path/to/standard/zef "$@";

That or providing an add'l installation argument for overriding the
standard location of rakudo that can live in the object with 
.prefix at which point the perl wrapper takes:

    with $.my_rakudo -> $path
    {
    }
    else
    {
    }

which solves it.

Q: Is there any reasonable way to install rakudo in such a way as
   to completely divorce the wrapper paths from any path to the
   rakudo used to complile the input?

e.g., if paths are /usr/bin/rakudo & /var/lib/rakudo then 
$*EXECUTABLE seems like the only reaonable choice for the #!.

Otherwise it's all living under a single installation prefix and 
searching up the tree for bin/rakudo will provide a working path.

Note that all of this >>>only<<< applies to systems which entirely
lack an "env" executable, which is pretty much nothing released 
since around System7 or embedded systrems -- and anyone using 
rakudo on an automobile engine prototype will have a well-described
way to override the wrappers. 

-- 
Steven Lembark
Workhorse Computing
lembark@wrkhors.com
+1 888 359 3508

niner:
On Mittwoch, 27. Mai 2020 17:03:11 CEST Nick Logan wrote:
> > So we can use `$*EXECUTABLE.absolute()` for non-relocatable builds and
> > only need a solution for the relocatable ones.
> If we are supporting the logic to do the magic lookup hoopla then why not
> always use it?

/bin/env is a useless indirection if you already know the full path to the 
executable you want to run. It makes it harder for packaging tools to auto-
detect dependencies (rpmbuild e.g. looks at the shebang line). For that reason 
/bin/env is discouraged in packaging guidelines and will cause warnings or 
even errors.
We have this lovely line in the .spec files for all module packages:
sed -i -e '1s:!/usr/bin/env :!/usr/bin/:' %{buildroot}/%{_datadir}/perl6/
vendor/bin/*

ugexe:
How do those packaging guidelines handle multiple versions of rakudo? I suspect there can only be one version, and if so then of course you have to package things that way for those platforms. We don’t place those arbitrary limitations on multiple versions of of the same module despite them existing for most packagers for instance.

ugexe:
I also say all of this as someone in a particularly unique position of having to had to deal with these version issues for modules and executables (as the author of zef and constantly getting issues caused by an old rakudo star install)

lembark:
On Wed, 27 May 2020 08:03:11 -0700
Nick Logan <notifications@github.com> wrote:

> > So we can use `$*EXECUTABLE.absolute()` for non-relocatable builds
> > and only need a solution for the relocatable ones.  
> 
> If we are supporting the logic to do the magic lookup hoopla then why
> not always use it?

Not sure what you mean by "hoopla"... the use of "#!/bin/env" as an 
absolute path to allow using the current path to look up an 
executable is well-established, has been for a long, long time.

So far as I know all current *NIX systems support the "env" 
executable. There may be systems where someone chooses not to 
install it. At that point neither of /bin/env nor /usr/bin/env
will exist. 

As an alternative, I'd propose using whatever version of rakudo 
is above the wrapper in the filesystem as a fallback. This allows
code from one installation of rakudo to install wrappers into another
one at initial install time -- or as a fix if something is broken. 

If someone is installing the wrapper into, say, /var/repo/rakudo
for their own reasons then $*EXECUTABLE is most likely what they 
need for any wrappers.

This guarantees that the path used with a #! is absolute, extant,
and available at the time a wrapper is installed. There isn't 
any mroe bullet-proof I can make it than that.

I'm not sure what you mean by "relocatable", perhaps that is the 
disconnect. The common approach of "#!/bin/env rakudo" allows
running a given wrapper with whatever version of rakudo is on
the current path -- which does allow the user to select whichever
rakudo they prefer at runtime via their path. 

This approach allows installing the wrappers pretty much anywhere
on the filesystem so long as rakudo is on the user's path at runtime.

The alternative of an absolute path to rakudo on the #! line still
allows the wrapper to be located anywhere on the filessytem -- at
installation time or after it. The wrapper can be moved at will 
becuase the path to rakudo is fixed at instal time. Catch there is
that if you want to run zef using another rakudo you have to use
"rakudo zef ..." since the #! line is absolute and the path won't
affect the execution.

You can still run zef with any copy of rakudo you like, it just
takes a bit more verbiage (one word). 

Q: What else do you mean by "relocatable"?

-- 
Steven Lembark
Workhorse Computing
lembark@wrkhors.com
+1 888 359 3508

lembark:
On Wed, 27 May 2020 07:59:29 -0700
Nick Logan <notifications@github.com> wrote:

> The code you posted isn’t equivalent to the output of what my
> solution outputs 

It doesn't use a hardwired /usr/bin/perl6 where one does not exist.
Other than that you get a #! wrapper with an absolute path to the
executable and whatever extant rakudo or raku or perl6 is available
from the installation environment. The '', '-j', '-m', etc, wrappers
all reference rakudo<-whatever> -- even if that executable does not
exist as the original code does.


-- 
Steven Lembark
Workhorse Computing
lembark@wrkhors.com
+1 888 359 3508

lembark:
On Wed, 27 May 2020 08:06:10 -0700
Nick Logan <notifications@github.com> wrote:

> A repository doesn’t even have to belong to a specific rakudo
> installation. 

e.g., /var/repo/rakudo with /usr/bin/rakudo?

Fine: In that case if /bin/env or /usr/bin/env exists the wrapper 
will be, say, #!/bin/env rakudo" which works perfectly or if 
env does not exist they'll get "#!$*EXECUTABLE" which leaves you
with an executable wrapper.

Either way the wrapper references an extant rakudo.

-- 
Steven Lembark
Workhorse Computing
lembark@wrkhors.com
+1 888 359 3508

lembark:
On Wed, 27 May 2020 08:07:35 -0700
Nick Logan <notifications@github.com> wrote:

> and fwiw I prefer dying to running scripts with the wrong version of
> rakudo sometimes and causing very confusing bugs 

I'd rather not die but you objected to the use of $*EXECUTABLE
previously. If I cannot use that as a final fallback then lacking
"env" and also lacking any rakudo above the wrapper in the filesystem
there is no way to write a shell wrapper. 

The ideal solution would probably be:

    env exists          => #!/path/to/env raukudo
    bin/rakdudo exists  => #!/path/to/bin/rakudo
                           #!$*EXECUTABLE

Absent the last alternative your choice is to either silently not
write the wrapper or write a wrapper that will fail when the shell
is incapable of executing Raku code.

-- 
Steven Lembark
Workhorse Computing
lembark@wrkhors.com
+1 888 359 3508

lembark:
> /bin/env is a useless indirection if you already know the full path
> to the executable you want to run. It makes it harder for packaging
> tools to auto- detect dependencies (rpmbuild e.g. looks at the
> shebang line). For that reason /bin/env is discouraged in packaging
> guidelines and will cause warnings or even errors.
> We have this lovely line in the .spec files for all module packages:
> sed -i -e '1s:!/usr/bin/env :!/usr/bin/:'
> %{buildroot}/%{_datadir}/perl6/ vendor/bin/*

My original problme was that the module installed wrappers with
"#!/usr/bin/env perl6" that were broken here due to the lack of
both /usr/bin/env and perl6 not being on my path.

The initial fix was using whchever exists of /bin/env or /usr/bin/env.

After that the issue was what to do if env didn't exist at all.

Given that rakudo is installed in /opt/rakduo here, accessed via
symlinks in /opt/bin that approach simply guarantees that the 
wrappers generate non-executable code. 

It also generates non-executable code on any system that uses the
stock GNU paths of /bin/exe.

If you prefer to avoid env entirely -- which has its bad points
if people expect their path to affect which executables are used
for commands -- then use $*EXECUTABLE by itself or find bin/rakudo
above the wrapper location with an alternative of $*EXECUTABLE.

I just have a preference for wrapper scripts that don't give me 
red-herring errors due to non-existant paths in their #! line.

-- 
Steven Lembark
Workhorse Computing
lembark@wrkhors.com
+1 888 359 3508

lembark:
On Wed, 27 May 2020 12:15:30 -0700
Nick Logan <notifications@github.com> wrote:

> How do those packaging guidelines handle multiple versions of rakudo?
> I suspect there can only be one version, and if so then of course you
> have to package things that way for those platforms. We don’t place
> those arbitrary limitations on multiple versions of of the same
> module despite them existing for most packagers for instance.

Using env handles multiple rakudo versions by simply using the version 
of rakudo that the user has selected via their path. At that point I
can have fourteen executables available and pick one via the path for
all of the wrappers.

That's why people picked up on using env when the #! lines got 
converted from relative to absolute in the POSIX standard. 

Reall that wrappers used to look like "#!perl", using the path, 
which is where "#!/bin/env perl" came from: same effect with an
absolute path to the executable following #! as per POSIX.


-- 
Steven Lembark
Workhorse Computing
lembark@wrkhors.com
+1 888 359 3508

lembark:
On Wed, 27 May 2020 12:17:05 -0700
Nick Logan <notifications@github.com> wrote:

> I also say all of this as someone in a particularly unique position
> of having to had to deal with these version issues for modules and
> executables (as the author of zef and constantly getting issues
> caused by an old rakudo star install)

I've had to deal with it on systems with multiple perl installs
floating around. It becomes a mess. Making code gracefully
deal with all of the options is a pain. The catch is that in the
end you can't ever deal with all of it. There are just too many
ways to deal with installation paths and breaking up filesystems
to handle them all gracefully.

The *NIX and perly way is to let the user decide: Select the 
exectable you want via the PATH and you will get it when you
run zef. If you export PATH="/var/tmp/rakudo-test/bin:$PATH" then
your next pass at "zef install foobar" will use *that* rakudo via
env. Period. You picked it becuase you wanted it, you got it. If 
your path doesn't include anything "rakudo" then you get a 
specific error about rakudo not being on your path. Bingo.

Hardwiring "/usr/bin/perl6" into the #! line won't help anyone
who has multiple versions run zef with the versin of their choice,
especially if the file is in their home direcory, a sandbox, or
/opt. It especially won't help someone who really does have 
rakudo on their path and keeps wondering why the wrapper fails
with /usr/bin/perl6 when /opt/bin/rakudo and /opt/bin/raku are 
right there. 

So far as I know there are no releases of rakudo that didn't have
a "bin/rakudo". If that's not true then adding a check for 
"bin/perl6" if bin exists and rakudo doesn't will solve that
with pretty much no overhead.

So far as I know the push now is to replace evreything perl6 with
raku or rakudo so the approach here seems workable.

-- 
Steven Lembark
Workhorse Computing
lembark@wrkhors.com
+1 888 359 3508

ugexe:
> Q: What else do you mean by "relocatable"?

rakudo can be built with a flag that allows the directory it is built in to be moved to a different location later

lembark:
On Fri, 29 May 2020 14:11:47 -0700
Nick Logan <notifications@github.com> wrote:

> > Q: What else do you mean by "relocatable"?  
> 
> rakudo can be built with a flag that allows the directory it is built
> in to be moved to a different location later

At which point:

    #!/bin/env rakudo

will simply find whatever is on their path allowing the executable
to be relocated wherever they like.

Bingo: 100% relocatable rakudo executable, if you can find it you
can run it. 

In this case there are really two options for composing the 
#! line: 

 -  Find env or die becuase there is no way whatever to use any
    absolute path to rakudo since the user has specified that 
    its path is dynamic at runtime. This guarantees that the 
    user can relocate rakudo at will with impunity.

 -  Assume that whatever rakudo is installing the werapper is the
    one they expect to use for real, at which point using
    $*EXECUTABLE makes complete sense: After relocating the 
    executable you simply re-run zef to install the wrappers
    using the actual path to the relocated executable. Move the
    executable again, re-use zef to install the wrappers. 

The ability to relocate rakudo -- or determine it runtime via the
path -- is why /bin/env is used in most cases (and why #!perl was
used originally).

In any case, hardiwiring "#!/usr/bn/env /usr/bin/perl6" is the most
anti-relocatable solution you can find as there is no sanity check
for the existance of either exeuctable when the wrapper is composed.


-- 
Steven Lembark
Workhorse Computing
lembark@wrkhors.com
+1 888 359 3508

lizmat:
This PR appear to have stalled.  Perhaps it should be closed?

lembark:
Probably. The notion of using the path to find which copy of raku to execute for a #! script seems foreign. At that point simply hardwiring it all to a non-existant /bin/raku fails equally well everywhere.

patrickbkr:
I do like to see this topic move forward, so please don't just close this. But I think we need to think about this more.

The following is my (potentially incorrect) understanding of how things work. Please correct me where I'm wrong.

A CURI (CompUnit::Repository::Installation) is a store represented as a folder for installed modules. It contains everything a module needs, including the actual `something.raku` script. All contents (including that script) are stored in a way opaque to the user. Running such an installed script requires calling the following Raku method:

    CompUnit::RepositoryRegistry.run-script('some_script');

Wrappers which basically only execute the above call are installed in the `curi_store/bin` subfolder. For those wrapper scripts to work their path must be in PATH or they need to be called with their absolute path.

There are three standard CURIs every Rakudo installation brings with it: `core`, `site` and `vendor` installed into `$prefix/share/perl6/`. In principle the `site` and `vendor` CURIs could be shared between multiple Rakudo installations, but that doesn't make much sense, as every installation brings its own set with it.
There can be additional CURIs in arbitrary places. Which CURIs are used is configurable. (How?)

Relocatable Rakudo installations have a well defined directory layout. Things (`rakudo` executables, CURIs, libs, ...) are always in the same place relative to each other.

There are different situations we need to consider:

- Standard CURIs, part of a non-relocatable installation: Typical case when Rakudo is installed by the distro. The CURIs belong to a specific Rakudo installation. -> Directly use the full path of the `rakudo` excutable in the shebang line.
- Standard CURIs, part of a relocatable installation: Different approaches are possible:
  - Use `env` in the shebang line and hope that the PATH points to the Rakudo this CURI belongs to. (That's *not* guaranteed.)
  - Use absolute paths and provide a tool to update the shebang lines. (That's what Strawberry Perl does.) One then needs to call that script whenever the installation is moved around.
  - Let the wrapper be a shell script or native executable and dynamically determine the `rakudo` path based on the location of the wrapper. Something like `$wrapper_path/../../../bin/rakudo`.
- Custom CURIs: Such CURIs don't belong to a specific rakudo installation, so it's difficult to classify them as relocatable or non-relocatable. The only options I see are:
  - Rely on `env` and just use the Rakudo that happens to be in the PATH.
  - Provide a tool to set / change the shebang lines. Then it's up to the user to configure the wrappers to his liking.

Sidenote: On Windows there are no shebang lines. There additional `.bat` files are generated that currently rely on PATH. In general the classification described above applies, but needs to be adapted accordingly.

I tend towards the following:

- We provide a tool written in Raku that is able to set shebang lines and generate wrapper scripts. It is capable of:
  - Detect whether a CURI is a standard one or not, and if it is locate the matching `rakudo` executable.
  - Generating Raku wrapper scripts with shebang lines that use `env` or absolute paths.
  - Generating wrappers that are either native executables or standard `sh` / `bat` scripts that locate `rakudo` based on the path of the wrapper.
- When building a non-relocatable Rakudo we default to using absolute paths in shebang lines on *nix and generate additional `.bat` files on Windows that use absolute paths as well.
- When building a relocatable Rakudo, we default to generating smart wrapper scripts that determine the path based on the wrapper scripts location.
- CURIs somehow remember (some configuration file) how its wrapper scripts should look. When using the tool mentioned above to reconfigure the wrappers in a CURI it will also change that configuration. The CURI implementation in Rakudo will respect this configuration and generate wrappers accordingly. The goal here is to never mix different wrappers in a specific CURI.

Sidenote: I am currently working on a tool that can generate native executables without requiring a compiler to locate programs (using absolute paths or paths relative to the executables location) and call them with configured and passed arguments. Maybe this tool can be of help here.

@niner, @ugexe Can you please check, whether there are any errors in my understanding and give your opinion?

ugexe:
> Which CURIs are used is configurable. (How?)

`raku -I inst#/path/to/curi ...` or `RAKULIB=inst#/path/to/curi`

There is not yet a way to change the default repos or their order.

Something possibly worth mentioning is we cant't rely on a given CURI prefix to work on any raku. see: the CURI method [method upgrade-repository()](https://github.com/rakudo/rakudo/blob/420cb1f4d7b5e45c9223946fc487390067b306ec/src/core.c/CompUnit/Repository/Installation.pm6#L113) which happens automatically when upgrading rakudo


> Generating wrappers that are either native executables or standard sh / bat scripts that locate rakudo based on the path of the wrapper.

Why wouldn't it use the rakudo that is running the proposed raku tool?

patrickbkr:
> Something possibly worth mentioning is we cant't rely on a given CURI prefix to work on any raku. see: the CURI method method upgrade-repository() which happens automatically when upgrading rakudo

Looking at the code it seems the repos are forward compatible (they can be upgraded), but not backwards compatible. (Older rakudos will fail when they encounter a newer CURI format. I didn't try, but I think they are incompatible.) There is no check for too new repo versions in there. Should there be one?

(A little off topic: Is there a good reason for the version marker to be written last in `upgrade-repository()` when the repo doesn't exist yet? When installation fails during setting up the CURI, any later re-runs of `upgrade-repository()` will then fail because the CURI is falsely assumed to be version 0.)

> > Generating wrappers that are either native executables or standard sh / bat scripts that locate rakudo based on the path of the wrapper.
> 
> Why wouldn't it use the rakudo that is running the proposed raku tool?

For the "Standard CURIs, part of a relocatable installation" situation the paths need to be dynamic. (I'm not sure I understood the question correctly, as the answer seems too obvious. If I'm right, can you elaborate a bit on the question?)

jubilatious1:
Just checking that this PR doesn't make adding a shebang line at the top of a script an absolute requirement. Thx.

ugexe:
> For the "Standard CURIs, part of a relocatable installation" situation the paths need to be dynamic. (I'm not sure I understood the question correctly, as the answer seems too obvious. If I'm right, can you elaborate a bit on the question?)

Basically I would expect `raku /path/to/site/bin/foo` (and `rakudo /path/to/bin/foo` etc) to run with the raku I've explicitly used and not try to guess. I was presuming these things made the dynamic options pretty limited.

patrickbkr:
(We are exclusively talking about the relocatable installation situation here.)

> Basically I would expect `raku /path/to/site/bin/foo` (and `rakudo /path/to/bin/foo` etc) to run with the raku I've explicitly used and not try to guess. I was presuming these things made the dynamic options pretty limited.

That's a usecase I didn't know exists up to now. Do I presume correctly, that the ability to call the scripts with a `raku` executable directly is a hard requirement? If yes, then the first solution that comes to mind is to always generate two files in the `bin/` folder.

- A `my-script.raku` file without any shebang line, containing just the naked `CompUnit::RepositoryRegistry.run-script('my-script');` call. That script does not have the `x` bit set on Unixes.
- A `my-script` / `my-script.bat` / `my-script.exe` file, that contains the aforementioned wrapper logic to determine the `raku` to use and then call the `my-script.raku` file with that `raku`. This file does have its executable bit set.

Then one could run the scripts like this: `raku /path/to/site/bin/my-script.raku`, and the wrapper scripts could still do their job.  (The wrapper scripts don't contain Raku code in this case. They are thus not runnable with `raku /path/to/site/bin/my-script`.)

I don't really like the idea of always having two files for each script in the `bin/` folder. But then we already do that on Windows. Also it introduces an inconsistency, as depending on how it's configured, the `my-script` is either a real Raku script and can be called as such or it's only the wrapper. In the second case the Raku script is located in `my-script.raku`.

The only alternative I see is to work with shebang lines with absolute paths and requiring the user to always trigger a regeneration of the scripts whenever the installation is moved. I'm unsure I like that solution better.

@ugexe Which solution would you prefer? Do you have any other ideas?

lizmat:
Is this PR still salvageable?

patrickbkr:
@lizmat I don't think there is much to be salvaged in the commit, but the discussion contains a lot of valuable ideas.

lizmat:
Issue #3635 will remain open and refer to this, so closing this now

PR #3724, opened 2020-05-27
Change INSTALL to markdown

coke:
Readability should be the same as .txt, but we get formatting for those viewing in GitHub.

Some other minor updates.


INSTALL.md
```diff
@@ -0,0 +1,171 @@
...
+
+# Running the test suite
+
+Entering "make test" will run a small test suite that comes bundled with
+Rakudo. This is a simple suite of tests, designed to make sure that the
+Rakudo compiler is basically working and that it's capable of running a
+simple test harness.
+
+Running "make spectest" will import the official Raku test suite from the
+"roast" repository <http://github.com/Raku/roast/> and run all of these
```
JJ, line 133: As long as you're transforming this, could you maybe also markdownify the links?

vrurg: Please, don't. GitHub [supports](https://gist.github.com/VEnis/7465176#automatic-links) the angle brackets for links. For plain text reading they're way more preferable over the original markdown format.

PR #3741, opened 2020-06-03
Fix possible race in multi-threaded precompilation

MasterDuke17:
by changing $!lock-count to an atomicint and using atomic operations to
interact with it. This code used to be inside a `.protect` call (and so
didn't need to be atomic), but that was removed in
ac87ea2ac84883bf701808485ac2da887f7496ed for other reasons.

### DIFF
--- a/src/core.c/CompUnit/PrecompilationStore/File.pm6
+++ b/src/core.c/CompUnit/PrecompilationStore/File.pm6
@@ -118,7 +118,12 @@ class CompUnit::PrecompilationStore::File
     has IO::Path:D $.prefix is built(:bind) is required;
 
     has IO::Handle $!lock;
+#?if moar
+    has atomicint $!lock-count;
+#?endif
+#?if !moar
     has int $!lock-count;
+#?endif
     has $!loaded;
     has $!dir-cache;
     has $!compiler-cache;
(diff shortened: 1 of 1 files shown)


ugexe:
Everywhere else atomicint is used there is a macro to not use it for the e.g. jvm. If we add macros to not do it on JVM then the race condition would then still exist on that platform.

MasterDuke17:
> Everywhere else atomicint is used there is a preprocessor to not use it for the e.g. jvm. If we add preprocessor to not do it on JVM then the race condition would then still exist on that platform.

@ugexe nice catch. Updated to use pre-processor directives. And yes, while the JVM and JS backends would still be subject to the race condition here, that's no different from all the other places atomics are used. @niner pointed out it looks like the JVM does support them https://docs.oracle.com/javase/7/docs/api/java/util/concurrent/atomic/AtomicInteger.html, we just need to implement that.

ugexe:
> that's no different from all the other places atomics are used.

That may not be entirely true. I can’t audit the code right now, but one might swap out an atomic for a regular int when the calculation doesn’t actually need to be 100% correct — for instance calculating heuristics to determine when to add a new thread to the thread pool — but it’s preferable if it is.

lizmat:
I'm not seeing `$!wont-lock` being set anywhere, nor a way to cleanly set it?

niner:
I fear this PR has bitrotted quite a bit. There is no `$!wont-lock` anymore in master. See commit 0500ed3ac093ceaf8926c282419c03e9ab3cd6e1

MasterDuke17:
> I fear this PR has bitrotted quite a bit. There is no `$!wont-lock` anymore in master. See commit [0500ed3](https://github.com/rakudo/rakudo/commit/0500ed3ac093ceaf8926c282419c03e9ab3cd6e1)

Rebased, conflicts fixed, force pushed. Let's see what CI says.

lizmat (approved):

PR #3744, opened 2020-06-04
Add read-from-input option, closes #3743

### LINKED ISSUE #3743: Behavior of (raku -) command
## The Problem

To get a Raku interactive shell, I have to type `raku` in command line but accidentally I typed `raku -` and the command line stays as it is without response even if I give further commands.

## Expected Behavior
I think it should exit or give some informative message.

## Actual Behavior

Stays as it is, seems like expecting more input, but does not response. 
To exit it, I have to type `Ctrl + C`.

## Steps to Reproduce
Once the `raku` binary is in PATH, type `raku -` in the powershell.

## Environment
* Compiler version (`raku -v`):

```
This is Rakudo version 2020.05.1 built on MoarVM version 2020.05
implementing Raku 6.d.
```

JJ:
This option was, apparently, not documented, but works as expected.

### DIFF
--- a/src/Perl6/Compiler.nqp
+++ b/src/Perl6/Compiler.nqp
@@ -143,6 +143,7 @@ With no arguments, enters a REPL (see --repl-mode option).
 With a "[programfile]" or the "-e" option, compiles the given program
 and, by default, also executes the compiled code.
 
+  -                    read program source from standard input until EOF is found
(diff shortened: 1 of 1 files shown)


lizmat:
Thanks!

However, I'm thinking of taking the Python approach here, which is to start up the REPL if it is reading from STDIN **and** STDIN is connected to a terminal.  But until then, this description is fine.

PR #3747, opened 2020-06-06
nativecall: Allow Callable arg to return List or IO::Path

softmoth:
This is needed to facilitate something like:

```perl6
sub find-foo() {
    %*ENV<FOO_LIBRARY> // ('foo', v1)
}

sub foo() is native(&find-foo) { * }
```

Without this change, `find-foo` must:

- use `guess_library_name()` directly (ugly, it should be an implementation detail),
- reproduce its functionality (error prone and onerous),
- give up and return a simple library name string (not portable or maintainable), or
- switch from `is native(&find-foo)` to `is native(find-foo)` (buggy, ENV value gets baked in during precompilation)

### DIFF
--- a/lib/NativeCall.rakumod
+++ b/lib/NativeCall.rakumod
@@ -209,19 +209,20 @@ my role NativeCallSymbol[Str $name] {
     method native_symbol()  { $name }
 }
 
-multi guess_library_name(IO::Path $lib) is export(:TEST) {
+multi guess_library_name(IO::Path $lib) returns Str is export(:TEST) {
     guess_library_name($lib.absolute)
 }
-multi guess_library_name(Distribution::Resource $lib) is export(:TEST) {
+multi guess_library_name(Distribution::Resource $lib) returns Str is export(:TEST) {
     $lib.platform-library-name.Str;
 }
-multi guess_library_name(Callable $lib) is export(:TEST) {
-    $lib();
+multi guess_library_name(Callable $lib) returns Str is export(:TEST) {
+    my \ret := $lib();
+    nqp::istype(ret, Str) ?? ret !! guess_library_name(ret)
 }
-multi guess_library_name(List $lib) is export(:TEST) {
+multi guess_library_name(List $lib) returns Str is export(:TEST) {
     guess_library_name($lib[0], $lib[1])
 }
-multi guess_library_name(Str $libname, $apiversion='') is export(:TEST) {
+multi guess_library_name(Str $libname, $apiversion='') returns Str is export(:TEST) {
     $libname.DEFINITE
         ?? $libname ~~ /[\.<.alpha>+ | \.so [\.<.digit>+]+ ] $/
             ?? $libname #Already a full name?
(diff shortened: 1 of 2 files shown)


niner:
The name of the library we load will already be baked into the precompilation file, except for Distribution::Resource objects which we handle in a smarter way. The whole native trait argument situation is a mess :/

softmoth:
Thanks, @niner. I don't doubt it's a mess; but calling `is native(&get-library)` does avoid baking in my testing:

### `./tmp/Foo.rakumod` test module

```raku
unit module Foo;
use NativeCall;

sub foo-lib()     { %*ENV<FOO_LIB> // 'libncursesw.so.6' }
sub foo-lib-ver() { %*ENV<FOO_LIB> // ('ncursesw', v6) }

# Not OK: Baked in at first compile!
sub vers_str1() returns Str is native(foo-lib) is symbol<curses_version> is export { * }

# This works!
sub vers_fun1() returns Str is native(&foo-lib) is symbol<curses_version> is export { * }

# This fails unless PR #3747 is applied
sub vers_fun2() returns Str is native(&foo-lib-ver) is symbol<curses_version> is export { * }

=finish
```

##### First run seems fine

    $ FOO_LIB=libncursesw.so.6 raku -Itmp -MFoo -e 'put try { $_() } // "FAIL {$_.WHY}" for &vers_str1, &vers_fun1'
    ncurses 6.1.20191109
    ncurses 6.1.20191109

##### Using `is native(Str)` has baked in the value, but `Callable` is correct

    $ FOO_LIB=nonsense raku -Itmp -MFoo -e 'put try { $_() } // "FAIL {$_.WHY}" for &vers_str1, &vers_fun1'
    ncurses 6.1.20191109
    FAIL fun1

##### Force a recompile, and all three fail as expected

    $ date >> tmp/Foo.rakumod
    $ FOO_LIB=nonsense raku -Itmp -MFoo -e 'put try { $_() } // "FAIL {$_.WHY}" for &vers_str1, &vers_fun1'
    FAIL str1
    FAIL fun1

##### But failed libraries are also baked in during compile

    $ FOO_LIB=libncursesw.so.6 raku -Itmp -MFoo -e 'put try { $_() } // "FAIL {$_.WHY}" for &vers_str1, &vers_fun1'
    FAIL str1
    ncurses 6.1.20191109

##### Now, compare `--> Str` (works) `--> List` (needs this patch)

    $ date >> tmp/Foo.rakumod
    $ raku -Itmp -MFoo -e 'put try { $_() } // "FAIL {$_.WHY}" for &vers_fun1, &vers_fun2'
    ncurses 6.1.20191109
    FAIL fun2
    $ ./install/bin/raku -Itmp -MFoo -e 'put try { $_() } // "FAIL {$_.WHY}" for &vers_fun1, &vers_fun2'
    ncurses 6.1.20191109
    ncurses 6.1.20191109
    $

lizmat:
Is this still a thing after new-disp has landed and @niner has done a lot of work on the NativeCall interface?

lizmat:
Looks like the return signature has already been removed in another commit, so closing.

Issue #3751, opened 2020-06-12
Find a version control conflict marker only after unspace


atroxaper:
## The Problem
For now, the compiler can detect VCS conflict section in the source file. The section looks like:
```
<<<<<<< HEAD
say 'your';
=======
say 'their';
>>>>>>> branch
``` 
The feature works only if the previous line ends with `\` symbol, which means quoted whitespace.

## Expected Behavior
The feature should work irrespective of `\` existence.

## Actual Behavior
The feature works only if the previous line ends with `\` symbol.

## Steps to Reproduce
Try to compile a file with
```
say 'first both';\
<<<<<<< HEAD
say 'your';
=======
say 'their';
>>>>>>> branch

say 'middle both';

<<<<<<< HEAD
say 'your';
=======
say 'their';
>>>>>>> branch
```
and receive only one SORRY for line 2 with the message `Found a version control conflict marker`.

## Environment
* Operating system: macOS
* Compiler version: This is Rakudo version 2020.05.1 built on MoarVM version 2020.05 implementing Raku 6.d

PR #3752, opened 2020-06-12
#3751 Fix vcs conflict marker detection

atroxaper:
Previously vcs conflict marker detection work only after unspace.
Now the marker can be detected as after unspace as without it.

Changes for issue: #3751
New tests: https://github.com/Raku/roast/pull/647


lizmat:
Thanks!

PR #3760, opened 2020-06-18
Type Parameter's methods

Kaiepi:
This affects the return values of some of `Parameter`'s methods (such as
`name`) that were returning values of inconsistent types.

Passes `make test` and `make spectest`. Should probably wait until after this next release.


lizmat:
Looks good to me, but would like to postpone merge until after this weekend's release

lizmat:
Thanks!

PR #3767, opened 2020-06-22
[WIP] Make Parameter.BUILD more robust

Kaiepi:
`Parameter.BUILD` can easily generate nonsensical parameters, and can
throw when attempting to generate more complex parameters. This should
make it so `Parameter.BUILD` always generates a sensible parameter for the
arguments given by establishing an order of preference for arguments
passed: `:flags`, followed by `:name`, followed by the remaining arguments.
If an argument does not make sense given an argument with higher
preference than it, then it is ignored. In addition, better support for
named parameters and type coercions has been added.

I'm a bit iffy on how errors are handled with this currently. Typed exceptions are not used, and unless a parameter name is outright impossible to parse properly, no errors are thrown. This means parameters that cannot be created in standard Raku, like `Int:D *@slurpy?`, become `*@slurpy`, for instance.

Depends on https://github.com/rakudo/rakudo/pull/3760.
Passes `make test` and `make spectest`, with some additional Roast tests.

### DIFF
--- a/src/core.c/Parameter.pm6
+++ b/src/core.c/Parameter.pm6
@@ -68,11 +68,11 @@ my class Parameter { # declared in BOOTSTRAP
         }
     }
 
-    sub definitize-type(Str:D $type, Bool:D $definite) {
+    sub definitize-type(Str:D $type, Bool:D $definite --> Mu) {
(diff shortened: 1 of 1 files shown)


lizmat:
Seems like this has stalled, so assume can be closed.  Please re-open if you disagree!

PR #3789, opened 2020-07-12
Revert broken master

ugexe:
This allows master to pass its tests again. Also, it reverts the interface regression where it was assumed all CURs are in the rakudo code base when the reality is there are externals CURs that were using the original interface (`MyCur.need(...)`).


lizmat:
Seems I forgot to push a change to roast.  I guess I won't do that anymore.

PR #3793, opened 2020-07-13
Also set the $NQP_HOME HLL symbol in the nqp compiler object

patrickbkr:
It's needed by NQP to determine the template.html path.

This is supporting [Raku/nqp#645](https://github.com/Raku/nqp/pull/645).


vrurg:
I'm confused. Since NQP needs it then why adding it to rakudo?

patrickbkr:
@vrurg That's the reason it's a draft PR. The code I added in NQP to set this same HLL symbol is not called when running rakudo (I added it to [src/NQP/Compiler.nqp](https://github.com/Raku/nqp/pull/645/files#diff-3404f0859da177921b33f02c70401d8a)). I'm currently searching for a place in NQP where this setup code fits well and will be called both, when running Rakudo and when running NQP. `src/HLL/Compiler.nqp` is a possibility, but adding it to its `BUILD` seems a bit strange. I'm open for suggestions.

vrurg:
I have an idea but don't have time right now to develop it into something final. The point is that the code using the symbols is located in `HLL::Compiler` and `HLL::Backend::MoarVM`. The second one we can generalize to just a backend because others may need this data too. Perhaps it makes sense to switch away from using a globalish `*hllsym` toward local data on the compiler or backend class? Besides, it would scratch another itch I spotted in your work in raku/nqp#645: code setting the `*_HOME` variables is duplicated for Rakudo and NQP. By having a method on the compiler class you'd have it all one place. The current compiler is easily obtainable with `nqp::getcomp`.

Hopefully this would help you find a cleaner way as setting a hllsym for nqp in Rakudo looks really somewhat confusing.

vrurg:
BTW, since you already patching `nqp-j.in`, maybe you could split it into *nix and win versions after all?

src/Perl6/Compiler.nqp
```diff
@@ -7,6 +7,31 @@ class Perl6::Compiler is HLL::Compiler {
...
+
+    method BUILD() {
+        # Determine Perl6 and NQP dirs.
+#?if jvm
+        my $sep := nqp::atkey(nqp::jvmgetproperties,'os.name') eq 'MSWin32' ?? '\\' !! '/';
+        my $execname := nqp::atkey(nqp::jvmgetproperties,'perl6.execname');
+#?endif
+#?if !jvm
+        my $config := nqp::backendconfig();
+        my $sep := $config<osname> eq 'MSWin32' ?? '\\' !! '/';
```
vrurg, line 20: If we gonna have pieces like this scattered all around it's no good. Feels to me like we should expose some of the most required platform-dependant variable from nqp-configure via hll-config to reduce the amounts of platform-detecting code.

This is not a request for change, just something to consider.

src/Perl6/Compiler.nqp
```diff
@@ -7,6 +7,31 @@ class Perl6::Compiler is HLL::Compiler {
     has $!language_modifier; # Active language modifier; PREVIEW mostly.
     has $!language_revisions; # Hash of language revision letters. See gen/<vm>/main-version.nqp
     has $!can_language_versions; # List of valid language version
+    has $!rakudo-home;
+
+    method BUILD() {
```
vrurg, line 12: BTW, I still not well known with nqp limitations, but this overrides `HLL::Config::BUILD` without invoking it, if memory serves me right.

PR #3805, opened 2020-07-19
Make X::Method::NotFound exception more user friendly

### LINKED ISSUE #3802: "No such method '!private' for invocant of type xxx Did you mean '!private'?"
## The Problem

In this program:

```
class Privateed  {
    method !private() { return "⌣"  }
    method call-private( $method-name )  {
	self."!$method-name"();
    }
}

my $obj = Privateed.new;
say $obj.call-private( "private" );
```

There's an error with the LTA message:

```
No such method '!private' for invocant of type 'Privateed'. Did you mean '!private'?
  in method call-private at private-test.p6 line 10
  in block <unit> at private-test.p6 line 15
```

## Expected Behavior

A different kind of message. Also maybe an error indicating you can't call private methods that way. How I arrived to that contrived syntax was by using the exclamation-mark less version, simply writing "private".

Also, not clear there's a way to call private messages by name. Other attempts (using `self!"private"`, for instance) fail.

## Environment
* Operating system:

raku 2020.06

vrurg:
The exception can now provide user with a few additional tips depending on the context of the method call and what kind of mistake the user possibly made. Tips include, in addition to the traditional suggestion list of like candidates, the following two:

- `self!"private"()` syntax if `self."!private"()` form is presumably used.

- If a private method invoked but all suggested candidates are public then tip the user about possibly using public method call. Same advise is given for public invocation if all suggestions are private.

No private methods are suggested if call is made outside of a class body. The condition is checked by matching invocator class to `$?CLASS`.

Measures are taken not to flood the user with surplus tips.

Should fix #3802

### DIFF
--- a/src/core.c/Mu.pm6
+++ b/src/core.c/Mu.pm6
@@ -1026,6 +1026,7 @@ my class Mu { # declared in BOOTSTRAP
 
     method dispatch:<!>(Mu \SELF: \name, Mu \type, |c) is raw {
         my $meth := type.^find_private_method(name);
+        my $class := nqp::getlexcaller('$?CLASS');
         $meth ??
             $meth(SELF, |c) !!
             X::Method::NotFound.new(
@@ -1033,6 +1034,7 @@ my class Mu { # declared in BOOTSTRAP
               method   => name,
               typename => type.^name,
               :private,
+              :in-class-call(nqp::eqaddr(nqp::what(SELF), $class)),
             ).throw;
     }
 
--- a/src/core.c/Mu.pm6
+++ b/src/core.c/Mu.pm6
@@ -1026,7 +1026,6 @@ my class Mu { # declared in BOOTSTRAP
 
     method dispatch:<!>(Mu \SELF: \name, Mu \type, |c) is raw {
         my $meth := type.^find_private_method(name);
-        my $class := nqp::getlexcaller('$?CLASS');
         $meth ??
             $meth(SELF, |c) !!
             X::Method::NotFound.new(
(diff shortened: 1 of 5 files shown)


ugexe:
If the error only makes such suggestions for code inside the class containing the private method then it makes sense, but e.g. `raku -e 'use Foo; Foo.new.baz'` should not suggest private methods like `!baz` or `!bar`.

vrurg:
> If the error only makes such suggestions for code inside the class containing the private method then it makes sense, but e.g. `raku -e 'use Foo; Foo.new.baz'` should not suggest private methods like `!baz` or `!bar`.

Argh, of course! But I don't currently see a concise way to distinguish one call from another. I would appreciate an advise if you have any.

vrurg:
I didn't realize it at first that we do actually suggest `!baz` if `Foo.new.baz` fails. So, my patch changes little from this respect.

Moreover, I don't currently see a smart way of determining the context from withing `method_not_found_error`.

ugexe:
Hmmm, then I wonder if the best solution is just to remove the suggestion on private methods then

vrurg:
I was considering it myself. Only suggest publics on a public call so far looks like the right thing to do.

vrurg:
I have implemented a seemingly reasonable solution to in/out of class method calls. Made tips given to the user as concise as possible. Considering that method call errors are quite a commonplace, should be much more helpful now.

vrurg:
Unless there're no objections to this PR, I'll add a test to rakudo. This is most certainly not a speccable case.

src/core.c/Mu.pm6
```diff
@@ -1026,13 +1026,15 @@ my class Mu { # declared in BOOTSTRAP
 
     method dispatch:<!>(Mu \SELF: \name, Mu \type, |c) is raw {
         my $meth := type.^find_private_method(name);
+        my $class := nqp::getlexcaller('$?CLASS');
```
ugexe, line 1029: Should this be inlined so there is no additional cost for the first ternary condition?

vrurg: It should. An overlooked remnant of a debug print which followed the statement. Thanx!

PR #3810, opened 2020-07-20
Deprecate nqp-lib

patrickbkr:
The option has been effect less for at least a year now. So remove all uses in Rakudo and add a deprecation notice.

This is for *after* the release.

### DIFF
--- a/src/Perl6/Compiler.nqp
+++ b/src/Perl6/Compiler.nqp
@@ -54,6 +54,10 @@ class Perl6::Compiler is HLL::Compiler {
             %options<doc> := 'Text';
         }
 
+        if nqp::existskey(%options, 'nqp-lib') {
+            note('Option `--nqp-lib` is deprecated, has no effect and will be removed in 2021.06.');
+        }
+
         my $argiter := nqp::iterator(@args);
         nqp::shift($argiter) if $argiter && !nqp::defined(%options<e>);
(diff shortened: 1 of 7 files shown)


patrickbkr:
This was an accidental push of latest master to this branch. That's why it got autoclosed. I'm reopening.

src/Perl6/Compiler.nqp
```diff
@@ -54,6 +54,10 @@ class Perl6::Compiler is HLL::Compiler {
             %options<doc> := 'Text';
         }
 
+        if nqp::existskey(%options, 'nqp-lib') {
+            note('Option `--nqp-lib` is deprecated, has no effect and will be removed in a future release.');
```
Altai-man, line 58: "Will be removed in 2021.06".

patrickbkr:
@Altai-man I think it turned into a noop during my big relocatability refactor a year ago. But I'm not sure anymore. It might actually have been a noop already.

Altai-man:
>I think it turned into a noop during my big relocatability refactor a year ago. But I'm not sure anymore. It might actually have been a noop already

No problems with squash-merge it once CI is green.

src/Perl6/Compiler.nqp
patrickbkr: Resolved.

PR #3812, opened 2020-07-21
Make Blob.gist work for all sizes of Blob

Kaiepi:
`Blob.gist` would throw with non-8-bit `Blob`s. This should make it so
`Blob.gist` will work for any size of `Blob`, including future 1/2/4-bit
ones.


lizmat:
I'm confused by the number of failures in CI, and whether they have anything to do with this PR.

ugexe:
Yes, master is again broken so PRs will show failing results

lizmat:
Was clean locally, and since it is all just Raku / nqp code, decided the CI test failures were no reason to not merge.

PR #3814, opened 2020-07-21
Fix release pipeline to not delete the linux build

patrickbkr:


Altai-man:
How can re re-run CI nicely?

patrickbkr:
I just tried the suggestion of ugexe to create an empty commit to re-trigger the CI. It works, but the empty commit will stay around.

JJ:
> I just tried the suggestion of ugexe to create an empty commit to re-trigger the CI. It works, but the empty commit will stay around.

Well, you can restart them in every one of the pages... Too much work, probably.

Altai-man:
>I just tried the suggestion of ugexe to create an empty commit to re-trigger the CI. It works, but the empty commit will stay around

No, we can squash and merge, github even has UI for it. :) I'll do it once CI is green.

patrickbkr:
@Altai-man I think the most obvious thing to do, when one wants to retrigger the CI because of new commits in master, is to just rebase and push. I just didn't think of that solution earlier. I'm pondering on these problems mostly because of [problem-solving#206](https://github.com/Raku/problem-solving/issues/206) at the moment.

ugexe:
> I just tried the suggestion of ugexe to create an empty commit to re-trigger the CI. It works, but the empty commit will stay around

You avoid the empty commit with `--amend` if you are ok rewriting history (similarly to rebasing)

patrickbkr:
> > I just tried the suggestion of ugexe to create an empty commit to re-trigger the CI. It works, but the empty commit will stay around
> 
> You can avoid the empty commit with `--amend` if you are ok rewriting history (similarly to rebasing)

Thanks for the hint!

Issue #3799, opened 2020-07-18
custom postcircumfix are neither allowed nor forbidden


gfldex:
```
class C { };

my \px = C.new;

sub postcircumfix:«< >»(C $c is raw) {
    dd $c;
}

px<ls -l>;
```
OUTPUT:
```
===SORRY!=== Error while compiling /home/dex/projects/raku/lib/raku-shell-piping/px.raku
Missing required term after infix
at /home/dex/projects/raku/lib/raku-shell-piping/px.raku:9
------> px<ls -l>⏏;
    expecting any of:
        prefix
        term
```

This is either LTA or Rakubug. In both cases it's ENODOC.

## Environment
evalable6

vrurg:
`< >` postcircumfix has a special [meaning](https://docs.raku.org/language/operators#postcircumfix_%3C_%3E) and can't be redefined. Hence the error: second `>` is considered to a `greater than` op.

What you really need is either

```raku
multi postcircumfix:<{ }>(C $c, $s) {
    dd $s;
}
multi postcircumfix:<{ }>(C $c, @s) {
    dd @s;
}

px{"ls"};
px<ls -l>;
```

vrurg:
Reopening as I realized that compiler should still be able to parse `< >` properly even when redefinition attempt took place.

vrurg:
I looked into feasibility of making overriding of `< >`, `<< >>`, and `« »` postcircumfixes possible. It seems to be. But aside of this I also realized that any changes needed to achieve this will be overridden by `RakuAST` implementation.

Therefore I think it is better to postpone this issues up until `RakuAST` is implemented.

PR #3825, opened 2020-07-25
Make sure overriding of angular associative postcircumfixes doesn't break compilation

vrurg:
Rakudo currently fails to compile `< >` postcircumfix if code defines a `&postcirumfix:«< >»` candidate. This behavior is caused by the fact that aside from other operators, `< >` is defined using `token postcircumfix:sym<ang>` instead of `postcurcumfix:sym«< >»`, what would be the standard way.

Renaming the token and corresponding action methods results in silent ignoring of user-defined routine. Same happens with two other associative ops: `« »` and `<< >>`. With this PR all three will result in a error message like this:

```
Cannot override postcircumfix operator '< >', as it is a special form handled directly by the compiler
(consider implementing '&postcircumfix:<{ }>', or method AT-KEY, or Associative role)
```

### DIFF
--- a/src/Perl6/Actions.nqp
+++ b/src/Perl6/Actions.nqp
@@ -6752,7 +6752,7 @@ class Perl6::Actions is HLL::Actions does STDActions {
         make $past;
     }
 
-    method circumfix:sym<ang>($/) { make $<nibble>.ast; }
+    method circumfix:sym«< >»($/) { make $<nibble>.ast; }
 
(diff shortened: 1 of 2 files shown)


vrurg:
Takes care of #3799. I don't consider this a fix. Perhaps the real one could be implemented as a resolution to Raku/problem-solving#214.

lizmat:
Feels to me this could be closed.  Please re-open if my feeling was wrong.

vrurg:
Non-vital, can wait for RakuAST.

PR #3826, opened 2020-07-25
Remove (()) from Map.new.raku

lizmat:
There is no need for them, so just don't create them.


lizmat:
$ raku -e 'Map.new(()) for ^10000'
    real	0m0.247s

    $ raku -e 'Map.new for ^10000'
    real	0m0.143s

    $ raku -e 'Nil for ^10000'
    real	0m0.134s

PR #3834, opened 2020-07-30
Partial fix of mixing in a role with attributes into a NQP class

vrurg:
When a Raku role with an attribute is mixed into a metamodel or NQP class rakudo throws `X::Method::NotFound`:

```raku
role Foo { has Mu $.bar }
"foo".HOW does Foo;
```

This PR is a fix. It also fixes `X::Method::NotFound` exception message code failing on `KnowHOW` objects because their `methods` method returns the method table, not a list of methods.

### DIFF
--- a/src/core.c/Attribute.pm6
+++ b/src/core.c/Attribute.pm6
@@ -26,7 +26,11 @@ my class Attribute { # declared in BOOTSTRAP
         if self.has_accessor {
             my str $name   = nqp::unbox_s(self.name);
             my $meth_name := nqp::substr($name, 2);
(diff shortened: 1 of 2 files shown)


vrurg:
The fix is partial as parameterized roles with an attribute still throw when role parameter is assigned to the attribute:

```raku
role Foo[$p] {
    has $.foo = $p;
}
"42".HOW does Foo["The Answer"];
```

With this PR the exception produced is `X::TypeCheck::Binding::Parameter`. Looks like a problem with a build plan produced for the attribute but can't tell more.

vrurg:
**Upd** Full fix includes Raku/nqp#650

Issue #3824, opened 2020-07-24
Cannot profile code anymore


lizmat:
## The Problem

No profile files are written anymore.

## Expected Behavior

Profile files should be written.

## Actual Behavior

    $ raku --profile -e 'say "foo"'
    foo
    Writing profiler output to profile-1595623269.3879838.html
    Some exceptions were thrown in END blocks:
      X::AdHoc: Cannot call method 'nqp-home' on a null object
          in any <main> at /Users/liz/Github/rakudo.moar/install/share/perl6/runtime/perl6.moarvm line 1

and *no* profile file is written.

This seems to have happened just before the 2020.07 release, or shortly after.

Altai-man:
Can confirm a 2020.07 regression, it's a call for a point.

AlexDaniel:
Bisected: (2020-07-16) https://github.com/rakudo/rakudo/commit/1b3780f72569f3a8a74ee4beb69c0a9ca2f2f333
Before and after: https://gist.github.com/8927fa4727f568968d9adfe6625eeabb

Though I think @patrickbkr already knows :)

patrickbkr:
The problem is that the `nqp-home()` method in the `HLL::Compiler` class can not be accessed uniformly. When calling some NQP code the method can be accessed with `nqp::getcomp('nqp').nqp-home()`. When running Raku code one needs to use `nqp::getcomp('Raku').nqp-home()`.

The profiler code in NQP in [src/vm/moar/HLL/Backend.nqp](https://github.com/Raku/nqp/blob/master/src/vm/moar/HLL/Backend.nqp#L702) tries to access the nqp home using the `nqp` compiler object. But the compiler is not registered under that name. So it seems the Compiler object is a poor choice to put that function.

I'm not sure where else we might put it. The code in that function needs to access

- the command line parameters (well, the corresponding `rakudo-home()` function in rakudo does)
- environment variables
- the compile configuration

Where could such a function live?

vrurg:
Can we consider introducing a `'default'` compiler? It's actually not the first time when it doesn't matter what particular compiler `nqp::getcomp` returns as all what's needed is `HLL::Compiler`-declared services.

patrickbkr:
@vrurg I've typed together something along the lines of your idea. It seems to work.

I'm unsure whether the `Compiler` object is the right place to put such generic functionality though. (I don't know of any better place!) Can you review?

Altai-man:
[...]

PR #3838, opened 2020-08-03
Add a Perl6::SysConfig class

patrickbkr:
This PR in combination with [Raku/nqp#653](https://github.com/Raku/nqp/pull/653) will hopefully fix [rakudo/rakudo#3824](https://github.com/rakudo/rakudo/issues/3824) for good.

I'm still pondering whether `rakudo-build-config` should just be left out of `SysConfig` or not.

I do value feedback on this, as I'm not entirely convinced this is the best way to actually solve #3824.


src/Perl6/Compiler.nqp
```diff
@@ -9,6 +9,10 @@ class Perl6::Compiler is HLL::Compiler {
     has $!can_language_versions; # List of valid language version
     has $!rakudo-home;
 
+    method config() {
```
vrurg, line 12: Same note as https://github.com/Raku/nqp/pull/653/files#r464615285

src/main.nqp
```diff
@@ -14,8 +17,8 @@ $comp.parsegrammar(Perl6::Grammar);
 $comp.parseactions(Perl6::Actions);
 $comp.addstage('syntaxcheck', :before<ast>);
 $comp.addstage('optimize', :after<ast>);
-hll-config($comp.config);
-nqp::bindhllsym('Raku', '$COMPILER_CONFIG', $comp.config);
+hll-config(nqp::gethllsym('default', 'SysConfig').rakudo-build-config);
+nqp::bindhllsym('Raku', '$COMPILER_CONFIG', nqp::gethllsym('default', 'SysConfig').rakudo-build-config);
```
vrurg, line 21: Similar note as for `config()` method name. Except that here a name like `$SYSTEM_CONFIG` fits even better than `$COMPILER_SYSCONFIG`.

vrurg:
Other than the minor doubts I, both PRs looks promising to me.

src/Perl6/Compiler.nqp
patrickbkr: Done.

src/main.nqp
patrickbkr: I removed `$COMPILER_SYSCONFIG` entirely. It was only used in a few places.

timo:
it randomly occurs to me: do we make sure to escape stuff properly when substituting into the .in files? If a path contains a ' in it, we could be in trouble ...

patrickbkr:
@timo: There are a set of macros in `nqp-configure/lib/NQP/Macros.pm` that do quoting for different situations, (Makefile depenency, sh, CMD, C source, maybe more) which should be used at the place of use in the *.in files.

I think there are some chars that can't be escaped in Makefiles (spaces?) for which the only solution is to put the thing into a Makefile variable, but that's not a concern of the templating system.

patrickbkr:
I'm going to merge this and the respective NQP PR now, so we'll have some days of testing left before the next release. If problems show up we'll have some time to react.

PR #3845, opened 2020-08-12
Start to use nqp::execname on the JVM backend

usev6:
The op is now available in Rakudo (after the last NQP bump).
(It's only NQP on the JVM backend that still needs a bootstrap.)


usev6:
Hmm, there seems to be something wrong (CI failures). I'll investigate.

usev6:
This failed, because nqp::execname returned null if execname was not found in the system properties. Should be fixed with https://github.com/Raku/nqp/pull/654. I'll open a new PR if things are ready.

usev6:
This failed, because nqp::execname returned null if execname was not found in the system properties. Should be fixed with https://github.com/Raku/nqp/pull/654. I'll open a new PR if things are ready.

PR #3848, opened 2020-08-15
Simplify binary release instructions a little

patrickbkr:
- They can now be copied more easily by using wildcards for some filenams.
- Also the code is now highlighted correctly.
- Some smaller changes to the commands.

In adapted form this pulls in the changes of PR#3287.

### DIFF
--- a/docs/release-guide-binary.md
+++ b/docs/release-guide-binary.md
@@ -63,22 +63,24 @@ As of 2019-07-08 CentOS 6 (using glibc 2.12) is a good pick.
 - Use a 64 bit Linux.
 - Install docker.
 - `docker run -it --name=rakudo-build centos:6 bash`
-
-    yum -y update && yum clean all
-    yum install git perl perl-core gcc make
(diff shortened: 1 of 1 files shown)


lizmat:
This **documentation only** PR would be unmergeable because the CI tests failed.

And I cannot invoke a re-test either, I apparently need some Circle-CI credentials that I do not appear to have.

Feels to me we should make sure that *all* core devs should be able to call for a re-test without additional hassle.  Because if not, PR's will just start falling by the wayside.

Altai-man:
Rakudo tests:
```
t/04-nativecall/08-callbacks.t                                (Wstat: 256 Tests: 0 Failed: 0)
  Non-zero exit status: 1
  Parse errors: Bad plan.  You planned 8 tests but ran 0.
Files=108, Tests=1812, 209 wallclock secs ( 0.76 usr  0.12 sys + 393.99 cusr 21.14 csys = 416.01 CPU)
Result: FAIL
make: *** [m-coretest5] Error 1
```

patrickbkr:
@lizmat I feel your sentiment very much. I'm working towards making our CI more robust and easy to use and hope the solution I strive to implement will indeed be an improvement over what we have now.

PR #3849, opened 2020-08-15
Make error when sorting with junctions more awesome

lizmat:
You cannot sort a list that contains Junction values.  The reason for
this, is really that we do not have a sensible way of collapsing a
Junction into an Order value.

Instead of dying with an internals error about not being able to
coerce a Junction to an int, it now dies with a X::Cannot::Junction
exception.

Spotted by AlexDaniel++

### DIFF
--- a/src/core.c/Rakudo/Sorting.pm6
+++ b/src/core.c/Rakudo/Sorting.pm6
@@ -10,6 +10,25 @@ my class Rakudo::Sorting {
         )
     }
 
+    # Define a local default compare function that is basically the core's
+    # infix:<cmp> for all values *but* Junctions.  If a Junction is involved
+    # in the comparison, it will throw an exception, as there is no logical
+    # way to handle that case.
+    proto sub compare(|) {*}
+    multi sub compare(Junction:D \a, \b) {
+        X::Cannot::Junction.new(
+          junction  => a.raku,
+          for       => "as a value when sorting",
+        ).throw
+    }
+    multi sub compare(\a, Junction:D \b) {
+        X::Cannot::Junction.new(
+          junction  => b.raku,
+          for       => "as a value when sorting",
+        ).throw
+    }
+    multi sub compare(\a, \b) { a cmp b }
+
     # https://en.wikipedia.org/wiki/Merge_sort#Bottom-up_implementation
     # The parameter is the HLL List to be sorted *in place* using simple cmp.
     method MERGESORT-REIFIED-LIST(\list) {
@@ -55,7 +74,7 @@ my class Rakudo::Sorting {
                           nqp::isge_i($j,$end)
                             || nqp::iseq_i(
                                  nqp::decont(  # for some reason we need this
-                                   nqp::atpos($A,$i) cmp nqp::atpos($A,$j)
+                                   compare(nqp::atpos($A,$i), nqp::atpos($A,$j))
                                      || nqp::cmp_i($i,$j)
                                  ), # apparently code gen with || isn't right
                                  -1
@@ -88,7 +107,7 @@ my class Rakudo::Sorting {
           ),
           nqp::if(
             nqp::islt_i($n,2)
-              || nqp::isle_i(nqp::atpos($A,0) cmp nqp::atpos($A,1),0),
+              || nqp::isle_i(compare(nqp::atpos($A,0), nqp::atpos($A,1)),0),
             list,  # nothing to be done, we already have the result
             nqp::p6bindattrinvres(list,List,'$!reified',  # need to swap
               IB2(nqp::atpos($A,1),nqp::atpos($A,0)))
@@ -234,9 +253,10 @@ my class Rakudo::Sorting {
                           nqp::isge_i($j,$end)
                             || (nqp::iseq_i(
                                  nqp::decont(
-                                   nqp::atpos($S,nqp::atpos_i($A,$i))
-                                     cmp nqp::atpos($S,nqp::atpos_i($A,$j))
-                                     || nqp::cmp_i($i,$j)
+                                   compare(
+                                     nqp::atpos($S,nqp::atpos_i($A,$i)),
+                                     nqp::atpos($S,nqp::atpos_i($A,$j))
+                                   ) || nqp::cmp_i($i,$j)
                                  ),
                                  -1
                                ))
@@ -275,8 +295,10 @@ my class Rakudo::Sorting {
           nqp::p6bindattrinvres(nqp::create(List),List,'$!reified',
             nqp::if(
               nqp::islt_i($n,2)
-                || nqp::isle_i(
-                    mapper(nqp::atpos($O,0)) cmp mapper(nqp::atpos($O,1)),0),
+                || nqp::isle_i(compare(
+                     mapper(nqp::atpos($O,0)),
+                     mapper(nqp::atpos($O,1))
+                   ),0),
               $O,  # nothing to be done, we already have the result
               IB2(nqp::atpos($O,1),nqp::atpos($O,0))  # need to swap
             )


AlexDaniel:
I wonder, is there no performance penalty for this?

lizmat:
Yes, some.  But for hot code paths, the guards should quickly inline the normal `compare` candidate.  But please, yes, profile it  :-)

lizmat:
I will answer your questions and remarks extensively this time, because I hope we can get to some understanding and a way of moving forward.

> The reason for all this special-casing is because Junctions are these weird objects that you can for some reason pass around.  This just shouldn't be the case.

If you wouldn't be able to pass Junctions around, they wouldn't make any sense.  Because even in something as simple as `my $foo = 4 + 3 | 5`, you've already passed `3|5` as a parameter to `infix:<+>`.  So it **needs** to be the case.

> People just need syntactic sugar to be able to compare against multiple values, plus a way to work with values in arrays, junctions don't need to be magical objects.

The magic is simple: if something accepts a `Mu`, it will get a raw `Junction` when given.  If it does **not** accept `Mu`, then the system will call whatever is being called multiple times for each eigenstate of the `Junction`.

> Also, I don't believe that Raku should be a testing ground for weird ideas

Then you're with the wrong programming language.  Raku intends to be where the puck will be, not where the puck has been.

> I want it to be a practical language.

So do I.  But apparently you think that Raku can not be both.  If you really think that, then again, you're with the wrong programming language.

> The fact that the compiler can't properly optimize stuff away forces us to shit into our own well by complicating the implementation of basic features.

To an extent, this is called bootstrapping.  And it is part of torturing the implementor for the user.  **I** don't mind doing this.  If you think that Raku is the only language that is "shitting into their own well", then I'd urge you to look a little closer to the implementations of quite a few other programming languages.

> The reason these compare multis exist is simply because the compiler is not capable enough.

No.  The reason these multis exist, is because the sorting algorithm expects `cmp` to return something that is `-1`, `0` or `1`, **not** a `Junction`.  However, the fact that `cmp` returns a `Junction` when at least one of its arguments is a `Junction`, is completely and utterly valid.  The real problem with sorting lists with `Junction`s in it, is that we cannot collapse a `Junction` of `Order` values into a single `Order` value.  If **you** know how to do that, then please enlighten me.

In any case, this is yet another attack by you on Raku: the compiler is capable enough to compile itself and actually make things work in production.  That's more than the first C compiler that I have worked with, actually did.  Which cost quite a lot of money at the time.

> So should we be doing stuff like this just to get an immediate perf gain

A 4x performance gain for such a simple, and straightforward change that hardly involves any extra `nqp` ops, I'd say **yes**.

> or should we make some sort of plans ahead in order to make sure that we stop polluting the source as soon as possible

I think we're way beyond that point.  Fortunately, thanks to @jnthn 's work on runtime optimizations, some of the "dirty" optimizations that I have done in the past, have been removed already.  And I hope te be removing a lot more of them in the future.  You can plan all you want, but if you want a useable and practical language, some dirty tricks *are* allowed under the hood in my book.  And in Larry's book as well, in case you wondered.

> start removing previously added perf hacks

What is the point of removing performance hacks?  Do you really think a user of Raku cares about how nice the code looks internally?  Really??

> I realized that this PR doesn't exactly inline the behavior of cmp, it defines it's own.

Indeed, that was an oversight on my part.  Fixed with 3f60d0a70b .

> I'm pretty sure there are more edge cases and this is just the one I managed to find.

I'm pretty sure there are not anymore.

> One should also think how custom cmp and stuff like that will work with this. To be honest, I failed to get sort to use my own cmp multis, but maybe I'm just doing something wrong.

If you have custom `cmp`s, you need to specify them explicitly.  The setting code only "sees" the `infix:<cmp>` of the core.

> Anyway, the compare sub will only fall back to cmp if none of the signatures matched, meaning that any types that derive from Int, Str, Num and such will always be forced to use one of the compare candidates and not your custom cmp (if that is at all allowed, I'm not sure).

I have no idea what you mean here, you appear to contradict yourself.
------------------------------------------------------------------

So far the direct responses.  Now for the innuendos.

You appear to want to change the Raku Programming Language into a direction that the majority of core developers and as far as I know, most of the users are *not* interested in.  If a user wants a smaller and simpler language, there are many smaller and simpler languages out there, some of which with large corporate backing.

You have been told again and again, that your style of communication is abrasive.  To the point that to my knowledge, now at least 3 people have left the #raku channel, so they don't have to exposed to your abrasiveness anymore.  As such, you have become a liability for the Raku community.

You bring up good points most of the time.  But you seem to have a fixed view in your mind of how things should happen, and start throwing tantrums if things don't go your way.  **GROW UP**  Try to find compromises, realize that Rome wasn't built in a single day, most things usually take more time than you ever expected.

Finally: you appear angry most of the time.  I assume there's at least something going on in your life that makes you thoroughly unhappy.  Please don't project that unhappiness onto the Raku community.

Thank you.

AlexDaniel:
> But you seem to have a fixed view in your mind of how things should happen, and start throwing tantrums if things don't go your way. GROW UP […] you appear angry most of the time.

Um, I was just trying to leave a review? The review was mostly positive, and I did find an actual flaw and saved us from a regression. My understanding was that you don't find the PR workflow the most comfortable so I try to review them ASAP so you can move forward. I don't understand.

Yes, I did say that we sort of cornered ourselves into the need for this PR because of the existence of Junctions as objects and the state of the compiler. But why get so mad about this? I'm trying to be as factual as possible.


I'll be very short: No, junctions don't have to be this way and can function differently (yes I mean that it can be a completely different feature that solves similar problems); yes, I'm with the wrong community; yes I think it can't be both; I urge you to look at other languages too; I was talking about perf multies that exist for the sole purpose of improving performance, I guess I wasn't clear; I'd say probably not, but it really depends; Ah, thanks, I didn't know just specifying my own `cmp` is not enough, that's OK.

AlexDaniel:
Now, I'm sorry that my comments are considered abrasive, but please bear with just one more because you said:

> I have no idea what you mean here, you appear to contradict yourself.

I guess I didn't explain it well so maybe it's a good idea to clarify what I meant.

You created a set of `compare` multis that ends with the last candidate that calls `cmp`. This means that you have two layers of multiple dispatch happening (a set of `compare` multis and a set of `cmp` multis). Multidispatch won't flatten these two sets (obviously?), so you have to be *very* careful because some signature can loosely match a given set of arguments. So when you say:

> I'm pretty sure there are not anymore.

I'm not. In fact given my previous experience I'd say I'm sure there are more cases that I just don't know about yet. And this is also why I say that the perf benefit is not a good argument for merging, you have to weigh in whether the PR complicates the code and whether it adds more room for edge cases.

So let's see. You have IntStr and NumStr candidates, good, but what can be a Str that will accidentally match the Str candidate? Right, RatStr!

```raku
c: sorting-with-junctions,HEAD dd (<6/5.5>, <25/25.0>, <1/0.5>).sort
```
```
¦sorting-with-junctions: «(RatStr.new(2.0, "1/0.5"), RatStr.new(1.0, "25/25.0"), RatStr.new(<12/11>, "6/5.5")).Seq␤»
¦HEAD(9d6d8dd): «(RatStr.new(1.0, "25/25.0"), RatStr.new(<12/11>, "6/5.5"), RatStr.new(2.0, "1/0.5")).Seq␤»
```

So my quick understanding is that it is using the Str value to do the sorting, which is very problematic. Please, don't be so sure.

Also, please judge my reviews on their own merits and not on the tone you *think* I had in mind when I wrote them. I have a feeling that anything I say you view from the negative perspective, but that's not the case.

lizmat:
Getting back to @AlexDaniel 's comments.

Sorry to not have answered sooner, but I was effectively afk for the whole of Tue.

> Um, I was just trying to leave a review?

It started out as just a review, but turned into something else after the edits, in my opinion.

> The review was mostly positive, and I did find an actual flaw and saved us from a regression.

And for that I thank you.

> My understanding was that you don't find the PR workflow the most comfortable so I try to review them ASAP so you can move forward.

Indeed, by my main beef with the PR workflow is really about the discussion dying out in some kind of agreement, and then not being merged.

> I don't understand.

I think that's part of the problem that several people have with your abrasive comments.

> Yes, I did say that we sort of cornered ourselves into the need for this PR because of the existence of Junctions as objects and the state of the compiler. But why get so mad about this? I'm trying to be as factual as possible.

"This just shouldn't be the case. People just need syntactic sugar to be able to compare against multiple values, plus a way to work with values in arrays, junctions don't need to be magical objects. Also, I don't believe that Raku should be a testing ground for weird ideas, I want it to be a practical language."

This isn't factual.  This is expressing an opinion.

> I'll be very short: No, junctions don't have to be this way and can function differently (yes I mean that it can be a completely different feature that solves similar problems); yes, I'm with the wrong community; yes I think it can't be both; I urge you to look at other languages too; I was talking about perf multies that exist for the sole purpose of improving performance, I guess I wasn't clear; I'd say probably not, but it really depends; Ah, thanks, I didn't know just specifying my own cmp is not enough, that's OK.

That's a lot of words.

> You created a set of compare multis that ends with the last candidate that calls cmp. This means that you have two layers of multiple dispatch happening (a set of compare multis and a set of cmp multis). Multidispatch won't flatten these two sets (obviously?), so you have to be very careful because some signature can loosely match a given set of arguments.

There's no technical reason why those wouldn't be flattened.  I think `cmp` having so many candidates, is not helping.  @jnthn please correct me if I'm wrong.

> So my quick understanding is that it is using the Str value to do the sorting, which is very problematic. Please, don't be so sure.

Point taken.  18cb45afdc and 42f1bd58fc should address the issues you've raised.

> Also, please judge my reviews on their own merits and not on the tone you think I had in mind when I wrote them. I have a feeling that anything I say you view from the negative perspective, but that's not the case.

I try not to, but it's hard given your past outbreaks on IRC and in privmsgs.  But I will try to do so harder.  I suggest you also try to do harder: "But why get so mad about this?".  Why do you think I was mad?

AlexDaniel:
@lizmat no need to reply to the entirety of my comments :) Lessons that were learned will linger for some time, but any conversation [needs an end](https://communitywiki.org/wiki/DiminishingReplies) :)

In my opinion it's OK for reviews to have some additional information for other reviewers and also some thoughts on what can happen in the future. I think we should want to encourage people to leave reviews, more eyes on the code is always better, but this also means that people can get confused as to what the underlying problem is and why all this additional code is needed (and potentially leave negative reviews). My two points answer that – the fix is for the edge case that is caused by the past decision to have junctions as objects, and all the extra code is required because we have to manually help the compiler in its current state. Yes, I did express how unfortunate both of these things are, but in the context of this PR I think it's fair.

> There's no technical reason why those wouldn't be flattened.

It's not about compiler details, it's the semantics you get from calling a different function (in this case an operator) from another one. I think you just have a different viewpoint when looking at this code, which isn't wrong, but the one I'm proposing allowed to “see” the bug without running any code (it actually took me a lot of trial and error to come up with a snippet for the bug I saw).

> outbreaks on IRC

These were rants, not outbreaks. I think it should be OK for people to express their frustration. I also hope we all can find some humor in the things I say. Privmsgs is another thing, there I just let my human out. :)

AlexDaniel (changes requested):
Sorry, am I doing something wrong?

> Could not find symbol '&Junction' in 'X::Cannot'

https://gist.github.com/Whateverable/20c508c40ef97d5b3122c53e4ac28709

lizmat:
> Yes, I did express how unfortunate both of these things are, but in the context of this PR I think it's fair.

That was not my argument.  My argument was that you stated these **opinions** to be **facts**.

> These were rants, not outbreaks.

These "rants" were enough for at least three people to leave the #raku IRC channel, one of which is Jonathan Worthington.  And in two of these cases, caused people to stop doing development work.  Rants shouldn't do that.  Rants should be agreeable, comedic even, with a some truth in them.  Think IgNobel research results.  Not Fox News.

> I think it should be OK for people to express their frustration.

You should be glad that I do not express my frustration with your actions more regularly.  Do you think it is OK that I should do that?  Don't you think that will make the atmosphere on the channel even worse than it is now whenever you start ranting again?

> Privmsgs is another thing, there I just let my human out. :)

Not funny.  And not even right.  If you really take this to be true, then please go away.  You have no place in the Raku community.  Or you can wait until the Raku community has an enforceable CoC.

lizmat:
> Sorry, am I doing something wrong?

No, I forgot to put back the `X::Cannot::Junction` stub.  72b8304b36 fixed that.  Thanks for the spot!

AlexDaniel:
> > Privmsgs is another thing, there I just let my human out. :)
>
> Not funny. And not even right. If you really take this to be true, then please go away. You have no place in the Raku community. Or you can wait until the Raku community has an enforceable CoC.

As if I'm privmsg-ing people all the time to bother them? Come on. The vast majority of PM conversations I had were not started by me.

> You should be glad that I do not express my frustration with your actions more regularly. Do you think it is OK that I should do that?

Of course I'm OK with others giving me raw feedback. Stop saving your words and hit where it hurts.


My rants are typically about the decisions in Rakudo which I think are wrong. If you think I'm not right then no big deal. But IMO the reason people get so butthurt is because some of my points are actually valid. I think the devs need a reality check. Nothing is doomed, there is a nice way forward, but people need to admit mistakes in order to progress.

AlexDaniel (changes requested):
As for this PR, I said multiple times that this approach is not going to work. Some ways of doing things just can't result in a desired result, and that's OK. I tried to indicate this early on to save you time. Instead you iterated on it and tried to plug all the holes I noticed.

There will be a moment when I will no longer be able to point out the regressions, but it's unlikely there will ever be time when all the issues are truly fixed in this PR. Generally, there are commits that simplify the code and reduce the amount of potential edge cases, and there are ones that do the opposite. This PR is simply using the latter. It's just that with every next iteration we get more complicated code and also more obfuscated bugs.

Reasons you should not continue:

1. As I already explained, the new `compare` multis are essentially reimplementing the `cmp` table. It is hard to get it completely right (which I hope was demonstrated), and even if you do it perfectly, further bugfixes in rakudo (which are [not uncommon](https://github.com/rakudo/rakudo/commit/085145f3a5)) require the developers to somehow know that they also need to port their fixes to the `compare` table to make sure that sorting is also fixed.

2. It also doesn't help that none of the issues I mentioned were uncovered by roast, meaning that `sort` is tested very poorly. It's possible there are more bugs but we're simply not testing for them.

3. You are introducing a role for allomorphs just because you need it for the perf improvement. Maybe this can come in handy for users too (the role is user-visible), but it being a means to just get a perf gain is a massive red flag.

4. This particular perf issue is not worth your time in the long run. Eventually the compiler will get somewhat faster, maybe even the language will get simpler which will make some code paths unnecessary. Just try to push it there instead.


Anyway, here's a regression:

```raku
c: sorting-with-junctions,HEAD say sort(NaN, Inf, -Inf)
```

```
¦sorting-with-junctions: «(NaN -Inf Inf)␤» ¦HEAD(88aa6db): «(-Inf Inf NaN)␤»
```


This is my last review. I won't be approving this PR for reasons stated above, which is fine, you just need another reviewer to discuss it (although with no merge policy you can just do whatever you want).

I learned quite a bit because I was looking at the Rakudo code (I no longer find black-box testing fun). Thanks!

AlexDaniel:
> It's been enlightening.

I hope so. Thing is, most subroutines can get junctions in their inputs, and will often end up *returning* a junction too when that happens. This makes the language very difficult to use – either you ignore junctions altogether and your code (especially if it's a library) can end up having bugs or behavior that you didn't intend (when junctions get involved), or you keep it in mind which adds to the mental load when writing your code. The former is what most people end up doing, including core developers. For example, when rakudo got return constraints on a bunch of subs/methods, it made it impossible to get a Junction from them (e.g. cmp of arrays). I don't think it was intentional, it's just that we tend to put it into a DIHWIDT category.

But if we all don't want to expect Junctions passed around as data, then perhaps it shouldn't even be possible. That would not only make the language easier to use and generally more predictable, but will also simplify any code that attempts to handle junctions separately.

I guess there are different thresholds at which different people will consider something as a misfeature. I don't know what's yours, but Junctions blew through mine.

[The recent blog](https://raku-advent.blog/2020/08/08/rfc-223-superpositions/#user-content-conclusion) post implicitly acknowledges some of this. I highlighted some text in bold:

> Junctions are, despite their **internal complexity** and rarity in programming languages are something that are so well thought out and integrated into common Raku coding styles **that most use them without any thought**.

… Oh no!

The article also goes about how it isn't just syntactic sugar, and provides the most common examples of using junctions, all of which can be implemented with syntactic sugar or a set of helper subs. I agree that writing `$x == 42 | 59` is very nice, but it *can* be done without junctions as objects.

I also object to the “well thought out” idea. I think Junctions were and are still an experiment, in fact I'd say one that belongs to a toy language (because of how it affects essentially any code), which Raku isn't.

<details>
  <summary>A bit of humor maybe to wrap it up? You know how some languages go to great lengths in order to help developers not get Nulls unexpectedly in their code? In Raku you can get multiple Nils at the same time, as a single value! :)</summary>

  <br>In some sense it's not even a joke. Some people [use `~~ Nil`](https://gist.github.com/3313ab7d8fcfe66d9bb092552f7ed7f8) to check if something is undefined. Let's not judge them for that because TIMTOWDTI. However, `all(Nil, 42) ~~ Nil` is False, meaning it's defined. Code like this will then go ahead and try to use the Junction, only to get a Nil which was “checked for”.
  
</details>

lizmat:
>> It's been enlightening.
> I hope so.

It was so because of the way allomorphs are implemented, **not** how Junctions are implemented, as I've done quite a bit of optimizing in that area.

> either you ignore junctions altogether

That was what the architects were going for.  And implementers of the language have been punished for that.

> For example, when rakudo got return constraints on a bunch of subs/methods, it made it impossible to get a Junction from them (e.g. cmp of arrays). I don't think it was intentional, it's just that we tend to put it into a DIHWIDT category.

Yeah, maybe `Junction`s should be special cased like that, just like `Failure`s.

> I agree that writing `$x == 42 | 59` is very nice, but it can be done without junctions as objects.

Then you should be glad to know that examples such as that are actually statically optimized to `$x == 42 || $x == 59`

> I also object to the “well thought out” idea.

You're entitled to your opinion.

> By the way, ever tried negating a Junction? It's just not well thought out.

Not sure what you mean:

    dd -(4|5)   # any(-4,-5)

> Some people use ~~ Nil to check if something is undefined.

You cannot protect people from writing bugs in their programs.

> A bit of humor maybe to wrap it up?

Let's try to keep the joke out of these discussions please.  Many people reading this might not get the joke.  And other may be offended by it.  Let's keep discussions here factual and technical, please.

AlexDaniel:
> Yeah, maybe Junctions should be special cased like that, just like Failures.

Generally, I recommend to aim for *less* special cases, not more.

> Then you should be glad to know that examples such as that are actually statically optimized to `$x == 42 || $x == 59`

In this case it's the semantics that bother me the most, not performance. But yes, that's good, pretty much proving my point that you can provide user convenience of junctions without passing them around as objects.

> Not sure what you mean:

Ah, I wasn't clear at all, sorry. It's about using negated ops with junctions.

For example, let's say you want to test that exactly *one* of the strings is not equal to something:

```raku
say <1 2 3 4>.one ne ‘42’ # True
```
Wait, but… why is it True? All of the values are not 42, not just one! Moreover, we are not even getting a Junction back.

The explanation is that negated ops are somehow special-cased to behave this way. But in this case `ne` doesn't even have a `!` so it looks even more confusing.

Here's a wonderful conversation I was able to find: https://colabti.org/irclogger/irclogger_log/perl6-dev?date=2017-06-23#l333

It has everything! There are mentions of:
* how only one case was properly considered
* jarring special casing
* how special casing turns features into something to avoid
* perl5 and an obligatory comparison to it :)
* a linguistic explanation that absolutely doesn't help
* how it's too late to change something once it's already there (“the additive nature of Raku development” is what I call it)
* people giving up and just putting it into the box of Raku warts
* a need for a well-defined behaviour
* there's also someone (me in this case) religiously defending Raku :)
* and also someone (again me…) not understanding any of the above

> You cannot protect people from writing bugs in their programs.

But you can! That's the point. Some languages have more pitfalls than others, and there are ways to avoid them.

PR #3852, opened 2020-08-18
Fix Baggy.pickpairs with a Callable

stoned:
Baggy.pickpairs' Callable should be invoked with the Baggy number of
elements as the pairs are picked independently of the elements' weight
in the Baggy object.

See proposed test included in Raku/roast#668


lizmat (approved):
Good catch!

PR #3853, opened 2020-08-19
Suggest method names without checking for capitalization

lizmat:
Currently, an expression like `say 42.which` will fail with:

  No such method 'which' for invocant of type 'Int'

*without* suggesting .WHICH.  This patch changes that to:

  No such method 'which' for invocant of type 'Int'.  Did you mean 'WHICH'?

### DIFF
--- a/src/core.c/Exception.pm6
+++ b/src/core.c/Exception.pm6
@@ -153,6 +153,10 @@ my class X::Method::NotFound is Exception {
     has $.typename;
     has Bool $.private;
     has $.addendum;
+    has $.message is built(False);
+    has @.suggestions is built(False);
+    has @.tips is built(False);
+
     # This attribute is an implementation detail. Not to be documented.
     has $.in-class-call;
 
@@ -162,14 +166,13 @@ my class X::Method::NotFound is Exception {
           !! "of type '$.typename'"
     }
 
-    method message() {
+    method TWEAK() {
         my @message = $.private
           ?? "No such private method '!$.method' for invocant $.of-type"
           !! "No such method '$.method' for invocant $.of-type";
 
         @message.push: $.addendum if $.addendum;
 
-        my @tips;
         my $indirect-method = $.method.starts-with("!")
                                 ?? $.method.substr(1)
                                 !! "";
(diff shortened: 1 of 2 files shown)


src/core.c/Exception.pm6
```diff
@@ -203,15 +203,26 @@ my class X::Method::NotFound is Exception {
...
+                next
+                  if $method_candidate.^name eq 'Submethod'  # a submethod
+                  && !$invocant_methods{$method_name};       # unknown method
+
+                if $.method.fc eq $method_name.fc {
+                    %suggestions{$method_name} = "";  # assume identity
+                }
+                else {
+                    my $dist =
+                      StrDistance.new(:before($.method), :after($method_name));
```
vrurg, line 221: Cancelled an accidental review and lost my comment. So, just repeating myself again: what if we just use `.fc` on both variants here?

```
StrDistance.new(:before($.method.fc), :after($method_name.fc));
```

It would then cover erratas like `.whicj` too.

lizmat: Good point.

lizmat: Added with 30c4b6e599

vrurg (approved):

t/05-messages/01-errors.t
```diff
@@ -222,7 +222,7 @@ throws-like { Blob.splice }, X::Multi::NoMatch,
         'a public method of the same name as the missing private method is suggested';
     throws-like q| class RT123078_3 { method !bar { }; method baz { } }; RT123078_3.new.bar |,
         X::Method::NotFound,
-        message => all(/<<"No such method 'bar'" \W/, /<<'RT123078_3'>>/, /\s+ Did \s+ you \s+ mean \s+ "'baz'"/),
+        message => all(/<<"No such method 'bar'" \W/, /<<'RT123078_3'>>/, /\s+ Did \s+ you \s+ mean/),
```
codesections, line 225: This test was previously testing that "a private method of the same name as the public missing method is not suggested for out-of-class call" – that is, that `!bar` was _not_ suggested.  This revised test no longer ensures that `!bar` is not suggested, which could allow a regression in future code changes.

Here is the alternate test I proposed in my accidental duplicate PR, which I believe addressed that issue. 

```raku
       message => all(/<<"No such method 'bar'" \W/, /<<'RT123078_3'>>/, /<<'baz'>>/,  none(/<<'!bar'>>/)),
```

codesections (commented):
I noticed one issue where the test could allow a regression; I've described the issue and proposed a solution inline.

vrurg (changes requested):

vrurg (approved):
Hopefully, I miss nothing.

vrurg:
@lizmat If there is no other reason to keep this opened, I'd say it's ready to merge.

PR #3854, opened 2020-08-19
Add Mu.WHAT for introspection

lizmat:
Yes, foo.WHAT will be codegenned without actually calling any method.
But to not have a method Mu.WHAT also means it won't show up in any
introspection or error messages.

Together with b7ff282fc5, this will suggest WHAT for "say 42.what".

### DIFF
--- a/src/core.c/Mu.pm6
+++ b/src/core.c/Mu.pm6
@@ -11,6 +11,10 @@ my class Mu { # declared in BOOTSTRAP
 
     method sink(--> Nil) { }
 
+    # This method is only here so that it can be introspected.  The codegen
+    # will actually generate an nqp::what in all cases.
+    method WHAT() { self.WHAT }
+
     proto method perl(|) {*}
     multi method perl(Mu \SELF: |c) { SELF.raku(|c) }
     # although technically not a documented method, some module authors have


src/core.c/Mu.pm6
```diff
@@ -11,6 +11,10 @@ my class Mu { # declared in BOOTSTRAP
 
     method sink(--> Nil) { }
 
+    # This method is only here so that it can be introspected.  The codegen
+    # will actually generate an nqp::what in all cases.
+    method WHAT() { self.WHAT }
```
vrurg, line 16: What if codegen changes later? Perhaps it worth just having `nqp::what` here despite it won't be invoked anyway? Just for the safety of it.

lizmat: If that codegen changes, then this should change as well, wouldn't it?

vrurg: Sure. But my point is that it could be overlooked. I'd prefer something that won't silently go into deep recursion.

lizmat: So you're implying that the method would actually be called at some point ?

vrurg: Exactly. And it's not only due to potentially changed codegen. It could be invoked indirectly, or via '$obj."$what-method-name"()`, etc.

jnthn: The code-gen change risk is low, but I'm a bit uncomfortable with this; I'd prefer that the name suggestion code carried the small list (`WHAT`, `HOW`, and `WHO`) of things that are true MOP primitives rather than adding these methods.

lizmat:
Will create a new PR for this.

PR #3856, opened 2020-08-20
Have Test::done-testing return Bool.

ziuq:
Done-testing returning Bool will easily allow instrumenting test files to eliminate  repeating tests unnecessarily.  This is aimed at the coder who tests several times an hour while writing and should not interfere with more formal testing.

### DIFF
--- a/lib/Test.rakumod
+++ b/lib/Test.rakumod
@@ -775,7 +775,8 @@ sub proclaim(Bool(Mu) $cond, $desc is copy, $unescaped-prefix = '') {
     $cond
 }
 
-sub done-testing() is export {
+sub done-testing(-->Bool:D) is export {
+    my $return = True;
     _init_io() unless $output;
     $done_testing_has_been_run = 1;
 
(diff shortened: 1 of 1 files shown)


lizmat (changes requested):
Why using a variable, when you could just `return False` inside the `if` statements.  And do a `True` at the end of the sub?

ugexe:
> This is aimed at the coder who tests several times an hour while writing and should not interfere with more formal testing.

Can you give a better example of what you're trying to do? For instance I would think you can already tell if done-testing retuned true/false by examining the exit code of the test.

> Why using a variable, when you could just return False inside the if statements. And do a True at the end of the sub?

Those if statements are not exclusive

lizmat:
> Those if statements are not exclusive

Good point, ugexe++

ziuq:
@lizmat:  For laziness and resilience, I maintained the control flow of the routine--the `if`s did not obviously (to me) prevent fall through.
@ugexe:  That is true.  But by capturing status sooner it is possible to  use a 'make'-like system within the test file which determines if anything has changed since the last successful  run and turn off the tests and substitute a single successful test.  This system when not enabled becomes a no-op; where that does not  comply the coder can throw this into a personal branch, etc.

So it is for the developer who tweaks and tests, tweaks and tests ... and mostly about decreasing the distraction of longer interruptions.

vrurg:
I don't see how `done-testing` returning a meaningful value could harm anything. I still don't get how exactly @ziuq is going to use it, but I can think of tests providing additional diagnostics or instructions conditionally. So, if nobody explicitly objects I'd merge it.

@lizmat I think your request for changes can be safely ignored now?

lizmat (approved):

vrurg:
@ziuq Thank you!

ugexe:
> I don't see how done-testing returning a meaningful value could harm anything.

To play devils advocate: there are other things this could return... why not an Int of passed tests or some nested data structure for instance? Returning a boolean value certainly helps demonstrate the use case, but it doesn't mean that is the ideal implementation or suggest alternatives were considered in a design.

That said I'm not against this. But I wanted to point out that this wasn't some clear-cut thing just because it was small and meaningful.

AlexDaniel:
> why not an Int of passed tests or some nested data structure for instance

If there's a need for more data in the future we can make it return an object that is truthy when necessary, and it can include extra fields for any other information.

Yeah, it's not clear-cut, but there's a reasonable path we can take without much pain.

vrurg:
> > I don't see how done-testing returning a meaningful value could harm anything.
> 
> To play devils advocate: there are other things this could return... why not an Int of passed tests or some nested data structure for instance? Returning a boolean value certainly helps demonstrate the use case, but it doesn't mean that is the ideal implementation or suggest alternatives were considered in a design.

I considered this too. But the common approach for other routines is to return a boolean, so I don't see why `done-testing` should be different here.

Would we ever consider having the number of sucess/failures/whatever else be available to a testing code it would be better to provide a specialized routine or a set of routines. This kind of data could be useful not only by the end of a suite. For example, if failure of any of a number of tests should result in the whole suite failure then it would be enough to have just one single check.

Though I anyway like the approach of my `Test::Async` more with a specialized managing object with corresponding public attributes: `test-suite.tests-failed`.

PR #3857, opened 2020-08-20
Streamline X::Multi::Ambiguous listing

lizmat:
- remove ":", it does not convey any information, clear it's a signature
- add two spaces to set it off the first line
- add "is default" status of the candidate when applicable


AlexDaniel (approved):
Looks good to me.

Issue #3859, opened 2020-08-20
X::Method::NotFound errors do not perform case-insensitive searches when suggesting methods


codesections:
# The Problem

When an `X::Method::NotFound` exception is thrown, it's `message` method suggests up to four similar methods that the user might have meant.  However, it computes "similar" methods based on a case-sensitive comparison of the user-supplied method and methods available on the object.

This case sensitivity can result in methods not being suggested even though they are largely similar (or even identical) to
the method the user supplied, apart from case. This can be a particular problem for methods that new users may have forgotten are all uppercase.  For example, `why` and `WHY` share no letters when compared in a case-sensitive way.
## Expected Behavior

```
Int.why  # THROWS: «No such method 'why' for invocant of type 'Int'.  Did you mean 'WHY'?»
```

## Actual Behavior

```
Int.why  # THROWS: «No such method 'why' for invocant of type 'Int'»
```

Because this is a very minor issue, and because I have been looking for a tiny issue to get my feet wet contributing to Rakudo, I am also submitting an accompanying PR that resolves the issue.

codesections:
Apologies for the clutter; this is already addressed in a pending PR #3853

PR #3860, opened 2020-08-20
Make X::Method::NotFound case-insensitive

codesections:
When an `X::Method::NotFound` exception is thrown, it's `.message` method suggests up to four similar methods that the user might have meant.  However, it computes "similar" methods based on a case-sensitive comparison of the user-supplied method and methods available on the object.

This case sensitivity can result in methods not being suggested even though they are largely similar (or even identical) to
the method the user supplied, apart from case. This can be a particular problem for methods that new users may have forgotten are all uppercase.  For example, `why` and `WHY` share no letters when compared in a case-sensitive way.

This commit resolves this issue by calling `.fc` before performing the comparison.  This may *very* slightly slow the comparison, but I am not concerned – when generating the message from `X::Method::NotFound`, the program is very likely to be in an unrecoverable state, where extremely slight performance issues are unlikely to have an impact.

This commit updates the corresponding test, and hopefully makes it slightly less fragile – it no longer tests against the suggested
methods being an exact match, but just tests that private methods are correctly excluded.

Solves #3859

### DIFF
--- a/src/core.c/Exception.pm6
+++ b/src/core.c/Exception.pm6
@@ -208,7 +208,7 @@ my class X::Method::NotFound is Exception {
                 my $method_name = code-name($method_candidate);
                 # GH#1758 do not suggest a submethod from a parent
                 next if $method_candidate.^name eq 'Submethod' && !@invocant_methods.first($method_name, :k).defined;
-                my $dist = StrDistance.new(:before($.method), :after(~$method_name));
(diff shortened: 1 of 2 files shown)


lizmat:
Thank you for your work.  But this appears to provide a subset of what https://github.com/rakudo/rakudo/pull/3853 provides?

codesections:
Oops, you are absolutely correct – I had not seen that PR.  I'm closing this, though writing this did give me one comment on your PR, which I will make in that thread.

PR #3873, opened 2020-08-24
Introducing the Allomorph class

lizmat:
- a subclass of Str
- IntStr/NumStr/RatStr/ComplexStr are now subclasses of allomorph
- replaced several nqp::ifs by ternaries
- streamlined many coercions, specifically RatStr.Numeric
- simplified 16 infix:\<eqv> candidates into a single one
- need further research to do same for infix:\<cmp> and infix:<===>
- make test and make spectest clean

The reason for this is to make it easier and more extendable to handle
allomorphs.  Having an `Allomorph` class should allow custom numeric
objects to also include their allomorphic version.  And having an
`Allomorphic` class should make typechecking for allomorphs more centralized
as is proven by the infix:\<eqv> simplification.


vrurg:
I wonder why all-lowercased name?

AlexDaniel:
@lizmat there is a change in behavior when subclassing the allomorphs, please consider it. That said, I don't know which one is better, so you decide. The difference appears because of the dispatch table (again):

```raku
c: introducing-allomorph-class,HEAD class Foo is IntStr {}; say IntStr.new(42, ‘forty two’) eqv Foo.new(42, ‘forty two’)
```
```
¦introducing-allomorph-class: «False␤»
¦HEAD(c1784bd): «True␤»
```

AlexDaniel:
BTW `^mro` is also affected, I don't know if it's something that can change easily. However, what worries me a little bit is that Int and Str are now in a different order… does it make a difference? I don't know.

```raku
c: introducing-allomorph-class,HEAD say IntStr.^mro
```
```
¦introducing-allomorph-class: «((IntStr) (allomorph) (Str) (Int) (Cool) (Any) (Mu))␤»
¦HEAD(c1784bd): «((IntStr) (Int) (Str) (Cool) (Any) (Mu))␤»
```

vrurg:
> @lizmat there is a change in behavior when subclassing the allomorphs, please consider it. That said, I don't know which one is better, so you decide. The difference appears because of the dispatch table (again):
> 
> ```perl6fe
> c: introducing-allomorph-class,HEAD class Foo is IntStr {}; say IntStr.new(42, ‘forty two’) eqv Foo.new(42, ‘forty two’)
> ```
> 
> ```
> ¦introducing-allomorph-class: «False␤»
> ¦HEAD(c1784bd): «True␤»
> ```

A result of `nqp::eqaddr` used in `eqv` op. Should be `nqp::istype`, I guess?

lizmat:
I would argue that I actually fixed a bug wrt `eqv`:

    class Foo is Int { }
    say Foo eqv Int;   # False

jnthn:
> I would argue that I actually fixed a bug wrt eqv

Yes; `eqv`, like `===`, is only ever `True` if its operands have the exact same type.

src/core.c/allomorphs.pm6
```diff
@@ -1,11 +1,39 @@
...
+    multi method Str(Allomorph:D:) {
+        nqp::getattr_s(self,Str,'$!value')
+    }
+
+    multi method WHICH(Allomorph:D:) {
+        nqp::box_s(
+          nqp::join('|',nqp::list_s(
+            self.^name,
+            self.Numeric.WHICH,
+            self.Str.WHICH
```
jnthn, line 16: Is there a case where the type and `Str` value would be the same, but `self.Numeric.WHICH` would be different? I struggle to imagine one.

src/core.c/allomorphs.pm6
```diff
@@ -1,11 +1,39 @@
...
+    multi method raku(Allomorph:D:) {
+        nqp::join("",nqp::list_s(
+          self.^name,'.new(',self.Numeric.raku,', ',self.Str.raku,')'
+        ))
+    }
+}
+
 # the uses of add_I in this class are a trick to make bigints work right
-my class IntStr is Int is Str {
+my class IntStr is Allomorph is Int {
```
jnthn, line 30: Why not `is Int is Allomorph`?

src/core.c/allomorphs.pm6
```diff
@@ -1,11 +1,39 @@
...
     method new(Int:D $i, Str:D $s) {
         my \SELF = nqp::add_I($i, 0, self);
         nqp::bindattr_s(SELF, Str, '$!value', $s);
-        SELF;
+        SELF
     }
-    multi method ACCEPTS(IntStr:D: Any:D \a) {
+    multi method ACCEPTS(IntStr:D: Any:D \a) is default {
```
jnthn, line 36: Not thrilled to see `is default`; what was the conflict?

src/core.c/allomorphs.pm6
```diff
@@ -189,26 +191,11 @@ multi sub infix:<cmp>(ComplexStr:D $a, RatStr:D     $b) { $a.Complex cmp $b.Rat
...
-multi sub infix:<eqv>(NumStr:D     $a, IntStr:D     $b --> False) {}
-multi sub infix:<eqv>(NumStr:D     $a, RatStr:D     $b --> False) {}
-multi sub infix:<eqv>(NumStr:D     $a, NumStr:D     $b) { $a.Num     eqv $b.Num     && $a.Str eqv $b.Str }
-multi sub infix:<eqv>(NumStr:D     $a, ComplexStr:D $b --> False) {}
-
-multi sub infix:<eqv>(ComplexStr:D $a, IntStr:D     $b --> False) {}
-multi sub infix:<eqv>(ComplexStr:D $a, RatStr:D     $b --> False) {}
-multi sub infix:<eqv>(ComplexStr:D $a, NumStr:D     $b --> False) {}
-multi sub infix:<eqv>(ComplexStr:D $a, ComplexStr:D $b) { $a.Complex eqv $b.Complex && $a.Str eqv $b.Str }
+multi sub infix:<eqv>(Allomorph:D $a, Allomorph:D $b --> Bool:D) is default {
```
jnthn, line 194: I get why the `is default` here is needed; it makes me a little uneasy to see it, but I didn't think up an example where it gets us into bother.

jnthn (commented):
I like that this makes things more uniform and more general (and would ease adding a further allomorph out of core, if somebody is implementing other numeric types and wishes to have one). Kind of wish `Allomorph` was a role, is all.

src/core.c/allomorphs.pm6
lizmat: Only if you artificially construct allomorphs.  Which people can do, and therefore will.

src/core.c/allomorphs.pm6
lizmat: Because I was under the impression that that would make `Allomorph` tighter?

src/core.c/allomorphs.pm6
lizmat: $ r '<42> ~~ <66>'
Ambiguous call to 'ACCEPTS(IntStr: IntStr)'; these signatures all match:
  (Str:D: Str:D \other, *%_)
  (IntStr:D: Any:D \a, *%_)

src/core.c/allomorphs.pm6
jnthn: I don't think so; tightness in the multi-dispatch sense is determined by comparing two types in the same position in a signature, and the MRO won't impact upon that type judgement.

src/core.c/allomorphs.pm6
jnthn: Ah yes, those are. Though we could add an `IntStr:D: Str:D` candidate to overcome that.

src/core.c/allomorphs.pm6
jnthn: Goodness, why... :-) But yes, I guess it doesn't hurt too much to be robust against it.

src/core.c/allomorphs.pm6
jnthn: Especially as we *already* have a branch for `Str` anyway.

lizmat: > Especially as we already have a branch for Str anyway.

I'm not quite sure I follow.  Where would this branch for `Str` be ?

lizmat: Aaah,... I see  :-)

lizmat: Alas:

    # You failed 112 tests of 229
    not ok 107 - .ACCEPTS
    # Failed test '.ACCEPTS'

I think this is caused by the fact that it changes the order in which the tests are performed.  When we have a `Str` candidate, it will be preferred above the `Any` candidate, and thus miss the `Numeric` check.  Will try to see how it works out with a `Numeric` candidate as well.

lizmat: Alas:

    Ambiguous call to 'ACCEPTS(IntStr: IntStr)'; these signatures all match:
      (IntStr:D: Numeric:D \a, *%_)
      (IntStr:D: Str:D \a, *%_)

lizmat: Result of having a Numeric candidate:

    Ambiguous call to 'ACCEPTS(IntStr: Str)'; these signatures all match:
      (Str:D: Str:D \other, *%_)
      (IntStr:D: Any:D \a, *%_)

So, I'm not sure how to fix this other with an `is default`

lizmat: 39bc2bdf28 replaces all `ACCEPTS` by a single candidate, but that still needs an `is default` to make tests pass

src/core.c/allomorphs.pm6
lizmat: hmmm... Maybe it should?

src/core.c/allomorphs.pm6
```diff
@@ -1,6 +1,14 @@
 my class Allomorph is Str {
     multi method Bool(::?CLASS:D:) { self.Numeric.Bool }
 
+    multi method ACCEPTS(Allomorph:D: Any:D \a) is default {
```
MasterDuke17, line 4: Why not multis for Numeric and Str?

lizmat: Because then we wind up with an ambiguous dispatch.

lizmat:
Unless someone comes out as against this, I will merge this PR in the coming days.

vrurg:
I'd join @jnthn suggestion of `Allomorph` being a role. A parametric one. Any particular reason we limit it to `Numeric` and `Str`? With strict binding to allomorphs been `Str` in first place? Some 3rd party module may use the concept to implement their own allomorphs which are non-numeric, non-str. After all, the term itself came to be initially from chemistry and then migrated into linguistic.

Something like this would look more natural to me:

```raku
role Allomorph[::T1, ::T2] { ... }
class StrNumeric does Allomorph[Str, Numeric] is Str { ... }
class IntStr is StrNumeric is Int { ... }
```

AlexDaniel:
This is not part of your PR, but something I noticed:
```raku
multi sub infix:<cmp>(Int:D \a, Rational:D \b) {
    a.isNaN || b.isNaN ?? a.Num cmp b.Num !! a <=> b
}
multi sub infix:<cmp>(Rational:D \a, Int:D \b) {
    a.isNaN || b.isNaN ?? a.Num cmp b.Num !! a <=> b
}
```
Am I just being dumb, or… how can an `Int` have an `.isNaN` method?

AlexDaniel:
Another interesting observation (not really important for this PR because this PR preserves the behavior):
```
<AlexDaniel> c: introducing-allomorph-class,HEAD say Duration.new(Inf) cmp Inf
<committable6> AlexDaniel, ¦introducing-allomorph-class,HEAD(09e4f23): «Less␤»
```
They should actually be Same, but the code seems to be checking against Inf/-Inf and assuming that the other value doesn't matter.
```
<AlexDaniel> c: introducing-allomorph-class,HEAD say Duration.new(Inf).Bridge cmp Inf.Bridge
<committable6> AlexDaniel, ¦introducing-allomorph-class,HEAD(09e4f23): «Same␤»
```

src/core.c/Order.pm6
```diff
@@ -28,12 +28,16 @@ multi sub infix:<cmp>(\a, Real:D \b) {
 }
 
 multi sub infix:<cmp>(Real:D \a, Real:D \b) {
-       (nqp::istype(a, Rational) && nqp::isfalse(a.denominator))
-    || (nqp::istype(b, Rational) && nqp::isfalse(b.denominator))
-    ?? a.Bridge cmp b.Bridge
-    !! a === -Inf || b === Inf
-        ?? Less
-        !! a === Inf || b === -Inf
+    nqp::istype(a,Rational) && nqp::istype(b,Rational)
```
AlexDaniel, line 31: I wish this had some comments for every branch, otherwise it's very hard to deduce which cases are being tested for. The type graph for Real is rather complicated so maybe the comments themselves could reveal some bugs.

AlexDaniel (approved):
I think this is an improvement to what we had before. We should be very careful with PRs like this (there are many edge cases that can be accidentally broken), but I don't see any changes in behavior. I think it's good to go.

As for the idea of turning it into a role, I don't have an opinion on that. Please judge yourself.

Issue #3748, opened 2020-06-08
any(0,4) != 0 is False


mscha:
## The Problem

```
> any(0,4) == 0
any(True, False)
> any(0,4) != 0;
False
```

## Expected Behavior

```
> any(0,4) == 0
any(True, False)
> any(0,4) != 0;
any(False, True)
```

## Actual Behavior

See above

## Steps to Reproduce

See above

## Environment
* Operating system: Linux
* Compiler version (`perl6 -v` or `raku -v`): Rakudo version 2020.05.1 built on MoarVM version 2020.05

coke:
One more note:
```
> any(0,4) == 0
any(True, False)
> so any(0,4) == 0
True
```
So I think we're first getting the Boolean, and then negating the boolean, not the junction.

pmichaud:
On Mon, Jun 08, 2020 at 08:28:26AM -0700, Will Coleda wrote:
> So I think we're first getting the Boolean, and then negating 
> the boolean, not the junction.

I recall that this is the specced behavior.  See the documentation for Junctions at https://docs.raku.org/type/Junction :

  Negated operators are special-cased when it comes to autothreading. 
  $a !op $b is rewritten internally as !($a op $b). The outer negation 
  collapses any junctions, so the return value always a plain Bool.

    my $word = 'yes';
    my @negations = <no none never>;
    if $word !eq any @negations {
        say '"yes" is not a negation';
    }

  Note that without this special-casing, an expression like $word ne 
  any @words would always evaluate to True for non-trivial lists on one side.

So, I suspect this ticket is actually "notabug".

Pm

AlexDaniel:
Documentation will not fix this. Not only it is special-cased, it is also fully unreadable. Just ban it if the code doesn't do what it says.

pmichaud:
I'm not claiming that the documentation is a fix; I'm simply saying that it was a deliberate design decision after a fair bit of discussion (i.e., not an arbitrary decision), and using the documentation as evidence of that.

There are tests in roast (autothreading.t) that enforce this behavior; the tests reference another issue ticket that describes more behind the issue.  https://github.com/Raku/old-issue-tracker/issues/1787

FCO:
> On 8 Jun 2020, at 16:54, Patrick R. Michaud <notifications@github.com> wrote:
> 
> 
> On Mon, Jun 08, 2020 at 08:28:26AM -0700, Will Coleda wrote:
[...]

PR #3874, opened 2020-08-24
Remove special handling of ne and != wrt Junctions

lizmat:
This is a proof of concept.  It breaks the following tests:

t/spec/S03-junctions/boolean-context.t                          (Wstat: 256 Tests: 74 Failed: 1)
  Failed test:  72
t/spec/S03-junctions/autothreading.t                            (Wstat: 768 Tests: 106 Failed: 3)
  Failed tests:  79-80, 85
t/spec/S03-operators/misc.rakudo.moar                           (Wstat: 256 Tests: 78 Failed: 1)
  Failed test:  74

t/04-nativecall/18-routine-sig-sanity.t                       (Wstat: 512 Tests: 35 Failed: 2)
  Failed tests:  3, 7

Should we decide to do this, we will probably need a version check for this.
So this will have to wait for the newdisp branch to be merged.

### DIFF
--- a/src/core.c/Numeric.pm6
+++ b/src/core.c/Numeric.pm6
@@ -308,9 +308,9 @@ multi sub infix:<=~=>(\a, \b, :$tolerance = $*TOLERANCE)    {
 # U+2245 APPROXIMATELY EQUAL TO
 my constant &infix:<≅> = &infix:<=~=>;
 
-proto sub infix:<!=>(Mu $?, Mu $?, *%) is pure  {*}
+proto sub infix:<!=>($?, $?, *%) is pure  {*}
 multi sub infix:<!=>($?)                    { Bool::True }
-multi sub infix:<!=>(Mu \a, Mu \b)          { not a == b }
+multi sub infix:<!=>(\a, \b)                { not a == b }
 # U+2260 NOT EQUAL TO
 my constant &infix:<≠> := &infix:<!=>;
 
--- a/src/core.c/Numeric.pm6
+++ b/src/core.c/Numeric.pm6
@@ -308,9 +308,13 @@ multi sub infix:<=~=>(\a, \b, :$tolerance = $*TOLERANCE)    {
 # U+2245 APPROXIMATELY EQUAL TO
 my constant &infix:<≅> = &infix:<=~=>;
 
-proto sub infix:<!=>($?, $?, *%) is pure  {*}
-multi sub infix:<!=>($?)                    { Bool::True }
-multi sub infix:<!=>(\a, \b)                { not a == b }
+proto sub infix:<!=>(Mu $?, Mu $?, *%) is pure  {*}
+multi sub infix:<!=>($?) { Bool::True }
+multi sub infix:<!=>(Mu \a, Mu \b) {
+    nqp::istype(($_ := a == b),Junction)
+      ?? $_ but .not
+      !! $_.not
+}
 # U+2260 NOT EQUAL TO
 my constant &infix:<≠> := &infix:<!=>;
 
--- a/src/core.c/Stringy.pm6
+++ b/src/core.c/Stringy.pm6
@@ -36,9 +36,8 @@ proto sub infix:<eq>($?, $?, *%)  is pure {*}
 multi sub infix:<eq>($x?)          { Bool::True }
 multi sub infix:<eq>(\a, \b)       { a.Stringy eq b.Stringy }
 
-proto sub infix:<ne>(Mu $?, Mu $?, *%) is pure {*}
+proto sub infix:<ne>($?, $?, *%) is pure {*}
 multi sub infix:<ne>($x?)            { Bool::True }
-multi sub infix:<ne>(Mu \a, Mu \b)   { a !eq b }
 multi sub infix:<ne>(Any \a, Any \b) { a.Stringy ne b.Stringy }
 
 proto sub infix:<lt>($?, $?, *%) is pure {*}
--- a/src/core.c/Stringy.pm6
+++ b/src/core.c/Stringy.pm6
@@ -36,9 +36,13 @@ proto sub infix:<eq>($?, $?, *%)  is pure {*}
 multi sub infix:<eq>($x?)          { Bool::True }
 multi sub infix:<eq>(\a, \b)       { a.Stringy eq b.Stringy }
 
-proto sub infix:<ne>($?, $?, *%) is pure {*}
-multi sub infix:<ne>($x?)            { Bool::True }
-multi sub infix:<ne>(Any \a, Any \b) { a.Stringy ne b.Stringy }
+proto sub infix:<ne>(Mu $?, Mu $?, *%) is pure {*}
+multi sub infix:<ne>($x?) { Bool::True }
+multi sub infix:<ne>(Mu \a, Mu \b) {
+    nqp::istype(($_ := a.Stringy eq b.Stringy),Junction)
+      ?? $_ but .not
+      !! .not
+}
 
 proto sub infix:<lt>($?, $?, *%) is pure {*}
 multi sub infix:<lt>($x?)          { Bool::True }


lizmat:
FWIW, this is the commit from 2011 that made `ne` to be the same as `!eq` wrt `Junction`s: 886f67743070f9d51bdd5

jnthn:
Given `ne` and `!=` are really just convenience forms of `!eq` and `!==` (part huffmanization, part cultural expectations), I don't see why we'd want to give them different semantics.

lizmat:
> Given `ne` and `!=` are really just convenience forms of !eq and !== (part huffmanization, part cultural expectations)

I'd say they're more cultural expectations than huffmanization, as prefixing any infix op with a `!` is a special thing of Raku, afaik.

So, I'd say that `ne` and `!=` are expected to work just as `eq` and `==` do, but not negated.  Rather than as the special case of `!` prefix of an infix op semantics, which semantically means to collapse any Junction.

     a == b
     a !== b is the same as !(a == b), so collapses
     a != b is culturally the same as a == b, but the opposite

It's hard to explain.

> I don't see why we'd want to give them different semantics

Well, that's just it.  I feel (and have felt, looking back at the clogs) that the current semantics are "different"  :-)

It takes a lot of work to grok how Junctions work.  The current behavior of `!=` and `ne` is a special case of that, for reasons I think are wrong.

We need to document the prefix ! to an infix to Boolify its result better.  And that `a !op b` is the same as `!(a op b)`, and that thus `!==` is **not** the same as `!=`.

jnthn:
> I'd say they're more cultural expectations than huffmanization, as prefixing any infix op with a ! is a special thing of Raku, afaik.

Yes, I think it's the bigger factor; I recall Larry noting (I'm heavily paraphrasing) that while regularity (everything via the `!` metaop) would in some ways be nice, forcing everyone to write `!eq` and `!==` would probably be a bit too counter-cultural. (I was gonna say that it would have saved us this discussion, but of course there'd have been 20 proposals to provide `!=` and `ne` by now if we'd gone that way... :-))

> So, I'd say that ne and != are expected to work just as eq and == do, but not giving the opposite result. Rather than as the special case of ! prefix of an infix op semantics

On "giving the opposite result" - I think I agree, but it leads me to the opposite conclusion too (more below). In terms of number of available semantics on offer, making `ne` and `!eq` different means that there's more to learn. It'd also mean we can't explain the `!` meta-op as a generalization of `!=` and `ne`; they become these negation-y things that stand alone from the general form of negation-y things. I don't dispute they make the operator implementations for `!=` and `ne` more regular, but I'm not sure that's the most pressing regularity to go after.

> which semantically means to collapse any Junction

But ultimately that's what Junctions are destined to do: collapse in a boolean context.

> The current behavior of != and ne is a special case of that, for reasons I think are wrong.

I've seen various explanations of why the "lift out the negation" semantics were chosen, but to me the biggest win is in the refactoring ergonomics. Being able to exchange `!eq` for `ne` is not all that persuasive, but if I have:

```
unless $food eq any <fish seafood> {
    eat $food;
}
```

And I decide I want an `else` branch, I can do the mechanical transformation of:

```
if $food ne any <fish seafood> {
    eat $food;
}
```

And given that's defined as:

```
if !($food eq any <fish seafood>) {
    eat $food;
}
```

I can be comfortable I've preserved the semantics. By contrast, I believe this PR would mean that:

```
if $food ne any <fish seafood> {
    eat $food;
}
```

Instead behaves like:

```
if any ($food ne 'fish'), ($food ne 'seafood') {
    eat $food;
}
```

Which is a result I'd find distasteful. :-)

> It takes a lot of work to grok how Junctions work

Not sure how far I agree, but either way I'd argue it's less work if one can perform common code transformations that normally work on `!=`/`eq`, and still have the right things happen even if the code involves junctions.

raiph:
Hopefully this comment is helpful. Apologies if it's not.

> It takes a lot of work to grok how Junctions work.

I 100% agree.

I'd say this is probably my top concern about junctions, along with the directly related issue of explaining it all.

> The current behavior of != and ne is a special case ... for reasons I think are wrong.

First a quick tangent. Presumably it's more than just `!=` and `ne`. What about `∉` and cousins?

My (dim) recollection of the reasoning for making these ops do what their aliases do was that doing so made it easy to explain their behaviour:

* Same as the operators they're aliases of

And this in turn meant it was easy to explain their behaviour *when used with junctions*:

* Same as the operators they're aliases of

* Which in turn means `a !op b` means `!(a op b)`

> It's hard to explain.

I'm not sure if you're saying what I just claimed was easy to explain is actually hard in your view, or if you're saying the way you're suggesting we switch to is hard to explain. But if it's the latter, my understanding is that the issue of explaining it is the primary reason they work the way they currently do. To return to my starting point:

> It takes a lot of work to grok how Junctions work.

So, all other things being equal, it's best if we avoid making things harder to explain.

The negations + junctions issue adds on top of other difficulties with junctions.

If we now add yet another layer of complexity to say that the ops `ne` and `!=`, which are already special cases in that they are convenience aliases for `!eq` and `!==`, are *doubly* special cased in that when they're used with Junctions, they *lose* their no longer act as the operators they are otherwise aliases of, well, why?

What is the *practical* benefit you're seeing that outweighs the additional complexity in explaining this metaspecial case in which their usual special casing is undone?

Or, conversely, if the benefit you see is that it's *easier to explain*, what is the full explanation of their operation that's easier to explain than the current:

* Same as the operators they're aliases of

?

jubilatious1:
Related to rakudo/rakudo#3748 ?

jnthn:
> Related to #3748 ?

Yes, indeed. About the linguistic argument sometimes mentioned - and referenced in that issue - it can be seen from my example above; we understand "I'll eat it unless it's fish or seafood" and "I'll eat it if it's not fish or seafood" to mean the same thing, and so it'd be quite surprising to be given a plate of octopus with the explanation "it's not fish". Trouble is that there seem to be cases that *don't* come out quite right under the linguistic argument too (generally, if the junction is on the left hand side of the comparison).

> What about ∉ and cousins?

I'd expect `∉` and similar to have the same junction semantics as `!=` and `ne`.

lizmat:
Regarding `∉` and cousins: QuantHashes are `.WHICH` based entities.  `Junction`s have a `.WHICH` that are not value types, just like `List`s, so they will only ever "match" themselves.

Going back to specific comments:

> we can't explain the ! meta-op as a generalization of != and ne; 

I don't see the `!` metaop of a generalization of `ne`.  To me, `ne` is an op just like `eq` or `gt` or `ge`.  It performs a logic action on its operands, and returns the result of that action.  And there is no visual cue for `!` in there anywhere.  And yes, logic shows that `ge` could be considered to be a `Rlt`, but that doesn't reduce the value of `ge` as an operator.  So in other words, I see `ne`, and by consequence `!=` as full operator citizens, rather than being derivatives of other operators.

> Which is a result I'd find distasteful. :-)

The example is correct.

> Not sure how far I agree, but either way I'd argue it's less work if one can perform common code transformations that normally work on !=/eq, and still have the right things happen even if the code involves junctions.

I guess the example is bad in the sense that a refactor of that should set off an alarm.  Worse case would of course be if it was *not* a visible junction.  But then again, you probably wouldn't have use such a construct.

The example is also bad in the sense that you could argue it is a poor use of `Junction`s.  A use of `Set`s would have been better here, I'd say:

    unless $food ∈ <fish seafood>.Set {
        eat $food
    }

can be easily and mechanically refactored to:

    if $food ∉ <fish seafood>.Set {
        eat $food
    }

But apart from the example's quality, I now *do* see the issue better with refactoring these **especially** when one of the operands is what I would call a "hidden" `Junction`.  Which got me close to closing the PR.

However, this gave me an idea: what if a `Junction` produced by `!=` or `ne` would have a (non-infective) flag, that would cause it to collapse into its opposite logical value.  That would remove the magical of `!=` and `ne`, and replace it by some other magic.

AlexDaniel:
Yes, this PR doesn't make it easier. Indeed, it adds a special case to what appears to be another special case. Again, the goal should be less special cases, not more. The point that was made [in this discussion](https://colabti.org/irclogger/irclogger_log/perl6-dev?date=2017-06-23#l333) is that any operator should work with junctions like any other. This way it is very easy to explain: any operator is applied to all junction values and you get a Junction back. That is, `==` is applied to all Junction values, and so should `!=`, and `!==`, and any other operator.

The way `!==` wraps negation over the whole junction result appears to be a special case. I can't really tell, maybe it's just an unfortunate implementation detail, but that justification doesn't make it easier for users. As a user, I see metaops as just ops. I can do any regular stuff with them. For example:

```raku
my $x = &[R!>]; say $x(42, 5) # True
```

Should I be thinking that these arguments were wrapped in some particular way? I'd say not. And then what about:
```raku
my $x = &[R>]; say $x(42, (5|6)) # any(False, False)
```
```raku
my $x = &[R!>]; say $x(42, (5|6)) # True
```
Well, if you're applying any given op (`$x`) to your arguments, one of which is a Junction, should you really expect it to do what it currently does?

As for code transformations, I think a few unrealistic assumptions are made. But first of all, the more or less bulletproof transformation for `unless` is `if not`. Flipping the op is of course also possible, but then it can get very confusing. Observe:
```raku
if ‘fish’ ne any <fish seafood> {
```
As I'm reading this, if ‘fish’ is not equal to any of these, the expression should be True. So in this case it will always be True? Not currently, but given what people are saying about negated ops with junctions, I think it's a very common interpretation. That is, this:
```raku
if ‘fish’ ne all <fish seafood> {
```
If ‘fish’ is not equal to all of these. But fish is equal to fish, so don't eat it. Makes sense, but it doesn't do that. There are simply two different interpretations.

Now, I think the wrong assumption is that you will be able to write code like this for actual projects. In reality, however, if good Raku::Critic existed, it would've banned the use of negated ops with junctions simply because the code is not explicit enough and it is *very common* for different people to read it differently. If this behavior is special-cased on purpose, then, unfortunately, it was added for code that will be discouraged. Anyway, the way to fix it is to remove that behavior for negated ops, *and not* to introduce more special cases for ops that just happen to *look* like normal non-negated ones. However, if the implementation has to be this way and there's no way to change it, then we will just have to leave it the way it is :(

Hope it makes sense.

All related discussions I could find:
* https://colabti.org/irclogger/irclogger_log/perl6?date=2016-08-21#l51
* https://colabti.org/irclogger/irclogger_log/perl6-dev?date=2017-06-23#l333
* https://github.com/rakudo/rakudo/issues/3748
* https://github.com/Raku/old-issue-tracker/issues/1787
* https://github.com/Raku/old-issue-tracker/issues/3100

lizmat:
@AlexDaniel 

> Yes, this PR doesn't make it easier. Indeed, it adds a special case to what appears to be another special case. Again, the goal should be less special cases, not more.

I **removed** the special casing of `!=` and `ne` in this PR.   How is that **more** special casing? *Confused*

jubilatious1:
> we understand "I'll eat it unless it's fish or seafood" and "I'll eat it if it's not fish or seafood" to mean the same thing...

Sometimes the language evolves to help dispel logical misconceptions. The "neither/nor" linguistic construct comes to mind:

_"I'll eat neither fish nor seafood."_

Actually, quite a famous quote in the US is built around the neither/nor construct: **_“Neither snow nor rain nor heat nor gloom of night stays these couriers from the swift completion of their appointed rounds”_**

https://uspsblog.com/how-is-new-york-city-related-to-famous-postal-quote/

[Now, how to write the above quote in Raku?]

lizmat:
The only failing tests are now the ones that specifically test on it not returning a `Junction`.  But the check for the resulting `Bool` value *passes*  :-)

jnthn:
> I don't see the ! metaop of a generalization of ne

While S03 doesn't precisely phrase it that way, it does say that [`!=` and `ne` are traditional shortcuts for the general form](https://design.raku.org/S03.html#Negated_relational_operators). The term "shortcut" hints at semantic equivalence, and confirms they were considered as related things during the initial language design. (It's conceivable in an alternative timeline where macros somehow featured early in the compiler implementation, they'd have been done as `macro infix:<!=>($a, $b) { quasi { $a ![&COMPILING:<< infix:<==> >>] $b } }` or something along those lines.)

> How is that more special casing?

Well, *if* one considers `!=` as a shortcut for `!==`, then it's making the shortcut a special case compared to the general negation form. This is the view I'd subscribe to. However, if one doesn't take that view, then the current behavior of `!=` and `ne` will look like the special case.

> The example is also bad in the sense that you could argue it is a poor use of Junctions. A use of Sets would have been better here

If all I'm interested in is a boolean result, I think reaching for Junctions is quite natural. It's when we want to know *which* things matched that a Junction is a poor choice.

lizmat:
> If all I'm interested in is a boolean result, I think reaching for Junctions is quite natural. It's when we want to know which things matched that a Junction is a poor choice.

I guess I have worked too much on `Set`s, but please note that `(elem)`, `∈`, `∉`, `(cont)`, `∋`, `∌` **all** return a Bool.  And as such are an excellent candidate for using a `Set`, I'd say.

In any case, I *think* my latest commit gives the best of both worlds: it returns a `Junction`, **but** one that will collapse in the way we currently expect.

AlexDaniel:
@lizmat fwiw, as far as I can see, in this snippet:

```raku
if $food ∉ <fish seafood>.Set {
```

`.Set` is not needed. This should work just as fine:
```raku
if $food ∉ <fish seafood> {
```
It even [seems to avoid](https://github.com/rakudo/rakudo/blob/cc19d3d85a788169379762a006dfaf64f2a1bfe7/src/core.c/set_elem.pm6#L51-L73) a creation of a Set object. So, yes, that particular example is even shorter with set ops. But that's just one bad example, sometimes you want to use a specific op so you kinda have to use junctions.

lizmat:
@jnthn not sure what your current idea is about this PR, and whether you've included https://github.com/rakudo/rakudo/pull/3874/commits/dc8c1f0f5ff29360fdbd8c08c4630d933e7e98d4 into your opinion

jnthn:
@lizmat I've pondered it some more over the days since the latest commit, but I think I can summarize my position as:

* I *do* see `!=` as a shortcut for `!==` and `ne` as a shortcut for `!eq`, and from that viewpoint any semantic difference between the shortcuts and their full forms feels undesirable. My pondering that maybe the shortcuts could be compiler special forms (or just standard library macros) that map into to `!==` and `!eq` wasn't entirely a throwaway one.
* Since the destiny of a Junction is to end up in a boolean context, collapsing a little earlier doesn't seem like a significant problem to me - and certainly not one that I'd throw something as costly as using `but` at. Did I miss something here?
* I agree some applications of negating operators with Junctions don't behave as a natural language reading might suggest, and that this is a source of confusion. However, it seems to me that the change proposed here would not solve that, but only change which cases are confusing - as well as have the negative refactorability effects as described earlier. Perhaps if there's a consistent pattern to the problematic cases (for example, where the Junction appears on the LHS of the negated operator) then it might be worth a `worry` (mention in "Potential Difficulties").

raiph:
A [tio program](https://tio.run/##dVLPbtowHL7nKT7CpIBUrKBKuzDoqmo3ph2KdurFSczq1djIduiyqXf2NHunvgi1nRQculqxlfj3/fMv3jItPh6GuBG8fIC9Z3j@t0dRW6skaKF2DFRWeKTcYq00pvnkModhpZKVIckQ36RowH5tNTOGK2mcBrVgOypqahmswkrXTkQzVNxsBW1Y5Xkr5VRYcCxVxQItM9C1vEAZwvggN77UpSmYUI@e@lVVfN10VGksldZ4I/dhtRKh0EukufM8vGLxWarabbdjjk85cuImy5Hl2WKWnIA7posIOJ9j4J85nvd7N//2wD@0qreTB@4a48FEKslAwkJl4xYhQG6ZBVlyY2PqB3FvJob/Zq3PdIYhptgw6lJWSmYWQTsi6B7hcpYk/ue0ByOl2hRcUusPPzpqjxcL8lNxOcoukI0xWbS2bS/@JMC7Cvp9Bd1T6DTarvmyf@sqXS1uUogQNo4YYNOEXO5Up664hk9xdRUHHgyQjuKN8atWOjtq/ddRnzu2nrr11G889bmnjj31uWerdrp@TjQNB2qb4dk9dGpog7s0Ytyl4Gt3aXsyKfny/XoZEwU11gP7kWPE8vp2hSwjzuG0/ZS8Eej3@cmNw@EF) that mass generates variations of `foo op bar` in a way that might be helpful to those exploring this issue.

lizmat:
FWIW, I'm still mulling this one over :-)

lizmat:
Still mulling :-)

lizmat:
Still mulling  :-)

lizmat:
Still mulling!

lizmat:
I give up

2colours:
https://github.com/Raku/problem-solving/issues/319
Apparently I managed to reiterate the very same issue without knowing.

I have to say that after reading this PR, it does seem to me this is indeed more Raku/problem-solving than rakudo/rakudo...

In that issue, I pinpointed something that got hinted here as well but maybe went a bit unnoticed: >= is NOT a negation of < but a proper "first-class" operator, same for gt and le. To say that the fact that I can express negation with the same junction type in one case but not in the other is "confusing" might even be a little euphemistic...

I think this is a whole separate issue from ne and != being equivalent to !eq and !==. I share lizmat's opinion that they shouldn't be, if this made it possible for them to become "first-class" operators like <= or le.

This is also a separate issue from accepting the behavior of the ! meta-operator for junctions _at all_, i.e whether the negation should go outside of the junction - keep in mind this isn't just a technical question of when to collapse junctions but also a semantical one since or beloved de Morgan equations apply when we take negations from one side of the parens to the other. :)

Of course these issues have certain consistency implications so they all need to be taken into account when defining the behavior - however, there can be more important points that can define the bigger picture while allowing multiple possible approaches that lead to that.

In my case, that would mean:
- lifting the negation from ne and != in particular, hence (in my opinion, arbitrarily) making them "fake operators" is a problem; I wouldn't be the first one to compare this design choice to sabotage. This is the main thing that, in my opinion, has to go.
- the behavior of the ! metaoperator can be argued for and against: if we take the current behavior for granted, consistency suggests that ne and != will need to break away from !eq and !== (which I would very welcome personally) - if the whole negation lifting doesn't happen (and I don't see why it _has to_), of course the whole "special treatment" discussion would be instantly over.

PR #3878, opened 2020-08-25
Add definiteness constraint to convenience methods

codesections:
The `set`/`unset` methods on `SetHash` and the `add`/`remove` methods on `BagHash` cannot be called on type objects; there's no meaningful way to add or remove an element from a non-definite item.  However, the methods were not previously constrained to require a definite invocant, which means that invoking the methods with a type object resulted in a slightly confusing error: «Cannot look up attributes in a $Type type object».

This commit adds the missing constraints, which improves the error message to: «Invocant of method '$method' must be an object instance of type '$Type', not a type object of type '$Type'.  Did you forget a '.new'?».

### DIFF
--- a/src/core.c/BagHash.pm6
+++ b/src/core.c/BagHash.pm6
@@ -312,7 +312,7 @@ my class BagHash does Baggy {
     }
 
 #--- convenience methods
-    method add(\to-add --> Nil) {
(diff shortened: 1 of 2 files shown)


lizmat (approved):
Good catch!

Issue #3882, opened 2020-08-26
T Lin_JVM build often hangs


Altai-man:
## The Problem

Our JVM build on CI often hangs forever, exceeding the CI timer and thus invalidates whole commit.

## Expected Behavior

This should not happen, CI should reliably pass if things are ok.

## Actual Behavior

Builds hang.

## Steps to Reproduce

I would:

1)Investigate logs of recent builds to see if there is a common point in all those hangs.
2)Create a branch with all possible logging enabled, enabled a test/wip PR and re-trigger it until hang happens and see if logging has anything interesting.

## Environment
* Operating system: CI env, JVM build.
* Compiler version (`perl6 -v` or `raku -v`): HEAD

usev6:
Now, that the CI builds for the JVM backend are successful again, I took a look at the logs of recent builds. I didn't see a hanging JVM build. So I think that the problem has gone away somehow over the last 1.5 years.

I'm closing this issue.

PR #3887, opened 2020-08-27
Add some missing files to the `make clean` target for the JVM backend

patrickbkr:
These files were left behind on `make clean` and `make realclean`.


patrickbkr:
The tests once again fell victim to #3882.

PR #3892, opened 2020-08-30
Sake to ake

Altai-man:
s/Sake/Ake/ to silence a warning, plus add a complete target without ssh business involved.


tools/releasable/Akefile
```diff
@@ -354,7 +364,7 @@ task ‘rakudo-manual-check’, { # Step ❶ (rakudo release guide)
     say ‘  + Review the RT queue’;
     say ‘  + Update the release announcement draft’;
     say ‘’;
-    say ‘Everything else will be done for you by this Sakefile’;
+    say ‘Everything else will be done for you by this Akefilefile’;
```
AlexDaniel, line 367: Akefilefile ?

AlexDaniel (changes requested):

AlexDaniel:
Makes sense to me, feel free to merge. FWIW, the idea for ssh target was to allow working on releases even if you have just slow laptop at hand. That is, it is meant to run tests on a faster machine while doing all other steps locally. Of course, this approach is not necessary if you have a fast enough computer, and especially so if flappers are fixed (because it currently reruns the whole test suite if it stumbles upon one). I'd say feel free to remove ssh stuff completely, but please judge yourself.

Altai-man:
>I'd say feel free to remove ssh stuff completely, but please judge yourself

I'll surely do. ;)
Need to remove some rsync mentions as well...

tools/releasable/Akefile
```diff
@@ -22,11 +22,6 @@ use File::Directory::Tree;
```
AlexDaniel, line 22: ↑ The table above still talks about “Local” and “Server” and I think it no longer makes sense.

AlexDaniel:
BTW it should've been possible to ssh to localhost too, I never tried it but I think it works.

Altai-man:
>BTW it should've been possible to ssh to localhost too, I never tried it but I think it works.

It is, but I don't have my ssh server enabled all the time and if we are tailoring the process...

I admire your efforts put into this and had some hard time thinking about doing this PR, but I don't plan to leave the position any time soon (unless people would chase me off with pitchforks and torches because I stopped posting email announcements for a couple of releases and now it's stupid to send them as a pack, stupid to skip them and pretend it did not happen, yet odd to make situation worse doing nothing about it :P ).

Hence we probably want to tailor the process to be _as dumb as possible_ and tie all the ties possible so that future releases won't be as, ahem, time consuming and annoying to the users we have as this month's ones were.

tools/releasable/Akefile
Altai-man: Hmmmm. It is pretty and I am not sure I want to remove it... Though yes, now it makes no sense.

AlexDaniel:
> I stopped posting email announcements for a couple of releases and now it's stupid to send them as a pack, stupid to skip them and pretend it did not happen, yet odd to make situation worse doing nothing about it :P

That's OK I guess, [they never really rendered](https://www.nntp.perl.org/group/perl.perl6.compiler/2019/11/msg18340.html) in the archive when I posted them :D IIRC unsigned emails render just fine, but the idea of not signing my release announcements sounded even worse to me.

> Hence we probably want to tailor the process to be as dumb as possible and tie all the ties possible so that future releases won't be as, ahem, time consuming and annoying to the users we have as this month's ones were.

Yes, simpler is better in this case. I'm not objecting, I'm agreeing :)

AlexDaniel (approved):

PR #3898, opened 2020-08-31
Remove the Cool.EVAL method

lizmat:
Just for reference and testing

### DIFF
--- a/src/core.c/Cool.pm6
+++ b/src/core.c/Cool.pm6
@@ -397,10 +397,6 @@ my class Cool { # declared in BOOTSTRAP
     method trim-leading () { self.Stringy.trim-leading  };
     method trim-trailing() { self.Stringy.trim-trailing };
 
-    method EVAL(*%opts) {
-        EVAL(self, context => CALLER::, |%opts);
-    }
-
     multi method Real() {
         nqp::if(
             nqp::istype((my $numeric := self.Numeric), Failure),


jnthn:
I'm curious what the fallout is. I suspect removing it we may not get away with, but making it macro-esque like `WHAT` and friends might do it (that is, it's a compiler rewrite into an `EVAL` sub call).

lizmat:
Hmmm... that shouldn't be too difficult to do, but how would we recognize the "core" `Cool.EVAL` method at compile time?  Or are you saying that *all* EVAL methods should "suffer" this fate.  In other words, don't call your methods with all capital letters either?  As they may have their semantics changed in future versions of Raku ?

jnthn:
> but how would we recognize the "core" Cool.EVAL method

There'd not be one; `$foo.EVAL` would compile into `EVAL($foo)`. Also worth nothing that `EVAL(...)` is realized as a sub, but that's a bit of a lie because the compiler recognizes it as a special form and knows now to optimize out lexicals when it sees it. One could argue that for 6.e *both* forms are rewrites (into something implementation defined).

lizmat:
So, assume I have a class:

    class Foo {
        method EVAL() { }
    }

then `Foo.new.EVAL` would be codegenned as `EVAL(Foo.new)` ?  So for that to actually work, one would have to add a `sub EVAL(Foo:D)` candidate and export that as well, for `Foo.new.EVAL` to work?

Or would the codegen be: `EVAL(Foo.new.Str)` ?

jnthn:
@lizmat You'd have to write `Foo.new."EVAL"()`, just as you do if you have a method called `WHAT`.

lizmat:
The spectest fallout is massive.  Closing this now, as the code change is trivial, so easily re-created if and when it is deemed necessary.

PR #3899, opened 2020-09-01
Add cmp candidate for Code

### LINKED ISSUE #3885: Warning on Hash[non-str, non-str].new( ... more than one pair ...)
## The Problem

$ `raku -e "(my Regex %t{Regex}=/'y'/,/1/, /'z'/,/0/).raku.say"`
Regex object coerced to string (please use .gist or .raku to do that)
  in block <unit> at -e line 1
Regex object coerced to string (please use .gist or .raku to do that)
  in block <unit> at -e line 1
(my Regex %{Regex} = /'y'/ => /1/, /'z'/ => /0/)
_why warnings there? ⬆︎ The warnings are the problem._

By contrast, this doesn't warn
$ `raku -e "(my Regex %t{Regex}=/'y'/,/1/).raku.say"`
(my Regex %{Regex} = /'y'/ => /1/)

## Expected Behavior
warning-free "raku" call:
$ `raku -e "(my Regex %t{Regex}=/'y'/,/1/, /'z'/,/0/).raku.say"`
(my Regex %{Regex} = /'y'/ => /1/, /'z'/ => /0/)

## Musing

I'd think that whatever the fix is, it should go in the Associative role, so that all derived classes will "freeze" warning-free.

## Environment
* Operating system: Mac OSX
* Compiler version (`perl6 -v` or `raku -v`): This is Rakudo version 2020.07 built on MoarVM version 2020.07
implementing Raku 6.d.

alabamenhu:
Fixes #3885 wherein gisting Hashes with non-Str keys could produce warnings.

The root of the problem is that `cmp` resorted to stringification, which generates warnings with `Code` objects (the only non-`Nil` types to do so).  Now `Code` objects will stringify using their `name`, which matches previous behavior but without the warnings.  Warnings elsewhere —where they are more useful— are thus preserved.

### DIFF
--- a/src/core.c/BagHash.pm6
+++ b/src/core.c/BagHash.pm6
@@ -312,7 +312,7 @@ my class BagHash does Baggy {
     }
 
 #--- convenience methods
-    method add(\to-add --> Nil) {
+    method add(BagHash:D: \to-add --> Nil) {
         nqp::bindattr(
           self,SetHash,'$!elems',nqp::create(Rakudo::Internals::IterationSet)
         ) unless $!elems;
@@ -321,7 +321,7 @@ my class BagHash does Baggy {
         );
     }
 
-    method remove(\to-remove --> Nil) {
+    method remove(BagHash:D: \to-remove --> Nil) {
         Rakudo::QuantHash.SUB-ITERATOR-FROM-BAG(
           $!elems, to-remove.iterator
         ) if $!elems;
--- a/src/core.c/SetHash.pm6
+++ b/src/core.c/SetHash.pm6
@@ -271,7 +271,7 @@ my class SetHash does Setty {
     }
 
 #--- convenience methods
-    method set(\to-set --> Nil) {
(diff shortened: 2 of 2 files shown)


src/core.c/Order.pm6
```diff
@@ -54,6 +54,16 @@ multi sub infix:<cmp>(int $a, int $b) {
     ORDER(nqp::cmp_i($a, $b))
 }
 
+multi sub infix:<cmp>(Code:D \a, Code:D \b) {
+    a.name cmp b.name
+}
+multi sub infix:<cmp>(Code:D \a, \b) {
+    a.name cmp b.Stringy
+}
+multi sub infix:<cmp>(Code:D \a, \b) {
```
niner, line 63: Isn't the Code:D type constraint on the wrong arg here?

alabamenhu: Yikes. Correct. I'll amend

PR #3902, opened 2020-09-02
Add cmp candidates for Code

### LINKED ISSUE #3885: Warning on Hash[non-str, non-str].new( ... more than one pair ...)

alabamenhu:
Fixes #3885 wherein gisting Hashes with non-Str keys could produce warnings.

The root of the problem is that `cmp` resorted to stringification, which generates warnings with `Code` objects (the only non-`Nil` types to do so).  Now `Code` objects will stringify using their `name`, which matches previous behavior but without the warnings.  Warnings elsewhere —where they are more useful— are thus preserved.

### DIFF
--- a/src/core.c/Order.pm6
+++ b/src/core.c/Order.pm6
@@ -54,6 +54,16 @@ multi sub infix:<cmp>(int $a, int $b) {
     ORDER(nqp::cmp_i($a, $b))
 }
 
+multi sub infix:<cmp>(Code:D \a, Code:D \b) {
+     a.name cmp b.name
+}
+multi sub infix:<cmp>(Code:D \a, \b) {
+     a.name cmp b.Stringy
+}
+multi sub infix:<cmp>(\a, Code:D \b) {
+     a.Stringy cmp b.name
+}
+
 multi sub infix:«<=>»(Int:D \a, Int:D \b) {
     ORDER(nqp::cmp_I(nqp::decont(a), nqp::decont(b)))
 }


lizmat:
I'm wondering whether we should just add a `quietly` to the gisting of Hashes.

alabamenhu:
> I'm wondering whether we should just add a `quietly` to the gisting of Hashes.

That works as a minimal fix for this one for sure (and allows `Nil` which does feel like it should emit a warning but YMMV). We'd also need to track down other places like in `Setty` that similarly uses sort for gisting, but if that's the preferred approach I can go track down the other places where it might happen and put in a `quietly`.

My thinking is that while a non-string key for a Hash might be rare, a mixed class array is more likely since it's of `Any` by default and if passsed to an external module that uses `cmp` or `sort`, the same issue will crop up

    sub external-sub-from-module (@a) { 
        for @a.sort { ... }
    }

    my @foo = '1', 2, <foo bar>, /foo/, { $^a + $^b};
    external-sub-from-module @foo;

The above user might try to wrap the external call in a `quietly`, but would then miss other warnings generated for non-stringification reasons.

niner (approved):
This improves not only gisting of hashes but all use cases where lists of routines get sorted (like `Any.^methods.list.sort.say`). Seems like a good thing to have.

PR #3903, opened 2020-09-03
Revise deprecation warnings to show only on v6.d

codesections:
My previous PR (#3900) inadvertently displayed the deprecation warning on v6.c and later; this fixes the code to only display the warning on v6.d.  Thanks to @timo for pointing this out to me on IRC.

I would have liked to put both the sub `parse-names` and the method `parse-names` in the new `core.d/Str.pm6` file for clearer
organization.  However, I could not see a way to correctly modify a `Str` method from inside that file.  If someone can point me in the direction of how to do that in a clean way, I'd be happy to revise this change.

### DIFF
--- a/src/core.c/Str.pm6
+++ b/src/core.c/Str.pm6
@@ -3230,8 +3230,10 @@ my class Str does Stringy { # declared in BOOTSTRAP
     }
 
     method parse-names(Str:D: --> Str:D) {
-        Rakudo::Deprecations.DEPRECATED('uniparse');
-        self.uniparse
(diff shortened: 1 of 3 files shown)


lizmat:
Since `parse-names` is a method, shouldn't we just be able to have a src/core.d/Str.pm6 that contains a `method parse-names` with the deprecation message, and a src/core.e/Str.pm6 that has a `method parse-names` that just dies with a `X::Method::NotFound` error?

lizmat:
Feels to me that this can be closed.

codesections:
Closing this as stale; I'll open a new PR if I come back to this issue.

PR #3905, opened 2020-09-03
Make IO::Notification::Change a value type

lizmat:
Which should make it better to filter for it, e.g. with
Supply.unique(expires => X).

### DIFF
--- a/src/core.c/IO/Notification.pm6
+++ b/src/core.c/IO/Notification.pm6
@@ -16,9 +16,7 @@ my class IO::Notification {
 
         multi method WHICH(Change:D: --> ValueObjAt:D) {
             nqp::box_s(
-              nqp::join('|',nqp::list_s(
-                self.^name,nqp::substr($!event.Str,4),$!path
-              )),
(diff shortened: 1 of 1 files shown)


src/core.c/IO/Notification.pm6
```diff
@@ -13,6 +13,15 @@ my class IO::Notification {
             "$.path: $.event";
         }
         method IO { $!path.IO }
+
+        multi method WHICH(Change:D: --> ValueObjAt:D) {
+            nqp::box_s(
+              nqp::join('|',nqp::list_s(
+                self.^name,nqp::substr($!event.Str,4),$!path
```
jnthn, line 20: Not sure the `substr` is a net win here?

jnthn (approved):
I don't see a reason why not to do this; it's immutable for sure.

src/core.c/IO/Notification.pm6
lizmat: I felt the "File" part of the enum was redundant.  Not a biggie, really.

PR #3907, opened 2020-09-07
Parallel module install

lizmat:
An attempt at parallelizing module installation, now that a blocking race condition has been fixed by nine.

### DIFF
--- a/src/core.c/CompUnit/Repository/Installation.pm6
+++ b/src/core.c/CompUnit/Repository/Installation.pm6
@@ -148,147 +148,191 @@ sub MAIN(:$name, :$auth, :$ver, *@, *%) {
     }
 
     method install(Distribution $distribution, Bool :$force) {
-        my $dist  = CompUnit::Repository::Distribution.new($distribution);
+        my $dist := CompUnit::Repository::Distribution.new($distribution);
         my %files = $dist.meta<files>.grep(*.defined).map: -> $link {
-            $link ~~ Str ?? ($link => $link) !! ($link.keys[0] => $link.values[0])
+            $link ~~ Str
+              ?? ($link => $link)
+              !! ($link.keys.head => $link.values.head)
         }
 
-        $!lock.protect( {
-        my @*MODULES;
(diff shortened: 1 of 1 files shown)


src/core.c/CompUnit/Repository/Installation.pm6
```diff
@@ -148,147 +148,200 @@ sub MAIN(:$name, :$auth, :$ver, *@, *%) {
...
+                        note "(Already doing $id)" if $verbose;
+                        next;
+                    }
+
+                    my $source      := $sources-dir.add($id);
+                    my $source-file := $repo-prefix
+                      ?? $repo-prefix ~ $source.relative($.prefix)
+                      !! $source;
+
+                    @promises.push: start {
```
ugexe, line 321: Spawning a process for every module listed under provides is going to reach diminishing returns pretty quickly.

niner:
> Spawning a process for every module listed under provides is going to reach
> diminishing returns pretty quickly.

The problem is that we don't know the dependencies between the modules in a 
dist. Ordinarily you'd walk the tree bottom up and keep a couple processes 
busy that way, but we don't know how the nodes connect to a tree. Doesn't help 
if you spawn X processes but they all are waiting for the same dependency, 
while there may be other leaf modules available.

So the next step seems to be a large and hard one involving those precomp 
processes communicating their status (compiling/waiting for another precomp) 
to the main process so the latter can throttle accordingly. And even then you 
need even more communication as finishing one dependency may suddenly unblock 
a whole horde of precomp processes.

vrurg:
I did not follow the whole story of parallelizing precomps and may miss a thing. But how hard would it be to make it a task queue based processing? So that limited number of worker threads are pre-spawned and each would pick a task from a channel?

ugexe:
> I did not follow the whole story of parallelizing precomps and may miss a thing. But how hard would it be to make it a task queue based processing? So that limited number of worker threads are pre-spawned and each would pick a task from a channel?

Precompilation currently spawns a new process and thus requires IPC

lizmat:
Closing this now, as it appears we should be able to solve slow installation better with making sure we only precomp a module once in the installation process.

PR #3908, opened 2020-09-12
Decamelify easter eggs

lizmat:
Remove easter egg regarding camels.

And no, this is *not* a request to remove all easter eggs from Rakudo.  This is about removing references to Perl.


japhb:
This particular reference feels sufficiently oblique to be "a humorous reference that some will realize refers to shared history" rather than "clearly a Perl reference at face value".  Also it can be argued that camel and beer are jokes as much about the overall silliness of Unicode (or even of nature itself) as they are about the Perl family of languages per se.

I'd rather keep the oblique camel humor; it makes me smile.  :-)

MattOates:
Why not 🦋 instead?

lizmat:
@MattOates for this to work, we need **2** codepoints, one with one butterfly, and one with two butterflies.

dumarchie:
It's "Camelia, the Raku bug", so we could use 🐛 and 🦋. That would be `.base("raku")` or `.base("bug")`.

dumarchie:
Or we could go for base 3, with 🥚, 🐛 and 🦋. The egg resembles a zero, so it actually makes more sense than using one hump to represent zero and two humps to represent one. There appears to be no symbol for the pupa life stage, but base 4 is square anyway.

japhb:
That almost makes me want to do one with dimensions: ‧‭, |‭, □‭, 🎲‭ or somesuch -- that gives a base-4 system at least.  Too bad I couldn't find a codepoint for tesseract or light cone to go for base-5.  :-)

lizmat:
This PR was about removing the camel easter egg.  Not about adding others.  Since the general feeling is that they should stay, I will close this PR.  Should you feel inclined to add other easter eggs for `.succ` and `.pred`, please do them in a separate PR.  Thank you!

PR #3913, opened 2020-09-14
Fix `--rakudo-home` Configure.pl parameter

### LINKED ISSUE #3537: libperl6_ops_moar.so installed in wrong directory
Hello

While packaging rakudo for Debian, lintian issue the following error:
```
E: rakudo: arch-dependent-file-in-usr-share usr/share/perl6/runtime/dynext/libperl6_ops_moar.so
N: 
N:    This package installs an ELF binary in the /usr/share hierarchy, which
N:    is reserved for architecture-independent files.
```
`libperl6_ops_moar.so` should be installed in `/usr/lib` directory. 

With `moarvm`, `Configure.pl` provides a  `--libdir` option to specify where to install libraries

Could you provide a similar option for rakudo's Configure.pl ?

All the best

## Environment 
* Operating system:  Debian Linux (unstable)
* Compiler version (`perl6 -v`): 2020.02

patrickbkr:
`Perl6::SysConfig` object construction was broken in so far as the
%rakudo-build-config hash was initialized externally after object
construction and thus wasn't available in the `BUILD()` method itself but
was used there. This is fixed by passing a pre-populated
%rakudo-build-config hash to the Perl6::SysConfig constructor.

Fixes #3537


src/Perl6/SysConfig.nqp
```diff
@@ -2,8 +2,14 @@ class Perl6::SysConfig is HLL::SysConfig {
...
-    method BUILD() {
-        %!rakudo-build-config := nqp::hash();
+    method new(:%rakudo-build-config) {
+        my $obj := nqp::create(self);
+        $obj.BUILD(:%rakudo-build-config);
+        $obj
+    }
+
+    method BUILD(:%rakudo-build-config) {
+        %!rakudo-build-config := %rakudo-build-config;
```
patrickbkr, line 12: Is this a valid and good way to construct a subclass with a named parameter in NQP?
A small test shows, that the superclass (HLL::SysConfig) is actually called, but I don't know how.

patrickbkr: Update: It seems, that the super class constructor is called, but the subclass then starts with a blank superclass. So this actually doesn't work.

vrurg:
I'd say there is no need to use the named. Just make it a positional. Besides, even though it doesn't really matter here, but nameds come with some performance price.

patrickbkr:
@vrurg I now changed it to a positional parameter. But superclass construction is still broken. Any idea how a working constructor can look?

vrurg:
I don't have much time on my hands right now. The thing about `HLL::SysConfig` `BUILD` being invoked makes me confused. NQP doesn't support inheritance in full. I.e., there is no way for the parent method to be called if redefined by a child. So, you'd have to setup the attributes in `Perl6::SysConfig` `BUILD`. But why do you see the parent method invocation? No idea, it must not happen.

Otherwise, it looks like you need a workaround to utilize parent class initialization code. Can't think at the moment of any other class doing similar thing. But basically you'd need a setup method on `HLL::SysConfig` which would be invoked by `Perl6::SysConfig`.

patrickbkr:
I now changed it, so that the `HLL::SysConfig` `BUILD` logic is in a separate method and call it from `Perl6::SysConfig`. Now the super class attributes in `Perl6::SysConfig` are initialized correctly. But the `note()` statements in the code make it pretty clear, that the NQP `BUILD()` method is somehow called automatically as well.

So I still need an expert to clarify how to correctly write a super/sub class relationship in NQP, where both super and sub class have initialization code.

vrurg:
By rejecting any possible act of magic, I'd say that `HLL::Sysconfig.BUILD` is most definitely not called when you create `Perl6::SysConfig`. Just to make sure I make no mistake:

```
class Foo {
    method new() { nqp::create(self).BUILD() }
    method BUILD() { note("Foo!"); self }
}
class Bar is Foo {
    method BUILD() { note("Bar!"); self }
}

note("-- Foo");
Foo.new;
note("-- Bar");
Bar.new;
```

Therefore, the only logical conclusion would be to assume that somewhere `HLL::SysConfig` object is created too. Grepping give me the most likely suspect: `nqp/src/HLL/Compiler.nqp`. I was never getting really deep into the way NQP modules are involved into Raku compiler bootstrapping, but one thing I'm sure about is that since NQP modules are getting loaded their mainline code is executed. Consequently, the line `nqp::bindhllsym('default', 'SysConfig', HLL::SysConfig.new());` from `nqp/src/HLL/Compiler.nqp` does get ran. This is where you get the false impression of parent `BUILD` being called. It is, in fact, for for a totally different object.

This actually makes me wonder if the compiler and the sysconfig initialization code at the bottom of `Compiler.nqp` belongs there. To my understanding, it is purely NQP specific and should only be executed when we compile NQP.

patrickbkr:
@vrurg That was a very good call. You were right about the logic being called via a different path. I moved the logic in NQP from `src/HLL/Compiler.nqp` (which is subclassed in `src/Perl6/Compiler.pm6`, which is `use`d in `src/main.nqp`, which initializes `Perl6::SysConfig`) to `src/NQP/Compiler.nqp`. Now the duplicate construction of `HLL::SysConfig` is gone.

PR #3915, opened 2020-09-15
Add CAVEATS.md file

vrurg:
Move all platform-specific notes into it.

### DIFF
--- a/CAVEATS.md
+++ b/CAVEATS.md
@@ -4,39 +4,39 @@
 
 ### Spectest requirements
 
-You need recent version of either Strawberry Perl or ActiveState Perl.
+You need a recent version of either Strawberry Perl or ActiveState Perl.
 
-If you are working with ActiveState Perl you need the Mingw gcc compiler.
+If you are working with ActiveState Perl you will need the Mingw gcc
+compiler.
 
-You need msys git installed and you need "\Program Files\Git\cmd" on your
-execution path and NOT "\Program Files\Git\bin".
+You need msys git installed and "\Program Files\Git\cmd" on your execution
+path and NOT "\Program Files\Git\bin".
 
-You need a win32 curl program.
+You will also need a win32 curl program.
 
 
 ## macOS
 
 ### Dynamic libraries
 
-If Rakudo is installed into $HOME directory (this is also the case when
-rakubrew <https://rakubrew.org/> is used) then some modules may fail to
-install due to missing dynamic libraries. This could be caused by a
-situation when the library is not provided by macOS system itself but
-installed using Homebrew or MacPorts package managers. In this case the
-library cannot be loaded by rakudo binary. This is a security precaution
-taken by modern versions of macOS. The precaution works by restricting the
-directories available for loading dynamic libraries from only to the
-system-predefined set and those where the application binary is actually
-located. For example, if "rakudo" binary is installed under "$HOME/raku/bin"
-then aside of the system-wide locations it can access libraries in
-"$HOME/raku/lib" only.
+If Rakudo is installed into the $HOME directory (this is also the case with
+rakubrew <https://rakubrew.org/>) then some modules may fail to install due
+to missing dynamic libraries. This could be caused by a situation in which
+the library is not provided by macOS system itself but installed using
+Homebrew or MacPorts package managers. In this case the library cannot be
+loaded by "rakudo" binary. This is a security precaution taken by modern
+versions of macOS. The precaution works by restricting the directories
+available for loading dynamic libraries only to the system-predefined set
+and those where the application binary is actually located. For example, if
+"rakudo" binary is installed under "$HOME/raku/bin" then aside of the
+system-wide locations it can access libraries in "$HOME/raku/lib" only.
 
 For the above stated reason, if a problem with availability of a dynamic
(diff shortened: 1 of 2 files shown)


coke:
"doesn't exists" should be "doesn't exist"

CAVEATS.md
```diff
@@ -0,0 +1,52 @@
...
+### Spectest requirements
+
+You need recent version of either Strawberry Perl or ActiveState Perl.
+
+If you are working with ActiveState Perl you need the Mingw gcc compiler.
+
+You need msys git installed and you need "\Program Files\Git\cmd" on your
+execution path and NOT "\Program Files\Git\bin".
+
+You need a win32 curl program.
```
JJ, line 14: *will also* need

CAVEATS.md
```diff
@@ -0,0 +1,52 @@
...
+
+
+## macOS
+
+### Dynamic libraries
+
+If Rakudo is installed into $HOME directory (this is also the case when
```
JJ, line 21: *the* $HOME directory

JJ: *with* rakubrew (not when)

CAVEATS.md
```diff
@@ -0,0 +1,52 @@
...
+rakubrew <https://rakubrew.org/> is used) then some modules may fail to
+install due to missing dynamic libraries. This could be caused by a
+situation when the library is not provided by macOS system itself but
```
JJ, line 24: *in which* the library

CAVEATS.md
```diff
@@ -0,0 +1,52 @@
...
+installed using Homebrew or MacPorts package managers. In this case the
+library cannot be loaded by rakudo binary. This is a security precaution
```
JJ, line 26: *the* rakudo binary. Also, will you capitalize Rakudo?

CAVEATS.md
```diff
@@ -0,0 +1,52 @@
...
+taken by modern versions of macOS. The precaution works by restricting the
+directories available for loading dynamic libraries from only to the
```
JJ, line 28: restricting ... *to*

CAVEATS.md
```diff
@@ -0,0 +1,52 @@
...
+
+- <rakudo binary location path>/../lib
+- $HOME/lib (should be created if doesn't exists yet)
+
+The latter is one of the allowed system locations.
+
+Also, sometimes the library version provided by macOS itself is older than a
+module expects. This could result in erratic behaviors of the module. In
+such a case a solution would be to install a package with newer version of
+the library and also create symlinks into one of the two locations.
```
JJ, line 46: Maybe add an example?

INSTALL.md
```diff
@@ -159,13 +159,6 @@ If you want to run the tests in parallel, you need to install a fairly
...
 
-You need recent version of either Strawberry Perl or ActiveState Perl.
-
-If you are working with ActiveState Perl you need the Mingw gcc compiler.
-
-You need msys git installed and you need "\Program Files\Git\cmd" on your
-execution path and NOT "\Program Files\Git\bin".
-
-You need a win32 curl program.
+- CAVEATS.md
```
JJ, line 164: Link to the file.

JJ (approved):
Mostly OK, maybe check out my comments.

INSTALL.md
vrurg: In which form? Full github URL in angle brackets following the file name?

CAVEATS.md
vrurg: The example would depend on macOS version. Besides, I can only think of libssl issues @coke had. But there is no full certainty that it was the case.

CAVEATS.md
vrurg: It is a file name here.

vrurg:
@JJ Thanks! Will make corrections and merge later today.

CAVEATS.md
```diff
@@ -0,0 +1,52 @@
+# Platform specific notes
+
+## Windows
+
+### Spectest requirements
+
+You need recent version of either Strawberry Perl or ActiveState Perl.
```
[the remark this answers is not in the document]
vrurg: How effective is it to move some old pieces of text from one location to another! We need to do it time after time to correct otherwise overlooked errors! :)

CAVEATS.md
```diff
@@ -0,0 +1,52 @@
+# Platform specific notes
+
+## Windows
+
+### Spectest requirements
+
+You need a recent version of either Strawberry Perl or ActiveState Perl.
+
+If you are working with ActiveState Perl you will need the Mingw gcc
+compiler.
```
jnthn, line 10: This is a curiosity; I've tended to use ActiveState Perl + MSVC and it worked fine (still do that on the occasions I do a Windows build). Is the point that you'll need a C compiler (whereas Strawberry Perl bundles one), and it's being maybe more specific than it need be in saying which one?

coke: Guessing this came from the install text. I also use msvc- we can update the text to refer to both

vrurg: @coke guessed it right.

I only do a few builds on Window once in a long while. So, it's for somebody with better expertise to fix this paragraph.

PR #3921, opened 2020-09-17
De-huffmanize Capture.raku

lizmat:
The \\(...) syntax is confusing to a lot of newbies, especially coming
from Perl.  In order to de-emphasize that syntax, change the .raku
output to use the .Capture coercion syntax, which I think makes it
much more clear what is meant.

This will require *one* change in spectest, that specifically tests
for the \\(...) format in output:

    t/spec/S12-introspection/walk.t   test 23

Deprecating / removing the \\(...) syntax would be something for
multiple language versions.

### DIFF
--- a/src/core.c/Capture.pm6
+++ b/src/core.c/Capture.pm6
@@ -136,7 +136,7 @@ my class Capture { # declared in BOOTSTRAP
         my int $has-hash  = nqp::isconcrete(%!hash) && nqp::elems(%!hash);
         my Mu  $raku     := nqp::list_s();
         if nqp::eqaddr(self.WHAT, Capture) {
-            nqp::push_s($raku, '\(');
+            nqp::push_s($raku, '(');
             if $has-list {
                 my Mu $iter := nqp::iterator(@!list);
                 nqp::push_s($raku, nqp::unbox_s(nqp::shift($iter).raku));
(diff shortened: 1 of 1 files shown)


jnthn (changes requested):
> In order to de-emphasize that syntax, change the .raku
> output to use the .Capture coercion syntax, which I think makes it
> much more clear what is meant.

They're not equivalent, though. Capture literals parse an argument list inside of them and apply argument list rules around named arguments. A normal list does not retain this distinction. Thus:

```
$ raku -e 'given \((a => 1), b => 2) { dd .list; dd .hash }'
(:a(1),)
Map.new((:b(2)))
$ raku -e 'given ((a => 1), b => 2).Capture { dd .list; dd .hash }'
()
Map.new((:a(1),:b(2)))
```

I notice that the `.raku` method on `Capture` actually fails to roundtrip though: any Pair that is in the list part must be wrapped in parentheses. So this is wrong:

```
$ raku -e 'say \((a => 1), b => 2).raku'
\(:a(1), :b(2))
```

It should produce `\((:a(1)), :b(2))`.

I guess one could also `.raku` it as `Capture.new(list => (...), hash => Map.new(...))`.

> Deprecating / removing the \(...) syntax would be something for
> multiple language versions.

And replacing it with what? The boilerplate of `Capture.new` won't feel like an improvement to those using capture literals for argument matching in `Test::Mock`...

lizmat:
Hadn't realized that `\( )` actually has a different grammatical property.  So closing this PR

PR #3923, opened 2020-09-18
Precalculate prime numbers in first 100K

lizmat:
- create a constant that has a bitmap of the first 100K primes
- check that bitmap if request number < 100000
- use normal nqp::isprime_I for any other positive values
- add tests to make sure the bitmap is correct
- does *not* change the repeat facter: that feels like another issue


lizmat:
Makes `my @a = ^100000 .grep: *.is-prime` about 23x as fast.

jnthn (approved):
Classic space/time trade-off, but I guess the ~12KB extra this costs is a drop in the ocean, and lazy deserialization on MoarVM hopefully keeps the cost of this down for any programs never calling `is-prime` (bytecode files are memory mapped shared and thus demand paged). Glad to see the test.

colomon:
I'm trying to understand this patch.  The bitmap isn't actually "the first 100K primes" but rather "the primality of numbers 0-99,999", right? But you only consult the bitmap if the number is in the range 2-99,999?

If that's right, naively it seems like only using the bitmap for odd numbers in the range 3-99,999 would allow you to only use half as much bitmap and double the speed of shifting to find the bit you are interested in?  Apologies if you've already tried this and found your way faster, I'm not sure how good my intuition is for speed of operations on bignums / efficiency of NQP operations.  (And a second thought ... would it make sense to move this logic into nqp::isprime_I?)

lizmat:
@colomon : @nwc10 already nudged me for that, c824c44 implements that idea, making it more than 2x as fast again.

lizmat:
@colomon re moving this logic into `nqp::isprime_I` ?   Probably, but that would be more up `nwc10`'s ally.  Disadvantage would be that we would not have a sanity check anymore, or we would have to adapt the sanity check.

lizmat:
c824c44 also addressed the test-failure, because the script did not emit a plan.

colomon:
I was imagining logic more like

If even
    Return self == 2
Else
    Bitmap or is prime call

lizmat:
@colomon: taken you advice in 9ab69d5, makes `my @a = ^100000 .grep: *.is-prime` a few % faster still.

MasterDuke17:
>     * does _not_ change the repeat facter: that feels like another issue

Based on the SO answer @timo linked (https://security.stackexchange.com/questions/4544/how-many-iterations-of-rabin-miller-should-be-used-to-generate-cryptographic-saf), why not just hard-code the value to 40 at whatever the lowest level possible is (e.g., `MVM_bigint_is_prime` https://github.com/MoarVM/MoarVM/blob/master/src/math/bigintops.c#L1324) and not even make it an option to `nqp::isprime_I`?

lizmat:
The Travis build is failing for different reasons.  I suggest we drop Travis or fix whatever it is that is making it fail.  In any case, I cannot test if it is a transient issue or not, as I apparently do not have rights to request a re-run  :-(

JJ:
it's failed to pull the source, and just in one of the many tests. I would
just ignore it.

lizmat:
Thing is, with a more strict policy having to have a green badge to allow a PR to be merged, this PR could now **NOT** be merged.  And this is what we're striving for, as far as I know?

AlexDaniel (approved):
I'm not sure how I feel about this PR.

So, the motivation for this was probably this discussion (and my recent mention of it): https://news.ycombinator.com/item?id=21638145

In case of Julia, `Primes` is not core functionality, but a separate package. The implementation is here: https://github.com/JuliaMath/Primes.jl/blob/65668ef30cd9bab3120ba136139503bf77e964c3/src/Primes.jl#L134-L165 (but keep in mind that `witnesses` are defined above).

Raku, on the other hand, calls into the VM and asks for the result. In MoarVM libtommath is used: https://github.com/libtom/libtommath/blob/6e779e69a89c828a02b20323c6c7b494750e8dcf/mp_prime_is_prime.c#L16

All this is making me wonder so many things. For example, is calling into the VM *so expensive* that lookup tables need to be implemented in Rakudo? Does it mean than *any* other function will benefit from lookup tables?

Or, if using a lookup table is so much faster in any language, then why others are not choosing the same approach? For example, why libtommath doesn't have that optimization?

What kind of functionality are you optimizing for, exactly? If people need a fast way to generate sequential primes, then maybe we need a function for that instead? But if people need to test a lot of randomly sized numbers, then aren't we making it slower for them?

But probably my biggest concern is the never-ending *addition* of *new* code in order to cover up other issues. Less code is almost always better. For example, this proposed `is-prime` implementation shadows the underlying `nqp::isprime`. Given that roast tests are essentially end-to-end (and given that MoarVM itself has no tests), how can we now know that MoarVM's handling of primes is correct?

So, yes, it is faster, but is it better? And if it is worse, then why are we making it faster? People on hackernews will find other benchmarks that run slow.

And if this PR is really justified, which I don't think it is, that large constant is a bit of a nuisance. How about generating it at build time?

AlexDaniel (changes requested):
Pressed the wrong button in the previous review :)

AlexDaniel:
This got me curious, how much time does it take to generate primes for the first 100K numbers?

```
julia> @time using Primes
  0.045792 seconds (32.27 k allocations: 2.854 MiB)  # ← just loading the external module

julia> @time primes(2, 100_000)[end]                 # ← printing the last one for convenience, to avoid filling the terminal
  0.126466 seconds (100.53 k allocations: 5.231 MiB) # ← as I understand, this includes compilation
99991

# now let's try to generate primes for 10 million (100 times more than our first timing)

julia> @time primes(2, 10_000_000)[end]
  0.027343 seconds (7 allocations: 7.630 MiB)
9999991

# actually, now that it is compiled, we can try 100K again:

julia> @time primes(2, 100_000)[end]
  0.000264 seconds (7 allocations: 101.547 KiB)
99991
```

So if compiled code can generate the same amount of primes in 0.000264 seconds, then I'm not surprised that nobody is doing it. Saving way less than a millisecond in exchange for ≈12KB is not worth it.

Now I'm convinced that this PR is attempting to optimize things at the wrong place. While rakudo is doing `my @a = ^100000 .grep: *.is-prime`, Julia can generate all primes for 100K numbers more than 2500 times.

The question here is how can we get Raku code to execute fast. A lookup table just for the prime numbers is like sweeping actual performance issues under the rug. Why the existing code is so slow? Isn't it just calling into the VM?

MasterDuke17:
> This got me curious, how much time does it take to generate primes for the first 100K numbers?

I don't think this is an exact comparison. The Julia code is *generating* all the primes up to a certain value, for which there is a fast algorithm. libtommath has an `mp_prime_next_prime` function, but we don't use it in MoarVM. @lizmat's example is *finding* all the primes in a list (which just happens to be a range). I don't know Julia, what are the timings for using the `Primes.isprime` function on all the elements in the range?

The paper @timo mentioned in chat (http://ceur-ws.org/Vol-1326/020-Forisek.pdf) is pretty recent (2015) and I don't see any reference to it in libtommath's `mp_prime_is_prime`. They do have a 256 element table they check against (they are pretty careful about code size since libtommath has a bunch of use on small embedded devices). I may experiment with implementing it for MoarVM and/or libtommath.

MasterDuke17:
> I don't know Julia, what are the timings for using the `Primes.isprime` function on all the elements in the range?

I got curious and tried it myself (I think this is the correct Julia code):
```julia
julia> using Primes

julia> @time filter(x -> isprime(x), 0:100_000)[end]
  0.126064 seconds (214.14 k allocations: 11.066 MiB)
99991
julia> @time filter(x -> isprime(x), 0:100_000)[end]
  0.033867 seconds (40.80 k allocations: 2.295 MiB)
99991
```
Compared with this Raku code:
```perl6
say (^100_000).grep(*.is-prime).tail;
say now - INIT now
```
With Rakudo on master, I get ~12.8s. With Rakudo on first-100000-primes I get ~0.16s. If I change the `nqp::isprime_I` op to not take a rounds option and then simplify the relevant MoarVM function to just hardcode the value to 40, Rakudo on master takes ~5.3s.

MasterDuke17:
I just looked the Julia source. It uses gmp (which is I think known to be faster than libtommath) for its BigInt version, and implements the paper previously referenced for native-sized integers.

AlexDaniel:
> I don't think this is an exact comparison.

It is not, because it isn't a comparison. I was thinking of clarifying that but then went to bed :) When I was first looking at this PR I got a false impression that the hardcoded constant provides a shortcut for work that would be otherwise hard to calculate. However, the lookup table is worth only 0.000264s on my laptop from 2013. And for the meaningless benchmark (`my @a = ^100000 .grep: *.is-prime`) there is plenty of time to generate it.

See, you can spend these 0.000264s on MoarVM's side the first time isprime is called. @jnthn mentioned lazy deserialization and somehow I have a feeling that generating the table is comparable to that in terms of speed, if not even faster. But that won't make the benchmark any faster. Or… will it?

Hold on… Wait a minute…

How fast is libtommath itself?

```c
#include <tommath.h>

#include <stdio.h>

mp_bool result;
mp_int number;

int main() {
    if (mp_init(&number) != MP_OKAY)
        printf("whoopsie!");

    for (size_t i = 0; i < 100000; i++) {
        if (mp_add_d(&number, 1uL, &number) != MP_OKAY)
            printf("whoopsie!");

        if (mp_prime_is_prime(&number, 1, &result) != MP_OKAY)
            printf("whoopsie!");

        if (result)
            printf("%u\n", mp_get_u32(&number));
    }
    return 0;
}
```

Not the most beautiful or correct code, I know, but it times at 20s (and that's with -O3). 😱 

Turns out we're using **too many rounds**!

If we change the number of rounds from 100 to 1 it can then run in 0.9s on my computer. That is still very slow, but the thing is that it works for numbers *way higher* than 100_000!

> I got curious and tried it myself (I think this is the correct Julia code):

It is, but I think it stutters a bit when compiling the provided lambda, or something like that. Try using the `isprime` function directly:

```
@time filter(isprime, 0:100_000)[end]
  0.009602 seconds (16.23 k allocations: 1.084 MiB)
99991
```
(this code also runs fast on the first run)

I think you can also define the function earlier.

Now, in some sense comparing to Julia is not really useful. That being said, why is libtommath so slow? I mean, answering my own questions from earlier, it seems like both Rakudo and MoarVM are not terribly slow in this particular case, it's the libtommath itself slowing things down.


----

So what does it all mean in terms of this PR? Well, first of all, we should reduce the number of rounds from 100 to maybe 1. Of course, it first needs to check if the number is actually small enough for a single round to suffice, and if not it's OK to fall back to 100. Some testing is probably required to figure this all out.

Then, even if the number of rounds is 1, why is it still so slow? How is it possible for Julia to be around 100 times faster than libtommath, even when a single round is used?

There's also this interesting comment: https://github.com/libtom/libtommath/blob/c18817cc229f9c60a0a70b76cf49077e91983f41/mp_prime_is_prime.c#L94-L98. Maybe it's relevant.

lizmat:
With all the discussions and research, it has become clear to me that this PR is not viable.  Which is fine by me!  As long as we get `is-prime` faster in the end.

I therefore have marked this PR as a draft now, so we can continue the discussion here for now.  Should the discussion move to somewhere else, this PR can be closed.

colomon:
I can't tell if @AlexDaniel 's last comment is meant to be a joke or not?  There is a very good chance 100 rounds is too many -- I have a memory that our initial implementation was very paranoid because the primality test seemed like magic.  (My memory is we combined this with a list of small primes for speed.)  On the other hand, unless there's something going on I don't understand at all, 1 round is absolutely not enough.  The excellent paper @MasterDuke17 posted from @timo talks about very specific cases where using the right numbers means 1 round is acceptable.

AlexDaniel:
@colomon I'm not joking. What's the first number when there's a difference between 1 round and 100?

colomon:
I have no idea what the first number it will be wrong on is -- indeed, since there is a random element, I'm not even sure it's possible to know.

In general if you only run one round, you will falsely identify composite numbers as prime 25% of the time.

colomon:
While https://en.wikipedia.org/wiki/Miller%E2%80%93Rabin_primality_test doesn't explicitly say it, it does present the opposite -- for instance, if n < 1,373,653, then you can test it with two specific rounds (a=2 and a=3) and always be  correct. That in turn suggests you will start seeing errors before that number if you only run one round.

japhb:
We don't have to pick a single value -- we can do a quick test for rough size range, and from there pick the number of rounds and witnesses to be either perfect (if small enough) or better than cosmic ray risk (if bigger than that).

To pick the bit length to switch to different `t`, I think this is one of those places that we can take advantage of already having MoarVM's "small big integers" at a known size ... we can determine the optimal tests to give the right answer for every "small" integer, and then fall back to a more conservative value (but yes, perhaps less than `t=100`) for true big integers.

(FWIW, it looks like the Julia code linked above is doing approximately this, with a few special choices of witnesses as well.)

AlexDaniel:
@colomon I have tested numbers up to 52_200_000 and found no differences between 1 and 30 rounds. If there's a random element that can skew the results in practice then 2 rounds is also fine-ish performance wise.

> We don't have to pick a single value -- we can do a quick test for rough size range, and from there pick the number of rounds and witnesses to be either perfect (if small enough) or better than cosmic ray risk (if bigger than that).

Yes, that was the point.

> (FWIW, it looks like the Julia code linked above is doing approximately this, with a few special choices of witnesses as well.)

Yes, but choosing a different amount of rounds won't exactly make it fast enough in our case. Even with 1 round it is still too slow. So, yes, it is a good start, but in the end something else should be figured out.

MasterDuke17:
I just added the FJ32_256 implementation supplied by the authors as a branch in `MVM_bigint_is_prime` if it's not actually big. Now on Rakudo master
```perl6
say (^100_000).grep(*.is-prime).tail;
say now - INIT now
```
takes 0.036s (and Rakudo passes `make m-test m-spectest`).

MasterDuke17:
> That being said, why is libtommath so slow?

I haven't been able to find exhaustive benchmarks of GMP vs libtommath, but there are some random comments with relevant info. Here are a couple factors:
- libtommath is exclusively C, no assembly (and I believe intends to stay that way)
- GMP is 10 years older than libtommath
- GMP is designed with speed as the primary goal (e.g., hand-optimized assembly for many different CPUs), for libtommath education is the primary goal (e.g., lots of comments, the documentation can be compiled into a book)
- I believe GMP implements more algorithms than libtommath (though they might all be the same for primality testing, I haven't checked)
- Anecdotal, but several of the comments I saw online said things like being within a factor of 3 of GMP is a fantastic achievement

As to why MoarVM chose libtommath in the beginning, I'm not sure, probably @jnthn would know. Some projects won't use GMP because it's dual-licensed under LGPL v3 and GPL v2, but libtommath is under the Unlicense. The API seems similar, but not identical, so switching wouldn't be a drop-in replacement.

MasterDuke17:
See https://github.com/MoarVM/MoarVM/pull/1352 for the MoarVM changes mentioned above.

AlexDaniel:
@MasterDuke17 fantastic work, thank you very much!

lizmat:
Closing as this approach is now dead  :-)

PR #3930, opened 2020-09-21
Remove unnecessary branch in .truncate

MasterDuke17:
This fixes the difference between `True.truncate` (which returns True)
and `False.truncate` (which returned 0) so they both return self.

### DIFF
--- a/src/core.c/Real.pm6
+++ b/src/core.c/Real.pm6
@@ -58,7 +58,7 @@ my role Real does Numeric {
     proto method exp(|) {*}
     multi method exp(Real:D: )           { self.Bridge.exp               }
     method truncate(Real:D:) {
-        self == 0 ?? 0 !! self < 0  ?? self.ceiling !! self.floor
+        self < 0  ?? self.ceiling !! self.floor
     }
     method isNaN { Bool::False }
 


lizmat:
@moritz what was your reasoning for the `== 0` branch in 1123ec5cec72bb2b364c23 ?

moritz:
It's too long for me to remember, but I can imagine two possible reasons:

* `.truncate` on any zero always returns an Integer 0 (and not, for example, a Rat)
* To avoid a method call in the zero case.

If it's not desirable, I'm fine with removing it.

lizmat:
Perhaps adding a `multi method truncate(Bool:) { self }` would be more appropriate, taking the "avoiding method call in case of 0" argument?

AlexDaniel (approved):

AlexDaniel:
> Perhaps adding a `multi method truncate(Bool:) { self }` would be more appropriate, taking the "avoiding method call in case of 0" argument?

@lizmat can we please start thinking in terms of edge cases and code complexity? As a rule of thumb, we should be making the logic simpler. Proof: `say Same.truncate; say More.truncate`.

MasterDuke17:
>     * To avoid a method call in the zero case.

FWIW, `sub a($a) { $a == 0 ?? 0 !! $a < 0  ?? $a.ceiling !! $a.floor }; my $b; $b = a(False) for ^1_000_000; say now - INIT now; say $b` takes ~0.2s, but `sub a($a) { $a < 0  ?? $a.ceiling !! $a.floor }; my $b; $b = a(False) for ^1_000_000; say now - INIT now; say $b` takes ~0.095s.

lizmat:
@AlexDaniel can *you* start thinking on replying in a less abrasive way?

Alternate reaction:

@lizmat Have you considered other edge cases and code complexity? As a rule of thumb, we should be making the logic simpler. Proof: `say Same.truncate; say More.truncate`

AlexDaniel:
> Alternate reaction:

Yes, I agree that it sounds better. In my view the issue I mentioned is not isolated to this pull request, and it is also not about any single developer. There is more frustration than abrasiveness, but you're right that I am not good at making that clear.

lizmat (approved):

PR #3938, opened 2020-09-30
Binary release pipeline fixes

patrickbkr:
- Fixes the pipeline on Windows
- Doesn't run spec test. The tests seem to hang or take a very long time on the CI.

### DIFF
--- a/tools/build/binary-release/build-windows.ps1
+++ b/tools/build/binary-release/build-windows.ps1
@@ -45,7 +45,7 @@ nmake install
 CheckLastExitCode
 
 echo "========= Testing Rakudo"
-prove -e ..\install\bin\perl6 -vlr t
+prove -e install\bin\raku -vlr t
 CheckLastExitCode
 
 echo "========= Cloning Zef"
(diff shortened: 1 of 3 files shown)


tools/build/binary-release/build-windows.ps1
```diff
@@ -45,7 +45,8 @@ nmake install
 CheckLastExitCode
 
 echo "========= Testing Rakudo"
-prove -e ..\install\bin\perl6 -vlr t
+rm -r t\spec
+prove -e install\bin\raku.exe -vlr t
```
lizmat, line 49: Also changing `perl6` to `raku.exe` here.  Two questions:

- is this intentional?
- why the `.exe` extension now, and not before on `perl6` ?

patrickbkr: The change to `raku` was intentional. I don't think we'll use the release pipeline for older releases not providing a `raku` executable, so it seems safe to make the move.

I'm not entirely sure anymore, but I vaguely recall `prove` not finding the executable without the `.exe` suffix. As to how it worked before - I can't tell. I'll give it a quick try what happens without the `.exe`.

patrickbkr: @lizmat The `.exe` is actually not needed. I removed it.

PR #3939, opened 2020-09-30
Build: A wrong NQP version should abort the build

patrickbkr:
Irrespective of using the `--with-nqp` configure flag a wrong NQP version is always fatal.
@vrurg The original code looked to be on purpose. That's why I'd like your reassurance that there is actually no reason to allow building with a wrong NQP version.

### DIFF
--- a/tools/lib/NQP/Config/Rakudo.pm
+++ b/tools/lib/NQP/Config/Rakudo.pm
@@ -663,7 +663,7 @@ sub gen_nqp {
     return unless defined($gen_nqp) || defined($gen_moar);
 
     {
-        my $user = $options->{'github-user'} // 'perl6';
+        my $user = $options->{'github-user'} // 'Raku';
 
         # Don't expect any specific default commit in nqp/ if the repo is
         # already checked out and we're force-rebuilding.
--- a/tools/lib/NQP/Config/Rakudo.pm
+++ b/tools/lib/NQP/Config/Rakudo.pm
@@ -643,18 +643,15 @@ sub gen_nqp {
 
         unless ( $force_rebuild
             || !$nqp_have
-            || $nqp_ver_ok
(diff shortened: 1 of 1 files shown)


vrurg:
I'm on vacation until Monday. Keeping my eye on notifications, but that's all I can until then.

vrurg (changes requested):
As long as I remember the purpose of using a warning is to make bisection of a regression in NQP possible when the bug is only revealing itself in a particular version of Rakudo. I relied on this for at least one time.

Overall, I think it's a good idea to fail upon incorrect NQP version which would prevent a number of bug reports from newbies. But the change must be accompanied with an option to turn the failure into a warning again would it ever be required. This way we'd achieve more user friendliness without prohibiting more sophisticated use cases.

tools/lib/NQP/Config/Rakudo.pm
```diff
@@ -665,7 +662,7 @@ sub gen_nqp {
 
     return unless defined($gen_nqp) || defined($gen_moar);
 
-    if ( defined $gen_nqp || defined $gen_moar ) {
+    {
         my $user = $options->{'github-user'} // 'perl6';
```
vrurg, line 666: BTW, isn't it time to replace _'perl6'_ with _'Raku'_?

patrickbkr: Done.

patrickbkr:
@vrurg I now made it listen to the `--ignore-errors` flag and depending on that either `note` or fail. Is that what you had in mind?

vrurg (approved):

vrurg:
Hm, I was certain I have replied and approved recently.

I thought about this to be a special case requiring a specialized option. But `ignore-errors` would probably be ok too.

PR #3940, opened 2020-09-30
Force "id" command to use POSIX locale

### LINKED ISSUE #3927: $*USER and $*GROUP gives Nil output with different locales
## The Problem
$*USER and $*GROUP return Nil when using non-English locale (zh_TW.UTF-8 and maybe others). This looks to be caused by the `id` command being used in Process.pm6. The `id` output on my machine, which uses the zh_TW.UTF-8 locale, produces the output:

    使用者id=1000(rypervenche) id群組=1000(rypervenche) 組=1000(rypervenche),10(wheel),18(audio),27(video)

The problem code can be found [here](https://github.com/rakudo/rakudo/blob/master/src/core.c/Process.pm6#L81).
## Expected Behavior
$*USER and $*GROUP should output actual username and group.
## Actual Behavior

$*USER and $*GROUP return Nil.

## Steps to Reproduce

1. Generate zh_TW.UTF-8 locale (varies by distro)
2. `LANG="zh_TW.UTF-8" raku -e 'say $*USER; say $*GROUP'`

## Environment
* Operating system: Gentoo Linux Testing
This is Rakudo version 2020.08.2 built on MoarVM version 2020.08
implementing Raku 6.d.

rypervenche:
This is a temporary workaround to ensure that $*USER and $*GROUP don't
return Nil on non-English locales.

Resolves #3927.

### DIFF
--- a/src/core.c/Process.pm6
+++ b/src/core.c/Process.pm6
@@ -77,7 +77,7 @@ Rakudo::Internals.REGISTER-DYNAMIC: '$*HOME', {
 
 {
     sub fetch($what) {
-        once if !Rakudo::Internals.IS-WIN && try { qx/id/ } -> $id {
+        once if !Rakudo::Internals.IS-WIN && try { qx/LC_MESSAGES=POSIX id/ } -> $id {
             if $id ~~ m/^
               [ uid "=" $<uid>=(\d+) ]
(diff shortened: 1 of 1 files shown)


niner (approved):

lizmat:
Thank you!

PR #3942, opened 2020-10-01
Update tests for Unicode 13.0, 13.1

thundergnat:
Modify testing scripts to deal with new Unicode 13.0, 13.1 data. Include Unicode 13.0 database files for testing.

MoarVM Pull request https://github.com/MoarVM/MoarVM/pull/1359 must be applied and Moar version bumped before applying this.

When that and this are done, https://github.com/Raku/roast/pull/676 can be applied.

This **WILL FAIL** CI until MoarVM 1359 is applied.


Altai-man:
Unicode versions are backward compatible? As in, won't we break expectations of e.g. 12.1 moving to 13?

thundergnat:
They should be backward compatible for the most part. Each version builds on the one before. There can be a few thing that get modified between version, not just added, but it is mostly error corrections, (minimal) removal of deprecated glyphs, alias updates, that kind of thing. By far the largest change is added codepoints & emoji and rule updates.

> won't we break expectations of e.g. 12.1 moving to 13?

Yes. Which is why I have the note in the pull request to not merge until MoarVM is updated to 13. There is a bunch of testing specifically for new content and rules introduced in 13 that will fail on version 12.1 because it doesn't know about them. 

That being said, 13 is not wildly different than 12.1. Almost all of the differences are new content. There are a few rule changes regarding grapheme composition of multiple combining characters, but they for they most part cover obscure combinations that weren't well defined before or regularize them to match how new combinations are defined.

samcv:
> There are a few rule changes regarding grapheme composition of multiple combining characters
@thundergnat 
Can you give me a link to where it mentions these changes? We can continue the discussion on graphemes at https://github.com/MoarVM/MoarVM/pull/1359 

I'm not concerned about back compatibility for the 12->13 unicode version jump. Though it is an open question how Raku wants to handle changing unicode versions, which will of course result in different behavior. For now we have accepted that there may be minor changes, but that is an acceptable tradeoff compared to keeping our unicode version out of date. In an ideal world you could choose which unicode version you want to use, but I don't think I, or other developers would want the fun of maintaining something like that.

PR #3943, opened 2020-10-04
Move Int candidates from Numeric.pm6 to Int.pm6

MasterDuke17:


lizmat:
I assume this is just because it makes more sense to have them there ?

lizmat (approved):

MasterDuke17:
> I assume this is just because it makes more sense to have them there ?

Yep.

PR #3944, opened 2020-10-04
Add Junction.collapse method

lizmat:
Inspired by Wim Vanderbauwhede's The strange case of the greedy junction

  https://wimvanderbauwhede.github.io/articles/greedy-junctions/

The implementation is rather different from the one proposed, but this
should perform a *lot* better.

### DIFF
--- a/src/core.c/Junction.pm6
+++ b/src/core.c/Junction.pm6
@@ -378,6 +378,22 @@ my class Junction { # declared in BOOTSTRAP
         # If we get here, wasn't actually anything to autothread.
         call(|args);
     }
+
+    method collapse(Junction:D:) {
+        if nqp::isgt_i(nqp::elems($!eigenstates),1) {
+            my \type := nqp::atpos($!eigenstates,0).WHAT;
+            my int $i = 0;
+            nqp::while(
+              nqp::islt_i(($i = nqp::add_i($i,1)),nqp::elems($!eigenstates)),
+              nqp::unless(
+                nqp::eqaddr(nqp::atpos($!eigenstates,$i).WHAT,type),
+                return Failure.new("Can't collapse Junction with {type.^name}")
+              )
+            );
+        }
+
+        $!eigenstates.List
+    }
 }
 
 proto sub any(|) is pure {*}


lizmat:
test failures are unrelated

AlexDaniel (changes requested):
It seems to me that the intention behind junctions was that their use would be transparent and you wouldn't need a way to inspect the values, so no way was added originally. This PR adds the ability to extract the values, somehow with a check that all values are of the same type.

Many questions. Why would they need to be of the same type? I don't understand. What does it “collapse”? At first glance the provided gist actually collapses the values, but this PR simply exposes the internal values of a junction.

Moreover, this functionality (whether it collapses or not) adds too much depth on the user's side, possibly introducing more bugs in code that is potentially already buggy due to the use of junctions. The user would need to handle cases of 1 value, more than one value, 0 values, an exception because of differing types (or values), and a yet another edge case of not even getting a junction back (therefore not getting something that has a .collapse method).

This PR as is needs to be modified, but I'd suggest to drop the whole idea of exposing junction values to users.

lizmat:
Did you read Wim Vanderbauwhede's blog post?

AlexDaniel:
Yes.

lizmat:
Then, how do you feel about: https://docs.raku.org/language/faq#index-entry-Junction_(FAQ)

Perhaps we should just make `eigenstates` a public attribute?   It feels strange to me that we on the one hand don't allow access, and on the other hand provide a very badly performing workaround.

AlexDaniel:
@lizmat the FAQ entry does everything it possibly can to discourage it and explain that you're not meant to do that. “If you want to do it anyway, you can *abuse* autothreading for that” – just because you can doesn't mean you should. This PR actually adds something people shouldn't do.

lizmat:
The same discouragement could be moved from the FAQ to the documentation of an `eigenstates` method, and would be more easily found that way.  Clearly, people are asking about this, otherwise it wouldn't be a FAQ.

AlexDaniel:
Please don't add functionality that is inherently discouraged. To me it reads like the provided solution in FAQ is there just to satisfy the curiosity of the users.

vendethiel:
`use MONKEY-SEE-NO-COLLAPSE;`

taboege:
If it is merged, is this only going to be a Rakudo feature or is it planned to become part of the Raku specification?

I always assumed that the insistence that the eigenstates cannot be queried is because implementations should not be restricted to implement junctions as a list of eigenstates. My naïve and layman imagination after watching that one talk by Conway was that junctions would become more than a list with a collapsing mode on Raku compilers targeting quantum computers. But then again, maybe the list-of-eigenstates representation is already effectively required when junctions are so closely associated with "autothreading" behavior.

Anyway, I think my first question still makes sense and is of some mid-term relevance.

lizmat:
Closing this now.  People interested in getting to the eigenstates of an object can use this: https://modules.raku.org/dist/eigenstates

PR #3947, opened 2020-10-07
Make "is rw" on parameters of MAIN worry

lizmat:
As noted in R#3929 by Zoffix++

### DIFF
--- a/src/Perl6/Actions.nqp
+++ b/src/Perl6/Actions.nqp
@@ -4025,9 +4025,18 @@ class Perl6::Actions is HLL::Actions does STDActions {
             ));
         }
 
-        # Set name.
+        # Set name if we have one
         if $<deflongname> {
-            $block.name(~$<deflongname>.ast);
+            my $name := ~$<deflongname>.ast;
+            $block.name($name);
+
+            # Check for 'is rw' parameters if MAIN
+            if $name eq 'MAIN' {
+                for $signature.params.FLATTENABLE_LIST -> $param {
+                    $/.worry("'is rw' on parameters of 'sub MAIN' usually cannot be satisfied.\nDid you mean 'is copy'?")
+                      if $param.rw;
+                }
+            }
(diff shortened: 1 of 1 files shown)


lizmat:
Associated tests: https://github.com/Raku/roast/commit/0ebacd297c

AlexDaniel (changes requested):
This PR assumes that user's don't name their scoped subs as MAIN. Once that's fixed I think it'll be good to go.

See https://gist.github.com/325171801278f97cab41b221a3e28059

lizmat:
AFAIK, there's no way at that stage to find out whether the MAIN sub has a scope that would make it run when called as a script.  That's why I made it a `worry` rather than a `sorry`.  To get rid of the worry, one would have to give the sub a different name.  Which, given the special meaning of MAIN, would be a *good* thing.

AlexDaniel:
If that's a good thing then also giving a worry on any sub called “MAIN” would be more consistent and predictable.

AlexDaniel:
BTW, an interesting bug:
```bash
perl6 -e 'my &MAIN = -> $x { say $x }'
```
```
No such method 'cando' for invocant of type 'Callable'
  in block <unit> at -e line 1
```

lizmat:
>> there's no way at that stage to find out whether the MAIN sub has a scope that would make it run when called as a script

> giving a worry on any sub called “MAIN” would be more consistent and predictable.

So, you want *ANY* script to have this worry??

AlexDaniel:
> So, you want _ANY_ script to have this worry??

I meant any sub that won't be used for command line stuff. Or are you saying that this is not possible to detect either?

lizmat:
> are you saying that this is not possible to detect either?

Yes, at the moment we do this test, I don't think we have enough info to tell whether that MAIN sub will be called as a CLI handler.  At least with my understanding of Actions.nqp  :-)

PR #3948, opened 2020-10-07
sub MAIN: Negation and explicit values with bundling and single param

patrickbkr:
When using the %*SUB-MAIN-OPTS<bundling> option negation and explicit
values are ambiguous when actually bundling multiple params together.

    -abc=foo # a = True, b = True, c = "foo"? a = b = c = "foo"?

When only a single param is given though, there is not ambiguity.

    -a=foo # a = "foo"

This change makes the latter possible.

@Leont I can't select you as a reviewer, but can you also have a look?

### DIFF
--- a/src/core.c/Main.pm6
+++ b/src/core.c/Main.pm6
@@ -80,7 +80,7 @@ my sub RUN-MAIN(&main, $mainline, :$in-as-argsfiles) {
                 # explicit value
                 if nqp::isgt_i(nqp::elems($split),1) {
                     my str $name = nqp::shift($split);
-                    die "Can't combine bundling with explicit arguments" if $bundling && nqp::iseq_s($0.Str, '-');
+                    die "Can't combine bundling with explicit arguments" if $bundling && nqp::iseq_s($0.Str, '-') && $name.chars > 1;
                     %named.push: $name => $1.chars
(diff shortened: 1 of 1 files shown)


src/core.c/Main.pm6
```diff
@@ -89,10 +89,10 @@ my sub RUN-MAIN(&main, $mainline, :$in-as-argsfiles) {
                 # implicit value
                 else {
                     if $bundling && nqp::iseq_s($0.Str, '-') {
-                        die "Can't combine bundling with explicit negation" if $1.chars;
+                        die "Can't combine bundling with explicit negation" if $1.chars && $arg.chars > 1;
```
Leont, line 92: Wouldn't that make more sense as part of the preceding `if` condition?

patrickbkr: @Leont I don't think so. If I would add that to the enclosing `if`, it would fall back to interpreting the parameter as a non-bundled parameter as soon as more than one letter is passed even when it started with a single "-". So `-/abc` would be interpreted as `abc = False`. I'm pretty sure that's not what we want.

Leont (approved):

lizmat (approved):

Issue #3945, opened 2020-10-06
Intersection meta-ops behave like union


nwellnhof:
say 1..3, 1..3 Z∩ 2..4, 2..4 # OUTPUT: (set(1 2 3 4) set(1 2 3 4))
    say zip with=>*∩*, (1..3, 1..3), (2..4, 2..4) # OUTPUT: (set(2 3) set(2 3))

There are similar issues with X meta-ops. Also, baggy multiplication ⊍ meta-ops behave like baggy addition.

lizmat:
Found the reason for the observed behaviour.  The code:

    1..3, 1..3 Z(&) 2..4, 2..4;

winds up in the **single** argument `(&)` candidate, which coerces its parameter to a `Set`.  So you basically get:

    (1..3, 2..4).Set

And it doesn't pass the `zip` operator logic at all, so that's why you *do* see the correct behaviour with `zip`.  I assume something similar is happening with the `X` metaops.

lizmat:
Thanks for the report, this should be in the 2020.10 release of Rakudo.

PR #3950, opened 2020-10-08
Fix for issue with Z / X with set operators

lizmat:
Make some set ops prec 'non' (structural)
    
This fixes the problem described in https://github.com/rakudo/rakudo/issues/3945 and is spectest clean.
    
The problem was caused by Z(&) basically codegenning to:

   infix:<(&)>( (left,right) )

for each entry on either side of the meta-op, instead of:

   infix:<(&)>( left, right )

The former just creating a Set of all elements in left/right (because it was using the single arg candidate), rather than actually applying the op to the left/right side.
    
This most likely also fixes the problem for other meta-ops, such as X.

### DIFF
--- a/src/Perl6/Actions.nqp
+++ b/src/Perl6/Actions.nqp
@@ -4025,18 +4025,9 @@ class Perl6::Actions is HLL::Actions does STDActions {
             ));
         }
 
-        # Set name if we have one
+        # Set name.
         if $<deflongname> {
-            my $name := ~$<deflongname>.ast;
-            $block.name($name);
-
-            # Check for 'is rw' parameters if MAIN
-            if $name eq 'MAIN' {
-                for $signature.params.FLATTENABLE_LIST -> $param {
-                    $/.worry("'is rw' on parameters of 'sub MAIN' usually cannot be satisfied.\nDid you mean 'is copy'?")
-                      if $param.rw;
-                }
-            }
+            $block.name(~$<deflongname>.ast);
         }
 
         # Finish code object, associating it with the routine body.
--- a/src/Perl6/Actions.nqp
+++ b/src/Perl6/Actions.nqp
@@ -4025,9 +4025,18 @@ class Perl6::Actions is HLL::Actions does STDActions {
             ));
         }
 
-        # Set name.
+        # Set name if we have one
         if $<deflongname> {
-            $block.name(~$<deflongname>.ast);
+            my $name := ~$<deflongname>.ast;
+            $block.name($name);
+
+            # Check for 'is rw' parameters if MAIN
+            if $name eq 'MAIN' {
+                for $signature.params.FLATTENABLE_LIST -> $param {
+                    $/.worry("'is rw' on parameters of 'sub MAIN' usually cannot be satisfied.\nDid you mean 'is copy'?")
+                      if $param.rw;
+                }
+            }
         }
 
         # Finish code object, associating it with the routine body.
--- a/src/Perl6/Grammar.nqp
+++ b/src/Perl6/Grammar.nqp
@@ -4139,21 +4139,21 @@ grammar Perl6::Grammar is HLL::Grammar does STD {
     token infix:sym<o>   { <sym>  <O(|%concatenation)> }
 
     token infix:sym<&>   { <sym> <O(|%junctive_and, :iffy(1))> }
-    token infix:sym<(&)> { <sym> <O(|%junctive_and)> }
-    token infix:sym«∩»   { <sym> <O(|%junctive_and)> }
-    token infix:sym<(.)> { <sym> <O(|%junctive_and)> }
-    token infix:sym«⊍»   { <sym> <O(|%junctive_and)> }
+    token infix:sym<(&)> { <sym> <O(|%additive)> }
+    token infix:sym«∩»   { <sym> <O(|%additive)> }
+    token infix:sym<(.)> { <sym> <O(|%multiplicative)> }
+    token infix:sym«⊍»   { <sym> <O(|%multiplicative)> }
 
     token infix:sym<|>    { <sym> <O(|%junctive_or, :iffy(1))> }
     token infix:sym<^>    { <sym> <O(|%junctive_or, :iffy(1))> }
-    token infix:sym<(|)>  { <sym> <O(|%junctive_or)> }
-    token infix:sym«∪»    { <sym> <O(|%junctive_or)> }
-    token infix:sym<(^)>  { <sym> <O(|%junctive_or)> }
-    token infix:sym«⊖»    { <sym> <O(|%junctive_or)> }
-    token infix:sym<(+)>  { <sym> <O(|%junctive_or)> }
-    token infix:sym«⊎»    { <sym> <O(|%junctive_or)> }
-    token infix:sym<(-)>  { <sym> <O(|%junctive_or)> }
-    token infix:sym«∖»    { <sym> <O(|%junctive_or)> }
+    token infix:sym<(|)>  { <sym> <O(|%additive)> }
+    token infix:sym«∪»    { <sym> <O(|%additive)> }
+    token infix:sym<(^)>  { <sym> <O(|%additive)> }
+    token infix:sym«⊖»    { <sym> <O(|%additive)> }
+    token infix:sym<(+)>  { <sym> <O(|%additive)> }
+    token infix:sym«⊎»    { <sym> <O(|%additive)> }
+    token infix:sym<(-)>  { <sym> <O(|%additive)> }
+    token infix:sym«∖»    { <sym> <O(|%additive)> }
 
     token prefix:sym<let>  { <sym><.kok> <O(|%named_unary)> { $*W.give_cur_block_let($/) } }
     token prefix:sym<temp> { <sym><.kok> <O(|%named_unary)> { $*W.give_cur_block_temp($/) } }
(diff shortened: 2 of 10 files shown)


AlexDaniel (commented):
This changes the precedence table so I'm a bit worried. Are there any changes in precedence that users will see?

lizmat:
I don't know.  It did *not* trigger any spectest issues and if *did* fix the issue.

AlexDaniel:
> I don't know. It did _not_ trigger any spectest issues

That doesn't instill confidence. The jump is across `named_unary` so some sort of combination of `temp` and set ops can have a change in behavior, and I'm totally not surprised that roast doesn't have a test for that (it's very patchy). Also, structural ops seem to be non-associative but set ops are, and I don't know if that's a positive or a negative.

> did fix the issue

The ticket comes from codegolf. I'd prefer to see more evidence for fixes like this.

lizmat:
> The ticket comes from codegolf.

Why does that matter?  I mean, I just fixed the `my &MAIN` nit you found.  Which was you looking for "trouble".

> I'd prefer to see more evidence for fixes like this.

Well, maybe a blin run would be a good thing to do then.

AlexDaniel:
> Why does that matter?

Because changing the precedence table is very scary. Fixing it is probably a good idea, but changing the precedence table carelessly means that you can cause bugs for actual users just because somebody cornered themselves into a particular buglet when golfing.

lizmat:
> Because changing the precedence table is very scary. Fixing a bug is probably a good idea

Whenever is fixing a bug **NOT** a good idea?

> but changing the precedence table carelessly

And here you are implying again that I'm apparently doing things `carelessly`.  And you wonder why people think you're abrasive.  Please **STOP** this manner of communication.

> means that you can cause bugs for actual users

That's why we have spectests.  If the coverage of the spectests is not good enough, we should fix *that*.  We also have at least a Blin run coming up, which potentially should show problems.

> just because somebody cornered themselves into a particular buglet when golfing.

For what it is worth, I do **NOT** consider this a "buglet".  I consider that a serious shortcoming of `Z` and `X` handling with regards to set operators.  Which I either implemented from scratch (in the case of the set operators) or significantly streamlined in the past years (in the case of `Z` and `X`).  How somebody found this, does **NOT** matter.  So I think I can say a know a little more about them than most (core) developers.

AlexDaniel:
> And here you are implying again that I'm apparently doing things carelessly. And you wonder why people think you're abrasive.

I don't wonder, I know because I'm saying things directly with zero sugar coating. I asked you if any changes in precedence will affect user code, you said you don't know but it is spectest clean. That's an extremely low bar for a change like this. I'm totally fine with somebody else reviewing your PRs if they can take your arrogance, but this PR was open for 4 days so 🤷

AlexDaniel:
BTW why are there two files? For example, this PR does not modify this:

https://github.com/rakudo/rakudo/blob/e0e24a8dbfd73b20697096467f499f8def5cea3a/src/Perl6/Grammar.nqp#L4153

lizmat:
> BTW why are there two files?

I assume you mean: "why are there **NOT**  two files?"

> For example, this PR does not modify this:

This kind of **factual** feedback helps: indeed, changing those as well, breaks spectests.  Thank you for noting this.

AlexDaniel:
> This kind of factual feedback helps: indeed, changing those as well, breaks spectests. Thank you for noting this.

You are choosing to ignore every opinion I express unless I then spend the time to find runnable snippets that back up what I say. You can save time for both me and yourself if you listened just a bit more.

lizmat:
> You are choosing to ignore every opinion I express unless I then spend the time to find runnable snippets that back up what I say.

opinions vs data.  I'll take data any time.

> You can save time for both me and yourself if you listened just a bit more

That feels like "Proven guilty until proven innocent".  I'd rather go the other way around: "Proven innocent until proven guilty".  You've just demonstrated "guilty".  And I'm acting on it.

AlexDaniel:
> I'll take data any time.

Except that when I give you data you then add a little bit of code to cover that data point, and the process repeats. It's good to rely on data, but you requiring data *every single time* is just arrogance. It should be normal to take into account someone's knowledge instead. Some approaches are just inherently wrong, and some changes are inherently disruptive. Don't touch the precedence table unless you have good reasons and good understanding of what is going to change.

AlexDaniel:
> > BTW why are there two files?
>
> I assume you mean: "why are there NOT two files?"

No, I actually ask why there are two files with the same information. The question isn't really within the scope of this PR, but still. Information shouldn't be duplicated like that.

… And because you really need me to come up with snippets to prove that it's not OK:

https://github.com/rakudo/rakudo/blob/e0e24a8dbfd73b20697096467f499f8def5cea3a/src/Perl6/Grammar.nqp#L4382

https://github.com/rakudo/rakudo/blob/e0e24a8dbfd73b20697096467f499f8def5cea3a/src/core.c/precedence.pm6#L170

Are these talking about the same op? And if so, why is the precedence different? Maybe these two files are meant to have different precedence levels? I don't know, hence the question.

lizmat:
> Information shouldn't be duplicated like that.

Agree.  And I'm pretty sure this will go in the rakuast branch.  It is my understanding that the precedence levels in the Grammar are used for codegenning, but that that information is basically lost in the Rakudo setting.  Therefore, they need to be set again (at the moment).  This is largely caused by the fact that the current Grammar / Actions need a World to build all the necessary data-structures.  And that's one of the goals of the rakuast branch to get rid of.

So why was it done this way?  Very simple: to make it work.  Could it be done better?  Yes, that's one of the reasons for the rakuast branch.

AlexDaniel:
OK, understandable. If it goes away eventually then no problem.

AlexDaniel:
I can appreciate the humour of that last commit and the consequent marking of the PR as ready to review.

In case it is not a joke, let me mention again that changing the precedence table (especially jumping even more precedence levels) has consequences. For example:

```raku
<AlexDaniel> c: setops-prec-non,HEAD say 5 (+) 3 xx 2
<committable6> AlexDaniel, ¦setops-prec-non: «(Bag(3 5) Bag(3 5))␤» ¦HEAD(e0e24a8): «Bag(3(2) 5)␤»
```

I'm surprised if the spectest is clean for that. As I mentioned, it's a very low bar to pass.

lizmat:
@AlexDaniel Would you have reviewed it if it was *not* marked as "Ready for review" ?  Then what does "Ready for review" mean?

AlexDaniel:
> @AlexDaniel Would you have reviewed it if it was _not_ marked as "Ready for review" ? If so, then what does "Ready for review" mean?

Well… that's correct. I was just not expecting the same precedence change just done differently (as that's what I already reviewed)… If you meant it to be reviewed then yes, you did it right.

This commit is completely different so all what I said before no longer applies. I don't know anything about metaop handling and NQP code is very hard to read for me. I can kinda understand what the change is doing, but, for your future commits, the commit message could've explained why. As in, not just that it fixes the bug, but *why* this new logic is resulting in a bugfix (“why” is usually the hardest to deduce).

Anyway, what about this? What should it produce?

```raku
<AlexDaniel> c: setops-prec-non,HEAD say 1..3, 1..3 Z∩ 2..4, 2..4 Z∩ 2..3, 2..3
<committable6> AlexDaniel, ¦setops-prec-non,HEAD(e0e24a8): «(Set(1 2 3 4) Set(1 2 3 4))␤»
```

lizmat:
Not this  :-(

lizmat:
$ raku -e 'say 1..3, 1..3 Z∩ 2..4, 2..4 Z∩ 2..3, 2..3'
    (Set(2 3) Set(2 3))

AlexDaniel:
What is going on in that last commit? Are you essentially changing set ops from list associativity to left associativity?

AlexDaniel (changes requested):
Let's stop the bruteforce and try to figure out what's going on here. The last fix is essentially making set ops left-associative when used with metaops. *Why?* Does it mean that other list associative ops are broken when used with Z and X? If not, why are set ops special?

It seems like something else is not working as expected and hacking the metaop mechanism is just applying a patch in the wrong place.

lizmat:
> What is going on in that last commit? Are you essentially changing set ops from list associativity to left associativity?

Well, using the same execution path as left associativity to prevent code duplication.  Because that is what it would do if it would get to the `**@p` candidates in the actual set operator handlers.

lizmat:
>  If not, why are set ops special?

**These** set ops are set ops that create another `QuantHash`.  All other set ops return a Bool.

> It seems like something else is not working as expected and hacking the metaop mechanism (for set ops specifically) is just applying a patch in the wrong place.

The problem with *these* set ops, is that the "list" associativity calls the set operator with a non-itemized `List`.  Ideally, this should be picked up by the `**@p` candidate of the set op, but it is not, and instead picked up by the `Any` candidate, which just turns the `List` into a `Set`.

Adding a specific `List` candidate to these set operators, is not the right thing either.

That's is why I went for the additional precedence and the "set" flag for special handling.

AlexDaniel:
It seems like the Any candidate was added here: https://github.com/rakudo/rakudo/commit/e8cb9a2f521847983385b8c5297d74a8f96afc08#

> Ideally, this should be picked up by the **@p candidate of the set op, but it is not, and instead picked up by the Any candidate

Any way we can be sure that the Any candidate should exist? Because it conflicting with `**@p` candidate is perhaps a bug by itself.

PR #3951, opened 2020-10-08
Add sanity check in infix handling, fixes R#3949

lizmat:
Before this, an "infix:(&)" (note the missing < > there) would cause a:

    $ raku -e 'infix:(&)'
    This type (QAST::WVal) does not support positional operations

crash without any indication on *where* in the code the actual error was.
After this, it will cause a correct compilation error:

    $ raku -e 'infix:(&)'
    You can't adverb :(&)
    at -e:1
    ------> infix:(&)⏏<EOL>


AlexDaniel (approved):
Looks good to me.

PR #3953, opened 2020-10-11
Implement the prefix:<||> operator in postcircumfixes

lizmat:
- for 6.e and later only
- described by https://design.raku.org/S09.html#line_419
- sort of implemented by Wenzel P.P. Peppmeyer in Slippy::Semilist
- does not actually add a prefix operator: it's just a double slip
- manipulates the AST in case of a postcircumfix, to select the ; variant
- the postcircumfix:<{; }> variant does the right thing for prefix:<||>

### DIFF
--- a/src/Perl6/Actions.nqp
+++ b/src/Perl6/Actions.nqp
@@ -1538,11 +1538,11 @@ class Perl6::Actions is HLL::Actions does STDActions {
 
     # Produces a LoL from a semicolon list
     method semilist($/) {
-        if $<statement> {
+        if $<statement> -> $statements {
             my $past := QAST::Stmts.new( :node($/) );
-            if nqp::elems($<statement>) > 1 {
+            if nqp::elems($statements) > 1 {
                 my $l := QAST::Op.new( :name('&infix:<,>'), :op('call') );
-                for $<statement> {
+                for $statements {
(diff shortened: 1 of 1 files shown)


src/Perl6/Actions.nqp
```diff
@@ -1538,19 +1538,32 @@ class Perl6::Actions is HLL::Actions does STDActions {
...
                 $past.push($l);
                 $past.annotate('multislice', 1);
             }
             else {
-                $past.push($<statement>[0].ast || QAST::WVal.new( :value($*W.find_single_symbol('Nil')) ));
+                my $ast := $statements[0].ast;
+
+                # anything like "||foo" ?
+                if nqp::istype($ast,QAST::Op) && $ast.name eq '&prefix:<|>' {
+                    if $*W.lang-ver-before("e") {
```
AlexDaniel, line 1557: Any particular reason why this check is before the next if? Once .e is released (and going forward), it'd be checking the version any time it stumbles upon a `|`, regardless of whether it is `||` or not. I guess checking for the version is very cheap, but still, wouldn't it be better to check the version inside the next `if`?

lizmat: Addressed with 8dcba57368

PR #3954, opened 2020-10-12
Make Hash.DELETE-KEY return Nil on non-existent keys

lizmat:
Currently, it returns the default value.  Which basically does not
allow you to differentiate between:

    my %h; dd %h.DELETE-KEY("a");             # Any

which does not remove a key, and:

    my %h = a => Any; dd %h.DELETE-KEY("a");  # Any

which *did* remove a key.  Making it return `Nil`, *would* allow you to
see the difference (well in all but the most contrived cases).

This is a breaking change, it breaks these spectests:

    t/spec/S32-basics/xxKEY.rakudo.moar                             (Wstat: 768 Tests: 129 Failed: 3)
      Failed tests:  18, 37, 56
      Non-zero exit status: 3
    t/spec/S32-hash/delete-adverb.t                                 (Wstat: 768 Tests: 131 Failed: 3)
      Failed tests:  21, 27, 39
      Non-zero exit status: 3
    t/spec/S32-exceptions/misc.rakudo.moar                          (Wstat: 512 Tests: 180 Failed: 2)
      Failed tests:  156-157
      Non-zero exit status: 2
    t/spec/integration/advent2013-day12.t                           (Wstat: 768 Tests: 32 Failed: 3)
      Failed tests:  9, 11-12
      Non-zero exit status: 3

### DIFF
--- a/src/core.c/Hash.pm6
+++ b/src/core.c/Hash.pm6
@@ -167,7 +167,7 @@ my class Hash { # declared in BOOTSTRAP
             nqp::getattr(self,Map,'$!storage'),
             key
           )),
-          $!descriptor.default,
+          Nil,
           nqp::stmts(
             nqp::deletekey(
               nqp::getattr(self,Map,'$!storage'),
@@ -540,7 +540,7 @@ my class Hash { # declared in BOOTSTRAP
                 nqp::getattr(self,Map,'$!storage'),
                 (my str $WHICH = key.WHICH)
               )),
-              TValue,
+              Nil,
               nqp::stmts(
                 nqp::deletekey(nqp::getattr(self,Map,'$!storage'),$WHICH),
                 nqp::getattr(value,Pair,'$!value')


vrurg:
I think this change doesn't make a lot of sense. The only reliable way would always be to test for a key existence prior to deleting it. Because:

```raku
my %h;
%h<A> := Nil;
dd %h<A>:delete; # Nil
```

So, in my code, possibly operating on third-party generated hashes, I wouldn't rely on `DELETE-KEY` returning `Nil` as there is non-zero chance it was intentionally bound to the key.

lizmat:
But why would you bind `Nil` to a key?  What would that mean?

vrurg:
No idea. We can predict nothing of what might come into one's mind. But since that kind of binding is not disallowed then it's possible and should be considered.

mendel:
One can also argue that it breaks some assumptions about the default value: Currently one can rely that `%h<foo>:delete` always returns the same as `%h<foo>` would (before deleting the key, of course). If this PR gets merged, that would not be true any more. It becomes even more apparent if this value is a defined value (cf. the less contrasting distinction between `Any` and `Nil`), which can be the case when the hash has a default value set.

For example, the current behaviour is:

    my %h is default(42);
    %h<nonexistent>         # 42
    %h<nonexistent>:delete  # 42

This PR would change the result of the second expression to `Nil` while leaving the first to be `42`.

Focusing on the original problem, which I understand is to differentiate between deleting a key whose value is `Any` and attempting to delete a non-existent key, it seems to me that it is easy to achieve when using the `&postcircumfix:<{}>` operator, by combining the `:delete` adverb with the `:k`/`:p` and/or the `:exists` adverbs (eg.`:delete :v` or `:delete :p`):

    my %h = foo => Any, bar => Any;

    %h<foo>:delete:v          # Any
    %h<nonexistent>:delete:v  # ()

    %h<bar>:delete:p          # :foo(Any)
    %h<nonexistent>:delete:p  # ()

So I assume this is rather an optimisation problem than just making something possible.

The implementation of the `&postcircumfix:<{}>` operator, when used with the `:delete :v` / `:delete :p` adverb combinations, indeed performs up to two method calls on the hash: first `%h.EXISTS-KEY($k)` then if the key exists, it's followed by a `%h.DELETE-KEY($k)` call. If I understand it correctly, the point was to eliminate the need for a separate `%h.EXISTS-KEY($k)` call.

@lizmat Is the above analysis correct? :-)

Issue #3582, opened 2020-03-31
RFE: shell / no shell switch


ToddAndMargo:
Hi All,

Windows 10-1909 Pro
This is Rakudo version 2020.01 built on MoarVM version
2020.01.1 implementing Perl 6.d.

There are times when having a big black shell pop up when running a Raku program and othes when it is very much not desirable, such are running somethign fromteh registry:
```
[HKEY_CLASSES_ROOT\*\shell\OpenWithFileAttributes.pl6\command]
@="\"C:\\rakudo\\bin\\raku.exe\" \"K:\\Windows\\NtUtil\\FileAttributes.pl6\" \"%1\""
```
Would you please consider adding a no shell command line switch to the run string that would disable the big black shell pop up?

Raku --noshell myprogram.pl6

Many thanks,
-T

ToddAndMargo:
Any movement on this?  I have customers killing the black box pop ups and causing havoc.

Altai-man:
Hi. It is hard to say what is going on without knowing more about the code you have.
Can you provide some of the code you use? I suspect it it runs external commands using `shell` or `run`?

patrickbkr:
I [read up](https://stackoverflow.com/a/574918/1975049) a bit about this. On Windows an executable is either marked as a `console` or a `windows` executable (or some obscurer others). I *think* that our `raku.exe` executables are marked as `console`. Which would mean they always spawn a console window in which the output is printed. If my guess is correct we should evaluate what the consequences of making our executables `windows` would be.

vrurg:
@patrickbkr no way. A specific `rakudo-w.exe` which is a Win version – yes.

vrurg:
Not really as I did very little win-programming. Literally – played for a moth with something years ago.

My point is that `rakudo` executable must open a console because it is likely to be desirable in many situations like double-clicking a _.raku_ file.

ToddAndMargo:
On 2020-07-20 07:39, Altai-man wrote:
> Hi. It is hard to say what is going on without knowing more about the 
> code you have.
> Can you provide some of the code you use? I suspect it it runs external 
> commands using |shell| or |run|?


It is being run from from Cobian Backup.  I am
not sure how he is calling things.

ToddAndMargo:
On 2020-07-20 08:02, Vadim Belman wrote:
> Not really as I did very little win-programming. Literally – played for 
> a moth with something years ago.
> 
> My point is that |rakudo| executable must open a console because it is 
[...]

PR #3955, opened 2020-10-13
Generate a `rakuw.exe` on Windows

patrickbkr:
On Windows executables are either marked as "console" applications or
"windows" applications. The first always spawn a console window to which
STDIN, STDOUT and STDERR are connected, the second spawns no such window
and STDIN, STDOUT and STDERR are not connected to anything. This can't be
decided at runtime, but is a flag set on the executable file itself. So the
best we can do to support both usage scenarios is provide separate binaries
for each scenario.
Python does the same. The `python.exe` executable is compiled as "console"
while `pythonw.exe` is a "windows" executable. There is no point in making
up a new naming scheme, so we just do what the others do and name it
`rakudow.exe` / `rakuw.exe`.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -106,6 +106,8 @@ perl6.rc
 /inst-perl6-debug.exe
 /inst-perl6-m.exe
 /inst-perl6-debug-m.exe
+/inst-perl6w.exe
+/inst-perl6w-m.exe
 /inst-rakudo
 /inst-rakudo-debug
 /inst-rakudo-m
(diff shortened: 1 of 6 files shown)


Altai-man:
There was a ticket asking for this, I wonder if we can look it up to close as this is merged?

patrickbkr:
@Altai-man It's https://github.com/rakudo/rakudo/issues/3582 That ticket was the reason for me to start working on this. I'll make sure to close it once it's actually solved.

PR #3957, opened 2020-10-14
Replace all simple nqp::if(42,foo,bar) with ternaries

lizmat:
If memory serves me well, about 1.5 years ago, optimizations in
Rakudo made the construct:

   42 ?? foo !! bar

equally fast as the nqp equivalent:

  nqp::if(42,foo,bar)

This commit converts all of the simple nqp::if() structures in the
core.c setting to their ternary equivalents.

- binary size of CORE.c.setting.moarvm file is unchanged (almost)
- speed should be on par with the nqp::if() equivalents
- readability of the core setting should be improved.

make test and make spectest clean.


AlexDaniel (approved):
Last time I checked there was still a consistent difference, even if minor. Maybe that's no longer true. Either way I find the readability improvement much more important. Looked through the changes and didn't find any accidental mistakes. Great work!

src/core.c/BagHash.pm6
```diff
@@ -149,7 +145,7 @@ my class BagHash does Baggy {
                 nqp::if(
                   nqp::existskey(elems,$key),
                   nqp::getattr(nqp::atkey(elems,$key),Pair,'$!value'),
-                  0
+                  # 0 the value of existskey if the key doesn't exist
```
AlexDaniel, line 148: Ah btw I don't know if this change was intended to be here.

lizmat: It is...   to make it similar to the approach for Bag.AT-KEY

AlexDaniel: OK! Just looked different in an otherwise repetitive diff, so made me wonder.

PR #3960, opened 2020-10-15
[RakuAST] (:!foo).IMPL-INTERPRET should have a False value

b2gills:
It seems like the code for `RakuAST::ColonPair::True.IMPL-INTERPRET()` was copied to `RakuAST::ColonPair::False` but without the proper alteration.


lizmat:
Looks like a good catch.  I guess we need more tests!   :-)

lizmat (approved):

PR #3962, opened 2020-10-16
Fix NativeCall/C++ to allow double instantiation

mid1221213:
The test of the arguments concreteness using the Op
<isconcrete> instead of <isconcrete_nd> was preventing
the instantiation more than once of the same class
using the following code:

my Class $i .= new;
my Class $j .= new;

The second line, without this commit, fails with the wrong
message: « Native call expected argument 1 with CPPStruct
representation, but got a P6opaque (Scalar) »

This commit possibly fixes other issues related to
argument passing to native calls.


niner:
Thanks a lot!

PR #3965, opened 2020-10-18
Remove RESTRICTED setting

AlexDaniel:
RESTRICTED settings looks as if it is a security feature, but this
issue cannot be approached with a blacklist. For example, the things
that it blacklists are still available through method calls and
nativecall, and there is just no way to plug all the possible
holes. Moreover, those who actually need to run code securely will
probably resort to seccomp or the like, with a whitelist of allowed
syscalls, so the compiler seems to be the wrong place to implement
this.

I guess RESTRICTED setting was useful as a “don't accidentally break
the evalbot” feature, but it stopped working completely after the
release of v6.d.

This commit passes `test` (and `spectest`) as there seem to be no
tests at all for the feature.


AlexDaniel:
@patrickbkr can you please check? I'm not sure about the logic in makefiles.

AlexDaniel:
It seems like the feature was introduced in 7c001ff636b528d99eb130e48267a0720e5e0b75:

> Add first cut of SAFE.setting, for the benefit of p6eval

Weird stuff!

vrurg:
Basically, it's a good idea. The only thing: there is currently a bug in setting loading logic which doesn't allow to use a custom setting. Fixing it requires time which rather scarce resource on my side currently. And when I eventually get there I'd like to have a ready to use setting to test against. So, it's nothing but my wish to keep it there for a while as it harms nobody.

AlexDaniel:
@vrurg anything wrong with `git revert 6ad69ab7d43` to conduct your tests?

vrurg:
@AlexDaniel The revert could be not possible without hard to resolve conflicts if too much changes happen in the meanwhile.

Anyway, I'm not against removing it. Just would like my point to be considered prior to merging.

lizmat:
I would be against merging this PR until we have a solution for the Camela bot, which relies on the RESTRICTED setting.

AlexDaniel:
@lizmat how is it useful for the bot? If it stopped working with v6.d release, then for two years now it didn't work at all, and even if it did, it wouldn't've protected against anybody trying to intentionally cause damage. @niner can probably comment more, but I think (I can be mistaken) meanwhile the bot was restricted just a tiny little bit more by making most of the filesystem non-writable, which should probably be enough to protect against unintentional damage.

lizmat:
Looks like this is ready for merging again.  Question is, should we?  @moritz @niner how would this affect Camelia nowadays?

src/core.c/Nil.pm6
```diff
@@ -1,5 +1,4 @@
 my class Nil is Cool { # declared in BOOTSTRAP
-
```
MasterDuke17, line 2: Looks like this is an artifact from the rebase.

lizmat: The empty line removal?  Probably.  Is it a problem?  Don't  think so?

MasterDuke17: Problem, no, just clutters the git history.

niner:
Camelia lives in her own VM, runs as an unprivileged user and is now further confined by AppArmor. I don't think the RESTRICTED setting would add anything to that and it has been completely ineffective for a long time anyway.

lizmat:
Ok, so good to merge then?

niner (approved):

vrurg (approved):

Issue #3971, opened 2020-10-19
Attempting to copy a directory to an existing file throws but destroys the file


tbrowder:
## The Problem

Given an existing directory and existing file, inadvertently attempt to copy the directory to the file.

## Expected Behavior

The attempt will result in an exception and the target file will be unaltered.

## Actual Behavior

The file is destroyed.

## Steps to Reproduce

execute:

````
my $f = "/tmp/f1";
spurt $f, "some text";
my $dir = "/tmp/dir";
mkdir $dir;
copy $dir, $f;
````

get OUTPUT:

````
Failed to copy '/tmp/dir' to '/tmp/f1': Failed to copy file: illegal operation on a directory
  in block <unit> at dirfil.raku line 5
````

check the original file:

````
$ ls -l /tmp/f1
ls: cannot access '/tmp/f1': No such file or directory
````

## Environment

* Operating system: Debian Buster (10)
* Compiler version (`perl6 -v` or `raku -v`):
This is Rakudo version 2020.09 built on MoarVM version 2020.09
implementing Raku 6.d.

lizmat:
Thanks for the spot and thanks for the fix and tests!

PR #3973, opened 2020-10-19
Fix issue #3971: attempt to copy a directory to a file destroys the file!

tbrowder:
This **very serious** issue apparently has been in core for a very long time and **there is no existing test for it**.  It should be incorporated **before the next release**.

**ISSUE:** Attempting to copy a directory to a file fails as expected, **but the file is deleted**. 

**FIX:** In **method copy** add an initial check for the **self** and **$to** IO types and fail immediately if **self** is a directory and **$to** is a file. (See roast PR 690 for an added test for this fix.)

File modified:

    src/core.c/IO/Path.pm6

### DIFF
--- a/src/core.c/IO/Path.pm6
+++ b/src/core.c/IO/Path.pm6
@@ -448,6 +448,13 @@ my class IO::Path is Cool does IO {
     }
 
     method copy(IO::Path:D: IO() $to, :$createonly --> True) {
+        # add fix for issue #3971 where attempt to copy a dir
+        # to a file clobbers the file.
+        self.d and $to.f and fail X::IO::Copy.new:
+            :from($.absolute),
+            :to($to.absolute),
+            :os-error('cannot copy a directory to a file');
+
         $createonly and $to.e and fail X::IO::Copy.new:
             :from($.absolute),
             :to($to.absolute),


niner:
Please specify the actual change and the reason for it in the commit message (especially in the subject). Just referring to some issue makes it hard to judge, whether one is qualified to review, makes it harder to dig through the code's history in git blame and makes it harder to write change logs.
Mentioning the issue number is helpful, only stating the issue number is not enough.

niner (approved):

lizmat (approved):

AlexDaniel:
I started writing this review and then… well, you'll see yourself. I added some comments afterwards, they look like this: “(… )”.

First of all, I suggest to write a few words when doing reviews of PRs. It helps to understand whether the review was actually made (in real world it happens sometimes that people see a change that sounds useful and just click “approve” without reading the code), and it also gives some light into what the reviewer was and wasn't considering. Note that I personally also expect some affirming comments like “this PR makes sense because …”, not just what could be done better next time. Especially for something that's at risk of wiping user data, you know.

So, looking at the PR. First of all it introduces a race condition because a file can be created after the check. That is, Rakudo can still delete a file when it shouldn't (… well, supposedly, because that's an argument being made here). So, really, the issue is not fixed but is just hidden a bit better. That's an OK temporary workaround, but it is not a fix. Did reviewers approve it having that in mind? I don't know, but then the original issue was closed so maybe that was simply overlooked. That's OK too, will reopen (… actually won't, because this whole issue is bonkers, a new ticket is needed I think).

Then there's this XY question: Why is the file deleted in the first place? I don't see a delete call anywhere around the lines changed by the PR. If it's not there, then the “fix” is in the wrong place. So this shouldn't be a PR to Rakudo. That said, I did check if `nqp::copy` is used anywhere else in rakudo, and it isn't, so the workaround is actually applied “everywhere”, which is good (… but still completely irrelevant)!

Then come the edge cases. What if it isn't a file, but something else? For example, what's going to happen if it's a symlink?

So I started testing…

Turns out, `copy` can't be used to copy directories. At all. What is this PR and the original issue is even talking about? I don't understand.

So let's try this (simplified from the roast PR):
```raku
mkdir ‘foodir’;
copy ‘foodir’, ‘destpath’;
```

That code creates an empty file `destpath`. Successfully. Let that sink in. Seeing the contributions in this PR and the actual behavior really made me wonder if I'm doing something completely wrong… I was on 2020.05.1, I upgraded to master, I'm still seeing the same. Please let me know if I'm overlooking something obvious, because at this point I'm starting to question my sanity :)

Assuming my understanding is correct, there's a lot wrong here:
* “Attempting to copy a directory to a file fails as expected” – no it *doesn't* fail as “expected”. It only fails if you have files in that directory. Otherwise it succeeds. If the directory is not empty, it fails with “Failed to copy file: illegal operation on a directory”. It's a wrong operation attempted, nothing to do with the existence of a destination file.
* “but the file is deleted” – it is not deleted, it is overwritten successfully. If there's a failure during the operation, *then* it is also deleted after it was overwritten.

Now, it is arguably true that trash is written into the destination. Any user that writes the code to `copy` a directory most likely expected it to actually work with directories, not to produce an empty file. So, if a workaround is meant to exist, it should be checking if the *source* is a directory. To clarify, “copying” a directory into a directory is also not possible.

I did briefly look into what a proper fix could be. NQP doesn't handle the copy logic and it goes straight to MoarVM. MoarVM then uses [uv_fs_copyfile](http://docs.libuv.org/en/v1.x/fs.html#c.uv_fs_copyfile), which then uses `uv_fs_open` to open the source. From what I understand, that is meant to work like [open](https://man7.org/linux/man-pages/man2/open.2.html), which, surprisingly, allows opening directories. I guess we should check how other libuv users are handling this.

FWIW, `move` also doesn't move directories, and similarly creates empty files.

tbrowder:
You are correct  that some weird behavior is going on, but this fix was to prevent immediate file destruction.

My fix DOES check that the source is a dir and fails if the destination is a file. No, I did not consider to check if an empty dir as source works.

I cannot find the issue again, but many of your concerns are listed there.

AlexDaniel:
@tbrowder don't stress it, the PR is probably a good starting point. What I find surprisingly is how it got two approving reviews, and then was merged, without iterating on it.

I recall https://github.com/rakudo/rakudo/pull/2043 but I think there was one more ticket or PR somewhere.

AlexDaniel:
Quick example using nodejs:
```
var fs = require('fs');

function callback(err) {
    if (err) throw err;
    console.log('Copied!');
}

fs.copyFile('foodir', 'bartesttest', callback);
```

* Creates an empty file if `foodir` is an empty directory.
* Overwrites the file if it exists.
* Fails with `illegal operation on a directory` if the directory is not empty.

In other words, behaves exactly the same way as Rakudo before this PR.

PR #3974, opened 2020-10-20
Add Mathematical char ranges to .succ / .pred handling

lizmat:
The title says it all.


lizmat:
Closing this PR as these char ranges are actually full of holes so that they would require a lot of extra handling, which I'm currently don't want to spend time on.

PR #3975, opened 2020-10-22
Expose nqp::indexingoptimized as Str.no-strands

lizmat:
Calling .no-strands on a Str will make sure that the string in
memory consists of a single array of codepoints, rather than possibly
as a collection of strands (which typically happens with a lot of
~= work on a string).

Benchmarking this is tricky, but a .comb on a string that consists
of 1000 strands, is about 7% faster with calling .no-strands on it
once.

YMMV

### DIFF
--- a/src/core.c/Str.pm6
+++ b/src/core.c/Str.pm6
@@ -53,6 +53,10 @@ my class Str does Stringy { # declared in BOOTSTRAP
         nqp::bindattr_s(self, Str, '$!value', nqp::unbox_s($value))
     }
 
+    method no-strands(Str:D: --> Str:D) {
+        nqp::bindattr_s(self,Str,'$!value',nqp::indexingoptimized($!value))
+    }
+
     multi method Bool(Str:D: --> Bool:D) {
         nqp::hllbool(nqp::chars($!value));
     }


coke:
Should we add tests for this (basically just that the method exists and returns a Str that compares same), or mark it as an internal to rakudo method?

coke:
Bikeshed: "collapse" or "compact" instead?

lizmat:
Well, we can bikeshed about the name.  I was just wondering whether it would make sense to expose this.  And how much it could make things faster.  My benchmark was really a worst case scenario, so it would have the most improvement.

Another idea could be that when spesh sees something happening on the same string over and over again, it could do a `nqp::indexingoptimized` under the hood.

vrurg:
Actually, since the method exposes internal implementation detail of MoarVM, it shouldn't be a part of the language spec. Not with the name given. But keeping in mind that other backends may have some kind of mem-optimizations for strings that would affect performance, a method of the kind could be specced. Though no good variant comes into mind as _flattening_ term has different meaning in Raku, and _optimizing_ is ambiguous.

Altai-man:
I imagine this kind of implementation detail should not be exposed? For a user, there is `Str` and API it provides. Be it implemented with strands or arrays or something else should not really matter for the user perspective, nor the user should know anything about it on a method level.
If a user feels like a pro, they can just call nqp, thus taking consequences on themselves. But exposing it as a method says "We will be responsible for this forever and ever".

lizmat:
> We will be responsible for this forever and ever

Well, that's the question, I guess.  I mean, for other implementations, it could be a no-op.  Having people force to use nqp, hurts compatibility with other implementations.  Allowing for a HLL interface, even if it is a no-op, would at least provide some sort of compatibility.

lizmat:
I'm closing this in favour of spesh potentially handling the use case

PR #3978, opened 2020-10-23
Add a Mu ACCEPTS candidate for Mu:D

Kaiepi:
Smartmatching against `Mu` type objects used to throw. I fixed this with
3b4794f, but at the time I didn't think to check if the same would occur
with `Mu` instances or not (it does). This makes it so `Mu` type objects
will smartmatch against instances of themselves, and `Mu` instances will
smartmatch against themselves.

Passes `make test` and `make spectest`, but is better left alone until after this upcoming release.

### DIFF
--- a/src/core.c/Mu.pm6
+++ b/src/core.c/Mu.pm6
@@ -21,11 +21,11 @@ my class Mu { # declared in BOOTSTRAP
     multi method ACCEPTS(Mu:U: Mu \topic) {
         nqp::hllbool(nqp::istype(topic, self))
     }
-    multi method ACCEPTS(Mu:D: Mu \topic) {
-        nqp::hllbool(nqp::eqaddr(nqp::decont(topic), nqp::if(
-          nqp::defined(topic),
-          self,
-          self.WHAT)))
+    multi method ACCEPTS(Mu:D: Mu:U \topic) {
+        nqp::hllbool(nqp::eqaddr(nqp::decont(topic), self.WHAT))
+    }
+    multi method ACCEPTS(Mu:D: Mu:D \topic) {
+        nqp::hllbool(nqp::eqaddr(nqp::decont(topic), self))
     }
     # Typically, junctions shouldn't be typechecked literally. There are
     # exceptions though, such as Junction in particular, so this probably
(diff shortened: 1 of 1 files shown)


src/core.c/Mu.pm6
```diff
@@ -21,6 +21,12 @@ my class Mu { # declared in BOOTSTRAP
     multi method ACCEPTS(Mu:U: Mu \topic) {
         nqp::hllbool(nqp::istype(topic, self))
     }
+    multi method ACCEPTS(Mu:D: Mu \topic) {
```
lizmat, line 24: Why not a `Mu:U \topic` and a `Mu:D \topic` candidate?

Kaiepi: I don't do that because it's possible to create values that won't typecheck against `Mu:U` or `Mu:D` (but will with `Mu`) with the MOP:

```perl6
class NonMuHOW {
    method new_type {
        Metamodel::Primitives.create_type: self.new, 'Uninstantiable'
    }
}

my Mu   $ := NonMuHOW.new_type; # OK
my Mu:U $ := NonMuHOW.new_type; # Dies
```

vrurg: It shouldn't die. Apparently, this is a bug.

vrurg: Actually, it's not that clear to me now. The behavior is certainly inconsistent. But wether it's bad or good – I'm not sure.

If first place, it dies because `ClassHOW` explicitly adds `Mu` as a parent if there're no explicit parents and no default parent is set yet. This makes all classes declared with `class` inherit from either `Any` or `Mu` by default. Thus, `Mu:U` explicitly requires a `Mu` descendant and fails.

The first statement doesn't die because of a special handling of Mu which causes type check to be skipped. This allows Raku to hold objects of any type in a `Mu`-typed symbol. This also seems to be logical because there should be a way to make it possible or we simply disallow custom types.

Wether this behavior should remain as it is or has to be changed is subject for a problem-solving ticket. But to my view `NonMuHOW.new_type ~~ Mu` must fail because this is what it is: a type object which is not `Mu`. And thus `Mu:U` is the right signature for the task.

Kaiepi:
Updated.

PR #3980, opened 2020-10-23
Add flag for creating symlinks without absolute path

mienaikage:
As `.absolute` is called on the path when `.symlink` is called, it is not currently possible to create a relative symlink using this method. The added `:literal` flag is intended to allow for the altering of this behavior, creating a symlink pointing to the path as it was given.

### DIFF
--- a/src/core.c/IO/Path.pm6
+++ b/src/core.c/IO/Path.pm6
@@ -496,8 +496,8 @@ my class IO::Path is Cool does IO {
         }}
     }
 
-    method symlink(IO::Path:D: IO() $name, Bool :$literal --> True) {
-        nqp::symlink($literal ?? ~self !! $.absolute, nqp::unbox_s($name.absolute));
+    method symlink(IO::Path:D: IO() $name, Bool :$absolute = True --> True) {
+        nqp::symlink($absolute ?? $.absolute !! ~self , nqp::unbox_s($name.absolute));
         CATCH { default {
             fail X::IO::Symlink.new:
                 :target($!os-path), :name($name.absolute), :os-error(.Str);
--- a/src/core.c/IO/Path.pm6
+++ b/src/core.c/IO/Path.pm6
@@ -496,8 +496,8 @@ my class IO::Path is Cool does IO {
         }}
     }
 
-    method symlink(IO::Path:D: IO() $name --> True) {
-        nqp::symlink($.absolute, nqp::unbox_s($name.absolute));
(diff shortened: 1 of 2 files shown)


lizmat:
Looks good from functionality point of view.  But why the choice for the name `literal`?  Wouldn't `relative` make more sense?

mienaikage:
I chose `literal` over `relative` as it allows you to specify a path without alteration, which may or may not be relative. `relative` seemed less descriptive to me for what is actually happening.

lizmat:
Ok, well, maybe a `:$absolute = True` flag would be more appropriate, which would be True by default, and you could indicate *not* calling absolute on the path with `:!absolute`.

mienaikage:
That does make sense, it was one of the options I had considered! (naming things is hard!)

lizmat:
Yeah.  FWIW, this feels very strange to me, being usually at the other end of this stick  :-)

I hope some other people will chime in!

tbrowder:
I like @lizmat's suggestion.

AlexDaniel:
So, ideally, how should it behave? I understand that there's a backcompat concern, but it's hard to judge if a particular kludge is actually the best we can do right now if an actual solution is not defined. For example, is named parameter better than an explicit call to `.absolute`? Is one default behavior better (less unexpected?) than the other?

lizmat:
Removing the call to `.absolute` on the inside of the `symlink` method would break code.

Adding a `:$absolute = True` named, would not.

> particular kludge

This is not a kludge, this is an addition to the existing API.  And a decent discussion about it.

AlexDaniel:
>  And a decent discussion about it.

I agree, and discussing what would be the best API would've been a great addition. As I've said, I understand that there is a backcompat concern, I get it. Doing “additions” to the API just because a wrong default was chosen in the past is also OK, but it doesn't hurt to stop and think what would've been the best solution because perhaps there's a way to get there.

mienaikage:
This seemed like the most appropriate solution to maintain backwards compatibility, but admittedly its current default behavior was not what I was expecting. I'd attempted to use `symlink` in the past, and forgotten why I had opted not to use it until I tried it again earlier today.

lizmat (approved):

lizmat:
@mienaikage Would be nice if you could add some tests for it as well, probably in t/spec/S32-io/io-path-symlink.t .

mienaikage:
@lizmat I shall look into putting some tests/docs together shortly.

Issue #3658, opened 2020-05-03
@a[-42..∞]:delete does not work


AlexDaniel:
Code:
```raku
my @a = 3, 4, 5;
@a[-5..∞]:delete; # same should work with *
say @a;
say @a.elems
```

Output:
```
[3 4 5]
3
```

It should work because -5..99999999 deletes all elements.

Output on all releases: https://gist.github.com/78189c36599a3dbb02724435723053b0

Related:
* (2017-10-08) https://github.com/rakudo/rakudo/commit/456358e3c380eeeb5fe5bc6260ba6f51c42a52ff
* (2018-02-08) https://github.com/rakudo/rakudo/commit/f330d7fc44d76258e651c9114360d08f929b25b9

AlexDaniel:
FWIW same happens with …

lizmat:
Negative indices are not allowed on slices, so it should die, should it not?

lizmat:
Working on this in the `faster-slice-access` :-)

lizmat:
The problem is not with infinity, but with laziness:
```raku
my @a = 1,2,3;
say @a[(-1..2).lazy];  # ()
say @a[(-1..2).lazy];  # Index out of range. Is: -1, should be in 0..^Inf
```

lizmat:
Also, it appears that slices produce a list with Failures.  The above error is caused by the `say` throwing a `Failure`, which `dd` doesn't:
```raku
my @a = 1,2,3;
dd @a[-1..9];
# (Failure.new(exception => X::OutOfRange.new(what => "Index", got => -1, range => "0..^Inf", comment => Any), backtrace => Backtrace.new), 1, 2, 3, Any, Any, Any, Any, Any, Any, Any)
```

lizmat:
So preliminary conclusion is that the laziness of the `Range` (and thus its iterator) causes the slice production to halt the production / adverb application of the slice if a `Failure`is produced (which is probably interpreted as an empty slot).  Compare:
```raku
my @a = 1,2,3;
dd @a[(1,-1,0)]:delete;
# (2, Failure.new(exception => X::OutOfRange.new(what => "Index", got => -1, range => "0..^Inf", comment => Any), backtrace => Backtrace.new), 1)
dd @a;
# [Any, Any, 3]
```
with
```raku
my @a = 1,2,3;
dd @a[(1,-1,0).lazy]:delete;
# (2,)
dd @a;
# [1, Any, 3]
```

PR #3983, opened 2020-10-24
Introducing Seq.slice(indices)

lizmat:
It appears that a "slice" method on Seq has been discussed extensively
in the longer ago past.  So I thought I'd have a go at it today.

The .slice method takes any number of indices that are monotonocally
increasing (that is, any later index value *must* be greater than any
previous index value), and produces the associated values from the
iterator of the given sequence.  To give an artificial example:

    my @a = ^10; dd @a.Seq.slice(0, 2 ... *);  # (0, 2, 4, 6, 8).Seq

is about 2x as fast as:

    my @a = ^10; dd @a.Seq[0, 2 ... *];       # (0, 2, 4, 6, 8)

This is because the .slice method does *not* have to do any caching.

Note that if an index is produced that does not follow the rule, an
exception will be thrown.

Turns out the underlying iterator had already been implemented for some
aspects of Match processing, so this was actually only creating a frontend
for that functionality.

### DIFF
--- a/src/core.c/Any-iterable-methods.pm6
+++ b/src/core.c/Any-iterable-methods.pm6
@@ -1519,6 +1519,9 @@ Consider using a block if any of these are necessary for your mapping code."
         (find-reducer-for-op(&with))(&with,1)(self)
     }
 
+    proto method slice(|) is nodal { * }
+    multi method slice(Any:D: *@indices) { self.Seq.slice(@indices) }
+
     proto method unique(|) is nodal {*}
 
     my class Unique does Iterator {
--- a/src/core.c/Any-iterable-methods.pm6
+++ b/src/core.c/Any-iterable-methods.pm6
@@ -1520,7 +1520,7 @@ Consider using a block if any of these are necessary for your mapping code."
     }
 
     proto method slice(|) is nodal { * }
-    multi method slice(Any:D: *@indices) { self.Seq.slice(@indices) }
+    multi method slice(Any:D: *@indices --> Seq:D) { self.Seq.slice(@indices) }
 
     proto method unique(|) is nodal {*}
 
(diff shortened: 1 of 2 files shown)


JJ:
Maybe take into account #3658 and throw on negative indices? Maybe check out Raku/old-issue-tracker#6233 while you're on it (with the corresponding Raku/roast#712)?

src/core.c/Any-iterable-methods.pm6
```diff
@@ -1556,6 +1556,9 @@ Consider using a block if any of these are necessary for your mapping code."
         (find-reducer-for-op(&with))(&with,1)(self)
     }
 
+    proto method slice(|) is nodal { * }
+    multi method slice(Any:D: *@indices) { self.Seq.slice(@indices) }
```
JJ, line 1560: Why not provide a return Type for this?

lizmat: I guess one could.

lizmat: Added with https://github.com/rakudo/rakudo/commit/9e61a71be7

lizmat:
The indexes provided must be monotonically increasing.  If they're not, it will fail.  `-1` is per definition lower than the lowest allowed index, so it already fails.

    $ raku -e '(1...*).slice(-1,2,4)'
    Provided index -1, which is lower than 0

lizmat:
> Maybe take into account #3658 and throw on negative indices? Maybe check out Raku/old-issue-tracker#6233 while you're on it (with the corresponding Raku/roast#712)?

This is being worked on in the `faster-slice-access` branch, fixing a large number of issues not even mentioned in these reports.

PR #3989, opened 2020-10-28
Add RAKU_TEST_DIAG_LINE_ON_FAIL

coke:
Defaults to 1. If set to empty/0, disables the diag output in testing
that adds file/line number information on failed tests.

Purpose is to conditionally hide additional diag output on failing tests. For files with individual tests, file::line is helpful. For tests that have loops and all fail in the same place, the diag output is useless. Leaving defaulted on in this PR, but could change the default to off easily.

### DIFF
--- a/lib/Test.rakumod
+++ b/lib/Test.rakumod
@@ -8,6 +8,8 @@ my int $raku_test_times =
   ?(%*ENV<RAKU_TEST_TIME> // %*ENV<PERL6_TEST_TIMES>);
 my int $die_on_fail =
   ?(%*ENV<RAKU_TEST_DIE_ON_FAIL> // %*ENV<PERL6_TEST_DIE_ON_FAIL>);
+my int $diag_line_on_fail =
+  %*ENV<RAKU_TEST_DIAG_LINE_ON_FAIL> // 1;
 
 # global state
 my @vars;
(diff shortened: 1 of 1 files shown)


coke:
```
$ RAKU_TEST_DIAG_LINE_ON_FAIL=1 install/bin/raku -MTest -e 'ok(0,"fail")'
not ok 1 - fail
# Failed test 'fail'
# at -e line 1
$ RAKU_TEST_DIAG_LINE_ON_FAIL=0 install/bin/raku -MTest -e 'ok(0,"fail")'
not ok 1 - fail
$ RAKU_TEST_DIAG_LINE_ON_FAIL= install/bin/raku -MTest -e 'ok(0,"fail")'
not ok 1 - fail
$ install/bin/raku -MTest -e 'ok(0,"fail")'
not ok 1 - fail
# Failed test 'fail'
# at -e line 1
```

lizmat:
FWIW, I don't really like the default to be 1.  Sometimes we have to for consistency's sake (like recently with the `IO::Path.symlink(:absolute)` flag.  But this is brand new functionality.  So I'd rather see the variable named `RAKU_TEST_NO_DIAG_ON_FAIL` and have it default to 0.

Should people think this is a bad idea, then I would suggest renaming it to `RAKU_TEST_DIAG_ON_FAIL`, as this is about suppressing all diagnostic output, not just the line number.

coke:
> FWIW, I don't really like the default to be 1. Sometimes we have to for consistency's sake (like recently with the `IO::Path.symlink(:absolute)` flag. But this is brand new functionality. So I'd rather see the variable named `RAKU_TEST_NO_DIAG_ON_FAIL` and have it default to 0.

OK. I can change the name and flip the sense of the test.

> Should people think this is a bad idea, then I would suggest renaming it to `RAKU_TEST_DIAG_ON_FAIL`, as this is about suppressing all diagnostic output, not just the line number.

I'm not trying to suppress all diag output. Just the file & line number when a test fails - so I'll keep "LINE" in the name of the var also.

lizmat:
Please don't change anything until more people have commented  :-)

lizmat:
Hmmm... no more comments..  what to do?

coke:
No feedback received, and it's been so long, I no longer need this change. Closing PR

PR #4000, opened 2020-11-01
Don't inherit mixin class language revision

### LINKED ISSUE #3998: use v6.e.PREVIEW does not allow mixin of custom roles
```
use v6.e.PREVIEW;
role R {};
my $a = Int but R;
```

fails with:
```
Type object Int+{R} of v6.c is not compatible with R of v6.e
  in block <unit> at /home/dex/tmp/tmp3.raku line 3
```

The expected output is none.

It would be nice to have this sorted before Christmas or testing modules before the release of v6.e might be difficult. I did not check if there are futher problems with precomp.

vrurg:
Use the compiler-default. This fixes #3998 and makes the following code
work:

    use v6.e.PREVIEW;
    role R { }; my $v = 0 but R;

Yet, `$v.^language-revision` will now return `e`, contrary to
`0.^language-revision` returning `c`.

### DIFF
--- a/src/Perl6/Metamodel/Mixins.nqp
+++ b/src/Perl6/Metamodel/Mixins.nqp
@@ -74,8 +74,6 @@ role Perl6::Metamodel::Mixins {
         # Create new type, derive it from ourself and then add
         # all the roles we're mixing it.
         my $new_type := self.new_type(:name($new_name), :repr($obj.REPR), :is_mixin);
-        $new_type.HOW.set_language_revision($new_type, $obj.HOW.language-revision($obj))
-            if $obj.HOW.is_composed($obj);
         $new_type.HOW.set_is_mixin($new_type);
         $new_type.HOW.add_parent($new_type, $obj.WHAT);
         for @roles {


lizmat:
This seems like a nice solution.  However, more generally, it feels that the way this is currently implemented needs a lot more thought.  A blog post with some more background would be greatly appreciated!

vrurg:
@lizmat Perhaps [the general description](https://vrurg.github.io/2019/12/16/new-role-of-roles-in-raku) of what's been done would be ok? Because writing about internal changes may turn into a big book with little value in it. :)

niner:
It occurs to me that 6.e roles are really only incompatible with 6.c classes 
in very specific circumstances. E.g. IIUC if the role doesn't contain a 
submethod, it can be composed into a 6.c class regardless of the role's 
language version. Can we restrict throwing of errors to those definitely 
incompatible cases?

vrurg:
@niner something like this I'm thinking about since the mixin issue arose. What concerns me most is switching from a generalized and easily managed rule of "no back compat over some language boundary" to a particular case. It's fine while the case is the only one, but what if there be more in time? We may end up with some kind of spaghetti code. So, I'm trying to figure out if there is a good way to get this done.

BTW, this is likely a task for the new dispatchers. Wish they be done eventually...

vrurg:
Meanwhile, with regard to this PR – I think it makes sense anyway, no matter if the specialized compatibility condition would be used or not.

Altai-man:
Sorry for the delay. I was asked to do a blin run of this at https://colabti.org/irclogger/irclogger_log/moarvm?date=2020-11-05#l156 but as usual I cannot go with a fork. Can this commit be submitted into a main repo branch so that relevant builds could be produced?

vrurg:
A blin run over this branch didn't show any breakages. Ready to merge.

Issue #4004, opened 2020-11-04
∈/(elem) given an an Int:D argument from sub MAIN and a Seq:D gives erroneous results


Kaiepi:
## The Problem
While toying with a script, I found that it wasn't behaving like I expected it to with the input I was giving. This is about as far as I can golf the problem:
```perl6
use v6;
unit sub MAIN(Int:D $x) {
    say $x ∈ [X+] (1..3) xx 2;
}
```

## Expected Behavior
This should output `True` given any possible sum of `1`, `2`, or `3` (e.g. `2`) as input.

## Actual Behavior
This outputs `False` given `2`.

The weird thing is this will work with other `Int:D`s:
```perl6
use v6;
unit sub MAIN(Int:D $x) {
    my Int:D $y = 2;
    say $y ∈ [X+] (1..3) xx 2;
}
```

But not `$x`. Not even clones of `$x`. Other ways of generating a new `Int:D` from `$x` with the same value will allow this to work though:

```perl6
use v6;
unit sub MAIN(Int:D $x) {
    say $x.succ.pred ∈ [X+] (1..3) xx 2;
}
```

## Steps to Reproduce

Save the test code as `test.raku` and run it with an integer as an argument.

## Environment
* Operating system:
OpenBSD bastille.kennel.qt 6.8 GENERIC.MP#116 amd64
* Compiler version (`perl6 -v` or `raku -v`):
Welcome to Rakudo(tm) v2020.10-51-g53b558fe0.
Implementing the Raku(tm) programming language v6.d.
Built on MoarVM version 2020.10-20-g70e3608cb.

AlexDaniel:
Int:D does not guarantee you an Int, as surprising as it sounds. In this case I think you're getting an IntStr.

taboege:
That statement is not what I would call surprising, because IntStr is a subclass of Int, so of course an Int constraint would accept it. By contrast, I would be surprised if, for the same reason, `Any:D` would refuse to accept anything whose type is more specific than `Any`.

Otherwise AlexDaniel is spot on. You get an IntStr and set semantics of (elem) rely on the value type behavior of IntStr, where `<2>.WHICH` is `IntStr|2`, distinct from `2.WHICH`. That's a hairy situation...

AlexDaniel:
[...]

PR #4005, opened 2020-11-04
Introduce %*SUB-MAIN-OPTS<coerce-allomorphs-to>

lizmat:
This new SUB-MAIN-OPTS setting allows one to specify how allomorphs
should be coerced **prior** to being dispatched.  So suppose you have
a sub MAIN:

    sub MAIN(Int:D $value) {
        say "value is a $value.^name()"
    }

calling such a script with "42" will show that the value is actually
an IntStr, **not** an Int.  Activating this new option with:

    my %*SUB-MAIN-OPTS = coerce-allomorphs-to => Int;

will ensure that all numeric values entered on the command will be
coerced to Ints.

The following type coercions are supported: Numeric, Int, Rat, Num,
Complex and Str.

Inspired by https://github.com/rakudo/rakudo/issues/4004

### DIFF
--- a/src/core.c/Main.pm6
+++ b/src/core.c/Main.pm6
@@ -54,45 +54,16 @@ my sub RUN-MAIN(&main, $mainline, :$in-as-argsfiles) {
         my &coercer = &val;
         if %sub-main-opts<coerce-allomorphs-to>:exists {
             my $type := %sub-main-opts<coerce-allomorphs-to><>;
-            if $type =:= Numeric {
+            if   $type =:= Numeric
+              || $type =:= Int
+              || $type =:= Rat
+              || $type =:= Num
+              || $type =:= Complex
+              || $type =:= Str {
+                my $method := $type.^name;
                 &coercer = -> \value {
                     (my \result := val(value)) ~~ Allomorph
-                      ?? result.Numeric
-                      !! result
-                }
-            }
-            elsif $type =:= Int {
-                &coercer = -> \value {
-                    (my \result := val(value)) ~~ Allomorph
-                      ?? result.Int
-                      !! result
-                }
-            }
-            elsif $type =:= Rat {
-                &coercer = -> \value {
-                    (my \result := val(value)) ~~ Allomorph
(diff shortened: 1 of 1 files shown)


src/core.c/Main.pm6
```diff
@@ -51,10 +51,60 @@ my sub RUN-MAIN(&main, $mainline, :$in-as-argsfiles) {
         my $positional := nqp::create(IterationBuffer);
         my %named;
 
+        my &coercer = &val;
+        if %sub-main-opts<coerce-allomorphs-to>:exists {
+            my $type := %sub-main-opts<coerce-allomorphs-to><>;
+            if $type =:= Numeric {
```
MasterDuke17, line 57: Since this probably isn't very hot code, why not:
```perl6
if $type =:= Numeric|Int|Rat|Num|Complex|Str {
    &coercer = -> \value {
        (my \result := val(value)) ~~ Allomorph
            ?? result.$type
            !! result
    }
} else {
```

lizmat: Sadly, that doesn't work for `Numeric`.  And the check for:

     $type =:= Int|Rat|Num|Complex|Str

doesn't work, because that would compare with the instantiated Junction, **not** with its constituents.

MasterDuke17: Huh, I hadn't realized that for `Numeric`. I guess you could do:
```perl6
if $type =:= Int or $type =:= Rat or $type =:= Num or $type =:= Complex or $type =:= Str {
    &coercer = -> \value {
        (my \result := val(value)) ~~ Allomorph
            ?? result.$type()
            !! result
    }
}
elsif $type =:= Numeric {
    &coercer = -> \value {
        (my \result := val(value)) ~~ Allomorph
            ?? result.Numeric
            !! result
    }
}
else {
```
but maybe it doesn't lose enough copy/pasting to make up for the extra complexity.

lizmat: Simplified further with https://github.com/rakudo/rakudo/commit/49eecd68c5

Issue #3996, opened 2020-10-31
Make the offered Rakudo (Windows) versions consistent


AntonOks:
## The Problem
Cannot automate Rakudo installation on Windows via [scoop ](https://scoop.sh/) and other package managers due to different Rakudo release information's on various places.

## Expected Behavior
Have the same Rakudo version information everywhere consistent. Especially the offered ZIP file and the included compressed Rakudo folder should have 100% the same logic / naming!

## Actual Behavior
The latest [release on GitHub](https://github.com/rakudo/rakudo/releases) says [2020.10](https://github.com/rakudo/rakudo/releases/tag/2020.10) but no corresponding Windows package (zip) is offered here.

There is NO version **_2020.10_** on the "[official download page](https://rakudo.org/downloads/rakudo)" (??!). Here we have a [2020.10**_-01_**](https://rakudo.org/dl/rakudo/rakudo-moar-2020.10-01-win-x86_64.zip) ZIP file, **_BUT_** within this ZIP file the directory, which should represent the Rakudo version is _2020.01_ instead of 2020.10**_-01_** (aka same as the Rakudo version).

## Steps to Reproduce

- Download https://rakudo.org/dl/rakudo/rakudo-moar-2020.10**_-01_**-win-x86_64.zip
- Extract it to C:\Temp\rakudo-moar and look into this directory. You will see "rakudo-**_2020.10_**" there instead of (expected related to the offered ZIP file name!) **_rakudo-2020.10-01_**

## Environment
* Operating system: Windows but this doesn't play a role ;)
* Compiler version (`perl6 -v` or `raku -v`): 2020.10 ZIP file ;)

AntonOks:
@patrickbkr 

Seems like the "problem" is https://github.com/rakudo/rakudo/blob/0a4eaf3110e8fe52f33f0f43cbd3d78908c2138c/tools/build/binary-release/build-windows.ps1#L67

Guess [$Env:REVISION](https://github.com/rakudo/rakudo/blob/0a4eaf3110e8fe52f33f0f43cbd3d78908c2138c/azure-pipelines.yml#L12) would also be known in this script so this line should be `mv install rakudo-$Env:VERSION-$Env:REVISION`... and L68 needs then also to reflect this change afterwards ;)

patrickbkr:
[...]

PR #4006, opened 2020-11-05
make the offered rakudo-moar-$VERSION-$REVISION-win-x86_64.zip file consistent with the included compressed directory name

AntonOks:
see #3996 

the finally generated  "rakudo-win.zip" file here is later renamed in [azure-pipelines.yml](https://github.com/rakudo/rakudo/blob/68196c99dfccbeb708851fdc1592d026bece3042/azure-pipelines.yml#L324) to reflect the $VERSION and $REVISION... so it is more consistent to have everything aligned to the same structure / naming

### DIFF
--- a/tools/build/binary-release/build-windows.ps1
+++ b/tools/build/binary-release/build-windows.ps1
@@ -64,6 +64,7 @@ cp -Force -r "tools\build\binary-release\Windows\*" install
 cp LICENSE install
 
 echo "========= Preparing archive"
-mv install rakudo-$Env:VERSION
-Compress-Archive -Path rakudo-$Env:VERSION -DestinationPath ..\rakudo-win.zip
+rakudoZipDirectory = "rakudo-moar-" + $Env:VERSION + "-" + $Env:REVISION
+mv install $rakudoZipDirectory
+Compress-Archive -Path $rakudoZipDirectory -DestinationPath ..\rakudo-win.zip
 


patrickbkr:
I hope to find some time to look into this tomorrow morning.

AntonOks:
Hi @patrickbkr Have you had a chance for a quick check? Any comments or feedback? Will this work ok from your point of view and could it be merged soon?

Thanks a lot!

@Altai-man

patrickbkr:
@AntonOks I think we need more discussion on this (over in the issue).
I suspect the issue is mostly about personal preference and less about one solution being objectively better than the others. So I'd prefer not to change this if no general preference for a different naming emerges. So at the moment this more or less blocks on feedback of other devs.

With respect to this PR more work is needed. If the goal is to unify the names of the archive and folder, then we should either name the folder just "rakudo" or give it the full name of the archive (this PR currently does something different). If we do the latter, we should also add the compiler tool chain (gcc, clang or msvc) to the name of the archive first, as that change is bound to happen sooner or later. Also we need to change the name not only in the script for Windows, but also in the ones for MacOS and Linux.

But for now I would just wait a bit longer until more feedback comes in and then decide on how we want to proceed.

But still: Thanks for your enthusiasm, work and willingness to push this forward!

Altai-man:
@patrickbkr note this is a (I guess) release blocker, so if some devs in particular have to be kind enough to give this some attention, we should ping them.

patrickbkr:
@AntonOks Given there now is a decision in #3996, do you want to update this PR accordingly or should I give it a try?
The following things need to be done:

- Make sure the folder gets the full name, identical to the achive name itself, with the addition of the toolchain.
- Update the scripts for MacOS and Linux as well. They are located in the same folder as the Windows build script.
- Adapt [azure-pipelines.yml](https://github.com/rakudo/rakudo/blob/master/azure-pipelines.yml) to name the archive files themselves correctly.

AntonOks:
> @AntonOks Given there now is a decision in #3996, do you want to update this PR accordingly or should I give it a try?
> The following things need to be done:
> 
>     * Make sure the folder gets the full name, identical to the achive name itself, with the addition of the toolchain.
>     * Update the scripts for MacOS and Linux as well. They are located in the same folder as the Windows build script.
>     * Adapt [azure-pipelines.yml](https://github.com/rakudo/rakudo/blob/master/azure-pipelines.yml) to name the archive files themselves correctly.

@patrickbkr 
Just started to look around and then found out [the new naming is already in place](https://rakudo.org/downloads/rakudo) . Wow... big THANKS to all involved!

patrickbkr:
@AntonOks The binary builds are currently done semi automatically by me (using the scripts this PR is about). Using the new naming was only a small manual tweak after the automated build went through. (Plus a [change](https://github.com/rakudo/rakudo.org/pull/55) to the rakudo.org website to provide access to the toolchain bit in its API.)

So the tasks I listed in a previous comment are still to be done. Do you want to give it a try or should I do it?

AntonOks:
> So the tasks I listed in a previous comment are still to be done. Do you want to give it a try or should I do it?

@patrickbkr 

if it's a nobrainer for you, pls do it. THANKS!

I had a look and already plenty questions, i.e. is "toolchain" really hard-coded and how and where to make it a environment variable? If at all.. See, I would need help anyhow and it for sure would take some weeks...

patrickbkr:
This PR is superseded by #4071. Closing.
@AntonOks Thanks for your work on this!

PR #4013, opened 2020-11-08
Expose Hash::Object/Hash::Typed roles

lizmat:
These roles used to be internal to the Hash class.  By exposing them
it is possible to use them for easy typechecking and dispatch.

This makes the rather esoteric check:

    nqp::eqaddr(foo.keyof,Str(Any))

much more readable as:

    nqp::istype(foo,Hash::Object)

albeit with inverted meaning, of course,  This check is about 2x as
fast as the nqp::eqaddr version.  Still it is little work, but some
of these happen in potentially hot code paths in set operator handling.

For easier maintainability, these roles have been put into separate
source-files.

This commit also changes all places where the faster Hash::Object check
could be used.


MasterDuke17:
> One could possibly bikeshed that these roles should be called
> Hash::ObjectHash and Hash::TypedHash respectively.

Or just Hash::Object and Hash::Typed? No strong feelings, just throwing things out there.

vrurg:
> Or just Hash::Object and Hash::Typed? No strong feelings, just throwing things out there.

So, I'm not the only one out there... :) Not that it really matters a lot as the core consist of mostly single-word names.

lizmat:
If `Hash::Typed` and `Hash::Object` are the consensus, then that's what they will be :-)

PR #4017, opened 2020-11-10
Fix issue preventing CLI handler MAIN from having a type constraint

mienaikage:
Rakudo 2020.10 has an issue preventing MAIN subs from having a type constraint:

```
raku -e 'sub MAIN (--> Nil) {}'
MAIN must be a 'sub' to allow it to be called as a CLI handler
  in block <unit> at -e line 1
```

Rakudo 2020.09:
```
raku -e 'sub MAIN (--> Nil) {}; &MAIN.^name.say'
Sub+{Callable[Nil]}
```

### DIFF
--- a/src/core.c/Main.pm6
+++ b/src/core.c/Main.pm6
@@ -288,7 +288,7 @@ my sub RUN-MAIN(&main, $mainline, :$in-as-argsfiles) {
     }
 
     sub find-candidates($capture) {
-        &main.^name eq 'Sub'
+        &main.^name ~~ /^ Sub [ '+{Callable[' .*? ']}' ]? $/
           ?? &main
                # Get a list of candidates that match according to the dispatcher
                .cando($capture)


src/core.c/Main.pm6
```diff
@@ -288,7 +288,7 @@ my sub RUN-MAIN(&main, $mainline, :$in-as-argsfiles) {
     }
 
     sub find-candidates($capture) {
-        &main.^name eq 'Sub'
+        &main.^name ~~ /^ Sub [ '+{Callable[' .*? ']}' ]? $/
```
niner, line 291: I wonder why we test for a class name here instead of doing a normal type check? The latter wouldn't be affected by mixins.

mienaikage: I did throw this together quite quickly last night, I'll take a closer look later today to see how this can be improved.

lizmat:
The better check was implemented with e3352cdd82  Thanks for spotting and trying to fix!

PR #4018, opened 2020-11-10
Allow new versioning values

lizmat:
This commmit allows for several version specification alternatives
*without* changing anything else.  It would merely allow people to
refer to the new version values *now*.  Allowed alternatives are:

  v2015          -> v6.c
  v2018          -> v6.d
  v2021.PREVIEW  -> v6.e.PREVIEW

This also simmplifies the version checking logic for the most common
cases.  This appears to have a beneficial effect on the duration of
a spectest run.

### DIFF
--- a/src/Perl6/Grammar.nqp
+++ b/src/Perl6/Grammar.nqp
@@ -864,7 +864,7 @@ grammar Perl6::Grammar is HLL::Grammar does STD {
           # we parse out the numeral, since we could have "6d"
           :my $version := nqp::radix(10,$<version><vnum>[0],0,0)[0];
           [
-          ||  <?{ $version == 6 }> { $*W.load-lang-ver: $<version>, $comp }
+          ||  <?{ $version == 6 || $version == 2015 || $version == 2018 || $version == 2021 }> { $*W.load-lang-ver: $<version>, $comp }
           ||  { $/.typed_panic: 'X::Language::Unsupported',
                   version => ~$<version> }
           ]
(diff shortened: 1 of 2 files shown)


vrurg:
First of all, I think we have to get it resolved in problem-solving. We're still in the limbo with regard to wether we gonna stick to `6.x` scheme, or move one to something different.

Then, the `if-elseif` chain looks wrong to me anyway. If there be any kind of aliasing to the existing letter-based revisioning, then it's better done via `tools/templates/RAKU_SPEC` with appropriate support from build tooling, apparently.

lizmat:
Ok, I was living under the impression that going for a year based language versioning, was sorta decided already.

In that light I wanted to make it possible for people to start using the `use v2015` and `use v2018` and `use v2021.PREVIEW` version indicators as early as possible.

I was not aware of the existence of `tools/templates/RAKU_SPECS`.  It however feels to me that the current versioning speccing feels over-engineered for something that will only change roughly once every 3 years.  And the number of allowed `use vxxx` strings is actually pretty limited.  Or am I missing something essential here?

if we want to do a problem-solving round, then please let us do that.

vrurg:
I don't think we need another round as long as there is raku/problem-solving#141. My point for now is that we release 6.e and then re-vivify the versioning discussion. There is no more pressure from Perl-naming anyway.

With regard to `RAKU_SPEC`, I came up with the idea when got bored of repetitive version-dependent code here and there. With this file adding a new version, or changing status of an old version, is mostly a matter of changing a line in that file. Plus I like the formalization of modifiers like `PREVIEW`.

lizmat:
> There is no more pressure from Perl-naming anyway.

I think it is *very* relevant from a PR point of view.  It's not for nothing that Perl 4 is basically Perl 3 but bumped to look better for the O'Reilly books.

So I have a strong feeling that the next language level should **NOT** be called `v6.e`.  Whether it should be `v21` or `v2021`, I think I've made my point clear before (that it should be `v2021`.

PR #4026, opened 2020-11-13
Gen faster code for some kinds of `when` cases

MasterDuke17:
If it's a type or literal, we don't need to do a full smartmatch.
Unfortunately this isn't dramatically faster, since the primary cost is
in generating a Backtrace for the Failure created when a case doesn't
match.

An example test case:
```raku
my $a;
my @a = 1, 5, Int, "hi", "bye", 50, Complex;
my $s = now;
for ^100_000 -> $i {
    $a = do given @a[$i % 7] {
        when Int  { "Int" }
        when 5    { "five" }
        when "hi" { "hello" }
        default   { "default" }
    }
};
say now - $s;
say $a
```
takes ~1.4s with this change, ~1.8s before.

### DIFF
--- a/src/core.c/Str.pm6
+++ b/src/core.c/Str.pm6
@@ -772,7 +772,7 @@ my class Str does Stringy { # declared in BOOTSTRAP
          )
         )
     }
-    multi method Numeric(Str:D: --> Numeric:D) {
+    multi method Numeric(Str:D: Bool :$fail-or-nil --> Numeric:D) {
 #?if !jvm
         # check for any combining characters
         nqp::isne_i(nqp::chars(self),nqp::codes(self))
@@ -800,7 +800,7 @@ my class Str does Stringy { # declared in BOOTSTRAP
                  self,0,nqp::chars(self)
                ) == nqp::chars(self)
               ?? 0                                # just spaces
-              !! val(self, :val-or-fail)          # take the slow route
+              !! val(self, :val-or-fail, :$fail-or-nil)          # take the slow route
     }
 
     multi method gist(Str:D:) { self }
--- a/src/Perl6/Actions.nqp
+++ b/src/Perl6/Actions.nqp
@@ -2339,16 +2339,30 @@ class Perl6::Actions is HLL::Actions does STDActions {
                 die("Unknown QAST::Want type " ~ $sm_exp[2].HOW.name($sm_exp[2]));
             }
 
+            # Coercion method to call on the given value
+            my $method_call :=
+                QAST::Op.new( :op('callmethod'), :name($method),
+                  QAST::Var.new( :name('$_'), :scope('lexical') ) );
+
+            # We don't need/want `val()` to `fail()` if `Numeric()` ends up calling it
+            # and it doesn't succeed, that creates an expensive Backtrace that we just
+            # end up throwing away
+            if $method eq 'Numeric' {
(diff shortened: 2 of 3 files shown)


src/core.c/Str.pm6
```diff
@@ -772,7 +772,7 @@ my class Str does Stringy { # declared in BOOTSTRAP
          )
         )
     }
-    multi method Numeric(Str:D: --> Numeric:D) {
+    multi method Numeric(Str:D: :$fail-or-nil --> Numeric:D) {
```
timo, line 775: should the `$fail-or-nil` parameter be typed Bool here and in the val sub?

MasterDuke17: Force pushed that change.

MasterDuke17:
FYI, this code:
```raku
class MyInt is Int {
	method Numeric {
		"MyInt == " ~ (self + 1).Str
	}
}

my MyInt $a .= new(3);

given $a {
	when 3 {
		say "three"
	}
	when 4 {
		say "four"
	}
	default {
		say "default"
	}
}
```
dies with `Cannot convert string to number: base-10 number must begin with valid digits or '.' in '⏏MyInt == 4' (indicated by ⏏)
  in block <unit> at -e line 1` on master, but prints 'default' on this branch. The current behavior isn't specced (i.e., this PR passes `make m-test m-spectest`), but should it be?

MasterDuke17:
This code:
```raku
class MyStr is Str {
	method Numeric {
		self.chars
	}
}

my MyStr $a .= new(:value("abc"));

given $a {
	when 3 {
		say "three"
	}
	when 4 {
		say "four"
	}
	default {
		say "default"
	}
}
```
prints 'three' both on master and this branch.

lizmat:
The additional `:fail-or-nil` solution feels like a hack upon a hack, and I wonder what effects that will have on the performance of `val`.  There's something really broken on the current implementation of `val`, which I tried to fix before with a total rewrite.  That failed, because there was no way to assemble a `num` from two strings in the same way that `nqp::numify` does.  The rewrite did *not* have the "create a `Failure` when it is not needed" issue that you're trying to circumvent here.

I would love to resuscitate that effort, but that stands or falls with being able to create a `num` from two `str` (before and after decimal point) in **exactly** the same way that `nqp::numify` does.  That would *really* help improve this.

MasterDuke17:
> The additional `:fail-or-nil` solution feels like a hack upon a hack, and I wonder what effects that will have on the performance of `val`.

Agree re hack, but I'm not sure it'll effect `val` performance. I didn't benchmark it directly, but did try a micro-benchmark of adding a second named parameter to a sub/method and didn't notice any slowdown.

> I would love to resuscitate that effort, but that stands or falls with being able to create a `num` from two `str` (before and after decimal point) in **exactly** the same way that `nqp::numify` does. That would _really_ help improve this.

I'm not sure I follow. You need/want to implement the same logic, but not use `nqp::numify`?

lizmat:
@MasterDuke17 because in the code I wrote, it would have 2 integer values.  The whole performance enhancement was based on not having to keep pointers.  Alas, given two integer values, I was unable to get the same result as what nqp::numify did.  And thus borked a number of pesky tests in roast.  And thus invalidated the whole approach.  :-(

lizmat:
Thanks!

Issue #4027, opened 2020-11-14
Flapper: S32-io/IO-Socket-Async.t


Kaiepi:
## The Problem
`S32-io/IO-Socket-Async.t` in Roast failed when I ran `make spectest` on a new fork of master:

```
Test Summary Report
-------------------
t/spec/S32-io/IO-Socket-Async.t                                 (Wstat: 256 Tests: 26 Failed: 0)
  Non-zero exit status: 1
  Parse errors: Bad plan.  You planned 40 tests but ran 26.
```

But passes when the file's tests are run individually.

## Environment
* Operating system:
OpenBSD bastille.kennel.qt 6.8 GENERIC.MP#116 amd64
* Compiler version (`perl6 -v` or `raku -v`):
Welcome to Rakudo(tm) v2020.10-65-g082763e75.
Implementing the Raku(tm) programming language v6.d.
Built on MoarVM version 2020.10-31-gb13f6a24c.

PR #4028, opened 2020-11-14
Throw an exception if &make is called w/o a valid match in $/

Kaiepi:
This adds a new `X::MakeWithInvalidMatch` exception, which is thrown by
`&make` when `$/` is not an `NQPMatchRole`. This fixes a LTA error when
`&make` is called, but `$/` hasn't been given a value.

Passes `make test` and `make spectest` (outside of https://github.com/rakudo/rakudo/issues/4027).

### DIFF
--- a/src/core.c/Match.pm6
+++ b/src/core.c/Match.pm6
@@ -1,3 +1,13 @@
+my class X::MakeWithInvalidMatch is Exception {
+    has Mu:_ $.match is built(:bind) is required;
+
+    method message(::?CLASS:D: --> Str:D) {
+        "&make called with an invalid match in lexical variable \$\/ ({
+            Rakudo::Internals.SHORT-STRING: $!match, :method<raku>
+        }).{' Is there any in scope?' if $!match =:= Nil}"
+    }
+}
+
 my class Match is Capture is Cool does NQPMatchRole {
 # from NQPMatchRole
 #    has int $!from;  # start position of match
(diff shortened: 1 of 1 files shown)


src/core.c/Match.pm6
```diff
@@ -322,10 +332,11 @@ multi sub infix:<eqv>(Match:D \a, Match:D \b) {
     );
 }
 
-
 sub make(Mu \made) {
-    nqp::bindattr(nqp::decont(nqp::getlexcaller('$/')),Match,'$!made',made)
+    nqp::if(
+      nqp::istype((my Mu:_ $match := nqp::decont(nqp::getlexcaller('$/'))),NQPMatchRole),
+      nqp::bindattr($match,Match,'$!made',made),
```
Kaiepi, line 338: This seems to work with slangs... but how?

lizmat:
Do you have any idea on performance issues of this change?  `make` is called an awful lot when e.g. compiling the setting.

Kaiepi:
Compiling the setting takes around the same amount of time with or without this for me, but in a benchmark, `&make` with a `Match` in `$/` runs around 10% slower than it did before. `NQPClassHOW` and `Metamodel::GrammarHOW` would have `NQPMatchRole` in their typecheck cache, so I think this checks out.

lizmat:
To be honest, I'm a bit reluctant to merge this because:

a. grammars are slow as it is, so making them slower again feels wrong
b. this seems inspired by a flapper causing `$/` to become undefined when it shouldn't

So the real life case for this error to appear to be very limited.  When you're setting `$/` yourself, then you either know what you're doing, or it is a case of DIWHIDT.

So I'm leaning towards -1 for this PR.  @jnthn maybe has other opinions.  If he doesn't merge in the next few days, I will close this PR.

PR #4029, opened 2020-11-14
Improve output of Attribute with Pod::To::Text

mienaikage:
Attributes with declarator blocks currently do not output correctly with `--doc`.

Before:
```
class Foo
This is the class

class Attribute.new
This is the attribute
```

After:
```
class Foo
This is the class

attribute Mu $!bar
This is the attribute
```


lizmat:
Thanks!

PR #4042, opened 2020-11-17
Make cas($target, &code) ~17% faster

dumarchie:


lizmat:
Wow, that surprises me.  Do you have a benchmark that I could use to verify?

niner:
I guess the speedup comes from not using `return`. Have you tried a similarly structured version in pure Raku for comparison? Maybe we can get the speedup while retaining the readability.

lizmat:
Argh, I read that backwards  :-(   Definitely *not* used to seeing other people optimizing hot code with nqp::ops  :-)

@dumarchie:  I wonder if this could be faster still with a `nqp::repeat_until` ?

dumarchie:
I threw the benchmark away, but I just took the time taken by `cas($scalar, &code)` minus the time taken by `$scalar = code($scalar)`.

I also think it was mainly the `return` that made it slow. I did not try a pure Raku restructuring because I could not think of a nice one and I assumed that an NQP loop would have less overhead anyway.

Sorry @lizmat, I don't see how a repeat until would improve the implementation.

dumarchie:
I wondered how much difference there would be on a MinGW build, so I recreated the benchmark script:

```
my $n      = 1_000_000;
my &code   = { $_ + 1 };

my $scalar = 0;
my $start  = now;
$scalar    = code($scalar) for 1..$n;
printf "assign:       %.3f\n", my $t1 = now  - $start;

$scalar = 0;
$start = now;
cas($scalar, &code) for 1.. $n;
printf "cas:          %.3f\n", my $t2 = now - $start;

printf "cas overhead: %.3f\n", $t2 - $t1;
```

These are runs 2..4 on a fresh install *without* the change:

```
c:\raku>rakudo\install\bin\raku.exe cas.raku
assign:       0.165
cas:          1.851
cas overhead: 1.686

c:\raku>rakudo\install\bin\raku.exe cas.raku
assign:       0.168
cas:          2.000
cas overhead: 1.833

c:\raku>rakudo\install\bin\raku.exe cas.raku
assign:       0.151
cas:          1.978
cas overhead: 1.827
```

These are runs 2..4 on a fresh install *with* the change:

```
c:\raku>rakudo\install\bin\raku.exe cas.raku
assign:       0.166
cas:          1.629
cas overhead: 1.463

c:\raku>rakudo\install\bin\raku.exe cas.raku
assign:       0.158
cas:          1.648
cas overhead: 1.490

c:\raku>rakudo\install\bin\raku.exe cas.raku
assign:       0.167
cas:          1.642
cas overhead: 1.475
```

So on average the change makes the cas overhead around 17% less.

How much difference do you see on your system, @lizmat?

lizmat:
Thanks!

Issue #4040, opened 2020-11-17
[JVM] Failing test in S12-coercion/coercion-types.t after switch to new coercion semantics


usev6:
## The Problem
After the merge of https://github.com/rakudo/rakudo/pull/3891 a test in S12-coercion/coercion-types.t started to fail: https://github.com/Raku/roast/blob/861bc6c8d4/S12-coercion/coercion-types.t#L77-L84

The problem is not specific to arrays, but to assignment:
```
$ ./rakudo-j -e 'my Rat(Str) $foo = 42'
Type check failed in assignment to $foobar; expected Rat(Str) but got Int (42)
  in block <unit> at -e line 1
```

I suspect the problem to be in https://github.com/rakudo/rakudo/blob/3e10cc6f8a/src/Perl6/bootstrap.c/BOOTSTRAP.nqp#L1634. The call to ```.coerce``` two lines below is not reached, because the conditional evaluates to false.

[Aside: If I'm not mistaken, MoarVM takes a different path and doesn't hit this code (for this example) -- probably because assignment is handled differently.]

## Environment
* Operating system: FreeBSD 11.3-RELEASE-p6
* Compiler version (`perl6 -v` or `raku -v`): Welcome to Rakudo(tm) v2020.10-104-g1f090e04d.
Implementing the Raku(tm) programming language v6.d.
Built on JVM.

Altai-man:
Can this be fixed before release this weekend?

usev6:
@Altai-man I hope @vrurg will find a way to fix it. But I wouldn't consider this a blocker -- given that the JVM backend isn't fully functional.

vrurg:
There is no need to fix this because JVM behaves correctly in this case. It is a bug in MoarVM, actually because `Int !~~ Rat(Str)`. There is, actually, a proof for this: `my Rat $foo = 42`. The question is how far this erroneous behavior stretches back.

vrurg:
One more check which do the right thing:

```raku
sub foo(Rat(Str) $v) {
    note $v.WHICH;
}
foo(1);
```

So, I wonder where exactly things go wrong with the assignment... O-o-okay...

PR #4043, opened 2020-11-17
Fix incorrect assignment passing with coercion

vrurg:
`my Rat(Str) $foo = 1` will now fail because `Int` is not
passing the type check.

Also made `Metamodel::CoercionHOW::coerce()` more strict about the value
it's trying to coerce. Now it will throw if it doesn't match against
`$!constraint_type`. Normally this shouldn't happen to a user unless it
invokes the method directly, but should be of great help in catching
core errors.


vrurg:
Closes #4040

melezhik:
I put RP test for that -  http://rakudist.raku.org/sparky/report/RakuPlay-3/1329

Issue #4041, opened 2020-11-17
[JVM] Failing tests in S06-currying/positional.t after switch to new coercion semantics


usev6:
## The Problem
After the merge of #3891 some tests in S06-currying/positional.t started to fail: https://github.com/Raku/roast/blob/861bc6c8d4/S06-currying/positional.t#L162-L166

A somewhat shortened code snippet from the first failing test:
```
$ ./rakudo-j --ll-exception -e 'my Int @AoI = 1,2; my $foo := sub (Str(Int) @a, Str(Int) @b) { }; dd $foo.assuming($@AoI)'
Method 'coerce' not found for invocant of class 'Perl6::Metamodel::CurriedRoleHOW'
  in trybind (EVAL_1:1)
  in <unit> (EVAL_1:1)
  in <unit-outer> (EVAL_1:1)
  in <anon> (gen/jvm/CORE.c.setting:57381)
  in EVAL (gen/jvm/CORE.c.setting:57374)
  in assuming (gen/jvm/CORE.c.setting:16397)
  in <unit> (-e:1)
  in <unit-outer> (-e:1)
  in eval (gen/jvm/stage2/NQPHLL.nqp:1216)
  in <anon> (gen/jvm/stage2/NQPHLL.nqp:1328)
  in command_eval (gen/jvm/stage2/NQPHLL.nqp:1325)
  in command_eval (gen/jvm/Compiler.nqp:116)
  in command_line (gen/jvm/stage2/NQPHLL.nqp:1309)
  in MAIN (gen/jvm/rakudo.nqp:133)
  in <mainline> (gen/jvm/rakudo.nqp:121)
  in <anon> (gen/jvm/rakudo.nqp)
```

## Environment
* Operating system: FreeBSD 11.3-RELEASE-p6
* Compiler version (`perl6 -v` or `raku -v`): Welcome to Rakudo(tm) v2020.10-104-g1f090e04d.
Implementing the Raku(tm) programming language v6.d.
Built on JVM.

Altai-man:
Can this be fixed before release this weekend? [2]

usev6:
It looks like the ```.assuming``` can be golfed away:
```
$ ./rakudo-j --ll-exception -e 'my Int @AoI = 1,2; sub (Str(Int) @a) { 42 }($@AoI)'
Method 'coerce' not found for invocant of class 'Perl6::Metamodel::CurriedRoleHOW'
  in <anon> (-e:1)
[...]

PR #4044, opened 2020-11-18
[JVM] Use archetypes to figure out coerciveness

usev6:
That matches the way .coercive is used in many other places.
Fixes https://github.com/rakudo/rakudo/issues/4041


vrurg (approved):

vrurg:
@usev6 what I was wondering about is why don't we use same nqp binder for JVM? Performance?

usev6:
@vrurg Sorry, I absolutely don't have an idea about that. I always just assumed there is a bootstrap problem or some such. @jnthn will know ;)

PR #4047, opened 2020-11-19
Type IO::Socket::INET family/type/protocol values

Kaiepi:
This changes the type of `IO::Socket::INET.new`'s `:$family` parameter from
`Int` to `ProtocolFamily:D(Int:D)` (removing their invalid `Int:U` typing) and
gives the class' `$!family`/`$!type`/`$!proto` attributes explicit
`ProtocolFamily:D`/`SocketType:D`/`ProtocolType:D` typings (meaning they no
longer accept raw `Int:D` values corresponding to those types' enum
values).

This was originally a change I had planned for the v6.e changeset of my solution for https://github.com/Raku/problem-solving/111, but is doable in v6.c now thanks to the new coercion API (@vrurg++). This doesn't appear to have a significant impact on the performance of `IO::Socket::INET`.

### DIFF
--- a/src/core.c/IO/Socket/INET.pm6
+++ b/src/core.c/IO/Socket/INET.pm6
@@ -1,17 +1,12 @@
 my class IO::Socket::INET does IO::Socket {
     my module PIO {
-        constant SOCK_PACKET    = 0;
-        constant SOCK_STREAM    = 1;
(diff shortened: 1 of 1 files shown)


lizmat:
Thanks!

PR #4049, opened 2020-11-19
[JVM] Test if method 'coercive' is available

usev6:
A test in MISC/bug-coverage-stress.t was failing with

  Method 'coercive' not found for invocant of class 'Archetypes'

A shortened version of the failing code (no module loading
involved) is:

  my role Bar {
      method compose(Mu \type) { callsame }
  }
  multi trait_mod:<is>(Attribute:D $attr, :$mytrait!) {
      $attr.package.HOW does Bar
  }
  my class Foo {
      has $!bar is mytrait
  }

There is a similar check in src/Perl6/Action.nqp, where we run

  nqp::can($ptype_archetypes, 'coercive')

So I guess it makes sense to verify that the method is available.


vrurg (approved):

vrurg:
Somehow (and it's not bad) it stumbled upon a NQP class. The check certainly makes sense because if `Archetypes` doesn't know about coerciveness – we shouldn't bother trying anything related.

Thanks for catching this one!

PR #4050, opened 2020-11-19
Optimize some array operators

dumarchie:
Raw slurpies are more efficient.


lizmat:
Cool!  Do you have any idea how much improvement this gives?

lizmat (approved):

dumarchie:
In this case I didn't bother to benchmark. These candidates don't appear to be used much, although `push(%h, ...)` and `append(%h, ...)` can be really helpful because they throw if the programmer accidentally provides a named argument instead of a `Pair`.

lizmat:
I'll wait with merging this *after* the 2020.11 release, if that's ok with you.  Just to be on the safe side.

Issue #2446, opened 2018-10-29
LTA error when making a subset out of a coercer


zoffixznet:
Was wondering if it would work and got a bad error:

```
+Zoffix │ m: subset Foo of Int(); -> Foo {}(42)
+camelia │ rakudo-moar 266af373d: OUTPUT: «===SORRY!===␤Cannot find method 'nominalize' on object of type Perl6::Metamodel::CoercionHOW␤»
```

MasterDuke17:
Fwiw, this patch:
```
diff --git a/src/Perl6/Metamodel/CoercionHOW.nqp b/src/Perl6/Metamodel/CoercionHOW.nqp
index fe6ebb9ff..f98fc24ba 100644
--- a/src/Perl6/Metamodel/CoercionHOW.nqp
+++ b/src/Perl6/Metamodel/CoercionHOW.nqp
@@ -54,6 +54,14 @@ class Perl6::Metamodel::CoercionHOW
         check_instantiated($coercion_type);
         nqp::typeparameterat($coercion_type, 1)
     }
+
+    method nominalize($obj) {
+        my $constraint_type := $obj.HOW.constraint_type($obj);
+        $constraint_type.HOW.archetypes.nominal ??
+            $constraint_type !!
+            $constraint_type.HOW.nominalize($constraint_type)
+    }
 }
 BEGIN {
     my $root := nqp::newtype(Perl6::Metamodel::CoercionHOW, 'Uninstantiable');
```
causes the error to change to:
```
Constraint type check failed in binding to parameter '<anon>'; expected Foo but got Int (42)
  in block <unit> at -e line 1
```

vrurg:
Fixed with #4058

PR #4058, opened 2020-11-25
Implement metmodel transparency of nominalizables

vrurg:
As a matter of fact, a nominalizable is a wrapper around a nominal or another nominalizable. As such, it makes sense to have their MOP classes transparent, similar to how their typeobjects are transparent down to the final nominal type. I.e., it means that for

    role R[::T] {
        my subset RS of T;
        method foo(RS() $v) {...}
    }

`RS().HOW.archetypes.generic` should be true at the compile time because `RS` wraps a generic. Similarly, `RS` would have it's `coercive` archetype set would it be declared as `my subset RS of T()`. The transparency lets the compiler and runtime to make sure no nominalization is missed no matter how deep they're nested into each other. In particular, it makes the sample above possible in theory. Though in practice implementing it is yet too complicated task impractical in the light of the upcoming RakuAST implementation.

A new metmodel role `Perl6::Metamodel::Nominalizable` is introduced which provides basic common functionality of nominalizables. In particular, it implements `.^wrappee` method which allows to fetch necessary nominalization from the nesting. `RS.^wrappee(:coercion)` would return `T()` for the latter `RS` example. Correspondingly, `RS().^wrappee(:subset)` would result in `RS` itself.

Implemented `instantiate_generic` for `Metamodel::SubsetHOW`.

Implemented partial support of the new transparency in `Actions.nqp`.

Improved the performance of the new binding code by replacing a call to `coercive` method on a parameter object with `nqp::getattr` and `nqp::bitand_i` pair. For now this is all I can squeeze of it until `nqp::curcode` can be made inlinable.


vrurg:
Yet, this draft currently passes spectests.

Altai-man:
Given the current situation we likely want to get some design-wise & performance-wise review before merging.

vrurg:
It turns out that supporting generics as subset refinees is way more complicated than expected if we aim at proper support of them everywhere. `where` blocks of a lexically scoped subset within a role declaration seemingly share same closure, to begin with. It results in the generic type used in such a `where` block resolving to a wrong typeobject. Usually to the one used with the latest role specialization.

Hopefully, this would be easier to implement with RakuAST some day.

vrurg:
Fixes #2446 anyway.

PR #4064, opened 2020-11-28
Fix unwanted references to other compilations by CompilerServices

niner:
CompilerServices used several lexical variables to store often used supposedly
constant QAST nodes and hashes. Turns out these nodes do get modified, i.e.
references get added or maybe it's just that they get added to a serialization
context. Anyway turning them into attributes and initializing them newly for
every compilation fixes issues with references to outer compiler's SCs.

This not only affects cases where multiple comp units are precompiled in the
same process but also cases where multiple compilations run in parallel.


lizmat:
WOW!

lizmat (approved):

PR #4065, opened 2020-11-28
Give stub packages created by package_at_key a proper longname

niner:
Stub packages created by simply accessing them like &Foo::Bar::note = sub { };
only got the last part of their name ("Bar"). Now they know who they really are
(Foo::Bar).


vrurg:
Any reason not to merge this yet?

PR #4072, opened 2020-11-29
Make Num coercer demand definite invocant

### LINKED ISSUE #1517: Type coercion turns optional named argument into required one substantially
## The Problem
Type coercion (e.g., `Num(Rat)`) turns optional named argument into required one substantially.
Even if explicitly specify `?` to a named argument, it requires to specify the named argument:
```
$ perl6 -e 'sub func(Num(Rat) :$v?) { say $v if $v.defined; }; func();'
Cannot look up attributes in a Rat type object
  in sub func at -e line 1
  in block <unit> at -e line 1
```

## Expected Behavior
Even if the named argument isn't given, it doesn't cause an error because the named argument is optional.

## Actual Behavior
It cause an error.

## Steps to Reproduce

```
$ perl6 -e 'sub func(Num(Rat) :$v?) { say $v if $v.defined; }; func();'
Cannot look up attributes in a Rat type object
  in sub func at -e line 1
  in block <unit> at -e line 1
```

## Environment
* Operating system: Debian stretch
```
$ perl6 --version
This is Rakudo version 2017.12-211-g029226fe9 built on MoarVM version 2017.12.1-30-ged7c1234f
implementing Perl 6.c.
```

vrurg:
This changes the error message reported back to user from a cryptic 

```
Cannot look up attributes in a <...> type object
```

to more correct 

```
Invocant of method 'Num' must be an object instance of type 'Rational',
not a type object of type 'Rat'.  Did you forget a '.new'?
```

Fixes #1517

### DIFF
--- a/src/core.c/Rational.pm6
+++ b/src/core.c/Rational.pm6
@@ -65,7 +65,7 @@ my role Rational[::NuT = Int, ::DeT = ::("NuT")] does Real {
 
     method nude() { $!numerator, $!denominator }
 
-    method Num(--> Num:D) {
+    method Num(Rational:D: --> Num:D) {
         nqp::p6box_n(nqp::div_In($!numerator,$!denominator))
     }
 


vrurg:
There is an inconsistency though which is then common to all numeric coercions. Whereas we seemingly cannot coerce from an undefined value we still can `Num(Rat) == 0` which is `True`. This can easily be fixed by adding a multi-candidate:

    multi method Num(Rational:U: --> Num:D) {
        self.Mu::Numeric;
        0e0
    }

But I'm not sure how much sense does it make. `Str()` does result in an empty string with a warning. But `Int()` doesn't. And yet it passes the `Int() == 0` check.

niner (approved):

Issue #1405, opened 2018-01-16
Coercers as `of` in subsets appear to silently fail


zoffixznet:
```
<Zoffix__> m: my subset B2 of Numeric(Cool) where Int:U|Num:D|Rat:D; say 42e0 ~~ B2
<camelia> rakudo-moar 2ac1ceaa1: OUTPUT: «False␤»
<Zoffix__> m: my subset B2 of Numeric(Cool) where Int:U|Num:D|Rat:D; say "42e0" ~~ B2
<camelia> rakudo-moar 2ac1ceaa1: OUTPUT: «False␤»
<Zoffix__> m: my subset B2 of Cool where Int:U|Num:D|Rat:D; say 42e0 ~~ B2
<camelia> rakudo-moar 2ac1ceaa1: OUTPUT: «True␤»
```

vrurg:
Fixed with the merge of #4073

PR #4073, opened 2020-11-29
Fix subsets of coercions

vrurg:
A subset would now coerce a value if built upon a coercion type.

    subset S of Num(Str) where {!.defined || $_ > 0};
    my S $foo;
    $foo = "42"; # 42e0

Also fixes smartmatching over subsets:

    "42" ~~ S;

    subset Foo of Num(Str);
    Str ~~ Foo;

Fix for #1405

### DIFF
--- a/src/Perl6/Actions.nqp
+++ b/src/Perl6/Actions.nqp
@@ -5141,8 +5141,7 @@ class Perl6::Actions is HLL::Actions does STDActions {
 
         # If we have a refinement, make sure it's thunked if needed. If none,
         # just always true.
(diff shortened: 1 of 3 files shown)


vrurg:
This PR should also introduce a little bit of optimization for subsets with no `where` block. Now instead of autogenerating an always-True block Rakudo will set  refinement to `nqp::null`. `SubsetHOW.accepts_type` would then bypass invocation of `ACCEPTS` method on refinement.

niner (approved):

PR #4080, opened 2020-12-01
fix missing =over =back in docs/running.pod

dod38fr:
Hi

Without this patch, `pod2man --name=perl6 docs/running.pod` returns a lot of errors because of missing `=back` `=over` and `=encoding UTF-8` tags which are required by Perl5's pod syntax.

This patch add the missing tags.

All the best

### DIFF
--- a/docs/running.pod
+++ b/docs/running.pod
@@ -1,3 +1,10 @@
+# This pod file is used in Debian to generate C<raku> man page when
+# building C<rakudo> package. At that stage only Perl5 C<pod2man>
+# program is available. So this file must follow Perl5's pod syntax.
+#
+# See https://github.com/rakudo/rakudo/pull/4080#issuecomment-736682853
+# for more information.
+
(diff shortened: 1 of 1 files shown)


coke:
Shouldn't we make this valid POD6 instead?

lizmat:
Probably  :-)

dod38fr:
In Debian, raku man page is generated from running.pod with Perl5's pod2man. 

We'd need a similar program shipped with rakudo if running.pod is converted to pod6. I'd rather avoid depending on another Raku module because build dependency loops are hard to manage.

coke:
OK. let's just add a comment at top of file explaining why it's in p5 pod. (in source that doesn't end up in the rendered man page)

dod38fr:
@coke: I've added a comment in my PR

coke (approved):

coke:
Thanks!

PR #4084, opened 2020-12-02
Fix a shaped array performance problem

jnthn:
When we want to turn a native array into a shaped native array, we
steal the meta-object and re-use it on a new type with the multi-dim
array representation. While that preserves the type behavior (aside from
representation), there's a new STable, which lacks the various caches.
Having those attached means there's now no non-cached type check or
method lookup MOP queries when we work with shaped native arrays, which
was the cause of the slow performance compared to non-shaped native
arrays.


lizmat (approved):

niner (approved):

lizmat:
Thanks!

Issue #3308, opened 2019-11-19
BOOTStr issue


j-mckim:
## The Problem

When I run the example code, it fails in various ways that lead me to believe what is being assigned to the variable is not a "normal" type.

This may be related to [rakudo/rakudo/issues/1800)] in which case I apologize for the noise.

## Expected Behavior

I should get an object of type string.

## Actual Behavior

I get something else - a BOOTStr. Subsequent code in the program is unable to deal with it.
An example diagnostic:

Type check failed in binding to parameter '$content'; expected Str but got BOOTStr (?)
  in sub do_something at bootstr-bug.p6 line 9
  in block <unit> at bootstr-bug.p6 line 11

## Steps to Reproduce

Run this script:

<
#!perl6

use v6;

my $user_name = Str($*USER);

#say $user_name; # fails, no method 'gist'

sub do_something(Str $content) {}

do_something($user_name); # fails, with mention of BOOTStr
>

## Environment
* Operating system:  Ubuntu 18.04
* Compiler version: Rakudo Star version 2019.03.1 built on MoarVM version 2019.03
implementing Perl 6.d.

kjkuan:
Curiously `$*USER` is a `IntStr` rather than a `Str`...hmm, and using `Str( )` on instances of `IntStr` does result it a `BOOTStr`:
```
my $n = Str(<123>);
say $n.^name;  # BOOTStr
say $n;   # fails
```

lizmat:
Oddly enough I *can* replicate the issue with `$*USER` on HEAD, but not the more generic `<123>` case.

lizmat:
bisectable points to: https://github.com/rakudo/rakudo/commit/b0a1b6c31c8ee93dec53ba117bd2d8b68bfcfd31

Kaiepi:
I think this points to a broader issue with the MOP than just the `name` meta-method. Take this for example:

```perl6
use v6.d;

class FooHOW
    does Metamodel::Naming
    does Metamodel::TypePretense
{
    method new_type(FooHOW:_: Str:D :$name --> Mu) {
        my ::?CLASS:D $meta := self.new;
        my Mu         $type := Metamodel::Primitives.create_type: $meta;
        $meta.set_name: $type, $name;
        $type
    }
}

BEGIN FooHOW.pretend_to_be: (Mu, Any);

my constant Foo = FooHOW.new_type: :name<Foo>;
say Metamodel::Primitives.is_type: Foo, Any;
```

[...]

PR #4096, opened 2020-12-08
Implement invocation protocol for roles

### LINKED ISSUE #4094: Roles are not auto-punned for postcircumfix:<( )>
This works as expected:

    class A {
        method CALL-ME(|c) { dd c }
    }
    A(42,666);  # \(42,666)

    role B {
        method CALL-ME(|c) { dd c }
    }
    B(42,666);  # Cannot invoke this object (REPR: Uninstantiable; B)
    (B.^pun)(42,666);  # \(42,666)

Seems to me some magic is missing to handle auto-punning in this situation, like would happen with:

    B.CALL-ME(42,666);  # \(42,666)

vrurg:
Basically, it all winds down to puning a role and then invoking the
resulting class. For that matter only the default handler is used
which is different from that used for other type objects.

Fixes #4094

### DIFF
--- a/src/Perl6/Metamodel/ClassHOW.nqp
+++ b/src/Perl6/Metamodel/ClassHOW.nqp
@@ -30,6 +30,7 @@ class Perl6::Metamodel::ClassHOW
     has @!role_typecheck_list;
     has @!fallbacks;
     has $!composed;
+    has $!is_pun;
     has $!pun_source; # If class is coming from a pun then this is the source role
 
     my $archetypes := Perl6::Metamodel::Archetypes.new(
@@ -312,6 +313,15 @@ class Perl6::Metamodel::ClassHOW
 
     method set_pun_source($obj, $role) {
         $!pun_source := nqp::decont($role);
+        $!is_pun := 1;
+    }
+
+    method is_pun($obj) {
+        $!is_pun
+    }
+
+    method pun_source($obj) {
+        $!pun_source
     }
 }
 
--- a/src/Perl6/Metamodel/CoercionHOW.nqp
+++ b/src/Perl6/Metamodel/CoercionHOW.nqp
@@ -157,28 +157,32 @@ class Perl6::Metamodel::CoercionHOW
             $coercion_method := $!nominal_target.HOW.name($!nominal_target);
             $method := nqp::tryfindmethod($value_type, $coercion_method);
             if nqp::defined($method) {
-                $coerced_value := $method($value)
+                $coerced_value := $method($value);
             }
+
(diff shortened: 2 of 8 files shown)


vrurg:
I decided not to implement invocation attributes and handlers because not sure if it makes any sense.

No tests added yet. Some specs are broken.

lizmat (approved):
Looks good to me, but then I should probably recuse myself from reviewing this.

vrurg:
With regard to broken spectests, the only fundamental change is that `Str(IntStr.new)` doesn't coerce into `Str` anymore. The test code is `Str($*USER).^name`.

The change is the result of using the now standard coercion protocol where previously it was explicit `$*USER.Str` used. The outcome makes full sense because `IntStr ~~ Str`, so `Str(IntStr) === IntStr`. But _t/S02-literal/allomorphic.t_ insist on `Str`.

@jnthn, would you mind changing the test? It's not yet a part of a released language version (added in Feb, 2020). I think something like `$*USER.Str` would be testing the same thing, just other way around.

vrurg:
I just've checked out #3308 and the fix for it in 3a86272. `$*USER.Str` doesn't make sense in their context. It now looks to me that `isa-ok Str($*USER), Str` should be the right thing to do for the test.

vrurg:
Ok, I have made a blin run against this PR, took care of two mistakes in failing modules. Otherwise no breakages found.

If nobody minds, I plan to merge it tomorrow (~12hr since this comment).

lizmat:
Do we have an idea on the performance impact of this?

vrurg:
I don't expect any significant impact as no hot paths are touched. The coercion case of `Role($value)` could be impacted because a direct method call is now replaced with coercion protocol. But that'd be little price paid for the unification of semantics.

Perhaps, we need some kind of a canary test for performance bundled with Rakudo. Just for the sake of running it manually against different versions. Because to my view, even if I come up with some kind of test case it most likely would be biased.

lizmat:
Yeah, I have been giving some thoughts about that.  Somehow picking up work on `MoarVM::Profiler` again, and build some kind of tests on that checking for number of frames run and number of allocations made and stuff like that.  Would that make sense?

vrurg:
I think it would. But something backend-agnostic measuring just timings would be good to have too. Something like Tux's reports on IRC.

lizmat:
Yeah, but since our profiler only runs on MoarVM, and these tests would specifically **not** test timings, as these would be too fickle and produce a lot of false positives.  The number of frames / allocations is a much better metric in that case I think.

Issue #4098, opened 2020-12-09
Warning on use of `contains` or `index` with collections; and other traps due to naming?


raiph:
Imo the warnings that are displayed when `contains` or `index` are called on `List`s (or subtypes thereof) are an excellent solution to the corresponding entry on the traps page.

----

I recently encountered [this](https://www.reddit.com/r/adventofcode/comments/k6nbmv/2020_day_4_part2_raku_off_by_one/) in the 2020 AdventOfCode, with the following bug and comment by the adventering Scala dev:
```
when /^(<digit>+)'cm'$/ { take (150..193).contains($/[0].Int) }
```

> I'm glad I asked for help, because I would not have found that myself!

Perhaps this is a perfect example of how we don't have to drive Raku into being a nanny PL. Someone can get stuck, ask, and it becomes a communal learning process, and that's OK.

But, maybe we do want to do something, and I imagine it's already been discussed:
```
say (10..30).contains: 5;                    # True
say (10..30).index: 5;                       # 16
say {:1a, :2b, :!c, :d, :e}.contains: 's';   # True
say {:1a, :2b, :!c, :d, :e}.index: 'b';      # 4
```

I've looked for, but have been unable to find, discussion of the above, and having warnings if these two routines are called on other `Iterable`s.

----

I can well imagine that several discussions have been had and I've just failed to find them. And that the decision, if the above examples were discussed, was that producing warnings when `contains` or `index` were applied to `List`s and `Array`s was sufficient, at least for the near term.

After all, the underlying technical issue is that these routines silently coerce to string, and, more generally that Raku fully embraces weak typing. So devs who guess what a routine does without checking the doc may get bit, and that can be perfectly cromulent, even if those used to strong static typing by default for all built ins might be put off.

And/or, that having warnings when these routines are applied to `List`s may well be just the right and sufficient touch, because newbies will likely quickly bump into the issue, and either see the warning, or be pointed to the warning if they don't spot it, or can be pointed to the `List` warning if they encounter the problem with some other collection type that doesn't warn, so perhaps it's all good.

----

While I'm writing this I can't help but also wonder if there are other routines or type names that are plausibly similarly problematic, and, if so, if a warning is plausibly the right solution for any of these but no warning has yet been introduced.

Specifically, Raku routines, types,, keywords, constants, etc with a name that is familiar to coders in other major PLs, but such that what that name implies in those PLs is sufficiently awkwardly inconsistent with what it implies in Raku that it may also deserve a warning.

[...]

PR #4099, opened 2020-12-09
Add warnings on improper use of .contains and .index

lizmat:
As part of the discussion in #4098.  Spectest clean.

### DIFF
--- a/src/core.c/Map.pm6
+++ b/src/core.c/Map.pm6
@@ -13,6 +13,18 @@ my class Map does Iterable does Associative { # declared in BOOTSTRAP
         self.new.STORE(@args, :INITIALIZE)
     }
 
+    multi method contains(Map:D: \needle) {
+        my $name := self.^name;
+        warn "Applying '.contains' to a $name will look at its .Str representation.  Did you mean '$name\{needle}:exists'?".naive-word-wrapper;
+        self.Str.contains(needle)
+    }
+
+    multi method index(Map:D: \needle) {
+        my $name := self.^name;
+        warn "Applying '.index' to a $name will look at its .Str representation.  Did you mean '$name\{needle}:exists'?".naive-word-wrapper;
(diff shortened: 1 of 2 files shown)


raiph:
Is `index` as a sub delegating to method form? If so, then great, if not,, then there's that.

lizmat:
It does.

raiph:
Does it make sense to at least consider having default implementations for `contains` and `index` in the `Iterable` role that produce a simple generic warning?

I say this because, while I see [Zoffix's point about retaining `contains` and `index` for non-Str objects that respond to a `.Str` coercion by producing a scalar string rendering that works fine for `contains` and `index` such as `Match` objects and `IO::Path` objects](https://github.com/rakudo/rakudo/pull/907#issuecomment-256741597) (I presume he's right about the latter), it seems to me that using `contains` or `index` on any `Iterable` is asking for trouble.

raiph:
Strawman proposed generic warnings for inclusion in `contains` and `index` methods in `Iterable`:

        my $name := self.^name;
        warn "Applying '.contains' to a $name will look at its .Str representation.".naive-word-wrapper;

        my $name := self.^name;
        warn "Applying '.index' to a $name will look at its .Str representation.".naive-word-wrapper;

lizmat:
I think we still need the discussion on whether we should do these warnings (which I'm in favour of) and for which methods: today on #raku, someone was using `.comb` on an array of arrays.  Same for `split` I guess.  And `starts-with`, `ends-with`, `eqat` etc.

lizmat:
Perhaps the subs / methods in question could take a `:quietly` parameter to indicate that you know what you're doing?

patrickbkr:
Wouldn't `.Str' do the same and be clearer about what's happening?

lizmat:
@patrickbkr good point.  Ok, scratch the `:quietly` idea :-)

PR #4103, opened 2020-12-10
Re-imagine basic slice assignment, 5.8x as fast

lizmat:
- makes @a[1,2,3] = 4,5,6 about 5.8x as fast
- breaks 2 spectests that assume handling lazy positions also means
  that the result should be lazy.  The tests in t/spec/S32-list/seq
  (23,25) are in fact marked as testing "experimental behaviour",
  so one could argue that they can be changed.

Should it be deemed necessary that a lazy position should also mean a lazy result, then that could be done relatively easy.  But I would argue that the current behaviour (assigning until a position is produced that is outside of the original `Positional`) makes more sense.

### DIFF
--- a/src/core.c/Rakudo/Iterator.pm6
+++ b/src/core.c/Rakudo/Iterator.pm6
@@ -3269,6 +3269,7 @@ class Rakudo::Iterator {
     # Positional indexing, the number of elements in the Positional.  The
     # iterator will produce positions until the given iterator is exhausted.
     # Also takes care of Callables being produced by the given iterator.
+    # Any non-numeric position will be returned as is.
     class PosWithCallables {
         has $!iterator;
         has $!elems;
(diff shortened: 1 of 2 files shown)


niner:
Which other cases have you benchmarked for this?

lizmat:
None, because these are the ones with the least overhead so makes it clear how much the process itself has improved.

Which ones would you like me to benchmark?

lizmat:
@Altai-man Am planning on merging this **after** the 2020.12 release.

lizmat:
Added the following PORTING CAVEAT:

    =head2 Microseconds vs Seconds

    Due to a misunderstanding, it was found that the the C<times> function returns
    B<microseconds> versus seconds (as the original Perl version does).  This was
    only found after two years of this module's existence, and deemed much more
    useful in the Raku context.

    Therefore it was decided to keep this behaviour in December 2020.

New version on its way to the ecosystem.  Thanks for the spot!

lizmat:
Converted this PR back to draft.  I'm going to probably close this PR, as I found some issues while working on making array slicing of Iterables faster (aka `@a[@i]`).  Specifically, it looks like some behaviours in this PR are different from behaviour or a basic array slice, so it's probably undertested, otherwise roast would have found them for this PR.

lizmat:
Closing in favour of https://github.com/rakudo/rakudo/pull/4131 which provides more functionality *and* better performance.

Issue #4104, opened 2020-12-11
Array.of throws on type objects


Leont:
## The Problem
All of these work:
```
dd Positional.of
dd Positional[Str].of
dd Array[Str].of
```
But `Array.of` throws `Cannot look up attributes in a Array type object`. It should return `Mu` or `Any`.

## Environment
```
Welcome to Rakudo(tm) v2020.11.
Implementing the Raku(tm) programming language v6.d.
Built on MoarVM version 2020.11.
```

lizmat:
if the PR is the right approach, this will have to be done for the `returns`, `name` and possibly other methods as well.

Leont:
> if the PR is the right approach, this will have to be done for the returns, name and possibly other methods as well.

Yeah, that makes sense

melezhik:
Added to RICH - http://rakudist.raku.org/sparky/report/RakuPlay-2/4803

Leont:
This has been fixed in the mean time

PR #4106, opened 2020-12-11
Make sure that .of works on type objects always

lizmat:
If a type supports the .of method, that is.

In response to https://github.com/rakudo/rakudo/issues/4104

### DIFF
--- a/src/core.c/Array.pm6
+++ b/src/core.c/Array.pm6
@@ -1321,13 +1321,13 @@ my class Array { # declared in BOOTSTRAP
     method name() {
         nqp::isnull($!descriptor) ?? Nil !! $!descriptor.name
     }
-    method of() {
-        nqp::isconcrete(self)
-          ?? nqp::isnull($!descriptor)
-            ?? Mu
-            !! $!descriptor.of
-          !! Mu
+
+    proto method of() {*}
+    multi method of(Array:U: --> Mu) { }
+    multi method of(Array:D:) {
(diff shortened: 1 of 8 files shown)


src/core.c/Array.pm6
```diff
@@ -1322,7 +1322,11 @@ my class Array { # declared in BOOTSTRAP
         nqp::isnull($!descriptor) ?? Nil !! $!descriptor.name
     }
     method of() {
-        nqp::isnull($!descriptor) ?? Mu !! $!descriptor.of
+        nqp::isconcrete(self)
```
niner, line 1325: Why not make a multi out of it and benefit from dispatch caching?

lizmat: NO real reason, I wanted to keep the diff as small as possible, to at least gauge whether this would be considered the right approach.  Because making it a multi then poses the question whether it should become available for every type, (aka having a Mu.of fallback) or not.

Also note that this would need to be done for `.returns` and `.name` as well.

niner (approved):

src/core.c/Baggy.pm6
```diff
@@ -12,7 +12,7 @@ my role Baggy does QuantHash {
 # Immutables aspects of Bag/Mix, need to live to Bag/Mix respectively.
 
 #--- interface methods
-    method of() { UInt }
```
b2gills, line 15: I think that UInt needs to actually be in the code since it is just a subset.

PR #4111, opened 2020-12-13
Use `exec` instead of `system` and then exiting with the return code.

nwc10:
The approach using `system` misses the case where the invoked program exits
with a signal - for example if the OOM killer takes out the setting build.

Before this change the wrapper script would calmly take the value 137
returned from `system`, shift it right by 8 and exit with status 0, which
make took as success and carried on. At which point the next command invoked
would mysteriously fail due to a missing input file.

Now we see:
    Stage start      :   0.001
    Stage parse      : 745.469
    Stage syntaxcheck:   0.000
    Stage ast        :   0.000
    Stage optimize   : 102.057
    Stage mast       : Killed
    make: *** [Makefile:1170: blib/CORE.c.setting.moarvm] Error 137

a useful error message and the build aborts.

Where we still need `system` in Configure.pl, simply check whether it
returned non-zero. This way we also catch the error case of death by fatal
signal.

I checked perlport - `exec` and `system` have identical portability
caveats, so this change *should* not cause problems (at least, not cause
problems that we didn't already expose ourselves to).

### DIFF
--- a/Configure.pl
+++ b/Configure.pl
@@ -16,7 +16,7 @@ BEGIN {
     my $set_config = !qx{git config rakudo.initialized};
     if ( !-e '3rdparty/nqp-configure/LICENSE' ) {
         my $code = system($^X, 'tools/build/update-submodules.pl', Cwd::cwd(), @ARGV);
-        exit 1 if $code >> 8 != 0;
+        exit 1 if $code;
         $set_config = 1;
     }
     if ($set_config) {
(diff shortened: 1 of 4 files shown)


lizmat (approved):

niner (approved):

vrurg:
I like the idea overall, but the consistent failure of all Windows builds makes me suspicious of `exec` not being well-emulated on the platform.

patrickbkr:
I think I remember the problem. (The CI failures look very familiar.) Perls exec does not seem to block on Windows. IIRC that was the reason I went with `system` in the runners. (After a looot of trail and error.)

lizmat:
So, can this be merged?  Or is more research on Windows needed?

nwc10:
More research needed - sorry, forgot to update this.

Problem is that I don't have any direct access to Windows, the documentation seems to be, um, LTA, and playing whack-a-mole with private branches and pushes is not very efficient.

So not sure how to proceed.

PR #4113, opened 2020-12-15
Displays default values for parameters in sub MAIN

Xliff:
This PR adds default values to the end of parameters with a .WHY.trailing element.

Take the following example:

```
sub MAIN (
        $dummy,
        :$name = 'rakudo', #= Name
        :$rat  = 22/7,     #= Rational
        :$int  = 314159,   #= Integer
        :$str = 'This string is too long for a proper default value!', #= Obnoxiously long string
        :$bigint = 12345678910111213141516 #= Obnoxiously long integer
) {
        say 'This line is not important'
}
```

This now displays:

```
Usage:
  default1.raku [--name=<Any>] [--rat=<Any>] [--int=<Any>] [--str=<Any>] [--bigint=<Any>] <dummy>
  
    --name=<Any>      Name *[default: 'rakudo']*
    --rat=<Any>       Rational *[default: 3.142857]*
    --int=<Any>       Integer *[default: 314159]*
    --str=<Any>       Obnoxiously long string *[default: 'This string is too …']*
    --bigint=<Any>    Obnoxiously long integer *[default: 123456789…1213141516]*
```

(Emphasis, mine)

Currently this PR does not add default values if there is no .WHY.trailing comment. If it is the consensus of the group that this feature be added, this PR will be updated to include those changes.


Altai-man:
I understand the API is not changed here, just the user-facing output?

Xliff:
@Altai-man: Correct.

src/core.c/Main.pm6
```diff
@@ -276,7 +276,32 @@ my sub RUN-MAIN(&main, $mainline, :$in-as-argsfiles) {
                     }
                     @positional.push($argument);
                 }
-                @arg-help.push($argument => $param.WHY.contents) if $param.WHY and (@arg-help.grep:{ .key eq $argument}) == Empty;  # Use first defined
+            		#@arg-help.push($argument => $param.WHY.contents) if $param.WHY and (@arg-help.grep:{ .key eq $argument}) == Empty;  # Use first defined
```
ugexe, line 279: This line isn't needed

src/core.c/Main.pm6
```diff
@@ -276,7 +276,32 @@ my sub RUN-MAIN(&main, $mainline, :$in-as-argsfiles) {
...
+            		if $param.WHY and (@arg-help.grep:{ .key eq $argument}) == Empty {
```
ugexe, line 280: Is the indentation on this correct?

Xliff: Probably not, I will need to rework the PR with respect to indentation. I was more concerned with functionality when I submitted this PR. Expect another commit to correct the whitespacing.

src/core.c/Main.pm6
Xliff: The comment was there to serve as a location mark. I will remove it.

codesections:
I really like this PR – it adds some greatly appreciated support for more complex CLI apps.

That said, I have one concern about the way this PR handles flags (aka options that don't take any arguments).  If I extend your `default` example by adding a `Bool :verbose(:$v)` flag, I get the following `USAGE`:

```sh
Usage:                                                                                                    
  default2 [--name=<Any>] [--rat=<Any>] [--int=<Any>] [--str=<Any>] [--bigint=<Any>] [--verbose|-v] <dummy>
                                                                                                          
    --name=<Any>      Name [default: 'rakudo']    
    --rat=<Any>       Rational [default: 3.142857]      
    --int=<Any>       Integer [default: 314159]                
    --str=<Any>       Obnoxiously long string [default: 'This string is too …']
    --bigint=<Any>    Obnoxiously long integer [default: 123456789…1213141516] 
    --verbose|-v      Print verbose output [default: False]
```

The `[default: False]` value for `--verbose` strikes me as out of place – at the very least, it's inconsistent with the way the vast majority of Linux CLIs print their usage info (for example, the `--help` output of commands like `grep` or `ls` would get _very_ cluttered if they listed `[default: False]` for every flag they accept; the presumption is that flags are off unless turned on).  Would you be open to changing the PR to not print defaults for flags?  (I note that this is consistent with our treatment of type constraints is `USAGE`: the signature for `MAIN` constrains `$v` to be a `Bool`, but we don't print this, since it's clear that flags can only be on or off.)

(One other very minor note: In your comment, you showed the output including asterisks (`*[defalut: 'rakudo']*`) but I notice that the output doesn't print them.  I'm not sure which you intended, but I'll say I much prefer the way you have it now, without the asterisks.  It's both cleaner and more consistent with how I've seen other programs print defaults.)

Xliff:
@codesections - For situations like this:

```raku
sub MAIN (
  Bool :verbose(:$v) = True
)
```

I believe displaying the default value is warranted. How about this? If the option is a Boolean parameter with a defined default value, it is displayed, otherwise it is not.

Would that be a better solution?

codesections:
Yep, that would work.  Is there a particular reason you think we should go with "a defined default value" rather than `True`?  They cover almost the same ground, and I can't think of a situation where someone would want to print that a flag defaults to `False`.  That said, I don't have strong feelings on the matter – needing a definite `False` default is also fairly rare (the only times it's come up for me was when I wanted to be able to do math with the `False` without Rakudo complaining.

Xliff:
Yes. It gives the coder more control, rather than having to make (arguably a rare) mistaken assumption.

```raku
sub MAIN (
  Bool :$right = True,
  Bool :$left  = False
  Bool :$ambidextrous
)
```

In the above case, the programmer has said "I only want two values to have defaults", so the current suggestion honors that wish.

codesections:
Hmm, I'm all for giving the coder more control.  But I think the example you provided actually moved me more in the other direction.  If we draw the line based on the defaults set in `MAIN`, a slightly expanded version of program would produce a usage message something like

```sh
Usage:                                                                                                    
  pitch [--left] [--right] [--ambidextrous] -- Throw a baseball to the batter
                                                                            -                              
    --right             Throw right-handed [default: True]
    --left              Throw left-handed [default: False]
    --ambidextrous      Throw with a random hand
    --speed=<Int>       Speed (in mph) at which to throw the ball
```

Putting myself in the head of the user of the CLI program (who may or may not know Raku, and certainly shouldn't need to be thinking about what is going on in `MAIN`), I think that'd be pretty confusing.  I imagine a thought process something like:

> It says that the default for `--left` is false, but doesn't list a default for `--ambidextrous`.   Did the programmer just forget to list one?  Or is it something other than false?  I don't think it could default to true.  Maybe it's an option, and I'm supposed to pass in a value?  But it doesn't look like `--speed`…

I still think that not printing a default for any flag that defaults to `False` is the way to go, just to keep the help text slightly less cluttered.  But after thinking through this example, I'm inclined to believe that basing what's shown on what default is explicit versus implicit in `MAIN` is splitting the baby in a way that's worse than either alternative.  So if you really don't want to hide `False` default values for flags, I think it'd be better to just print all default values (and accept that this will mean CLIs with a lot of flags will have somewhat cluttered `--help` text.)  But I'd still prefer not to print it for `False` flags :grinning:

Xliff:
@codesections - If there is an overwhelming response for this behavior, I'll update the PR. Thanks for the coments!

Xliff:
On a lark, could always add the "is hidden-from-defaults" trait....

codesections:
It just occurred to me that many standard *nix CLI programs use "default" in two slightly different ways, and this PR covers only one of them.

`ls` does a good job of distinguishing between these two types of "default", and includes the following lines in its `--help` output:

```
    --hyperlink[=WHEN]      hyperlink file names; WHEN can be 'always'
                            (default if omitted), 'auto', or 'never'
    --indicator-style=WORD  append indicator to with style WORD to entry name:
                            none (default), slash (-p),
                            file-type (--file-type), classify (-F)
```
that is `default` in the text for `--indicator-style` lists the default value that is used if you call `ls` without the `--indicator-style` option; this is the same thing that `[default: ____]` means in this PR.  The way `ls` uses `default` in `--hyperlink`, however, is different: there, `default if omitted` says what happens if you call `ls` _with_ the `--hyperlink` option, but don't pass the (optional) argument to `--hyperlink`.  This PR doesn't provide a way to express that second type of `default`.  (And while `ls` is careful to distinguish between the two, not all programs are – for example, `cp` uses `default` to mean the same thing that `ls` means by `default if omitted`.)

I mention this partly because the one CLI program we _know_ raku users are using – `raku` – uses `default` in the "default if omitted" sense.  I'm not _positive_ that any of this really has an impact for this PR – maybe we want to keep this PR as-is, and save the question of how, if at all, to list default-if-omitted values for another day.  (Now that I think about it, we might not even have a good way to _create_ a default-if-omitted option through the signature for `MAIN`, even though both `raku` and `zef` use default-if-omitted values.  Maybe I should open a separate issue for that?).

But, in any event, I wanted to raise these issues here so that we can at least *intentionally* hold off on addressing them in this PR!

Xliff:
Yes, this can be closed.

PR #4120, opened 2020-12-17
Add missing `-I lib` to an is-run test

MasterDuke17:
Otherwise it passes when run under make (e.g., `make
t/02-rakudo/99-misc.t` or `make m-test`), but not when run directly
(e.g., `./rakudo-m -I lib t/02-rakudo/99-misc.t`).

### DIFF
--- a/t/02-rakudo/99-misc.t
+++ b/t/02-rakudo/99-misc.t
@@ -142,7 +142,7 @@ group-of 2 => 'collation experiment' => {
         use experimental :collation;
         $*COLLATION.set: :primary;
         print 'pass'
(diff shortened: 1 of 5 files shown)


lizmat:
Could this be missing at more places?

MasterDuke17:
> Could this be missing at more places?

It was needed because of the `use experimental`, but I didn't check anywhere else.

MasterDuke17:
> It was needed because of the `use experimental`, but I didn't check anywhere else.

Added the other one I found.

MasterDuke17:
> Added the other one I found.

Looks like `use Test` in an `is-run` needs it too, so there are some more I'll add.

lizmat (approved):

PR #4129, opened 2020-12-20
Move to CentOS 7 in binary release linux build

patrickbkr:
CentOS 6 has gone EOL in 2020-11. According to
https://gist.github.com/wagenet/35adca1a032cec2999d47b6c40aa45b1
CentOS 7 is now the Distro with the oldest glibc (2.17). So we move to
that. This means that from now on the binary releases will not work on CentOS 6 anymore.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -231,7 +231,7 @@ stages:
       pool:
         vmImage: 'ubuntu-18.04'
       container:
-        image: centos:6
+        image: centos:7
         options: "--name raku-build-container -v /usr/bin/docker:/tmp/docker:ro"
       workspace:
         clean: all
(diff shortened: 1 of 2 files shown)


patrickbkr:
@Altai-man Should this change be announced or noted somewhere?

Altai-man:
In the next changelog special, I think.

As for the place to specify it as a fact somewhere, I don't think we have that, https://github.com/Raku/problem-solving/pull/219 got trapped into infinite "no human resources to consider" state.

patrickbkr:
WRT Raku/problem-solving#219 that ticket is on my table. I am (slowly) working on the implementation of a bot to orchestrate our CI pipelines. So it's not blocked, just progressing slowly. Supporting tony-o with the ecosystem rewrite has higher priority for me at the moment though. (Robust ecosystem > robust CI pipeline)

Altai-man:
@patrickbkr those are awesome news! I am very glad to hear that.

I think this can be merged now, given we just had a release.

Issue #2628, opened 2019-01-19
[JVM] build is broken (error in stage parse during generation of CORE.d.setting.jar)


usev6:
## The Problem
Trying to build rakudo-jvm on current head (304feae9f0) leads to this error during generation of CORE.d.setting.jar:
```
NQP_LIB=blib java -Xss1m -Xms500m -Xmx3000m -cp .:/usr/home/christian/perl6/tmp/rakudo_jvm_constant_6d/install/share/nqp/runtime/asm-4.1.jar:/usr/home/christian/perl6/tmp/rakudo_jvm_constant_6d/install/share/nqp/runtime/asm-tree-4.1.jar:/usr/home/christian/perl6/tmp/rakudo_jvm_constant_6d/install/share/nqp/runtime/jline-1.0.jar:/usr/home/christian/perl6/tmp/rakudo_jvm_constant_6d/install/share/nqp/runtime/jna-4.0.0.jar:/usr/home/christian/perl6/tmp/rakudo_jvm_constant_6d/install/share/nqp/runtime/nqp-runtime.jar:/usr/home/christian/perl6/tmp/rakudo_jvm_constant_6d/install/share/nqp/lib/nqp.jar:rakudo-runtime.jar:perl6.jar:/usr/home/christian/perl6/tmp/rakudo_jvm_constant_6d/install/share/nqp/lib perl6 --setting=NULL.d --ll-exception --optimize=3 --target=jar --stagestats --output=CORE.d.setting.jar --nqp-lib=blib gen/jvm/CORE.d.setting
Stage start      :   0.001
Stage parse      : Method 'CORE' not found for invocant of class 'QAST::SVal'
  in set (gen/jvm/stage2/QASTNode.nqp:142)
  in new (gen/jvm/stage2/QASTNode.nqp:447)
  in perl6_module_loader_code (src/Perl6/World.nqp:1412)
  in load_setting (src/Perl6/World.nqp:853)
  in loading_and_symbol_setup (src/Perl6/World.nqp:608)
  in comp_unit (gen/jvm/Perl6-Grammar.nqp:827)
  in TOP (gen/jvm/Perl6-Grammar.nqp:559)
  in parse (gen/jvm/stage2/QRegex.nqp:2303)
  in parse (gen/jvm/stage2/NQPHLL.nqp:1546)
  in execute_stage (gen/jvm/stage2/NQPHLL.nqp:1462)
  in run (gen/jvm/stage2/NQPHLL.nqp:1495)
[...]

Issue #1666, opened 2018-03-28
[JVM] Native arrays are miscompiled in core settings


usev6:
For some reasons its not possible to use native arrays when starting the supervisor thread (src/core/ThreadPoolScheduler.pm6) on the JVM backend.

Trying to create a native num array with ```my num @last-utils``` came out as a VMArrayInstance instead of VMArrayInstance_n. This problem already lead to the following workaround: https://github.com/rakudo/rakudo/commit/581edd58e9

I'm filing this issue, since another workaround is needed for using nqp::getrusage in src/core/ThreadPoolScheduler.pm6 and I want to reference a bug report.

usev6:
As I found out in https://github.com/rakudo/rakudo/pull/4295 it seems to be possible to use native arrays if they are declared with
```
my @last-utils := array[num].new;
```
instead of
```
my num @last-util;
```

The latter form should work, too (so I won't close this issue), but it's good to have a workaround.

usev6:
The same problem happens with precompilation. (It popped up in https://github.com/rakudo/rakudo/issues/4354.)

So this blows up:
```
$ cat lib/Foo.rakumod
use nqp;                                                                                                                                                                                                                                       
#no precompilation;                                                                                                                                                                                                                            
                                                                                                                                                                                                                                               
class Foo {                                                                                                                                                                                                                                    
    method foo() {                                                                                                                                                                                                                             
        #my @data := array[int].new;                                                                                                                                                                                                           
        my int @data;                                                                                                                                                                                                                          
        nqp::bindpos_i(@data,0,42);                                                                                                                                                                                                            
        @data;                                                                                                                                                                                                                                 
    }                                                                                                                                                                                                                                          
}
$ ./rakudo-j -Ilib -e 'use Foo; dd Foo.new.foo'
VMArray representation does not implement bind_pos_native
  in method foo at /path/to/rakudo.jvm/../rakudo.jvm/lib/Foo.rakumod (Foo) line 8
  in block <unit> at -e line 1
```

The code works with either ```no precompilation``` or ```my @data := array[int].new``` (instead of ```my int @data```).

niner:
> The same problem happens with precompilation. (It popped up in
> https://github.com/rakudo/rakudo/issues/4354.)

In that case I'd say the problem _is_ precompilation. That's probably, why it 
affects the setting: because that's always precomiled.

usev6:
[...]

PR #4131, opened 2020-12-21
Re-imagine array slicing, fixing issues, between 2x and 10x faster

lizmat:
This PR makes array slice accesses with positions from an Iterable (aka `@a[@i]`) about 5x as fast *and* fixes several issues found in the previous implementation.  It also creates a framework for handling adverbs that are allowed with postcircumfix adverbs (in a separate source-file Rakudo/Internals/PostcircumfixAdverbs).

It also makes Whatever slices (aka `@a[*]`) about 8x as fast.  And it makes a bare zen-slice (aka `@a[]`) 12x as fast, and a zen-slice with adverbs 6x as fast.

Specific fixes are with Iterables within Iterables as index.  For example,

  dd <a b c d e>[0,(1,2,3),4] # ("a", ("b", "c", "d"), "e")

worked, but:

  dd <a b c d e>[0,(1,2,3),4]:k # (0,4)

did not, simply ignoring the inner iterable (1,2,3).  This is now
fixed.  This was apparently not implemented correctly since its initial
implementation.

This is spectest clean, but clearly more tests are needed since the
above error has not been spotted before now


src/core.c/array_slice.pm6
```diff
@@ -183,10 +183,513 @@ multi sub postcircumfix:<[ ]>( \SELF, Any:D \pos, Bool() :$v!, *%other ) is raw
...
+        # Good to go!
+        $index = $lookup;
+    }
+
+    my &accept := do if $index {
+
+        # Handlers for lazy positions and adverbs
+        if positions.is-lazy {
+               nqp::iseq_i($index,1) || nqp::iseq_i($index,2)
+              ?? -> \pos {    # lazy, :kv | :!kv
```
vrurg, line 226: Since the approach with a dispatch table is chosen, wouldn't it be faster to keep all the candidates in a local table and then pull `&accepts` from that table? I think it should be some performance gain comparing to the long chain of conditionals.

lizmat: I'm glad you ask this question.  And the sad answer is: this approach is chosen for a reason.

The reason is that every code block that is **accessed** in a block, will case a clone to be made of that block **and** of its closure.  And initializing it into an nqp::list counts as accessing.  If you have a dispatch table with 34 entries, that's an additional 68 allocations *before* you actually start to do anything.  And these allocations and closure taking completely wipe out **any** performance gain.

On the other hand, one could think of initializing the dispatch table at compile time.  But then they have taken the wrong closure.

To make dispatch tables like this work, the cloning of the block and closure taking should become lazy, I'd think.  Or provide some syntactic / pragma / nqp::code support to mark a block as intended to be cloned lazily.

@jnthn @niner

vrurg: `nqp::getstaticcode`? Pass `$result` as a parameter and get `$done` as a return value? Note that I checked out only few variants and don't know if any other lexical is used elsewhere.

lizmat: Possibly.  But I don't see it documented in https://github.com/Raku/nqp/blob/master/docs/ops.markdown ?  Is it documented somewhere else?  If not, maybe it should be documented ?

lizmat: Passing `$results` is not enough, `SELF` should then also be passed.  Which makes *3* parameters needing passing.  Which, if I recall correctly as I *have* tried that approach, basically kills performance again   :-(

vrurg: > Possibly. But I don't see it documented in https://github.com/Raku/nqp/blob/master/docs/ops.markdown ? Is it documented somewhere else? If not, maybe it should be documented ?

There're a lot of ops awaiting for documentation. Yet, `getstaticcode` is extensively used by `BOOTSTRAP`. I'd say it's implicitly documented. :) There is also one documented, `markcodestatic`. But I'm not really sure I understand it correctly. And if I am then it is only related to serialization.

> Passing `$results` is not enough, `SELF` should then also be passed. Which makes _3_ parameters needing passing. Which, if I recall correctly as I _have_ tried that approach, basically kills performance again :-(

What if the data to process is containerized in a class? I mean, it is really confusing that a long chain of `if`s works faster than anything else.

niner: Is there any way to use multi dispatch here? In Inline::Perl5 I had a similar dispatch chain. Since I try very hard to avoid any NQP there, I came up with this: https://github.com/niner/Inline-Perl5/commit/97ae4c7a3bcff892e7494e6f5824d973d282684d?branch=97ae4c7a3bcff892e7494e6f5824d973d282684d&diff=split and that made it faster and the code much nicer at the same time.

lizmat: @niner The previous implementation used multiple dispatch.  But since the adverbs are all passed as nameds, and the dispatcher gets really confused when having multiple required named parameters (let alone the maintainer), it quickly boils down to checking for one named parameter, and then looking at `%_` for additional named parameters.  Which totally sucks.  But as soon as you want to take all of that handling out of dispatch, you will need to handle the "no named parameters" case as well, as there is no way to dispatch on "no named arguments" if one of the candidates slurps all of the named arguments.

@vrurg ok, so making 2 x 34 classes from a role, and use a dispatch table to select which class to create the handler object with.  Hmmm...   That might work, but will bloat the core significantly, as each class comes with a lot of memory / disk use, no?

vrurg: > ok, so making 2 x 34 classes from a role, and use a dispatch table to select which class to create the handler object with. Hmmm... That might work, but will bloat the core significantly, as each class comes with a lot of memory / disk use, no?

No. I meant to use the table-based dispatch. Then make a single class hold all the parameters we may need and pass its instance as a single parameter to all subs. Something like:

```
my class DispatchParams {
    has $!SELF;
    has $!result;
    has $!pos;
}
my $dparam := nqp::create(DispatchParams);
nqp::bindattr($dparam, DispatchParams, '$!result', $result); # bind to the container
@table[$index].($dparam);

# inside the handler code:
nqp::getattr($dparam, DispatchParams, '$!result') = True; # should work because $!result =:= $result
```

This would eliminate the 3-parameter issue.

Hope I'm not saying anything stupid because I'm not fully awoken yet. :)

lizmat: I've taken @vrurg's suggestion and created a role with a class to implement the specific handling and run this in the branch.  The idea being that we would have a dispatch table of classes, and then just select the right class to do the work.  See code below.

Sadly, this approach for the most simple case, is about 2.5x as slow, basically making it perform as before.  But with a massive greater codegenned footprint, as this approach would require 68 consuming classes of the role to be made.

@vrurg is that what you had in mind?

```raku
use MONKEY;

my role PostcircumfixSlice {
    has $!result;
    has $!elems;
    has $!iterable;
    has int $!done;

    method !SET-SELF(\iterable) {
        $!result   := nqp::create(IterationBuffer);
        $!elems    := nqp::null;
        $!iterable := iterable;
        self
    }
    method new(\iterable) { nqp::create(self)!SET-SELF(iterable) }

    # The method that needs to be implemented by the consuming class
    method !accept(\pos) { ... }

    # Helper method for deleting elements, making sure that the total number
    # of elements is fixed from *before* the first deletion, so that relative
    # positions such as *-1 will continue to refer to the same position,
    # even if the last element of an array was removed (which shortens the
    # array).
    method !delete(\pos) {
        $!elems := $!iterable.elems if nqp::isnull($!elems);
        $!iterable.DELETE-POS(pos)
    }

    method slice(\iterator) {

        # Handle iterator in the generated positions: this will add a List
        # with the elements pointed to by the iterator to the result.  Because
        # these positions can also be non-Int, some trickery needs to be
        # done to allow this being called recursively.
        method !handle-iterator(\iterator) {

            # basically push the current result on a stack
            my int $mark = nqp::elems($!result);

            nqp::until(
              $!done || nqp::eqaddr((my \pos := iterator.pull-one),IterationEnd),
              nqp::if(
                nqp::istype(pos,Int),
                self!accept(pos),
                self!handle-nonInt(pos)
              )
            );

            # Take what was added and push it as a List
            if nqp::sub_i(nqp::elems($!result),$mark) -> int $added {
                my $buffer;
                if $mark {
                    $buffer := nqp::slice($!result,$mark,$added);
                    nqp::setelems($!result,$mark);
                }
                else {
                    $buffer := $!result;
                    $!result := nqp::create(IterationBuffer);
                }
                nqp::push($!result,$buffer.List);
            }
        }

        # Handle anything non-integer in the generated positions
        method !handle-nonInt(\pos) {
            nqp::istype(pos,Iterable)
              ?? nqp::iscont(pos)
                ?? self!accept(pos.Int)
                !! self!handle-iterator(pos.iterator)
              !! nqp::istype(pos,Callable)
                ?? nqp::istype(
                     (my $real := (pos)(
                       nqp::ifnull($!elems,$!elems := $!iterable.elems)
                     )),
                     Int
                   )
                  ?? self!accept($real)
                  !! self!handle-nonInt($real)
                !! nqp::istype(pos,Whatever)
                  ?? self!handle-iterator(
                       (^nqp::ifnull(
                           $!elems,
                           $!elems := $!iterable.elems
                         )).iterator
                     )
                  !! self!accept(pos.Int)
        }

        # Do the actual building of the result
        nqp::until(
          $!done || nqp::eqaddr((my \pos := iterator.pull-one),IterationEnd),
          nqp::if(
            nqp::istype(pos,Int),
            self!accept(pos),
            self!handle-nonInt(pos)
          )
        );

        $!result.List
    }
}

class Array::PostcircumfixSlice does PostcircumfixSlice {
    method !accept(\pos --> Nil) {
        nqp::push($!result,$!iterable.AT-POS(pos))
    }
}

# @a[@i]
sub foo(\SELF, Iterable:D \positions, *%_) is raw {
    return postcircumfix:<[ ]>(SELF, positions.Int, |%_) 
      if nqp::iscont(positions);

    # Get the dispatch index
    my int $index;
    if nqp::getattr(%_,Map,'$!storage') {
        if nqp::istype(
             (my $lookup := Rakudo::Internals.ADVERBS_TO_DISPATCH_INDEX(%_)),
             X::Adverb
           ) {
            $lookup.what   = "slice";
            $lookup.source = try { SELF.VAR.name } // SELF.^name;
            return Failure.new($lookup);
        }

        # Good to go!
        $index = $lookup;
    }
    Array::PostcircumfixSlice.new(SELF).slice(positions.iterator)
}

my @a = <a b c d e f g h i j>;

my $then;
my @i = 1,2,3;

$then = now;
@a[@i] for ^100000;
say now - $then;

$then = now;
foo(@a, @i) for ^100000;
say now - $then;
```

lizmat: Hmmm... looking at a profile of this, the top one is: `dispatch:<!>` at **28%**.  So apparently the private method dispatch is not being optimized.  And the `!accept` method is not being optimized in any way, shape or form (stays red in the profile).

lizmat: Ok, that's an interesting development: turning the whole thing into just a class (without the role business), makes this run **faster** than my current implementation.  Apparently, role composition is making a number of optimizations not work?

lizmat:
@usev6 looks like this PR is breaking JVM.  Could you give me a clue on what is causing this?

usev6:
> @usev6 looks like this PR is breaking JVM. Could you give me a clue on what is causing this?

I didn't find the time to try this out, but from glancing over the code I would suspect some of the 'my constant' declarations with nqp::list in  src/core.c/Rakudo/Internals/PostcircumfixAdverbs.pm6 to be at fault. Or maybe the use of ```my uint16 @map``` in https://github.com/rakudo/rakudo/pull/4131/files#diff-21e81e42acf857f07e8faae1d82c1ad846d0c77aa138d625e6322c4c3c6664aaR133

I think we had to special case 'my constant' declarations with nqp::hash and nqp::list in the past. The issue reminds me of https://github.com/rakudo/rakudo/issues/2628 where @pmurias mentioned some weirdness with serialization. (The mentioned issue also was about explosion during generation of CORE.d.setting.jar.)

src/core.c/Rakudo/Internals/PostcircumfixAdverbs.pm6
```diff
@@ -0,0 +1,509 @@
...
+    }
+
+    # Constants to allow mapping of valid slice adverb combinations to
+    # a value that can be used in lookup table to get a dispatch table
+    # lookup value.
+    my constant SLICE_NO_ADVERBS = 0x0000;
+    my constant SLICE_DELETE     = 0x0001;  # :delete
+    my constant SLICE_EXISTS     = 0x0002;  # :exists
+    my constant SLICE_NOT_EXISTS = 0x0004;  # :!exists
+    my constant SLICE_KV         = 0x0008;  # :k
```
usev6, line 102: Very minor remark: The comment should be ```:kv``` here. (I just stumbled upon it.)

src/core.c/Rakudo/Internals/PostcircumfixAdverbs.pm6
```diff
@@ -0,0 +1,509 @@
...
+
+    # Array that contains dispatch table value for postcircumfix
+    # adverbs.  When mapping into this array results in 0, then it
+    # was an unsupported combination of adverbs.  Any other value
+    # indicates an index value into a dispatch table that should
+    # be initialized in the order in which these combinations are
+    # initialized here.
+    my constant @pc-adverb-mapper = do {
+        my int $i = -1;
+        my uint16 @map;
```
usev6, line 133: This seems to be the (only) place that make the JVM backend explode. If I just remove the ```uint16``` then rakudo compiles.

usev6: ... so it seems to be the same kind of problem as reported in https://github.com/rakudo/rakudo/issues/1666

lizmat: Hopefully addressed with https://github.com/rakudo/rakudo/commit/cbdc32bbc6

src/core.c/Rakudo/Internals/PostcircumfixAdverbs.pm6
lizmat: Addressed with 5a6f6cf907be731961f50a56

src/core.c/Rakudo/Internals/PostcircumfixAdverbs.pm6
usev6: I'm afraid, that won't help. It compiles with a non-native array (```my @map```). But probably that will cause problems in other places ...

I have no clue what goes wrong with native arrays on the JVM backend when compiling the core :(

lizmat:
This PR is winding to a close.  To summarise, these are the results (apart from being quite a lot faster):

- nested indexes are now handled correctly, think `@a[1,2,(3,4)]`
- all adverbs are now handled correctly, also with nested indexes
- error messages are now more specific (this causes some spectest breakage)

src/core.c/Rakudo/Internals/PostcircumfixAdverbs.pm6
```diff
@@ -296,7 +296,12 @@ augment class Rakudo::Internals {
         );
 
         # Perform the actual lookup and handling
+#?if moar
         my int $index = nqp::atpos_i(@pc-adverb-mapper,$bitmap);
+#?endif
+#?if !moar
+        my int $index = nqp::atpos(@pc-adverb-mapper,$bitmap);
```
usev6, line 303: If I'm not mistaken ```@pc-adverb-mapper``` comes out as an Array on the JVM backend. So using ```nqp::atpos``` does not work. But if this line and line 411 are changed to 
```
my int $index = @pc-adverb-mapper[$bitmap];
```
then things seem to look better.

lizmat: Fixed with https://github.com/rakudo/rakudo/commit/76d3e33811 I hope.  Thanks for the nudges!

lizmat:
Since a Blin run didn't show any obvious regressions, I intend to merge this in the next 24 hours.

src/core.c/Rakudo/Internals/PostcircumfixAdverbs.pm6
usev6: It turned out that this wasn't enough. I had to remove the ```ìnt``` from the above line, too: https://github.com/rakudo/rakudo/pull/4163

But now the JVM backend seems to be fine (no additional spectest failures).

PR #4135, opened 2020-12-23
Implement "is implementation-detail" for classes

lizmat:
This implements the "is implementation-detail" trait on classes.
It also implements a "is-implementation-detail" method on classes,
that will either return 0 if the class **is not** an implementation
detail, and 1 if the class *is* an implementation detail.

    class Foo is implementation-detail { }
    class Bar { }

    say Foo.^is-implementation-detail;   # 1
    say Bar.^is-implementation-detail;   # 0

### DIFF
--- a/src/Perl6/Metamodel/ConcreteRoleHOW.nqp
+++ b/src/Perl6/Metamodel/ConcreteRoleHOW.nqp
@@ -160,6 +160,10 @@ class Perl6::Metamodel::ConcreteRoleHOW
             nqp::null()
         }
     }
+
+    method is-implementation-detail($obj) {
+        @!roles[0].is-implementation-detail($obj)
+    }
 }
 
 # vim: expandtab sw=4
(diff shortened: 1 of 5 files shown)


vrurg:
For ease of introspection I'd  expect the method to be also declared on `CurriedRoleHOW` and `ConcreteRoleHOW`. For curried roles it's a simple re-delegation to `$!curried_role`. For concretizations it's also re-delegation but to the first elements of `@!roles`.

lizmat:
@vrurg Can you verify [this](https://github.com/rakudo/rakudo/pull/4135/commits/aa469bd08e3437b1c8157f1a42bb0bfcb369a04e) is what you meant?

vrurg (approved):

vrurg:
Looks good.

JJ:
Came a bit too late here... Semantically speaking, it might have been a better idea to Boolify the return value to True and False, unless this kind of questions are usually answered with 1 and 0 (in which case *all* of them should be Boolified too).

JJ:
I see also the return value is not specified, so this could be done without changing tests, actually. I would very much prefer anyhow to give a Bool type to the return value.

lizmat:
@JJ I would so too, but the default method lives in NQP-land, and NQP doesn't have a HLL `True` / `False`.  So the override *could* technically return `True`, but the default can not easily.  After consultation with @jnthn I decided to have both just return a simple `int`.

Issue #4117, opened 2020-12-15
routine run bad works in windows 10


interlab:
## The Problem
function **run** does not work correctly

```
7-Zip [64] 16.04 : Copyright (c) 1999-2016 Igor Pavlov : 2016-10-04



Command Line Error:
Unsupported command:
Files/7-Zip/7z.exe
```

## Expected Behavior
no error

## Actual Behavior
error

## Steps to Reproduce
```
my $zipd = run 'C:\Program Files\7-Zip\7z.exe',
    'a',
    '-t7z',
    'D:\b\t.7z',
    'C:\Users\quest\Desktop\t',
    '-mhe', '-mx7',
    '-ppassword';
```

## Environment
* Operating system: windows 10
* Compiler version (`perl6 -v` or `raku -v`): 

> Welcome to Rakudo(tm) v2020.11.
> Implementing the Raku(tm) programming language v6.d.
> Built on MoarVM version 2020.11.

**in rakudo-moar-2020.05.1-01-win-x86_64\rakudo-2020.05.1 has no error**

lizmat:
- What is the error?
- Does this happen with *any* `run` call, or just when calling this specific external command?

interlab:
an archive ашду should be created. the path to the executable file is not correctly processed.

lizmat:
@patrickbkr I just note that the path has spaces in it.  Could this be a quoting issue?

patrickbkr:
Yes it looks like the space is at fault. I have an idea what might be wrong.

patrickbkr:
This turned out to be a conceptual problem in the [Raku/problem-solving#20 solution](https://github.com/Raku/problem-solving/blob/master/solutions/language/Windows-process-argument-quoting.md).

The libuv API to spawn a process differentiates between program to execute and program as part of the command string passed to the process. That's technically necessary for some libuv trickery like PATH inspection to work. In the usual case the first must not be escaped, the second must be. Neither the NQP `spawnprocasync` opcode nor the Rakudo `run` and `Proc` APIs differentiate between the two. This mismatch gives us grief for two reasons:

1. We want to do Windows argument quoting in Rakudo, but the differentiation of the two first happens in Moar.
2. Rakudos current API currently provides no way to specify program and command string separately, but one needs to do that when using the `win-verbatim-args` flag and having a program path that needs escaping (like spaces in the programs path).

I see no easy way out. Especially for the second problem.

interlab:
```
[...]

PR #4138, opened 2020-12-23
Fix calling programs with spaces in their path on Windows

patrickbkr:
`procspawnasync` will now explicitly take the program name as a
separate argument. This allows us to pass a different value for
program name and arg0. That's necessary in some cases (e.g. a space
in the program name on Windows).

Fixes rakudo/rakudo#4117

Writing a test for this is difficult, as that requires executing an executable that has a character in its path that triggers quoting on Windows (typically that would be a space).

Depends on a Moar and NQP PR, so don't merge yet!

Moar PR: https://github.com/MoarVM/MoarVM/pull/1408
NQP PR: https://github.com/Raku/nqp/pull/689


lizmat:
Sadly, I don't have any feedback other than: great that you're fixing this.  I have no way to try this on Windows (it has been banned from our household, just as HP :-)

niner (approved):

PR #4143, opened 2020-12-28
Add "Solaris" to $?Perl.KERNELnames

nwc10:
Technically "Solaris 2", as SunOS 4 was retconned to "Solaris 1", but
that's all froth in the Marketing Layer, as the Engineering Layer still
reports "SunOS 5.x". eg

    $ uname -sr
    SunOS 5.11


lizmat:
Thanks!

PR #4146, opened 2020-12-30
Change IO::Path.new("") error message string

JJ:
Also DRY by putting it into a constant. Closes Raku/old-issue-tracker#6588 when accepted.


lizmat:
Thanks!

PR #4148, opened 2020-12-30
Implement `sec`, `asec` and `sech` in terms of other NQP trigonometric functions.

nwc10:
There is some inconsistency in MoarVM - as well as the usual trigonometric
functions `sin`, `cos`, `tan` and their inverses, *some* of the others are
implemented directly as NQP ops, but not all. So we have nqp::sec and
nqp::asec for secants, but no OPs for cosecants or cotangents (or their
inverses), with the result that these are implemented by Num.pm6 as
arithmetic relations using the OPs that exist.

Similarly, `sech` exists as a wrapper, but not `cosech` or `cotanh` (or any
of the inverses).

Looking at MoarVM's interp.c I see that the 3 OPs in question *don't* call
C library functions for the primitives - each is C code implementing the
operation in terms of arithmetic on other library functions - library
functions that we expose as NQP ops.

Hence it's much simpler if we move this arithmetic from MoarVM to the CORE:
1) We can remove the OPs from MoarVM
2) The code can now be JITted (everything that wrapper uses is in the JIT,
   whereas the 3 synthesised functions are not)
3) The 3 OPs don't seem to be implemented in the JVM or JS backends, so we
   also improve their test pass percentage :-)

### DIFF
--- a/src/core.c/Num.pm6
+++ b/src/core.c/Num.pm6
@@ -190,11 +190,11 @@ my class Num does Real { # declared in BOOTSTRAP
     }
     proto method sec(|) {*}
     multi method sec(Num:D: ) {
-        nqp::p6box_n(nqp::sec_n(nqp::unbox_n(self)));
+        nqp::p6box_n(nqp::div_n(1e0, nqp::cos_n(nqp::unbox_n(self))));
     }
     proto method asec(|) {*}
     multi method asec(Num:D: ) {
-        nqp::p6box_n(nqp::asec_n(nqp::unbox_n(self)));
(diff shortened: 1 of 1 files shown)


nwc10:
@jnthn: This eliminates the last (only?) use of these 3 in Rakudo. If we think that this is a good plan, my intent was to delete the relevant regression tests from NQP for these 3, and deprecate and remove the OPs from MoarVM.

jnthn (approved):
@nwc10 That plan sounds reasonable to me. I was curious if C libraries did expose the things we miss (out of concern that maybe we lose precision somehow by not using them directly), but it appears not.

lizmat:
Thanks!

PR #4152, opened 2021-01-01
Prefer api sorting over version sorting

ugexe:
When sorting distributions my intention was for the highest api to be preferred (with the highest version being used if there was a tie in api version). This change aligns raku behavior with those intentions instead of preferring version over api.


JJ:
Again, same errors... And again I'd like to request a test for this, as a Rakudo test.

niner:
Since we decided that api is numeric and sortable, this does make sense indeed.

lizmat:
Thanks!

PR #4153, opened 2021-01-01
Use less common names for installation shim arguments

ugexe:
A mostly unknown feature of the bin script shims CURI installs is
that it allows choosing the distribution to load the script from,
which allows loading an e.g. older version of some script. However
to do so required using the :$name :$auth :$ver :$api arguments
which prevents any installed raku bin script from taking any
arguments of those names; the shim scripts breaks things such as
providing a `bin/foo --ver` that outputs version information. This
changes the argument names to use something far less likely to be
used by arbitrary scripts (which is fine for rarely used options).

### DIFF
--- a/src/core.c/CompUnit/RepositoryRegistry.pm6
+++ b/src/core.c/CompUnit/RepositoryRegistry.pm6
@@ -285,10 +285,6 @@ class CompUnit::RepositoryRegistry {
     }
 
     method run-script($script, :$name, :$auth, :$ver, :$api) {
-        shift @*ARGS if $name;
-        shift @*ARGS if $auth;
-        shift @*ARGS if $ver;
-
         my @installations = $*REPO.repo-chain.grep(CompUnit::Repository::Installation);
         my @metas = @installations.map({ .files("bin/$script", :$name, :$auth, :$ver).head }).grep(*.defined);
         unless +@metas {
--- a/src/core.c/CompUnit/Repository/Installation.pm6
+++ b/src/core.c/CompUnit/Repository/Installation.pm6
@@ -51,8 +51,8 @@ __END__
 :endofperl
 ';
     my $perl_wrapper = '#!/usr/bin/env #perl#
-sub MAIN(:$name, :$auth, :$ver, *@, *%) {
-    CompUnit::RepositoryRegistry.run-script("#name#", :$name, :$auth, :$ver);
+sub MAIN(:$raku-dist-name, :$raku-dist-auth, :$raku-dist-ver, :$raku-dist-api, *@, *%) {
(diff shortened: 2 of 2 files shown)


JJ:
I don't really get the tests, although the two circle-ci tests failures are consistent. However, it would probably be best to add a test for this new feature as a Rakudo-only test.

niner:
I wonder if command line arguments are the right tool for this job at all. Looking for prior art in this problem space, [update-alternatives](https://linux.die.net/man/8/update-alternatives) comes to mind. Packages install their files into private directories like /usr/lib/postgresql13/bin/psql vs. /usr/lib/postgresql94/bin/psql and update-alternatives gets called by the package manager to maintain symlinks to the preferred version:
```
/usr/bin/psql -> /etc/alternatives/psql
/etc/alternatives/psql -> /usr/lib/postgresql13/bin/psql
```

This makes it easy to have things DWIM usually and also to run a certain version if need be. It interferes neither with command line processing nor with environment variables (which may be a but cumbersome to use on Windows anyway). Of course, the downside would be that we need directories within the repo's bin directory for each distribution shipping scripts. And these directories would need to retain user readable names.

ugexe:
> And these directories would need to retain user readable names.

At that point we might as well ditch CURI altogether for something that uses human readable names for everything (which I don't disagree with), but that is a much larger untaking ( ...which I've toyed with in https://github.com/ugexe/Perl6-CompUnit--Repository--Lib ) that I don't see happening any time soon. In that respect I think we should either change those argument names to something else or remove them altogether such that their existence doesn't help encourage their use in the mean time.

src/core.c/CompUnit/RepositoryRegistry.pm6
```diff
@@ -285,10 +285,6 @@ class CompUnit::RepositoryRegistry {
     }
 
     method run-script($script, :$name, :$auth, :$ver, :$api) {
-        shift @*ARGS if $name;
-        shift @*ARGS if $auth;
-        shift @*ARGS if $ver;
-
```
ugexe, line 291: This change probably isn't sufficient -- it would need to remove `raku-dist-name`, `raku-dist-ver`, etc from `@*ARGS` so they don't get used by the script it will run.

Of course the current logic is terribly broken as well since it incorrectly assumes the order of `@*ARGS` and will happily remove parameters such as `foo` or `bar` instead of `name` and `auth` based on the order.

niner:
Yeah, it's a mess. Well, we definitely don't want to occupy the currently used 
argument names in the future, so much is clear. So a rename will definitely 
improve things and be worthwhile.

ugexe:
@niner what are your feelings on removing them altogether? Needing logic to fix up @*ARGS sways me a bit in that direction...

niner:
I'm not really opposed to that. Since they are not even documented, I can 
hardly see how they are in use. They cover a fringe case anyway.

PR #4154, opened 2021-01-01
Switch spawnprocasync to use a separate arg for the program name

patrickbkr:
That's a lot more obvious than putting the program name at the start of the
args list.

Depends on MoarVM/MoarVM#1416 and Raku/nqp#691


patrickbkr:
I consider this PR and the other two this depends on to be ready for merge. They pass NQP `make test` and Rakudo `make test` and `make spectest`.

PR #4166, opened 2021-01-15
Implement explicitly setting arg0 in `Proc`s

patrickbkr:
On basically all operating systems the process calling API usually expects
arg0 to be the program being called. But not necessarily so. With this
change it is now possible to optionally explicitly set arg0 to a different
value. This is analogous to the node.js API.

Tests in: https://github.com/Raku/roast/pull/714
Docs in: https://github.com/Raku/doc/pull/3784

### DIFF
--- a/src/core.c/Proc/Async.pm6
+++ b/src/core.c/Proc/Async.pm6
@@ -101,6 +101,7 @@ my class Proc::Async {
     has $.w;
     has $.enc = 'utf8';
     has $.translate-nl = True;
+    has $.arg0;
     has $.win-verbatim-args = False;
     has Bool $.started = False;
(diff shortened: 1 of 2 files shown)


src/core.c/Proc/Async.pm6
```diff
@@ -130,6 +131,11 @@ my class Proc::Async {
 
     submethod TWEAK(--> Nil) {
         $!encoder := Encoding::Registry.find($!enc).encoder(:$!translate-nl);
+
+        unless $!arg0.defined {
+            $!arg0 = $!path;
```
lizmat, line 136: $!arg0 //= $!path

lizmat (approved):

PR #4168, opened 2021-01-16
Fix proclaim stack travesral

vrurg:
If a test function invoked not from the main thread but somewhere where there is no `.nqp` file found in the stack then `proclaim` dies when encounters the stack bottom by attempting `ends-with` method on `Mu`.


niner (approved):
Looks sensible. I do wonder though how one can trigger this case?

vrurg:
In my case it's `ok` in a supply block.

PR #4170, opened 2021-01-17
Make Array.ASSIGN-POS about 15% faster

lizmat:
By reducing the actual fast path to a single nqp::assign, and reworking
the tests into ternaries.  Also returns the actual container now rather
than a deconted value (just like Hash::ASSIGN-KEY does).

Eliminates the FAST-PATH private method.  Removes the index < 0 check
from the SLOW-PATH as that is now being checked in the actual ASSIGN-POS
method (but not really acted upon, just used as a flag to take the slow
path).


lizmat:
Verified that `Array.ASSIGN-POS` is still small enough to get inlined, so merging.

Issue #4175, opened 2021-01-21
Dateish `later` and `earlier` should accept multiple units


0racle:
## The Problem

Passing multiple time units to a `Date` or `DateTime` object dies `More than one time unit supplied`.

There's no reason why we can't support multiple time units.

## Expected Behavior

```perl6
my $dt = DateTime('2020-12-31T22:30:00')
say $dt.later(:1hour, :30minutes);
# OUTPUT: 2021-01-01T00:00:00Z
```

This would allow a Hash of time units to be `Slip`d to the methods,

```perl6
my %units = (hours => 1, minutes => 30);
$dt .= later(|%units);
```

FWIW, this is supported in Python

```python
from datetime import datetime, timedelta
dt = datetime(2020, 12, 31,  22, 30, 0)
print(dt + timedelta(hours=1, minutes=30))
# OUTPUT: datetime.datetime(2021, 1, 1, 0, 0)
```

0racle:
I had hoped I could PR this one myself, but it would require a bit of re-work, as currently the `later` method can simply exit early. 

The simplest solution would be call `later` recursively until `%units` was exhausted.
A naïve solution would try to wrap the current logic in a loop, which would create a new Date/DateTime object with each loop. (Even though I would guess most loops would be <3, it's not ideal)

An optimal solution would need to accumulate time and then add it once at the end, but I'm not sure how best to tackle it.

At the very least, I think it would allow us to `!move-die` method in `Date` and just die inside `later` if `%units` is empty.

jnthn:
My recollection is that the reason it doesn't at present is because named arguments are not ordered, and that the additions of different date units do not commute. Example:

```
$ raku -e 'say Date.new(2021, 1, 30).later(months => 1).later(days => 1)'
2021-03-01
$ raku -e 'say Date.new(2021, 1, 30).later(days => 1).later(months => 1)'
2021-02-28
```

To be clear, `.later(months => 1, days => 1)` and `.later(days => 1, months => 1)` are not possible to distinguish on the callee side, so you'd have to pick which one to do first, and that could be rather confusing.

If this problem doesn't exist for hours/minutes/seconds, only for months/days, then maybe a partial liberalization is possible without increasing risk (that is, permit multiple units only when the addition of those units commutes).

salortiz:
> My recollection is that the reason it doesn't at present is because named arguments are not ordered, …

[...]

PR #4181, opened 2021-01-27
Implement Dateish.earlier/later with multiple units

lizmat:
In response to https://github.com/rakudo/rakudo/issues/4175 and the
discussion following it, this implements being able to move a Date
or a DateTime with multiple units, as long as they are in a List of
Pairs, rather than as simple named parameters.

### DIFF
--- a/src/core.c/Dateish.pm6
+++ b/src/core.c/Dateish.pm6
@@ -177,7 +177,8 @@ my role Dateish {
         nqp::join($sep,$parts)
     }
 
-    method earlier(*%unit --> Dateish:D) {
+    proto method earlier(|) {*}
+    multi method earlier(Dateish:D: *%unit --> Dateish:D) {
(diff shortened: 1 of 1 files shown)


vrurg (approved):

0racle:
I've tested the PR, and this works fine.

```raku
my $dt = DateTime('2020-12-31T22:30:00');
say $dt.later((:1hour, :30minutes));
```

However this - which is I think something people would expect to work - gives an LTA error

```raku
say $dt.later([:1hour, :30minutes]);
# This type cannot unbox to a native string: P6opaque, ContainerDescriptor::Untyped
```

Any reason why Array of Pairs can't work?

lizmat:
Hmmm... not sure, investigating.  Probably a missing `nqp::decont`

0racle:
Looks good now.

PR #4182, opened 2021-01-27
Provide single arg semantics for @array = QuantHash

lizmat:
Inspired by

  https://donaldh.wtf/2021/01/if-sets-would-dwim/

and the subsequent discussion at

  https://www.reddit.com/r/rakulang/comments/l5qub2/if_sets_would_dwim/

this provides an Array.STORE candidate that will handle a single QuantHash
argument with single argument semantics, just like `for` does.

### DIFF
--- a/src/core.c/Array.pm6
+++ b/src/core.c/Array.pm6
@@ -297,6 +297,17 @@ my class Array { # declared in BOOTSTRAP
         );
         nqp::p6bindattrinvres(self,List,'$!reified',buffer)
     }
+    multi method STORE(Array:D: QuantHash:D \qh --> Array:D) {
+        qh.iterator.push-all(
+          ArrayReificationTarget.new(
+            (my \buffer = nqp::create(IterationBuffer)),
+            nqp::decont($!descriptor)
+          )
+        );
+        nqp::bindattr(self,List,'$!todo',Mu);  # exhausted
+        nqp::bindattr(self,List,'$!reified',buffer);
+        self
+    }
     multi method STORE(Array:D: Mu \item --> Array:D) {
         nqp::push(
           (my \buffer = nqp::create(IterationBuffer)),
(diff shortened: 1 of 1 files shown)


vrurg (approved):

0racle:
I will reiterate my reasons for why I'm not sold on this, in that it makes behaviour inconsistent.

I often initialise a new Array with a single element (which I may push onto later) like so

```perl6
my @a = $foo;
```

but with this proposed behaviour, `@a` could have more than one element _only_ if `$foo` is a Set. This is inconsistent and unintuitive.

You might even say that making these return 2 different things

```perl6
my @a = <a b c d e> (-) <b d>;
my $a = <a b c d e> (-) <b d>;
```

takes us a step in the direction of Perl 5's scalar/list context semantics. I love Perl 5, but context-sensitive behaviour doesn't fit Raku.

If there's enough support for this change, I'll happily accept it... but I just want to make sure that it has been adequately considered.

vrurg:
@0racle I haven't had chance to try this PR in action, but if it works the same for quant hashes as it does for, say, arrays, then there is no inconsistency:

```raku
my $a = [1,2,3];
my @a = $a; # [[1, 2, 3]]
my @b = $a<>; # [1, 2, 3]
```

I expect it to be the same for quant hashes. I.e.

```raku
my $s = set(<a b c>);
my @a = $s; # [Set(a b c)]
my @b = $s<>; # [:a, :b, :c]
```

And if you need to explicitly itemize a quant hash then you can always use an extra comma for that, as we currently can with, say, positionals:

```raku
my @a = [1, 2, 3]; # [1, 2, 3]
my @b = [1, 2, 3],; # [[1, 2, 3]]
```

0racle:
> I expect it to be the same for quant hashes

I've just tested this PR and no, it's not the same. It does this

```perl6
my $s = set(<a b c>);
my @a = $s; # [:a, :b, :c]
```

Currently - provided `$a` is a Scalar container - `my @a = $a` always produces an Array with a single item (unless I'm missing something). This PR changes that.

vrurg:
@0racle Have you tried it with 9ebecb0?

0racle:
Yes, I tried it... works as per your examples.

```perl6
my $s = set(<a b c>);
# Set $s = Set.new("c","b","a")
my @a = $s; # [Set(a b c)]
# Array @a = [Set.new("c","b","a")]
my @b = $s<>; # [:a, :b, :c]
# Array @b = [:c, :b, :a]
```

PR #4185, opened 2021-01-29
Provide $*REPO with an associative interface

lizmat:
So that you can access your home repo with $*REPO\<home>, and the core
repo with $*REPO\<core>, etc.

### DIFF
--- a/src/core.c/CompUnit/Repository.pm6
+++ b/src/core.c/CompUnit/Repository.pm6
@@ -59,7 +59,7 @@ role CompUnit::Repository {
 }
 
 multi sub postcircumfix:<{ }>(CompUnit::Repository:D \repo, Str() $key) {
-    repo.repo-chain.first(quietly *.?name eq $key)
+    repo.repo-chain.first({ quietly (.?name // .id) eq $key })
 }
 
 # vim: expandtab shiftwidth=4
--- a/src/core.c/CompUnit/Repository.pm6
+++ b/src/core.c/CompUnit/Repository.pm6
@@ -58,4 +58,8 @@ role CompUnit::Repository {
     }
 }
 
+multi sub postcircumfix:<{ }>(CompUnit::Repository:D \repo, Str() $key) {
+    repo.repo-chain.first(quietly *.?name eq $key)
+}
+
 # vim: expandtab shiftwidth=4


jonathanstowe (approved):

ugexe:
Do we really need a shorter version of the existing api we provide for doing this?

```
raku -e 'say CompUnit::RepositoryRegistry.repository-for-name("site")'
inst#/Users/ugexe/.rakubrew/versions/moar-2020.11/install/share/perl6/site
```

ugexe:
One example of why this may not be a good idea is how people think of `%?RESOURCES` as a hash because its typically accessed associatively, and wonder why they can't call e.g. `.keys` on it.

ugexe:
Grepping on multiple fields makes me even more wary of this fwiw

lizmat:
re `repository-for-name`:

- the existence was not directly known to me, now that I've seen it, I remember seeing it before
- It is currently *not* documented, so unclear if that is actually a public interface.
- `CompUnit::RepositoryRegistry.repository-for-name("site")` is quite a mouthful
- It currently does *not* know about the AbsolutePath / NQP / Perl5 repos.

@ugexe Are we sure that any additional repos added are added internally to `$customlib` ?

ugexe:
> CompUnit::RepositoryRegistry.repository-for-name("site") is quite a mouthful

Thats fine... it is not common to need to do these things regularly (and I say this as one of the heaviest consumers of this functionality)

> It currently does not know about the AbsolutePath / NQP / Perl5 repos.

Yeah, each repo works differently and this PR is basically only useful for CURIs. Searching by `id` is generally of no real use
```
raku -e 'say CompUnit::RepositoryRegistry.repository-for-spec("foo")'
file#/Users/ugexe/repos/zef/foo
```

lizmat:
Re: associative interface giving the wrong signal: the fact that the sigil is `$` instead of `%` (as is with `%?RESOURCES`) should be indication enough?

lizmat:
Huh?  So `CompUnit::RepositoryRegistry.repository-for-spec("foo")` will actually *create* a `CompUnit::Repository::FileSystem` for "foo"?  I thought it was purely introspective?

lizmat:
In any case, we definitely are lacking documentation in this area :-(

jonathanstowe:
> Do we really need a shorter version of the existing api we provide for doing this?
> 
> ```
> raku -e 'say CompUnit::RepositoryRegistry.repository-for-name("site")'
> inst#/Users/ugexe/.rakubrew/versions/moar-2020.11/install/share/perl6/site
> ```

Ahah!  I have to admit that this is partially my fault as I had a need to get the prefix path for `site` for the current `raku` installation and presented a rather over-elaborate snippet for doing that in blissful ignorance of the above API.  I may have even known about it once.  So yes, documentation is probably the problem that needs to be solved.

lizmat:
It appears that this is a documentation issue, so closing this PR now.

JJ:
> It appears that this is a documentation issue, so closing this PR now.

I would appreciate if somebody raised this issue...

Issue #4187, opened 2021-01-31
[JVM] Creating DateTime objects from positional arguments is busted


usev6:
## The Problem
Creating a DateTime object from positional arguments fails with a NullPointerException on the JVM backend:
```
$ ./rakudo-j --ll-exception -e 'dd DateTime.new(2021,1,3,0,0,0)'
java.lang.NullPointerException
  in new-from-positional (gen/jvm/CORE.c.setting:58237)
  in new (gen/jvm/CORE.c.setting:58261)
  in new (gen/jvm/CORE.c.setting:58260)
  in <unit> (-e:1)
  in <unit-outer> (-e:1)
  in eval (gen/jvm/stage2/NQPHLL.nqp:1216)
  in <anon> (gen/jvm/stage2/NQPHLL.nqp:1328)
  in command_eval (gen/jvm/stage2/NQPHLL.nqp:1325)
  in command_eval (gen/jvm/Compiler.nqp:119)
  in command_line (gen/jvm/stage2/NQPHLL.nqp:1309)
  in MAIN (gen/jvm/rakudo.nqp:133)
  in <mainline> (gen/jvm/rakudo.nqp:121)
  in <anon> (gen/jvm/rakudo.nqp)
```

The problem goes away if 0c8d238a8c is reverted.

A golfed version of the problem
```
$ ./rakudo-j -e 'use nqp; sub foo(Int() $x) { nqp::islt_I($x, 0) }; say foo(7)'
java.lang.NullPointerException
  in sub foo at -e line 1
  in block <unit> at -e line 1
```

As far as I can see there is a problem with using coercion to Int (```Int() $n```) together with NQP ops for bigints. This one also fails:
```
$ ./perl6-j -e 'use nqp; sub foo(Int() $x) { nqp::isbig_I($x) }; say foo(7)'
java.lang.NullPointerException
  in sub foo at -e line 1
  in block <unit> at -e line 1
```

I'm not sure if this needs a fix in Rakudo or in NQP. (If it is the latter, commit https://github.com/Raku/nqp/commit/7190914a19 might be relevant. The NPE happens in https://github.com/Raku/nqp/blob/master/src/vm/jvm/runtime/org/raku/nqp/runtime/Ops.java#L6763, because the result from ```((P6OpaqueREPRData) type.st.REPRData).flattenedSTables[hint]``` is null.)

## Environment
Compiler version (`perl6 -v` or `raku -v`):
```
$ ./rakudo-j --version
[...]

PR #4189, opened 2021-01-31
[JVM] Add bandaid for DateTime from positionals

usev6:
Creating DateTime objects from positional arguments is currently busted
on the JVM backend: https://github.com/rakudo/rakudo/issues/4187.

This workaround should be removed once the underlying problem is fixed.

### DIFF
--- a/src/core.c/DateTime.pm6
+++ b/src/core.c/DateTime.pm6
@@ -173,14 +173,34 @@ my class DateTime does Dateish {
             %extra,
     --> DateTime:D) {
         self!oor("Month",$month,"1..12")
+#?if jvm
+          if nqp::islt_I($month.Int,1) || nqp::isgt_I($month.Int,12);
+#?endif
+#?if !jvm
           if nqp::islt_I($month,1) || nqp::isgt_I($month,12);
+#?endif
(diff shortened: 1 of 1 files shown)


lizmat:
So basically you're saying that coercions to types in parameters do not work on the JVM backend?  Surely that should create more issues than just this one?

lizmat:
If we're going to bandaid this, I'd rather prefer that we use the old code for the JVM case.  For the future, when we don't actually need to use nqp ops anymore to get better performance.

Working on that now.

lizmat:
f627ccb9046d46755e0d is the JVM bandaid with the original code (well, apart from the call to DAYS-IN-MONTH).  @usev6++ for this PR.  Closing it now.

usev6:
Thanks @lizmat for taking care of this. 

> So basically you're saying that coercions to types in parameters do not work on the JVM backend? Surely that should create more issues than just this one?

I don't really understand what goes wrong, but I think it's related to the ```bigint``` nature of values of ```Int```objects. Otherwise coercions to types in parameters seem to work fine. I'll try to do further investigating in https://github.com/rakudo/rakudo/issues/4187.

PR #4201, opened 2021-02-12
Add support for whitespace-separated CLI options

codesections:
This PR resolves the exiting TODO item:
> Allow both = and space before argument of double-dash args

More specifically, before this commit, a user could pass a `foo` command that took a named `bar` values in nine ways:

    $ foo --bar=value
    $ foo -bar=value
    $ foo :bar=value
    $ foo --bar      # equivalent to --bar=True
    $ foo -bar
    $ foo :bar
    $ foo --/bar     # equivalent to --bar=False
    $ foo -/bar
    $ foo :bar

(The ways that involve using a single `-` do not work when bundling is turned on, since that would be ambiguous.)

This commit adds three more ways:

    $ foo --bar value
    $ foo -bar value
    $ foo :bar value

(Note that this does not limit space separated values to the `--bar` form.  I'd be happy to add an extra check if anyone would like it, but I didn't see a reason to do more checks to add that limit)

Supporting space-separated options raises one common difficulty.  If `foo` accepts both an option and a positional argument, then the following expression is initially ambiguous:

    $ foo --bar baz

Is that a option `:bar("baz")` and no positional argument?  Or is it `:bar(True)` and a positional argument of `"baz"`?  This PR resolves that ambiguity by treating all values after an option as values for that option _unless_ they start with a `-` or `:` or if the `--bar` is type constrained to be a `Bool` (which indicates that it isn't an option at all, but a flag.)  Thus, if `foo` has the signature

    MAIN(:$bar, Bool :$v, $pos)

It will be parsed as follows:

   $ foo --bar baz     # $bar = "baz"
   $ foo --bar -v baz  # $bar = True;  $v = True; $pos = "baz"

If a user wants to pass a positional argument after an option, they can use `--` to end parsing of non-positional arguments (which was supported prior to this PR).

I looked over the current tests and did not see anywhere that looked like it should include a test for this functionality.  But if anyone would like me to include a test, I'd be happy to do so – just point me in the right direction!

If this PR generally looks good, I'll plan to submit another addressing the CSV TODO in the same file.  Between addressing those two TODOs and #4113, Raku is adding in the last few pieces in its CLI story!

### DIFF
--- a/src/core.c/Main.pm6
+++ b/src/core.c/Main.pm6
@@ -113,7 +113,7 @@ my sub RUN-MAIN(&main, $mainline, :$in-as-argsfiles) {
 
                 # explicit value (option) using whitespace
                 elsif $arg ∈ @script-options && !@args[0].starts-with('-'|':') {
-                    %named.push: $arg => @args.shift
+                    %named.push: $arg => thevalue(@args.shift)
                 }
 
                 # implicit value (flag)
--- a/src/core.c/Main.pm6
+++ b/src/core.c/Main.pm6
@@ -77,11 +77,19 @@ my sub RUN-MAIN(&main, $mainline, :$in-as-argsfiles) {
               !! coercer(a)
         }
 
-        my @script-options = &main.candidates.map(*.signature.params).flat.map( {
-            when .named && .type !~~ Bool { |.named_names } # Bools indicate flags, not options
-            default                       { Empty }
-        });
-
+        my %options-with-req-arg = Hash.new;
+        my %short-opts = Hash.new;
+        for &main.candidates {
+            for .signature.params -> $param {
+                if !$param.named { next }
+                for $param.named_names -> str $name {
+                    if nqp::iseq_i(nqp::chars($name), 1) { %short-opts.push($name => True) }
+                    my int $accepts-true = $param.type.ACCEPTS: True;
+                    for $param.constraint_list { $accepts-true++ if .ACCEPTS: True}
+                    if !$accepts-true { %options-with-req-arg.push($name => True) }
+                }
+            }
+        }
         while @args {
             my str $passed-value = @args.shift;
 
@@ -91,49 +99,61 @@ my sub RUN-MAIN(&main, $mainline, :$in-as-argsfiles) {
                 last;
             }
 
-            # no longer accepting nameds
-            elsif $no-named-after && nqp::isgt_i(nqp::elems($positional),0) {
+            if ($no-named-after && nqp::isgt_i(nqp::elems($positional),0)) {
                 nqp::push($positional, thevalue($passed-value));
+                nqp::push($positional, thevalue($_)) for @args;
+                last;
             }
 
-            # named
-            elsif $passed-value
-              ~~ /^ ( '--' | '-' | ':' ) ('/'?) (<-[0..9\.]> .*) $/ {  # 'hlfix
-                my str $arg = $2.Str;
-                my $split  := nqp::split("=",$arg);
-
-                # explicit value (i.e., option) using =
-                if nqp::isgt_i(nqp::elems($split),1) {
-                    my str $name = nqp::shift($split);
-                    die "Can't combine bundling with explicit arguments" if $bundling && nqp::iseq_s($0.Str, '-') && $name.chars > 1;
-                    %named.push: $name => $1.chars
-                      ?? thevalue(nqp::join("=",$split)) but False
-                      !! thevalue(nqp::join("=",$split));
+            my str $optstring;
+            my int $negated   = 0;
+            my int $short-opt = 0;
+            # long option
+            if nqp::eqat($passed-value, '--', 0) {
+                if nqp::eqat($passed-value, '/', 2) {
+                    $optstring = nqp::substr($passed-value, 3);
+                    $negated = 1;
                 }
-
-                # explicit value (option) using whitespace
-                elsif $arg ∈ @script-options && !@args[0].starts-with('-'|':') {
-                    %named.push: $arg => thevalue(@args.shift)
-                }
-
-                # implicit value (flag)
-                else {
-                    if $bundling && nqp::iseq_s($0.Str, '-') {
-                        die "Can't combine bundling with explicit negation" if $1.chars && $arg.chars > 1;
-                        my @chars = nqp::split('',$arg);
-                        for @chars -> $char {
-                            %named.push: $char => !($1.chars);
-                        }
-                    }
-                    else {
-                        %named.push: $arg => !($1.chars);
-                    }
-                }
+                else { $optstring = nqp::substr($passed-value, 2) }
             }
-
+            # short option
+            elsif nqp::eqat($passed-value, '-', 0) || nqp::eqat($passed-value, ':', 0) {
+                $short-opt = 1;
+                if nqp::eqat($passed-value, '/', 1) {
+                    $optstring = nqp::substr($passed-value, 2);
+                    $negated = 1;
+                }
+                else { $optstring = nqp::substr($passed-value, 1) }
+            }
             # positional
             else {
                 nqp::push($positional, thevalue($passed-value));
+                next;
+            }
+
+            my $split  := nqp::split("=",$optstring);
+            $optstring = nqp::shift($split);
+            my str $arg = nqp::join('=', $split);
+            if nqp::existskey(%options-with-req-arg, $optstring) {
+                if !$arg { $arg = @args.shift // '' }
+                %named.push: $optstring => ($negated ?? thevalue($arg) but False !! thevalue($arg));
+            }
+
+            elsif $short-opt && $bundling {
+                my int $cursor = 1;
+                my str $short-opt = nqp::substr($optstring, 0, 1);
+                my str $next      = nqp::substr($optstring, 0, 1);
+                # TODO: simplify?
+                while ($next && nqp::existskey(%short-opts, $next)) {
+                    %named.push: $short-opt => True;
+                    $short-opt = nqp::substr($optstring, $cursor++, 1);
+                    $next      = nqp::substr($optstring, $cursor,   1);
+                }
+            }
+            else {
+
+                if !nqp::elems($split) { %named.push: $optstring => ($negated ?? False !! True) }
+                else { %named.push: $optstring => $negated ?? thevalue $arg but False !! thevalue $arg }
             }
         }
         Capture.new( list => $positional.List, hash => %named )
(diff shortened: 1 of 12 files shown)


ugexe:
Should all of these not be the equivalent?
```
$ raku -e 'sub MAIN(:$int) { say $int }' --int 1
1

$ raku -e 'sub MAIN(Int :$int) { say $int }' --int 1
Usage:
  -e '...' [--int=<Int>]

$ raku -e 'sub MAIN(Int :$int) { say $int }' --int=1
1
```

> If a user wants to pass a positional argument after an option, they can use -- to end parsing of non-positional arguments 

If the current behavior doesn't require this then this functionality might need to go behind a feature flag -- if sufficient existing examples / instructions already expect the original behavior then we shouldn't break it.

codesections:
> Should all of these not be equivalent? [snip]

You are absolutely correct.  My code was still type checking space-separated arguments – which of course makes no sense because they will always be strings.  Thanks for the catch; I've fixed the issue.

> If the current behavior doesn't require this then this functionality might need to go behind a feature flag

You might be right about that; I'd be interested to hear other's thoughts.  On the one hand, it's not a change to the semantics of _Raku_ itself, and I'm not sure what API stability guarantees we make there.  And the issue would only come up in the (in my experience) rare case where a script has something that can validly be *both* a flag and an option (that is, where something has meaningful behavior both when not passed an argument _and_ when passed one).

One the other hand, you're right that this would break code if all those conditions are met.  I can think of four options:

1. A literal feature flag, gating this until a future release.
2. Adding space-separated options to the `%*SUB-MAIN-OPTS` configuration hash, with them off by default (at least for now)
3. Changing the default behavior so that any currently valid code would remain valid (that is, making `foo --color full` parse into `:color(True); $positional = "full"` whenever `MAIN` accepts a positional argument.
4. Accept the slight edge-case change.

Of these, I'm inclined most strongly towards 2 or 4, and against 3 – it seems like it would make the code more complicated _and_ make the behavior more surprising.  But I don't have firm thoughts and would be interested in any views you/others might have.

ugexe:
> rare case where a script has something that can validly be both a flag and an option

It might not be as rare as you think. zef allows e.g. `--wrap`,`--/wrap`,`--wrap=60` to enable wrapping, disable wrapping, and enable wrapping with an explicit width. And `--update`,`--/update`, `--update=cpan` to enable updates, disable udpates, and to enable updates only for cpan. And thats just off the top of my head for only one of my projects.

codesections:
That's a good point, and has given me a bit more to think about.  I was doing some more reading about correct [conventions for command line arguments](https://nullprogram.com/blog/2020/08/01/) and that's causing me to rethink the approach I took in this PR.

Specifically, that guidance distinguishes between **options that take arguments** (where the option-argument is required) and **optional option arguments**.  That guidance states that a space should only be allowed when option-argument is required; with optional-option arguments, the space indicates that the argument wasn't passed an option.

I hadn't really considered this approach, since I almost never use optional option arguments – I'd probably have written `--update` as one flag and then had a separate `--update-target` option.  But now that I've read the description, this approach makes a lot of sense.  Adopting it would not only ensure that we're fully backwards-compatible with existing `=` syntax, it would also help bring our implementation more in line with the one in GNU getopt (which isn't a requirement by any means, but would likely make Raku less surprising.)

In any event, I think this means I should take a different approach.  I'm going to convert this PR to a draft until I can spend a bit more time with the getopt man page and then take a different stab at a revised implementation.

Thanks for the help!

JJ (commented):
I can get really obnoxious on this, but if you add new functionality, you should add *tests* for that new functionality. These are not spec, I guess, so they would be local Rakudo tests.

codesections:
> if you add new functionality, you should add tests for that new functionality.

I very much agree and, as I said in the OP, the only reason I didn't add tests was that I wasn't sure where to put them/the best approach to take.

> These are not spec, I guess, so they would be local Rakudo tests.

That was my first thought as well, but your comment inspired me to do a bit more digging and I think I found a way that this can/should be in the spec.  Specifically, [S19](https://design.raku.org/S19.html) states that "Values are passed to options with the following syntax --option=value or --option value."  Of course, I know that the Synopsis aren't official unless/until they make it into Roast and there isn't anything in the [relevant section of Roast](https://github.com/Raku/roast/tree/master/S19-command-line-options).  But it seems like it would fit well there, so I'll plan a companion PR for Roast.

src/core.c/Main.pm6
```diff
@@ -78,6 +77,11 @@ my sub RUN-MAIN(&main, $mainline, :$in-as-argsfiles) {
               !! coercer(a)
         }
 
+        my @script-options = &main.candidates.map(*.signature.params).flat.map( {
```
MasterDuke17, line 80: Truly a minor nitpick, but you have a space between the opening `(` and `{` here, but not between the closing `}` and `)`.

src/core.c/Main.pm6
```diff
@@ -107,7 +111,12 @@ my sub RUN-MAIN(&main, $mainline, :$in-as-argsfiles) {
                       !! thevalue(nqp::join("=",$split));
                 }
 
-                # implicit value
+                # explicit value (option) using whitespace
+                elsif $arg ∈ @script-options && !@args[0].starts-with('-'|':') {
```
MasterDuke17, line 115: If this is running at startup and we really want to be concerned about performance, Junctions are very slow. I'd recommend `&& !(@args[0].starts-with('-') || @args[0].starts-with(':'))`

codesections: Thanks!  I've actually been refactoring this PR quite a bit to address the more substantive issues raised above (and avoid any incompatible changes).  So this line won't be in the PR. and (I think) the new version should generally be faster.  But I appreciate the tip and will keep an eye out for similar performance issues. :+1:

codesections:
Updating the conversation here with the most recent commit message.  I believe this PR is now ready to be merged, along with the companion tests in Raku/roast/pull/716.  @ugexe I believe this solves the backwards-incompatibility issues you noted above.  @moritz, I thought you might be interested in this PR, because it is based in part on an IRC conversation between you and Larry from 2011.

-----

This commit adds an additional feature to the command-line programs Raku generates: users may now pass arguments to options separated by whitespace.  That is, if a program takes a --port option, users can now call it with `--port 4242` in addition to with `--port=4242`. This behavior is described in S06 and S19; it also makes the syntax for invoking Raku CLI's more similar to standard Linux utilities.

This feature was previously discussed in 2011, but was not implemented because no resolution could be found to two issues.  First, whitespace delimited arguments can be ambiguous: does `--profile foo` pass `foo` as an argument to `--profile` or as a positional argument to the program?  (One potential solution to this ambiguity is to inspect whether `foo` starts with `-`.  But this can create hard-to-debug issues for unusual `foo`s and is generally not reliable enough for use in scripts.) Second, is there a way to pass multiple arguments to a single option that allows users to use shell globing?  That is, can a call like `--input-files *.jpg` be made to DWIM?

This commit resolves these issues in a way that is inspired by getopt(1) and thus should be familiar to Linux users.  It also avoids making any changes that would be incompatible with existing Raku CLIs. Before discussing the details, here is an example.  Consider a `demo` program with the following source:

    subset name of Any where Str|True;
    subset port of Str;
    multi MAIN(
        $file,
        name :$profile,    #= Write profile information to a file
        port :$debug-port, #= Listen for debugger connections on the specified port
        Bool :v($verbose), #= Display verbose output

    ) {}
    multi MAIN("--process-files", *@images) {}

This program generates the following usage message:

    Usage:
      demo [--profile[=name]] [--debug-port=<port>] [-v] <file>
      demo --process-files [<images> ...]

        --profile[=name]       Write profile information to a file
        --debug-port=<port>    Listen for debugger connections on the specified port
        -v                     Display verbose output

The following are valid ways to call `demo`:

    demo --profile ~/foo
    demo --profile=/tmp/bar ~/foo
    demo --debug-port 4242 ~/foo
    demo --debug-port=4242 ~/foo
    demo -v ~/foo
    demo --process-files *.jpg

These, however, are not valid

    demo --profile /tmp/bar ~/foo
    demo --debug-port ~/foo

The first is invalid because `/tmp/bar` and `~/foo` are both parsed as positional arguments, which means `demo` was called with too many positional arguments.  The second is invalid because `~/foo` is parsed as an argument to `--debug-port`, and thus `demo` lacks the required positional argument.

Here's how it works:

Raku now distinguishes between three types of options:

 * Boolean options (like `-v`), which _never_ take an argument; they
   are ether present or absent.
 * Options with a mandatory argument (like `--debug-port`), which
   always take an argument.  If you give them an argument with `=`,
   they will use that; if not, they'll take the following argument.
 * Options with an optional argument (like `--profile`), which are
   valid both with and without an argument.  You can _only_ give these
   arguments an option with the `=` syntax; if there is a space after
   the option, that means it was called without an argument.

And here's the signature that produces each type of argument:

 * Boolean options: A `Bool` type constraint.
 * Options with a mandatory argument: A type that does not `.ACCEPT` a
   `Bool`
 * Options with an optional argument: A type that `.ACCEPTS` a
   `True` (because passing an option without an argument is equivalent
   to passing `True`)

This fully resolves the ambiguity discussed above in a way that should feel fairly intuitive to most users (despite how long
explaining it takes!) because it basically boils down to "space separated arguments are fine, unless you don't need to specify a
value; then you've got to use =".  This is the same solution employed by getopt(1) and should be familiar from standard utilities like `ls` and `cp`.

However, distinguishing between options with mandatory/required arguments introduces one potential trap into the language: Types from which `Bool`s inherit will produce an argument that takes an optional argument.  This group includes some numeric types, which may trip up some users.  Thus `sub MAIN(Int :$debug-port)` requires `--debug-port` to be called with `=` (or get a default value of 1), which may surprise some users.  After this feature is added, I will update the documentation to clearly note this issue.

The code in this commit takes an additional step to mitigate this footgun: it modifies the usage message to indicate which option arguments are optional by marking them with `[...]` (as was already done for optional positional arguments).

Finally, the desire to allow multiple arguments to be passed to a single option can be accommodated using the subcommand syntax, as shown in the `--process-files` example above.

Because argument parsing is so often used in interactive CLI programs, it is important that it be as fast as possible and I have written this commit perform as well as possible.  As a result I believe that, despite the increased functionality, the revised code runs ~1ms faster (out of ~6ms previously spent parsing arguments).  This decrease, however, is within the margin of error reported by my benchmarking tool and reflects only one machine.

I have prepared a companion commit that adds 10 tests to Roast (and removes 5 tests marked `todo`) that test this functionality.

MasterDuke17:
> I have prepared a companion commit that adds 10 tests to Roast (and removes 5 tests marked `todo`) that test this functionality.

Looks like there are some rakudo tests that need addressing also https://dev.azure.com/Rakudo/rakudo/_build/results?buildId=1108&view=logs&j=1f3a7134-3b80-549a-eaff-49041f6b71c8&t=31ee51d3-bad8-5596-9e13-6187996f79fc&l=2655

codesections:
> Looks like there are some rakudo tests that need addressing

Thanks, fixed.

codesections:
After taking another look at this, I realized that, despite passing all tests, it does not handle all bundled arguments correctly.  I also noticed that the, due to the rebase, the diff for this PR is pulling in some unrelated changes.  Due to these two issues, I'm going to close this PR for now and re-open a clean PR once I have resolved the issue with bundled arguments (and added tests to prevent it from reoccurring).

PR #4202, opened 2021-02-13
Add multilevel list output for Pod::To::Text

hopkinsza:
Improves the output of Pod::To::Text by adding indentation for pod6 items based on their level.

For example, given the following script:

```
use Pod::To::Text;

=begin pod
=item item
=item2 item2
=item3 item3
=item item
=item2 item2
=end pod

say pod2text($=pod);
```

The current output would be:

```
  * item

  * item2

  * item3

  * item

  * item2
```

After this modification, the output is:

```
  * item

    * item2

      * item3

  * item

    * item2
```


JJ (approved):
I don't think Azure errors really are related to this, so I guess it's OK.

lizmat:
Thank you!

Issue #1632, opened 2018-03-20
Meta R op messes up fake infix colonpairs


zoffixznet:
```
12:13 		m: class {}.new :foo(42 - 3)
12:13 	camelia 	rakudo-moar ec5416a9a: ( no output )

12:13 	Zoffix 	m: class {}.new :foo(42 R- 3)
12:13 	camelia 	rakudo-moar ec5416a9a: OUTPUT: «Default constructor for '<anon|1>' only takes named arguments␤  in block <unit> at <tmp> line 1␤␤»
```

atroxaper:
Please, take a look at [fix](https://github.com/rakudo/rakudo/pull/4204) and [test](https://github.com/Raku/roast/pull/715);

atroxaper:
@lizmat close this, please.

patrickbkr:
Fixed by @atroxaper, so closing. Thanks!

PR #4204, opened 2021-02-14
Fix deleting colonpair with R-metaop in value by the Optimizer

atroxaper:
Optimizer was replacing the call op with METAOP_REVERSE metaop
to the METAOP_REVERSE call only. If the original op had named
parameter then it was lost. After such optimization a colon pair
become only the value.

Issue: #1632 
Tests: [#715](https://github.com/Raku/roast/pull/715)

### DIFF
--- a/src/Perl6/Optimizer.nqp
+++ b/src/Perl6/Optimizer.nqp
@@ -2337,8 +2337,11 @@ class Perl6::Optimizer {
       elsif self.op_eq_core($metaop, '&METAOP_REVERSE') {
         return NQPMu unless nqp::istype($metaop[0], QAST::Var)
           && nqp::elems($op) == 3;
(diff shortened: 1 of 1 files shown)


atroxaper:
@timo Hello. Do you wait me to fix something in autochecks or is it expected behaviour?

MasterDuke17:
> Hello. Do you wait me to fix something in autochecks or is it expected behavior?

I re-ran the failed Azure pipeline and the build+tests are fine, but it can't publish the artifacts because they're duplicated. The AppVeyor failure was just in one of the inlining tests, which are performance tests and the slow VM is to blame.

atroxaper:
@lizmat @MasterDuke17 @timo Thank you!

timo:
no, thank **you**

PR #4208, opened 2021-02-20
Fix single item assigment at List creation,

salortiz:
Add missing candidate for `STORE(List:D: Mu \item, :INITIALIZE($)! --> List:D)`
Without it `my @L is List = 'foo';` fails. (Resulting `List` is empty).


salortiz:
Roast tests added in https://github.com/Raku/roast/pull/717

lizmat (approved):

Issue #3029, opened 2019-07-04
Typed arguments to MAIN interact poorly


kjpye:
## Problem

When MAIN has both typed named arguments and named array arguments the argument parsing can fail, producing a usage message rather than running the routine.

For example, the script

```
#!/usr/bin/env perl6

sub MAIN(:@a, Bool :$b) { say 'ran'; }
```

when run, produces the following...

```
# ./testcli
ran

# ./testcli --a=c
ran

# ./testcli --a=c --a=d
ran

# ./testcli --b
ran

# ./testcli --/b
ran

# ./testcli --a=c --b
Usage:
  ./testcli [-a=<Any> ...] [-b]

# ./testcli --a=c --/b
Usage:
  ./testcli [-a=<Any> ...] [-b]

# ./testcli --a=c --a=d --b
ran

# ./testcli --a=c --a=d --/b
ran

# ./testcli --a=c --a=d --a=e --b
ran

# ./testcli --a=c --a=d --a=e --/b
ran

# ./testcli --b --a=c
Usage:
  ./testcli [-a=<Any> ...] [-b]

#2917  ./testcli --b --a=c --a=d
ran
```

## Expected behaviour

It is expected that each case would produce "ran".

## Analysis

It appears that when there is exactly one named parameter corresponding to an array argument, and another named argument which is restricted in type are both specified, then a usage message is produced rather than running the code. 

This failure is also evident if you replace "Bool" with "Int".

A workaround is to remove the type restriction, in which case the code runs as expected.

## Environment

The above examples were run on OpenSuse Leap15 and Tumbleweed systems with the following Rakudo version:

```
This is Rakudo version 2019.03.1-667-gd4ceb97e0 built on MoarVM version 2019.05-95-g065b2b427
implementing Perl 6.d.
```

The behaviour is the same (with a differently worded usage message) on the following rakudo versions:

```
This is Rakudo version 2016.09 built on MoarVM version 2016.09
implementing Perl 6.c.

This is Rakudo version 2017.01 built on MoarVM version 2017.01
implementing Perl 6.c.

This is Rakudo version 2018.01 built on MoarVM version 2018.01
implementing Perl 6.c.

This is Rakudo version 2019.03.1 built on MoarVM version 2019.03
implementing Perl 6.d.
```

treyharris:
This appears to overlap with issues raised in #2794.

kjpye:
[...]

Issue #2794, opened 2019-03-24
`MAIN`: `Str` arguments do not accept `True` or `False`


treyharris:
## Expected Behavior

```
% for arg in hi 42 3.5 Nil True False; do
perl6 -e 'sub MAIN(Str :$flag) { say $flag.perl }; BEGIN @*ARGS="--flag='$arg'"'
done
"hi"
IntStr.new(42, "42")
RatStr.new(3.5, "3.5")
"Nil"
"True"
"False"
```

(Or, possibly the last two should be a currently-missing `BoolStr` allomorph or a compound type of some sort.)

## Actual Behavior

```
% for arg in hi 42 3.5 Nil True False; do
perl6 -e 'sub MAIN(Str :$flag) { say $flag.perl }; BEGIN @*ARGS="--flag='$arg'"'
done
"hi"
IntStr.new(42, "42")
RatStr.new(3.5, "3.5")
"Nil"
Usage:
  -e '...' [--flag=<Str>]
Usage:
  -e '...' [--flag=<Str>]
```

If `Str` is replaced with `Str()`, this works much as would have been expected:

```
% for arg in hi 42 3.5 Nil True False; do
perl6 -e 'sub MAIN(Str() :$flag) { say $flag.perl }; BEGIN @*ARGS="--flag='$arg'"'
done
"hi"
IntStr.new(42, "42")
RatStr.new(3.5, "3.5")
"Nil"
"True"
"False"
```

I think that, in the context of `MAIN`, a `Str` argument should always accept anything from a non-validating CLI like a Unix or Windows shell, so the `Str()` should, perhaps, be implicit.

## Steps to Reproduce

See above for examples that can be directly entered into the shell.

## Environment
* Operating system: Ubuntu 18.04.2 LTS
* Compiler version (`perl6 -v`): This is Rakudo Star version 2018.10 built on MoarVM version 2018.10
implementing Perl 6.c.

lizmat:
I assume the order of the arguments was intended to be: `% for arg in hi 42 3.5 Nil True False; do` ?

So that Nil => "Nil", True -> "True", etc...

lizmat:
Looks like the most obvious solution would be to introduce a `EnumStr` allomorph.  This would not only handle `True` and `False`, but any other enum such as `Planned`, `Kept`, `Broken`, `Less`, `More`, `Same`, etc., etc.

lizmat:
[...]

PR #4209, opened 2021-02-21
Implement space-delimited CLI arguments (S06; S19)

codesections:
This PR is a slightly revised version of #4201 with a cleaner diff; the companion spectests are at  Raku/roast/pull/716.  The previous commit message is below.

---

This commit resolves the exiting TODO item:
> Allow both = and space before argument of double-dash args

and thereby implements missing functionality described in S06 and S19.

Specifically, this commit adds an additional feature to the command-line programs Raku generates: users may now pass arguments to options separated by whitespace.  That is, if a program takes a `--port` option, users can now call it with `--port 4242` in addition to with `--port=4242`.  This behavior is specified in S06 and S19; it also makes the syntax for invoking Raku CLIs more similar to standard Linux utilities.

This feature was previously discussed in 2011, but was not implemented because no resolution could be found to two issues.  First, whitespace delimited arguments can be ambiguous: does `--profile foo` pass `foo` as an argument to `--profile` or as a positional argument to the program?  (One potential solution to this ambiguity is to inspect whether `foo` starts with `-`.  But this can create hard-to-debug issues for unusual `foo`s and is generally not reliable enough for use in scripts.) Second, is there a way to pass multiple arguments to a single option that allows users to use shell globing?  That is, can a call like `--input-files *.jpg` be made to DWIM?

This commit resolves these issues in a way that is inspired by getopt(1) and thus should be familiar to Linux users.  It also avoids making any changes that would be incompatible with existing Raku CLIs. Before discussing the details, here is an example.  Consider a `demo` program with the following source:

    subset name of Any where Str|True;
    subset port of Str;
    multi MAIN(
        $file,
        name :$profile,    #= Write profile information to a file
        port :$debug-port, #= Listen for debugger connections on the specified port
        Bool :v($verbose), #= Display verbose output

    ) {}
    multi MAIN("--process-files", *@images) {}

This program generates the following usage message:

    Usage:
      demo [--profile[=name]] [--debug-port=<port>] [-v] <file>
      demo --process-files [<images> ...]

        --profile[=name]       Write profile information to a file
        --debug-port=<port>    Listen for debugger connections on the specified port
        -v                     Display verbose output

The following are valid ways to call `demo`:

    demo --profile ~/foo
    demo --profile=/tmp/bar ~/foo
    demo --debug-port 4242 ~/foo
    demo --debug-port=4242 ~/foo
    demo -v ~/foo
    demo --process-files *.jpg

These, however, are not valid

    demo --profile /tmp/bar ~/foo
    demo --debug-port ~/foo

The first is invalid because `/tmp/bar` and `~/foo` are both parsed as positional arguments, which means `demo` was called with too many positional arguments.  The second is invalid because `~/foo` is parsed as an argument to `--debug-port`, and thus `demo` lacks the required positional argument.

Here's how it works:

Raku now distinguishes between three types of options:

 * Boolean options (like `-v`), which _never_ take an argument; they
   are ether present or absent.
 * Options with a mandatory argument (like `--debug-port`), which
   always take an argument.  If you give them an argument with `=`,
   they will use that; if not, they'll take the following argument.
 * Options with an optional argument (like `--profile`), which are
   valid both with and without an argument.  You can _only_ give these
   arguments an option with the `=` syntax; if there is a space after
   the option, that means it was called without an argument.

And here's the signature that produces each type of argument:

 * Boolean options: A `Bool` type constraint.
 * Options with a mandatory argument: A type that does not `.ACCEPT` a
   `Bool`
 * Options with an optional argument: A type that `.ACCEPTS` a
   `True` (because passing an option without an argument is equivalent
   to passing `True`)

This fully resolves the ambiguity discussed above in a way that should feel fairly intuitive to most users (despite how long
explaining it takes!) because it basically boils down to "space separated arguments are fine, unless you don't need to specify a
value; then you've got to use =".  This is the same solution employed by getopt(1) and should be familiar from standard utilities like `ls` and `cp`.

However, distinguishing between options with mandatory/required arguments introduces one potential trap into the language: Types from which `Bool`s inherit will produce an argument that takes an optional argument.  This group includes some numeric types, which may trip up some users.  Thus `sub MAIN(Int :$debug-port)` requires `--debug-port` to be called with `=` (or get a default value of 1), which may surprise some users.  After this feature is added, I will update the documentation to clearly note this issue.

The code in this commit takes an additional step to mitigate this footgun: it modifies the usage message to indicate which option arguments are optional by marking them with `[...]` (as was already done for optional positional arguments).

Finally, the desire to allow multiple arguments to be passed to a single option can be accommodated using the subcommand syntax, as shown in the `--process-files` example above.

Because argument parsing is so often used in interactive CLI programs, it is important that it be as fast as possible and I have written this commit perform as well as possible.  As a result I believe that, despite the increased functionality, the revised code runs ~1ms faster (out of ~6ms previously spent parsing arguments).  This decrease, however, is within the margin of error reported by my benchmarking tool and reflects only one machine.

I have prepared a companion commit that adds 10 tests to Roast (and removes 5 tests marked `todo`) that test this functionality.  Additionally, I have ensured that this test does not break bundling of short flags (a feature that is not specified/tested in Roast but that Raku supports) and have added tests to verify that compatibility.

### DIFF
--- a/src/core.c/Main.pm6
+++ b/src/core.c/Main.pm6
@@ -1,6 +1,5 @@
 # TODO:
 # * Command-line parsing
-#   * Allow both = and space before argument of double-dash args
 #   * Comma-separated list values
 #   * Allow exact Raku forms, quoted away from shell
 # * Fix remaining XXXX
@@ -44,7 +43,7 @@ my sub RUN-MAIN(&main, $mainline, :$in-as-argsfiles) {
     }
 
     # Convert raw command line args into positional and named args for MAIN
-    sub default-args-to-capture($, @args is copy --> Capture:D) {
+    sub default-args-to-capture(&main, @args is copy --> Capture:D) {
         my $no-named-after = nqp::isfalse(%sub-main-opts<named-anywhere>);
         my $bundling = nqp::istrue(%sub-main-opts<bundling>);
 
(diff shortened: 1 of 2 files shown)


codesections:
(It appears that the Rakudo-specific test I added has some issues on Windows.  I see that a similar test in the same file is skipped on Windows with the message `is-run() routine doesn't quite work right on Windows`.  I've adopted the same approach for the tests in this PR, but please feel free to suggest an alternate approach.)

Leont:
A long time ago Raku decided on doing doing blind argument parsing, instead of guided argument parsing like almost everyone else. This issue is but one example of where that fails flat on its face: #3029 and #2794 are other prominent examples. This category of issues was much of the motivation for me to write [Getopt::Long](https://github.com/Leont/getopt-long6/).

Long story short, I think that allowing space delimited CLI arguments is excellent. But instead of this solution, I would argue for a broader solution that solves all of these issues by doing a full analysis of all arguments before trying to parse anything (much like `Getopt::Long` does).

codesections:
Thanks for the comment @Leont, and for introducing me to the terms "blind argument parsing" and "guided argument parsing": I hadn't come across those terms before but a quick Internet search led me to your [helpful post](https://dev.to/leontimmermans/better-argument-parsing-with-getopt-long-2o13) on the subject.

You are absolutely correct, of course, that there's no way to satisfactorily parse whitespace-delimited options without looking at the options that `MAIN` accepts (what you call "guided parsing").  Indeed, the code in this PR [does just that](https://github.com/rakudo/rakudo/pull/4209/files#diff-c8624e29f9409b6c4fd348407adc812112ed9332f3c13c6ca3176f729b85e403R81-R90).  I also agree that it would probably be helpful to extend the built-in argument parser to take greater advantage of this sort of introspection (at least to the extent that we can do so without slowing down parsing speed or breaking existing code – something I'm optimistic about).

As you said, moving the parser in this direction will probably play a large role in eliminating a whole class of bugs and, potentially, in adding new features to Raku argument parsing.  This will likely make Raku's default argument parsing more similar to `getopt-long`'s in some ways, though I am sure that there will still be an important place for `getopt-long` as well, especially for users who want the greater flexibility of the OOP interface. 

I believe that we should take an incremental approach to getting Raku's argument parser to that point and that this PR is a good first step.  Once we've merged this PR, I'd be happy to work on a next step that continues to move in that direction and closes existing issues/TODOs.  I'm not entirely clear whether you agree or disagree with this approach.  When you said that you'd prefer a broader solution "instead of this solution", does that mean you'd prefer that we not merge this PR?  If so, I disagree – I think that merging a number of incremental PRs is a better/safer/more stable approach than attempting a single large re-write of the parser.

More practically, is there any reason _not_ to merge the PR as it stands?  It would both add a feature from the spec and move the parser in the direction we both think it should go in, after all.

Leont:
> More practically, is there any reason not to merge the PR as it stands?

Given that it's already written and works, not merging this now is not helpful. But for the other issues I would favor going back to the drawing board first instead of picking them off one by one. The latter feels like more work in the long run, and much more likely to give a messy end-result.

Leont:
As a side-note, this is subtly incompatible with how `Getopt::Long` handles this (it allows the space for optional arguments, but only if the argument doesn't start with `--`). I may add a compatibility option to `Getopt::Long` once this is accepted.

lizmat:
Thank you!

Leont:
> As a side-note, this is subtly incompatible with how `Getopt::Long` handles this (it allows the space for optional arguments, but only if the argument doesn't start with `--`). I may add a compatibility option to `Getopt::Long` once this is accepted.

Dit [exactly that](https://github.com/Leont/getopt-long6/commit/226e768567d2aa4f3f44ffc178986bc72bb1458b) in 0.3.3 :-)

codesections:
@Leont great!  I'm glad getopt-long gives people both options.  The main reason we didn't adopt that approach in this PR (even though it's pretty much what I started with my first attempt) was for backwards compatibility.  But the other set of factors were the arguments in [this blog post](https://nullprogram.com/blog/2020/08/01/).  As that post notes, though, Python's argparse goes with the "guess based on what the next argument looks like" option, so it's nice to have getopt long give users that choice.

patrickbkr:
@Leont Can you elaborate a bit on what hinders us from "just" replacing the current command line parser with Getopt::Long given, that Getopt::Long is a drop-in replacement?

Leont:
> Can you elaborate a bit on what hinders us from "just" replacing the current command line parser with Getopt::Long given, that Getopt::Long is a drop-in replacement?

Basically it's a 95% compatible replacement, and not 100%. In particular there are a few edge cases when dealing with multiple MAIN subroutines that have incompatible arguments; it's not an easy problem to solve.

PR #4218, opened 2021-02-24
Optimize param typecheck

vrurg:
Constraints are rather easy to typecheck against. It is sufficient to typecheck against coercion target and constraint types. They can, in turn, be other nominalizables with complex typechecking. But at least one level of invocation can be eliminated. The optimization makes loop in the following example to run 8x to 9x faster:

```raku
        sub foo(Bool() $v) { !$v }
        for ^10_000_000 { my $r = foo(False); }
```

### DIFF
--- a/src/Perl6/Actions.nqp
+++ b/src/Perl6/Actions.nqp
@@ -9206,9 +9206,10 @@ class Perl6::Actions is HLL::Actions does STDActions {
 
             # Add type checks.
             my $param_type := %info<type>;
+            my $ptype_archetypes := $param_type.HOW.archetypes;
             my int $is_generic := %info<type_generic>;
-            my int $is_coercive := $param_type.HOW.archetypes.coercive;
-            my $nomtype    := !$is_generic && $param_type.HOW.archetypes.nominalizable
+            my int $is_coercive := $ptype_archetypes.coercive;
+            my $nomtype    := !$is_generic && $ptype_archetypes.nominalizable
                                 ?? $param_type.HOW.nominalize($param_type)
                                 !! $param_type;
             my int $is_rw := $flags +& $SIG_ELEM_IS_RW;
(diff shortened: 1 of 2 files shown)


vrurg:
A few small stats. Tested with this script:

```raku
sub foo(Bool() $v) { !$v }
sub bar(Bool $v)   { !$v }

my $st = now;
for ^10_000_000 {
    my $r = foo(False);
    #my $r = foo("yes"); # To typecheck against constraint type of coercion
}
my $et = now;
my $coercive = $et - $st;

$st = now;
for ^10_000_000 {
    my $r = bar(False);
}
$et = now;

my $nominal = $et-$st;

say "Time: coercive: ", $coercive, "; nominal: ", $nominal;
say "Netto: ", $coercive - $nominal;
```

The script revealed a really funny case where typechecking against coercion is event faster for plain `Bool` value, i.e. for this exact variant:

```
Time: coercive: 0.3807399; nominal: 1.43136106
Netto: -1.0506211921373
```

For the same version rakudo 2021.02 results in:

```
Time: coercive: 3.6128886; nominal: 1.4128102
Netto: 2.2000784438914
```

When coercion is needed (the commented out `foo` call with a string argument) numbers are less impressive, but we must remember that they include time needed to invoke a coercion method.

This PR:

```
Time: coercive: 9.7010204; nominal: 1.432363
Netto: 8.2686573083
```

2021.02:

```
Time: coercive: 13.77604515; nominal: 1.450659328
Netto: 12.3253858234812
```

I also tested the one-liner provided by @MasterDuke17 on IRC:

```raku
my %h = a => 42, b => 666; my $b; my $st=now; $b := %h<a>:exists for ^10_000_000; my $et = now; say $et-$st
```

The loop takes ~2.4-2.5 less time for this PR.

MasterDuke17:
> The script revealed a really funny case where typechecking against coercion is event faster for plain `Bool` value, i.e. for this exact variant:

You've run into the same problem guifa did on #raku a couple days ago. Testing multiple variants of code in the same file can cause difficulty for the dynamic optimizer (i.e., it optimizes for the earlier code, which might hurt the later code). If I just use a single loop and then edit what's done in the body and re-run I get timings more like you'd expect:

2021.02-16-g7d6e43224:
foo(False) - 2.5s
foo("yes") - 10.3s
bar(False) - 0.155s

optimize-param-typecheck:
foo(False) - 0.155s
foo("yes") - 6.8s
bar(False) - 0.155s

PR #4221, opened 2021-02-24
Release 2021.02.1

Altai-man:


docs/ChangeLog
```diff
@@ -1,3 +1,13 @@
+New in 2021.02.1:
+
+  + Fixes:
+    + Fix a regression of the 2021.02 release where a lazy iterable usage could suddenly
```
MasterDuke17, line 4: Wasn’t it a lazy iterable *slice*?

lizmat: Not even that, it only happened when a lazy *sublist* occurred in a slice.  So, `@a[^Inf]` was ok, `@a[1,^Inf]` would hang.  And `Text::CSV` had a test using that feature, which is how we noticed the issue.

Altai-man: The fix commit had a "a lazy iterable as part of a non lazy list of indexes in a slice" and I had to generalize/simplify it. Maybe "where a specific usage" would be better in this case, but this is a matter of subjective wording to me.
What matters, I think, is that there was a regression related to lazy iterables, it broke things, we fixed it, celebrate.

PR #4226, opened 2021-02-26
$x ** ±Inf is sometimes 0

kurahaupo:
inside_unit_circle ** Inf == 0
outside_unit_circle ** -Inf == 0


lizmat:
Thanks, and sorry that it took so long!

PR #4228, opened 2021-02-27
Add a predicated wait for Lock::ConditionVariable

Leont:
## The Problem
Currently, the `wait` method of `Lock::ConditionVariable` follows the paradigm of the C functions. I would suggest adding a more rakuish variant that takes a predicate. Much like the `protect` method in`Lock`, this makes it much easier to do the correct thing (and people will do things incorrectly if the easiest way is not also the correct way). So I propose adding this:

```
multi method wait(&predicate) {
    self.wait until predicate;
}
```

To take an example from `Awaiter`, instead of writing:
```
loop {
    $l.protect: {
        last if $remaining == 0;
        $ready.wait;
    }
}
```
One could now write
```
$l.protect: {
    $ready.wait: { $remaining == 0 }
}
```

### DIFF
--- a/src/core.c/Lock.pm6
+++ b/src/core.c/Lock.pm6
@@ -9,7 +9,9 @@ my class Lock {
         method new() {
             X::Lock::ConditionVariable::New.new.throw
         }
-        method wait() { nqp::condwait(self) }
+        proto method wait(|) {*}
+        multi method wait(--> Nil) { nqp::condwait(self) }
+        multi method wait(&predicate --> Nil) { nqp::condwait(self) until predicate; }
         method signal() { nqp::condsignalone(self) }
         method signal_all() { nqp::condsignalall(self) }
     }
--- a/src/core.c/Awaiter.pm6
+++ b/src/core.c/Awaiter.pm6
@@ -92,13 +92,8 @@ my class Awaiter::Blocking does Awaiter {
                 });
             }
 
-            # Block until remaining is 0 (need the loop to cope with suprious
-            # wakeups).
-            loop {
-                $l.protect: {
-                    last if $remaining == 0;
-                    $ready.wait;
-                }
+            $l.protect: {
+                $ready.wait: { $remaining == 0 }
             }
 
             # If we got an exception, throw it.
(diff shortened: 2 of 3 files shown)


src/core.c/Lock.pm6
```diff
@@ -9,7 +9,9 @@ my class Lock {
         method new() {
             X::Lock::ConditionVariable::New.new.throw
         }
-        method wait() { nqp::condwait(self) }
+        proto method wait(|) {*}
+        multi method wait() { nqp::condwait(self) }
+        multi method wait(&predicate) { nqp::condwait(self) while !predicate; }
```
vrurg, line 14: Why not `until predicate`?

Leont: Because I only thought of that after I wrote the code?

vrurg:
One thing I'd like to ask you about is to be somewhat more elaborate in commit messages. It's so helpful for code archeologists. :)

Will merge when get back home, unless somebody does it by then.

Leont:
> One thing I'd like to ask you about is to be somewhat more elaborate in commit messages. It's so helpful for code archeologists. :)

Fair point!

src/core.c/Lock.pm6
```diff
@@ -9,7 +9,9 @@ my class Lock {
         method new() {
             X::Lock::ConditionVariable::New.new.throw
         }
-        method wait() { nqp::condwait(self) }
+        proto method wait(|) {*}
+        multi method wait() { nqp::condwait(self) }
+        multi method wait(&predicate) { nqp::condwait(self) until predicate; }
```
jnthn, line 14: Hm, I doubt these return anything useful or that we've spec'd that, but I'd suggest while we're at this, we throw a `--> Nil` on the two candidates.

jnthn (commented):
This makes sense to me; it's very much in the spirit of providing factored out forms of common use cases (aside from the `protect` method on locks, the same kind of thing is done by proving a block form of `cas` too).

src/core.c/Lock.pm6
Leont: Done

jnthn (approved):

Issue #4191, opened 2021-02-02
[JVM] Make 'make test' usable again


usev6:
## The Problem
Currently ```make test``` isn't usable on the JVM backend. The primary problem is that ```t/harness5``` uses the EvalServer -- and the EvalServer heavily leaks threads and memory: https://github.com/Raku/old-issue-tracker/issues/6527.
I'm pretty sure that there are other problems, since very likely no-one has tried to run Rakudo's tests on the JVM backend for a long time.

Some ideas how to make ```make test``` usable again:
* Don't use the EvalServer, but run the test files one by one (in the same way as it is done on moar). This could be done by adding an option ```--evalserver``` to ```t/harness5``` and avoid the EvalServer unless that option is used.
* It should be possible to run ```make test```before ```make install```. There is an old issue about that: https://github.com/Raku/old-issue-tracker/issues/6524.
* Make sure the correct test files are run (requires a change to the generated Makefile). 
* Look at the fallout and fudge failing tests. (Very likely there is still a major problem with nativecall tests: https://github.com/Raku/old-issue-tracker/issues/6525.)

patrickbkr:
Can you give a guesstimate of how much runtime of `make test` would drop when dumping the EvalServer?

usev6:
Well, I'd expect the tests to run for more than one hour, but hopefully less than two hours (that is with ```TESTS_JOBS_1```). There seems to be one hanging test at the moment -- that would need fixing first.

Comparing with the status quo isn't really possible, because the EvalServer stops to work (due to the memory/thread leaks) after a couple of test files.

usev6:
Another failing test should be fixed with https://github.com/rakudo/rakudo/pull/4268

usev6:
A couple of tests fail because the JVM backend emits a slightly different error message when throwing X::Bind or X::Delete for native arrays. Since there are other inconsistencies for those error messages (and some tests) I've opened a separate issue: https://github.com/rakudo/rakudo/issues/4269

usev6:
All tests for NativeCall are failing. See
* https://github.com/rakudo/rakudo/issues/3043
* https://github.com/Raku/old-issue-tracker/issues/6525

usev6:
All test files for Telemetry are exploding. See https://github.com/rakudo/rakudo/issues/4354

usev6:
Just to add an update here: Most of the tests for Telemetry are passing now (https://github.com/rakudo/rakudo/issues/4354 has been resolved).

[...]

PR #4230, opened 2021-02-27
Tweak some tests to work on JVM backend

usev6:
Part of making 'make test' usable again on the JVM backend (https://github.com/rakudo/rakudo/issues/4191).

### DIFF
--- a/t/02-rakudo/03-corekeys-6c.t
+++ b/t/02-rakudo/03-corekeys-6c.t
@@ -761,6 +761,38 @@ my @expected = (
   Q{𝑒},
 );
 
-has-symbols CORE::, @expected, "Symbols in 6.c CORE::";
+my %nyi-for-backend = (
+  'jvm' => (
+    Q{&atomic-add-fetch},
+    Q{&atomic-dec-fetch},
+    Q{&atomic-fetch-add},
+    Q{&atomic-fetch-dec},
+    Q{&atomic-fetch-inc},
+    Q{&atomic-fetch-sub},
+    Q{&atomic-inc-fetch},
+    Q{&atomic-sub-fetch},
+    Q{&full-barrier},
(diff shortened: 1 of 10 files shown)


vrurg:
I'd merge it already but have a feeling that this way it provokes to postpone implementation of the missing features. Perhaps fudging would be a better option?

Sometimes I miss a "test warning" feature. Not just a `diag` comment, but something to be reported in summaries.

usev6:
I see your point. Actually you made a similar point in https://github.com/Raku/roast/pull/652#issuecomment-659408253 last year, so I wondered if there is an easy way to fudge the failing tests for the JVM backend. But I didn't see an obvious solution and in the end I tried the easy way again.

The following two points make me believe that it's "good enough" to just exclude the missing features:
* For all of those symbols there are actually tests in roast. They are fudged or otherwise disabled for the JVM backend, but still. Wouldn't roast as the language specification be a more appropriate place to "press" for the implementation of missing features?
* The JVM backend is lagging behind MoarVM since many years. As I see it the visibility of missing features is not the main problem, it's just that no-one has the tuits|energy|knowledge|time|interest to implement them. So IMHO we wouldn't lose much if we hide the missing features here.

But I acknowledge that these are not very strong arguments.

vrurg:
Sorry for late reply, but it's only today I found github notification in a pile of old unread emails...

Anyway, your reasoning about having roast to test for features and rakudo suit only testing what's actually implemented, appeals to me. Aside of this, I just realized that it _will_ complain when it finds a new symbol because the way you've done it doesn't ignore missing ones but simply doesn't expect them to appear in the list. Any newly implemented one will cause a test failure unless removed from the ignore list.

I'm too exhausted today to do thorough re-review of this PR before merging. But hope to get to it tomorrow.

Thank you!

PR #4231, opened 2021-03-01
Add support for userspace encodings in Blob.decode

codesections:
This adds support for decoding Blobs with userspace encodings registered in Encoding::Registry (this parallels the exiting support for userspace encodings in Str.encode).  I initially thought that we could do this by using the Encoding::Registry decoder in all cases (both builtin and user-defined) since it the registry should include all builtin decodings.

However, due to what appears to be a bug in MoarVM, the version of the builtin decoders in the registry do not correctly normalize some encodings into NFG.  I have opened an issue regarding that bug, MoarVM/MoarVM#1443.

In the meantime, I wrote this code to use the builtin decoders when decoding a supported encoding and to try the registry only if passed an unknown encoding.

Because this change involves decoding, I ran the full stresstest.  I have also prepared accompanying Roast tests for inclusion in S32-encoding/registry.t.


src/core.c/Buf.pm6
```diff
@@ -291,23 +291,34 @@ my role Blob[::T = uint8] does Positional[T] does Stringy is repr('VMArray') is
...
-            Rakudo::Internals.NORMALIZE_ENCODING($encoding),
-            $replacement.defined ?? $replacement !! nqp::null_s(),
-            $strict ?? 0 !! 1))
+        if Rakudo::Internals.BUILTIN_ENCODINGS{$encoding} {
+            nqp::p6box_s( nqp::decoderepconf(self,
+                            Rakudo::Internals.NORMALIZE_ENCODING($encoding),
+                            $replacement.defined ?? $replacement !! nqp::null_s(),
+                            $strict ?? 0 !! 1))
+       }
+       else { my $decoder = CORE::Encoding::Registry.find($encoding).decoder(:$replacement, :$strict);
```
lizmat, line 309: Does it always find a decoder?

src/core.c/Buf.pm6
```diff
@@ -291,23 +291,34 @@ my role Blob[::T = uint8] does Positional[T] does Stringy is repr('VMArray') is
 
     proto method decode(|) {*}
     multi method decode(Blob:D: $encoding = self.encoding // "utf-8") {
-        nqp::p6box_s(
-          nqp::decode(self, Rakudo::Internals.NORMALIZE_ENCODING($encoding))
-        )
+        if Rakudo::Internals.BUILTIN_ENCODINGS{$encoding} {
```
lizmat, line 294: I fail to see the use of `Rakudo::Internals.BUILTIN_ENCODINGS`.  I think the condition could be better written as:

```` raku
if Rakudo::Internals.NORMALIZE_ENCODING($encoding) -> $normalized {
    nqp::p6box_s(nqp::decode(self,$normalized))
}

and the same at similar places in this PR?

**Review by lizmat — CHANGES_REQUESTED** [review]

File: src/core.c/Buf.pm6 (hunk 1 above)
Comment (codesections) [reply src/core.c/Buf.pm6:309]: I'm not sure I understand the question.  If the user specifies an encoder that is builtin or that they've added to the registry, then it always finds a decoder.  If they specify a decoder that is neither builtin nor in the registry, then it throws an error `Unknown string encoding: '$encoding'`.  This is the same behavior as before this PR.

(Actually, it's slightly different because the exception is now appropriately typed: it's an `X::Encoding::Unknown.new(name => "non-existant-encoding")`  instead of an `X::AdHoc.new(payload => "Unknown string encoding: 'non-existant-encoding'")` exception.  But it's the message it displays is the same.)

File: src/core.c/Buf.pm6 (hunk 2 above)
Comment (codesections) [reply src/core.c/Buf.pm6:294]: That was actually my first thought as well!  However, that doesn't work because `NORMALIZE_ENCODING` falls back to returning a lower-cased version of the encoding when passed an unknown encoding:

```raku
say Rakudo::Internals.NORMALIZE_ENCODING('Foo')  # OUTPUT: «foo»
```

(I suppose the thought here is that we're _normalizing_ encodings, not checking whether they're known.)  This means that the if block you suggested would always return `True`.  Of course, we could modify the behavior of `NORMALIZE_ENCODINGS`, but I thought it would be better to suggest the smaller change.

lizmat: https://github.com/rakudo/rakudo/commit/22ea9d903f makes R:I:NORMALIZE_ENCODING return `Nil` rather than a lowercase version of the input.

codesections: Fixed in 68dbb99, thanks.

lizmat:
I'd like to hold off merging this until we're sure there's no Blin fallout from https://github.com/rakudo/rakudo/commit/22ea9d903f015559e8b5e15f51c7c1d9438da42b

codesections:
> I think we need to look at any location in the code that uses Rakudo::Internals.NORMALIZE_ENCODING.

Hmm?  Grepping over the codebase for "NORMALIZE_ENCODING" only shows it being used in `Buf.pm6` and `IO/Path.pm6`.

What other areas could use support for userspace encodings and would it make sense to address some/all of them in separate PRs?

codesections:
I'm with you now.  I'll update the PR to address `Path.slurp` and `Path.spurt-string` (the two methods other than `Blob.decode` where `NORMALIZE_ENCODING` is used.  I'll also update all three to pass the invalid decoding on to be used in the error message.

lizmat:
I wonder whether we shouldn't have some generic "give me a decoder / encoder for this encoding" interface in Rakudo::Internals, which would also check the registry.  Or alternately, maybe we should just *always* check the registry?

codesections:
> Or alternately, maybe we should just always check the registry?

I think that we should, in the end.

But right now that still runs into the MoarVM bug I mentioned when I submitted this PR (MoarVM/MoarVM#1443).  Since I don't know MoarVM (or, if I'm honest, C) well enough to hunt that bug down, I think it makes sense to get a working implementation here.  We can always revisit it when the MoarVM issue is resolved.

codesections:
I revised the approach in the PR to ensure that we pass the name of any unknown encoding on, and can therefore throw a correct error.  I also added tests for that error.

While doing so, I realized that some functions that were trying to set a default encoding (when the user didn't specify one) were failing to do so, but that bug was hidden by the fact that `NORMALIZE_ENCODING` _also_ set the same default.  I have fixed that issue and set the default only once, at the public API boundary.

codesections:
Any objections to me merging the revised version of this PR?  I'm hoping to get it merged in time for the next release.

src/core.c/io_operators.pm6
```diff
@@ -125,10 +125,9 @@ multi sub close(Channel:D $channel) { $channel.close }
 
 proto sub slurp(|) {*}
 multi sub slurp(*%_) { $*ARGFILES.slurp(|%_) }
-multi sub slurp(IO::Handle:D $fh, *%_) { $fh.slurp(|%_) }
-multi sub slurp(IO() $path, :$bin!) { $path.slurp(:$bin) }
-multi sub slurp(IO() $path, :$enc ) { $path.slurp(:$enc) }
-multi sub slurp(IO() $path        ) { $path.slurp(:enc<utf8>) }
+multi sub slurp(IO::Handle:D $fh, *%_)    { $fh.slurp(|%_) }
+multi sub slurp(IO() $path, :$bin!)       { $path.slurp(:$bin) }
+multi sub slurp(IO() $path, :$enc='utf8') { $path.slurp(:$enc) }
```
lizmat, line 130: Why the reduction in candidates here?  The extra candidate exists to make dispatch faster.

codesections: The only multi candidate that I cut is `multi sub slurp(IO() $path) { $path.slurp(:enc<utf8>) }`  (the diff shows three others as modified because I adjusted alignment).

I cut that one because, unless I'm confused, it's never called: `slurp('file')` dispatches to `slurp(IO() $path, $enc)` because it matches and is narrower.

lizmat:
@codesections is this still salvageable?  If not, please close  :-)

Issue #3293, opened 2019-11-09
utf32 encode documented but not implemented


thundergnat:
## The Problem
Either a documentation problem or an implementation problem.

UTF32 encode / decode is not implemented. 

This page: https://docs.perl6.org/type/Encoding::Registry specifically implies that it is.

```
# also check that it isn't just a case / spelling problem
for <utf8 utf16 utf32 UTF32 utf-32 UTF-32> {
    say '🇺🇸'.encode($_);
    CATCH { default { .say } }
}
```
```
utf8:0x<F0 9F 87 BA F0 9F 87 B8>
utf16:0x<D83C DDFA D83C DDF8>
Unknown string encoding: 'utf32'
  in block <unit> at utf32.p6 line 2

Unknown string encoding: 'utf32'
  in block <unit> at utf32.p6 line 2

Unknown string encoding: 'utf32'
  in block <unit> at utf32.p6 line 2

Unknown string encoding: 'utf32'
  in block <unit> at utf32.p6 line 2
```

## Environment
* Operating system:  

Linux Mint 19.2

* Compiler version (`perl6 -v`): 

This is Rakudo version 2019.07.1 built on MoarVM version 2019.07.1
implementing Perl 6.d.

codesections:
@thundergnat thanks for pointing this issue out to me.

I believe that this issue is caused by https://github.com/MoarVM/MoarVM/issues/1348, which I opened earlier today.  If I am reading the code correctly, Rakudo and NQP both expect VMs to implement an operation that handles utf32 encoding, but MoarVM doesn't.  As I described in that issue:

> [MoarVM does not have] a `string/utf32.c` file, nor is there a `MVM_string_utf32_encode_substr` function, which I would expect to be defined in that file.  If MoarVM did implement a `MVM_string_utf32_encode_substr` function, then it could be added as a `case` in the  [`MVM_string_encode_config`](https://github.com/MoarVM/MoarVM/blob/master/src/strings/ops.c#L1723-L1725) function which would, I believe, add support for encoding utf32.

lizmat:
Hmmmm  How is utf32 different from the Uni representation?

MasterDuke17:
> Hmmmm How is utf32 different from the Uni representation?

I think they're pretty similar. One int (i.e., 32-bits) per codepoint.  But I don't know how easy it would be for Rakudo to use `Uni` when `utf32` is used as an encoding?

lizmat:
[...]

PR #4232, opened 2021-03-01
Remove utf32 from encodings lists for MoarVM

codesections:
Rakudo currently lists utf32 as one of the encoding/decoding operations it supports.  However, this support has never been added to the MoarVM backend and attempting to use it throws a somewhat confusing exception.

Accordingly, this commit removes utf32 from relevant lists in Rakudo::Internals and Encoding::Registry when Rakudo is compiled for MoarVM (the JVM backend supports utf32).  This causes spectest failures because Roast was checking whether an encoding was listed in the Encoding::Registry without testing whether the encoding could actually be used.

I have prepared a companion PR that marks these tests as todo for rakudo.moar and that adds substests to ensure that registered encodings can actually be used.

Related issues: #3293 and  MoarVM/MoarVM#1348


lizmat:
Thanks!

PR #4235, opened 2021-03-04
Make .keys/values/kv/pairs/antipairs/invert work on Enums

lizmat:
It appears to be a recurring WAT when people try to find out which
possibilities there are when using a specific enum.  Before this
commit, "say Bool.keys" would show "()", after this commit it will
show "(True False)" (or "(False True)" depending on hash order).

This is achieved by adding a check for enums to the Any:U candidates
of these methods, and then calling the appropriate method on the
.enums method.

cog++ for the nudge

### DIFF
--- a/src/core.c/Any.pm6
+++ b/src/core.c/Any.pm6
@@ -92,39 +92,39 @@ my class Any { # declared in BOOTSTRAP
     multi method end(Any:D:) { self.list.end }
 
     proto method keys(|) is nodal {*}
-    multi method keys(Any:U:) {
-        nqp::istype(self.HOW,Metamodel::EnumHOW) ?? self.enums.keys !! ()
-    }
+    multi method keys(Enumeration:) { self.enums.keys }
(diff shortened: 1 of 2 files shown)


vrurg:
This implementation feels a little wrong to me. It is likely to kick back with optimizations. I'd rather think of a multi-candidate for `Enumeration`; and adding dedicated methods on `Bool`. Since it's a special case, so let it be a special case which would be easier for VM to optimize.

While technically this approach is concise and allows one to create a custom `enum` without consuming `Enumeration`, I'd rather require such custom typeobjects to implement the methods. This is already done implicitly by this PR because it expects `enums` method to be present and `Bool` does implement it explicitly.

lizmat:
```` raku
dd Bool ~~ Enumeration;  # False
````

I guess we would need to make a separate candidate for `Bool` then.

vrurg (approved):

PR #4239, opened 2021-03-06
Optimize truth checks of arrays in whiles and ifs

MasterDuke17:
E.g., `while @a { ... }` and `if @a { ... }` are now converted into
`while +@a { ... }` and `if +@a { ... }`.

Rakudo builds ok and passes `make m-test m-spectest`, but the broken `make install` is due to the `if` optimization and I haven't figured out why/what the fix is yet.

Before `my @a = ^1_000_000; my $b; my $s := now; $b := @a.shift while @a; say now - $s; say $b` took ~0.9s, after it takes ~0.15s.

### DIFF
--- a/src/Perl6/Optimizer.nqp
+++ b/src/Perl6/Optimizer.nqp
@@ -1354,6 +1354,28 @@ class Perl6::Optimizer {
             }
         }
 
+        # Optimize `while @a { ... }` and `if @a { ... }` into
+        # `while +@a { ... }` and `if +@a { ... }`
+        if $optype eq 'while' {
+            my $conditions := $op[0];
+            if nqp::istype($conditions, QAST::Var) && nqp::substr($conditions.name, 0, 1) eq '@' {
+                $op[0] := QAST::Op.new( :op('callstatic'), :name('&prefix:<+>'), $conditions );
+            }
+            self.visit_op_children($op);
+            return $op;
+        }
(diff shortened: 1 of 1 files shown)


MasterDuke17:
It dies at https://github.com/rakudo/rakudo/blob/master/src/core.c/Parameter.pm6#L331

MasterDuke17:
Changing https://github.com/rakudo/rakudo/blame/master/src/core.c/Parameter.pm6#L332 to just `(my int $elems = nqp::elems(@!named_names))` does build and `make install` ok, but causes a bunch of fails in `make m-test`. E.g.,
```
    ok 1 - 'multi sub f(int $foo is rw) { }; f(42)' died
    not ok 2 - right exception type (X::Comp)
    # Failed test 'right exception type (X::Comp)'
    # at SETTING::src/core.c/Exception.pm6 line 62
    # Expected: X::Comp
    # Got:      X::TypeCheck::Binding::Parameter
    # Exception message: Type check failed in binding to parameter '<anon>'; expected Any but got Mu (Mu)
    # You failed 1 test of 2
not ok 2 - calling multi sub that expects a rw native argument with a literal is caught at compile time
# Failed test 'calling multi sub that expects a rw native argument with a literal is caught at compile time'
# at t/02-rakudo/13-exceptions.t line 13
```

MasterDuke17:
Closing this because now after new-disp has landed it actually slows the example down, from ~0.12s to ~0.13s (and still has the same not-understood failures).

PR #4240, opened 2021-03-06
Add --> Order:D for Num's cmp and don't create Bools

codesections:
This is another commit following up on b2ae599a75 by adding a return constraint to `&infix:<cmp>` operators.  This commit covers `Num`, and also avoids using `&infix:<or>`, which was needlessly constructing a `Bool`.


lizmat:
Good catch.  But the `--> Order:D` is not necessary, as it is only inspected on the `proto`, and that has it.

codesections:
... of course.  I knew that.  Thanks, fixed.

lizmat:
Thanks!

PR #4244, opened 2021-03-09
Add new routine 'frac'

tbrowder:
This PR adds a **frac** routine to Rakudo. The routine takes a real or integer argument and returns the fractional part as a positive number (0 <= n < 1). It will throw on Inf or NaN arguments.

This PR is accompanied by a PR to roast adding a new test file for **frac**.

Although not commonly found in modern programming languages, such a routine is used frequently in astronomical calculations and is described in the book *Celestial Calculations: A Gentle Guide  to Computational Astronomy* (J. L. Lawrence, The MIT Press, 2019) on page 7.

Adding this routine I believe would add a positive note to Raku's reputation.

### DIFF
--- a/src/core.c/Real.pm6
+++ b/src/core.c/Real.pm6
@@ -60,7 +60,7 @@ my role Real does Numeric {
     multi method exp(Real:D: )           { self.Bridge.exp               }
     
     method frac(Real:D:) {
-        self.abs - self.Int.abs
+        abs(self - self.Int)
     }
     method truncate(Real:D:) {
         self < 0  ?? self.ceiling !! self.floor
--- a/src/core.c/Real.pm6
+++ b/src/core.c/Real.pm6
@@ -58,6 +58,10 @@ my role Real does Numeric {
     multi method log(Real:D: Real $base) { self.Bridge.log($base.Bridge) }
     proto method exp(|) {*}
     multi method exp(Real:D: )           { self.Bridge.exp               }
+    
+    method frac(Real:D:) {
+        self.abs - self.Int.abs
+    }
     method truncate(Real:D:) {
         self < 0  ?? self.ceiling !! self.floor
     }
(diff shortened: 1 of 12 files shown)


tbrowder:
On Tue, Mar 9, 2021 at 08:19 Daniel Sockwell ***@***.***> wrote:
> I like this idea in principle
...
> I might bikeshed on the name, though. frac is a pretty short word,

An alternative is to have one be an alias for the other (fraction vs
frac). Precedent is established with 'log' and siblings as well as
'truncate' and 'Int'.

The flip side is other languages have frac (or frac-like: Frac or modf)

+ Pascal - Frac
+ R - Frac
+ C - modf
+ Irie Pascal - frac

jonathanstowe:
> The flip side is other languages have frac (or frac-like: Frac or modf)

Interestingly (or not,) it appears to have appeared in FORTRAN-95 as `FRACTION`, but earlier 3rd party FORTRAN library implementations seem to have `frac` (I was curious about this because the language would tend to speak to long time usage in the domain you originally mentioned.)

tbrowder:
On Tue, Mar 9, 2021 at 11:43 Jonathan Stowe ***@***.***>
wrote:

> Interestingly (or not,) it appears to have appeared in FORTRAN-95
>
> I didn't think about that, I should look in my old FORTRAN IV book from
1961 if I can find it!

lizmat:
FWIW, I seem to remember seeing it when I was doing FORTRAN in the 1976 / 1977 timeframe.  On punchcards, no less!

jonathanstowe:
I think we should be quiet about this now, people might get the wrong idea about the language 🤣🤣🤣🤣

src/core.c/Real.pm6
```diff
@@ -58,6 +58,10 @@ my role Real does Numeric {
     multi method log(Real:D: Real $base) { self.Bridge.log($base.Bridge) }
     proto method exp(|) {*}
     multi method exp(Real:D: )           { self.Bridge.exp               }
+    
+    method frac(Real:D:) {
+        self.abs - self.Int.abs
```
vrurg, line 63: Wouldn't it be a bit faster if we do `abs(self - self.Int)`? One less method call anyway.

tbrowder: You are right. By my count 4 ops with my code, 3 ops with yours.

vrurg (approved):

vrurg:
@lizmat would you mind if we merge?

lizmat (approved):

Issue #4234, opened 2021-03-04
Minor issues when building NQP then Rakudo on AIX7: -bmaxdata:0x80000000 and usage of "ln"


ItchyPlant:
## The Problem

There are still two minor issues when building Raku on AIX7 which can be solved easily. 

### This is shown for both **NQP** and **Rakudo** when issuing `make`
```
gcc: error: ,-bmaxdata:0x80000000: No such file or directory
```

### This is only for **Rakudo** during `make install`
```
...
+++ Creating Raku executable alias
Usage: ln [-f|n] [-s] src [target]
   or: ln [-f|n] [-s] src1 ... srcN directory
make: *** [Makefile:1509: m-runner-default-install] Error 1
```

## Expected Behavior
No issues

## Actual Behavior
Some tunes need to be applied. 

## Steps to Reproduce

Before doing anything:  
```
# export LIBPATH=/opt/raku/lib
```

### 1. MoarVM

Used **MoarVM-2021.02**. 

Preparation in `build/setup.pm` (but it is OK and it compiles without it anyway):  
In `OS_AIX` section, replaced `ldmiscflags         => '-Wl,-brtl',` to `ldmiscflags         => '-Wl,-bmaxdata:0x80000000,-brtl',`

The complete block after chaning:  
```
...
our %OS_AIX = (
    %OS_POSIX,

    defs                => [ qw( _ALL_SOURCE _XOPEN_SOURCE=500 _LINUX_SOURCE_COMPAT ) ],
    syslibs             => [ @{$OS_POSIX{syslibs}}, qw( rt dl perfstat ) ],
    ldmiscflags         => '-Wl,-bmaxdata:0x80000000,-brtl',
    ldrpath             => '-L"/@libdir@"',
    ldrpath_relocatable => '-L"/@libdir@"',
...
```

#### Build commands used:
```
# perl Configure.pl --has-libffi --prefix /opt/raku
...
# make
...
# make install
...
```
Result: 
```
# /opt/raku/bin/moar --version
This is MoarVM version 2021.02
```

### 2. NQP

Used **nqp-2021.02**.

Configure command used: 
```
# perl Configure.pl --prefix /opt/raku
...
```
Then in `Makefile`, this had to be fixed: 
```
...
M_LD_INST_NQP=gcc -Wl,-bmaxdata:0x80000000,-brtl -O3 -DNDEBUG ,-bmaxdata:0x80000000 -L"//opt/raku/lib"
...
[...]

PR #4246, opened 2021-03-10
Fix `ln -nfs` on AIX

vrurg:
Add a substitute script for `ln -nfs`

Currently it is for AIX only where there is no flag with the same meaning as `-n`/`-h` in GNU and BSD implementations.


vrurg:
Should fix #4234

PR #4248, opened 2021-03-12
Introduce CX::Returning and use it to expose the payload of most control exceptions.

gfldex:
Control exceptions are exposed to the user via CONTROL. Yet the payload that some CX can carry is not available. That limits the usability of control exceptions quite badly. By introducing the role `CX::Returning` user defined CX can get the same treatment.

Tests can be found at https://gist.github.com/c5f5b4d0192a1abd17b48b9554be8004 .

More work needs to be done to make CX more useful. This PR is the first step.


src/core.c/Exception.pm6
```diff
@@ -351,31 +351,36 @@ my class X::Pragma::OnlyOne is Exception {
 
 my role X::Control is Exception {
 }
-my class CX::Next does X::Control {
+my role CX::Returning {
```
vrurg, line 354: Can we find a better name for the role? To my view neither `take` nor `emit` are not a return-kind things.

gfldex: Sure. What is your suggestion?

gfldex: Thinking about it a little longer I would like to point the following out. If I borrow your lawn mower and give it to your wife after using it, did I return the lawn mower? Do I have to put it back into your shed or will the garage do? Let's have a look of the common usage we have for the english term "return" in Raku.

```
sub borrow($thing) {
    return "short sharp knife";
}

say borrow("lawn mower");
```

I strongly doubt many functions are actually "returning" anything. Some functions even have a completely empty signature and yet, they "return" a value that was never given to them. I'm quite sure we can come up with a differently wrong english phrase. Not sure if that would be better.

taboege: To my mind, what `return` returns (and the returner utilized on behalf of the returnee) is *control over the process*. Maybe structured programming is so natural today that this meaning doesn't stand out anymore.  Raku has a number of different words for this same concept, which are used depending on who transfers control (a `sub` call, a `supply` block, a `when` block, a `gather` block, ...). In fact, this meaning of "returning" could be said to apply to all control exceptions, save for CX::Warn, which is opposite of vrurg's concern. The new CX::Returning applies to control exceptions which *carry a value* to the returnee, but in that direction all I can come up with are `Valuable` / `Valued` / `Payloaded` neither of which sounds good.

vrurg: > Sure. What is your suggestion?

Unfortunately, my English fails me here. In a personal project I'd name it `Payloadish`, but not sure if everybody would like it here. 

Alternatively, my brain have came up with `CX::Carrier` variant a few minutes ago.

niner:
CX::WithPayload?

src/core.c/Exception.pm6
gfldex: How about `class CX::SomeNameHere does Payload {...}` ?

vrurg: > How about `class CX::SomeNameHere does CX::Payload {...}` ?

Look undoubtedly better than `CX::Returning`. I'd stop here.

I still like `CX::Carrier` anyway... ;)

gfldex:
`CX::WithPayload` does what's on the tin. Tests in https://gist.github.com/gfldex/c5f5b4d0192a1abd17b48b9554be8004 have been updated. Local `make spectest` have been fine 3x in a row.

vrurg:
There are a lot of stray commits in this PR. Looks like an incorrect merge took place.

lizmat:
Is this PR still salvageable?  If not, then I guess we better close it.

lizmat:
Closing as no response in the past 4 months.

Issue #4250, opened 2021-03-14
Add algebraic function "fractional"


tbrowder:
## Environment
* Operating system:
* Compiler version (`perl6 -v` or `raku -v`):

ugexe:
fwiw I think having the unhufmanized frac is confusing enough that I wouldn’t have included it myself... but having both frac and fraction? I don’t agree that makes things any less confusing.

tbrowder:
I would like to add both (or all three) versions, but defer to better consensus.

vrurg:
Maybe I'm not fully into the situation, but can't a named parameter be added to the routine which would define it's behavior? So far, from what I see on Wikipedia, it looks like the best way of doing things would be:

1. `frac` will default to the third definition. I.e. `frac(-1.3) == -0.3`.
2. The current implementation is then as easy to get as `abs(frac($x))`.
3. The first definition, i.e. the `0.7` outcome, is implemented with `frac(-1.3, :floor)`

tbrowder:
I revamped the issue.

codesections:
Since this issue is where the discussion should take place, I'll copy what I wrote in the merged revert PR:

> I think `frac` is common/useful enough to have in core, but not common/familiar enough to justify being hufmanized down to 4 characters.
> 
> It seems like a more descriptive name would solve the ambiguities discussed upthead. And, for someone working in a domain that uses this function, it's easy enough to add a my `&frac = &absolute-fractional-part` or whatever. (That name is just an example, not a suggestion).

Personally, I'd find `frac(-1.3, :floor) == 0.7` very confusing, given that `floor(frac(-1.3)) == -1`.  (I know that the two operations aren't commutative, but my point is that I wouldn't have clear expectations of how a `:floor` argument would work).

Here's a probably off-the-wall idea: could we address the same need by adding a `prefix:<⌊>` floor operator and letting users define their own `frac` inline?  Looking at the formulas you posted, I can't help but notice that `$x - ⌊$x` is only one character longer than `frac($x)` – and totally avoids the ambiguity.

tbrowder:
I've been editing the revamped issue steadily and it's now stable--at least I believe the formulae are correct. I propose adding the functions in the table to Raku.

I will publish very soon a module implementing them.

tbrowder:
Module published: see in [modules.raku.org](https://modules.raku.org) module **Math::FractionalPart**.

b2gills:
[...]

PR #4251, opened 2021-03-14
Revert "Add new routine 'frac'"

tbrowder:
Reverts rakudo/rakudo#4244

Too much discovered too late, and I acted too hastily. The issue needs more discussion. See references with issue #4250.

**UPDATE**

In the meantime, I plan to publish a module (**Math::FractionalPart**) to more clearly show some options and help discussions toward the best way to proceed.

### DIFF
--- a/src/core.c/Cool.pm6
+++ b/src/core.c/Cool.pm6
@@ -56,7 +56,6 @@ my class Cool { # declared in BOOTSTRAP
     method floor()          { self.Numeric.floor        }
     method ceiling()        { self.Numeric.ceiling      }
     method truncate()       { self.Numeric.truncate     }
-    method frac()           { self.Numeric.frac         }
 
     ## string methods
 
--- a/src/core.c/Real.pm6
+++ b/src/core.c/Real.pm6
@@ -58,10 +58,6 @@ my role Real does Numeric {
     multi method log(Real:D: Real $base) { self.Bridge.log($base.Bridge) }
     proto method exp(|) {*}
     multi method exp(Real:D: )           { self.Bridge.exp               }
-    
-    method frac(Real:D:) {
-        abs(self - self.Int)
-    }
     method truncate(Real:D:) {
         self < 0  ?? self.ceiling !! self.floor
     }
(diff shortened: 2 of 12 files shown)


lizmat:
It feels indeed like `frac` was merged too soon: I was not aware of the different definitions of `frac` and `fractional`.  I do feel there's a benefit in there being an easy way to get to the fractional part of a real value, but as to its name or *exact* semantics, I have very little to add to the discussion, I'm afraid.

vrurg (approved):
Same on my side, as it  is for lizmat: I had a feeling that not everything that easy with negatives.

Altai-man:
IMO astronomical computations seem like a pretty specific field to keep something in core. What are the net benefits over just keeping a module with helpers for the field?

lizmat:
> IMO astronomical computations seem like a pretty specific field to keep something in core

Wanting to get to the fraction of a real value, definitely is not something just being done in astronomical calculations.  :-)

Altai-man:
>Wanting to get to the fraction of a real value, definitely is not something just being done in astronomical calculations

Fair enough, never mind me then.

codesections:
My thoughts haven't really changed since we [first discussed](https://github.com/rakudo/rakudo/pull/4244#issuecomment-793958857) the idea of adding `frac` – I think its common/useful enough to have in core, but not common/familiar enough to justify being hufmanized down to 4 characters.

It seems like a more descriptive name would solve the ambiguities discussed upthead.  And, for someone working in a domain that uses this function, it's easy enough to add a `my &frac = &absolute-fractional-part` or whatever.  (That name is just an example, not a suggestion). 

>  tbrowder requested review from … codesections 22 hours ago 

I'm happy to review, but what does an approving review mean in the context of a reversion?  Approving the revert?  Approving the original change?  Or was it just a ping for discussion?

lizmat:
I've merged this so that we don't have a release with potentially changed semantics on `frac`.  I'm welcoming a new PR to be discussed and possibly merged *after* the 2021.03 release.

tbrowder:
See my revamped issue #4250 which is now my current position.

tbrowder:
> My thoughts haven't really changed since we [first discussed](https://github.com/rakudo/rakudo/pull/4244#issuecomment-793958857) the idea of adding `frac` – I think its common/useful enough to have in core, but not common/familiar enough to justify being hufmanized down to 4 characters.
> 
> It seems like a more descriptive name would solve the ambiguities discussed upthead. And, for someone working in a domain that uses this function, it's easy enough to add a `my &frac = &absolute-fractional-part` or whatever. (That name is just an example, not a suggestion).
> 
> > tbrowder requested review from … codesections 22 hours ago
> 
> I'm happy to review, but what does an approving review mean in the context of a reversion? Approving the revert? Approving the original change? Or was it just a ping for discussion?

The latter, and, as I hoped, a reviewer (@lizmat) did as I had hoped one would do.

PR #4253, opened 2021-03-16
Fix race condition when setting up continuation in ThreadPoolScheduler

niner:
The block containing the continuationinvoke gets pushed onto the
ThreadPoolScheduler's queue before the continuation is actually taken. It's
possible that a worker thread takes the block out of the queue while
$continuation is still uninitialized.

But how is that possible? All of this happens under a lock and there's a
protected section right before we push the resume block onto the queue.
Shouldn't this protected section run after we have taken the continuation
and released the lock and therefore the push only happen after the continuation
is set up?

Turns out that the protected section runs on the same thread as the setup of
the continuation. Since our locks are re-entrant, we don't actually wait there
at all. Thus we can reach the push before the lock is set up.

Fix by acquiring the same lock when invoking the continuation.

### DIFF
--- a/src/core.c/ThreadPoolScheduler.pm6
+++ b/src/core.c/ThreadPoolScheduler.pm6
@@ -151,7 +151,9 @@ my class ThreadPoolScheduler does Scheduler {
                             }
                             if $resume {
                                 nqp::push($!queue, {
-                                    nqp::continuationinvoke($continuation, nqp::null())
+                                    $l.protect: {
+                                        nqp::continuationinvoke($continuation, nqp::null())
+                                    }
                                 });
                             }
                         });
--- a/src/core.c/ThreadPoolScheduler.pm6
+++ b/src/core.c/ThreadPoolScheduler.pm6
@@ -127,7 +127,9 @@ my class ThreadPoolScheduler does Scheduler {
             if $num-handles {
                 my $continuation;
                 my $exception;
+                my $thread-id = $*THREAD.id;
                 my $l = Lock.new;
+                note "Locking in thread $thread-id with lock { nqp::where($l) }";
                 $l.lock;
                 {
                     my int $remaining = $num-handles;
@@ -151,9 +153,13 @@ my class ThreadPoolScheduler does Scheduler {
                             }
                             if $resume {
                                 nqp::push($!queue, {
-                                    $l.protect: {
-                                        nqp::continuationinvoke($continuation, nqp::null())
+                                    unless nqp::isconcrete($continuation) { # cheap check for the common case that $continuation is already set
+                                        note "Thread { $*THREAD.id } continuation not defined yet, locking {nqp::where($l)}";
+                                        $l.lock; # lock gets released as soon as $continuation is initialized
+                                        $l.unlock; # no need to hold the lock while running the continuation - no one else is gonna take it
                                     }
+                                    note "Thread { $*THREAD.id } invoking continuation { nqp::reprname($continuation) }";
+                                    nqp::continuationinvoke($continuation, nqp::null())
                                 });
                             }
                         });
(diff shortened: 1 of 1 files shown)


lizmat:
Wow!

lizmat (approved):

src/core.c/ThreadPoolScheduler.pm6
```diff
@@ -151,7 +151,9 @@ my class ThreadPoolScheduler does Scheduler {
                             }
                             if $resume {
                                 nqp::push($!queue, {
-                                    nqp::continuationinvoke($continuation, nqp::null())
+                                    $l.protect: {
+                                        nqp::continuationinvoke($continuation, nqp::null())
+                                    }
```
jnthn, line 156: We may be able to make this more efficient.

1. I think we only need to use the lock for synchronization purposes if `nqp::isconcrete($continuation)` comes back false.
2. I wonder if we really need to protect the invoke operation, or if we just want to do a `$l.lock; $l.unlock;` beforehand.

niner:
On Dienstag, 16. März 2021 13:06:11 CET Jonathan Worthington wrote:

> We may be able to make this more efficient.
> 
> 1. I think we only need to use the lock for synchronization purposes if
> `nqp::isconcrete($continuation)` comes back false. 2. I wonder if we really
> need to protect the invoke operation, or if we just want to do a `$l.lock;
> $l.unlock;` beforehand.

We certainly don't need to protect the operation itself. The lock is not used 
anymore after the continuation has been initialized.

What about nqp::until(nqp::defined($continuation), nqp::null) before the 
continuationinvoke? We're actually not that interested in the lock but if the 
continuation is initialized.

jnthn:
> We certainly don't need to protect the operation itself. The lock is not used anymore after the continuation has been initialized.

Thinking about it some more, protecting the operation itself might also cause more significant problems, because non-blocking `await` will not happen if the thread we're running on is holding a `Lock`. Instead it'll fall back on a blocking `await`. So we really do need to make sure we only use the lock for sync.

> What about nqp::until(nqp::defined($continuation), nqp::null) before the continuationinvoke?

Only if it's for a bounded number of iterations (say, 1000) and then with a fallback to taking the lock. I've been down this road plenty of times before, and "it's short enough to busy wait" is reliably regrettable. Threads can be suspended whenever the kernel fancies, and if it suspends the one we're busy-waiting for, we can loop away an entire time-slice of CPU cycles. The chances increase under system load, and can produce profiling oddities in tools like `callgrind`, which serialize all of the threads onto a single real one.

niner:
This is extremely weird. I just pushed a commit that checks $continuation once and if it's not concrete, acquires the lock and immediately afterwards releases it (just to make sure we wait for $continuation to be set. Nevertheless it still happens that we try to invoke $continuation's Scalar container afterwards. I added debug output and could catch it with a surprising result: $continuation should already be set by thread 11 when we check it in thread 16. Even the lock is already released before the other thread checks $continuation. That's why locking and unlocking doesn't help. We're not waiting for anything. It really looks like the variable bound to by one thread maintains it's original value in the other thread.

Locking in thread 11 with lock 139914746844160
Setting continuation in thread 11 with lock 139914746844160
Unlocked in thread 11
Thread 15 invoking continuation MVMContinuation
Thread 16 continuation not defined yet, locking 139914746844160
Thread 16 invoking continuation P6opaque
Locking in thread 9 with lock 139914807116816
Setting continuation in thread 9 with lock 139914807116816
Unlocked in thread 9
Thread 9 invoking continuation MVMContinuation
Unhandled exception in code scheduled on thread 16
continuationinvoke expects an MVMContinuation, got P6opaque Scalar instead

niner:
Considering yesterday's result, I figured the most likely explanation is that that debug output is a bit misleading (i.e. out of order due to async IO) and at the time of the first check the variable is not yet initialized, but that it does get initialized immediately after the check and that the real issue is that we're reading that lexical into a local register which doesn't get updated between the check and passing it to continuationinvoke. Alas looking at the bytecode that's clearly not the case. We do have 2 getlex             loc_7_obj, lex_Frame_19648_lex3_obj ops. And the getlex op does not cache any results either.

lizmat:
So, having had a 2021.03 release, is this ready to go in?

niner:
I just pushed https://github.com/rakudo/rakudo/commit/98839c8022a7a1b9fc740ac187e3830017833df1 containing just a `$l.lock; $.unlock` to get the race condition fixed, leaving a potential optimization for later.

PR #4256, opened 2021-03-20
Make signal handler setup synchronous to avoid race conditions

niner:
On MoarVM setup of a signal handler is itself an asynchronous operation. Wait
for it to complete before returning as not only our spec tests but probably
every single user expects the signal handler to be in place at that point. This
was already true for the JVM and JS backends.

Fixes a race condition in t/spec/S17-procasync/kill.t

### DIFF
--- a/src/core.c/signals.pm6
+++ b/src/core.c/signals.pm6
@@ -45,10 +45,24 @@ multi sub signal(*@signals, :$scheduler = $*SCHEDULER) {
             submethod BUILD(:$!scheduler, :$!signal) { }
 
             method tap(&emit, &, &, &tap) {
-                my $cancellation := nqp::signal($!scheduler.queue(:hint-time-sensitive),
+                my $queue := $!scheduler.queue(:hint-time-sensitive);
+#?if moar
+                my $setup-semaphore := Semaphore.new(0);
+                my $cancellation := nqp::signal(
+                    $queue,
+                    -> { $setup-semaphore.release },
+                    $queue,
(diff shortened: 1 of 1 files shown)


MasterDuke17:
Relies on https://github.com/MoarVM/MoarVM/pull/1445

niner:
I updated the commit to use a Semaphore instead of the Promise and await. With this, signal handler setup will behave synchronously even when done as part of react/whenever setup, making https://github.com/Raku/roast/commit/099d13e3b38fa48714b785076d5cf4d74f768fee obsolete.

src/core.c/signals.pm6
```diff
@@ -45,10 +45,24 @@ multi sub signal(*@signals, :$scheduler = $*SCHEDULER) {
             submethod BUILD(:$!scheduler, :$!signal) { }
 
             method tap(&emit, &, &, &tap) {
-                my $cancellation := nqp::signal($!scheduler.queue(:hint-time-sensitive),
+                my $queue := $!scheduler.queue(:hint-time-sensitive);
+#?if moar
+                my $setup-semaphore = Semaphore.new(0);
```
lizmat, line 50: := perhaps?  Saves one Scalar and indicates immutability?

niner:
> := perhaps?  Saves one Scalar and indicates immutability?

Right. Updated the commit.

niner:
Rebased and merged manually (to get the NQP bump in)

PR #4264, opened 2021-03-24
Set "allocate in gen2" while creating objects that are permarooted.

nwc10:
Given that these objects will persist for the lifetime of the interpreter,
it makes sense to create them directly in gen2.

### DIFF
--- a/src/vm/moar/ops/perl6_ops.c
+++ b/src/vm/moar/ops/perl6_ops.c
@@ -43,6 +43,9 @@ static void p6init(MVMThreadContext *tc, MVMuint8 *cur_op) {
     if (!initialized) {
         initialized = 1;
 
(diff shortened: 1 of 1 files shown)


lizmat:
I'm not sure how to interpret the breakage of tests.  Nor do I see a way to re-test.

niner:
I think `MVM_gc_allocate_gen2_default_set` and `MVM_gc_allocate_gen2_default_clear` need to be marked `MVM_PUBLIC` like `MVM_gc_root_add_permanent` for MSVC's linker to find them.

lizmat:
Does this PR still make sense after all the changes in MoarVM and new disp having landed?

niner:
No, that code is gone :)

PR #4265, opened 2021-03-25
Use new nqp::time instead of nqp::time_(i|n)

MasterDuke17:
nqp::time_(i|n) were deprecated in favor of nqp::time which returns
integer nanoseconds since the epoch.

Requires https://github.com/MoarVM/MoarVM/pull/1450 and https://github.com/Raku/nqp/pull/706


Kaiepi:
`Trait::Traced` may pop up in blin as a result of this. If I haven't updated the module by that point, ping me.

PR #4274, opened 2021-03-30
Build: Deal with signals in `system` calls, use `exec` where possible

patrickbkr:
This is a rework of #4111 taking the quirks of Windows into account.
Closes #4111 
@nwc10 Can you have a look?

We missed the case where the invoked program exits with a signal - for
example if the OOM killer takes out the setting build.

Before this change the wrapper script would calmly take the value 137
returned from `system`, shift it right by 8 and exit with status 0, which
make took as success and carried on. At which point the next command invoked
would mysteriously fail due to a missing input file.

Now we see:
    Stage start      :   0.001
    Stage parse      : 745.469
    Stage syntaxcheck:   0.000
    Stage ast        :   0.000
    Stage optimize   : 102.057
    Stage mast       : Killed
    make: *** [Makefile:1170: blib/CORE.c.setting.moarvm] Error 137

a useful error message and the build aborts.

We sadly can't use `exec` on Windows, because `exec` does not block there,
but spawns a child process and instantly continues. So we still rely on
`system` on Windows.

### DIFF
--- a/Configure.pl
+++ b/Configure.pl
@@ -16,7 +16,7 @@ BEGIN {
     my $set_config = !qx{git config rakudo.initialized};
     if ( !-e '3rdparty/nqp-configure/LICENSE' ) {
         my $code = system($^X, 'tools/build/update-submodules.pl', Cwd::cwd(), @ARGV);
-        exit 1 if $code >> 8 != 0;
+        exit 1 if $code;
         $set_config = 1;
     }
     if ($set_config) {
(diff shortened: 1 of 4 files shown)


nwc10:
Thanks for looking into this and finding a working solution for Win32. It's annoying that we have to work around what seems to be a deficiency in how Win32 perl deals with `exec`, but I can't see a better way, and even if it was fixed upstream it would be years until we could take advantage of it.

I tested this using `ulimit -S -t ...` to set a CPU time limit that would be exceeded when building the core setting. With master (as expected) the build gets confused. With this branch (merged to master to cope with the `time_i` changes) the build exits sensibly.

I think that this is good to merge.

Issue #3043, opened 2019-07-11
Cannot import NativeCall on the JVM


Kaiepi:
## The Problem
See title.

## Expected Behavior
The module should be possible to import.

## Actual Behavior
This gets thrown:
```
===SORRY!=== Error while compiling -e
Cannot import the following symbols from NativeCall, because they already exist in this lexical scope:  wint_t, wchar_t, void, ulonglong, ulong, ssize_t, size_t, longlong, long, char32_t, char16_t, bool, Pointer, OpaquePointer, CArray, &trait_mod:<is>, &refresh, &postcircumfix:<[ ]>, &nativesizeof, &nativecast, &explicitly-manage, &cglobal
at -e:1
------> <BOL>⏏say 1
Unhandled exception: Wrong number of arguments passed; expected 1..1, but got 1
  in dispatch:<::> (gen/jvm/CORE.setting:1775)
  in clone (gen/jvm/CORE.setting:43519)
  in <anon> (gen/jvm/CORE.setting:39443)
  in EXCEPTION (gen/jvm/CORE.setting:39437)
  in <anon> (gen/jvm/CORE.setting)
  in <anon> (gen/jvm/CORE.setting)
  in dispatch:<::> (gen/jvm/CORE.setting:1775)
  in clone (gen/jvm/CORE.setting:43519)
  in <anon> (gen/jvm/CORE.setting:39443)
  in EXCEPTION (gen/jvm/CORE.setting:39437)
  in <anon> (gen/jvm/CORE.setting:4939)
  in <anon> (gen/jvm/CORE.setting)
  in <anon> (gen/jvm/CORE.setting:1844)
  in <anon> (gen/jvm/CORE.setting:4779)
  in pull-one (gen/jvm/CORE.setting:4766)
  in push-all (gen/jvm/CORE.setting:3044)
  in coremap (gen/jvm/CORE.setting:4820)
  in coremap (gen/jvm/CORE.setting:4738)
  in deepmap (gen/jvm/CORE.setting:55509)
  in deepmap (gen/jvm/CORE.setting:55508)
  in infix (gen/jvm/CORE.setting:54782)
  in infix (gen/jvm/CORE.setting:54561)
  in HYPER (gen/jvm/CORE.setting:55504)
  in dispatch:<hyper> (gen/jvm/CORE.setting:1845)
  in report (gen/jvm/CORE.setting:55799)
  in report (gen/jvm/CORE.setting:55793)
[...]

PR #4277, opened 2021-03-30
Unbreak 'use NativeCall' on JVM

usev6:
... by not assigning nqp::hash at compile time. This blows up on the
JVM backend -- see https://github.com/rakudo/rakudo/issues/3043.

### DIFF
--- a/lib/NativeCall.rakumod
+++ b/lib/NativeCall.rakumod
@@ -5,7 +5,7 @@ use NativeCall::Types;
 use NativeCall::Compiler::GNU;
 use NativeCall::Compiler::MSVC;
 
-my constant $repr_map = nqp::hash(
+my $repr_map := nqp::hash(
   "CArray",    "carray",
   "CPPStruct", "cppstruct",
   "CPointer",  "cpointer",
@@ -133,7 +133,7 @@ my constant $signed_ints_by_size =
   nqp::list_s( "", "char", "short", "", "int", "", "", "", "longlong" );
 
 # Gets the NCI type code to use based on a given Raku type.
-my constant $type_map = nqp::hash(
+my $type_map := nqp::hash(
   "Bool",       nqp::atpos_s($signed_ints_by_size,nativesizeof(bool)),
   "bool",       nqp::atpos_s($signed_ints_by_size,nativesizeof(bool)),
   "Callable",   "callback",


usev6:
I'm not totally sure, but I was under the impression, that the JVM backend *did* support calls into native C libraries (via ```com.sun.jna```) -- and that the code just bitrotted. If I'm not mistaken the first commit in that area was https://github.com/Raku/nqp/commit/cc25bbc39d -- that's even before MoarVM did know about NativeCall. Maybe someone with more knowledge of the JVM backend could shed some light on this?

But anyway, your point still stands: We'd remove an optimization to fix something that no-one uses. Do you have a rough guess how much we'd lose without the optimization?

Of course, fixing the underlying problem with ```my constant $foo = nqp::hash()``` on the JVM backend would be best. But I'm afraid that's above my paygrade.

niner:
I'm pretty sure that we could get NativeCall back on JVM quite easily. I just 
forgot about JVM when introducing the nativeinvoke op. While implementing that 
op would be pretty hard, there's no hard requirement to use it.

We'd just need to keep the generic function body instead of replacing it here:
https://github.com/rakudo/rakudo/blob/master/lib/NativeCall.rakumod#L315

usev6:
Thanks for the pointer. I'll try to look into this -- unless you want to do that.

I think, we'll also need a ```nativecallinvoke``` op, but that's probably just a copy of the current ```nativecall``` op.

usev6:
I've opened a PR to add the op ```nativecallinvoke``` (as a copy of ```nativecall```): https://github.com/Raku/nqp/pull/708

> We'd just need to keep the generic function body instead of replacing it here:
> https://github.com/rakudo/rakudo/blob/master/lib/NativeCall.rakumod#L315

I looked a while, but I don't see/understand how this would help. As far as I can see we'd have to avoid calling ```!create-jit-compiled-function-body``` -- because that method uses ```nativeinvoke```.

For experimentation I've just changed https://github.com/rakudo/rakudo/blob/ac9f40d871/lib/NativeCall.rakumod#L541 to call ```!create-function-body``` a second time.

This allowed some progress, but showed another problem: The JVM backend doesn't know about ```$!call```, introduced with https://github.com/rakudo/rakudo/commit/9ef7a7e9dd.

With another workaround added, a couple of test files started to pass. But there seems to be a severe problem if an extra block is added around the definition of a native sub:

```
$ ./rakudo-j -e 'use NativeCall; { sub abs(uint32 $n --> uint32) is native { * }; say abs(-42) }'
java.lang.NullPointerException
  in sub abs at /home/christian/perl6/perl6-roast-data/rakudo.jvm/lib/NativeCall.rakumod (NativeCall) line 597
  in block <unit> at -e line 1
```

It works without the extra block:
```
$ ./rakudo-j -e 'use NativeCall; sub abs(uint32 $n --> uint32) is native { * }; say abs(-42)'
42
```

usev6:
> Of course, fixing the underlying problem with my constant $foo = nqp::hash() on the JVM backend would be best. But I'm afraid that's above my paygrade.

Well, again stubbornness seems to have paid off. I think I've found the underlying problem: https://github.com/Raku/nqp/pull/717

usev6:
This PR is no longer needed. Closing.

Issue #4266, opened 2021-03-26
subtest doesn't print header


Leont:
## The Problem

In other TAP implementations, a subtest always prints a header first (one of the TAP drafts even mandated this). E.g.
```
$ perl -MTest::More -e 'subtest "foo", sub { ok 1 }; done_testing'
# Subtest: foo
    ok 1
    1..1
ok 1 - foo
1..1
```

However, `Test.rakumod` doesn't:

```
$ raku -MTest -e 'subtest "foo", { ok 1 }; done-testing'
    ok 1 - 
    1..1
ok 1 - foo
1..1
```
I would argue that it should, not only because it's consistent but also because it's makes analysing what's going wrong much more easy when one can actually find the name of the subtest.

Note that this should be printed to `$*STDOUT`, not `$*STDERR`, so `diag` is not appropriate here.

## Environment
```
Welcome to Rakudo(tm) v2020.12.
Implementing the Raku(tm) programming language v6.d.
Built on MoarVM version 2020.12.
```

vrurg:
Implemented by #4280.

vrurg:
#4280 is merged, closing.

PR #4280, opened 2021-03-31
Make subtest output a header

vrurg:
Something I always wanted it to do. Now it behaves similar to Perl's
Test::More module and prints `# Subtest: description` before actually
starting doing its work.

This commit also introduces `:force-informative` argument for `_diag`.

Implements #4266

### DIFF
--- a/lib/Test.rakumod
+++ b/lib/Test.rakumod
@@ -1,3 +1,4 @@
+use v6.c;
 use MONKEY-GUTS;          # Allow NQP ops.
 
 unit module Test;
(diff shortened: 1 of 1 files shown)


vrurg:
I always wanted this feature.

vrurg:
Specs are adjusted with raku/roast#726

lizmat (approved):

lib/Test.rakumod
```diff
@@ -1,3 +1,4 @@
+use v6.c;
```
JJ, line 1: Any reason why you're doing this? It makes sense, but it's unrelated to the other changes.

JJ (approved):

lib/Test.rakumod
vrurg: I don't want a separate commit, let alone a PR, for just one line.

PR #4288, opened 2021-04-02
Add new methods to class DateTime

tbrowder:
The following methods have been added to class DateTime to fill in some of the missing well-known time values of the instant:

+ *day-fraction* - The instant's hours, minutes, and seconds as a fraction of a 24-hour day
+ *mjd* - The real number, formed as the sum of *daycount* (the integral part of the *Modifed Julian Date*) and *day-fraction*, known as the *Modified Julian Date* (MJD)
+ *juliandate* - The real number, formed as the sum of the *MJD* and *2_400_000.5*, known to astronomers as the *Julian Date* or *Julian Day Number*

### DIFF
--- a/src/core.c/DateTime.pm6
+++ b/src/core.c/DateTime.pm6
@@ -398,6 +398,21 @@ my class DateTime does Dateish {
         Instant.from-posix: self.posix + $!second % 1, $!second >= 60;
     }
 
+    method day-fraction(DateTime:D: --> Real:D) {
+        my $sec = self.hour * 60 * 60;
+        $sec   += self.minute * 60;
+        $sec   += self.second;
+        $sec / (24.0 * 60 * 60)
+    }
+
+    method mjd(DateTime:D: --> Real:D) {
+        self.daycount + self.day-fraction
+    }
+
+    method juliandate(DateTime:D: --> Real:D) {
+        self.mjd + 2_400_000.5
+    }
+
     method posix(DateTime:D: $ignore-timezone? --> Int:D) {
         return self.utc.posix if $!timezone && !$ignore-timezone;
 
--- a/src/core.c/DateTime.pm6
+++ b/src/core.c/DateTime.pm6
@@ -405,12 +405,12 @@ my class DateTime does Dateish {
         $sec / (24.0 * 60 * 60)
     }
 
-    method mjd(DateTime:D: --> Real:D) {
+    method modified-julian-date(DateTime:D: --> Real:D) {
         self.daycount + self.day-fraction
     }
 
-    method juliandate(DateTime:D: --> Real:D) {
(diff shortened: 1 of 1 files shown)


ugexe:
It’s strange that one compound word uses dashes to separate words while one does not. It’s also strange verbose wording is used for two of the methods, where one method is an abbreviated form of mostly one of the verbose method named.

tbrowder:
I agree that's not consistent. What do you think about these:

+ julian-date
+ modified-julian-date

Or?

jonathanstowe (approved):

lizmat:
In the light of the .fract / .fractional discussion elsewhere, would it make sense to just the `day-fractional` method called `.fract` or `.fractional` ?  Then, like other situations, you could consider a `DateTime` object just as a special kind of number?

ugexe:
> In the light of the .fract / .fractional discussion elsewhere, would it make sense to just the day-fractional method called .fract or .fractional ? Then, like other situations, you could consider a DateTime object just as a special kind of number?

I don't think most users would assume when they see a `.fractional` method on such a object that it is specifically referring to the day instead of any other date or time part that objects represents like e.g. a year fraction for determining what quarter of the year it is.

lizmat:
@ugexe Hmm.... intriguing thought: perhaps `month-fraction` and `year-fraction` could then also be a thing...  hmmm...

tbrowder:
On Fri, Apr 2, 2021 at 11:19 Elizabeth Mattijsen ***@***.***>
wrote:

> @ugexe <https://github.com/ugexe> Hmm.... intriguing thought: perhaps
> month-fraction and year-fraction could then also be a thing... hmmm...
>
Easy to do, but I don't have a use case fot it, except completeness.

>

lizmat:
I don't have an immediate use for it, but then again, I don't have that for `day-fraction` either.

The reason for me to add something like that, is that for the calculation you would need to know how many days there are in a month / year, and that's easier to fetch internally than externally.  And externally, it runs the risk of becoming out of sync (well, unlikely, but then again, who thought that we would ever have a negative leap-second: yet this appears to be on the table in some circles  :-)

tbrowder:
I just looked at the NASA JPL time site and it looks like they do show a day fraction. They also show the day number in the year (day 92 on 2021-04-02).

ugexe:
I don't think its necessarily important to include those methods. My argument was just that the methods should be named in such a way that they could be included at some time without the naming being inconsistent.

tbrowder:
So I'm happy to leave other new methods for another day. I would love to get this PR into the upcoming release if possible. What say my advisors?

lizmat:
Thank you!

tbrowder:
You're welcome, @lizmat. And thank you!

PR #4290, opened 2021-04-02
Fix recent JVM breakage (Rakudo::Unicodey)

usev6:


lizmat:
So not even an `my int @a` works in the setting on the JVM backend?  Even that late in the setting?

lizmat:
I've just pushed https://github.com/rakudo/rakudo/commit/496f986e6c in the hope it will stick on the JVM backend.  If it doesn't, let me know, and I'll make it use an untyped array.  Thing is that I will need do make all of the `nqp::shift_i` occurrences backend specific  :-(

usev6:
> So not even an my int @a works in the setting on the JVM backend? Even that late in the setting?

I'm afraid that is the case. As far as I can see native arrays are mis-compiled into non-native arrays in the setting.

So, after https://github.com/rakudo/rakudo/commit/496f986e6c the JVM backend builds, but method ```ords``` for strings does not work:
```
$ ./rakudo-j -e 'dd "foo".ords'
VMArray representation does not implement push_native
  in block <unit> at -e line 1
```

>  Thing is that I will need do make all of the nqp::shift_i occurrences backend specific :-(

Do you refer to upcoming changes? Because for now things seem to work if an untyped array is used (as in https://github.com/rakudo/rakudo/pull/4290/commits/6014cfa81b99f5d3b4069bbb0fbcb3220308b1e6).

```
$ ./rakudo-j -e 'dd "foo".ords'  ## with an untyped array and without nqp::push_i
(102, 111, 111).Seq
```

lizmat:
Hopefully fixed with https://github.com/rakudo/rakudo/commit/b88e1ca3bf  Please close this PR if so  :-)   Thanks for the nudging and the PR!

usev6:
It looks like we finally found a solution: https://github.com/rakudo/rakudo/pull/4293 -- and it doesn't even require many quirks.

The main discovery for me is that
```
my int @array;
```
is miscompiled in the setting (on the JVM backend, that is), while
```
my @array := array[int].new;
```
seems to work. That's good to know.

I'm closing this PR.

PR #4292, opened 2021-04-03
Add tests for former REPL crashes

usev6:
We didn't have regression tests for evaluations that caused the REPL to crash badly.
The crashes are described in two bug reports from the old issue tracker:

* https://github.com/Raku/old-issue-tracker/issues/2917
* https://github.com/Raku/old-issue-tracker/issues/3254

See commits for some more details.

### DIFF
--- a/t/02-rakudo/repl.t
+++ b/t/02-rakudo/repl.t
@@ -3,7 +3,7 @@ use lib <t/packages>;
 use Test;
 use Test::Helpers;
 
-plan 45;
+plan 46;
 
 my $eof = $*DISTRO.is-win ?? "'^Z'" !! "'^D'";
 my $*REPL-SCRUBBER = -> $_ is copy {
@@ -244,6 +244,13 @@ is-run-repl ['Nil'], /Nil/, 'REPL outputs Nil as a Nil';
         :out(/'The value is 42'/),
     'variables persist across multiple lines of input';
 }
+
+# https://github.com/Raku/old-issue-tracker/issues/2917
+is-run-repl "my \\a = any set <1 2 3>;\nsay 1 ~~ a",
+    :out{
+        .contains('any') and .contains('False') and not .contains: 'True';
+    },
+    'REPL does not report True and False for junction';
 
 {
     # If the REPL evaluates all of the previously-entered code on each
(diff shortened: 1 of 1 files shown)


vrurg:
May I ask you to have better description in the first comment and your commits in general? When PR is merged the comment is used as the default message for the merge commit. And overall, just referring to issues is not sufficient because the links may be gone with time (as it already happened with the old tracker). We even have a guide on writing good commit messages [here](https://github.com/rakudo/rakudo/blob/master/CONTRIBUTING.md#commit-messages)

vrurg (changes requested):
Fix commit messages

usev6:
I didn't realize that the first comment on a PR will be used as the default message for the merge commit. Thanks for pointing that out.

I kind of see your point, that the two commit messages are very terse (maybe overly terse). On the other hand the commits just added relatively simple tests that should be self-explaining. I was aware of our guide on writing good commit messages, but I didn't interpret it in a way, that the content of a referenced issue should be summed up in the commit message. Maybe the following part should be reworded to make that clearer:

> If you fixed a ticket, or the commit relates to a specific ticket, please mention the ticket in the title or the body as #12345 so that GitHub will automatically link to the corresponding issue. If there was an IRC conversation that can give some background or useful information, you can link to it by visiting the IRC logs and linking the link provided by the timestamps on the left side of the page.

I have to admit that your requests for ```better descriptions [...] in your commits in general``` hurts me a bit (the ```in general``` part). (At least usually) I try to write good commit messages -- but maybe I'm not able to meet the standards. I'll ponder on this a bit.
(Please note, that I'm fully aware that you -- or other reviewers -- have better things to do than trying to train people to write better commit messages. At my day job I'm the one who has to do that.)

vrurg:
> I have to admit that your requests for `better descriptions [...] in your commits in general` hurts me a bit

Oops... I'm sorry as I didn't mean that! I only referred to the two commits of this PR because they're as terse the the PR comment. Not for a second I thought about any other ones than these two. I sincerely apologize for this! My certainly imperfect English plays some tricks with me sometimes too.

Also, there is no need to close this PR. The comment can be edited, the commits squashed, amended, and force-pushed back. That's what I'd do.

usev6:
Thanks for clearing up the misunderstanding! So all is good :).
I'll rework the commits and this PR at a later time.

usev6:
@vrurg I hope, I have addressed the points you raised and the current version is good to merge.

vrurg (approved):
Looks good. Thank you!

PR #4295, opened 2021-04-03
[JVM] Limit special cases to variable declaration

usev6:
We have a couple of workarounds for native arrays in the setting not
working correctly on the JVM backend. Having something like

  my int @arr;

was compiled to a VMArray instead of a VMArray_i. It turned out that
binding to a new native array does work:

  my @arr := array[int].new;

By using this construct for the JVM backend the scope of the needed
special casing is reduced. (Also a really bad hack was removed from
src/core.c/ThreadPoolScheduler.pm6, were we relied on the mentioned
miscompilation.)

### DIFF
--- a/src/core.c/ThreadPoolScheduler.pm6
+++ b/src/core.c/ThreadPoolScheduler.pm6
@@ -565,8 +565,12 @@ my class ThreadPoolScheduler does Scheduler {
 
                 scheduler-debug "Supervisor started";
                 my int $last-rusage-time = nqp::time;
+#?if jvm
+                my @rusage := array[int].new;
+#?endif
 #?if !jvm
                 my int @rusage;
+#?endif
                 nqp::getrusage(@rusage);
                 my int $last-usage =
                   nqp::mul_i(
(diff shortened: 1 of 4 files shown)


lizmat:
Why not use the `my @arr := array[int].new;` form for MoarVM backend as well?

Or fix the miscompilation?  :-)

If the latter is problematic, perhaps then we can do the former instead?  I really would like to reduce the amount of conditional compilation in the setting, not increase it or replace it by something else  :-)

usev6:
Fixing the miscompilation would be the best thing for sure. I'm afraid I'm not able to do that.

For why I didn't went for the form ```my @arr := array[int].new;```:
* I'm unsure whether that might have some other impact on the MoarVM backend (be it a semantic difference or changes in performance).
* The version used for MoarVM (```my int @arr;```) seems to be cleaner, so I have a (weak) preference to keep it.
* My line of thought was that isolating the scope of the special cases already helps a lot. It makes it clear that it is just a workaround, no deeper difference in behaviour between the backends.

So, a change of the code for MoarVM could (and IMHO should) be done in a separate commit. But I'm not sure if it's worth it.

lizmat:
Looking at the `--target=ast` output, they codegen quite differently, so merging this as it is now.  Thanks!

usev6:
For the sake of completeness: The workaround has been removed with https://github.com/rakudo/rakudo/pull/4400

PR #4300, opened 2021-04-04
Add NFC/NFD/NFKC/NFKD as subs

lizmat:
These functions were only available as Cool.methods.  This makes them
also available as subs for consistency.

Also contains changes to tests for the now globally visible subs.


jnthn:
Should they not just work "naturally" (and with identical syntax) as type coercions?

lizmat:
You mean something like `my NFC() @a = "foobar";`  ??

jonathanstowe:
It already works as a coercion:

```
raku -e 'say NFC("foobars")'
NFC:0x<0066 006f 006f 0062 0061 0072 0073>
```

lizmat:
Well, yes, but I guess we need the sub versions for introspection?

jnthn:
> Well, yes, but I guess we need the sub versions for introspection?

The types in scope are just as introspectable, I think?

lizmat:
I guess  :-)

PR #4302, opened 2021-04-05
Support backtrace creation from nqp::null on JVM

usev6:


usev6:
Spectest for MoarVM still running, but it looked good on JVM backend.

niner (approved):

PR #4303, opened 2021-04-05
Don't add a delegated method if target already has it

vrurg:
Resolve a case like:

```raku
    class C {
        has $.backend handles <Str>;
        method Str { ... }
    }
```

Whereas this particular snipped may look silly, things are less fun if
the attribute is declared in a role; espcially in one which is not
controlled by us.

### DIFF
--- a/src/core.c/traits.pm6
+++ b/src/core.c/traits.pm6
@@ -449,11 +449,14 @@ multi sub trait_mod:<handles>(Attribute:D $target, $thunk) {
         }
 
         method add_delegator_method($attr: Mu $pkg, $meth_name, $call_name) {
-            my $meth := method (|c) is rw {
-                $attr.get_value(self)."$call_name"(|c)
-            };
-            $meth.set_name($meth_name);
-            $pkg.^add_method($meth_name, $meth);
+            # Don't add a method to the destination type object if it already has a method with the same name.
+            unless $pkg.^declares_method($meth_name) {
+                my $meth := method (|c) is rw {
+                    $attr.get_value(self)."$call_name"(|c)
+                };
+                $meth.set_name($meth_name);
+                $pkg.^add_method($meth_name, $meth);
+            }
         }
 
         method apply_handles($attr: Mu $pkg) {


jnthn:
> Whereas this particular snipped may look silly

It does, and I think that's the kind of silly that should be an error.

> things are less fun if the attribute is declared in a role

In this case I'd expect the method in the class to win, as is the general principle.

A less sweeping solution (only silently letting the class method win when the attribute and its implied method came from a role) seems fine to me, but I think the exact case specified in the example should remain a redeclaration error.

vrurg:
Ok, I finally for more time to elaborate on this case. First, this is how I got here. I have a role `Container` which looks alike to this draft:

```raku
role Container {
    has $.payload handles @coercion-methods;
}
```

How exactly `@coercion-methods` is built is not really relevant. Let's assume it is provided externally. Then there is actual implementation class which does some lazy string expansion. One of the ways to trigger the expansion is to coerce a class instance into a string. Something like:

```raku
class Expandable::String does Container {
    method Str {
        self!maybe-expand;
        $!payload.Str
    }
}
```

For this case it is really clear that `handles` must not install method `Str`. There is another case, though. What if the list delegated methods is defined by a 3rd party module?

```raku
unit module Foo;

our @delegations = build-delegation-list();
```

```raku
use Foo;
class Bar {
    has $.a handles @Foo::delegations;
    method Str { ... }
}
```

In this case I may not have control over what `@Foo::delegations` content is. Consequently, I will need to take special measures to filter out `Str` from the list _if_ it is there.

It all gets even more confusing if the attribute with `handles` is declared in a role. All a developer would know about the role is that it provides a couple of methods. One of them the consuming class want to override. But, alas! it can't because it's a delegated one:

```raku
unit role R;
has $!a handles <Str Int Bool>;
```

```raku
use R;
class C does R {
    method Int {...} # SORRY! Package 'C' already has a method 'Int'
}
```

Isn't it too much of confusion for the only reason of preventing the silliness of the first example?

lizmat:
Hopefully unstalling this PR with this nudge.

vrurg:
@lizmat Perhaps it can be closed as @jnthn is not happy about the idea. Perhaps it worth considering implementing the feature for roles only, to cover the problems I mentioned. But not sure if it will eventually make it into the code and thus if it makes sense to spend time on it.

PR #4304, opened 2021-04-05
Make CLIENT:: work for code invoked from NQP world

vrurg:
Replace use of `===` with `nqp::eqaddr` because the former expects `Any`
arguments.


lizmat:
Cool!  Wonder if there are more places like that in the setting that are using `===` this way.  OTOH, the `nqp::eqaddr` semantics coincide with `=:=`, so did we change the semantics here subtly, or should it have been a `=:=` in the first place?

vrurg:
`=:=` would do. My only reason for using `nqp::eqaddr` is because `PseudoStash` is already a heavy user of `nqp::*`.

With regard to semantics, overall `===` appeals to me because it is documented as `value identity operator`. When applied to type objects, this is what I usually expect: no matter if they're containerized or not, one gets the result. I would rather consider making its arguments `Mu`-typed as there is nothing in it binding it to `Any`.

PR #4308, opened 2021-04-07
Remove uniprop-(int|str|bool)

lizmat:
They appear to have been added at some point, but they are not tested
nor are they documented, nor are they needed in the setting nor any
core modules.

Their functionality seems to intended as a performance shortcut for the
more general "uniprop" and "uniprops" functions.  My argument would be
that *if* one knows about these, one might as well use the underlying
nqp::ops as these appear to be a very thin HLL wrapper for them.


raiph:
If someone wants them later they can add them back.

But "My argument would be that .... one might as well use the underlying nqp::ops as these appear to be a very thin HLL wrapper for them." seems like an extraordinary argument. NQP is an implementation detail. If someone *needs* functionality, then surely the *last* thing we should be doing is suggesting they use an implementation detail?

I mean, if there's no other way, and the person is clearly told that under no circumstances are they helping Raku if their use of unsupported implementation details ever directly or indirectly, deliberately or accidentally, ends up putting pressure on core devs to not break their code in future years, then fair enough. But would it be that way?

lizmat:
> one might as well use the underlying nqp::ops as these appear to be a very thin HLL wrapper for them

This was really a continuation of the thought that these methods were for core dev debugging, something which was later confirmed by @samcv on IRC.

> If someone wants them later they can add them back.

I could just put those subs into a module :-)   Adding them back would then be something a user could do for their context.

japhb:
Sigh, I was using those: https://github.com/japhb/Text-MiscUtils/blob/b11149150b83f735cda87ba979e93d389ebbaed0/lib/Text/MiscUtils/Layout.pm6#L20

That said, no need to roll back; I'll switch to the general ones.

Before deleting existing Rakudo functionality though, it would be good to check the ecosystem as well.  Don't even need to use Blin, I think -- Greppable6 should do, yes?

japhb:
Text-MiscUtils 0.0.6 released with `uniprop` fixes.

lizmat:
@japhb Well, I was assuming a Blin run *would* occur before the release, and it would find this issue then.

PR #4312, opened 2021-04-10
Add support for ∊ and ∍ as aliases for ∈ and ∋

moon-chilled:


src/core.c/set_elem.pm6
```diff
@@ -92,8 +94,9 @@ proto sub infix:<(cont)>($, $, *%) is pure {*}
 multi sub infix:<(cont)>(\a, \b --> Bool:D) { b (elem) a }
 
 # U+220B CONTAINS AS MEMBER
-proto sub infix:<∋>($, $, *%) is pure {*}
-multi sub infix:<∋>(\a, \b --> Bool:D) { b (elem) a }
```
lizmat, line 96: Good catch!

lizmat (approved):

lizmat:
@jnthn would this pose a difficult issue for Comma, or would it be just adding these ops as alias somewhere?

jnthn:
@lizmat Not really. I mean, the translation of `(elem)` into a Unicode form will probably just stay as `∈` rather than going so far as offering the choice, but the Unicode to ASCII translation for both shouldn't be an issue.

I do wonder if at some point we may want to make things like `∊` be defined as a macro that rewrites into `∈`, so that folks wanting to override them only need do the one form. That question extends rather beyond just these operators, though.

I got curious why there are two forms, and if there are other small versions of other set operators, but it appears not.

lizmat:
Yeah and googling for semantic differences between the two forms, doesn't appear to bring up anything.

jnthn:
It has [a comment](https://www.fileformat.info/info/unicode/char/220a/index.htm) "originates in math pi fonts; not the straight epsilon", which gives a bit more to go on (for example, [this SO question](https://stackoverflow.com/questions/13188587/conversion-of-mathematicalpi-symbol-names-to-unicode).

Anyway, I guess we accept it on the same basis we accept smart quotes: people could easily find content using it, paste it into their Raku programs, and it's nice if it just works rather than failing because it's the wrong kind of Unicode element of.

lizmat:
Also for future reference: there appear to be more small versions around

https://aplwiki.com/wiki/Unicode

That page may contain other candidates.

Most notable:
- * vs ⋆    (ASTERISK vs STAR OPERATOR)
- ~ vs ∼  (TILDE vs TILDE OPERATOR)

PR #4319, opened 2021-04-11
Use correct api value when calculating dist id

ugexe:
When calculating a dist id we treat empty strings as value distinct from 0, but they should be (and generally are) treated the same.


ugexe:
This allows the following to actually uninstall the META6.json file for the distribution:
`raku -e 'my $cur = CompUnit::RepositoryRegistry.repository-for-name("site"); my $dist = $cur.candidates("File::Find").head; say $cur.uninstall($dist);`

ugexe:
Actually this isn't the correct fix, it just happens to fix the 1 liner albiet not for the reason I was thinking

PR #4320, opened 2021-04-12
Speedup exponentiation by Ints

MasterDuke17:
Before, `my $a; $a := $_ ** 2 for 1..10_000_000; say now - INIT now;`
took ~7.1s, now it takes ~5.7s. However, `my $a; $a := $_ ** -2 for 1..1_000_000; say now - INIT now;`
took ~2.3s and now takes ~3.7s. The slowdown in the -2 case is because
of the non-default (smaller) epsilon passed to .Rat, but without it
`1_000_000 ** -2` would result in 0 instead of 0.000000000001.

### DIFF
--- a/src/core.c/Int.pm6
+++ b/src/core.c/Int.pm6
@@ -368,12 +368,14 @@ my constant UINT64_UPPER = nqp::pow2_I(2, 64, Int);
 
 multi sub infix:<**>(Int:D $a, Int:D $b --> Real:D) {
     # when a**b is too big nqp::pow2_I returns (Int)
-    nqp::isge_I($b, 0)
-      ?? nqp::isconcrete((my $power := nqp::pow2_I($a, $b, Int))) 
+    nqp::isconcrete((my $power := nqp::pow2_I($a, $b, Int)))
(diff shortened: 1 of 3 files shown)


MasterDuke17:
Forgot to mention timings in the 3rd commit message, but `my $a; $a := $_ ** 2 for 1..10_000_000; say now - INIT now` is now ~4.07s and `my $a; $a := $_ ** -2 for 1..1_000_000; say now - INIT now` is now ~3.1s.

MasterDuke17:
Where it stands now (at the 6th version):
`my $a; $a := $_ ** 2 for 1..10_000_000; say now - INIT now` is ~4.2s, compared to ~7.1s for master.
`my $a; $a := $_ ** -2 for 1..1_000_000; say now - INIT now` is ~1.5s, compared to ~2.3s for master.

lizmat:
Is this PR salvageable?

MasterDuke17:
> Is this PR salvageable?

Yeah. I was actually looking at it again a little while ago. I'm not 100% sure what it'll look like in the end, but at least some Rakudo-only changes should land.

PR #4324, opened 2021-04-13
Allow negative Unix (POSIX) values for DateTime instantiation

tbrowder:
The purpose of this PR is to allow negative Unix epoch values to be used so dates earlier than the Unix epoch can be instantiated with instants from older epochs, particulary the Julian date epoch.

### DIFF
--- a/src/core.c/Any-iterable-methods.pm6
+++ b/src/core.c/Any-iterable-methods.pm6
@@ -874,7 +874,7 @@ Consider using a block if any of these are necessary for your mapping code."
             );
             $!index = $i;
         }
-        method deterministic(--> Bool:D) { $!iter.deterministic }
+        method is-deterministic(--> Bool:D) { $!iter.is-deterministic }
     }
     method !grep-k(Callable:D $test) { Seq.new(Grep-K.new(self,$test)) }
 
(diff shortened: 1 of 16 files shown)


lizmat:
I would put the code handling `epoch < 0` into a separate private method (e.g. `!bc`, and then do something like:

    return self!bc($epoch) if $epoch < 0;

tbrowder:
> I would put the code handling `epoch < 0` into a separate private method (e.g. `!bc`, and then do something like:
> 
> ```
> return self!bc($epoch) if $epoch < 0;
> ```

Good idea, WILCO!

tbrowder:
I lied on #raku-dev. I had not installed my change before testing. Fundamentally the problem is DateTime's handling of negative dates is the break point, not negative epochs. I made another epoch defined at 0001-01-01T00:00:00Z and compare unix epochs against it. All epochs >= to it are handled as usual, but epochs less than that get the special handling.

Given that change, all current spectests pass, as well as the few tests I have created but not added to roast yet.

Now I will work on adding tests across the new domain of, say, years -4000 to +4000.

tbrowder:
I'm discovering more nuances in the current code that interfere with my approach. For one thing, my experiments show that the Claus Tondering algorithm implementation is wrong. I believe the algorithms by Meeus (as used in Perl's Astro::Montenbruck by Sergey Krushinsky) are the ones to be used. 

I am closing this PR to work on another attempt after @MasterDuke17's PRs are merged.

PR #4325, opened 2021-04-13
Add a `repl` sub to enter the repl from within a program

lizmat:
This would allow one to enter the REPL from a non-interactive program.
Ideally, this should be aware of its context, but apparently setting
$!save_ctx is not enough.  Maybe @vrurg has some ideas about that.

Sorta inspired by some comments on the #raku channel today.

### DIFF
--- a/src/core.c/REPL.pm6
+++ b/src/core.c/REPL.pm6
@@ -440,13 +440,13 @@ do {
 
             $!history-file.absolute
         }
-
-        method here() {
-            my $repl := self.new(nqp::getcomp("Raku"),%_);
-            nqp::bindattr($repl,REPL,'$!save_ctx',nqp::ctxcaller(nqp::ctx));
-            $repl.repl-loop
-        }
     }
+}
+
+sub repl(*%_) {
+    my $repl := REPL.new(nqp::getcomp("Raku"),%_);
+    nqp::bindattr($repl,REPL,'$!save_ctx',nqp::ctxcaller(nqp::ctx));
+    $repl.repl-loop
 }
 
 # vim: expandtab shiftwidth=4
--- a/src/core.c/REPL.pm6
+++ b/src/core.c/REPL.pm6
@@ -440,6 +440,12 @@ do {
 
             $!history-file.absolute
         }
+
+        method now() {
+            my $repl := self.new(nqp::getcomp("Raku"),%_);
+            nqp::bindattr($repl,REPL,'$!save_ctx',nqp::ctxcaller(nqp::ctx));
+            $repl.repl-loop
+        }
     }
 }
 
--- a/src/core.c/REPL.pm6
+++ b/src/core.c/REPL.pm6
@@ -441,7 +441,7 @@ do {
             $!history-file.absolute
         }
 
-        method now() {
+        method here() {
             my $repl := self.new(nqp::getcomp("Raku"),%_);
             nqp::bindattr($repl,REPL,'$!save_ctx',nqp::ctxcaller(nqp::ctx));
             $repl.repl-loop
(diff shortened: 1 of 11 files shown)


Altai-man:
No merge before release.

ugexe:
We should avoid methods named identically to those of other core methods that do something completely different (I still 🤦 over creating `CUR.candidates` when `Routine.candidates` already existed)

lizmat:
@ugexe Good point.  I chose "now" because, well because I want a REPL now.  Maybe `here` would be a better name.

Any suggestions on making it see its context?

lizmat:
@Altai-man Wasn't planning on merging this before release.  I changed it to "Draft" to make that clearer!

vrurg:
@lizmat Unfortunately, I can't be really helpful with the context. It looks like it should work. Except that, perhaps, passing it as `:outerctx($caller-ctx)` argument to `repl-loop` method would do the same but cleaner. I then traced it down to `HLL::Compier` `eval` method where it is used exactly for that – to set the outer context of the compiled code. Hence can't tell what's going on without debugging. Will see if I can spare some time for this, but can't make any promise now.

ugexe:
`here` seems too generic for how often I expect this specific functionality to be used... after all I literally want any line of code I write to be `here` (and usually for it to run `now`).

I'd probably go more towards something like `REPL.enter-foo-mode` or `REPL.initiate-loop`

vrurg:
> I'd probably go more towards something like `REPL.enter-foo-mode` or `REPL.initiate-loop`

`enter-loop`?

lizmat:
FWIW, I see this as a debugging tool, so I would like to have it as short as possible.

Making it a sub called `repl` might also be an option?

src/core.c/REPL.pm6
```diff
@@ -440,6 +440,12 @@ do {
 
             $!history-file.absolute
         }
+
+        method here() {
+            my $repl := self.new(nqp::getcomp("Raku"),%_);
+            nqp::bindattr($repl,REPL,'$!save_ctx',nqp::ctxcaller(nqp::ctx));
```
jnthn, line 446: This isn't going to go well. The compiler is free to optimize out lexicals as it pleases. We do have a list of magical sub names, such as `EVAL`, which inhibit compiler optimization in all outer scopes. (Others include `eval-lives-ok` for `Test`.) I'm not thrilled about this, but at least hope that list can go away in the future by making all such things instead be macros that do...something...to inhibit the optimizations (the point being that the compiler doesn't need to keep a magical list of names).

But that means the things in question should be *sub* calls today, *not* method calls. We do, I grant, have a `.EVAL` method, which I consider an unfortunate mistake, although we could make that a `macro postfix:<.EVAL>` or something (and hope it wins on the LTM longest literal prefix rule). (I'd personally lean towards deprecation, though.)

Anyway, if we really should have this, please make it a `sub`. It'll have to go in the same magical list of optimization inhibitors that `EVAL` is also in.

lizmat: OK, I will make it a sub called `repl`.

@jnthn where does the magical sub names list live?

jnthn: > OK, I will make it a sub called repl.

That's succinct.

> where does the magical sub names list live?

`src/Perl6/Optimizer.nqp` I believe.

lizmat: Aah, that's why I couldn't find it: `eval-lives-ok` is **not** in that list, but `throws-like` is.

lizmat:
Trying to catch a user doing Ctrl-d, or resetting the eof / error flag on STDIN, is still out of reach.  But at least typing "exit" now in `repl` will **not** actually exit the program, but just exit the repl-loop, making it re-entrant and thus a simple debugging tool as well.

Therefore I marked this PR now ready for review.  As this is new functionality not touching on current functionality, I'd say it's pretty safe to merge before 2021.05 release.

PR #4333, opened 2021-04-18
Fix 'too many open files' during module uninstall

ugexe:
Recently uninstalling modules might give a 'too many open files' error. By making a call to `.dir` eager this error goes away, although I'm not clear why or what changed in e.g. libuv that would require this now.

### DIFF
--- a/src/core.c/CompUnit/Repository/Installation.pm6
+++ b/src/core.c/CompUnit/Repository/Installation.pm6
@@ -81,7 +81,7 @@ sub MAIN(:$name, :$auth, :$ver, *@, *%) {
 
         for $short-dir.dir -> $dir {
             $dir.add($id).unlink;
-            $dir.rmdir unless $dir.dir;
+            $dir.rmdir unless $dir.dir.elems;
         }
     }
 


ugexe:
see https://gist.github.com/ugexe/453b7f6ce7832fc0af33c6c3426c1d12

ugexe:
While it seems like it only started happening recently the following code shows the same issue on other versions of rakudo:

```
✗ raku -e 'for 1..100 -> $a { mkdir $a; for 1..10 -> $b { $a.IO.add("foo.txt.$b").spurt("666"); }; }; for $*CWD.dir.eager -> $dir { for 1..10 -> $b { $dir.add("foo.txt.$b").unlink; $dir.rmdir unless $dir.dir; }; }'

Failed to get the directory contents of '/Users/ugexe/repos/foo/17': Failed to open dir: Too many open files
  in block  at -e line 1
  in block <unit> at -e line 1
```

Changing `$dir.rmdir unless $dir.dir;` to `$dir.rmdir unless $dir.dir.elems;` (similar to what this PR does) allows it to run to completion on all the versions of rakudo I tried.

lizmat:
The problem is really two-fold, in my opinion.

- When `.Bool` is called on the `Seq`, **one** element is fetched from the iterator, and then the iterator is basically abandoned.
- `dir` creates an `Iterator` that gets embedded into a `Seq`.  Calling `.Bool` on such a `Seq` will fetch one directory element, and then abandon the iterator **leaving the internal dirhandle open**.

Now, I'm pretty sure that when the Iterator object gets garbage collected, the `$!dirhandle` attribute is also closed.  But maybe we should verify that: if it is not, we may need to add a `DESTROY` method to these iterators (`Rakudo::Iterator::Dir` and `Rakudo::Iterator::DirTest`).

Meanwhile, I've found one case where the dirhandle was *not* closed immediately after exhausting it, and fixed it with https://github.com/rakudo/rakudo/commit/62c976a43c .

lizmat:
Perhaps a more general solution to this type of issue, is to create a subclass of `Seq` that would have a special `Bool` method, which in turn would call the `Bool` method on the iterator (which is *not* in the standard `Iterator` API), which would allow the iterator to special case that.

For `IO::Path.dir`, this would basically mean:

````raku
method Bool(--> Bool:D) {
    my $has-entries := nqp::hllbool(nqp::nextfiledir($!dirhandle));
    nqp::closedir($!dirhandle);
    $has-entries
}

b2gills:
A more general fix would be for `Seq.Bool` to just sink the iterator after calling `.pull-one`, by following it with`.sink-all`.
Presumably a directory iterator would know to just close the handle when `.sink-all` is called on it.

The problem is that `.cache` and `PositionalBindFailover` are things which exist.

(Personally I think that the whole sneaky caching of `Seq` to be a fundamental mistake. If you want a cached `Seq` turn it into a `List` first. Basically you would only ever be able to reliably do a single operation on `Seq`, better make it count.)

lizmat:
> Presumably a directory iterator would know to just close the handle when .sink-all is called on it.
(Just checked, no it doesn't know that.)

That would be easily fixed.

> Personally I think that the whole sneaky caching of Seq to be a fundamental mistake. If you want a cached Seq turn it into a List first. Basically you would only ever be able to reliably do a single operation on Seq, better make it count

I happen to wholeheartedly agree with that.  But changing that, would at least require a language version, but more importantly, a good way to dispatch on the **caller's** language version.

PR #4352, opened 2021-05-02
Replace use of P6EX hllsym with Metamodel::Configuration.throw_or_die

vrurg:
The advantages of this move are:

- less boilerplate code as one doesn't need to add an entry to `P6EX`
  hash if wants to throw an exception from Metamodel code
- much better readability
- should also reduce the number of closure allocations as the code
  objects stored in `P6EX` are now gone

What is not solved yet:

- Java/JS backends. I just don't know how to deal with them
- perl6_ops.c which is currently the only case where `P6EX` is used

### DIFF
--- a/src/Perl6/Metamodel/Configuration.nqp
+++ b/src/Perl6/Metamodel/Configuration.nqp
@@ -39,43 +39,20 @@ class Perl6::Metamodel::Configuration {
     }
     method role_to_role_applier_type() { $role_to_role_applier_type }
 
-    my $X_package := nqp::null();
-    method set_X_package($X) {
-        $X_package := $X;
-    }
-    method find_exception($exception) {
-        my $ex_type := nqp::null();
-        unless nqp::isnull($X_package) {
-            my @parts := nqp::split('::', $exception);
-            # If the first part of the long name is not X then pretend there is no such exception. Which is, actually
-            # and most likely, is true.
-            if nqp::iseq_s(nqp::shift(@parts), 'X') {
-                my $who := $X_package.WHO;
-                while +@parts {
-                    my $name := nqp::shift(@parts);
-                    if $who.EXISTS-KEY($name) {
-                        if +@parts {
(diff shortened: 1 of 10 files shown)


vrurg:
With `throw_or_die` it is now much easier to replace a number of `nqp::die` with real Raku exceptions.

jnthn:
I think this probably belongs more under `Rakudo::Internals` or some such (looking at whats in `Metamodel::Configuration`, which I'd not noticed before now, perhaps the entire class belongs there).

> perl6_ops.c which is currently the only case where P6EX is used

`p6finddispatcher` will be gone once we're using the new dispatcher, which will in turn eliminate the P6EX usage in `perl6_ops.c`.

vrurg:
I don't agree that `throw_or_die` belongs to Raku land because currently it is available to the metamodel code at any time, including the early stages of core setting compilation. Encapsulated fallback to plain `nqp::die` until the actual exception classes are available is what makes it appealing in my eyes.

What could be done is `find_exception` method can be offloaded to Raku code in order to avoid duplication of the symbol lookup functionality. This way involves more overhead, but it shouldn't be critical because exceptions are expensive anyway and they normally happen only once.

jnthn:
Sorry for the confusion; I wasn't meaning that the code belongs in `CORE.setting`, but rather that things under the `Metamodel::` namespace are candidates for gradual standardization, and so things that are implementation-specific plumbing probably are better elsewhere.

vrurg:
Time after time I'm thinking about the metamodel standardization. It should become a problem-solving issue at some point because we'd need to decide what methods must be part of Raku specs, and what are Rakudo specific. 

`Rakudo::Internals` is unlikely candidate for standardization. But `Metamodel::Primitives` is certainly is.

vrurg:
I think this PR is ready. Tests and spectests are passing. With 2c4a006 I tried to demo how easy is it now to make metamodel more informative in error reporting.

PR #4358, opened 2021-05-07
Use "#!/usr/bin/env rakudo" instead of perl6 in module scripts

voegelas:
As far as I can see there's no need to put "#!/usr/bin/env perl6" any longer into scripts like cro and zef when these modules are built and installed with recent Rakudo versions. This pull request replaces perl6 with rakudo, perl6-m with rakudo-m etc. in src/core.c/CompUnit/Repository/Installation.pm6.

BTW, if scripts use #!/usr/bin/env the package tool rpmbuild adds automatic dependencies on the package coreutils, which provides /usr/bin/env, instead of rakudo. But that's another issue that I handle in the RPM spec file.

### DIFF
--- a/src/core.c/CompUnit/Repository/Installation.pm6
+++ b/src/core.c/CompUnit/Repository/Installation.pm6
@@ -38,19 +38,19 @@ class CompUnit::Repository::Installation does CompUnit::Repository::Locally does
     my $windows_wrapper = '@rem = \'--*-Perl-*--
 @echo off
 if "%OS%" == "Windows_NT" goto WinNT
-#perl# "%~dpn0" %1 %2 %3 %4 %5 %6 %7 %8 %9
-goto endofperl
+#raku# "%~dpn0" %1 %2 %3 %4 %5 %6 %7 %8 %9
+goto endofraku
 :WinNT
-#perl# "%~dpn0" %*
-if NOT "%COMSPEC%" == "%SystemRoot%\system32\cmd.exe" goto endofperl
-if %errorlevel% == 9009 echo You do not have Perl in your PATH.
+#raku# "%~dpn0" %*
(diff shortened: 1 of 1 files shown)


patrickbkr:
Is https://github.com/rakudo/rakudo/pull/3716 related?

voegelas:
> Is #3716 related?

Sort of. But I think that systems that haven't got /usr/bin/env and only provide /bin/env are non-standard even if no standard defines the exact location. I wouldn't add complex code that supports both /usr/bin/env and /bin/env. Better don't use the env command at all und use the configured prefix to set the interpreter to #!/usr/bin/rakudo, for example. That would also make package tools like rpmbuild and rpmlint happy.

But my pull request isn't about the use of the env command. It just replaces perl6 by rakudo.

patrickbkr:
Ah. Thanks for the explanation!

Looking good. Do you have permissions to merge this yourself?

patrickbkr (approved):

lizmat:
I think at this stage, where the status of the CompUnit modules is still a bit in question (is it implementation specific or not?), that `rakudo` is actually the proper thing to do.

jnthn:
I thought quite a lot of aspects of the `CompUnit::` stuff had made it into spectest by now?

lizmat:
Ah, indeed. Hmmm...

jnthn:
I did just take a glance through the tests, and I think `S11-repository/cur-candidates.t` is sort of covering something in this territory.

voegelas:
`raku` would require `raku-m`, `raku-j` and `raku-js` command aliases. That's why I replaced `perl` with `raku` in most places, but decided to use `rakudo` in the subst() calls. For example, when zef is installed you get the scripts zef, zef-m, zef-j and zef-js with corresponding shebang lines.

PR #4359, opened 2021-05-08
Remove special cases for hash constants for JVM

usev6:
The underlying problem has been fixed for the JVM backend with
https://github.com/Raku/nqp/commit/00408eea93.

Unfortunately, the special cases are probably still needed for the
JS backend. So I switched the logic from 'use hash constants only
for MoarVM' to 'use hash constants everywhere except for JS'.

### DIFF
--- a/src/core.c/CompUnit/RepositoryRegistry.pm6
+++ b/src/core.c/CompUnit/RepositoryRegistry.pm6
@@ -363,10 +363,10 @@ class CompUnit::RepositoryRegistry {
         }
     }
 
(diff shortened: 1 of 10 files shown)


usev6:
For the record: @pmurias stated in https://github.com/rakudo/rakudo/issues/2628#issuecomment-455823489, that using ```my constant nqp::hash``` posed a problem for the JS backend, too. I don't know if that is still the case.

lizmat (approved):

PR #4368, opened 2021-05-18
Add :throw named argument to run()

lizmat:
Basically, this will just run the same logic as sinking, causing a
failure to be thrown immediately, instead of silently continuing.
And this not producing any output with :out.

Also some style changes to run() to make it better readable

### DIFF
--- a/src/core.c/Proc.pm6
+++ b/src/core.c/Proc.pm6
@@ -237,11 +237,26 @@ my class Proc {
 }
 
 proto sub run(|) {*}
-multi sub run(*@args where .so, :$in = '-', :$out = '-', :$err = '-',
-        Bool :$bin, Bool :$chomp = True, Bool :$merge,
-        Str  :$enc, Str:D :$nl = "\n", :$cwd = $*CWD, :$env, :$arg0, :$win-verbatim-args = False) {
-    my $proc := Proc.new(:$in, :$out, :$err, :$bin, :$chomp, :$merge, :$enc, :$nl);
+multi sub run(*@args where .so,
+       :$in  = '-',
+       :$out = '-',
+       :$err = '-',
+  Bool :$bin,
+  Bool :$chomp = True,
+  Bool :$merge,
+  Str  :$enc,
+  Str  :$nl = "\n",
+       :$cwd = $*CWD,
+       :$env,
+       :$arg0,
+       :$win-verbatim-args = False,
+       :$throw,
+--> Proc:D) {
+    my $proc := Proc.new(
+      :$in, :$out, :$err, :$bin, :$chomp, :$merge, :$enc, :$nl
+    );
     $proc.spawn(@args, :$cwd, :$env, :$arg0, :$win-verbatim-args);
+    $proc.sink if $throw;
     $proc
 }
 


lizmat:
Meh.  Wrong # in commit

ugexe:
This seems like a bad idea to me. `:throw` isn't just making an exception happen without having to sink it yourself, its changing the blocking entirely. Such a chance would bring inconsistencies for what I see as a very minor benefit.

lizmat:
The intent of this is to allow for oneliners such as:

```
dd run(<foo bar>, :out).out.slurp;
```

which at the moment, silently just returns `""`.

Sadly, this *also* just returns the empty string silently:

```
my $p = run(<foo bar>, :out);
say $p.out.slurp;
```
I think there are no blocking issues here, as it would wait for the process to finish anyway.

In either case, a way to make it throw (aka `autodie` in Perl) feels to me to have a good thing to have in the toolchest.

ugexe:
> I think there are no blocking issues here, as it would wait for the process to finish anyway.

That example is too trivial -- work can be done between `run` and when the process gets sunk.

lizmat:
If you want work to be done between `run` and the process being sunk, then you shouldn't use `:throw`.  Maybe it should be called `:wait` instead then?  With the throwing as a side-effect?

lizmat:
Just discusses this with *jnthn*: it seems that it would be better to catch the case that the external process didn't actually start, and have `.out` and `.err` throw when that happens.  That would catch the original op's case and possibly many others, and would be automatic for all users, without needing to know about any additional parameters.

So closing this PR now, in lieu of a new one.

unclechu:
@lizmat by the way, is it okay to call `.sink` before anything is read from `out`? Is there some buffer limit for `out` that can be overfilled while the thread waits for `.sink`?

unclechu:
For now I’m writing something like this:
``` raku
with run «foo bar baz», :out { LEAVE {.sink}; .out.slurp(:close).chomp }
```
I’m not sure if `LEAVE` block is necessary.

lizmat:
You can call .sink, but then you would be blocking on the process actually starting.  Which may have performance impact.

unclechu:
@lizmat yes, I understand. I mean if I open `:out` and not reading anything yet before the child process is done waiting by `.sink`, where the contents for `.out` is stored that I’m going to read from **after** the child process finishes?

lizmat:
Somewhere in memory, which could be a serious issue in some situations.

jnthn:
For non-trivial process handling, I'd suggest using [`Proc::Async`](https://docs.raku.org/type/Proc::Async) instead, where things are rather more explicit. (`Proc`, and thus `run`, are wrappers around this that try to let one pretend things are synchronous, but IMO they lead to more questions than answers.)

unclechu:
@jnthn this is exactly about trivial blocking cases. The original issue raised from my one-liner: `run(«foo bar baz», :out).out.slurp(:close).chomp` which just silently returns empty string for whatever failure which I think is pretty confusing.

jnthn:
@unclechu I guess as soon as there's reading output from the process involved, I already consider it non-trivial. :-)

I fully agree `Proc` is confusing in this situation. In hindsight, having handle-like `out` and `err` things on `Proc` was a design mistake. Instead, it should probably have always run until exit, and only then returned an object that provides access to the entire data read from stdout/stderr along with the exit code. If the process could not be spawned at all, it would be an error to even ask for those (and probably it would need to be defused somehow anyway if there was a non-zero exit code). I wonder if we can get there from where we are...

unclechu:
@jnthn I think having a simple way to call child processes and capturing their output like they are just local functions is really nice to have, for some cases even crucial. I believe it satisfies UNIX-way principle and making scripting things a lot more pleasant. I like that you can decide whether to forward stdout/stderr or to capture it. If what you have in mind can be written in similar level of simplicity as `run(«foo bar baz», :out).out.slurp(:close).chomp` then I agree. But `Proc::Async` is out of scope of what I actually need in my particular case, it would transform that one-liner into few lines that is already too much for a simple capture of the child command output. There are of course applications for `Proc::Async` but there also are applications for blocking `Proc`.

Issue #4286, opened 2021-04-01
A DateTime object should output a real number for its posix method


tbrowder:
## The Problem

A DateTime object can be created from a fractional posix timestamp, but its *posix* method shows only the integral part of that timestamp.

## Expected Behavior

    my $t = DateTime.new(123.3);
    1970-01-01T00:02:03.300000Z
    say $t.posix
    123.3

## Actual Behavior

    say $t.posix
    123
    say $t.utc
    1970-01-01T00:02:03.300000Z

## Environment

Debian Buster
Raku release 2021.03

tbrowder:
Unfortunately such output would not always rountrip exactly because the input value isn't saved. But the output should be within acceptable error bounds (and checked with tests).

tbrowder:
After looking more at the temporal code in Raku, I see the fractional (Real) POSIX time information is currently available via use of the Instant method, and more complete documentation would help that be more obvious.

As one example of confusion, the Instant is said to be not based on any epoch, but it seems to be in fact based on the POSIX time epoch.

I am working on a docs PR to add more details of the situation.

Closing this issue as not needed.

tbrowder:
Note PR #4432 provides a 'posix-real' method which shows the posix value with fractional seconds.

tbrowder:
Reopening awaiting incorporation in version 6.e.

In the meantime, interested users can use module DateTime::Posix for that functionality.

PR #4371, opened 2021-05-18
Provide a method returning a real 'posix' value

tbrowder:
+ Add new multi method "DateTime.posix" which returns a real value
    - Add a required named option ":$real" to the new multi to select the new method
+ Satisfies closed issue #4286
+ Replaces draft PR #4350 (which has been closed)


lizmat:
Sorry for the delay: implemented as https://github.com/rakudo/rakudo/commit/1ffedd662b as the PR had bitrotted.

PR #4372, opened 2021-05-18
Make DateTime.posix return a fractional value in 6.e

lizmat:
This is really a proof-of-concept.  It basically saves the DateTime
class that is known at the beginning of parsing the 6.e core setting
as "DateTimeD", and then creates a DateTime class that inherits from
"DateTimeD".  Several notes to make about this:

- this exposes DateTimeD, it probably shouldn't.  But I don't see
  a way to do that at the moment, and still make the new DateTime
  class work in 6.e code.
- because DateTimeD is exposed, several tests in "make test" fail.
  This is easily fixable should we decide to make it ok to expose
  DateTimeD.
- Because the new DateTime is a subclass of the DateTimeD, attributes
  can only be reached with calling the accessors, or to use nqp::getattr.
  I've used the former for all but .second, as that currently always
  returns an Int apparently.  It would be nice if there were some
  HOW stuff to make DateTimeD trust DateTime, so that direct access
  to the attributes were possible.

### DIFF
--- a/src/core.e/core_prologue.pm6
+++ b/src/core.e/core_prologue.pm6
@@ -7,9 +7,7 @@ my constant CORE-SETTING-REV = 'e';
 
 my constant DateTimeD = DateTime;
 class DateTime is DateTimeD {
-    method posix(DateTime:D: $ignore-timezone? --> Numeric:D) {
-        return self.utc.posix if $.timezone && !$ignore-timezone;
-
+    multi method posix(DateTime:D: --> Real:D) {
         # algorithm from Claus Tøndering
         my int $a = (14 - $.month) div 12;
         my int $y = $.year + 4800 - $a;
@@ -19,6 +17,7 @@ class DateTime is DateTimeD {
         ($jd - 2440588) * 86400
           + $.hour      * 3600
           + $.minute    * 60
+          - $.timezone
           + nqp::getattr(self,DateTimeD,'$!second')
     }
 }
--- a/src/core.e/core_prologue.pm6
+++ b/src/core.e/core_prologue.pm6
@@ -1,7 +1,26 @@
 use nqp;
 
 # This constant must specify current CORE revision
-# Must preceede class declarations to allow correct recording of their respective language version.
+# Must precede class declarations to allow correct recording of their
+# respective language version.
 my constant CORE-SETTING-REV = 'e';
+
+my constant DateTimeD = DateTime;
+class DateTime is DateTimeD {
+    method posix(DateTime:D: $ignore-timezone? --> Numeric:D) {
+        return self.utc.posix if $.timezone && !$ignore-timezone;
+
+        # algorithm from Claus Tøndering
+        my int $a = (14 - $.month) div 12;
+        my int $y = $.year + 4800 - $a;
+        my int $m = $.month + 12 * $a - 3;
+        my int $jd = $.day + (153 * $m + 2) div 5 + 365 * $y
+            + $y div 4 - $y div 100 + $y div 400 - 32045;
+        ($jd - 2440588) * 86400
+          + $.hour      * 3600
+          + $.minute    * 60
+          + nqp::getattr(self,DateTimeD,'$!second')
+    }
+}
 
 # vim: expandtab shiftwidth=4


jnthn:
I somehow can't find it now, but I'm sure there was a `problem-solving` issue on the topic of changing method semantics over language versions, where I proposed that - with `new-disp` in place - we make it possible to have multi candidates filtered by the language version of the caller. I know it's frustrating to wait, but I'd prefer we provide the proposed 6.e behavior for `DateTime.posix` - which on the surface looks agreeable to me - once we have that mechanism available, which I think would lack all the shortcomings of the approach taken here:

* Wouldn't need a `DateTimeD`
* Thus wouldn't need mitigation in the tests
* Wouldn't have the encapsulation issues because the method goes into the same class

lizmat:
I understand all of those reasons.  But would it really hurt if we would implement it like this for now, and then replace it with the proper solution when new-disp lands?

niner:
> I understand all of those reasons.  But would it really hurt if we would
> implement it like this for now, and then replace it with the proper
> solution when new-disp lands?

It would get us all the backwards incompatibility headaches (like the need to 
deprecate something) for a feature that can be easily supplied by module.

jnthn:
> But would it really hurt if we would implement it like this for now, and then replace it with the proper solution when new-disp lands?

I guess I'm naturally inclined to avoid re-work or to warn people when I expect things to be a throwaway. The only particular risk I see is forgetting to clear up something after it when switching to doing it a different way.

> It would get us all the backwards incompatibility headaches (like the need to
> deprecate something) for a feature that can be easily supplied by module.

Is this an argument against the `DateTime.posix` change itself, or rather than the temporary solution being proposed? Since it would seem to require a `6.e.PREVIEW` to get at it, we don't really have a problem with deprecation.

niner:
> Is this an argument against the `DateTime.posix` change itself, or rather
> than the temporary solution being proposed? Since it would seem to require
> a `6.e.PREVIEW` to get at it, we don't really have a problem with
> deprecation.

Against a temporary solution.

vrurg:
> I somehow can't find it now, but I'm sure there was a `problem-solving` issue on the topic of changing method semantics over language versions, where I proposed that - with `new-disp` in place - we make it possible to have multi candidates filtered by the language version of the caller.

I would strongly disagree with this feature. Even though I was once considering ways to implement it. First of all, one of the right decisions taken long ago about Perl6 semantics was about taking the knowledge about the calling context away from the code objects. Implementing language revision dependent dispatch is basically not so different from having Perl's `wantarray` available. In a way, it's even worse because it would introduce a very confusing kind of side effects in cases where core objects created by, say, 6.e environment are passed over into 6.d/6.c code where they will behave differently! I mean, the last thing I'd expect is when my `DateTime` object results in different output depending on wether I print its stringification done my code, or when it is passed into, say, `print` where it's stringified in 6.c context.

Also, we must consider cases where objects or typeobjects are used by metamodel NQP code. There is no language revision context to rely upon. Hence, there is an ambiguity which doesn't have a solution.

What I do like with the approach taken by @lizmat in this PR is that by inheriting from 6.c version of `DateTime` the continuity of the class is preserved. I have recently stumbled upon an issue with `EVAL` which doesn't accept 6.e version of `PseudoStash` because 6.c version is expected. But it's only today I realized that `CORE::v6*` namespace is not only supported by 6.e `PseudoStash` but also provided by `World` at the compile time.

So, the right solution for both classes is:

```raku
my DateTime is CORE::v6c::DateTime {...}
```

This way we get no extra symbol. 

My bad I overlooked this possibility back then when the feature was introduced. Looks like I have a job to be done now. A core with 6.e `PseudoStash` declared this way has been successfully compiled a few minutes ago.

jnthn:
> Implementing language revision dependent dispatch is basically not so different from having Perl's wantarray available.

I'm not sure `wantarray` was so much judged as a bad thing by itself, it's just an impossible thing when one has multiple dispatch too, since the context is unknowable, and multiple dispatch was judged the far more useful thing to keep. I'm pretty sure `wantarray` survived in the early language design for quite a while, because I can hazily remember it going away, and I wasn't involved until the better part of a decade into the process.

By contrast, the language version is a statically known (that is, lexical) property of the callsite, and so the problems that make `wantarray` unreasonable in Raku don't apply. Also, somewhat amusingly, I think you were [the first of us to cast the stone of implicit lexical properties being taken into account in a dispatch](https://github.com/rakudo/rakudo/blob/master/src/vm/moar/spesh-plugins.nqp#L23). :-)

> my DateTime is CORE::v6c::DateTime {...}

Then any 6.e code cannot receive a `DateTime` object created in 6.d code because the type check fails. While *sometimes* such type incompatibility is a feature, it feels like a bit of a heavy hammer in this situation.

vrurg:
> By contrast, the language version is a statically known (that is, lexical) property of the callsite, and so the problems that make `wantarray` unreasonable in Raku don't apply. Also, somewhat amusingly, I think you were [the first of us to cast the stone of implicit lexical properties being taken into account in a dispatch](https://github.com/rakudo/rakudo/blob/master/src/vm/moar/spesh-plugins.nqp#L23). :-)

How on earth do you manage to remember such details??? :) As an excuse I can point out that it is a spesh which I'd rather consider as a part of the context itself, contrary to the method it finds. Alternative implementation would require a dynamic symbol.

> Then any 6.e code cannot receive a `DateTime` object created in 6.d code because the type check fails. While _sometimes_ such type incompatibility is a feature, it feels like a bit of a heavy hammer in this situation.

Probably not that "heavy". I consider `CORE::v6c::DateTime:D $date` notation safer than version-based dispatch. We can also provide coercion from 6.c into 6.e version of the class. And make this coercion easily available with a syntax sugar of a kind.

Aside of above, strict separation of class versions would allow one to have version-based dispatch and handle different behaviors explicitly whenever necessary:

```raku
multi method foo(CORE::v6c::DateTime:D $d) {
    # take care of the older behavior ...
}
multi method foo(CORE::v6e::DateTime:D $d) {
    # ... and the newer behavior
}
```

Not sure if there is a way to make this possible with context-dependent method.

Meanwhile, in #4373 I have implemented a proof of concept for 6.e class inheriting from 6.c. Done it for `PseudoStash` because this is where the difference is big enough for the separation to make sense.

lizmat:
"where I proposed that - with new-disp in place - we make it possible to have multi candidates filtered by the language version of the caller."

@jnthn I assume this is not yet in place  :-)  Any pointers how someone like me, could be adding that?

lizmat:
Closing this as new-disp has landed and this PR is now really outdated.

PR #4373, opened 2021-05-20
Make 6.e PseudoStash a child of 6.c PseudoStash

vrurg:
This PR should fix a problem with `EVAL` which only takes 6.c `PseudoStash` as its `:context` argument.

Also add support for `:implementation-detail` argument, akin to how 0c36feb07d added it to 6.c.

### DIFF
--- a/src/Perl6/World.nqp
+++ b/src/Perl6/World.nqp
@@ -5127,6 +5127,12 @@ class Perl6::World is HLL::World {
                 QAST::SVal.new( :value('GLOBAL') ) );
         }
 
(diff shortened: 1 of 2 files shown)


vrurg:
For now this draft is a proof of concept. Complete implementation requires `:implementation-detail` argument support for methods like `keys`, `pairs`, etc. See 6.c version of `PseudoStash`.

Also see #4372 for related discussion.

vrurg:
Not a draft anymore, ready to merge.

PR #4380, opened 2021-05-28
Tweak worker addition for many small tasks

jnthn:
Today we add additional workers to the thread pool based upon lack of
progress. This works well when work items are relatively large, and so
hold on to a worker for a while. For applications that schedule lots of
small pieces of work that are well below the threshold at which the
supervisor runs, however, some work is always completed, and so the
condition for adding further threads to the pool is not met. This could
be observed by hitting a Cro HTTP server that does relatively little
work to produce the response with a lot of concurrent requests, and
seeing nearly no workers being added, and CPU usage staying very low as
a result.

With this change, we look at the amount of outstanding work also. If
there is more work outstanding than workers available, we add an extra
worker. This has been observed to lead to additional worker creation in
a Cro HTTP server under load, which in turn improves request rates.

### DIFF
--- a/src/core.c/ThreadPoolScheduler.pm6
+++ b/src/core.c/ThreadPoolScheduler.pm6
@@ -736,11 +736,22 @@ my class ThreadPoolScheduler does Scheduler {
             }
         }
 
-        # If we didn't complete anything, then consider adding more threads.
+        # Consider adding more worker threads when:
+        # 1. We didn't complete anything since the last supervision. This is
+        #    likely because some long-running tasks are holding onto the
+        #    workers. We should at least think about adding more (and the
+        #    code below will try to determine if that's beneficial).
+        # 2. The number of tasks in the queue is greater than the number of
+        #    workers. Such a situation suggests we are under-resourced, and
+        #    so liable to fall behind. Consider it like checkout lanes at a
+        #    supermarket: if 20 people are queueing and there are only 2 open
+        #    checkout lanes, it makes sense to open more, but if there are 20
+        #    people waiting and 30 open checkout lanes, there's little to be
+        #    won by opening another one at this point.
         my int $total-workers = nqp::elems($!general-workers)
           + nqp::elems($!timer-workers)
           + nqp::elems($!affinity-workers);
-        if $total-completed == 0 {
+        if $total-completed == 0 || nqp::elems(queue) > nqp::elems(worker-list) {
             if $total-workers < $!max_threads {
                 # There's something in the queue and we haven't completed it.
                 # If we are still below the CPU core count, just add a worker.


lizmat (approved):

niner (approved):
Oh that explains a lot! I guess I'm not the only one who wondered why a simple (or even not so simple) test case wouldn't scale as well as it should.

I do wonder about using the number of workers as the threshold though. On the first look it seems to be a logical choice and intuitively make sense. But on second though it seems more like just an arbitrary number. If I have as an extreme example just one worker and the queue holds 2 tiny tasks, starting up another worker probably won't be worth it. If I've got 30 workers falling behind and there are already 20 tasks in the queue, waiting for another 10 seems like a bit of a waste.

Or maybe I'm overthinking this or just overlooking the obvious connection. In any case this change will improve quite a few use cases and we can always fine tune later on.

jnthn:
It's certainly a simplistic heuristic, and so will certainly get things wrong (although I'd argue the addition of it means we get more things right). I picked it because:

* The principle behind it is easy to state: if there's more than one outstanding task for each worker, it suggests we're under-resourced or liable to fall behind
* It's somewhat self-throttling: the more workers we have, the more reluctant it is to create the next one. There's only one addition per supervision, meaning that a pile of small tasks landing at once (especially around startup) will at least somewhat be smoothed out by that, rather than resulting in a burst of worker creations.
* It's really cheap to implement
* I saw it have a useful effect in a non-trivial benchmark (the Cro application itself is trivial, but the HTTP processing pipeline is certainly a lot beyond being a microbenchmark)

That said, how it behaves as it moves from/towards limits is interesting indeed.

> just one worker and the queue holds 2 tiny tasks, starting up another worker probably won't be worth it

We could `max` the queue length with something to avoid this. It just adds a little latency in scaling up if we really are under-provisioned, but it's a very small increase in complexity to do this. Of course, that makes us pick a magic number.

> if I've got 30 workers falling behind and there are already 20 tasks in the queue, waiting for another 10 seems like a bit of a waste

The thinking was that if there's 30 workers and 20 tasks in the queue that's less than one job for each worker to pick up as its next task, so we're in a kind of "steady state" - that is, we're at this point keeping up. If by the next supervision there were already 40 tasks, that'd suggest we'd benefit from another worker.

niner:
> I picked it because:
> 
> * The principle behind it is easy to state: if there's more than one
> outstanding task for each worker, it suggests we're under-resourced or
> liable to fall behind * It's somewhat self-throttling: the more workers we
> have, the more reluctant it is to create the next one. There's only one
> addition per supervision, meaning that a pile of small tasks landing at
> once (especially around startup) will at least somewhat be smoothed out by
> that, rather than resulting in a burst of worker creations. * It's really
> cheap to implement
> * I saw it have a useful effect in a non-trivial benchmark (the Cro
> application itself is trivial, but the HTTP processing pipeline is
> certainly a lot beyond being a microbenchmark)

And now we've got this reasoning documented :) Can you add it to the commit 
message please?

> > just one worker and the queue holds 2 tiny tasks, starting up another
> > worker probably won't be worth it
> We could `max` the queue length with something to avoid this. It just adds a
> little latency in scaling up if we really are under-provisioned, but it's a
> very small increase in complexity to do this. Of course, that makes us pick
> a magic number.
> > if I've got 30 workers falling behind and there are already 20 tasks in
> > the queue, waiting for another 10 seems like a bit of a waste
> The thinking was that if there's 30 workers and 20 tasks in the queue that's
> less than one job for each worker to pick up as its next task, so we're in
> a kind of "steady state" - that is, we're at this point keeping up. If by
> the next supervision there were already 40 tasks, that'd suggest we'd
> benefit from another worker.

Actually, what we're really trying to determine is: "are we falling behind"? 
Which can be rephrased as "is the queue growing"? To answer that question, we 
really want to compare the queue length with the length at the previous 
supervisor run. Or maybe something like the maximum queue length of the 
previous n supervisor runs to get a more stable base for comparison. So we 
don't increase the pool of 23 workers just because the queue regressed from 8 
to 9 tasks.

lizmat:
> So we don't increase the pool of 23 workers just because the queue regressed from 8 to 9 tasks.

But that wouldn't happen, because 23 > 9 ?   It would only happen if the queue regressed from 23 to 24 outstanding tasks?  But then you could argue, it *does* make sense as there are more tasks than workers ?

Or am I  missing something?

niner:
> But that wouldn't happen, because 23 > 9 ?   It would only happen if the
> queue regressed from 23 to 24 outstanding tasks?  But then you could argue,
> it *does* make sense as there are more tasks than workers ?

I meant if we repace the `elems(queue) > elems(workers)` check with 
`elems(queue) > previous-queue-elems`

niner (approved):

jnthn:
I'll merge this, given there's approval. We also have some suggestions for tweaks and an alternative strategy, but so far as I can tell this makes things better for some use cases, and nobody pointed out any that it will make things worse for.

jnthn:
> I meant if we replace the `elems(queue) > elems(workers)` check with
> `elems(queue) > previous-queue-elems`

BTW, I guess my hesitancy on this is that it doesn't have the natural "brake" on worker creation that exists in the current solution, and my feeling is that such a brake is a good thing.

PR #4385, opened 2021-05-31
RFC for a .pick(**) candidate

lizmat:
Like .pick(*), this would exhaust all possible values.  But instead
of stopping providing values, continues with a fresh set again.

The advantage over just using .roll(*) is that you're guaranteed a
more even distribution of values on shorter runs exceeding the
original pool size.

For now this is only implemented for `List`, to test the waters.

### DIFF
--- a/src/core.c/List.pm6
+++ b/src/core.c/List.pm6
@@ -992,6 +992,9 @@ my class List does Iterable does Positional { # declared in BOOTSTRAP
         }
         method is-deterministic(--> False) { }
     }
+    multi method pick(List:D: HyperWhatever) {
+        self.pick(*).Slip xx *
+    }
     multi method pick(List:D: $number is copy) {
         return self.fail-iterator-cannot-be-lazy('.pick from') if self.is-lazy;
         my Int $elems = self.elems;


b2gills:
I wonder what `.roll(**)` would hypothetically do?

(If there is truly no sensible thing for `.roll(**)` then I have no arguments against `.pick(**)`.)

lizmat:
Well, conceptually `.roll(**)` would repeat the infinite sequence of random values an infinite number of times.  But as is my understanding, `Inf * Inf` is still `Inf`, so there would be no semantic difference.

In any case, I would not want `.roll(**)` to actually work: people have trouble enough understanding the difference between `.roll` and `.pick`, so I don't want people to do a `.roll(**)` and expecting the semantics of `.pick(**)`.  Not implementing a `.roll(HyperWhatever)` candidate appears the right way to do that to me.  If this turns out to be an issue, then maybe create a `Mu.roll(HyperWhatever)` candidate that would throw a more specific error.

b2gills:
My point is that if `.roll(**)` is nonsensical, then there is nothing for `.pick(**)` to work similar to.  
So then it is free to do its own thing.

lizmat:
Implemented with  https://github.com/rakudo/rakudo/commit/cd4a72b6a7 so closing this PR now.

Issue #4370, opened 2021-05-18
IO::Socket::Async is not closing connection after timeout on OSX


Zer0-Tolerance:
## The Problem
IO::Socket::Async is not closing TCP connection after several SYN_SENT and the connect timeout on OSX , this is also the same behavior with IO::Socket::INET.
It's not even possible to force close by using .close method as the promise is broken. 
## Expected Behavior
Connection is closed and file handle is released.  
## Actual Behavior
Connection is flagged as 'CLOSED' and remains forever ... 

## Steps to Reproduce
We need to connect to a server not responding 
`my $r=await IO::Socket::Async.connect('127.0.0.1',12345)`
wait for timeout 80 sec on OSX
`lsof -Pni |ack  raku`
result: 
`rakudo    88573 user   19u  IPv4 0xa6242a87c7115feb      0t0  TCP 127.0.0.1:64833->127.0
.0.1:12345 (CLOSED)`
## Environment
* Operating system: `Darwin 19.6.0 Darwin Kernel Version 19.6.0: Tue Jan 12 22:13:05 PST 2021; root:xnu-6153.141.16~1/RELEASE_X86_64 x86_64`
* Compiler version (`perl6 -v` or `raku -v`): 
```
Welcome to Rakudo(tm) v2021.04.
Implementing the Raku(tm) programming language v6.d.
Built on MoarVM version 2021.04.
```

Zer0-Tolerance:
Hi There,
Quick update: same behavior with latest version 2021.05.

Kaiepi:
I think `IO::Socket::Async`'s failure to close the socket and `IO::Socket::INET`'s are two separate issues:

- In MoarVM, in the event an error occurs during the invocation of the `asyncconnect` op that backs `IO::Socket::Async.connect`, the socket isn't closed when an [error is emitted asynchronously](https://github.com/MoarVM/MoarVM/blob/master/src/io/asyncsocket.c#L554..L567), unlike when it occurs [immediately](https://github.com/MoarVM/MoarVM/blob/master/src/io/asyncsocket.c#L609..L610).
- In the case of `IO::Socket::INET`, the socket isn't closed because `IO::Socket::INET.new` doesn't do anything if its `connect` or `bindsock` op calls throw. I don't think these ops should be concerned with closing the socket on error like `asyncconnect` does; `CATCH`ing from Rakudo instead keeps these ops simple.

Zer0-Tolerance:
[...]

PR #4391, opened 2021-06-03
Clean up sockets created by IO::Socket::INET.new on error

Kaiepi:
Related to #4370.

### DIFF
--- a/src/core.c/IO/Socket/INET.pm6
+++ b/src/core.c/IO/Socket/INET.pm6
@@ -57,14 +57,13 @@ my class IO::Socket::INET does IO::Socket {
             split-host-port :host($localhost), :port($localport), :$family
         orelse fail $_) unless $family == PF_UNIX;
 
-        #TODO: Learn what protocols map to which socket types and then determine which is needed.
         self.bless(
             :$localhost,
             :$localport,
             :$family,
             :listening($listen),
             |%rest,
-        )!initialize()
+        )!initialize(nqp::socket(0))
     }
 
     # Open new connection to socket on $host:$port
@@ -81,13 +80,12 @@ my class IO::Socket::INET does IO::Socket {
             :$family,
         ) unless $family == PF_UNIX;
 
-        # TODO: Learn what protocols map to which socket types and then determine which is needed.
         self.bless(
             :$host,
             :$port,
             :$family,
             |%rest,
-        )!initialize()
+        )!initialize(nqp::socket(10))
     }
 
     # Fail if no valid parameters are passed
@@ -96,8 +94,8 @@ my class IO::Socket::INET does IO::Socket {
             ~ "Invalid arguments to .new?";
     }
 
-    method !initialize() {
-        my $PIO := nqp::socket($!listening ?? 10 !! 0);
+    method !initialize(Mu $PIO is raw) {
+        CATCH { nqp::closefh($PIO) }
 
         # Quoting perl's SIO::INET:
         # If Listen is defined then a listen socket is created, else if the socket type,
--- a/src/core.c/IO/Socket/INET.pm6
+++ b/src/core.c/IO/Socket/INET.pm6
@@ -63,7 +63,7 @@ my class IO::Socket::INET does IO::Socket {
             :$family,
             :listening($listen),
             |%rest,
-        )!initialize(nqp::socket(0))
+        )!initialize(nqp::socket(10))
     }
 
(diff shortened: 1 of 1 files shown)


lizmat (approved):

lizmat (approved):

Zer0-Tolerance:
Hi , any idea when this will be merged ?

vrurg:
@Kaiepi any reason for this to wait?

Kaiepi:
I need to double check that the JVM copes with this OK. Past that, no.

Kaiepi:
Good to go.

vrurg:
@Kaiepi thank you!

Zer0-Tolerance:
hi all, 
Just check the behavior in 2021.07 and Raku is still not closing connection after timeout. Can you confirm on your side ? Pretty annoying because it's seems to prevent any kind of TCP Scan of more than 16k IP adresses.

Kaiepi:
@Zer0-Tolerance Are they synchronous sockets? `IO::Socket::INET`'s sockets are getting closed on timeout for me locally, but I'm on an OpenBSD system, and don't have an OS X one to test with offhand.

lizmat:
Is there a golf that I can use to check on MacOS ?

Kaiepi:
I can't write a straight golf for this, but OS X should be BSD-ish enough for the strategy I used locally to apply:
1. Run:
```raku
shell "ktrace -p $*PID";
say try { IO::Socket::INET.connect: '127.0.0.1', 12345 } // $!;
sleep;
```
2. An error should pop up eventually, and should be `ETIMEOUT`-related. If not, `docs.raku.org` seems apt to time out on the same port.
3. In the same directory, run:
```sh
$ kdump | less
```
4. There may or may not be multiple `socket(2)` calls depending on the host. `IO::Socket::INET`'s socket should be the last of these, followed closely by a `connect(2)` call. Note the address given and see if a call to `close(2)` for this socket appears at any point afterwards, in which case don't end the script yet.
5. If closed, `$ netstat -p tcp` should not include any sockets with the connection's address.

lizmat:
````
$ ktrace -p 7987
ktrace: unknown command `-p'
usage:	ktrace info
	ktrace trace [-AnNrsStu] [--remotetime] [-T <timeout>] [-f <filter-desc>] [-b <buffer-size-mb>] [-R <raw file> | -E]  [-C <code file> [...]] [-x <process name|pid> [...] | -p <process name|pid> [...]] [-F] [-c <command> [...]]
	ktrace dump [-E] [-l <compression-level>] [-T <timeout>] [-f <filter-desc>] [-b <buffer-size-mb>] [-p <pid-or-process-name>] [<path>] [-c <command> [...]]
	ktrace init -b <buffer-size-mb> | -n <n-events>
	ktrace setopt [-x <pid-or-process-name> | -p <pid-or-process-name>] [-w] [-f <filter-desc>]
	ktrace enable
	ktrace disable
	ktrace decode <debugid> [<debugid> ...]
	ktrace remove
	ktrace reset
	ktrace emit <debugid> [<arg1> [<arg2> [<arg3> [<arg4>]]]]
	ktrace symbolicate <path>
	ktrace stacks [-R <raw file> | -E]  [-C <code file> [...]] [-x <process name|pid> [...] | -p <process name|pid> [...]] [-F]
	ktrace compress [-l fast|balanced|small] <path>
		fast: fast speed, small compression ratio
		balanced: balanced speed and compression ratio
		small: slow speed, high compression ratio (default)
	ktrace artrace [-nr] [-t <time>] [-i <interval>] [-o <filename>] [-b <buffer-size-mb>] [-f <filter-desc>] [-F <filter-desc>] [-p <pid-or-process-name>] [--remote[=<remote-device>]] [--type=full|profile|lite|morelite] [--kperf <sample-desc>] [-c <command> [...]]

	-R <raw file> : read events from the trace file at path <raw file>
	-E : use the existing trace session, instead of setting up a new one
	-C <code file> : provide codes files that name the events
	-x <process name|pid> : ignore the events from processes specified
	-p <process name|pid> : show only events from processes specified
	-F : print the filter description used to configure ktrace
````

Zer0-Tolerance:
> @Zer0-Tolerance Are they synchronous sockets? `IO::Socket::INET`'s sockets are getting closed on timeout for me locally, but I'm on an OpenBSD system, and don't have an OS X one to test with offhand.

Yes async socket from IO::Socket::Async::SSL

PR #4393, opened 2021-06-07
Speed up setup of NativeCall subs and reduce memory usage

niner:
When replacing a native sub's stub body with a generated one, we need to remove
the scheduled fixups, lest they undo our replacement. The easiest way to do
this is to just run the compiler thunk, which takes care of this after dynamic
compilation. This however is a very expensive operation as it not only compiles
that body but has to create setup code for the lexical scope. This makes
compiling a computation unit containing native subs scale O(n^2) with the
number of native subs in the lexical scope.

We can avoid this by doing a more targeted removal of the stub code.
This speeds up compiling a comp unit containing some 3K native subs from almost
5 minutes to just 2 1/2 minutes and reduces memory usage from 16 GiB to 8 GiB.

### DIFF
--- a/src/Perl6/World.nqp
+++ b/src/Perl6/World.nqp
@@ -2682,10 +2682,10 @@ class Perl6::World is HLL::World {
                 # do dynamic compilation.
                 %!code_object_fixup_list{$cuid} := $fixups;
             }
-
-            # Stash the QAST block in the comp stuff.
-            @compstuff[0] := $code_past;
         }
+
+        # Stash the QAST block in the comp stuff.
+        @compstuff[0] := $code_past;
 
         # If this is a dispatcher, install dispatchee list that we can
         # add the candidates too.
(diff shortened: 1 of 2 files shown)


src/Perl6/World.nqp
```diff
@@ -2982,6 +2982,26 @@ class Perl6::World is HLL::World {
...
+
+            my %sub_id_to_sc_idx := self.context().sub_id_to_sc_idx();
+            my $code_ref := nqp::getattr($code, $code_type, '$!do');
+            if nqp::existskey(%sub_id_to_sc_idx, $subid) {
+                nqp::markcodestatic($code_ref); # maybe $!do instead
+                self.update_root_code_ref(%sub_id_to_sc_idx{$subid}, $code_ref);
+            }
+            if nqp::existskey(%!code_object_fixup_list, $subid) {
+                my $fixups := %!code_object_fixup_list{$subid};
+                $fixups.pop() while $fixups.list;
```
MasterDuke17, line 3000: We need to actually pop the existing list? We can't just assign a new empty list?

src/Perl6/World.nqp
```diff
@@ -2982,6 +2982,26 @@ class Perl6::World is HLL::World {
...
+        my @compstuff := nqp::getattr($code, $code_type, q<@!compstuff>);
+        unless nqp::isnull(@compstuff) {
+            my $subid := @compstuff[0].cuid;
+
+            nqp::bindattr($code, $code_type, '@!compstuff', nqp::null());
+
+            my %sub_id_to_sc_idx := self.context().sub_id_to_sc_idx();
+            my $code_ref := nqp::getattr($code, $code_type, '$!do');
+            if nqp::existskey(%sub_id_to_sc_idx, $subid) {
+                nqp::markcodestatic($code_ref); # maybe $!do instead
```
MasterDuke17, line 2995: What does this comment mean? `$code_ref` is already `$!do`, isn't it?

PR #4395, opened 2021-06-08
Micro-optimize levenshtein sub

MasterDuke17:
By adding types, using the right operator for the types it's operating
on, removing explicit returns, and manually contructing a string instead
of using nqp::join.

For this completely made up benchmark:
```raku
my int $i := 0;
my $d;
while $i++ < 50_000 {
    $d := levenshtein(~$i, ~($i + 1));
}
say($d);
```
the original runtime was ~4.0s, with 812,893 deoptimizations. This
reduces those numbers to ~2.6s, with 0 deoptimizations. Additionally
there are fewer allocations overall, but I don't have the exact numbers
for those.

### DIFF
--- a/src/Perl6/World.nqp
+++ b/src/Perl6/World.nqp
@@ -60,43 +60,40 @@ nqp::bindcurhllsym('nqplist', &nqplist);
 # when deriving from an unknown/typo'd class.
 sub levenshtein($a, $b) {
     my %memo;
-    my $alen := nqp::chars($a);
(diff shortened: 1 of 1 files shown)


MasterDuke17:
For a real-life example, `time raku -e 'sat "hi"'` takes ~0.103s on master, but ~0.095s with this change.

lizmat (approved):
It's nice to make it more efficient.  But I guess it won't make a difference if any normal execution?

PR #4396, opened 2021-06-10
Compile-time micro-optimization

vrurg:
Move a call to `$*W.find_single_symbol` outside of a `for` loop.

### DIFF
--- a/src/Perl6/Actions.nqp
+++ b/src/Perl6/Actions.nqp
@@ -1549,12 +1549,13 @@ class Perl6::Actions is HLL::Actions does STDActions {
 
     # Produces a LoL from a semicolon list
     method semilist($/) {
+        my $Nil := $*W.find_single_symbol('Nil');
         if $<statement> -> $statements {
             my $past := QAST::Stmts.new( :node($/) );
             if nqp::elems($statements) > 1 {
                 my $l := QAST::Op.new( :name('&infix:<,>'), :op('call') );
                 for $statements {
-                    my $sast := $_.ast || QAST::WVal.new( :value($*W.find_single_symbol('Nil')) );
+                    my $sast := $_.ast || QAST::WVal.new( :value($Nil) );
                     $l.push(wanted($sast, 'semilist'));
                 }
                 $past.push($l);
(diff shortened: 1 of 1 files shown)


niner (changes requested):
Isn't it also possible that we do not need to look up Nil at all? Indeed I think that's actually the common case.

lizmat (approved):

lizmat:
@niner possbly, but one could also argue that the lookup for `Nil` could be moved even further out, possibly to the outer scope of `Actions.nqp`.  And replace other lookups for `Nil` as well?

MasterDuke17:
A long time ago I tried creating constants for some of the most popular looked up symbols, but I don't remember why it didn't happen. However, I do have a mix of stats. https://gist.github.com/MasterDuke17/0cb912b3d0cb46e1e862df8fa8ccf0cc has stats for what name is looked up in `find_single_symbol()` during CORE.c compilation. Below is the entry in the 'Routines' tab from a --profile-compile of the CORE.c compilation taken six months ago (it silently fails now). Sorted by exclusive time, `find_single_symbol()` is the 9th routine.

&nbsp; | 271123 | find_single_symbol gen/moar/World.nqp:4991 | 99.8% jit605,504 | 1,142.46ms / 1,696.93ms1.89µs /  2.80µs per entry
-- | -- | -- | -- | --

niner:
On Donnerstag, 10. Juni 2021 10:16:09 CEST Elizabeth Mattijsen wrote:
> @niner possbly, but one could also argue that the lookup for `Nil` could be
> moved even further out, possibly to the outer scope of `Actions.nqp`.  And
> replace other lookups for `Nil` as well?

Unfortunately no, since Nil could have been overridden in lexical scope, so we 
must do that lookup at the point we're currently compiling.

lizmat:
@niner Well, some values have special meanings in Raku: `Mu` and `Nil` and `IterationEnd` come to mind.  Lexically overriding those feels like a severe case of DIHWIDT.  And thus these cases could be considered special?

In any case, this will all be moot when RakuAST lands.

vrurg:
> Isn't it also possible that we do not need to look up Nil at all? Indeed I think that's actually the common case.

Aside of the remaining discussion, do you mean replacing it with a constant? `nqp::null`?

I'm having a hard morning at work, possibly lagging behind a bit. :)

niner:
No, just that looking at that function, Nil is only used in two places and only conditionally: `$_.ast || Nil` and `$ast || Nil`. So if we have an AST there, we don't need Nil at all. I think the case where we do need that Nil is for cases like `(1;2;;3)` i.e. two consecutive semicolons. That doesn't sound very common to me.

vrurg:
@niner makes sense. Thanks, I'm closing it.

PR #4397, opened 2021-06-11
Restore Pod6 E<> html entity support

raydiak:
See https://logs.liz.nl/raku/2021-06-11.html#01:38 . This restores lost functionality for pod escapes like `E<mdash>` which use the names of HTML entities. Untested on JVM. Please review carefully as I haven't contributed in a very long time and this restores code from 2014.


src/Perl6/Pod.nqp
```diff
@@ -1665,6 +1665,261 @@ class Perl6::Pod {
 
     } # end sub table
 
+    # all-lowercase HTML5 character entities, derived from
+    # https://github.com/w3c/html/blob/master/entities.json
+    my %entities;
+    # NQP doesn't support assignment of entire hashes :(
+    # so we have to iterate
```
vrurg, line 1672: We actually don't have to. `my %entitities := nqp::hash('aacute',[225],'aacute',[225], ... );`

vrurg (changes requested):
Looks ok except for the hash assignment.

src/Perl6/Pod.nqp
lizmat: Let's start with reverting the offending commit, and then look at improvements later?

vrurg: Ok. I didn't notice it's a revert at the first glance.

vrurg (approved):

vrurg:
BTW, it seemingly breaks JVM again. But I'd rather for JVM to be fixed than for pulling out documented functionality.

raydiak:
> BTW, it seemingly breaks JVM again. But I'd rather for JVM to be fixed than for pulling out documented functionality.

I saw that. @usev6 volunteered to take a look at the JVM problem hopefully this weekend. The IRC discussion is at https://logs.liz.nl/raku-dev/2021-06-11.html#09:19

usev6:
As far as I can see, Java's general limit of 64K bytes for method size is hit.
I tried to use nqp::hash(), but that exploded with the same errror.

I did some experiments and just creating a nqp::hash with enough entries and running that code with ```nqp-j``` eventually blows up with the following error:
```
java.lang.ClassFormatError: Invalid method Code length 65536 in class file D1B41BFF11C2F55874301B5C73AE87E759349E91
  in jvm (gen/jvm/stage2/NQPHLL.nqp:101)
  in execute_stage (gen/jvm/stage2/NQPHLL.nqp:1481)
  in run (gen/jvm/stage2/NQPHLL.nqp:1519)
  in <anon> (gen/jvm/stage2/NQPHLL.nqp:1522)
  in compile (gen/jvm/stage2/NQPHLL.nqp:1514)
  in eval (gen/jvm/stage2/NQPHLL.nqp:1186)
  in evalfiles (gen/jvm/stage2/NQPHLL.nqp:1460)
  in command_eval (gen/jvm/stage2/NQPHLL.nqp:1350)
  in command_line (gen/jvm/stage2/NQPHLL.nqp:1309)
  in MAIN (gen/jvm/stage2/NQP.nqp:4183)
  in <mainline> (gen/jvm/stage2/NQP.nqp:4179)
  in <anon> (gen/jvm/stage2/NQP.nqp)
```

Adding more entries leads to the error we are seeing in the CI (```java.lang.NoClassDefFoundError: org/objectweb/asm/tree/MethodNode```).

Interestingly, the same code does not blow up when it is run with ```rakudo-j```. For the sake of correctness: I've added a ```use nqp;``` to the top, so my test code looks like this:
```
use nqp;
my $foo := nqp::hash(                                                                                                                                                                                                                          
    'lharul',
    [10602],                                                                                             
    'lhblk',
    [9604],
    [... many more entries ...]
);
```

Anyway, I'd say the code in  ```src/Perl6/Pod.nqp``` should be put between ```#?if !jvm``` and ```#?endif``` markers and in ```src/Perl6/Actions.nqp``` the old code should be kept for the JVM backend (again with markers for the preprocessor). (I could do that myself, once this is merged.)

usev6:
I've created an issue for nqp to at least document the problem with the JVM backend: https://github.com/Raku/nqp/issues/727
Maybe the examples over there make the problem a bit clearer.

raydiak:
The old implementation was restored for JVM only, so it builds. Please let me know if I can do anything else to get this merged.

lizmat:
Thanks!

Issue #4394, opened 2021-06-07
Make test fails on OSX Apple M1 architecture


Zer0-Tolerance:
## The Problem
I've built from source the latest Raku 2021.05 but the make test fails to pass even with libffi installed using brew. This is related to #4364. 
```
+++ Checking for moar NQP version
'/Users/xxx/perl5/perlbrew/perls/perl-5.24.4/bin/perl' -I'/Users/xxx/install/raku/rakudo-2021.05/tools/lib' -I'/Users/xxx/install/raku/rakudo-2021.05/3rdparty/nqp-configure/lib' t/harness5 --moar t/01-sanity t/02-rakudo t/04-nativecall t/05-messages t/06-telemetry t/07-pod-to-text t/08-performance t/10-qast t/09-moar
t/01-sanity/03-op-logic.t ....................................... ok
t/01-sanity/02-op-math.t ........................................ ok
t/01-sanity/04-op-cmp.t ......................................... ok
t/01-sanity/01-literals.t ....................................... ok
t/01-sanity/05-var.t ............................................ ok
t/01-sanity/06-op-inplace.t ..................................... ok
t/01-sanity/09-end-blocks.t ..................................... ok
t/01-sanity/12-counter.t ........................................ ok
t/01-sanity/11-tap.t ............................................ ok
t/01-sanity/10-regex.t .......................................... ok
t/01-sanity/08-var-array.t ...................................... ok
t/01-sanity/13-equal.t .......................................... ok
t/01-sanity/07-op-string.t ...................................... ok
t/01-sanity/15-sub.t ............................................ ok
t/01-sanity/17-isa.t ............................................ ok
t/01-sanity/16-eqv.t ............................................ ok
t/01-sanity/19-say.t ............................................ ok
t/01-sanity/21-try.t ............................................ ok
t/01-sanity/20-defined.t ........................................ ok
t/01-sanity/18-simple-multisubs.t ............................... ok
t/01-sanity/14-if.t ............................................. ok
t/01-sanity/98-test-deprecated.t ................................ ok
t/01-sanity/53-transpose.t ...................................... ok
t/02-rakudo/01-is_approx.t ...................................... ok
t/01-sanity/56-use-isms.t ....................................... ok
t/01-sanity/55-use-trace.t ...................................... ok
t/01-sanity/99-test-basic.t ..................................... ok
t/02-rakudo/03-corekeys-6d.t .................................... ok
t/02-rakudo/03-corekeys.t ....................................... ok
[...]

PR #4399, opened 2021-06-12
Fix for #4394

tony-o:
The test is already skipped on OSX, the M1 architecture is now being identified as `macos` rather than `macosx`, this update amends the skip to include both.

### DIFF
--- a/t/04-nativecall/02-simple-args.t
+++ b/t/04-nativecall/02-simple-args.t
@@ -48,7 +48,7 @@ is TakeInt64(0xFFFFFFFFFF), 9, 'passed int64 0xFFFFFFFFFF';
 sub TakeUint8(uint8) returns int32 is native('./02-simple-args') { * }
 sub TakeUint16(uint16) returns int32 is native('./02-simple-args') { * }
 sub TakeUint32(uint32) returns int32 is native('./02-simple-args') { * }
-if $*DISTRO.name eq 'macosx' {
+if $*DISTRO.name ~~ 'macosx'|'macos' {
     #
     # For some reason, on OS X with clang, the following test fails with -O3
     # specified.  One can only assume this is some weird compiler issue (tested


t/04-nativecall/02-simple-args.c
```diff
@@ -78,16 +78,16 @@ DLLEXPORT int TakeInt64(int64_t x)
...
+    if ((0xFF&(unsigned int)x) == 0xFE)
         return 10;
     return 0;
 }
 
-DLLEXPORT int TakeUint16(unsigned short x)
+DLLEXPORT int TakeUint16(char x)
 {
-    if (x == 0xFFFE)
+    if ((0xFFFF&(unsigned int)x) == 0xFFFE)
```
niner, line 90: These changes do not make sense to me. We explicitly want to test functions that take unsinged characters and signed shorts. Why would we change both to char or int?

niner (changes requested):

t/04-nativecall/02-simple-args.c
tony-o: @niner the compiler is optimizing out the actual LSB part of the check in clang on arm64.  `w0, w8` are 32bits and actually contain `0xFFFF - 1`.  This test fails even with `-O0` on apple's M1.  Using a 4 byte type and checking explicitly more closely matches what is actually happening when the C is compiled.

The disassembler shows the problem more clearly (this is compiled from the .c prior to my changes).

clang (apple m1, arm64):
```
0000000000003ec4 <_TakeUint8>:
    3ec4: 1f f8 03 71  	cmp	w0, #254   // w0 is 4 bytes and may or may not be 0 padded, eg could be 0x00FE or 0xFFFE

0000000000003ed4 <_TakeUint16>:
    3ed4: c8 ff 9f 52  	mov	w8, #65534  // same problem as with w0.
    3ed8: 1f 00 08 6b  	cmp	w0, w8
```

`%dil` is explicitly checking the LSB, `%di` is checking the 2 LSBs.
gcc (debian on elf64-x86_64):
```
0000000000001260 <TakeUint8>:
    1260:	40 80 ff fe          	cmp    $0xfe,%dil  //explicit LSB check

0000000000001280 <TakeUint16>:
    1280:	66 83 ff fe          	cmp    $0xfffe,%di // explicit 2 LSB check
```

jonathanstowe: Is there any virtue in putting some `#ifdef` so other platforms keep the original behviour?

jonathanstowe: Is there any virtue in putting this in some `#ifdef` so other pl

niner: I have not had enough time to investigate this properly. But it feels rather like this is an issue in NativeCall itself. It's quite possible that the platform calling ABI expects us to clear the upper bits of those registers and we don't do that. I'd rather get a definitive answer on what we're expected to do and whether we do this or not, before making any changes to our code.

At least we now know what to look for. That's already a huge step forward on this long standing issue!

MasterDuke17: Hm, I wonder if this is at all related to the errors I'm getting in 03-simple-returns.t with my gmp branch on Windows? A negative int64 is getting corrupted somewhere.

tony-o: @niner the clang implementation says that it can go either way (either the upper bytes are `F|0` but nothing else depending on the signedness coming in) and it's not expected that the upper bits are cleared (even if they aren't this isn't a problem on GCC+x86_64 because of `%dil` rather than the comparison of the entire register).  Either NC can force proper signedness for coercion and try to handle what the compiler _might_ do or we can special case clang on osx to do the right thing.  I noticed the uint16 test is skipped in all cases.  Most likely because they are effectively promoted to `int`s during comparison the value is unsigned -1 for 4|8 bytes. I don't think it's NC's problem to try to handle compiler weirdness.

@MasterDuke17 unless you're using clang this is unrelated.  Use -g3 to create the object and `objdump` can give you more info.

tony-o: If you'd rather use a macro to detect clang on arm to just use int and otherwise let it be, let me know and I'll make it these commits that way

MasterDuke17: > @MasterDuke17 unless you're using clang this is unrelated. Use -g3 to create the object and `objdump` can give you more info.

No, I'm getting the errors with MSVC. Oh well, this sounded vaguely similar/related, but I guess not.

lizmat:
I'm not sure of the state of this PR.  Can it be closed?  Should it be merged?

PR #4400, opened 2021-06-12
[JVM] Remove special case for native arrays on JVM

usev6:
It's now possible to use native arrays in the setting.
Fixed with https://github.com/Raku/nqp/pull/720 and
https://github.com/Raku/nqp/pull/722.


lizmat (approved):
Hurray!

PR #4404, opened 2021-06-15
Allow Nil to represent holes in sparse arrays

lizmat:
This commit affects the way arrays are rendered with .raku, and it
affects the way Nil is interpreted as a value when initializing an
array.

This is achieved by the following conceptual changes:

1. in .raku, output a hole (nqp::isnull) as `Nil`

To achieve this, several Iterator classes have their private method
`!hole` turned into a outside visible method `hole` with the
"is implementation-detail" trait.  This allows Array.raku to override
the handling of a hole by mixing in a "hole" method into the iterator
that produces "Nil".

2. in STORE and circumfix:<[ ]> (and its helpers), interprete Nil as
   *skipping* an element to be initialized.  If there was only one
   element specified, and it was Nil, then return an empty Array.

3. fix some Rakudo::Internals helper methods to recognize nqp::null
   values without throwing a tantrum.

This breaks a few tests in roast, that expect that [Nil] will return
[Any], where now it will return [Nil].  There is no further test breakage.

This does *not* fix the case of Nil in arrays constrained with a type
smiley.  Generating the .raku of these is fine, but it won't roundtrip
because `my Int:D @a = 1,Nil,3` is a runtime error.  FWIW, I think this
is actually incorrect behaviour, or at least unproductive behaviour.

This work was done in reply to

  https://github.com/Raku/problem-solving/issues/279

One of the proposed solutions was to introduce a Hole class with special
semantics.  Should that proposal be accepted, then it is mostly replacing
`Nil` by `Hole` in most of this commit.


MasterDuke17:
https://dev.azure.com/Rakudo/rakudo/_build/results?buildId=1520&view=logs&j=07e75ff5-ce69-54f0-467d-376df9cd87cd&t=954af8ab-70c7-51a5-28d5-58c3b1b44a57&l=1828 is new

lizmat:
Can not reproduce this locally :-(

Also, all of the changes I did are no where near any native nums, so this feels like a flapper to me.

MasterDuke17:
> Also, all of the changes I did are no where near any native nums, so this feels like a flapper to me.

Yeah, I meant it's a new flap, I don't think it's related to these changes.

vrurg:
First of all, I haven't analyzed the patch itself sufficiently. I'm also not well-familiar with this part of the core. So, might overlook or misinterpret something.

I'm afraid, replacing `Nil` with `Hole` won't resolve everything. Mostly because where one needs a series of holes they'd need to either `Nil xx $n` or `Hole xx $n` but it's rather expensive both on speed and on memory for the task. That's why I think `Hole` must have `$!count` and `$!until`.

Another question is about stringification. Would `gist` still report the default value for a hole? Or will it report `Nil` instead?

lizmat:
> Mostly because where one needs a series of holes they'd need to either Nil xx $n or Hole xx $n but it's rather expensive both on speed and on memory for the task. That's why I think Hole must have $!count and $!until.

I'm not seeing that: that's merely moving the  work to another place?

> Another question is about stringification. Would gist still report the default value for a hole? Or will it report Nil instead?

Need to  double check that.  AFIAK, `.gist` calls `.AT-POS` and as such would report the default value.

vrurg:
> > Mostly because where one needs a series of holes they'd need to either Nil xx $n or Hole xx $n but it's rather expensive both on speed and on memory for the task. That's why I think Hole must have $!count and $!until.
> 
> I'm not seeing that: that's merely moving the work to another place?

Sorry, not seeing where? I'm losing track a bit. If you mean the attributes, then I was mentioning them in https://github.com/Raku/problem-solving/issues/279#issuecomment-861016978. 

I'd try to guess what you meant and sorry if the guess is wrong. 

I made a mistake at that statement as I assumed that `Seq` produced by `xx` will be unrolled into a `List`. But unless `&infix:<,>` has an error in it, it must not actually happen. So, there is probably no advantage in allocating a single `Hole` object over allocating a `Seq`. Unless the latter comes with hidden expenses I'm not aware of. 

Though another aspect remains the same. For a `Seq` we'd need to iterate to know about how many holes to create. But with `$hole.count` we'd know it right away and this would spare us quite a number of CPU cycles.

As to the "another place" statement – well, perhaps. I'm not sufficiently well familiar with `Array` internals to tell for sure. My initial idea was that holes are to be handled by `&infix:<,>`exclusively. `List`/`Array` `$!reified` would then be taken one-to-one from the `IterationBuffer` produced by the operator. Again, to know for sure if this is feasible and is the right approach I'd need more time to investigate and dig deeper into the code.

jnthn:
Today we can explain `Nil` as "it sets the `Scalar` container it's assigned into back to its default value" (and we explain an `Array` as being made up of `Scalar` containers). With this change, we'd need to add "except when it's in an `Array` composer, or assignment into an `Array`, when it instead makes that element be considered deleted". I'm not against all irregularities, but as in natural languages I'd expect them to show up in the common things, not in dusty corners that have taken years for anybody to really notice.

lizmat:
From the docs:

> The value Nil may be used to fill a spot where a value would normally go, and in so doing, explicitly indicate that no value is present.

I don't think this commit changes that.

> When assigned to a container, the Nil value (but not any subclass of Nil) will attempt to revert the container to its default value; if no such default is declared, Raku assumes Any.

I think the real issue is that "we explain an Array as being made up of Scalar containers", which is *obviously* not true when you do `my @a; @a[2] = 42` and look at `@a[1]`, or do `@a[0] := 666`.

So I really think that this simplification is what is biting us.

raiph:
See also: https://github.com/rakudo/rakudo/issues?q=is%3Aissue+is%3Aopen+hole+in%3Atitle

PR #4406, opened 2021-06-16
Introduce [1,2,3]:Int syntax

lizmat:
Arrays created with [] do not have typing.  It is one of Raku's traps
that people do: sub foo(Int @a) { }, and then get a typecheck error
when they pass [1,2,3] as an argument.  This commit would allow
specification of [1,2,3]:Int as a valid argument.

This implementation uses the fact that you *can* specify adverbs on
operators.  It basically looks if there is a single adverb, takes
its name and does a type-lookup for that.  And then uses that type
to create the correctly constrained Array.

Because we cannot have multi candidates that take *no* named arguments
and take any named arguments, this implementation affects the performance
of the standard logic of [].  So this should probably not be merged as
such.

One way around that now, would be to only allow for classes in Core to
be used as a type: then we could add specific candidates for these.

Another solution could be to think of completely new syntax for this
feature.

### DIFF
--- a/src/core.c/Array.pm6
+++ b/src/core.c/Array.pm6
@@ -182,46 +182,48 @@ my class Array { # declared in BOOTSTRAP
             !! Rakudo::Iterator.Empty            # nothing now or in the future
     }
     method from-iterator(Array:U: Iterator $iter --> Array:D) {
+        my \array := self.new;
         nqp::if(
           nqp::eqaddr(
             $iter.push-until-lazy(
               my \target := ArrayReificationTarget.new(
                 (my \buffer := nqp::create(IterationBuffer)),
-                BEGIN nqp::getcurhllsym('default_cont_spec')
+                nqp::getattr(array,Array,'$!descriptor')
               )
             ),
             IterationEnd
           ),
-          nqp::p6bindattrinvres(nqp::create(self),List,'$!reified',buffer),
+          nqp::p6bindattrinvres(array,List,'$!reified',buffer),
           nqp::stmts(
-            nqp::bindattr((my \result := nqp::create(self)),
-              List,'$!reified',buffer),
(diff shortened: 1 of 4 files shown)


jnthn:
> It is one of Raku's traps
> that people do: sub foo(Int @A) { }, and then get a typecheck error
> when they pass [1,2,3] as an argument. 

This is true, but...

> This commit would allow
> specification of [1,2,3]:Int as a valid argument.

If they knew to do *this*, they'd have known to like it like `Array[Int](1, 2, 3)` or similar. Granted that is a bit verbose, but taking a look at the error output today:

```
$ raku -e 'sub foo(Int @a) { }; foo [1,2,3]'
Type check failed in binding to parameter '@a'; expected Positional[Int] but got Array ([1, 2, 3])
  in sub foo at -e line 1
  in block <unit> at -e line 1
```

Aside from the fact that this one should be possible to detect at compile time, the error doesn't provide much of a clue on what to do. We could suggest using `Array[Int](...)` instead, or adding a type to a declaration. (Maybe can see if there's a value returned by `.name` on the `Array` to decide which of those to suggest.)

As for the proposed syntax, it has the end weight problem (the reason that Raku has `m:g/x/y/` rather than Perl's `m/x/y/g`). I'm certainly not against looking for a more succinct spelling than the admittedly verbose `Array[Int](...)`, but it should solve more than just that (e.g. offer a solution for typed hashes too, and perhaps a way to omit the type and have it inferred from the first element).

patrickbkr:
I'd whole-heartedly welcome a better error message here. I have seen this error many times and after some searching around (and not realizing what the real problem was) always ended up just removing the type constraint of the routine.

lizmat:
Closing this in the hope that someone will make a PR with a better error message.

PR #4407, opened 2021-06-17
Support generics with consumption/inheritance in roles

vrurg:
Make the following two constructs possible:

    role R1[::T] does T {}
    role R2[::T] is T {}

These were assumed possible, according to a comment in ParametricRoleHOW
code. But a few final touches needed to make them possible:

- traits must accept generics when applied to a parametric type
- typechecks must be done against instantiated typeobjects too,
  including parents
- additional validation for archetype of a consumed typeobject after
  instantiation since it bypassed trait validation

Spectests are passing.


vrurg:
I currently see this rather useful for cases like:

```raku
role R[::T Foo] is T { }
class C does R[FooChild] { }
```

which would certainly be better than something like:

```raku
role R[::T Foo] { }
class C does R[FooChild] is FooChild { }
```

Especially when `R` declared in a module.

PR #4408, opened 2021-06-17
Test mkdir when parent was created but child fails

McSinyx:
Move Raku/roast@a0d3a9aa27aaa9c2d204db34a33579dd2fdc3f22 from Raku/roast#736 due to the lack of portability of the condition that is tested against (on a few file system like Reiser the maximum filename length is larger than 255).  This test is supposed to show the MoarVM bug where `mkdir` reports success after creating some of the parent directory but (silently) failed to create the final directory.

### DIFF
--- a/t/02-rakudo/19-mkdir.t
+++ b/t/02-rakudo/19-mkdir.t
@@ -1,0 +1,9 @@
+use lib <t/packages/>;
+use Test;
+use Test::Helpers;
+
+plan 1;
+
+# mkdir soft-fails when directory name is too long (over 255 bytes).
+my $path = make-temp-dir.add("foobar").add("universe" x 42);
+fails-like { $path.mkdir }, X::IO::Mkdir;


vrurg:
One minor note: I don't think the test belongs to moar only. Event though the original bug popped up in this backend, but it may re-appear under other backends too. So, perhaps _t/02-rakudo/_ is better location for it.

But otherwise I'm more curious about the failed CI checks. Looks like we have what was discussed in roast PR: mkdir doesn't fail. Perhaps the path length is still not sufficiently long? It'd be nice to diagnose it on a local linux box. If the length is the culprit then we can either try making it more than 4096, or even 32768 which the longest what NTFS theoretically supports.

McSinyx:
> One minor note: I don't think the test belongs to moar only.
> Event though the original bug popped up in this backend,
> but it may re-appear under other backends too.
> So, perhaps _t/02-rakudo/_ is better location for it.

Agreed, I moved the test.

> But otherwise I'm more curious about the failed CI checks.
> Looks like we have what was discussed in roast PR: mkdir doesn't fail.

The CI status is expected to me: all MoarVM checks fail but JVM.
The MoarVM patch hasn't been merged yet.

vrurg (approved):
Ready for merge when MoarVM/MoarVM#1507 lands in moar.

McSinyx:
@zhuomingliang, do you think MoarVM/MoarVM#1507 is ready to be merged now, i.e. could you merge it?  Thanks all for the reviews BTW.

zhuomingliang:
Anyone could merge it after next release.

vrurg:
@McSinyx could you, please, rebase the PR on top of the latest master?

McSinyx:
Sure, no problem.

zhuomingliang:
Parameters

lpPathName

The path of the directory to be created.

For the ANSI version of this function, there is a default string size limit for paths of 248 characters (MAX_PATH - enough room for a 8.3 filename). To extend this limit to 32,767 wide characters, call the Unicode version of the function and prepend "\\?\" to the path. For more information, see Naming a File
Tip  Starting with Windows 10, version 1607, for the unicode version of this function (CreateDirectoryW), you can opt-in to remove the 248 character limitation without prepending "\\?\". The 255 character limit per path segment still applies. See the "Maximum Path Length Limitation" section of Naming Files, Paths, and Namespaces for details.

zhuomingliang:
see https://docs.microsoft.com/zh-cn/windows/win32/api/fileapi/nf-fileapi-createdirectoryw

Parameters

lpPathName

The path of the directory to be created.

For the ANSI version of this function, there is a default string size limit for paths of 248 characters (MAX_PATH - enough room for a 8.3 filename). To extend this limit to 32,767 wide characters, call the Unicode version of the function and prepend "\\?\" to the path. For more information, see Naming a File

```Tip  Starting with Windows 10, version 1607, for the unicode version of this function (CreateDirectoryW), you can opt-in to remove the 248 character limitation without prepending "\\?\". The 255 character limit per path segment still applies. See the "Maximum Path Length Limitation" section of Naming Files, Paths, and Namespaces for details.```

vrurg (changes requested):
I would put this PR on hold until it is fixed.

lizmat:
Is this ready for merging now?

McSinyx:
Oops sorry I forgot about this.  No, the tests are supposed to pass after the MoarVM patch was merged.

McSinyx:
@lizmat I think I wasn't clear, I meant it was not ready for merging because the tests failed while they should pass.  I should have converted this to draft.

lizmat:
The test passed for me, I checked.

I assumed the CI failures were caused by lack of MoarVM merge.  My bad.

vrurg:
I'm confused as the MoarVM PR was merged 17 days ago. I thought it should have been bumped since then, but a week ago this PR was still failing.

McSinyx:
It seems that [it's indeed passing](https://dev.azure.com/Rakudo/rakudo/_build/results?buildId=1611&view=results) after dependencies bumping (2bf59702e8fa5bd108a2d3e1332badbe47622db7).  The failing tests seems to be unrelated.

vrurg:
Ok, I will re-revert the PR later today and make sure everything is ok with the latest NQP/Moar bump.

@McSinyx thanks!

McSinyx:
Thanks everyone for the reviews and troubleshooting!

PR #4412, opened 2021-06-17
Introduce "done foo" as shortcut for "emit foo; done"

lizmat:
As discussed in https://github.com/Raku/problem-solving/issues/249

This makes the "done" sub a multi, and adds logic in the "run-done"
handlers in Rakudo::Supply to handle anything specified that is not
an exception.

Should we want to be able to emit an Exception, then we will have to
add some more logic for handing the QUIT phaser's handling of "run-done".

### DIFF
--- a/src/core.c/Rakudo/Supply.pm6
+++ b/src/core.c/Rakudo/Supply.pm6
@@ -129,14 +129,6 @@ class Rakudo::Supply {
         }
 
         method run-done(--> Nil) {
-            my \ex := nqp::exception();
-            unless nqp::isnull(ex) {
-                my \value := nqp::getpayload(ex);
-                unless nqp::isnull(value) || nqp::istype(value,Exception) {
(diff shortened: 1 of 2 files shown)


jnthn:
Really, I'd expected the change to amount to making `done` a `multi` and then adding a candidate:

```
multi done(Mu \value --> Nil) {
    emit value;
    done;
}
```

jnthn:
> Should we want to be able to emit an Exception, then we will have to
> add some more logic for handing the QUIT phaser's handling of "run-done".

I think we should be able to; I like being able to explain features in terms of equivalence to simpler features ("desugaring"), so I find it awkward that we'd have a discontinuity with exceptions. I'd really rather fix it by not changing `run-done` at all though, and doing the boring `emit` + `done` call sequence.

jnthn (approved):

Issue #4410, opened 2021-06-17
Bogus "emit without supply or react" when using hyper


niner:
## The Problem

Using `.hyper` when processing an array leads to an occasional "emit without supply or react" error in a worker. This can be trivially reproduced with:

```
> rakudo -e 'my @a; loop { @a.hyper[0] }'
A worker in a parallel iteration (hyper or race) initiated here:
  in block <unit> at -e line 1
  in block <unit> at -e line 1

Died at:
    emit without supply or react
      in block <unit> at -e line 1
      in block <unit> at -e line 1
```

Tested on Rakudo commit 28ff12a8b884ac01044e4c3ecd07feeab8e2b73f

lizmat:
Confirmed on MacOS.

niner:
It's *not* a spesh issue. Can reproduce it with MVM_SPESH_DISABLE=1 set, but it takes longer to fail.
It's also *not* caused by commit 6f42d19eb0. I suspected that one but reverting it also turned out to just defer the inevitable.

niner:
@dogbert17 noted some suspicious tsan output: https://gist.github.com/dogbert17/dea0a2c20be1826851615980be3b1710
I'm skeptical about a data race though. AFAICT cache_node_head can only ever change from NULL to a valid pointer, but not back, so I don't think this can cause any issues.

lizmat:
A full stack traces, just in case

```
$ rl 'my @a; loop { @a.hyper[0] }'
emit without supply or react
   at SETTING::src/core.c/Exception.pm6:62  (/Users/liz/Github/rakudo.moar/install/share/perl6/runtime/CORE.c.setting.moarvm:throw)
 from SETTING::src/core.c/Exception.pm6:569  (/Users/liz/Github/rakudo.moar/install/share/perl6/runtime/CORE.c.setting.moarvm:print_control)
 from SETTING::src/core.c/Exception.pm6:591  (/Users/liz/Github/rakudo.moar/install/share/perl6/runtime/CORE.c.setting.moarvm:<anon>)
 from gen/moar/stage2/NQPHLL.nqp:2094  (/Users/liz/Github/rakudo.moar/install/share/nqp/lib/NQPHLL.moarvm:)
 from <unknown>:1  (/Users/liz/Github/rakudo.moar/install/share/nqp/lib/NQPHLL.moarvm:)
[...]

PR #4414, opened 2021-06-18
Fix bogus "emit without supply or react" in Supplier's TapList

niner:
Avoid auto-vivification of the Scalar container for the $!tappers
attribute when reading. That could lead to a Mu overwriting the list
in $!tappers after we've already set $added to True in method tap,
causing us to blow up when we try to call .emit on that Mu.

Fixes GH #4410

### DIFF
--- a/src/core.c/Supplier.pm6
+++ b/src/core.c/Supplier.pm6
@@ -17,6 +17,11 @@ my class Supplier {
         # need lock on modification).
         has Mu $!tappers;
 
+        # Avoid auto-vivification of the Scalar container for the $!tappers
+        # attribute when reading. That could lead to a Mu overwriting the list
+        # in $!tappers after we've already set $added to True in method tap.
+        submethod BUILD() { $!tappers := Mu; }
+
         method tap(&emit, &done, &quit, &tap) {
             my $tle := TapListEntry.new(:&emit, :&done, :&quit);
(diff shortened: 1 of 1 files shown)


lizmat:
```
has Mu $!tappers is built(False) is built(:bind) = Mu;
```
achieves the same result, but does not have a costly `BUILD` call.  At least for me?

jnthn:
@lizmat Why the two `is built` trait applications? I'd expect the first to be redundant given the second?

lizmat:
Because `is built(:bind)` implies that it can be built with `.new`.

But indeed, the order is wrong: it should have been `is built(:bind) is built(False)` to get the semantics from before.

jnthn:
Maybe `is built(:bind)` should go away in favor of making `has Mu $!tappers := Mu` work (it's an error today).

lizmat:
There's that  :-)

niner:
I merged it as the fix is tested and works as is and is low risk and I would really like to have this in the release. Since the BUILD will only run on a Supplier.new call, performance shouldn't be much of an issue. You will usually create a Supplier only during setup of a threaded or asynchronous processing pipeline, i.e. maybe a hand full.

When we have support for `has Mu $!tappers := Mu` we can use that as a more elegant way of course :) Or maybe we'll even get rid of the underlying issue with lazy initialization of Scalar containers.

PR #4415, opened 2021-06-18
Implement `last` with value

lizmat:
This allows for the following constructs:

```raku
say ^10 .map: { $_ == 5 ?? (last)   !! $_ }  # (0 1 2 3 4)
say ^10 .map: { $_ == 5 ?? last(42) !! $_ }  # (0 1 2 3 4 42)
```

and also in `.grep`:
```raku
say ^10 .grep: { $_ == 5 ?? last       !! $_ %% 2 }  # (0 2 4)
say ^10 .grep: { $_ == 5 ?? last(True) !! $_ %% 2 }  # (0 2 4 5)
```

and any other loop constructs that return values, e.g.:

```raku
my $i; say do while ++$i < 10 { $i == 5 ?? (last)   !! $i }  # (1 2 3 4)
```
vs
```raku
my $i; say do while ++$i < 10 { $i == 5 ?? last(42) !! $i }  # (1 2 3 4 42)
```

### DIFF
--- a/src/core.c/Any-iterable-methods.pm6
+++ b/src/core.c/Any-iterable-methods.pm6
@@ -156,7 +156,11 @@ Consider using a block if any of these are necessary for your mapping code."
                     'REDO', ($stopped = 0),
                     'LAST', nqp::stmts(
                       ($!did-iterate = 1),
-                      ($result := IterationEnd)
+                      nqp::if(
+                        nqp::isnull($result := nqp::getpayload(nqp::exception)),
+                        ($result  := IterationEnd),
+                        ($!source  := Rakudo::Iterator.Empty),
+                      )
                     )
                   ),
                   :nohandler
@@ -224,7 +228,13 @@ Consider using a block if any of these are necessary for your mapping code."
                             ?? ($done = 1)
                             !! ($stopped = 0)),
                         'REDO', ($stopped = 0),
-                        'LAST', ($done = $!did-iterate = 1)
+                        'LAST', nqp::stmts(
+                          ($done = $!did-iterate = 1),
+                          nqp::unless(
+                            nqp::isnull($value := nqp::getpayload(nqp::exception)),
+                            target.push($value)
+                          )
+                        )
                       )
                     ),
                     :nohandler
@@ -357,8 +367,11 @@ Consider using a block if any of these are necessary for your mapping code."
                     ),
                     'REDO',
                     ($redo = 1),
-                    'LAST',
-                    ($result := IterationEnd)
+                    'LAST', nqp::if(
+                      nqp::isnull($result := nqp::getpayload(nqp::exception)),
+                      ($result := IterationEnd),
+                      ($!source := Rakudo::Iterator.Empty),
+                    )
                   ),
                 ),
               :nohandler);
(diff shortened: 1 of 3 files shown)


lizmat:
Looks like I missed some other types of looping structures.

lizmat:
I've just realized that making `next value` should not be too difficult to implement as well :-)

But I'll wait for this to be merged first.

lizmat:
Converted back to draft as it needs support for Slips in values:

```raku
dd ^10 .map: { last |(42,66) if $_ == 5; $_ }
```
should be `(0,1,2,3,4,42,666).Seq`, not `(0,1,2,3,4,slip(42,66)).Seq`

lizmat:
Support for `Slip` added

moon-chilled:
```raku
do LABEL: for ^10 {
    last LABEL
}
```

What if I want to emit the label object?

moon-chilled:
It might also be nice to be able to do something like this:

```raku
do LABEL: for ^10 {
    INNER: loop {
        last LABEL, 17
    }
}
```

Which would let you disambiguate my example above with `last LABEL, LABEL`.

vrurg:
> What if I want to refer to the value of the label object instead of calling `LABEL.last`?

I'd say that you don't want it. :) Seriously, you can wrap it into a class-container. Something like:

```
class LastCont {
    has Mu $.payload;
}
#...
last LastCont.new(:payload(LABEL));
```

There are ways to simplify the syntax and reduce it to `LastCont(LABEL)`, but that out of the scope.

`last LABEL, $value` is of more interest, I'd say.

lizmat:
I think `last LABEL, $value` is basically a no-no, as that exits the iterator directly, so has no way of getting the value into the iterator's value stream.

This also gets into the territory of having `last` take multiple values, as a shorter way to describe a `Slip`.  AKA:

```raku
last 42, 666;
```
to be the equivalent of:
```raku
last slip(42,666);
```
Would a `Label` object in there mean something?

I tried to wrap a `Label` in a container and see if `last $(LABEL)` would work.  First of all, that would only make sense in a situation where the values are being collected.  I have not been able to create an example that would allow that without running into `labeled last without loop construct` error, so I have given up on that idea for the moment.

melezhik:
@lizmat  I've run r3 against your branch, and not all test passes (ex 3.1, ex7 failed )

https://gist.github.com/melezhik/3dcc2db1297bade97f1803bbd954c155

HTH

lizmat:
@melezhik
```raku
say ^10 .grep: { $_ == 5 ?? last       !! $_ %% 2 }
```
is not valid Raku code.  You need to either do `last()` or `(last)`.

melezhik:
@lizmat  I suspected that , thanks. I just used the examples in the ticket ...

melezhik:
I changed last to (last), now all tests pass except the last one

https://gist.github.com/melezhik/fcc4523f66cfce66c8fa5dc42ad63911

melezhik:
Hi @lizmat , if this supposed to work? I tried both with and without parenthesis after last, both cases do not give expected `(0,1,2,3,4,42,666).Seq`

```
[root@2834387600a8 r3tool]# $RAKUBIN -e 'dd ^10 .map: { last |(42,66) if $_ == 5; $_ }'
Too many positionals passed; expected 0 or 1 arguments but got 2
  in block <unit> at -e line 1
```

```
[root@2834387600a8 r3tool]# $RAKUBIN -e 'dd ^10 .map: { last() |(42,66) if $_ == 5; $_ }'
WARNINGS for -e:
Useless use of "|" in expression "last() |(42,66)" in sink context (line 1)
(0, 1, 2, 3, 4).Seq
```

lizmat:
with `last |(42,66)` you're actually not slipping a list, but you're flattening a list as positional parameters (the original use of prefix `|`).  You want `last slip(42,666)`, which I think actually reads better as well.

melezhik:
Ok. `dd ^10 .map: { last slip(42,66) if $_ == 5; $_ }` now gives `(0, 1, 2, 3, 4, 42, 66).Seq` is it correct result?

lizmat:
Yup

melezhik:
thanks, I have fixed all r3 tests now.

PR #4416, opened 2021-06-19
Fix freezes caused by concurrently produced iterators

### LINKED ISSUE #4413: Weird interaction between array interpolation, given and hyper
## The Problem

There is some kind of weird interaction / interference between array interpolation, given and hyper.

Kind of bogus example: Pay no attention to the grep operation, or the actual printed result. It is immaterial. Just here for illustration purposes.

The hyper is obviously counterproductive for a such a small range, but I tripped over this when I was trying to use a ^1_000_000 range where hyper helps a lot.

(**EDIT:** golfed down the examples a little, cut out some extraneous clutter. )

Works fine: ( no hyper grep )
```
say "{+$_} found:\n{.batch(10).join: "\n"}" given (^100).grep: { .is-prime };
```
Also fine: ( explicitly expose the topic as an array when interpolating: @$_ )
```
say "{+$_} found:\n{@$_.batch(10).join: "\n"}" given (^100).hyper.grep: { .is-prime };
```
Hangs forever: ( well, 10 minutes at least before I killed it. )
```
say "{+$_} found:\n{.batch(10).join: "\n"}" given (^100).hyper.grep: { .is-prime };
```

Interestingly enough, it seems like the {+$_} is important. Works as expected if you remove it.
```
say " found:\n{.batch(10).join: "\n"}" given (^100).hyper.grep: { .is-prime };
```
Even moving the +$_ outside of the string interpolation doesn't help.
```
say .elems ~ " found:\n{.batch(10).join: "\n"}" given (^100).hyper.grep: { .is-prime };
```
Hangs

It may not be related to the interpolation. This also hangs:
```
say .elems ~ " found:\n" ~ .batch(10).join: "\n" given (^100).hyper.grep: { .is-prime };
```

## Environment
* Operating system: Linux Mint 20.1 Ulyssa
* Compiler version (`perl6 -v` or `raku -v`): $ raku -v
Welcome to Rakudo(tm) v2021.05.
Implementing the Raku(tm) programming language v6.d.
Built on MoarVM version 2021.05.

vrurg:
Freezes were happening with code similar to:

    (.list, .list) given (^10).hyper;

The cause was tracked down to two iterators created for the same
`$!work-stage-head` object. In an identical situation `Seq` would throw
`X::Seq::Consumed` which seems logical. Similar approach was taken for
`HyperSeq` and `RaceSeq`.

Aside of this, I noticed that `HyperSeq` and `RaceSeq` share basicall
all their code except for `hyper` and `race` methods. I found it
reasonable to move all commonalities into a role.

Fixes #4413

### DIFF
--- a/src/core.c/Exception.pm6
+++ b/src/core.c/Exception.pm6
@@ -3184,11 +3184,12 @@ my class X::MultipleTypeSmiley does X::Comp {
 }
 
 my class X::Seq::Consumed is Exception {
-    has Str:D $.kind = "Seq";
+    has $.kind = Seq;
     method message() {
-        ("The iterator of this $!kind is already in use/consumed by another $!kind" ~
-        " (you might solve this by adding .cache on usages of the $!kind, or " ~
-        "by assigning the $!kind into an array)").naive-word-wrapper
+        my $kind_name = $!kind.^name;
+        ("The iterator of this $kind_name is already in use/consumed by another $kind_name" ~
+        " (you might solve this by adding .cache on usages of the $kind_name, or " ~
+        "by assigning the $kind_name into an array)").naive-word-wrapper
     }
 }
 
--- a/src/core.c/Exception.pm6
+++ b/src/core.c/Exception.pm6
@@ -3184,10 +3184,11 @@ my class X::MultipleTypeSmiley does X::Comp {
 }
 
 my class X::Seq::Consumed is Exception {
+    has Str:D $.kind = "Seq";
     method message() {
-        "The iterator of this Seq is already in use/consumed by another Seq\n" ~
-        "(you might solve this by adding .cache on usages of the Seq, or\n" ~
-        "by assigning the Seq into an array)"
+        ("The iterator of this $!kind is already in use/consumed by another $!kind" ~
+        " (you might solve this by adding .cache on usages of the $!kind, or " ~
+        "by assigning the $!kind into an array)").naive-word-wrapper
     }
 }
 
--- a/src/core.c/HyperSeq.pm6
+++ b/src/core.c/HyperSeq.pm6
@@ -1,45 +1,58 @@
 # A HyperSeq performs batches of work in parallel, but retains order of output
 # values relative to input values.
 #?if !js
-my class HyperSeq does Iterable does Sequence {
+
+my role Sequence::Hyper does Iterable does Sequence {
     has HyperConfiguration $.configuration;
     has Rakudo::Internals::HyperWorkStage $!work-stage-head;
+    has $!joiner;
 
     submethod BUILD(:$!configuration!, :$!work-stage-head!) {}
 
-    method iterator(HyperSeq:D: --> Iterator) {
-        my $joiner := Rakudo::Internals::HyperToIterator.new:
+    method iterator(::?CLASS:D: --> Iterator) {
+        X::Seq::Consumed.new(:kind(::?CLASS.^name)).throw
+            if nqp::isconcrete($!joiner);
+        $!joiner := Rakudo::Internals::HyperToIterator.new:
             source => $!work-stage-head;
-        Rakudo::Internals::HyperPipeline.start($joiner, $!configuration);
-        $joiner
+        Rakudo::Internals::HyperPipeline.start($!joiner, $!configuration);
+        $!joiner
     }
 
-    method grep(HyperSeq:D: $matcher, *%options) {
+    method grep(::?CLASS:D: $matcher, *%options) {
         Rakudo::Internals::HyperRaceSharedImpl.grep:
             self, $!work-stage-head, $matcher, %options
     }
 
-    method map(HyperSeq:D: $matcher, *%options) {
+    method map(::?CLASS:D: $matcher, *%options) {
         Rakudo::Internals::HyperRaceSharedImpl.map:
             self, $!work-stage-head, $matcher, %options
     }
 
-    method invert(HyperSeq:D:) {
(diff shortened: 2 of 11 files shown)


vrurg:
I'm not marking this PR as draft because it is basically done and ready. The only thing I'm not sure about is the new role name. Is `Sequence::Hyper` ok or is it better to move it under `Rakudo::Internals` and make something like `Rakudo::Internals::HyperSequence` for uniformity purpose?

Also, when the role is moved into a dedicated file it's gonna be almost the first case of adding a new one since Raku extensions were approved. So, it's tempting to make it the first `.rakumod` in the core.

lizmat:
As this is mostly @jnthn's brainchild, I'm deferring review to him.  I would definitely *not* merge this for a 2021.06 release.

On other notes:

- I'm ok with the name
- It can live in Rakudo::Internals, at least for now.

jnthn:
Given `Sequence` is the base role for every kind of `Seq`, I figure something like `ParallelSequence` would be more fitting.

src/core.c/Exception.pm6
```diff
@@ -3184,10 +3184,11 @@ my class X::MultipleTypeSmiley does X::Comp {
 }
 
 my class X::Seq::Consumed is Exception {
+    has Str:D $.kind = "Seq";
```
jnthn, line 3187: I think it'd be more appropriate to pass the type object or the `Sequence` itself into the exception class, and leaving `message` to do the stringification.

src/core.c/Exception.pm6
```diff
@@ -3184,10 +3184,11 @@ my class X::MultipleTypeSmiley does X::Comp {
...
     method message() {
-        "The iterator of this Seq is already in use/consumed by another Seq\n" ~
-        "(you might solve this by adding .cache on usages of the Seq, or\n" ~
-        "by assigning the Seq into an array)"
+        ("The iterator of this $!kind is already in use/consumed by another $!kind" ~
+        " (you might solve this by adding .cache on usages of the $!kind, or " ~
+        "by assigning the $!kind into an array)").naive-word-wrapper
```
jnthn, line 3191: Hm what is `naive-word-wrapper`? I'd expect such things to be applied at the point of exception printing, not in the message production...

src/core.c/HyperSeq.pm6
```diff
@@ -1,46 +1,59 @@
...
+    has $!joiner;
 
     submethod BUILD(:$!configuration!, :$!work-stage-head!) {}
 
-    method iterator(HyperSeq:D: --> Iterator) {
-        my $joiner := Rakudo::Internals::HyperToIterator.new:
+    method iterator(::?CLASS:D: --> Iterator) {
+        X::Seq::Consumed.new(:kind(::?CLASS.^name)).throw
+            if nqp::isconcrete($!joiner);
+        $!joiner := Rakudo::Internals::HyperToIterator.new:
```
jnthn, line 15: I wonder if using `cas` or even just a native int attribute and an atomic increment would be more robust here. That way we'd catch attempts to consume it in parallel too (and don't have to be concerned about whether we're extending the lifetime of something by storing it).

src/core.c/Exception.pm6
vrurg: It was the initial plan, but then... Let's put it that way: I was a bit tired and for some reason decided `Seq` is not available here.

src/core.c/Exception.pm6
vrurg: But we currently don't do it anyway. There are a couple other exceptions doing this. This particular message was previously using newlines inside the message to wrap it. So, I'd keep it this way until the printing starts wrapping.

src/core.c/HyperSeq.pm6
vrurg: With all my love for parallelizing it's the kind of error I tend to repeat. Too much Perl in the veins, perhaps. ;) Also good point on preserving the joiner when we don't really need it. Because it's a case of boolean semantics, cas over atomicint would do the best.

Thanks!

vrurg:
@jnthn All notes were taken into account except for the `naive-word-wrapper` due to the concerns I wrote about.

I also considered making `Seq` thread-safe too but it's probably never meant to be. So, I dropped the idea.

src/core.c/Exception.pm6
jnthn: Yes, to be fair this was a more general grumble rather than anything that should block this PR.

src/core.c/ParallelSequence.rakumod
```diff
@@ -0,0 +1,44 @@
+# ParallelSequence role implements common functionality of HyperSeq and RaceSeq classes.
+my role ParallelSequence does Iterable does Sequence {
+    has HyperConfiguration $.configuration;
+    has Rakudo::Internals::HyperWorkStage $!work-stage-head;
+    has $!has-iterator;
```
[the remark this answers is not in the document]
jnthn: (What you have relies on interning of `Int`, which is probably reliable, but still...)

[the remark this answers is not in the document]
vrurg: > I think this should be `has atomicint $!has-iterator`

`atomicint` was my choice too, but it is not generally available at this point.  I tried changing the order of build with no success. But there was another error which I located later. Perhaps it was the problem, not the order. So, I will try that later today.

jnthn (approved):

vrurg:
I will merge when get JVM backend fixed. It would probably have to deal with `Int`.

PR #4425, opened 2021-06-22
Fix unimatch for non-base properties

codesections:
Prior to PR #4172, unimatch worked for non-base properties such as
```raku
    unimatch '[', ']', 'Bidi_Mirroring_Glyph'; # True
```
That PR, however, provided a speedup for base properties, but broke the more advanced use.  This adds back support for the more advanced use case as a fallback _after_ testing for a base property (which should avoid performance impacts in most cases).

If this generally looks good, I'll add Roast tests to [S15/unimatch-general](https://github.com/Raku/roast/blob/dffc2a4cec4a1b91a14289a88cead3ceecd5f5b4/S15-unicode-information/unimatch-general.t) which currently tests the base property but not others (such as those that return strings).

Thanks to @alabamenhu for help tracking down this regression.

### DIFF
--- a/src/core.c/unicodey.pm6
+++ b/src/core.c/unicodey.pm6
@@ -77,7 +77,7 @@ my class Rakudo::Unicodey is implementation-detail {
         my $prop := nqp::unipropcode($propname);
         nqp::hllbool(
           nqp::matchuniprop($code,$prop,nqp::unipvalcode($prop,$pvalname))
-        )
+        ) || uniprop($code, $propname) eq $pvalname
     }
 
     my constant $gcprop = nqp::unipropcode("General_Category");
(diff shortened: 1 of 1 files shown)


lizmat (approved):

melezhik:
works for me, when test against pr branch:

```
[root@42fb7c28b625 r3tool]# tomty --color pull_4425
[repository] :: index updated from file:///root/repo/api/v1/index
[pull_4425] :: cat $root_dir/task.bash
[pull_4425] ::
[pull_4425] :: $RAKUBIN -v
[pull_4425] ::
[pull_4425] :: $RAKUBIN -e "say '[', ( unimatch '[', ']', 'Bidi_Mirroring_Glyph' ), ']'"
[pull_4425] :: Welcome to 𝐑𝐚𝐤𝐮𝐝𝐨™ v2021.06-10-g0a1256afa.
[pull_4425] :: Implementing the 𝐑𝐚𝐤𝐮™ programming language v6.d.
[pull_4425] :: Built on MoarVM version 2021.06-6-g6977d0591.
[pull_4425] :: [True]
[task check] stdout match <[True]> True
```

PR #4428, opened 2021-06-24
Support (nearly) all HTML5 entities in Pod E<>

raydiak:
This PR adds HTML entities with mixed case which were missing from the hash (e.g. `Assign` for ≔), and fixes parsing of entity names containing numbers (e.g. `sup2` for ²). Spectest is clean. `E<>` coverage in roast seems to be lacking, so I also performed a few manual sanity checks to ensure we haven't broken any of the various forms of integer codepoints by allowing digits in entity names. I intend to add some extra tests for this in roast in the near future. I've also taken @vrurg's suggestion in rakudo/rakudo#4397 and modernized this to an `nqp::hash` instead of a `for` loop.

A few (11) HTML entities have uppercase versions. All but one of these are synonyms for their lowercase counterparts and exist only for backwards compatibility with older HTML versions. The one exception is `THORN` vs `thorn` (Þ vs þ). For whatever reason, this is the only character in the entire W3C HTML5 entity list to have its uppercase form represented by an all-uppercase name, instead of title case. A quick check in Firefox confirmed that it is actually implemented this way in practice. The only living language which uses this character according to https://en.wikipedia.org/wiki/Thorn_(letter) is modern Icelandic. It might also be (somewhat erroneously) used to represent the similar rune it is derived from which has significance in some still-practiced belief systems. It could also occur in a historical context as part of a few dead languages.

Uppercase names currently always parse as Unicode names, not HTML. This limitation could theoretically be addressed in the action method by checking for an HTML entity as a fallback if the string isn't a valid Unicode name. Accordingly, the 11 uppercase names are retained in the hash in case somebody decides to implement that.


vrurg (approved):

vrurg:
Unless somebody minds, I would merge it soon.

melezhik:
may I see a couple of code examples for this feature request, please?

raydiak:
As noted in my initial comment, in Pod `E<Assign>` now yields `≔`, and `E<sup2>` now yields `²`, as two examples. See the commit for the full list of entity names, which contains a hash of all currently-specified HTML5 entites.

It's less of a feature request, and more of a completion of the implementation as it was always intended. This feature was originally speculated in https://design.raku.org/S26.html#Entities. Mixed case identifiers were already supported by the grammar, but were missing from the %entities hash. The grammar support was extended to allow entity names which contain digits but don't successfully parse as integer codepoints.

After @coke mentioned https://logs.liz.nl/raku/2021-06-11.html#01:38, I began #4397 and continued with #4427 and Raku/roast#740. Since I was already working in that area recently and had noticed the entities hash was incomplete, as well as having bumped into a parse failure while attempting to test those changes with `E<sup2>`, it seemed natural to complete the work I started.

vrurg:
Simply put, `E<>` is defined as the way to include HTML5 entities into a doc. So, supporting the full HTML5 set makes full sense.

melezhik:
works for me (in r3)

```
[root@42fb7c28b625 r3tool]# SP6_FORMAT_TERSE=1 tomty --color   pr_4428
index updated from file:///root/repo/api/v1/index
cat $root_dir/task.bash

$RAKUBIN -v

cat $root_dir/test.raku

$RAKUBIN --doc=Text $root_dir/test.raku
Welcome to 𝐑𝐚𝐤𝐮𝐝𝐨™ v2021.06-19-ge7ba577f2.
Implementing the 𝐑𝐚𝐤𝐮™ programming language v6.d.
Built on MoarVM version 2021.06-6-g6977d0591.
=begin pod
Hey, I have some html5 code!

E<sup3>

E<sup2>

=end pod


say "I am fine";
Hey, I have some html5 code!

³

²
[task check] stdout match (s) <Hey, I have some html5 code!> True
[task check] stdout match (s) <^^ $$> True
[task check] stdout match (s) <^^ '³' $$> True
[task check] stdout match (s) <^^ $$> True
[task check] stdout match (s) <^^ '²' $$> True
```

raydiak:
Cool deal. I have spectests I'll push to roast as well, once this is merged

melezhik:
Sure. You can use R3 to test against your branches, before it's merged 😊

raydiak:
I put the roast tests into a PR (Raku/roast#741) instead of pushing them directly, in hopes that it will increase visibility of this PR here and prevent it from falling into awaiting-merge limbo. :) Whoever merges this, feel free to merge those tests as well.

PR #4433, opened 2021-06-25
Enable rendering of nested blocks in Pod::To::Text

raydiak:
This corrects a number of omissions when rendering S26 (and possibly elsewhere). Rakudo tests and spectests all pass cleanly. A diff of the output from S26 before and after this PR revealed no undesirable side effects.

### DIFF
--- a/lib/Pod/To/Text.rakumod
+++ b/lib/Pod/To/Text.rakumod
@@ -49,7 +49,7 @@ sub item2text($pod) {
 sub named2text($pod) {
     given $pod.name {
         when 'pod'  { pod2text($pod.contents)     }
-        when 'para' { para2text($pod.contents[0]) }
+        when 'para' { pod2text($pod.contents) }
         when 'config' { }
         when 'nested' { pod2text($pod.contents).indent(4) }
         default     { $pod.name ~ "\n" ~ pod2text($pod.contents) }
--- a/lib/Pod/To/Text.rakumod
+++ b/lib/Pod/To/Text.rakumod
@@ -51,7 +51,7 @@ sub named2text($pod) {
         when 'pod'  { pod2text($pod.contents)     }
         when 'para' { para2text($pod.contents[0]) }
         when 'config' { }
-        when 'nested' { }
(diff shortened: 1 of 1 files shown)


melezhik:
@raydiak  may I see an example of "nested blocks"? thanks

raydiak:
@melezhik https://design.raku.org/S26.html#Nesting_blocks gives explanation and examples. It is used in many places in that document's source itself, which can be viewed at https://github.com/Raku/old-design-docs/blob/master/S26-documentation.pod, and also occurs in roast at https://github.com/Raku/roast/blob/0a328684045702eed1d0589a665f5160ae46ecfc/S26-documentation/02-paragraph.t#L122.

There is another instance in roast at https://github.com/Raku/roast/blob/0a328684045702eed1d0589a665f5160ae46ecfc/S02-lexical-conventions/unspace.t#L163, but it is fudged and I am unclear on the syntactic quirks being tested there (it may not even be valid, given that the blame for that line dates all the way back to January 2008).

lib/Pod/To/Text.rakumod
```diff
@@ -49,9 +49,9 @@ sub item2text($pod) {
 sub named2text($pod) {
     given $pod.name {
         when 'pod'  { pod2text($pod.contents)     }
-        when 'para' { para2text($pod.contents[0]) }
+        when 'para' { pod2text($pod.contents) }
```
lizmat, line 52: OOC, is para2text still referred to in other places?

raydiak: Actually no afaik, good catch. Though given the naming conventions here: https://github.com/rakudo/rakudo/blob/178a00f86fe830c09be338a7e3b6f2f55a7f475a/lib/Pod/To/Text.rakumod#L13-L31 para2text should probably be called from line 18, with the code currently there in para2text instead. If that looks right to you, I'll add a commit for it to this PR.

raydiak: I decided to take the silence as agreement :)

PR #4434, opened 2021-06-27
Add 'modf' to the Raku core's numerical routines

tbrowder:
This routine is commonly found in many popular
programming languages (including Python) but
was not included in the original Raku release.

The routine as added follows the basic implementation
as found in C and C++ with some slght modifications
to take advantage of existing Raku features.
As implemented in Raku routine 'modf' provides:

+ for any number input it returns a list of the integral and
  fractional parts
+ both parts retain the sign of the input number
+ if either part is zero, that part is coerced to an integer
  zero
+ the returned integral part is truly an integer (as opposed
  to some other languages such as C and C++)
+ an optional second parameter ('places') can be entered to
  force the fractional part to the desired number of
  decimal places (useful for testing)

### DIFF
--- a/src/core.c/Num.pm6
+++ b/src/core.c/Num.pm6
@@ -168,6 +168,29 @@ my class Num does Real { # declared in BOOTSTRAP
     multi method sin(Num:D: ) {
         nqp::p6box_n(nqp::sin_n(nqp::unbox_n(self)));
     }
+
+    proto method modf(|) {*} 
+    multi method modf(Num:D: Num $places? --> List) { # method source of the return values, patterned after sub log
+        my $x = self; 
+        my \sign = $x.sign;
+        $x .= abs;  # now signless
+
+        # operate on absolute parts
+        my $int  = $x.Int;  
+        my $frac = $x - $int;
+        # restore original sign
+        $int  *= sign; 
+        $frac *= sign; 
+        $int  = 0 if $int  == 0;
+        $frac = 0 if $frac == 0;
+
+        if $places.defined and $places > 0 {
+            $frac = sprintf '%.*f', $places, $frac;
+        }
+
+        $int, $frac
+    }
+
     proto method asin(|) {*}
     multi method asin(Num:D: ) {
         nqp::p6box_n(nqp::asin_n(nqp::unbox_n(self)));
@@ -541,6 +564,26 @@ multi sub acosec(Num:D \x) {
 
 multi sub log(num $x --> num) {
     nqp::log_n($x);
+}
+
+multi sub modf(num $x is copy, num $places? --> List) { # sub source of the return values, patterned after sub log
+        my \sign = $x.sign;
+        $x .= abs;  # now signless
+
+        # operate on absolute parts
+        my $int  = $x.Int;  
+        my $frac = $x - $int;
+        # restore original sign
+        $int  *= sign; 
+        $frac *= sign; 
+        $int  = 0 if $int  == 0;
+        $frac = 0 if $frac == 0;
+
+        if $places.defined and $places > 0 {
+            $frac = sprintf '%.*f', $places, $frac;
+        }
+
+        $int, $frac
 }
 
 multi sub sin(num $x --> num) {
--- a/src/core.c/Numeric.pm6
+++ b/src/core.c/Numeric.pm6
@@ -72,6 +72,12 @@ proto sub sign($, *%) is pure {*}
 multi sub sign(Numeric \x) { x.sign }
 multi sub sign(Cool \x)    { x.Numeric.sign }
 
+proto sub modf($, $?, *%) is pure {*}
+multi sub modf(Numeric $x --> List)                  { $x.modf                  }
+multi sub modf(Numeric $x, Numeric $places --> List) { $x.modf($places)         }
+multi sub modf(Cool $x --> List)                     { $x.Numeric.modf          }
+multi sub modf(Cool $x, Cool $places --> List)       { $x.Numeric.modf($places) }
+
 proto sub log($, $?, *%) is pure {*}
 multi sub log(Numeric $x) { $x.log }
 multi sub log(Numeric $x, Numeric $base) { $x.log($base) }
--- a/src/core.c/Cool.pm6
+++ b/src/core.c/Cool.pm6
@@ -40,6 +40,10 @@ my class Cool { # declared in BOOTSTRAP
     multi method log(Cool:D: )      { self.Numeric.log          }
     multi method log(Cool:D: $base) { self.Numeric.log($base.Numeric) }
 
+    proto method modf(|) {*}
+    multi method modf(Cool:D --> List)          { self.Numeric.modf                  }
+    multi method modf(Cool:D $places --> List)  { self.Numeric.modf($places.Numeric) }
+
     proto method exp(|) {*}
     multi method exp(Cool:D: )      { self.Numeric.exp          }
     multi method exp(Cool:D: $base) { self.Numeric.exp($base.Numeric) }
--- a/src/core.c/Real.pm6
+++ b/src/core.c/Real.pm6
@@ -53,6 +53,11 @@ my role Real does Numeric {
         Complex.new(self.cos, self.sin);
     }
     method Complex() { Complex.new(self.Num, 0e0) }
+
+    proto method modf(|) {*}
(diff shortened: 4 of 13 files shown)


tbrowder:
Tests for this PR are in roast PR #742.

melezhik:
I run some tests against this branch,  example 8 fails, is it bug? (I did not expect `18` in fraction part .... )

```
[root@2834387600a8 r3tool]# SP6_FORMAT_TERSE=1 tomty pull_4434
index updated from file:///root/repo/api/v1/index
cat $root_dir/task.bash

$RAKUBIN -v

$RAKUBIN -e 'say "ex1 ", modf(0.0).perl'

$RAKUBIN -e 'say "ex2 ", modf(-0.0).perl'

$RAKUBIN -e 'say "ex3 ", modf(0).perl'

$RAKUBIN -e 'say "ex4 ", modf(.6).perl'

$RAKUBIN -e 'say "ex5 ", modf(.6,2).perl'

$RAKUBIN -e 'say "ex6 ", modf(123456789123456789.6,2).perl'

$RAKUBIN -e 'say "ex7 ", modf(+6.2,2).perl'

$RAKUBIN -e 'say "ex8 ", modf(+6.2,100).perl'
Welcome to ������™ v2021.06-22-g1e96d09df.
Implementing the ����™ programming language v6.d.
Built on MoarVM version 2021.06-6-g6977d0591.
ex1 (0, 0)
ex2 (0, 0)
ex3 (0, 0)
ex4 (0, 0.6e0)
ex5 (0, "0.60")
ex6 (123456789123456784, "0.00")
ex7 (6, "0.20")
ex8 (6, "0.2000000000000001800000000000000000000000000000000000000000000000000000000000000000000000000000000000")
[task check] stdout match <ex1 (0, 0)> True
[task check] stdout match <ex2 (0, 0)> True
[task check] stdout match <ex3 (0, 0)> True
[task check] stdout match <ex4 (0, 0.6e0)> True
[task check] stdout match <ex5 (0, "0.60")> True
[task check] stdout match <ex6 (123456789123456784, "0.00")> True
[task check] the last example should contain .2 followed by 99 zeros
[task check] stdout match <ex8 (6,"0.2000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000\")> False
=================
TASK CHECK FAIL
```

tbrowder:
I appreciate the feedback, I'll check it out. Thanks.

melezhik:
Glad it helps. Oh yeah in ex6 integral part should be 123456789123456789, ? i missed that, need to fix my test ...

tbrowder:
I think there are some things I've missed in the guts of Raku. Like needing Rats and FatRats when numbers get so big. I think I may need some big gun devs to investigate if I can't figure it out fairly soon.

melezhik:
I changed tests as well:

```
[root@2834387600a8 r3tool]# SP6_FORMAT_TERSE=1 tomty pull_4434
index updated from file:///root/repo/api/v1/index
cat $root_dir/task.bash

$RAKUBIN -v

$RAKUBIN -e 'say "ex1 ", modf(0.0).perl'

$RAKUBIN -e 'say "ex2 ", modf(-0.0).perl'

$RAKUBIN -e 'say "ex3 ", modf(0).perl'

$RAKUBIN -e 'say "ex4 ", modf(.6).perl'

$RAKUBIN -e 'say "ex5 ", modf(.6,2).perl'

$RAKUBIN -e 'say "ex6 ", modf(123456789123456789.6,2).perl'

$RAKUBIN -e 'say "ex7 ", modf(+6.2,2).perl'

$RAKUBIN -e 'say "ex8 ", modf(+6.2,100).perl'

$RAKUBIN -e 'say "ex9 ", modf(+6.2,4).perl'

$RAKUBIN -e 'say "ex_A ", modf(+6.2,10).perl'
Welcome to ������™ v2021.06-22-g1e96d09df.
Implementing the ����™ programming language v6.d.
Built on MoarVM version 2021.06-6-g6977d0591.
ex1 (0, 0)
ex2 (0, 0)
ex3 (0, 0)
ex4 (0, 0.6e0)
ex5 (0, "0.60")
ex6 (123456789123456784, "0.00")
ex7 (6, "0.20")
ex8 (6, "0.2000000000000001800000000000000000000000000000000000000000000000000000000000000000000000000000000000")
ex9 (6, "0.2000")
ex_A (6, "0.2000000000")
[task check] stdout match <ex1 (0, 0)> True
[task check] stdout match <ex2 (0, 0)> True
[task check] stdout match <ex3 (0, 0)> True
[task check] stdout match <ex4 (0, 0.6e0)> True
[task check] stdout match <ex5 (0, "0.60")> True
[task check] stdout match <ex6 (123456789123456789, "0.60")> False
[task check] stdout match <ex7 (6, "0.20")> True
[task check] the last example should contain .2 followed by 99 zeros
[task check] stdout match <ex8 (6,"0.2000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000000\")> False
[task check] stdout match <ex9 (6, "0.2000")> True
[task check] stdout match <ex_A (6, "0.2000000000")> True
=================
TASK CHECK FAIL
```

tbrowder:
Okay, I've been able to fix those errors by converting the input numbers to FatRats in my modf module, so tomorrow I will try that on this PR.

tbrowder:
I have had no luck getting two values out of the modf routine in core. I'm sure I'm missing some kind of magic, but I've tried using Lists as a return and using named hashes in the parameter list. Taking a break for a while.

lizmat:
Why the many `proto`s ?  Why not have a single `proto` ?

src/core.c/Numeric.pm6
```diff
@@ -72,6 +72,12 @@ proto sub sign($, *%) is pure {*}
 multi sub sign(Numeric \x) { x.sign }
 multi sub sign(Cool \x)    { x.Numeric.sign }
 
+proto sub modf($, $?, *%) is pure {*}
+multi sub modf(Numeric $x --> List)                  { $x.modf                  }
```
lizmat, line 76: Problably want to add a type smiley on all types here.  Can't take a `modf` of a type object, can you?

src/core.c/Num.pm6
```diff
@@ -543,6 +566,26 @@ multi sub log(num $x --> num) {
     nqp::log_n($x);
 }
 
+multi sub modf(num $x is copy, num $places? --> List) { # sub source of the return values, patterned after sub log
```
lizmat, line 569: Why is this not simply deferring to the `.modf` version?

tbrowder: Because I was blindly following the example of the log example.

tbrowder: In that regard, is it better to implement a sub in terms of the better or vice versa?

In my own modules I have generally (if not always) defined the sub and used it in the corresponding class method.

src/core.c/Numeric.pm6
tbrowder: > Return two smileys in a List?

lizmat:
The idea of having a `sub foo(\x)` call `x.foo`, is to allow custom classes to just work with `sub foo` by adding a `foo` method.

````raku
class Foo {
    method log() { 3.14e0 }
}
my $x = Foo.new;
say log(x);  # 3.14
````

Now, for native types, that probably doesn't make any sense, as you cannot (easily anyway) create a new native type, or inherit from it.

````raku
class Foo is num { }
# ===SORRY!=== Error while compiling -e
# num does not support inheritance, so Foo cannot inherit from it
````

Although it *is* possible to dispatch on a native num, this appears to be very bad for performance (about 7x as slow).  So that's why the native sub versions do the thing "locally" without dispatching to the method.

MasterDuke17:
@jnthn will this be any changed with new-disp? And if so, think it will be fast enough that we could switch to using the sub-calls-method paradigm with natives?

tbrowder:
This PR I think needs a rewrite. I will close it and resubmit after I get my act together better. I am closing the asociated roast PR also (roast PR #742).

src/core.c/Numeric.pm6
vrurg: @lizmat what about "use of undefined in numeric context" warning?

lizmat: Ah, I see.  Hmmm....  I think the reason the multis for math ops are set up this way, is because they precede the existence of type smileys.

OK, skip the type smileys for now, that is a more general issue :-)

PR #4436, opened 2021-07-03
Handle updates to curried role typecheck lists more lazily

Kaiepi:
The following throws upon composition of `Bar`:

```raku
role Foo[Any ::T]         { }
role Bar[::T] does Foo[T] { }
```

This happens because in order to compose `Bar`'s role typecheck list, it needs to know `Foo`'s as well, but this completes the parameterization of `Foo` before `T` has a chance to get instantiated, making it a `Mu` during
selection of a candidate for `Foo`.

`Metamodel::CurriedRoleHOW.complete_parameterization` and related metamethods assume that the role typecheck list needs to be updated at the same time. These are the only places the `update_role_typecheck_list` metamethod is called, implying the `role_typecheck_list` getter that depends on it needs to update this beforehand for one reason or another, which makes sense; adding roles as late as it does should invalidate any typecheck list cached. If completing the parameterization here is problematic, then defer it by allowing the cache be cleared and updating it when need be.

Should pass `make test` and `make spectest` . Roast's `S32-io/encode.t` fails here for me, but on master as well.

### DIFF
--- a/src/Perl6/Metamodel/CurriedRoleHOW.nqp
+++ b/src/Perl6/Metamodel/CurriedRoleHOW.nqp
@@ -96,21 +96,24 @@ class Perl6::Metamodel::CurriedRoleHOW
                 my @result := $candidate-how.body_block($!candidate)(|@pos_args, |%!named_args);
                 $type_env := @result[1];
             }
-            for $candidate-how.roles($!candidate, :!transitive) -> $role {
-                if $role.HOW.archetypes.generic && $type_env {
-                    $role := $role.HOW.instantiate_generic($role, $type_env);
+            if $candidate-how.roles($!candidate, :!transitive) -> @roles {
+                for @roles -> $role {
+                    if $role.HOW.archetypes.generic && $type_env {
+                        $role := $role.HOW.instantiate_generic($role, $type_env);
+                    }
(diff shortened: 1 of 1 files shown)


Kaiepi:
The new `clear_role_typecheck_role_list` metamethod may not be necessary. Testing.

vrurg:
It feels like method `roles` must no use `complete_parameterization` too.

Kaiepi:
Right. I don't know if the MOP alone can do this, but traits are applied early enough to trigger similar errors in `roles`:

```raku
multi sub trait_mod:<is>(Mu:U \T, :$throwing-now!) {
    T.^roles[0].^roles
}

role Foo[Any ::A]                         { }
role Bar[::A] does Foo[A] is throwing-now { }
```

vrurg (approved):
So elegant!

vrurg:
Do you also have roast PR for this?

Kaiepi:
Not at the moment. I'll get started on one tomorrow.

lizmat:
> Roast's S32-str/encode.t fails here for me, but on master as well.

After reverting 1ed28638909feb1a444ec0156fe5fa8b27fc9922 , it passes for me.  So I guess the assertion that it failed on master for you as well, was incorrect  :-(

Issue #1366, opened 2018-01-05
Proc.status should be a private method


moritz:
In perl6/doc#1710 we've discussed whether `.status` should be documented, but I think it's not really part of the public API. I propose we make it private.

My plan is to add a private method as a copy, and add a deprecation notice to the public one.

Altai-man:
As the deprecation has happened, this can be closed and eventually the code will be routinely deleted in a year from now.

PR #4442, opened 2021-07-06
Deprecate `status` method of `Proc` in place of exitcode` and `signal`

Altai-man:
Does what https://github.com/rakudo/rakudo/issues/1366 asked for 3 years ago (wow).


lizmat:
Generally, traits ons callables (e.g. `is export`) only need to be done on the proto, if they are to be applied on all candidates.

Only when you want to e.g. deprecate a specific candidate, should you apply that to that candidate.

PR #4449, opened 2021-07-10
Add Associative HyperWhatever support

gfldex:
I took the lack of complaint in https://github.com/Raku/problem-solving/issues/287 as silent consent and implemented the first part in v6.e. Support for @a[**] could be part 2.

Tests can be found in https://gist.github.com/381b402a5356bd97fb2ff7194a25dd96


t/02-rakudo/03-corekeys.t
```diff
@@ -794,6 +794,7 @@ my @allowed =
         Q{$¢},
         Q{&postcircumfix:<[; ]>},
         Q{&postcircumfix:<{; }>},
+        Q{&postcircumfix:<{ }>},
```
lizmat, line 797: I'm surprised this is needed?

gfldex: Without it the test fails. And there is a new symbol in CORE::v6e so it is not complaining wrongly.

lizmat:
Sorry it had to take this long!

PR #4450, opened 2021-07-11
Allow openbsd to locate shared libraries by version

casaca24:


Kaiepi:
Related to 5603128e019a9d40df55e718196cb10aa13b02b4.

casaca24:
thank you for taking the time to review this. i'll keep the patch locally. this allowed NativeLibs tests to pass, but i guess it causes more harm than good.

niner:
I think this PR is the better way to go in the long term. Honestly, the commit @Kaiepi pointed out does not do a great job at explaining the reason for ignoring versions on OpenBSD. Just ignoring versions also leaves no way to deal with situations where only a versioned library is in the search path (which at least on Linux is the default case!). Obviously this bites users as well. Maybe just other users in other situations. With this PR (i.e. reverting the workaround), we're following the intent of the programmer and give them the option to have a fallback to an unversioned library. It also increases pressure on authors to get their version story in order. If multiple versions are OK, that should better be stated explicitly in the module.

lizmat:
So, should we merge this?

PR #4452, opened 2021-07-13
Fix CURI probing on Windows

patrickbkr:
libUVs `stat` returns unusable writability information on Windows. It
can't be trusted for actual writeability. This means `IO::Path.w` is
just as unusable. Thus in `CURI.can-install()` on Windows we now
actually probe for writeability by trying to create a file.

This fixes installing things on Windows with Rakudo being installed
in "C:\Program Files\Rakudo".

LibUVs comments on `stat` on Windows: https://github.com/libuv/libuv/blob/v1.x/src/win/fs.c#L1731:L1749

### DIFF
--- a/src/core.c/CompUnit/Repository/Installation.pm6
+++ b/src/core.c/CompUnit/Repository/Installation.pm6
@@ -9,6 +9,7 @@ class CompUnit::Repository::Installation does CompUnit::Repository::Locally does
     has %!dist-metas;
     has $!precomp-stores;
     has $!precomp-store;
+    has $!prefix-writeable-cache;
 
     my $verbose = nqp::getenvhash<RAKUDO_LOG_PRECOMP>;
 
@@ -23,8 +24,28 @@ class CompUnit::Repository::Installation does CompUnit::Repository::Locally does
         }
     }
 
+    method !prefix-writeable {
+            if Rakudo::Internals.IS-WIN {
+                $!prefix-writeable-cache //= do {
+                    my $writable = False;
+                    try {
+                        my $check-file = $.prefix.add('test-file');
+                        if my $handle = $check-file.open(:create, :w) {
+                            $handle.close;
+                            $check-file.unlink;
+                            $writable = True;
+                        }
+                    }
+                    $writable
+                }
+            }
+            else {
+                $.prefix.w;
+            }
+    }
+
     method writeable-path {
-        $.prefix.w ?? $.prefix !! IO::Path;
+        self!prefix-writeable ?? $.prefix !! IO::Path;
     }
 
     method !writeable-path {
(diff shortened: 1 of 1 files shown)


ugexe:
Are you sure its not actual permission issues?
```
C:\Program Files>raku -e "say $*CWD.w"
False

C:\Program Files>mkdir foo
Access is denied.
```
```
C:\Users\ugexe\foo bar>raku -e "say $*CWD.w"
True
```

patrickbkr:
Can you try

    raku -e "say 'C:\Windows'.IO.w; 'C:\Windows\Non-existing.txt'.IO.spurt('foo')"

patrickbkr:
To quote libuvs documentation:

       Currently it's based on whether the 'readonly' attribute is set, which
       makes little sense because the semantics are so different: the 'read-only'
       flag is just a way for a user to protect against accidental deletion, and
       serves no security purpose. Windows uses ACLs for that.

src/core.c/CompUnit/Repository/Installation.pm6
```diff
@@ -23,16 +23,33 @@ class CompUnit::Repository::Installation does CompUnit::Repository::Locally does
         }
     }
 
+    method !prefix-writeable {
+        if Rakudo::Internals.IS-WIN {
```
ugexe, line 27: I think it might be ok to cache the result of the windows condition. It adds an edge case when the repo e.g. becomes unwritable when it was writable moments earlier, but it potentially saves enough file creation attempts in the 99%+ cases to be worth it.

src/core.c/CompUnit/Repository/Installation.pm6
```diff
@@ -23,16 +23,36 @@ class CompUnit::Repository::Installation does CompUnit::Repository::Locally does
         }
     }
 
+    method !prefix-writeable {
+        state $writable = False;
+        once {
```
ugexe, line 28: Two things: first I would only cache the result on windows... checking file existence the normal way (by not creating a file) is fine to do repeatedly. Second I would not use `state` or `once` to do this because they are not thread safe and may return undef when a second thread reaches it before the first thread has set the value.

ugexe: `has $!prefix-writable-cache;` + `$!prefix-writable-cache //= do { ... }` is a thread safe way to cache the value (the previously mentioned race condition would instead just write the file again instead of getting back undef)

ugexe: (fwiw similar changes need to be made to zef for its various probes)

patrickbkr: What's a good alternative? Saving to an attribute and returning that when it's defined?

ugexe: yes, something like `has $!prefix-writable-cache;` + `$!prefix-writable-cache //= do { ... }`

ugexe (approved):

PR #4455, opened 2021-07-16
Create checksums for release files

patrickbkr:
Checksums for release files have been requested at least once. I consider them good form.

<del>This PR adapts the release docs and `Akefile`, so that two additional files for each file is generated:

- <del>a `filename.checksums.txt` containing checksums
- <del>a `filename.checksums.txt.asc` containing checksums and also a signature (e.g. Fedora provides checksums in this format)

*Edit:* I made up my mind. I think there is no justification for providing a separate checksums and signed checksums file. I think there is no thing you can do with the non signed file that one couldn't do with the signed one. So this PR now only generates one additional file, a `somefile.checksums.txt` which contains checksums and an inline signature.

In general I'd like for us to not change the format again once we have decided on one as checksum checking tends to be a thing people like to automate. Thus I'd like to discuss the alternatives.

Providing separate checksum file for each file produces a lot of files, but that's how the Star Bundle releases are currently doing it. The obvious alternative being having a single checksum file for all files of each release.

See https://rakudo.org/downloads/star for the files the Star Bundle currently provides.
See https://getfedora.org/en/security/ to take a peek at Fedoras checksuming.
See https://cloudflare.cdn.openbsd.org/pub/OpenBSD/6.9/ for how OpenBSD does it.

I don't have a good overview of the landscape. Is there a dominant format most projects have settled on?

### DIFF
--- a/docs/release-guide-binary.md
+++ b/docs/release-guide-binary.md
@@ -19,14 +19,25 @@ The process of building is not started automatically, but has to be triggered ma
 - Verify all three archives have a size >10MB.
 - Verify all three archives decompress successfully.
 
-Sign the files:
+Sign the files and create checksum files:
 
-   gpg -b -u your@gpgkey.org --armor rakudo-moar-2020.01-01-linux-x86_64.tar.gz
-   gpg -b -u your@gpgkey.org --armor rakudo-moar-2020.01-01-macos-x86_64.tar.gz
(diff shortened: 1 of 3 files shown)


patrickbkr:
Ping @niner @Altai-man

Altai-man (approved):

PR #4470, opened 2021-07-20
Add Iterable cmp Iterable candidates

lizmat:
If either side of an `infix:<cmp>` is an `Iterable`, then use the iterator
on both sides to determine the result, rather than stringifying the
Iterable(s).  This is, I think, more correct, and much more memory
friendly as there is not `Str` representation created of the argument.

This change is spectest error free.  It however changes the outcome
of:

    dd "a b" cmp ("a","b")

from: `Order::Same` (which I consider to be incorrect), to `Order::More`.

The `is default` is needed to prevent ambiguous dispatch errors with
the `Any` / `Real` and `Real` / `Any` candidates.

### DIFF
--- a/src/core.c/Order.pm6
+++ b/src/core.c/Order.pm6
@@ -16,6 +16,15 @@ multi sub infix:<cmp>(\a, \b) {
       ?? Same
       !! a.Stringy cmp b.Stringy
 }
+multi sub infix:<cmp>(Iterable:D \a, Iterable:D \b) {
+    a.iterator cmp b.iterator
+}
+multi sub infix:<cmp>(Iterable:D \a, \b) is default {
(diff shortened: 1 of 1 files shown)


gfldex:
Since `"a b" ~~ <a b>`  is false, I expect `cmp` not to be `Order::Same`. Which one of the two is `More`, I'm not so sure.

With `Iterables` being acutally iterated, I wonder how LoLs and HaHs should be handled.

gfldex:
Thinking about this a little more, I believe that stringification was deliberate, because the decision, if comparing apples with peaches, is then offloaded to the Unicode Consortium. We don't got `Order::Dunno`. One could `fail` ofc.

PR #4472, opened 2021-07-21
Allow IO::Path.add to take multiple values

lizmat:
So that "foo".IO.add(<bar baz>) yields an "foo/bar/baz" path.

### DIFF
--- a/src/core.c/IO/Path.pm6
+++ b/src/core.c/IO/Path.pm6
@@ -377,9 +377,9 @@ my class IO::Path is Cool does IO {
           $!SPEC.join('', $!path, child.Str)
     }
 
-    method add (IO::Path:D: \child) {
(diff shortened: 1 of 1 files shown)


src/core.c/IO/Path.pm6
```diff
@@ -377,9 +377,9 @@ my class IO::Path is Cool does IO {
           $!SPEC.join('', $!path, child.Str)
     }
 
-    method add (IO::Path:D: \child) {
+    method add (IO::Path:D: *@children) {
```
jnthn, line 380: Wonder if it's worth a pair of `multi`s given the common case is just one child to add, and we can save an amount of work?

jnthn (approved):
Seems like a convenient thing to provide.

lizmat:
Should we lift this to 6.e ?

PR #4473, opened 2021-07-21
Add Cool.Order "coercer"

lizmat:
Convert a Cool value to the associated Order enum:
- < 0  -> Less
-   0  -> Same
- \> 0  -> More

The only other way to currently do this, is by basically writing the
ORDER (implementation-detail) subroutine again, or just do a call to
ORDER.  This allows syntactic sugar to not have to do it like that.

### DIFF
--- a/src/core.c/Cool.pm6
+++ b/src/core.c/Cool.pm6
@@ -40,6 +40,8 @@ my class Cool { # declared in BOOTSTRAP
     multi method log(Cool:D: )      { self.Numeric.log          }
     multi method log(Cool:D: $base) { self.Numeric.log($base.Numeric) }
 
+    method Order(Cool:D:) { ORDER(self.Int) }
+
     proto method exp(|) {*}
     multi method exp(Cool:D: )      { self.Numeric.exp          }
     multi method exp(Cool:D: $base) { self.Numeric.exp($base.Numeric) }


raiph:
I suspect I've got the wrong end of the stick, but hopefully you'll forgive me for writing nonsense if that's what the following is. :)

First, for a value `foo`, wouldn't `foo <=> 0` be the same operation?

Second, if so, but a method is desired for brevity/clarity, is `Cool` more compelling as a home for it than `Numeric`? Aiui `Cool` is meant to contain routines that coerce from string to number or vice-versa, or from scalar to list or vice-versa, when such routines are not so conveniently declared in their more specific classes. So perhaps it's more convenient to put it in `Cool` than ``Numeric`. But somehow it just doesn't seem right to have the routine you describe in `Cool`, given that it feels like `Order`, in the context of `Cool` (as against `Numeric`) ought by rights be agnostic about the order being numeric or lexicographic.

Third, wouldn't a method name like `.cmp0` be, er, in order, to make the three things it's doing clear, namely coercing to numeric, doing a `cmp`, and comparing with zero?

lizmat:
> First, for a value foo, wouldn't foo <=> 0 be the same operation?

It would.

> Second, if so, but a method is desired for brevity/clarity, is Cool more compelling as a home for it than Numeric? Aiui Cool is meant to contain routines that coerce from string to number or vice-versa, or from scalar to list or vice-versa, when such routines are not so conveniently declared in their more specific classes. So perhaps it's more convenient to put it in Cool than ``Numeric. But somehow it just doesn't seem right to have the routine you describe in Cool`, given that it feels like `Order`, in the context of `Cool` (as against `Numeric`) ought by rights be agnostic about the order being numeric or lexicographic.

The reason for putting it in `Cool`, is to make `"42".Order` work as well.  And have it throw for non-numeric values.

> Third, wouldn't a method name like .cmp0 be, er, in order, to make the three things it's doing clear, namely coercing to numeric, doing a cmp, and comparing with zero?

The reason I called it `.Order` is because it produces an `Order` enum.

raiph:
OK. I suspect the name `.cmp0` would be better, but accept the merit of `Order`.

----

For the record:

>> First, for a value `foo`, wouldn't `foo <=> 0` be the same operation?

> It would.

OK. So `foo cmp 0` has the advantage it would be explicit about what it's doing. But the disadvantage of being noisier. And the advantage it eliminates method call overhead. But compiler optimization and/or macroization can render that advantage moot.

> The reason for putting it in `Cool`, is to make `"42".Order` work as well. And have it throw for non-numeric values.

OK. Then I *did* understand what you intend, and hopefully you understand why I found it odd, and suspect others might too.

(Because `Cool` groups together operations related to numbers as strings and strings as numbers, and the `Order` enum is agnostic whether it's used in relation to lexicographic or numeric order. So why is `.Order` numeric all of a sudden? And the answer is because it's useful that way, whereas `.Order` returning an `Order` wouldn't mean anything useful when applied to a string because there's no equivalent of a natural mid point to compare against.)

PR #4474, opened 2021-07-21
Use 'inlinable' instead of 'inlineable'

MasterDuke17:


lizmat:
OOC, apart from the spelling, is there another reason for this change?

MasterDuke17:
Not really. But I was grepping through the repos, didn't find what I was looking for, and then realized I had to search for both values so thought it would be good to make it consistent. Wiktionary calls the version with 'e' an alternative. However, while it's slightly less common in NQP and Rakudo, it's quite a bit more common in MoarVM. Most annoyingly, we have public(ish) APIs for all with different spellings. E.g., `is inlinable` in Rakudo, but `StorageSpec` in MoarVM and NQP's JVM backend has a public attribute `inlineable`.

PR #4478, opened 2021-07-26
Introduce curried role binding

Kaiepi:
As [before](https://github.com/rakudo/rakudo/pull/4436), the following should be valid:

```raku
role Foo[Any ::T] { }
role Bar[::T] does Foo[T] { }
```

But we now also consider:

```raku
role Foo[Any ::T] { }
role Bar[::T] is Foo[T] { }
```

And:

```raku 
role OfIndex[Enumeration ::T] does T { }
```

Curries of parametric role groups already have a notion of a binding in their cached `$!candidate`. Formalizing this via new `bind` metamethods on `Metamodel::CurriedRoleHOW` and `Metamodel::ParametricRoleGroupHOW` makes it possible to fix several issues with role typechecking because if a binding is allowed to fail, it follows operations pertaining to them can follow suit.

`Metamodel::CurriedRoleHOW.bind` defers to its role's `bind` metamethod (if any, in which case it accepts the curry it's invoked from in addition to a metaobject) to produce a final parameterization:

```
bastille% raku -e 'say Blob[uint8].^bind =:= Blob.^bind: Blob[uint8]'
True
bastille% raku -e 'say Blob[uint8].^bind.HOW.^name'
Perl6::Metamodel::ParametricRoleHOW
```

There should be opportunities for optimizations depending on this here. While this passes all tests locally, this is a draft because it has changes reminiscent of those for subsets that were too breaking for v6.c. The effect of any revisions to this on performance should be dampened by the optimizations made now, but aren't made at this point to avoid making them in paranoia.

### DIFF
--- a/src/Perl6/Metamodel/CurriedRoleHOW.nqp
+++ b/src/Perl6/Metamodel/CurriedRoleHOW.nqp
@@ -182,6 +182,10 @@ class Perl6::Metamodel::CurriedRoleHOW
         @!pos_args
     }
 
+    method role_named_arguments($obj) {
+        %!named_args
+    }
+
     method roles($obj, :$transitive = 1, :$mro = 0) {
         self.complete_parameterization($obj);
         self.roles-ordered($obj, self.roles_to_compose($obj), :$transitive, :$mro)
--- a/src/Perl6/Metamodel/CurriedRoleHOW.nqp
+++ b/src/Perl6/Metamodel/CurriedRoleHOW.nqp
@@ -24,7 +24,7 @@ class Perl6::Metamodel::CurriedRoleHOW
     does Perl6::Metamodel::InvocationProtocol
 {
     has $!curried_role;
-    has $!candidate;                # Will contain matching candidate from curried role group
+    has $!binding;
     has @!pos_args;
     has %!named_args;
     has @!role_typecheck_list;
@@ -79,24 +79,38 @@ class Perl6::Metamodel::CurriedRoleHOW
         nqp::settypecheckmode($type, 2);
     }
 
-    method parameterize_roles($obj) {
-        my @pos_args;
-        nqp::push(@pos_args, $obj);
-        for @!pos_args {
-            nqp::push(@pos_args, $_);
+    # A curried role may be bound to form a role other than its origin, e.g.  a
+    # parametric role group and the candidate we're not quite ready to fetch at
+    # construction time. If we're still not ready yet whenever we need it, the
+    # binding is ourself for now. If the origin lacks a means of producing a
+    # binding, we assume it's its own by default.
+    method bind($obj) {
+        if $!binding =:= NQPMu {
+            if nqp::can($!curried_role.HOW, 'bind') {
+                my $binding := $!curried_role.HOW.bind($!curried_role, $obj);
+                $!binding := $binding unless nqp::decont($binding) =:= $obj;
+                $binding
+            }
+            else {
+                $!binding := $!curried_role
+            }
         }
-        if nqp::istype($!curried_role.HOW, Perl6::Metamodel::ParametricRoleGroupHOW) {
-            $!candidate := $!curried_role.HOW.select_candidate($!curried_role, @pos_args, %!named_args);
-            my $candidate-how := $!candidate.HOW;
+        else {
+            $!binding
+        }
+    }
 
-            self.set_language_revision($obj, $candidate-how.language-revision($!candidate));
+    method parameterize_roles($obj) {
+        my $binding := self.bind($obj);
+        unless nqp::decont($binding) =:= $obj {
+            self.set_language_revision($obj, $binding.HOW.language-revision($binding));
 
             my $type_env;
             try {
-                my @result := $candidate-how.body_block($!candidate)(|@pos_args, |%!named_args);
+                my @result := $binding.HOW.body_block($binding)($obj, |@!pos_args, |%!named_args);
                 $type_env := @result[1];
             }
-            for $candidate-how.roles($!candidate, :!transitive) -> $role {
+            for $binding.HOW.roles($binding, :!transitive) -> $role {
                 if $role.HOW.archetypes.generic && $type_env {
                     $role := $role.HOW.instantiate_generic($role, $type_env);
                 }
@@ -115,7 +129,7 @@ class Perl6::Metamodel::CurriedRoleHOW
             # Contrary to roles, we only consider generic parents. I.e. cases like:
             # role R[::T] is T {}
             if $type_env {
-                for $candidate-how.parents($!candidate, :local) -> $parent {
+                for $binding.HOW.parents($binding, :local) -> $parent {
                     if $parent.HOW.archetypes.generic {
                         my $ins := $parent.HOW.instantiate_generic($parent, $type_env);
                         nqp::push(@!parent_typecheck_list, $ins)
@@ -210,9 +224,6 @@ class Perl6::Metamodel::CurriedRoleHOW
             }
         }
         self.complete_parameterization($obj) unless $!is_complete;
-        if !($!candidate =:= NQPMu) && $!candidate.HOW.type_check_parents($!candidate, $decont) {
-            return 1
-        }
         for @!parent_typecheck_list -> $parent {
             if nqp::istype($decont, $parent) {
                 return 1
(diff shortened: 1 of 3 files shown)


Kaiepi:
The tail end of this needs more work.

src/Perl6/Metamodel/CurriedRoleHOW.nqp
```diff
@@ -79,77 +79,94 @@ class Perl6::Metamodel::CurriedRoleHOW
...
+    # A curried role may be bound to form a role other than its origin, e.g.  a
+    # parametric role group and the candidate we're not quite ready to fetch at
+    # construction time. If we're still not ready yet whenever we need it, the
+    # binding is ourself for now. If the origin lacks a means of producing a
+    # binding, we assume it's its own by default.
+    method bind($obj) {
+        if $!binding =:= NQPMu {
+            if nqp::can($!curried_role.HOW, 'bind') {
+                my $binding := $!curried_role.HOW.bind($!curried_role, $obj);
+                $!binding := $binding unless nqp::decont($binding) =:= $obj;
```
vrurg, line 91: If you ever expect `bind` to return a container (which is, perhaps, expected for Raku-land implementations of the method), then deconting must take place the line above or you end up with a container in `$!binding`.

src/Perl6/Metamodel/CurriedRoleHOW.nqp
```diff
@@ -79,77 +79,94 @@ class Perl6::Metamodel::CurriedRoleHOW
...
+    }
 
-            self.set_language_revision($obj, $candidate-how.language-revision($!candidate));
+    method parameterize_roles($obj) {
+        if self.archetypes.generic {
+            $!is_complete := 1;
+        }
+        else {
+            my $binding := self.bind($obj);
+            unless nqp::decont($binding) =:= $obj {
```
vrurg, line 109: Same as above, better assume that deconting is not needed.

src/Perl6/Metamodel/CurriedRoleHOW.nqp
```diff
@@ -79,77 +79,94 @@ class Perl6::Metamodel::CurriedRoleHOW
...
-        for @!pos_args {
-            nqp::push(@pos_args, $_);
+    # A curried role may be bound to form a role other than its origin, e.g.  a
+    # parametric role group and the candidate we're not quite ready to fetch at
+    # construction time. If we're still not ready yet whenever we need it, the
+    # binding is ourself for now. If the origin lacks a means of producing a
+    # binding, we assume it's its own by default.
+    method bind($obj) {
+        if $!binding =:= NQPMu {
+            if nqp::can($!curried_role.HOW, 'bind') {
```
vrurg, line 89: I'd rather make method `bind` mandatory on all role HOWs, except for `ConcreteRoleHOW` which is impossible candidate for currying. Then this condition and it's `else` block can be removed. `bind` on `ParametricRoleHOW` would then simply return itself.

src/Perl6/Metamodel/CurriedRoleHOW.nqp
```diff
@@ -79,77 +79,94 @@ class Perl6::Metamodel::CurriedRoleHOW
...
-            my $candidate-how := $!candidate.HOW;
+        else {
+            $!binding
+        }
+    }
 
-            self.set_language_revision($obj, $candidate-how.language-revision($!candidate));
+    method parameterize_roles($obj) {
+        if self.archetypes.generic {
+            $!is_complete := 1;
```
vrurg, line 105: It is kind of a good idea for optimization, but it breaks typechecking of partially parameterized roles:

```raku
role R1[::T] { }
role R2[::T1, ::T2] does R1[::T1] { } 
role R3[::T] does R2[Int, T] { }
say R3.^candidates.head ~~ R1[Int];
```

This is _True_ on the current master but _False_ on your branch. Perhaps this makes sense when all arguments are generic, but what if there are too many of them?

src/Perl6/Metamodel/CurriedRoleHOW.nqp
```diff
@@ -79,77 +79,94 @@ class Perl6::Metamodel::CurriedRoleHOW
...
-                    my $target-name := $obj.HOW.name($obj);
-                    my $role-name := $role.HOW.name($role);
-                    Perl6::Metamodel::Configuration.throw_or_die(
-                        'X::Composition::NotComposable',
-                        $role-name ~ " is not composable, so " ~ $target-name ~ " cannot compose it",
-                        :$target-name,
-                        composer => $role,
-                    )
+                if $error {
+                    # XXX: Could use a wrapper exception.
```
vrurg, line 120: _Must_ use. We must depart from using anonymous adhocs.

src/Perl6/Metamodel/CurriedRoleHOW.nqp
```diff
@@ -205,43 +226,43 @@ class Perl6::Metamodel::CurriedRoleHOW
...
     }
 
     method accepts_type($obj, $checkee) {
-        # First, we locate candidate curryings to check against. If
-        # the checkee is itself a curried role, it also goes in. Note
-        # that we only want those that have the same parametric role
-        # as us.
+        # Generics' incapacity for typechecking propagates.
+        if self.archetypes.generic {
+            return 0;
```
vrurg, line 246: Same note about partially parameterized cases.

src/Perl6/Metamodel/ParametricRoleGroupHOW.nqp
```diff
@@ -90,6 +90,20 @@ class Perl6::Metamodel::ParametricRoleGroupHOW
         $curried
     }
 
+    method bind($obj, $curried) {
+        my $candidate;
+        my %named_args := $curried.HOW.role_named_arguments($curried);
+        my @pos_args := [$obj];
+        nqp::splice(@pos_args, $curried.HOW.role_arguments($curried), +@pos_args, 0);
```
vrurg, line 97: BTW, why do you use `nqp::splice` instead of `nqp::push`?

Kaiepi: Splicing at the end of an array like I do here seems to run faster than pushing from a `for`. Elsewhere, there's usually a need to map over the arguments, in which case I'd expect the loop to be more efficient instead.

src/Perl6/Metamodel/CurriedRoleHOW.nqp
Kaiepi: I thought the overhead of the identity binding for `Metamodel::ParametricRoleHOW` would be greater than the `nqp::can`s, but that doesn't seem to be the case.

src/Perl6/Metamodel/ParametricRoleGroupHOW.nqp
vrurg: Makes sense. I forgot that `push` is for single items only.

Kaiepi:
Bar the ad-hoc error, the issues brought up so far should be addressed. The strategy to optimizing role typechecking is now to eliminate the recursion from the doee typechecks. If that's not reasonable to do, some of the performance gains might be salvageable with fast/slow paths for doee typechecking.

The following should also be taken into consideration now, but aren't at this point:

```raku
enum Index <A B C>;

class Foo does Index { }
Foo ~~ Index;
```

```raku
class Foo does Buf[uint8] { }
Foo ~~ Buf;
Foo ~~ Buf[uint8];

class Bar does Buf { }
Bar ~~ Buf;
Bar ~~ Buf[uint8];
```

These should share a common theme in their solution.

src/Perl6/Metamodel/CurriedRoleHOW.nqp
```diff
@@ -79,28 +79,36 @@ class Perl6::Metamodel::CurriedRoleHOW
...
-        nqp::push(@pos_args, $obj);
-        for @!pos_args {
-            nqp::push(@pos_args, $_);
+    # A curried role may be bound to form a role other than its origin, e.g.  a
+    # parametric role group and the candidate we're not quite ready to fetch at
+    # construction time. If we're still not ready yet whenever we need it, the
+    # binding is ourself for now.
+    method bind($obj) {
+        if $!binding =:= NQPMu {
+            my $binding := nqp::decont($!curried_role.HOW.bind($!curried_role, $obj));
```
vrurg, line 88: I see two inconsistencies here. 

First, you still use deconting with `bind`, but it's the only place where you do so. I think you can safely throw it away.

The other one worries me more. Your `bind` methods has different signatures. It means that the code invoking the method must know exactly what kind of role it currently works with. Would it always be the case? I wouldn't rely on this. So, the second parameter must be introduced but made optional where it's not used.

Kaiepi:
Closing in favour of https://github.com/rakudo/rakudo/pull/4478.

PR #4479, opened 2021-07-27
Enhance signature-signature typechecking

Kaiepi:
`Signature.ACCEPTS(Signature:D: Signature:D)` has been a pet peeve of mine for a while now because it seemed repetitive, but I couldn't make it any more efficient. There should only be a need to iterate over arguments of a signature and a topic signature once in most cases. With a refactor of `ACCEPTS` to iterate more sparingly, the following can become valid more readily:

```raku
:(:a($x), :b($y)) !~~ :(:a(:b(:c($x))))
```

All cases should get improved performance, but this isn't the sort of algorithm I can give a sweeping "x% faster":

- The following scenario becomes ~9x faster because it has predictable positional parameter placement:

```raku
:($a, $b, $c) ~~ :($a, $b, $c)
```

- The following becomes ~6x faster because positional slurpy and named parameters take more work to typecheck:

```raku
:($a, Int:D $b, :$c = :$b, :$d) ~~ :($a, :$c, +, *%)
```

- The following becomes ~9x faster again because typechecks fail earlier in all cases:

```raku
:($a, Int:D $b, :$c = $b, Int:D :$d!) ~~ :($a, :$b, +, *%)
```

### DIFF
--- a/src/core.c/Signature.pm6
+++ b/src/core.c/Signature.pm6
@@ -32,64 +32,78 @@ my class Signature { # declared in BOOTSTRAP
         nqp::hllbool(nqp::p6isbindable(self, nqp::decont($topic)));
     }
     multi method ACCEPTS(Signature:D: Signature:D $topic) {
-        my $sclass = self.params.classify({.named});
-        my $tclass = $topic.params.classify({.named});
-        my @spos := $sclass{False} // ();
-        my @tpos := $tclass{False} // ();
+        my @r-params := self.params;
+        my @l-params := $topic.params;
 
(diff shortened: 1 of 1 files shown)


Kaiepi:
Benchmarks in the OP are skewed because I was running them too many threads for the machine I was borrowing. The first scenario should run ~10x faster, the second and third ~8.5x faster.

lizmat (approved):

lizmat:
After merge, I will look at further optimizing it, unless someone beats me to it.

PR #4482, opened 2021-07-28
Fix a race in ClassHOW new_type

vrurg:
Anonymous class ID is storead in a class `my` variable and was incremented by `new_type` method without a lock protection. In my testing the probablity of creating a clashing name was ~0.25% for concurrently invoked `new_type` methods.

### DIFF
--- a/src/Perl6/Metamodel/ClassHOW.nqp
+++ b/src/Perl6/Metamodel/ClassHOW.nqp
@@ -43,6 +43,7 @@ class Perl6::Metamodel::ClassHOW
         nqp::findmethod(NQPMu, 'BUILDALL')(nqp::create(self), |%named)
     }
 
+    my $id_lock := NQPLock.new;
     my $anon_id := 1;
     method new_type(:$name, :$repr = 'P6opaque', :$ver, :$auth, :$api, :$is_mixin) {
         my $metaclass := self.new();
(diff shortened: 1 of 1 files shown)


src/Perl6/Metamodel/ClassHOW.nqp
```diff
@@ -54,7 +55,9 @@ class Perl6::Metamodel::ClassHOW
             $new_type := nqp::newtype($metaclass, $repr);
         }
         my $obj := nqp::settypehll($new_type, 'Raku');
-        $metaclass.set_name($obj, $name // "<anon|{$anon_id++}>");
+        $id_lock.protect: {
+            $metaclass.set_name($obj, $name // "<anon|{$anon_id++}>");
+        }
```
niner, line 60: Probably won't make a difference, but you could improve parallelization by only protecting `$anon_id++` like in https://github.com/rakudo/rakudo/blob/22f78f4e6c87dcff25b5ce971b1d354a6cf0fc1f/src/Perl6/Grammar.nqp#L484

niner (approved):

src/Perl6/Metamodel/ClassHOW.nqp
vrurg: I thought about it, but how much sense does it make? The current way is more readable.

In either case, `nqp::cas_i` would be even better here, but NQP seemingly doesn't support it.

niner (approved):

PR #4484, opened 2021-07-29
Make .sort aware of any &[cmp] additions

lizmat:
This makes .sort aware of any local &[cmp] additions from where .sort
is called, if called without an explicit comparator.

This will use CALLERS:: to find out if the outer &[cmp] is different
from the "system" &[cmp], and will use that if that is the case.

The use of CALLERS:: is a bit icky, and I'm not sure that this workaround
will be needed as soon as new-disp lands.

Also took the opportunity to reduce the amount of nqp code in this logic.

Has no discernable performance effect on any sorting, except on sorting
empty lists.

### DIFF
--- a/src/core.c/List.pm6
+++ b/src/core.c/List.pm6
@@ -1309,79 +1309,56 @@ my class List does Iterable does Positional { # declared in BOOTSTRAP
         )
     }
 
+    # Helper method for reifying what needs to be sorted
+    method !do-todo(--> Nil) {
+        $!todo.reify-until-lazy;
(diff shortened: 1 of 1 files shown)


lizmat:
@jnthn hope you have some time for some thoughts on this :-)

vrurg:
@lizmat I was solving similar case for `Range`: https://github.com/rakudo/rakudo/blob/522e49a7ae7545c1df62801458622de6c7bb2356/src/core.c/Range.pm6#L461-L472

PR #4489, opened 2021-08-08
Make sure Float.NaN/Double.NaN can be passed by using the NaN

titsuki:
Hi,
I found that nativecall test for NaN is missing when creating a xgboost [0] bindings.
In some machine learning libraries like xgboost, NaN is usable for handling missing feature values.

Cheers,

[0] https://github.com/dmlc/xgboost


titsuki:
Thanks for your review!

PR #4491, opened 2021-08-11
Make Scalar a non-value-object

vrurg:
Which it is not.

Resolves Raku/problem-solving#291


vrurg:
This PR currently breaks roast's _S02-types/list.t_. But I consider the test itself using incorrect approach and will submit a fix shortly.

niner (approved):

lizmat (approved):

PR #4492, opened 2021-08-13
Raise priority of `let` and `temp` operators

vrurg:
Make their behavior more intuitive when used with `.=`


vrurg:
Resolves Raku/problem-solving#292

Leont:
That line should probably be moved up, to keep it in precedence order

vrurg:
This is why it's a draft. I'd like it to be first considered and accepted in problem-solving.

vrurg:
PR is passing spectests and otherwise ready for merge.

PR #4496, opened 2021-08-19
Make concurrent use of %!conc_table safe

MasterDuke17:
Because it's used by MRO method dispatch, doing things such as parsing a
grammar in a start block can cause concurrent access and modification.
Locking didn't seem to work, but clone->modify-clone->replace-with-clone
does. Fixes the various hash-related MVM_oops in S17-promise/start.t.
dogbert17++, nine++, vrurg++

### DIFF
--- a/src/Perl6/Metamodel/Concretization.nqp
+++ b/src/Perl6/Metamodel/Concretization.nqp
@@ -36,11 +36,13 @@ role Perl6::Metamodel::Concretization {
     }
 
     method !rebuild_table() {
+        my %new_conc_table := nqp::clone(%!conc_table);
         for @!concretizations {
-            nqp::scwbdisable();
-            %!conc_table{~nqp::objectid(nqp::decont($_[0]))} := nqp::decont($_[1]);
-            nqp::scwbenable();
+            %new_conc_table{~nqp::objectid(nqp::decont($_[0]))} := nqp::decont($_[1]);
         }
+        nqp::scwbdisable();
+        %!conc_table := %new_conc_table;
+        nqp::scwbenable();
     }
 
     # Returns a list where the first element is the number of roles found and the rest are actual type objects.


lizmat (approved):
I have only one small nit about the naming of the new hash: `conc_table_clone`  It's not a clone of the original table, it it's the **new** state of the table.  So I would have called it something like `new_conc_table`.  That's all

nwc10:
Agree that this should fix the oops, but doesn't this approach just hide a different problem?

In that, what we seem to have is 2+ threads both trying to make **modifications** to the same hash (potentially) concurrently. And (surely) both threads expect their modification to persist? (Else why write it to the hash in the first place?) And with this commit, if the two threads "arrive" at the same time

1. thread 1 copies the hash
2. thread 2 copies the hash
3. thread 1 does its modification locally
4. thread 2 does its modification locally
5. thread 1 writes back (presumably de-facto atomically, given it's a pointer)
6. thread 2 writes back

and with that, thread 1's modification is lost.

(re-order who writes back first, but someone always loses)

So surely the correct solution to the underlying problem has to be locking and a hash, or something that isn't a hash at all?

lizmat:
Agree.

Before we go hunting this down: would this change with new-disp?  I assume it won't with RakuAST.

nwc10:
That is a good question that I hadn't considered. I don't know enough about either (or the current code) to even know how to figure this out.

(This is basically me saying I can't help further - sorry, but not wanting anyone to wait thinking that I might have a clue)

MasterDuke17:
I don't think so, iirc dogbert17 found on new-disp and master.

lizmat:
That it was found on new-disp, doesn't necessarily mean it won't change on new-disp?

niner:
This is part of the meta model. There's no reason to assume that it will change with new-disp and I don't think RakuAST will affect it in any way.

Of couse @nwc10 is right and I should have noticed it yesterday already. We do need a lock in `!rebuild_table`. The advantage of the clone-and-replace trick is that we need that lock _only_ in `!rebuild_table` and not for read-only access.

MasterDuke17:
Yeah, my bad, I got a little trigger happy with the merge button.
Unfortunately, I'm unlikely to have any time to re-visit this until Sunday
evening.

lizmat:
@Altai-man FWIW, it's an improvement on the previous behaviour.  And as such good enough for the 2021.08 release.

PR #4501, opened 2021-08-22
Don't lose concurrent modifications to %!conc_table

MasterDuke17:
They were just fixed to not cause MVM_oop'es by https://github.com/rakudo/rakudo/pull/4496,
but as @nwc10++ pointed out, we might now silently lose them. So also
add a lock such that all thread's modifications are kept.

### DIFF
--- a/src/Perl6/Metamodel/Concretization.nqp
+++ b/src/Perl6/Metamodel/Concretization.nqp
@@ -3,6 +3,8 @@ role Perl6::Metamodel::Concretization {
     has @!concretizations;
     has %!conc_table;
 
+    my $lock := NQPLock.new;
+
     method add_concretization($obj, $role, $concrete) {
         @!concretizations[+@!concretizations] := [$role, $concrete];
         nqp::scwbdisable();
@@ -36,13 +38,15 @@ role Perl6::Metamodel::Concretization {
     }
 
     method !rebuild_table() {
-        my %new_conc_table := nqp::clone(%!conc_table);
-        for @!concretizations {
-            %new_conc_table{~nqp::objectid(nqp::decont($_[0]))} := nqp::decont($_[1]);
-        }
-        nqp::scwbdisable();
-        %!conc_table := %new_conc_table;
-        nqp::scwbenable();
+        $lock.protect: {
+            my %new_conc_table := nqp::clone(%!conc_table);
+            for @!concretizations {
+                %new_conc_table{~nqp::objectid(nqp::decont($_[0]))} := nqp::decont($_[1]);
+            }
+            nqp::scwbdisable();
+            %!conc_table := %new_conc_table;
+            nqp::scwbenable();
+        };
     }
 
     # Returns a list where the first element is the number of roles found and the rest are actual type objects.
--- a/src/Perl6/Metamodel/Concretization.nqp
+++ b/src/Perl6/Metamodel/Concretization.nqp
@@ -37,27 +37,33 @@ role Perl6::Metamodel::Concretization {
         @conc
     }
 
-    method !rebuild_table() {
+    method !maybe_rebuild_table() {
+        my %new_conc_table := %!conc_table;
         $lock.protect: {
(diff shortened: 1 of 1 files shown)


lizmat (approved):

vrurg (approved):

vrurg:
BTW, as far as I understand clone->change->assign is still needed to keep the serializer happy?

niner:
> BTW, as far as I understand clone->change->assign is still needed to keep
> the serializer happy?

No, clone->change->assign is so we can have concurrent readers without having 
to take a lock on every read access (which presumably are much more common 
than writes).

vrurg:
Ah, of course. Ok, Houston, you know... The lock only helps with table rebuilds. But `concretization_lookup` is still reading from the hash in an unsafe way.

To make things more atomic, `!rebuild_table` must be invoked unconditionally. The condition must be checked inside it to decide if actual rebuild is needed. The `!rebuild_table` must return the hash it considers to be valid and `concretization_lookup` must preserve that and work with it, not with the attribue itself.

`rebuild_table` must then become `maybe_rebuild_table`.

vrurg (changes requested):
The fix is still incomplete as `concretization_lookup` method remains unsafe.

MasterDuke17:
> The fix is still incomplete as `concretization_lookup` method remains unsafe.

Opinions on db12008?

Now I'm wondering how many other metamodel classes are potentially unsafe?

vrurg (approved):

src/Perl6/Metamodel/Concretization.nqp
```diff
@@ -35,25 +37,31 @@ role Perl6::Metamodel::Concretization {
...
-        for @!concretizations {
-            %new_conc_table{~nqp::objectid(nqp::decont($_[0]))} := nqp::decont($_[1]);
-        }
-        nqp::scwbdisable();
-        %!conc_table := %new_conc_table;
-        nqp::scwbenable();
+    method !maybe_rebuild_table() {
+        my %new_conc_table;
+        $lock.protect: {
+            %new_conc_table := nqp::clone(%!conc_table);
```
vrurg, line 43: We can do a micro-optimization here by doiing initial binding of `%new_conc_table` to `%!conc_table` and moving `clone` inside the following `if`. Since we never modify the content of the hash already bound to the attribute it is safe to return it without cloning if no changes are needed. And only clone when there are new entries.

vrurg:
> Now I'm wondering how many other metamodel classes are potentially unsafe?

You've touched a dangerous ground here. I'd take the responsibility of admitting, that most of them are. Whenever there is a `set_*` method – there is a flag to be raised.

For example, role `Naming`. If `set_name` is invoked concurrently, then code stability depends on the binding operator and `nqp::setdebugtypename` to be safe. And even if they are, the name in `$!name` could end up being different from the one recorded by the backend.

`MultipleInheritance` role:

- `exlude_parent` pushes into a class `my` list
- `!rebuild_hides_ids` has about the same issue, as this PR fixes (after all, both are pieces of my code written about the same time). Just contrary to the concretizations, it is extremely unlikely to be invoked concurrently. And it operates on an attribute, not a class lexical. Yet, it is still possible to imagine a custom-built monster-class with hidden parents added concurrently by user code
- `add_parent` itself adds entries to `@!hides` and `@!parents`

It is hard to say, how much code is exposed to implicit concurrency. I.e. to cases where legitimate concurrent Raku code re-delegates to the metamodel behind the scenes. Like, say, with the method dispatching causing test flappers. But we certainly mostly unprotected from "malicious" user code.

MasterDuke17:
> > Now I'm wondering how many other metamodel classes are potentially unsafe?
> 
> You've touched a dangerous ground here.

I asked this over in #moarvm (https://colabti.org/irclogger/irclogger_log/moarvm?date=2021-08-24#l58) and @jnthn's main point was "There's a general expectation that they should be used from a single thread during construction, and effectively immutable after compose. Should a meta-object want to have mutable state post-compose then it should take care of threading concerns."

PR #4506, opened 2021-09-01
Fix escaping of # in regexes

### LINKED ISSUE #1324: '#' literals in Grammars: syntax error
## The Problem

There does not seem to be a way of including hashes (`#`) in grammars except by using `\c'

## Expected Behavior

Pretty much as it works in regular expressions, Escaping it should work, like here

```
> '#' ~~ /\#/
｢#｣
```

## Actual Behavior

```
grammar Fails {
    regex TOP { \# <stuff> }
    token stuff { \w+ }
}

my $parsed = Fails.parse( "#stuff");
```

Fails with:
```
$ perl6 Fails.p6
===SORRY!===
Missing required term after infix
at /home/jmerelo/Code/perl6/dev.to-code/perl6/Fails.p6:3
------>     token stuff { \w+ ⏏}
    expecting any of:
        prefix
        term
Other potential difficulties:
    Space is not significant here; please use quotes or :s (:sigspace) modifier (or, to suppress this warning, omit the space, or otherwise change the spacing)
    at /home/jmerelo/Code/perl6/dev.to-code/perl6/Fails.p6:3
    ------>     token⏏ stuff { \w+ }
```

## Steps to Reproduce

Save the code above and run with perl6 fails.p6. It does not matter if it's a token either.

## Environment

* Operating system:  Ubuntu
```
This is Rakudo version 2017.11 built on MoarVM version 2017.11
implementing Perl 6.c.
```

codesections:
In the `MAIN` language, `foo \# ` is correctly parsed as `foo` followed by an unspace and then a comment.  In the regex Slang, however, unspaces are not allowed; so that code should be parsed the same as `foo '#'`.

However, a 2013 commit (9ed945d) to add support for embedded comments in regexes inadvertently allowed unspaces as well, leading to misparsing of `\#` in some cases.  This PR changes the implementation of the 2013 commit to allow embedded comments without incorrectly allowing unspaces.

Closes #1324 and Raku/doc/issues/3947.  See also raku/old-issue-tracker/issues/6654 (a closed issue describing the same problem)

### DIFF
--- a/src/Perl6/Grammar.nqp
+++ b/src/Perl6/Grammar.nqp
@@ -5759,7 +5759,8 @@ grammar Perl6::RegexGrammar is QRegex::P6Regex::Grammar does STD does MatchPacka
     method throw_solitary_quantifier() { self.typed_panic('X::Syntax::Regex::SolitaryQuantifier') }
     method throw_solitary_backtrack_control() { self.typed_sorry('X::Syntax::Regex::SolitaryBacktrackControl') }
 
-    token normspace { <?before \s | '#'> <.LANG('MAIN', 'ws')> }
+    token rxws { [ \s | <.LANG('MAIN', 'comment')> ]* }
+    token normspace {  <?before \s | '#'> <.rxws>  }
 
     token rxstopper { <stopper> }
 


lizmat:
@codesections Can this PR be closed?

codesections:
Closing for now; it might be worth coming back to this one day with a more limited approach.

PR #4508, opened 2021-09-01
Update the concretization table more sparingly

Kaiepi:
This shouldn't need to be cleared upon each addition to the concretization list; because that can only be added to from its end and because we know how many concretizations are already in any given table,we should know exactly which concretizations will be missing from the clone during a rebuild.

The racey grammar test from `S17-promise/basic.t` seems to react OK.

### DIFF
--- a/src/Perl6/Metamodel/Concretization.nqp
+++ b/src/Perl6/Metamodel/Concretization.nqp
@@ -7,9 +7,6 @@ role Perl6::Metamodel::Concretization {
 
     method add_concretization($obj, $role, $concrete) {
         @!concretizations[+@!concretizations] := [$role, $concrete];
-        nqp::scwbdisable();
-        %!conc_table := nqp::hash(); # reset the cache
-        nqp::scwbenable();
     }
 
     method concretizations($obj, :$local = 0, :$transitive = 1) {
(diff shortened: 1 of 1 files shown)


Kaiepi:
This moves the `%new_conc_table` declaration inside the lock. This is to prevent a scenario where two threads see an updated `@!concretizations` and both go ahead with rebuilding based off the same prior `%!conc_table`.

src/Perl6/Metamodel/Concretization.nqp
```diff
@@ -38,21 +35,29 @@ role Perl6::Metamodel::Concretization {
     }
 
     method !maybe_rebuild_table() {
-        my %new_conc_table := %!conc_table;
+        # Capturing the concretization list is first and foremost because we
+        # depend on its size to know whether or not a rebuild is necessary, but
+        # there may be a wait before then. Try for more predictable output.
+        my int $captured := nqp::elems(@!concretizations);
```
vrurg, line 41: Why isn't this inside the lock?

Kaiepi: It belonging in there would also bring the write in `add_concretization` into question. I erred on the side of the current behaviour there being correct. If a concretization gets added after a call to this has begun, then it shouldn't influence the resulting table. Taking the size of the list ASAP like this aims to prevent that from happening during the remainder of the call.

PR #4510, opened 2021-09-01
Add info to CannotPrecomp exception

codesections:
The `CannotPrecomp` error message was LTA because it did not provide actionable information to new users (who may not be aware of the distinction between pre-compiled and interpreted files).  This revision clarifies that users should remove the item that cannot be pre-compiled from a Module.

See [this StackOverflow answer](https://stackoverflow.com/a/69018308/10173009) discussing the LTA error message.

### DIFF
--- a/src/core.c/Exception.pm6
+++ b/src/core.c/Exception.pm6
@@ -343,7 +343,7 @@ my class X::Pragma::NoArgs is Exception {
 }
 my class X::Pragma::CannotPrecomp is Exception {
     has $.what = 'This compilation unit';
-    method message { "$.what cannot be pre-compiled and thus cannot be used in a Module" }
+    method message { "$.what cannot be pre-compiled and thus cannot be used in a module" }
 }
 my class X::Pragma::CannotWhat is Exception {
     has $.what;
--- a/src/core.c/Exception.pm6
+++ b/src/core.c/Exception.pm6
@@ -343,7 +343,7 @@ my class X::Pragma::NoArgs is Exception {
 }
 my class X::Pragma::CannotPrecomp is Exception {
     has $.what = 'This compilation unit';
-    method message { "$.what may not be pre-compiled" }
+    method message { "$.what cannot be pre-compiled and thus cannot be used in a Module" }
 }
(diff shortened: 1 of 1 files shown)


src/core.c/Exception.pm6
```diff
@@ -343,7 +343,7 @@ my class X::Pragma::NoArgs is Exception {
 }
 my class X::Pragma::CannotPrecomp is Exception {
     has $.what = 'This compilation unit';
-    method message { "$.what may not be pre-compiled" }
+    method message { "$.what cannot be pre-compiled and thus cannot be used in a Module" }
```
MasterDuke17, line 346: I'll still +1 it regardless, but I'd suggest 'module' instead of 'Module'. I took a quick look in the docs, and while the keyword is 'Module', most uses of the word not at the beginning of a sentence are lowercase.

codesections: Good point; fixed.

MasterDuke17 (approved):

src/core.c/Exception.pm6
```diff
@@ -343,7 +343,7 @@ my class X::Pragma::NoArgs is Exception {
 }
 my class X::Pragma::CannotPrecomp is Exception {
     has $.what = 'This compilation unit';
-    method message { "$.what may not be pre-compiled" }
+    method message { "$.what cannot be pre-compiled and thus cannot be used in a module" }
```
jnthn, line 346: While the case in question (about `use lib`) results in a sensible hint, I wonder if other values of `$!what` end up more confusing? For example, given the default `what`, it'd say "This compilation unit cannot be pre-compiled  and thus cannot be used in a module", which is a bit odd.

jnthn: Hm, looking at that one case, it seems you'd pretty much have to have invoked `raku --target=mbc --output=foo.moarvm something-with-no-precompilation.rakumod` in order to see that message, which...most probably nobody does, since we manage precompilation automatically. If a tree falls in the forest and nobody hears it... :-)

PR #4511, opened 2021-09-02
Disallow creating Uni with out-of-range codepoints

moon-chilled:


lizmat:
Thank you for this!  Two questions / requests:

- are we checking for negative values as well?
- instead of nqp::die, please create a named exception class with the necessary information as attributes, and do a `.new.throw` on that.

moon-chilled:
Done!  Also added a check for surrogate codepoints.

I wonder if there's a way to make [this](https://github.com/MoarVM/MoarVM/blob/master/src/strings/utf8.c#L639) moarvm routine throw this exception?

moon-chilled:
Oh huh, I guess that absent an isgt_u, the array[uint32] candidate needs a <0 check too.

lizmat:
I don't think this is ready to be merged, right?

moon-chilled:
I think I addressed all the feedback.  Did I miss anything?

lizmat:
Thanks!

PR #4516, opened 2021-09-07
Add a candidate of Promise.at that accepts a DateTime

jonathanstowe:
Passing a DateTime to Promise.at seems natural and it strikes me that a
common use case will be to pass a date time that one might have at hand.

This adds a candidate with a coercion to Instant for `$at` (so will also
accept a custom class that implements .Instant,) and refines the
existing candidate to coerce to Numeric() so as to make the error on
being passed something that can't be converted to Numeric clearer.

### DIFF
--- a/src/core.c/Promise.pm6
+++ b/src/core.c/Promise.pm6
@@ -303,7 +303,14 @@ my class Promise does Awaitable {
         $scheduler.cue({ $vow.keep(True) }, :in($seconds));
         $p
     }
-    method at(Promise:U: $at, :$scheduler = $*SCHEDULER) {
+
+    proto method at(|) { * }
+
+    multi method at(Promise:U: Instant() $at, :$scheduler = $*SCHEDULER) {
+        self.in( $at - now, :$scheduler )
+    }
+
+    multi method at(Promise:U: Numeric() $at, :$scheduler = $*SCHEDULER) {
         self.in( $at - now, :$scheduler )
     }
 


vrurg:
There is no reason to create a duplicate method, especially since `Instant` is a `Numeric`. Instead it'd be more correct to fix `DateTime` coercion, which is apparently broken in this aspect. It looks like simple re-delegation to method `Instant` would do.

jonathanstowe:
Yes, a Numeric coercion on DateTime would work equally well (I actually thought there was one.)

lizmat:
So, can we expect a new PR for infix:<-> or shall I do that?  :-)

jonathanstowe:
Isn't just add a:

```
method Numeric(-->Numeric) {
    self.Instant.Numeric
}
```

To `DateTime` ?

jonathanstowe:
Which of course then opens the whole can of worms regarding arithmetic on DateTime objects.

lizmat:
FWIW, I was just thinking of adding `infix:<->(DateTime:D,Instant:D --> Instant:D)` and `infix:<->(Instant:D, DateTime:D --> Instant:D)` candidates

vrurg:
> FWIW, I was just thinking of adding `infix:<->(DateTime:D,Instant:D --> Duration:D)` and `infix:<->(Instant:D, DateTime:D --> Duration:D)` candidates

I'd say it makes sense irrelevant to numeric coercion.



> Isn't just add a:
> 
> ```
> method Numeric(-->Numeric) {
>     self.Instant.Numeric
> }
> ```

There is no need for the trailing `.Numeric` because `Instant` `does` the role already.

MasterDuke17:
> FWIW, I was just thinking of adding `infix:<->(DateTime:D,Instant:D --> Duration:D)` and `infix:<->(Instant:D, DateTime:D --> Duration:D)` candidates

Completely off topic, but wasn't there some proposal a while ago for a way to say `foo(A|B, A|B) { A.bar; B.baz; }` without have to have `multi sub foo(A, B) { A.bar; B.baz; }; multi sub foo(B, A) { B.baz; A.bar; }`?

lizmat:
https://github.com/rakudo/rakudo/pull/4518

vrurg:
And  #4519

jonathanstowe:
I'll close this as those other things seem to have this fairly well covered.

Issue #4531, opened 2021-09-19
Undefined initialiser


moon-chilled:
```
$ raku -e 'my Int:D() $x; say $x'
(Int)
```

I would expect either 0 or an error about a missing initializer.

moon-chilled:
P.S. bizarre error message I encountered:

```
$ raku -e 'Int:D.isa(Int:D, Int)'
Cannot resolve caller isa(Int:D:U: Int:D:U, Int:U); none of these signatures match
```

What is an Int:D:U?  Can it hurt me?

moon-chilled:
```
$ raku -e 'constant A = Int:U; constant B = A:U; constant C = B:U; say C:U'
(Int:U:U:U:U)
```

jonathanstowe:
I'm beginning to think that the type smiley with coercion is disarming the effect of the type smiley generally.

PR #4532, opened 2021-09-22
Clarify some edge cases of using nominalizables

vrurg:
This is a follow-up to #4531. While considering the original issue I have realized a couple of situations where behaviors of nominalizables needs fixing.


vrurg:
Suspended, awaiting for Raku/problem-solving#299

PR #4533, opened 2021-09-22
Use new dispatcher-drop-n-args syscall

timo:
Can reduce allocations of temporary MVMCapture and MVMCallsite objects.


jnthn:
Manually merged a slightly tweaked version; thanks!

PR #4536, opened 2021-09-24
Consistently throw X::Bind errors when rebinding readonly parameters or sigilless terms

codesections:
According to [S06](https://design.raku.org/S06.html#Parameters_and_arguments), "A [readonly] parameter may also not be rebound; trying to do so results in a compile time error."  See also [S02](https://design.raku.org/S02.html) and [S07](https://design.raku.org/S07.html) (both mentioning that parameter binding has `::=` semantics) and [S03](https://design.raku.org/S03.html) (describing the semantics of `::=`).

However, it appears that this behavior is not currently tested in Roast, creating a bit of a [hole in the spec](https://perl6.party/post/Perl6-On-Specs-Versioning-Changes-And-Breakage#thefutureofthespec) where it doesn't fully test the intend behavior.  Perhaps as a result, the non-rebindability of parameters is inconsistently enforced by Rakudo.

Specifically, prior to this PR, the following code generated a compile-time `X::Bind` error

```raku
sub f($a) { $a := 0 }
```

but neither of the following two lines did:

```raku
sub f(Mu $a) { $a := 0   }
sub g(   @a) { @a := [0] }
```

The first commit in this PR fixes that discrepancy by calling the existing `is_lexical_marked_ro` method to determine whether a parameter is readonly.  It also annotates parameters as readonly or not readonly in a few locations (which is necessary because `is_lexical_marked_ro` is used before the `Parameter` objects are constructed).

Finally, this PR's first commit changes the one location where Rakudo code took advantage of the previous bug by rebinding a readonly parameter.

Because signature destructuring operates by binding the parameters in a signature, this commit also fixes the incorrectly-allowed rebinding of variables with values set in that way.  Both for functions and standalone signatures, non-readonly parameters (that is, any that are `copy` or `rw`) can be freely rebound. Thus,

```raku
my ($a is copy, $b) := (0, 1); $a := 42;
```

is correctly allowed, but generates a compilation error without the `is copy`.

This PR's second commit addresses the similarly inconsistent behavior between `my \a = 0; a := 42` (an `X::Bind` error) and `my Mu \a = 0; a := 42` (previously allowed).  Just like rebinding of parameters, rebinding terms declared with `\` is not supposed to be allowed:

> We sometimes call these "sigilless variables", but they differ from normal variables in one significant way. Unlike a variable binding, a term binding is fixed for the rest of the lexical scope. The term may only be rebound be re-entering the scope. This is useful to enforce a programming style known as SSA, Single Static Assignment (really binding), which, among other benefits, gives the optimizer more guarantees about which symbols can be considered "temporarily immutable".

[S06](https://design.raku.org/S06.html).  See also f303efcea4953268cf51b9cd02728309839203df (commit by @jnthn with the commit message "Enforce SSA on sigilless in `my (\x, \y) = 1, 2;`.")

After this PR, neither form is allowed; unlike with parameters, `\` terms do not have `is copy` or `is rw` as escape hatches.

This PR passes all tests and spectests, and applies cleanly/passes all tests on the `new-disp` branch as well. 

---

I have not prepared any corresponding Roast tests.  Assuming that this PR looks good, is there consensus on pinning down the behavior described in S02, S03, S06, and S07?  I'm inclined to think that we should, given that this behavior is fairly fundamental to Raku's underlying semantics and it would be very helpful to have it documented and speced in Roast.  But I know that specifying new behavior in Roast is a long-term commitment, so I wanted to double check. If there's consensus about adding Roast tests, I'd be happy to prepare a PR.

I spent a while figuring out the intended behavior for rebinding, and asked several people for help.  Thanks especially to @raiph  and @moritz for helping me figure out what the intended behavior is.  The following links point to conversations I had or resources I found when researching the topic (in addition to the previously linked Synopses, of course)

* [StackOverflow: What are the rules for re-binding](https://stackoverflow.com/questions/69231506/what-are-the-rules-for-re-binding)
* [StackOverflow: Should sigilless "variables" with type constraints be re-bindable?](https://stackoverflow.com/questions/69049465/should-sigilless-variables-with-type-constraints-be-re-bindable)
* [Stack Overflow Chat](https://chat.stackoverflow.com/rooms/237343/discussion-between-codesections-and-raiph) (discussing the "What are the rules for re-binding" question at more length)
* [#raku IRC channel, 2021-09-22](https://logs.liz.nl/raku/2021-09-22.html) and [2021-09-23](https://logs.liz.nl/raku/2021-09-23.html) (discussing same SO question).
* [StackOverflow: Is there a purpose or benefit in prohibiting sigilless variables from rebinding?](https://stackoverflow.com/questions/50399784/is-there-a-purpose-or-benefit-in-prohibiting-sigilless-variables-from-rebinding/50401989#50401989) (2018-05-17 answer explaining SSA in `\` terms)
* Raku/doc#3950 (discussing whether current behavior `\` is correct, and how to document it)

### DIFF
--- a/src/Perl6/Actions.nqp
+++ b/src/Perl6/Actions.nqp
@@ -1243,11 +1243,11 @@ class Perl6::Actions is HLL::Actions does STDActions {
             my $type := $*OFTYPE.ast;
             $cur_lexpad[0].push(QAST::Var.new( :$name, :scope('lexical'),
                 :decl('var'), :returns($type) ));
-            $cur_lexpad.symbol($name, :$type, :scope<lexical>);
+            $cur_lexpad.symbol($name, :$type, :scope<lexical>, :ro(1));
         }
         else {
             $cur_lexpad[0].push(QAST::Var.new(:$name, :scope('lexical'), :decl('var')));
-            $cur_lexpad.symbol($name, :scope('lexical'));
+            $cur_lexpad.symbol($name, :scope('lexical'), :ro(1));
         }
         make $<defterm>.ast;
     }
--- a/src/Perl6/Actions.nqp
+++ b/src/Perl6/Actions.nqp
@@ -7591,14 +7591,14 @@ class Perl6::Actions is HLL::Actions does STDActions {
         }
         # Everything else is a (re)binding error.  Check the types to give a more specific msg
         elsif nqp::istype($target, QAST::WVal)
-        && $target.value.raku eq $target.node {    # A type (class, role, etc)
+            && $target.value.raku eq $target.node {    # A type (class, role, etc)
             $*W.throw($/, ['X', 'Bind', 'Rebind'], :target($target.value.raku), :is-type(1))
         }
         elsif nqp::istype($target, QAST::WVal) {   # A constant
             $*W.throw($/, ['X', 'Bind', 'Rebind'], :target(nqp::escape($target.node)))
         }
         elsif nqp::istype($target, QAST::Op)
-        && $target.op eq 'ifnull' {                # Code (subs, regex, etc)
+            && $target.op eq 'ifnull' {                # Code (subs, regex, etc)
             $*W.throw($/, ['X', 'Bind', 'Rebind'], target => nqp::escape($target[0].name));
         }
         else {             # Items that can never be bound (PsudoPackages, literals, etc)
--- a/src/Perl6/Actions.nqp
+++ b/src/Perl6/Actions.nqp
@@ -7540,7 +7540,7 @@ class Perl6::Actions is HLL::Actions does STDActions {
                     $was_lexical := 1;
                 }
                 if $*W.is_lexical_marked_ro($target.name) || !$was_lexical {
-                    $*W.throw($/, ['X', 'Bind']);
+                    $*W.throw($/, ['X', 'Bind', 'Rebind'], target => $target.name);
                 }
             }
 
@@ -7589,8 +7589,19 @@ class Perl6::Actions is HLL::Actions does STDActions {
             $dynbind.nosink(1);
             make $dynbind;
         }
-        # XXX Several more cases to do...
-        else {
+        # Everything else is a (re)binding error.  Check the types to give a more specific msg
+        elsif nqp::istype($target, QAST::WVal)
+        && $target.value.raku eq $target.node {    # A type (class, role, etc)
+            $*W.throw($/, ['X', 'Bind', 'Rebind'], :target($target.value.raku), :is-type(1))
+        }
+        elsif nqp::istype($target, QAST::WVal) {   # A constant
+            $*W.throw($/, ['X', 'Bind', 'Rebind'], :target(nqp::escape($target.node)))
+        }
+        elsif nqp::istype($target, QAST::Op)
+        && $target.op eq 'ifnull' {                # Code (subs, regex, etc)
+            $*W.throw($/, ['X', 'Bind', 'Rebind'], target => nqp::escape($target[0].name));
+        }
+        else {             # Items that can never be bound (PsudoPackages, literals, etc)
             $*W.throw($/, ['X', 'Bind']);
         }
     }
--- a/src/Perl6/Actions.nqp
+++ b/src/Perl6/Actions.nqp
@@ -7539,7 +7539,7 @@ class Perl6::Actions is HLL::Actions does STDActions {
                     }
                     $was_lexical := 1;
                 }
-                unless $was_lexical {
+                if $*W.is_lexical_marked_ro($target.name) || !$was_lexical {
                     $*W.throw($/, ['X', 'Bind']);
                 }
             }
--- a/src/Perl6/World.nqp
+++ b/src/Perl6/World.nqp
@@ -338,7 +338,7 @@ class Perl6::World is HLL::World {
                 $i := $i - 1;
                 my %sym := @!PADS[$i].symbol($name);
                 if %sym {
-                    return nqp::existskey(%sym, 'ro');
+                    return nqp::existskey(%sym, 'ro') && nqp::atkey(%sym, 'ro');
                 }
             }
             0;
@@ -1795,7 +1795,7 @@ class Perl6::World is HLL::World {
             return nqp::null();
         }
 
-        $block.symbol($name, :value($cont));
+        $block.symbol($name, :value($cont), :ro(0));
         self.install_package_symbol_unchecked($package, $name, $cont) if $scope eq 'our';
 
         # Tweak var to have container.
@@ -2392,6 +2392,7 @@ class Perl6::World is HLL::World {
             # Add variable as needed.
             my int $flags := nqp::getattr_i($param_obj, $param_type, '$!flags');
             my $varname := $_<variable_name>;
+
             if $varname && ($flags +& $SIG_ELEM_IS_RW || $flags +& $SIG_ELEM_IS_COPY) {
                 my %sym := $lexpad.symbol($varname);
                 if +%sym && !nqp::existskey(%sym, 'descriptor') {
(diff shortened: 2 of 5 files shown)


patrickbkr:
Wow. I enthusiastically support this PR. I'd like for https://github.com/Raku/doc/issues/3950 (make sure our documentation of sigils / variables is excelent) to be solved along with this PR as a way to verify the behaviors sanity and consistency.

src/Perl6/Actions.nqp
```diff
@@ -7486,7 +7486,7 @@ class Perl6::Actions is HLL::Actions does STDActions {
                     }
                     $was_lexical := 1;
                 }
-                unless $was_lexical {
+                if $*W.is_lexical_marked_ro($target.name) || !$was_lexical {
```
vrurg, line 7489: Since we explicitly know the cause of failed binding, I'd rather add a new exception and be more elaborate on the cause.

vrurg:
I have strong feeling against `my ($a, $b) := (0,1); $a := 42` to be an error. It is semantically no different from `my $a := 0; my $b := 1;`.

vrurg:
> Actually it is: the form `my ($a, $b) := (0,1);` is defined in terms of signature binding.

But then it will most certainly will be causing WAT effect.

codesections:
To elaborate on what @jnthn said: the appearance of the syntax makes it look as though 

> It is semantically no different from `my $a := 0; my $b := 1;`.

and I thought so myself for a long time.

But, as mentioned, the `my ($a, $b) := (0, 1);` form has the semantics of signature binding, which creates several significant (though subtle) knock-on semantic differences. 

Most notably, signature binding binds the `capture` to the `signature` (without a scalar), but still creates scalars for the `parameter`s / variables:

```raku
my ($a, $b) := (0, 1); 
for $a, $b { say $a.VAR.WHAT } # OUTPUT: «(Scalar)␤(Scalar)»
```
Conversely, `my $a := 0` of course binds directly:

```raku
my $a := 0; my $b := 1; 
for $a, $b { say $a.VAR.WHAT } # OUTPUT: «(Int)␤(Int)»
```

Similarly, `my ($a, $b) = (0, 1)` does _not_ support variable traits that can be used with `my $a := 0`, but _does_ support traits that can be used in a `signature`

For example, 
```raku
my ($a of Int, $b) := (0, 1);
```

throws a compiler error, but

```raku
my ($a is rw, $b) := ($=0, 1);
```
is allowed.

That's not an exhaustive list of differences, of course; just a sampling.

> But then it will most certainly will be causing WAT effect.

I agree (and, to be honest, I kind of wish we'd required the rhs to be `\(0, 1)`, which seems like it would have added clarity at the cost of a single char.  But too late for that now and, at the time, there may have been perl5-trap reasons for not wanting to do so).

But I think it's better to have an obvious WAT that users encounter early than subtle ones that they run into later.  As you and I both show, it's very easy to write a lot of Raku code with the mental model that `my ($a, $b) := (0, 1)` has the same semantics as `my $a := 0; my $b := 1`.  But, as I've discovered, that incorrect assumption can cause a lot of confusion and difficult-to-debug errors (especially when trying to write more immutable code).

Having said that, to the extent that this does cause a WAT, it's _especially_ important to have a good error message.  So your previous point about having a different `Exception` with a more descriptive message is spot on.  I think I let my desire to keep this PR minimal/show that it isn't a big change distract me from the in-hindsight-obvious inadequacy of `Cannot use bind operator with this left-hand side` as an error message for rebinding.

I'll update the PR with an `X::Rebind` error with a significantly more descriptive error message.

vrurg:
> I'll update the PR with an `X::Rebind` error with a significantly more descriptive error message.

We have a family of `X::Bind` exceptions. Better be kept under the namespace.

> I also have a WIP doc PR [Raku/doc#3962](https://github.com/Raku/doc/pull/3962) that should help avoid a WAT here.

I've got a second thought about this PR before even seeing your comment. It's the best answer to any future confusions.

lizmat:
So is this good to merge?

raiph:
> So is this good to merge?

+1 from me provided roast isn't being changed.

codesections:
> So is this good to merge?

Oh, oops, I think we were waiting on me to fix the Exception the way @vrurg suggested.  I'll do that now and make sure the branch still merges with passing tests.

> +1 from me provided roast isn't being changed.

I don't think Roast needs to be changed as part of this PR.  I do think it would be a good idea for Roast to more clearly specify the behavior in this area, but that seems like a fairly separate issue.

codesections:
Ok, I've added the `X::Bind::Rebind` error message, as @vrurg suggested.  Sorry about the delay – it turned out to be a bit harder than I expected to determine where to throw a `Rebind` versus `Bind` error with the information available in `Actions.nqp`.

Because I wanted to make sure I got that right, I also added several tests.  I believe that these tests could also form the basis for a future Roast PR but, as I said above, I don't think this PR needs to wait for that.  Does anyone have feedback on those tests before I draft a Roast PR along similar lines? (@raiph, you raised issues re: Roast upthread, so I thought you might have some thoughts).

In any event, I believe this PR is ready to be merged at least as far as I know.

vrurg:
Looks like you need to rebase your branch. 95 commits pushed...

codesections:
> Looks like you need to rebase your branch. 95 commits pushed...

The time between when I pushed to my fork and when I noticed the 95 commits and realized I hadn't _force_ pushed to my fork after rebasing was about 3 minutes; after that, it was fixed.  But you just **had** to call me out for it anyway :D

vrurg:
Damn github not notifying about forced pushes! My apologies! (facepalm)

src/Perl6/Actions.nqp
```diff
@@ -7589,8 +7589,19 @@ class Perl6::Actions is HLL::Actions does STDActions {
...
+        # Everything else is a (re)binding error.  Check the types to give a more specific msg
+        elsif nqp::istype($target, QAST::WVal)
+        && $target.value.raku eq $target.node {    # A type (class, role, etc)
+            $*W.throw($/, ['X', 'Bind', 'Rebind'], :target($target.value.raku), :is-type(1))
+        }
+        elsif nqp::istype($target, QAST::WVal) {   # A constant
+            $*W.throw($/, ['X', 'Bind', 'Rebind'], :target(nqp::escape($target.node)))
+        }
+        elsif nqp::istype($target, QAST::Op)
+        && $target.op eq 'ifnull' {                # Code (subs, regex, etc)
```
vrurg, line 7601: Nitpicking on indentations here and above, at line 7594.

vrurg (changes requested):
I'll try to have a closer look at the semantics side of things tomorrow. So far, aside of the bad indentations, I see no apparent problems. Tests look like they are ready for roast.

I wonder if this makes us one step closer to `::=`.

src/Perl6/Actions.nqp
codesections: > Nitpicking on indentations here and above, at line 7594.

Do you mean the fact that the `&&` isn't indented relative to the line above?  That's not my preferred style either, but it seems to be the most common one in use in this file, including in this fn at line 7581.  I'm happy to change it if you'd like, but I figured that opting for local consistency was the lesser evil here.

vrurg: I'd better change the other indentations. Looks like either 7581 and others are artifacts of incorrectly expanded tabs, or a weird habit of Zoffix. Anyway, it seems we have such a diversity of styles in the file that it's hard to stick to a single one. So, let's choose something readable. :) My own preference is to have logic op in front of its RHS, indented. It's also what Liz does. Looks like Larry and Jonathan prefer the op to follow its RHS. So, just choose whichever one is better for you. :)

vrurg (approved):
I'd propose moving the tests into Roast.

PR #4538, opened 2021-09-28
Don't make signature part of a callable named parameter

### LINKED ISSUE #4537: Cannot use a signature constraint on a named callable argument
## The Problem

Named `&` parameters work, and `&` parameters with signature constraints work, but using those two features together does not. 

## Expected Behavior
```raku
sub named(:&fn)                       { 'ok ' }
sub constrained(&fn:(Int))            { 'ok ' }
sub named-and-constrained(:&fn:(Int)) { 'ok ' }
sub fn(Int) {}
say named(:&fn), constrained(&fn), named-and-constrained(:&fn);
# OUTPUT: «ok ok ok»
```

## Actual Behavior
That code throws `No such method 'signature' for invocant of type 'Callable' in sub named-and-constrained`.

## Environment
* tested on `v2021.06` and master `v2021.09-4`

vrurg:
When a named parameter is declared as `:&foo:(Int)` the name used for
parameter binding was `foo:(Int)` resulting in falling back to
parameter's default - `Callable`. A visible effect of this error was
parameter checking to attempt calling `signature` method on `Callable`
role.

Resolves #4537

### DIFF
--- a/src/Perl6/Actions.nqp
+++ b/src/Perl6/Actions.nqp
@@ -5693,19 +5693,7 @@ class Perl6::Actions is HLL::Actions does STDActions {
         }
 
         if $<name><sigterm> || $<sigterm> -> $sig {
-            unless %*PARAM_INFO<post_constraints> {
-                %*PARAM_INFO<post_constraints> := [];
-            }
-            my $get_signature_past := QAST::Op.new(
-                :op('callmethod'),
-                :name('signature'),
-                WANTED(QAST::Var.new( :name('$_'), :scope('lexical') ),'param_var')
-            );
-            my $fakesig := $sig<fakesignature>;
-            my $closure_signature := $fakesig.ast;
-
-            my $where := make_where_block($fakesig, $closure_signature, $get_signature_past);
-            %*PARAM_INFO<post_constraints>.push($where);
+            %*PARAM_INFO<signature_constraint> := $sig<fakesignature>.ast.value;
         }
 
         if $<arrayshape> {
@@ -5759,7 +5747,7 @@ class Perl6::Actions is HLL::Actions does STDActions {
     method named_param($/) {
         %*PARAM_INFO<named_names> := %*PARAM_INFO<named_names> || nqp::list_s();
         if $<name>               { nqp::push_s(%*PARAM_INFO<named_names>, ~$<name>); }
-        elsif $<param_var><name> { 
+        elsif $<param_var><name> {
             my $name := $<param_var><name>;
             nqp::push_s(%*PARAM_INFO<named_names>, ~($name<subshortname> // $name));
         }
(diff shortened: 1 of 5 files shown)


niner (approved):

vrurg:
PR needs one more change. Declarations like `:&foo:(Int)` cannot be optional because they would fail to match against parameter default `Callable`. Need to add either a compile-time warning and make the parameter implicitly non-optional, or make it an error.

vrurg:
I have fixed the case when a signature-constrained named parameter results in `signature` method called on `Callable` if a routine is not supplied with corresponding named argument. The fix makes any such parameter required implicitly unless `?` is used which is a compile-time error.

There is a couple of notes on feasibility of making signature-constrained parameters optional. The primary problem about them is that the default value (which is `Callable`) doesn't have a `signature` method. And even if it would have it (I was playing with making `Code` the default for signature-constrained parameters); and the method would return some valid signature object (nonsense, but what if?) - even then it is unlikely to match user's constraint.

There are possible workarounds. 

The first is to mixin a custom role into the parameter default with a `signature` method which returns the user's constraint signature.

The second is to add `$!signature` attribute to the `Parameter` class. Comes at the cost of extra memory footprint only to take care of rather rarely used feature. A workaround to this, again, is to mixin a role with the attribute. In this case we would need to move argument signature validation code from a post-constraint `where` block into signature binding. The downside: it's a lot of work for the basically deprecated compiler code facing the replacement with RakuAST.

User-based solution for optional callable parameter within the approach taken by this PR would be `:&foo where { .defined && .signature ~~ :(Int) }`.

Opinions are welcome.

vrurg:
There is a bug related to type captures:

```raku
sub foo(::T, &bar:(T)) { ... }
```

The constraining signature is not getting instantiated. Perhaps, instantiating post-constraints list would work as expected. If not then attaching the constraining signature onto the `Parameter` object would be a better solution.

vrurg:
This PR is ready for merge. CI JVM failures are there on master, they are not caused by the PR.

Signature constraint information has been relocated from inside a `when` block in `@!post_constraints` list to a new attribute `$!signature_constraint`. Argument's signature is checked explicitly at parameter bind time. This move also allowed me to implement support for generics. So, `sub foo(::T, &fn:(T)) {...}` and `role R[::T] { method foo(&fn:(T)) {...} }` are both working now.

Aside of that, it should be possible now to consider the constraining signature in multi-dispatch deliberately, using it to better narrow down the selection of matching candidates. Unfortunately, I'm not sufficiently familiar with multi-dispatch guts to tweak them.

I have not found a reliable way to determine if a named parameter, or an optional positional, was actually passed with a routine call or not. For this reason I deduct the fact of it being missing by comparing parameter variable value to parameter's default type. It would be totally OK with the default `&parameter` declaration as `Callable` cannot be used as an argument for signature-constrained parameters. But if a trait changes `Parameter` `$!type` to something different this could become a potential problem.

src/Perl6/Actions.nqp
```diff
@@ -9716,6 +9710,72 @@ class Perl6::Actions is HLL::Actions does STDActions {
...
+                                    :op<getattr>,
+                                    $inst_param,
+                                    QAST::Op.new(:op<what>, $inst_param),
+                                    QAST::SVal.new(:value('$!signature-constraint')));
+                }
+                $var.push(QAST::ParamTypeCheck.new(
+                    QAST::Op.new(
+                        # If argument is a type object and is the same as parameter default then skip signature
+                        # matching. So far, this is the best way I know to determine if corresponding argument was
+                        # passed or not without inspecting the capture which is too slow.
```
niner, line 9757: Inspecting captures and matching types is what new-disp does all the time and efficiently too thanks to caching and guards. Sounds like this would be an opportunity for utilizing the new possibilities.

So instead of generating code into each routine, the dispatcher should contain the logic for handling arguments with signatures.

niner (changes requested):

src/Perl6/Actions.nqp
vrurg: @niner  Unfortunately, I'm not into dispatchers yet and will need time to learn them sufficiently to start actually doing something. The whole signature binding could be done as a dispatcher, I think. But before this is done I'd rather have a temporary fix in place. This would also allow to merge roast tests, so when dispatched solution arrives we have them in place to ensure it works as intended.

Aside of this, I think the whole thing is going to be different with RakuAST anyway. It is likely that dispatching will come as part of RakuAST implementation.

So, if there are no other objections, I'd be for merging this solution for now. As I'm involuntarily getting more and more involved into signature-related stuff, perhaps, when time allows, I'd be the one to start rewriting it. :)

niner: Fair enough :) Please, go ahead

PR #4544, opened 2021-10-01
Remove bad substr candidate

moon-chilled:
Was causing an infinite loop with `"whatever".substr`


lizmat:
Thanks for the report!  The problem was not a bad candidate, but a missing `Str.substr` candidate.  Added with https://github.com/rakudo/rakudo/commit/93f0d86fcd .

PR #4554, opened 2021-10-06
Potential simplification to working with suggestions for misspellings

MasterDuke17:
This is probably what I intended when I originally added suggestions
for methods, but I just didn't notice sometimes it was a number and
sometimes it was a StrDistance.


lizmat:
Thanks!

PR #4555, opened 2021-10-06
Raku is a legally registered trademark now in the USA, UK and Europe.

nige123:
More details can be found here:

https://news.perlfoundation.org/post/raku-foundation-and-tm-update


lizmat:
Ok, so what's the difference between (tm) and (R) then?  I was under the impression we should use (tm) ???

coke:
@codesections, any feedback on this?

Leont:
> Ok, so what's the difference between (tm) and (R) then? I was under the impression we should use (tm) ???

It's a very American distinction, AFAICT

PR #4559, opened 2021-10-10
Change guard to track value be non-Nil

vrurg:
Without this guard under certain conditions it is possible that a
container with non-nominal `.of` (like a subset) would take over the
optimization path of assigning `Nil` to a nominal container.

There're two scenarios where this problem occurs. The first is
`nqp::clone` of attribute's `auto_viv_container` and assign to the clone
bound to a lexical.

The second path is to `nqp::create(Scalar)` and set its descriptor to
attribute's container descriptor.

Don't know if there is way to experience this behavior without
involvement of nqp ops.

### DIFF
--- a/src/vm/moar/dispatchers.nqp
+++ b/src/vm/moar/dispatchers.nqp
@@ -396,9 +396,7 @@
                     }
                     else {
                         # Non-nominal or type check error.
-                        my $tracked-of := nqp::dispatch('boot-syscall', 'dispatcher-track-attr',
-                                $tracked-desc, ContainerDescriptor, '$!of');
-                        nqp::dispatch('boot-syscall', 'dispatcher-guard-literal', $tracked-of);
+                        nqp::dispatch('boot-syscall', 'dispatcher-guard-not-literal-obj', $tracked-value, Nil);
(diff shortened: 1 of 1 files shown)


vrurg:
Here is the best golf I've got to catch the problem:

```raku
use nqp;

sub asgn($obj, Mu \value) {
    my $attr := nqp::decont($obj.^attributes.grep(*.name.contains('foo')).head);
    #my $v := nqp::clone($attr.auto_viv_container.VAR);
    my $v := nqp::create(Scalar);
    nqp::bindattr(
        $v, Scalar, '$!descriptor',
        nqp::getattr(nqp::decont($attr), Attribute, '$!container_descriptor')
    );
    $v = value; # This is where it throws without this PR
    nqp::bindattr(nqp::decont($obj), $obj.WHAT, $attr.name, $v);
}

{
    my class Foo {
        has Int $.foo is rw where * <= 42;
    }
    my $inst = Foo.new;
    asgn($inst, 42);
    note $inst.foo;
}

{
    my class Foo {
        has Str $.foo;
    }
    my $inst = Foo.new;
    asgn($inst, Nil);
    note $inst.foo.raku;
}
```

PR #4560, opened 2021-10-10
[JVM] Unbreak build after merge of new-disp

usev6:
There where three errors during the build of the JVM backend that
have been fixed / worked around as follows:

* The change to src/vm/jvm/Perl6/Ops.nqp prevents an explosion with
  "getcodeobj can only be used with a CodeRef".

* After fixing that the build died with "Error while compiling op
  dispatch, no registered operation handler". This seems to be caused
  by https://github.com/rakudo/rakudo/commit/de24aac1b8 and just
  re-introducing the old behaviour for the other backends brought up
  the third error:

* "Could not instantiate role '<anon|1>': CodeRef representation does
  not support attributes". It looks like this happened since
  commit https://github.com/rakudo/rakudo/commit/25d2f634dd. Again,
  just re-introduing the old behaviour for the other backends avoided
  the error.

The first two errors and possible workarounds were spotted by
 MasterDuke++.

### DIFF
--- a/src/Perl6/Actions.nqp
+++ b/src/Perl6/Actions.nqp
@@ -9074,6 +9074,10 @@ class Perl6::Actions is HLL::Actions does STDActions {
         # handling.
         if $need_full_binder {
             $block.custom_args(1);
+#?if !moar
+            $block[0].push(QAST::Op.new( :op('p6bindsig') ));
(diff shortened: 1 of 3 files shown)


MasterDuke17:
Nice.

PR #4566, opened 2021-10-12
Make GLOBAL available to compile time code

### LINKED ISSUE #4557: Error while compiling "No Such Method 'EXISTS-KEY'" on second instance of use Inline::Python
## The Problem

Rakudo fails to compile a module when more than one of its sub-modules contains "use Inline::Python".

Golfed down to [this gist](https://gist.github.com/ggoebel/7831a0b2269fce57c4cd366c84e7e164)

Note: if Inline::Python is replaced with Inline::Perl5, then the error goes away.

## Expected Behavior

compile successfully

## Actual Behavior

$ raku -MFoo -I./lib -e 'say "ok"'
===SORRY!=== Error while compiling /home/ggoebel/git/test/lib/Foo.rakumod (Foo)
===SORRY!=== Error while compiling /home/ggoebel/git/test/lib/Baz.rakumod (Baz)
No such method 'EXISTS-KEY' for invocant of type 'BOOTHash'.  Did you
mean any of these: 'EXISTS-KEY', 'EXISTS-POS'?
at /home/ggoebel/git/test/lib/Baz.rakumod (Baz):2

at /home/ggoebel/git/test/lib/Foo.rakumod (Foo):3

at -e:1

## Steps to Reproduce

Golfed down to [this gist](https://gist.github.com/ggoebel/7831a0b2269fce57c4cd366c84e7e164)

## Environment
* Operating system: arch linux
* Compiler version: v2021.09-251-gff6951f59

vrurg:
This should resolve #4557 but primarily this fixes
CompUnit::RepositoryRegistry::short-id2class which throws on
`::($short-id)` statement because `GLOBAL` is not available at this
point despite it is already known either from compiler options, or from
unit's outers, or via a new one created by the World object.

As far as I'm aware, this also makes `GLOBAL` the same symbol for
compunit's mainline when it is executed at load time, and for run-time
code.

### DIFF
--- a/src/Perl6/World.nqp
+++ b/src/Perl6/World.nqp
@@ -724,6 +724,8 @@ class Perl6::World is HLL::World {
             self.pkg_compose($/, $*GLOBALish);
         }
 
+        nqp::bindhllsym('Raku', 'GLOBAL', $*GLOBALish);
+
         # Create or pull in existing EXPORT.
         if $!have_outer && $*UNIT_OUTER.symbol('EXPORT') {
             $*EXPORT :=


vrurg:
This PR passes spectests.

Issue #4539, opened 2021-09-28
Heredocs get confused about lexical scope when used with some block constructs


codesections:
The string created by a `q:to` heredoc is conceptually supposed to occur on the line in which the heredoc is started.  This means that we can correctly write code like

```raku
sub f { my $a = 'foo'; qq:to/END/ }
       $a
       END
say f; # OUTPUT: «foo»
```

This works because `$a` is evaluated in `&f`'s lexical scope, despite the fact that the _token_ `$a` occurs after `f`'s closing bracket.

However, it seems that some block contructs confuse the heredoc about this point.  Here's a golfed example:
```raku
sub f { {my $a = 'foo'; qq:to/END/ } }
       $a
       END
say f;
# OUTPUT: «===SORRY!=== Error while compiling /n Variable '$a' is not declared»
```

I've also run into this issue with `if` and `when` blocks – it seems that many nested blocks can trigger the bug shown above.

tbrowder:
I'm surprised the first example worked, much less the second.

codesections:
> I'm surprised the first example worked, much less the second.

Yeah, it's not _super_ well-documented at the moment, but it's a consequence of the fact that, as [the docs say](https://docs.raku.org/language/quoting#Heredocs:_:to)

> The contents of the heredoc always begin on the next line, so you can (and should) finish the line.

And the accompanying example shows that in action:

```raku
my $escaped = my-escaping-function(q:to/TERMINATOR/, language => 'html'); 
Here are the contents of the heredoc.
Potentially multiple lines.
TERMINATOR
```

I'd be happy to add more details to the docs, but don't want to do so before we debug some of the corner cases (to avoid creating an attractive nuisance).

jnthn:
Generally Raku is very strict about lexical scopes being delimited by curly braces. Glancing at the block parsing rule in the grammar, we find this:

```
    token blockoid {
        :my $*CURPAD;
        :my %*HANDLERS;
        <.finishpad>
        :my $borg := $*BORG;
        :my $has_mystery := $*MYSTERY ?? 1 !! 0;
        { $*BORG := {} }
        [
        | '{YOU_ARE_HERE}' <you_are_here>
        | :dba('block')
            '{'
            <!!{ $*VARIABLE := '' if $*VARIABLE; 1 }>
            <statementlist(1)>
            [<.cheat_heredoc> || '}']
            <?ENDSTMT>
        || <.missing_block($borg, $has_mystery)>
        ]
[...]

PR #4567, opened 2021-10-12
Prevent heredocs from using block variables when continuing after the block

tbrowder:
This change addresses proper handling of heredocs
raised in issue #4539 where the docs implied a heredoc
could be defined with its terminator following a block.

Jonathan said that was not intended and explained
how to fix it. That was the only change to the
code and it had the desired effect: a heredoc
cannot be defined with its terminator following a block, 
but it can be defined with its terminator following a function 
call line.

New tests have been added for this change in roast file 
'S02-literals/heredocs.t'.

Note the docs should be updated to discuss the ramifications
of this fix.

### DIFF
--- a/src/Perl6/Grammar.nqp
+++ b/src/Perl6/Grammar.nqp
@@ -1069,11 +1069,11 @@ grammar Perl6::Grammar is HLL::Grammar does STD {
             '{'
             <!!{ $*VARIABLE := '' if $*VARIABLE; 1 }>
             <statementlist(1)>
+            { $*CURPAD := $*W.pop_lexpad() }
             [<.cheat_heredoc> || '}']
             <?ENDSTMT>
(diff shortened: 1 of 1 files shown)


jnthn:
> This change addresses proper handling of heredocs
> raised in issue #4539 where the docs implied a heredoc
> could be defined with its terminator following a block.
> 
> Jonathan said that was not intended and explained
> how to fix it.

Hm, that's not quite what I intended. It's fine if the heredoc's terminator is outside of the block. It's *not* fine if the variables in the block are visible beyond the closing `}` of the block. So:

```
if $x { say q:to/END/; }
    This is alright
    END
```

Is no problem, but:

```
if $x { my $var = 42; say q:to/END/; }
    Should not be able to see $var
    END
```

Should be an error. I didn't expect the change I proposed to prevent the first, only the second.

PR #4568, opened 2021-10-12
Make GLOBAL shared at compunit load time

vrurg:
Previously a compunit was installing own GLOBALish as GLOBAL
unconditionally, effectively making it different from loading process'
GLOBAL. This resulted in cases where module's globals are kind of lost
at run time. For example:

    unit module Foo;
    GLOBAL::<$FOO> = 42;
    say $*FOO; # 42
    sub foo is export {
        say $*FOO; # ERROR!
    }

If `foo` is invoked by a script then `$*FOO` is not visible because it
is imported as `$FOO` into script's UNIT namespace and the GLOBAL &foo
sees at run-time is not module's GLOBALish, but scipt's GLOBAL which
lacks the symbol.

### DIFF
--- a/src/Perl6/World.nqp
+++ b/src/Perl6/World.nqp
@@ -2582,9 +2582,6 @@ class Perl6::World is HLL::World {
         # we ever try to run it during compilation.
         my $precomp;
         my $compiler_thunk := {
-            # Fix up GLOBAL.
-            nqp::bindhllsym('Raku', 'GLOBAL', $*GLOBALish);
-
             # Compile the block.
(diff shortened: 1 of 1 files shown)


vrurg:
This PR doesn't change the situation with compile-time `GLOBAL`.

```
unit module Foo;
BEGIN {
    GLOBAL::<$FOO> = pi;
    say $*FOO; # 3.141...
}
say $*FOO; # Error
```

Fixing this case wouldn't be possible without explicitly splitting `GLOBAL` and `UNIT`/`GLOBALish` at compile time and then doing separate merge of globals for both.

PR #4572, opened 2021-10-13
Use distribution specific comp unit ids in CURFS

ugexe:
Previously CU::R::FileSystem would only use the short name when generating a
comp unit id. However when two different versions of the otherwise same
module were loaded the same precomp file would get used. This changes to use
the distribution id (which already accounts for version/api/auth) in
combination with the short name of the requested module.

### DIFF
--- a/src/core.c/CompUnit/Repository/FileSystem.pm6
+++ b/src/core.c/CompUnit/Repository/FileSystem.pm6
@@ -75,7 +75,7 @@ class CompUnit::Repository::FileSystem
 
         with self!matching-dist($spec) {
             my $name = $spec.short-name;
-            my $id   = self!comp-unit-id($name);
+            my $id   = self!comp-unit-id($_.id ~ $name);
             my $*DISTRIBUTION  = $_;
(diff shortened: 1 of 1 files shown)


ugexe:
This fixes the following:
```
❯ tree
.
├── foo1
│   ├── META6.json
│   └── lib
│       └── Foo.rakumod
└── foo2
    ├── META6.json
    └── lib
        └── Foo.rakumod


4 directories, 4 files


❯ cat foo1/lib/Foo.rakumod
unit package Foo;

our sub my-version { 1 }


❯ cat foo2/lib/Foo.rakumod
unit package Foo;

our sub my-version { 2 }


❯ cat foo1/META6.json
{
"name":"foo",
"version":1,
"provides": { "Foo": "lib/Foo.rakumod" }
}


❯ cat foo2/META6.json
{
"name":"foo",
"version":2,
"provides":{ "Foo": "lib/Foo.rakumod" }
}


❯ perl6 -I foo1/ -I foo2/ -e '{ use Foo:ver<1>; say Foo::my-version(); }; { use Foo:ver<2>; say Foo::my-version(); }'
1
1
```

Issue #4558, opened 2021-10-08
Signatures with subsignatures smartmatch incorrectly in both directions 


codesections:
The [docs](https://docs.raku.org/routine/ACCEPTS#class_Signature) say that a `Signature.ACCEPTS` any other `Signature` if "if anything accepted by [that other `Signature] would also be accepted by the invocant".

If that's correct, it means that a `Signature` with a subsignature should match the same signature without that subsignature – for example, `:$foo ($, $)` requires a named argument `:foo` with a value that can be destructured into two values; that should match `:$foo`, which also requires a named `:foo` but places no constraints on its value.

However, that's currently not what happens:

```raku
say :(:$foo ($, $)) ~~ :(:$foo) # OUTPUT: «False»
```

[**EDIT**: Actually, the issue is broader than I realized when first opening this issue, and can result in false positives as well.  For example:

```raku
say :(:$foo ($, $)) ~~ :(:$bar ($, $)) # OUTPUT: «True»
```

That's more clearly a bug.  Maybe `.ACCEPTS` is somehow focusing only on the inner subsignature?  Or using the "only-look-at-values" semantics of `Pair.ACCEPTS` when it shouldn't be? ]

I tested on a fresh Rakudo built today (2021-10-08, v2021.09-253-gd2fdf2d0d).  I'm mentioning an exact build because a recent commit (66ae8612972efd27af4039aefbbbb1020f51abf2) fixed a related issue where `Signature`s wouldn't match against `|` but did _not_ fix the behavior described in this issue.

PR #4573, opened 2021-10-14
Fix handling of constrained Mu parameters in signature smartmatching

Kaiepi:
`Signature.ACCEPTS` detects the absence of a type for a parameter by comparing its type object to `Mu`. The problem is this doesn't consider any type constraints (where clauses, subsignatures, `Callable` signatures) the parameter may have, leading to false positives when the type object is lacking. `Parameter.ACCEPTS` also mishandles type constraints: it has paranoid guards against constraints in the LHS lacking in the RHS that prevents subtyping of parameters in this way, leading to false negatives.

While fixing `Parameter.ACCEPTS` is primarily a matter of removing said conditions, the absence of any type-related information in a `Parameter:D` whatsoever isn't very efficient to introspect from outside `Parameter`, thus the introduction of the `untyped` getter. Now fixing `Signature.ACCEPTS` is just a matter of replacing comparisons to `Mu` with calls to `Parameter.untyped`.

Fixes https://github.com/rakudo/rakudo/issues/4558.


patrickbkr:
Kaiepi: Do you have a commit bit?

Kaiepi:
Yeah, sorry. I was assuming merging my own PR was a no-no.

patrickbkr:
Forgiveness > Permission. :-)
I guess having a review on a PR is a good thing, but nothing against merging ones own PRs.

PR #4574, opened 2021-10-15
[JVM] Restore pre-new-disp behaviour for phasers

usev6:
This basically undoes the changes from 84da856ae6 for the non-moar
backends. This unbreaks a couple of tests for the JVM backend, that
now die with "CodeRef representation does not support attributes".


usev6:
The failing CI checks are clearly not related to the change.

PR #4577, opened 2021-10-17
Revert CurFS fixes for now

lizmat:
Until we can figure out why this is causing the spectest to take
about 2x as long (from 357 -> 616 wallclock for yours truly)


lizmat:
Slowdown has been fixed @ugexe++

PR #4579, opened 2021-10-18
Add is-wrapped method to Routine

vrurg:
The method was added with previous attempt to fix dispatching and
correspondingly was specced in S06-advanced/dispatching.t. Makes sense
as an implementation independent way of determining if a routine is
wrapped.

### DIFF
--- a/src/core.c/Routine.pm6
+++ b/src/core.c/Routine.pm6
@@ -75,9 +75,9 @@ my class Routine { # declared in BOOTSTRAP
         $raku
     }
 
-    method soft( --> Nil ) { }
+    method soft(--> Nil) { }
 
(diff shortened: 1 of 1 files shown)


vrurg:
> What is the motivation for having this?

My motivation was to have a helper for spectesting. There was something else back then, but I don't remember what it was. Generally speaking, there is a mutable state and having a straight forward and simple way to get the state makes sense to me.

Out of imaginary situations, there could be a trait making decision based on this info. Or one could filter out wrapped methods for analysis.

> Also, I think all new names we add to Raku objects should follow the `-` convention

I think so too. Back then it was heavy influence of the metamodel, yesterday it was reluctance to modify roast. :) Both times I was clearly wrong.

Issue #1433, opened 2018-01-23
`let`/`temp` fail to restore holes in arrays


zoffixznet:
```
<Zoffix__> m: (my @a is default(Nil))[3] = 42; say @a[1]:exists; say @a; { let @a; Nil }; say @a[1]:exists; say @a
<camelia> rakudo-moar 0a1641e94: OUTPUT: «False␤[Nil Nil Nil 42]␤True␤[(Any) (Any) (Any) 42]␤»

<Zoffix__> m: (my @a is default(Nil))[3] = 42; say @a[1]:exists; say @a; { temp @a; Nil }; say @a[1]:exists; say @a
<camelia> rakudo-moar 0a1641e94: OUTPUT: «False␤[Nil Nil Nil 42]␤True␤[(Any) (Any) (Any) 42]␤»
```

raiph:
I just tried this and the holes are now restored to the default. But that's not the same as restoring the variable, and its holes; note the difference in `@a[1]:exists`:
```
(my @a is default(Nil))[3] = 42; say @a[1]:exists; say @a; { let @a; Nil }; say @a[1]:exists; say @a;
# OUTPUT: «False␤[Nil Nil Nil 42]␤True␤[Nil Nil Nil 42 ]␤»

(my @a is default(Nil))[3] = 42; say @a[1]:exists; say @a; { temp @a; Nil }; say @a[1]:exists; say @a;
# OUTPUT: «False␤[Nil Nil Nil 42]␤True␤[Nil Nil Nil 42 ]␤»
```

vrurg:
While generally speaking I'd rather agree that it makes sense for a hole to be restored, I'm not sure it worth the cost to implement this functionality. Because technically `let` is a routine and `@a[$i]` results in an auto-vivified container which then gets passed into `let` as an argument, there is no way for the operator to know that it operates on an array element.

The task could be done by the phaser code, the compiler installs to actually implement the restoration process. But that code is agnostic about what it operates upon. In order to function correctly (including itemized arrays like in `$a[$i]` case) it would have to introspect the object. This  would involve cases like `I-return-array()[$i]`, `$obj.a[$i]`, etc.

Ok, the postcircumfix tips the compiler about possible line of actions to be taken. But what about `$obj.a.AT-POS($i)`? Or any kind of alternate method implementing similar functionality on a `Positional`?

[...]

PR #4602, opened 2021-10-29
Clarify let and temp operators

vrurg:
Make sure they preserve and restore holes in arrays (though preserving currently depends on `clone` method implementation) and conterization of elements.

The new implementation shifts the locus of control to `Array` and `Hash` themselve, leaving it up to `Rakudo::Internals` to deal with `Scalars`. The latter was done to allow the backend VM take better care of
optimizations.

With regard to the non-scalar containers, they're expected to provide two methods: `TEMP-LET-LOCALIZE` and `TEMP-LET-RESTORE`. The first is expected to return a handle which would let `TEMP-LET-RESTORE` recover the initial state of the container. The handle object is not constrained in neither way and it is totally up to the implementation to choose what the handle is.

The reason for this new approach is that normally a container would know better how to preserve and restore its internal state. This new approach also allows for third-party containers to provide full support for `let` and `temp` operators with ease.

### DIFF
--- a/src/Perl6/World.nqp
+++ b/src/Perl6/World.nqp
@@ -2787,8 +2787,7 @@ class Perl6::World is HLL::World {
             QAST::Var.new( :name($value_stash), :scope('lexical'), :decl('var') ),
             QAST::Op.new(
               :op('create'),
-              QAST::WVal.new( :value(self.find_single_symbol('IterationBuffer', :setting-only))),
-            )));
+              QAST::WVal.new( :value(self.find_single_symbol('IterationBuffer', :setting-only))))));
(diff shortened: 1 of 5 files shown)


vrurg:
Should resolve #1433. 

I have some doubts about possible side effects on todo-ed arrays. What would happen if a `temp`-orized array gets reified and then restored? I guess any `Seq` involved would throw as already consumed. But as long as localization process only re-initializes an array from a clone, reification could only happen as a result of user operations in which case it's rather their responsibility to take care of this situation. As a helper means, we could issue a warning for non-reified arrays.

PR #4604, opened 2021-10-30
use dispatcher-replace-arg in some dispatchers

timo:
relies on https://github.com/Raku/nqp/pull/745 which relies on https://github.com/MoarVM/MoarVM/pull/1585

passes spectest on my machine


MasterDuke17:
Can this be re-reverted? It was undone in https://github.com/rakudo/rakudo/commit/a507f7c0473a4c34d46ff8cb71d00048b5a673e4, but I'm not sure why.

PR #4605, opened 2021-10-31
Fix gisting a Buf or Blob parameterized with an unsized type

MasterDuke17:
The unsized natives don't have their $!nativesize set, so just assume
64-bit for them.

Rakudo builds ok and passes `make m-test m-spectest`. The example @dogbert17++ showed (`sub a (int $a) { Blob[int].new($a) }; say a 12;`) now prints `Blob[int]:0x<000000000000000C>` instead of dying with `Cannot unbox a type object (NQPMu) to int.`.

### DIFF
--- a/src/core.c/Buf.pm6
+++ b/src/core.c/Buf.pm6
@@ -350,8 +350,10 @@ my role Blob[::T = uint8] does Positional[T] does Stringy is repr('VMArray') is 
     );
 
     multi method gist(Blob:D:) {
-        my int $todo =
-          nqp::elems(self) min nqp::div_i(200,nqp::div_i(T.^nativesize,4) || 1);
+        # (u)int don't have a nativesize, so just assume 64-bit for them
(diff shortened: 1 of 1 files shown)


lizmat (approved):

src/core.c/Buf.pm6
```diff
@@ -350,8 +350,10 @@ my role Blob[::T = uint8] does Positional[T] does Stringy is repr('VMArray') is
     );
 
     multi method gist(Blob:D:) {
-        my int $todo =
-          nqp::elems(self) min nqp::div_i(200,nqp::div_i(T.^nativesize,4) || 1);
+        # (u)int don't have a nativesize, so just assume 64-bit for them
+        my int $nativesize = nqp::div_i(T.^nativesize // int64.^nativesize, 4) || 1;
```
Kaiepi, line 354: `$?BITS`?

PR #4606, opened 2021-10-31
Add a Str.substr-between method

lizmat:
Returns the substring that is between the two given strings in a
string.  Effectively a shortcut to:

    with $string.match(/ '$before' <( .*? )> '$after') {
        .Str
    }
    else {
        Nil
    }

But does not use the regex engine, so is up to 5x as fast.

This is a proof of concept.  If so accepted, support for regexes for before / after and support for :ignorecase and :ignoremark could be added.

### DIFF
--- a/src/core.c/Str.pm6
+++ b/src/core.c/Str.pm6
@@ -248,6 +248,23 @@ my class Str does Stringy { # declared in BOOTSTRAP
         )
     }
 
+    multi method substr-between(Str:D $before, Str:D $after, Int:D $pos = 0) {
+        my int $left = nqp::index(self, $before, $pos);
+        if $left > -1 {
+            my int $right = nqp::index(self, $after, $left);
+            if $right > -1 {
+                my int $offset = $left + nqp::chars($before);
+                nqp::substr(self, $offset, $right - $offset)
+            }
+            else {
+                Nil
+            }
+        }
+        else {
+            Nil
+        }
+    }
+
     multi method substr-eq(Str:D:
       Str:D $needle, Int:D $pos, :i(:$ignorecase)!, :m(:$ignoremark)
     --> Bool:D) {
--- a/src/core.c/Cool.pm6
+++ b/src/core.c/Cool.pm6
@@ -171,6 +171,13 @@ my class Cool { # declared in BOOTSTRAP
         (SELF = self.Str).substr-rw(from, want)
     }
 
+    proto method substr-between(|) {*}
+    multi method substr-between(Cool:D:
+      Cool:D $before, Cool:D $after
+    --> Bool:D) {
+        self.Str.substr-between($before.Str, $after.Str)
+    }
+
     proto method substr-eq(|) {*}
     multi method substr-eq(Cool:D:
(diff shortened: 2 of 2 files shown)


vrurg:
Why not a multi-candidate of `substr`? This applies to all three new `substr-*` methods

```raku
multi method substr(Str:D: Str:D $from, Str:D $to, Int:D $pos) {...}
multi method subsrt(Str:D: Int:D $pos = 0, Str:D :$before) {...}
multi method subsrt(Str:D: Int:D $pos = 0, Str:D :$after) {...}
```

src/core.c/Str.pm6
```diff
@@ -248,6 +248,23 @@ my class Str does Stringy { # declared in BOOTSTRAP
         )
     }
 
+    multi method substr-between(Str:D $before, Str:D $after, Int:D $pos = 0) {
+        my int $left = nqp::index(self, $before, $pos);
+        if $left > -1 {
+            my int $right = nqp::index(self, $after, $left);
+            if $right > -1 {
+                my int $offset = $left + nqp::chars($before);
+                nqp::substr(self, $offset, $right - $offset)
```
vrurg, line 257: This will not work on overlapping strings:

src: "0123456789";
before: "2345"
after: "456"

This will result in `$left=2`, `$right=4`, `$offset=6` -> `$right - $offset = -2`. Perhaps `$right` should be obtained on `$left + nqp::chars($before)`?

vrurg (changes requested):

src/core.c/Str.pm6
lizmat: Perhaps  :-)

lizmat:
Creating additional candidates for `.substr` might not be as bad as it would have been before new-disp.  So that's also an option.

ugexe:
fwiw I'm not convinced this warrants core humanizing nor language speccing -- is doing this thing so common that it a) many users actually expect a shorter way and b) many users would expect such a niche helper to be optimized to a high degree with nqp?

Consistency is important - all these variants lack `-rw` variants (and I'm not advocating adding those variants but rather pointing out how `substr-*` methods cant scale to arbitrary additions to save a couple lines line). Eventually (hopefully) nqp code won't needed for performance reasons and thus half the reason for implementing these special cases in the first place would no longer be valid.

patrickbkr:
Would it be feasable to alternatively add special cases to our regex protcessing to detect such "most commonly used" cases and transparently switch to a fast path? Then we wouldn't add to our API and would not require users to learn about these performance workarounds. (I seem to recall that Perl does similar auto-downgrading of regexes to simple searches when it detects a feasable regex.)

lizmat:
@patrickbkr Yes, that is also an option!

jnthn:
> Would it be feasable to alternatively add special cases to our regex protcessing to detect such "most commonly used" cases and transparently switch to a fast path? Then we wouldn't add to our API and would not require users to learn about these performance workarounds. (I seem to recall that Perl does similar auto-downgrading of regexes to simple searches when it detects a feasable regex.)

Yes, it would, but the matching/compilation handling of it all is a bit of a pain to do at the moment. Something for the new RakuAST-based compiler frontend, IMO.

PR #4608, opened 2021-10-31
Add a Str.substr-before method.

lizmat:
Returns the substring that is before a given string in a
string. Effectively a shortcut to:

    with $string.match(/ .*? )> '$before') {
        .Str
    }
    else {
        Nil
    }

But does not use the regex engine, so is up to 5x as fast.

Also a proof of concept, with support for regex and :ignorecase and :ignoremark to be added.

### DIFF
--- a/src/core.c/Cool.pm6
+++ b/src/core.c/Cool.pm6
@@ -171,6 +171,11 @@ my class Cool { # declared in BOOTSTRAP
         (SELF = self.Str).substr-rw(from, want)
     }
 
+    proto method substr-before(|) {*}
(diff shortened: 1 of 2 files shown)


jubilatious1:
Does this work with $before acting like a Zero-Width Assertion?  I know we've had a few (recent) issues regarding whether or not certain LookAheads/Behinds/Arounds are consuming characters (incorrectly).

lizmat:
I'm not sure.  The way it is implemented now, is to just search for the string and return anything before it if it is found, else `Nil`.

PR #4610, opened 2021-10-31
Add a Str.rinse method

lizmat:
Returns the substring that cleansed of the given string.
Effectively a shortcut to:

     $string.subst($dirt, :global);

but a bit faster.  At least this may be more semantically
pleasing.

If accepted, could have support for regexes added.  This would then most likely still be faster than the equivalent `.subst` because no `Match` objects would have to be made.

### DIFF
--- a/src/core.c/Cool.pm6
+++ b/src/core.c/Cool.pm6
@@ -171,6 +171,11 @@ my class Cool { # declared in BOOTSTRAP
         (SELF = self.Str).substr-rw(from, want)
     }
 
+    proto method rinse(|) {*}
+    multi method rinse(Cool:D: Cool:D $dirt, Int:D $pos = 0) {
+        self.Str.rinse($dirt.Str, $pos)
+    }
+
     proto method substr-eq(|) {*}
     multi method substr-eq(Cool:D:
       Cool:D $needle, :i(:$ignorecase)!, :m(:$ignoremark) --> Bool:D) {
--- a/src/core.c/Str.pm6
+++ b/src/core.c/Str.pm6
@@ -248,6 +248,15 @@ my class Str does Stringy { # declared in BOOTSTRAP
         )
     }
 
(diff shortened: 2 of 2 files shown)


vrurg (approved):

codesections:
Hmm, I'm not sure I'm a big fan of this from an API design perspective.  One aspect of Raku that I really cherish is that, despite being a pretty big language, we actually don't have all *that* many methods in our standard library – because we have good, composable functions and operators that let us avoid the need for many single-purpose methods that bulk out other standard libraries.  (That was even the theme of my [Raku Conf talk](https://conf.raku.org/talk/149)).  And `Str.rinse` feels like exactly the sort of single-purpose method we usually avoid.

I realize that there's precedent for adding single purpose Str methods to avoid the performance cost of building a Match (`.starts-with` and `.ends-with`), but I view those more as compromises that we needed to make for performance reasons than as a model to emulate.  

I'm hoping that once Raku AST lands, it will be possible to build a compiler pass that addresses many of these performance issues via static optimizations rather than new methods.  For instance, in the code you showed (`$string.subst($dirt, :g)`), the compiler has more than enough info to statically optimize that into the `.rinse` nqp implementation in this PR – there's no way for that Match to escape, we could optimize it out.  And that same optimization strategy would be available in other Str or Bool contexts (`if` statements would be especially handy).

Finally (and less importantly) I'm also not 100% sure about the name `.rinse`.  I didn't find the meaning immediately obvious; it fits, but I can think of several things that "rinsing" a string could mean – in fact, my first thought was that it had something to do with sanitizing input.  "rinse" also isn't a method name I'm aware of other languages using for substring removal, so it seems unlikely to benefit from built-in familiarity.  I tried a quick Internet search to see if other languages have a consensus term for substring removal and I didn't see one right off, mostly because many languages don't have such a method and instead direct user to substitute an empty string – which only makes me feel more strongly that this is too single-use to be worth including right now.

PR #4613, opened 2021-11-02
Make "raku" the authority of core modules

lizmat:
Seems to me that "perl" is a bit...  outdated.  And changing it to "raku" doesn't break any tests.

### DIFF
--- a/tools/build/install-core-dist.raku
+++ b/tools/build/install-core-dist.raku
@@ -44,7 +44,7 @@ $REPO.install(
     Distribution::Hash.new(
         {
             name     => 'CORE',
-            auth     => 'perl',
+            auth     => 'raku',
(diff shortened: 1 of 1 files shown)


vrurg (approved):

lizmat:
I'm not 100% sure this is safe to do yet.

vrurg:
> I'm not 100% sure this is safe to do yet.

Why so?

ugexe:
Among other things it would need to be decided what happens to the existing modules when rakudo is upgraded

lizmat:
Locally, after having run this branch, this says:

````raku
CompUnit::RepositoryRegistry.repository-for-name("core").installed.map(*.meta).elems.say;  # 2
````

PR #4617, opened 2021-11-03
Abstract virtual core META6.json into a module

lizmat:
The install-core-dist.raku script kept the information about which files are actually part of the Rakudo core distribution to itself.

This commit extracts that information into a separate module, so that other programs can get at the meta information in a reliable manner.

This adds a `lib/Rakudo/CORE/META.rakumod` file that makes the meta information available in `%Rakudo::CORE::META` when loaded.

````raku
use Rakudo::CORE::META;
say "Contains %Rakudo::CORE::META<provides>elems() modules";
````

### DIFF
--- a/lib/Rakudo/CORE/META.rakumod
+++ b/lib/Rakudo/CORE/META.rakumod
@@ -1,0 +1,38 @@
+unit class Rakudo::CORE;
+
+our %META = BEGIN { 
+    my %provides = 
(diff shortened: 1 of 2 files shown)


lizmat:
Closing in light of discussion at https://logs.liz.nl/raku-dev/2021-11-03.html#21:36

PR #4618, opened 2021-11-04
Disallow cmp with a Junction

MasterDuke17:

Now `say (9, 3, 1, any(4, 5, 6, 7, 500), "a").sort` dies with `Cannot use Junction 'any(4, 5, 6, 7, 500)' with cmp, Junctions are not comparable.` instead of printing `(a any(4, 5, 6, 7, 500) 1 3 9)`.

### DIFF
--- a/src/core.c/Order.pm6
+++ b/src/core.c/Order.pm6
@@ -11,6 +11,20 @@ sub ORDER(int $i --> Order) is implementation-detail {
 }
 
 proto sub infix:<cmp>($, $, *% --> Order:D) is pure {*}
+# stub needed otherwise you get `Could not find symbol '&Junction' in 'X::Cannot'` at runtime
+my class X::Cannot::Junction { ... }
+multi sub infix:<cmp>(\a, Junction:D $b) {
+    X::Cannot::Junction.new(
+        junction => $b.raku,
+        for      => 'with cmp, Junctions are not comparable',
+    ).throw
+}
+multi sub infix:<cmp>(Junction:D $a, \b) {
+    X::Cannot::Junction.new(
+        junction => $a.raku,
+        for      => 'with cmp, Junctions are not comparable',
+    ).throw
+}
 multi sub infix:<cmp>(\a, \b) {
     nqp::eqaddr(nqp::decont(a), nqp::decont(b))
       ?? Same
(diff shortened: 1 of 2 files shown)


src/core.c/Order.pm6
```diff
@@ -11,6 +11,22 @@ sub ORDER(int $i --> Order) is implementation-detail {
 }
 
 proto sub infix:<cmp>($, $, *% --> Order:D) is pure {*}
+multi sub infix:<cmp>(\a, Junction:D $b) {
+    Perl6::Metamodel::Configuration.throw_or_die(
```
lizmat, line 15: What is this magic?  :-)

MasterDuke17: I originally tried just `X::Cannot::Junction.new(...).throw`, which built fine, but the test case instead died with `Could not find symbol '&Junction' in 'X::Cannot'`.

lizmat: That means just you need to stub the class before using it.

````raku
my class X::Cannot::Junction { ... }
````

moon-chilled:
Hrm, not-entirely-serious suggestion:

`(1, any(0, 2)).sort` -> `any((0, 1), (1, 2))`

src/core.c/Order.pm6
vrurg: @lizmat thanks for the right suggestion! `throw_or_die` is not meant to be used outside of the metamodel.

vrurg:
I have reverted this because, as I mentioned in raku/problem-solving#297, it is not `cmp` which does it wrong, but it is `sort` which doesn't know how to handle junctions.

lizmat:
Then what should `say so (1 cmp 2 | 3)` say ?

vrurg:
It is the same kind of problem as with `say "WHAT???" if 1 cmp 2`.

lizmat:
ok, reverted

vrurg:
BTW, I tried to solve this issue by making enum values of `Order` altered with appropriate booleans. Unfortunately, it didn't work. Most problematic is the fact that the current state of things is encoded in `Test` module and roast. But the fact that `1 cmp 1` is falsy is truly annoying and this is the thing we should, probably, consider. I wish I could get this done right, but not with the current state of things on my side.

PR #4620, opened 2021-11-05
DEPRECATE smartmatching on Junctions

lizmat:
In response to https://github.com/Raku/problem-solving/issues/297 this commit DEPRECATEs the use of Junctions on the left hand-side of a smart match, and as an argument to .ACCEPTS.  Dis-allowing this feature has quite a bit of spectest fallout, so this will need to be looked at more carefully in the future.

### DIFF
--- a/src/core.c/Mu.pm6
+++ b/src/core.c/Mu.pm6
@@ -27,6 +27,10 @@ my class Mu { # declared in BOOTSTRAP
     # while preserving compatibility with existing code that has any ACCEPTS
     # candidates for Mu or Junction.
     multi method ACCEPTS(Mu:U \SELF: Junction:D \topic) is default {
+        DEPRECATED(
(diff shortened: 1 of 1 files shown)


codesections:
Just in case anyone hasn't been following the discussion, this would only deprecate using junctions as the _LHS_ of a smartmatch, not the RHS (which would be a much larger change!).  So `$a ~~ 3|4` would still be fine, it's just `3|4 ~~ $a` that would be deprecated if this is merged.

vrurg:
And my negative opinion on this PR can be found in raku/problem-solving#297

raiph:
I've realized I've seriously misinterpreted, or at least misrepresented, what Larry was saying. See my edit of my opening comment at https://github.com/Raku/problem-solving/issues/297.

PR #4622, opened 2021-11-06
Fix segfaults when multiple threads initialize dynamics

niner:
When multiple threads try to access an uninitialized dynamic like $*PROGRAM-NAME
at the same time, they all will run the initialization code. This is mostly
benign (just a little wasteful) but the actual assignment into the PROCESS::
stash may lead to segfault as that's unsafe writing into a hash. Indeed, this
not only affects PROCESS variables, but stashes in general. Since stashes are
usually global data, any write access (e.g. when loading modules) can lead to
threading issues.

Since we can assume stashes to be read-mostly data structures, we fix this by
cloning the stash's storage, modifying that and replacing the storage while
holding a lock to protect against other concurrent writers. This way read
access will be safe without further concurrency protections.

### DIFF
--- a/src/Perl6/bootstrap.c/BOOTSTRAP.nqp
+++ b/src/Perl6/bootstrap.c/BOOTSTRAP.nqp
@@ -3663,9 +3663,11 @@ BEGIN {
 
     # Set up Stash type, which is really just a hash with a name.
     # class Stash is Hash {
-	#     has str $!longname;
+    #     has str $!longname;
+    #     has $!lock;
     Stash.HOW.add_parent(Stash, Hash);
     Stash.HOW.add_attribute(Stash, Attribute.new(:name<$!longname>, :type(str), :package(Stash)));
+    Stash.HOW.add_attribute(Stash, Attribute.new(:name<$!lock>, :type(Any), :package(Stash)));
     Stash.HOW.compose_repr(Stash);
 
     # Configure the stash type.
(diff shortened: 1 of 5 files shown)


src/core.c/Stash.pm6
```diff
@@ -2,6 +2,10 @@ my class Stash { # declared in BOOTSTRAP
     # class Stash is Hash
     #     has str $!longname;
```
MasterDuke17, line 4: Add a `#     has Lock $!lock;` here?

src/core.c/Stash.pm6
```diff
@@ -29,6 +33,35 @@ my class Stash { # declared in BOOTSTRAP
...
+                nqp::bindattr(self, Map, '$!storage', $storage);
+                scalar
+            };
+        }
+        else {
+            nqp::p6assign(existing-key, assignval);
+        }
+    }
+    multi method ASSIGN-KEY(Stash:D: \key, Mu \assignval) is raw {
+        my str $key = key.Str;
```
lizmat, line 61: Why not `self.ASSIGN-KEY(key.str, assignval)` ?

src/Perl6/bootstrap.c/BOOTSTRAP.nqp
```diff
@@ -3666,6 +3666,7 @@ BEGIN {
 	#     has str $!longname;
```
MasterDuke17, line 3666: Just noticed the pre-existing-but-errant tab here, but the next line could use a `# has Lock $!lock;` also.

niner:
Adressed all comments by amending.

MasterDuke17 (approved):

jnthn (approved):

PR #4626, opened 2021-11-09
Add an Any(iterable).infer method

lizmat:
Given an iterable thing of values, will return the type / role that
is tightest for the values given.

As proposed by jnthn on https://logs.liz.nl/moarvm/2021-11-09.html#10:25-0001

### DIFF
--- a/src/core.c/Any-iterable-methods.pm6
+++ b/src/core.c/Any-iterable-methods.pm6
@@ -2381,8 +2381,8 @@ Consider using a block if any of these are necessary for your mapping code."
 
     method !slow-infer($iterator, Mu $type is copy, Mu $pulled is copy) {
         # set up types to check
-        my $mro   := nqp::clone(nqp::getattr($type.^mro,List,'$!reified'));
-        my $roles := nqp::clone(nqp::getattr($type.^roles,List,'$!reified'));
+        my $mro :=
+          nqp::clone(nqp::getattr($type.^mro(:roles),List,'$!reified'));
 
         nqp::repeat_until(
           nqp::eqaddr(($pulled := $iterator.pull-one),IterationEnd)
@@ -2392,21 +2392,6 @@ Consider using a block if any of these are necessary for your mapping code."
             nqp::stmts(                       # not the same base type
               nqp::shift($mro),
               ($type := nqp::atpos($mro,0)),  # assume next type for now
-              nqp::if(                        # check all roles, if any left
-                nqp::elems($roles),
-                nqp::stmts(
-                  (my $new-roles := nqp::list),
-                  nqp::while(
-                    nqp::elems($roles),
-                    nqp::if(
-                      nqp::istype((my $role := nqp::pop($roles)),$type)
-                        && nqp::istype($pulled,$role),
-                      ($type := nqp::unshift($new-roles,$role))
-                    )
-                  ),
-                  ($roles := $new-roles)
-                )
-              )
             )
           )
         );
--- a/src/core.c/Any-iterable-methods.pm6
+++ b/src/core.c/Any-iterable-methods.pm6
@@ -2362,44 +2362,55 @@ Consider using a block if any of these are necessary for your mapping code."
         my $iterator := self.iterator;
         nqp::if(
           nqp::eqaddr((my $pulled := $iterator.pull-one),IterationEnd),
-          Nil,                                    # nothing to check
+          Nil,                                            # nothing to check
           nqp::stmts(
-            (my $mro := nqp::clone(               # set up types to check
-              nqp::getattr($pulled.^mro,List,'$!reified')
-            )),
-            (my $roles := nqp::clone(
-              nqp::getattr($pulled.^roles,List,'$!reified')
-            )),
-            (my $type := nqp::atpos($mro,0)),     # initial type
+            (my $type := $pulled.WHAT),                   # initial type
             nqp::until(
               nqp::eqaddr(($pulled := $iterator.pull-one),IterationEnd)
-                || nqp::eqaddr($type,Mu),
-              nqp::until(
-                nqp::istype($pulled,nqp::atpos($mro,0)),
-                nqp::stmts(                       # not the same base type
-                  nqp::shift($mro),
-                  ($type := nqp::atpos($mro,0)),  # assume next type for now
-                  nqp::if(                        # check all roles, if any left
-                    nqp::elems($roles),
-                    nqp::stmts(
-                      (my $new-roles := nqp::list),
-                      nqp::while(
-                        nqp::elems($roles),
-                        nqp::if(
-                          nqp::istype((my $role := nqp::pop($roles)),$type)
-                            && nqp::istype($pulled,$role),
-                          ($type := nqp::unshift($new-roles,$role))
-                        )
-                      ),
-                      ($roles := $new-roles)
-                    )
-                  )
-                )
-              )
+                || nqp::not_i(nqp::eqaddr($pulled.WHAT,$type)),
+              nqp::null
             ),
-            $type
+            nqp::if(
+              nqp::eqaddr($pulled,IterationEnd),
+              $type,                                      # all the same
+              self!slow-infer($iterator, $type, $pulled)  # find out what
+            )
           )
         )
+    }
+
+    method !slow-infer($iterator, Mu $type is copy, Mu $pulled is copy) {
+        # set up types to check
+        my $mro   := nqp::clone(nqp::getattr($type.^mro,List,'$!reified'));
+        my $roles := nqp::clone(nqp::getattr($type.^roles,List,'$!reified'));
+
+        nqp::repeat_until(
+          nqp::eqaddr(($pulled := $iterator.pull-one),IterationEnd)
+            || nqp::eqaddr($type,Mu),
+          nqp::until(
+            nqp::istype($pulled,nqp::atpos($mro,0)),
+            nqp::stmts(                       # not the same base type
+              nqp::shift($mro),
+              ($type := nqp::atpos($mro,0)),  # assume next type for now
+              nqp::if(                        # check all roles, if any left
+                nqp::elems($roles),
+                nqp::stmts(
+                  (my $new-roles := nqp::list),
+                  nqp::while(
+                    nqp::elems($roles),
+                    nqp::if(
+                      nqp::istype((my $role := nqp::pop($roles)),$type)
+                        && nqp::istype($pulled,$role),
+                      ($type := nqp::unshift($new-roles,$role))
+                    )
+                  ),
+                  ($roles := $new-roles)
+                )
+              )
+            )
+          )
+        );
+        $type
     }
 
     proto method nodemap(|) is nodal {*}
--- a/src/core.c/Any-iterable-methods.pm6
+++ b/src/core.c/Any-iterable-methods.pm6
@@ -2357,6 +2357,51 @@ Consider using a block if any of these are necessary for your mapping code."
         Seq.new(Rakudo::Iterator.Rotor(self.iterator,@cycle,$partial))
     }
 
+    proto method infer(|) {*}
+    multi method infer(List:D:) {
+        my $iterator := self.iterator;
+        nqp::if(
+          nqp::eqaddr((my $pulled := $iterator.pull-one),IterationEnd),
+          Nil,                                    # nothing to check
+          nqp::stmts(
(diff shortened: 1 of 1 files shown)


src/core.c/Any-iterable-methods.pm6
```diff
@@ -2357,6 +2357,62 @@ Consider using a block if any of these are necessary for your mapping code."
...
+              self!slow-infer($iterator, $type, $pulled)  # find out what
+            )
+          )
+        )
+    }
+
+    method !slow-infer($iterator, Mu $type is copy, Mu $pulled is copy) {
+        # set up types to check
+        my $mro   := nqp::clone(nqp::getattr($type.^mro,List,'$!reified'));
+        my $roles := nqp::clone(nqp::getattr($type.^roles,List,'$!reified'));
```
vrurg, line 2385: Wouldn't `.^mro(:roles)` simplify the code?

lizmat: That doesn't seem to work?
````
$ raku -e 'dd Int.^mro(:roles)'
Cannot find method 'find_method' on object of type Perl6::Metamodel::ConcreteRoleHOW
````

Comment (lizmat) [reply src/core.c/Any-iterable-methods.pm6:2385]: If that would work, and the roles be in the right order, yes, that would simplify the code quite a bit  :-)

**Comment by ugexe** [comment]
I agree with choosing a better name

**Comment by lizmat** [comment]
`@a.tightest-common-type` ??

File: src/core.c/Any-iterable-methods.pm6 (hunk 1 above)
Comment (vrurg) [reply src/core.c/Any-iterable-methods.pm6:2385]: It works, but - concrete role... Hm, they're functionally limited, that's true. Considering the most popular pattern of `mro` use, I should probably rework it to return a parametric group.

Comment (vrurg) [reply src/core.c/Any-iterable-methods.pm6:2385]: > If that would work, and the roles be in the right order, yes, that would simplify the code quite a bit :-)

The order is right, `Int.^mro(:roles).map(*.^name)` gives `(Int Real Numeric Cool Any Mu)`. But there is a little problem. This parameter's primary purpose is to support 6.e semantics of submethods. It has to provide a list of typeobjects suitable for looking up submethods in them. Concretizations are the best candidates for this.

Perhaps it'd be better to have two named variants for each purpose: `:roles` and `:concretizations`. I need to think over it.

**Comment by lizmat** [comment]
or maybe just `@a.common-type` ?

**Comment by jnthn** [comment]
> @a.tightest-common-type

Too long, I'd say. The method was proposed as a replacement for the `.all ~~ Int` idiom, which is at least a threatened species at this point. But further to that, we've also sometimes pondered trying to make the typed/untyped boundary softer, and one of the ways to help with that is list/array/hash literals that would take on a type based upon the values they are constructed with - that is, the type that we would `infer`, and thus the name (that is to say, the semantics for such a thing would be identical to those of the `infer` proposed here).

Mostly I dislike `infer` in so far as Raku has generally tried to avoid going with "academic-y" names rather than those that read nicely, and `.all ~~ Int` reads a bit nicer than `.infer ~~ Int`. In `infer`'s slight favor, "type inference" is at least probably a well known term by this point.

File: src/core.c/Any-iterable-methods.pm6 (hunk 1 above)
Comment (vrurg) [reply src/core.c/Any-iterable-methods.pm6:2385]: @lizmat  #4635 implements more relevant `.^mro(:roles)` semantics.

Comment (vrurg) [reply src/core.c/Any-iterable-methods.pm6:2385]: @lizmat #4635 is merged, you could try simplifying your algorithm.

**Review by vrurg — APPROVED** [review]

**Comment by raiph** [comment]
@jnthn wrote:

> the `.all ~~ Int` idiom, which is at least a threatened species at this point.

If the threat is entirely due to optimization problems you feel should be eliminated, then fair enough.

To the degree it's wholly, or even in part, due to the 
https://github.com/Raku/problem-solving/issues/297 issue I filed, I want you to know that I currently think Larry did *not* threaten that idiom. Instead I think it was *my* accidental thinko / misrepresentation of what he wrote that led to some thinking that to be so.

I've now edited the issue's opening comment to try clarify things. I apologize for my original mistakes and any in my latest edit.

**Comment by jnthn** [comment]
> If the threat is entirely due to optimization problems you feel should be eliminated, then fair enough.

Well, performance wise, ideally, we'd be able to compile `$foo ~~ SomeType` (where we statically know `SomeType` resolves to a type object) into the `nqp::istype($foo, SomeType)` primitive, modulo boolifying the result properly if needed. Either:

1. The LHS of `~~` is not threaded through the `Junction`, so we can do this. This implies `.all ~~ Int` would not work because it's like asking "is a `Junction` of type `Int`", and the answer is no. Thus the idiom would cease to work. I thought enabling such optimization was the crux of the issue.
2. The LHS of `~~` is intended to thread the type check through the `Junction` values. In that case, we're largely left with runtime optimization.

It's not like we don't have a lot of good infrastructure for runtime optimization now, but all of it implies some warm-up cost, for example.

**Comment by raiph** [comment]
@jnthn

> Either:
>
> The LHS of `~~` is not threaded ... `.all ~~ Int` would ... cease to work. I thought enabling such optimization was the crux of the issue.

I am 99.9% sure that what you've just written reflects a misunderstanding that **I** accidentally created, which I'm now trying to undo.

First, yes to Larry trying to *reinforce* the need to pay attention to *optimizability*, and that being a primary reason for why he said what he said, namely to not auto-thread the LHS.

But he did *not* write, or appear to me (as I *now* read his words) to imply, that the idiom `.all ~~ Type` would stop working. Further, given what he wrote in the email thread in May/June, which specifically focused on a smart matching invocant that was an *instance* object, quite conspicuously (at least to me, in retrospect) *not* a type object, I think it's near certain *he was intending that `foo ~~ Type`  would stay working*.

What I think he was trying to get at is his view that we should ensure Raku compilers can still take advantage of the optimizability that had clearly been envisaged in the original smart matching design, as explicitly mentioned in S03's [Smart matching](https://design.raku.org/S03.html#Smart_matching) section:

> The optimizer is allowed to assume that no additional match operators are defined after compile time, so if the pattern types are evident at compile time, the jump table can be optimized.

And I think, in his 2021 comments, he meant to focus on that optimizability still being available for *instance* invocants of `ACCEPTS`, even while he'd taken it away for type objects in 2015.

That is, I think he decided in 2015 to make `foo ~~ Type` work, at the expense of optimization in that case, presumably as an *exception* that proves the rule that *in general* smart matching needed to not auto-thread (with the rationale being to ensure that, *in general*, it would be optimizable). Hence, I think, his qualifier "In general" in his sentence "In general, the left side of `~~` should not be auto-threading".

----

Are you aware of / interested in Larry's 2015 change that enabled auto-threading of the junction in `junction ~~ Type`? This middle section of my comment covers that.

I discovered my mistake about `foo ~~ Bar` (where `Bar` is a type object) due to codesection's excellent detective work. He showed that it was Larry himself who talked about, and implemented, the idiom `foo ~~ Type`, where it auto-threads `foo`, in 2015.

He modified roast, and Rakudo, based on reasoning he first suggested in [a very brief IRC exchange with Carl Masak](https://logs.liz.nl/perl6/2015-06-14.html#19:18-0001). It should take you no more than a minute or two to read and understand that exchange if you care to read it.

The next day he made the change by landing [this commit](https://github.com/rakudo/rakudo/commit/28a76979d9):

> Any types autothread when smartmatching junctions
> This is probably how most people will expect smartmatching to work.

```
Mu.pm6:

    proto method ACCEPTS(|) { * }
+   multi method ACCEPTS(Mu:U: Any \topic) {
+       nqp::p6bool(nqp::istype(topic, self))
+   }
+   multi method ACCEPTS(Mu:U: Mu:U \topic) {
        nqp::p6bool(nqp::istype(topic, self))
    }

Junction.pm6:

+   multi method ACCEPTS(Junction:U: Mu:D \topic) {
+       nqp::p6bool(nqp::istype(topic, Junction));
+   }
+   multi method ACCEPTS(Junction:U: Any \topic) {
+       nqp::p6bool(nqp::istype(topic, Junction));
+   }

Any.pm6:

    multi method ACCEPTS(Any:D: Mu \a) { self === a }
+   multi method ACCEPTS(Any:U: Any \topic) { # use of Any on topic to force autothreading
+       nqp::p6bool(nqp::istype(topic, self)) # so that all(@foo) ~~ Type works as expected
+   }
 
The code in his commit has, since 2015, changed, per current master, as shown in the following blames. (All changes are in core.c; core.d doesn't contain any of these three modules, and neither currently does core.e.)

* [Mu.pm6](https://github.com/rakudo/rakudo/blame/master/src/core.c/Mu.pm6#L19-L31)

* [Junction.pm6](https://github.com/rakudo/rakudo/blame/master/src/core.c/Junction.pm6#L179-L235)

* [Any.pm6](https://github.com/rakudo/rakudo/blame/master/src/core.c/Any.pm6#L18-L21)

----

> [Or:]
>
> The LHS of `~~` is intended to thread the type check through the Junction values. In that case, we're largely left with runtime optimization.

Aiui Larry was deliberately accepting that slow down to runtime opts for an `ACCEPTS` invocant that's a type object, but *not* for an *instance* invocant.

That said, I note that, in what he wrote in 2021, Larry carefully did not exclude the possibility that his decision in 2015 (to make an exception for type object invocants) would be unwound. He avoided touching on it, and perhaps that was very deliberate.

lizmat:
`@a.tct` ??  in the same vein as `gcd` and `lcm` ?

raiph:
I had no idea what `tct` stood for despite the "common" hint via `gcd` and `lcm`. I concluded you might have gotten confused and posted your comment in the wrong issue. I considered writing a comment to say as much but decided I should read all the comments in this one to be sure it made no sense. Then, when I saw your `.tightest-common-type` suggestion it instantly *did* make sense. :)

Still, I wondered if I could come up with something better. What about:
```
# Already works:
say 42        .isa: Int;         # True

# Suggestion along the same lines:
say (42, 0)   .are: Int;         # True
say (42, 1.5) .are;              # (Cool)
say <1 a>     .are;              # (Str)

# Already works:
say (@)       .does: Positional; # True
say (@)       .does: Iterable;   # True

# Suggestion along the same lines:
say (@,%)     .do: Iterable;     # True
say (@,%)     .do: Positional;   # False
say (@,%)     .do;               # (Iterable)
say (42, 1.5) .do: Numeric;      # True
say (42, 1.5) .do;               # (Real)

# For `are/do` combination aka `.tct`:
say 42        .types;            # (Int Real)
say (42, 1.5) .types;            # (Cool Real)
say (42, 1e0) .types;            # (Cool Numeric)
say <1 a>     .types;            # (Str)
say (@)       .types;            # (Any Iterable Positional)
say (@,@)     .types;            # (Any Iterable Positional)
say (@,%)     .types;            # (Any Iterable)
```

lizmat:
Since a lot of discussion was a. about the name, and b. about maybe needing additional functionality, I'm going to merge this as is.  Slightly academically, but the "type inference" idiom is good enough for me.

And we still have a few weeks to revert should we decide it shouldn't be in core anyway.

vrurg:
@lizmat BTW, @raiph suggestion about `are` as a development of `isa` appeals to me. :) Not the parameterized version though as with my recent fixes to smartmatching logic `@types.all ~~ Int` is now the right approach again.

codesections:
I'll add another +1 for `are` (and a corresponding -1 for `infer`).

I have two concerns with `infer`: First, although I agree with @jnthn that "type inference" is a fairly well-known term, I'm not sure that `Seq.infer` will make everyone think of _type_ inference, instead of anything else they might want to infer.  (The next item in a geometric sequence, perhaps?).

Second, even for people who get the "type inference" connection, I don't think that "inferring the tightest common type" is the only sort of type inference that'd come to mind. Personally, my first thought would be that it'd be type inference as in the below:

```raku
sub takes-a-list(List $l) {}
takes-a-list (^3).map(*+1).infer;
```

(That is, that `.infer` would infer what iterable is needed and cast the current iterable to that type (cf. Rust's [collect](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.collect) method).

In contrast, `.are` avoids these issues.  However, I don't think we need the corresponding `.do` method (which risks confusion with `do` blocks); I think `are` can apply to both Classes and Roles.  In fact, imo it reads pretty well for (many) Roles: `([1, 2], <3 4>, (5, 6)).are: Positional`.

lizmat:
So done with https://github.com/rakudo/rakudo/commit/186bd0b026

PR #4629, opened 2021-11-09
New disp nativecall

niner:
This is a dispatcher based rework of how native routines are called, replacing the runtime compiled native call sub bodies. As much work as possible like unboxing arguments is pushed into the HLL dispatcher code where it can be speshed away. JIT compilation of native calls is completely re-done to be more in line with other ops. We no longer JIT compile calls when building the native call site, but let spesh handle this via the new sp_runnativecall famliy of ops which doesn't require any special handling by HLL code.

Requires MoarVM's new-disp-nativecall-libffi branch and nqp's new-disp-nativecall branch


lib/NativeCall.rakumod
```diff
@@ -260,6 +263,144 @@ sub resolve-libname($libname) {
     $libname.platform-library-name.Str
 }
 
+use MONKEY-SEE-NO-EVAL;
+my $use-dispatcher = so $*RAKU.compiler.?supports-op('dispatch_v') && EVAL q:to/CODE/;
```
jnthn, line 267: Could the `EVAL` be done at `BEGIN` time, so we don't have to do the parse/compile of this every time `NativeCall` is first used in a program?

lib/NativeCall.rakumod
```diff
@@ -260,6 +263,144 @@ sub resolve-libname($libname) {
...
+    use nqp;
+    sub raku-native-dispatch-deproxy($callee, *@args) {
+        $callee(|@args) # implicitly fetches values from proxies
+    }
```
jnthn, line 271: Uh, that's clever! Not so runtime efficient as the proxy reader approach, but... :-) As a nitpick, I wonder if `**` would be better (so we don't flatten anything away)?

lib/NativeCall.rakumod
```diff
@@ -260,6 +263,144 @@ sub resolve-libname($libname) {
...
+    my &raku-nativecall := -> $capture is raw {
+        my $track_callee := nqp::dispatch('boot-syscall', 'dispatcher-track-arg', $capture, 0);
+        nqp::dispatch('boot-syscall', 'dispatcher-guard-literal', $track_callee);
+        my $callee := nqp::captureposarg($capture, 0);
+        $callee.setup;
```
jnthn, line 276: Ahh, the things we get to lift out of every call... :heart_eyes_cat:

lib/NativeCall.rakumod
```diff
@@ -260,6 +263,144 @@ sub resolve-libname($libname) {
...
+
+        my Mu $args := nqp::dispatch('boot-syscall', 'dispatcher-drop-arg', $capture, 0);
+        my $signature := nqp::getattr($callee, Code, '$!signature');
+        my int $readonly = nqp::getattr_i($signature, Signature, '$!readonly');
+        my int $pos-args = nqp::captureposelems($args);
+        my int $i = 0;
+        my int $continue = 1;
```
jnthn, line 283: I wonder if `$no-proxy` would be more descriptive, given that's what this is for.

lib/NativeCall.rakumod
```diff
@@ -260,6 +263,144 @@ sub resolve-libname($libname) {
...
+            # If it should be passed read only, and it's an object...
+            if nqp::captureposprimspec($args, $i) == 0 {
+                # If it's in a Scalar container...
+                my $track-arg := nqp::dispatch('boot-syscall', 'dispatcher-track-arg', $args, nqp::unbox_i($i));
+                nqp::dispatch('boot-syscall', 'dispatcher-guard-type', $track-arg);
+                nqp::dispatch('boot-syscall', 'dispatcher-guard-concreteness', $track-arg);
+                my $arg := nqp::captureposarg($args, $i);
+                my $track-value;
+                my $cstr = False;
+                if nqp::isconcrete_nd($arg) && nqp::what_nd($arg) =:= Scalar {
```
jnthn, line 294: I wonder it we should permit subclasses of `Scalar` too (make it `nqp::istype_nd($arg, Scalar)`).

niner: If we permit subclasses here, we should probably permit them in pass-decontainerized (where I stole this from) as well: https://github.com/rakudo/rakudo/blob/master/src/vm/moar/dispatchers.nqp#L2873

jnthn: Fair. Though in the place you stole it from it's an optimization issue (worst case is the callee does the deconting), here I think it's a correctness one (there's nothing else to do the deconting).

PR #4630, opened 2021-11-09
Add Any(iterable).isa-all(type) method

lizmat:
This returns a boolean whether all values produced by the iterable match the given type.  Intended to be used in e.g. constraints such as:
````raku
sub foo(@a where .isa-all(Int))
````
as an alternative to:
````raku
sub foo(@a where .all ~~ Int)
````
The .isa-all method is about 3.5x as fast as the `.all ~~ Int` approach for an array consisting of 10 `Int`s.  The efficiency increase for non-matching types will be *much* bigger, as the `isa-all` method can short-circuit on any non-matching value, which the `.all ~~ Int` approach can not.  So that would be *at least* 3.5x as fast, but potentially orders of magnitude faster for bigger arrays.

### DIFF
--- a/src/core.c/Any-iterable-methods.pm6
+++ b/src/core.c/Any-iterable-methods.pm6
@@ -2357,6 +2357,17 @@ Consider using a block if any of these are necessary for your mapping code."
         Seq.new(Rakudo::Iterator.Rotor(self.iterator,@cycle,$partial))
     }
 
+    proto method isa-all(|) {*}
+    multi method isa-all(Any:D: Mu $type --> Bool:D) {
+        my $iterator := self.iterator;
+        nqp::until(
+          nqp::eqaddr((my $pulled := $iterator.pull-one),IterationEnd)
+            || nqp::not_i(nqp::istype($pulled.WHAT,$type)),
+          nqp::null
+        );
+        nqp::hllbool(nqp::eqaddr($pulled,IterationEnd))
+    }
+
     proto method nodemap(|) is nodal {*}
     multi method nodemap(Associative:D: &op) {
         self.new.STORE: self.keys, self.values.nodemap(&op), :INITIALIZE


lizmat:
Alternate name: `all-is-type`

codesections:
A technical question: when you say that 

>  The efficiency increase for non-matching types will be much bigger, as the `isa-all` method can short-circuit on any non-matching value, which the `.all ~~ Int` approach can not.

Is that a limitation of the _current_ nature of Junctions, or something inherent to their design?  My understanding is that, once auto-threading is fully enabled, then an `.all` junction would short-circuit – which is to say that it would evaluate the LHS in parallel, and would break out of that evaluation as soon as it found the first non-`Int`.  That understanding is from [S03](https://design.raku.org/S03.html#Junctive_operators), which says:

> They can be evaluated in any order or in parallel. They can short-circuit as soon as any of them return 42, and not run the others at all. Or if running in parallel, the first successful thread may terminate the other threads abruptly. In general you probably want to avoid code with side effects in junctions.

I know that the Synopses aren't kept up to date; is that no longer the plan?  Or would this be about getting a performance win for the period between now and when full autothreading (and the short-circuiting behavior that comes with it) can be implemented?

I also have some bigger-picture concerns with this series of PRs, but I think those might be more appropriate for a problem solving thread, so I'll hold off on then here.

lizmat:
The (current) problem is really that when you do `.all ~~ Int`, you create a `Junction` of `Bool`s, and only when you collapse this, can it short-circuit.  In other words, the threading of `~~` does not currently know it needs to collapse to a `Bool` at the end.

With RakuAST we could probably create better code-gen that *would* allow us to short-circuit the smart-matching for cases as `where`, `when` and `if`.

CIAvash:
> Alternate name: `all-is-type`

I know we have `isa` method, but to me `is-all` reads better.

patrickbkr:
Generally I oppose adding bits to our API solely to work around performance issues. Fast forward a few years and our junction support is optimized enough to not need this method anymore. But then we can't get rid of this method again, because people are then relying on it. So in the long term we inflate our API with dead weight that lowers the elegance and approachability of Raku.

PR #4631, opened 2021-11-09
Add Any(iterable).isa-any(type) method

lizmat:
This returns a boolean whether *any* values produced by the iterable match the given type. Intended to be used in e.g. constraints such as:
````raku
sub foo(@a where .isa-any(Int))
````
as an alternative to:
````raku
sub foo(@a where .any ~~ Int)
````
This is the logical counterpart of `Any.isa-all` from https://github.com/rakudo/rakudo/pull/4630


lizmat:
Alternate name: `any-is-type`

PR #4632, opened 2021-11-09
Add Any(iterable).isa-none(type) method

lizmat:
This returns a boolean whether none of the values produced by the iterable
match the given type. Intended to be used in e.g. constraints such as:
````raku
sub foo(@a where .isa-none(Str))
````
as an alternative to:
````raku
sub foo(@a where .none ~~ Str)
````
This is the logical counterpart of `Any.isa-any` from https://github.com/rakudo/rakudo/pull/4631


lizmat:
Alternate name: `none-is-type` ?

PR #4634, opened 2021-11-09
Make concurrent access to an enum by value safe

MasterDuke17:
Previously multiple threads all trying to get an enum key by value could
race to modify `%!value_to_enum`, which would cause an `MVM_oops`.

Also make concurrently adding a value to an enum safe in the same
manner.

### DIFF
--- a/src/Perl6/Metamodel/EnumHOW.nqp
+++ b/src/Perl6/Metamodel/EnumHOW.nqp
@@ -25,7 +25,7 @@ class Perl6::Metamodel::EnumHOW
     has %!values;
 
     # Reverse mapping hash.
-    has %!value_to_enum;
+    has $!value_to_enum;
 
     # List of enum values (actual enum objects).
     has @!enum_value_list;
@@ -70,6 +70,9 @@ class Perl6::Metamodel::EnumHOW
     method add_enum_value($obj, $value) {
         %!values{nqp::unbox_s($value.key)} := $value.value;
         @!enum_value_list[+@!enum_value_list] := $value;
+        nqp::scwbdisable();
+        $!value_to_enum := NQPMu;
+        nqp::scwbenable();
     }
 
     method set_export_callback($obj, $callback) {
@@ -85,13 +88,18 @@ class Perl6::Metamodel::EnumHOW
     }
 
     method enum_from_value($obj, $value) {
-        unless %!value_to_enum {
+        my $value_to_enum := $!value_to_enum;
+        unless $value_to_enum {
+            $value_to_enum := nqp::hash;
             for @!enum_value_list {
-                %!value_to_enum{$_.value} := $_;
+                $value_to_enum{$_.value} := $_;
             }
+            nqp::scwbdisable();
+            $!value_to_enum := $value_to_enum;
+            nqp::scwbenable();
         }
-        nqp::existskey(%!value_to_enum, $value)
-            ?? %!value_to_enum{$value}
+        nqp::existskey($value_to_enum, $value)
+            ?? $value_to_enum{$value}
             !! nqp::null()
     }
 


niner:
Does the lookup hash actually provide performance benefits over just iterating through the value list? The lock will cost quite a bit and I assume that most Enums will have a rather short list of values. Even hash access by itself is more costly than array access, so e.g. for Bool with just 2 values and a 50 % success rate on first try, iteration must be faster.

MasterDuke17:
I just tried this patch:
```
diff --git src/Perl6/Metamodel/EnumHOW.nqp src/Perl6/Metamodel/EnumHOW.nqp
index 8cf0c042b..7e34bc211 100644
--- src/Perl6/Metamodel/EnumHOW.nqp
+++ src/Perl6/Metamodel/EnumHOW.nqp
@@ -24,9 +24,6 @@ class Perl6::Metamodel::EnumHOW
     # Hash representing enumeration keys to values.
     has %!values;
 
-    # Reverse mapping hash.
-    has %!value_to_enum;
-
     # List of enum values (actual enum objects).
     has @!enum_value_list;
 
@@ -85,14 +82,10 @@ class Perl6::Metamodel::EnumHOW
     }
 
     method enum_from_value($obj, $value) {
-        unless %!value_to_enum {
-            for @!enum_value_list {
-                %!value_to_enum{$_.value} := $_;
-            }
+        for @!enum_value_list {
+            return $_ if $_.value eq $value;
         }
-        nqp::existskey(%!value_to_enum, $value)
-            ?? %!value_to_enum{$value}
-            !! nqp::null()
+        nqp::null()
     }
 
     method enum_value_list($obj) {
```
For this example: `MVM_SPESH_BLOCKING=1 raku -e 'my $a; for ^1_000 { my enum E <a b c d>; race for ^1_000 { $a := E($_ % 4) }; }; say now - INIT now; say $a'` the current PR is essentially the same, both are ~3.3s.
For this example: `MVM_SPESH_BLOCKING=1 raku -e 'my $a; for ^400 { my enum E ("a".."zz"); race for ^400 { $a := E($_ % 702) }; }; say now - INIT now; say $a'` the current PR is ~1s and the above patch is ~7.2s.

niner (approved):

src/Perl6/Metamodel/EnumHOW.nqp
```diff
@@ -85,11 +96,17 @@ class Perl6::Metamodel::EnumHOW
...
+            unless %value_to_enum {
+                for @!enum_value_list {
+                    %value_to_enum{$_.value} := $_;
+                }
+                nqp::scwbdisable();
+                %!value_to_enum := %value_to_enum;
+                nqp::scwbenable();
             }
-        }
+        };
```
jnthn, line 109: I think a better fix here would be to:

1. Switch to a `$` attribute
2. Create the hash in a lexical
3. Bind the hash into the attribute

Of course two threads may produce it in duplicate, but that's not a correctness issue, and it's almost certainly still more efficient than having a Global Enum Lock.

src/Perl6/Metamodel/EnumHOW.nqp
```diff
@@ -70,6 +72,15 @@ class Perl6::Metamodel::EnumHOW
...
         %!values{nqp::unbox_s($value.key)} := $value.value;
         @!enum_value_list[+@!enum_value_list] := $value;
+        my %value_to_enum := %!value_to_enum;
+        $lock.protect: {
+            if %value_to_enum {
+                %value_to_enum{$value.value} := $value.key;
+            }
+            nqp::scwbdisable();
+            %!value_to_enum := %value_to_enum;
+            nqp::scwbenable();
```
jnthn, line 82: And here you'd just reset the `$!value_to_enum` to `NQPMu` and let it be regenerated later.

jnthn (changes requested):
I think this can be done without a lock, which will introduce contention over all threads that try to coerce with any (even different) enums...

src/Perl6/Metamodel/EnumHOW.nqp
```diff
@@ -85,13 +88,17 @@ class Perl6::Metamodel::EnumHOW
...
+                $value_to_enum{$_.value} := $_;
             }
+            nqp::scwbdisable();
+            $!value_to_enum := $value_to_enum;
+            nqp::scwbenable();
         }
-        nqp::existskey(%!value_to_enum, $value)
-            ?? %!value_to_enum{$value}
+        nqp::existskey($!value_to_enum, $value)
+            ?? $!value_to_enum{$value}
```
jnthn, line 101: Ah, I should have added: it's important only read the `$!value_to_enum` attribute once into a lexical, not multiple times, otherwise it could have `NQPMu` assigned along the way. That won't make a SEGV, but it would make an exception.

MasterDuke17:
@jnthn I think I've addressed your comments. I just did a couple rough timing (on my laptop now, so numbers might not be directly comparable to the ones I previously reported).

`MVM_SPESH_BLOCKING=1 raku -e 'my $a; for ^1_000 { my enum E <a b c d>; race for ^1_000 { $a := E($_ % 4) }; }; say now - INIT now; say $a'` is now essentially the same as master (~2.7s).
`MVM_SPESH_BLOCKING=1 raku -e 'my $a; for ^400 { my enum E ("a".."zz"); race for ^400 { $a := E($_ % 702) }; }; say now - INIT now; say $a'` is now essentially the same as master (~1s).

When this is deemed correct I'll either push a squashed rebase or just squash when merging.

src/Perl6/Metamodel/EnumHOW.nqp
```diff
@@ -85,13 +88,18 @@ class Perl6::Metamodel::EnumHOW
     }
 
     method enum_from_value($obj, $value) {
-        unless %!value_to_enum {
+        my $value_to_enum := $!value_to_enum;
+        unless $value_to_enum {
+            my $value_to_enum := nqp::hash;
```
jnthn, line 93: I don't think this should be redeclared `my` here, otherwise the outer `$value_to_enum` will be `NQPMu` the first time this is called (I'm surprised it doesn't crash in some way). Otherwise, it's looking fine.

src/Perl6/Metamodel/EnumHOW.nqp
```diff
@@ -85,13 +88,18 @@ class Perl6::Metamodel::EnumHOW
     }
 
     method enum_from_value($obj, $value) {
-        unless %!value_to_enum {
+        my $value_to_enum := $!value_to_enum;
+        unless $value_to_enum {
+            $value_to_enum := nqp::hash;
```
jnthn, line 93: Hm, I thought in NQP it had to be `nqp::hash()`, but apparently it somehow works anyway.

src/Perl6/Metamodel/EnumHOW.nqp
jnthn: Hah, yes, apparently it's fine!

```
$ nqp -e 'my $x := nqp::hash; say($x.HOW.name($x))'
BOOTHash
```

jnthn (approved):
LGTM, though please squash for cleaner history.

PR #4635, opened 2021-11-10
Get either parametric or concrete roles in MRO when requested

vrurg:
Change semantics of `:roles` named parameter to returning parametric
groups instead of concretizations to make it conform to `.^roles`
methods.

Add `:concretizations` named for cases when concretizations are more
useful. For example, for methods traversal by `WALK`.


Kaiepi:
I think `Metamodel::ParametricRoleHOW.mro` would need an update for this.

vrurg:
@Kaiepi does 5794a9b address your concern? 

Thanks!

Kaiepi:
@vrurg yep!

PR #4636, opened 2021-11-11
Raku on windows

tbrowder:
Start an installation and usage guide for Raku on Windows.


tools/raku-on-windows/raku-on-windows.pod6
```diff
@@ -0,0 +1,65 @@
...
+                ;C:\Program Files\Git\usr\bin
+=end code
+
+Note: any command terminals opened before the installation will have to be reopened to update the path to the Git commands.
+
+
+## Install Perl 6 / Raku:
+
+Install Raku (Perl 6) for Windows:
+Note: only the 64 bit version is supported.  For a 32 bit version, as the mailing list for help.
```
coke, line 59: "ask" the mailing list. Also, I'd remove this note and say that 64 bit is the supported version; if we have instructions, let's add them; if not, I don't think the mailing list is going to help.

WINDOWS-INSTALL.md
```diff
@@ -0,0 +1,63 @@
...
+
+![Fig. 1](/tools/raku-on-windows/pics/Raku-WinInstall-fig-1-git-setup.png)
+
+    2) Or, after the fact:
+       --> <win><r> sysdm.cpl
+         --> Advanced (tab at the top)
+          --> Environmental Variables (button, lower right)
+           --> Path, Edit, add to the end
+                    ;C:\Program Files\Git\usr\bin
+
```
coke, line 37: I use git bash when on windows and I don't recall ever having to manually adjust my path, I got it for free as part of the install.

tools/raku-on-windows/raku-on-windows.pod6
tbrowder: Feel free to edit as necessary. This first chunk was a rough translation from Todd's input and I haven't had a chance to do anything else with it.

WINDOWS-INSTALL.md
tbrowder: Ditto from my last comment.

PR #4646, opened 2021-11-19
MAIN default values

Xliff:
-  Adds the ability to display any default values in sub MAIN to the default USAGE message.


src/core.c/Main.pm6
```diff
@@ -302,7 +302,33 @@ my sub RUN-MAIN(&main, $mainline, :$in-as-argsfiles) {
...
                 }
-                @arg-help.push($argument => $param.WHY.contents) if $param.WHY and (@arg-help.grep:{ .key eq $argument}) == Empty;  # Use first defined
+            		#@arg-help.push($argument => $param.WHY.contents) if $param.WHY and (@arg-help.grep:{ .key eq $argument}) == Empty;  # Use first defined
+                if $param.WHY and (@arg-help.grep:{ .key eq $argument}) == Empty {
+                    my $why = $param.WHY.contents; # Use first defined
+                    if $param.default -> $d {
+                          constant MAXCHARS = 20;
+                          # $middle is for long integers and other argument types
+                          # that are better split in the middle.
+                          if ( my $def = $d() ).defined {
```
lizmat, line 312: Perhaps better written as `with $d() -> $def {` ?

lizmat:
Cool!

Could you give some before / after examples?

src/core.c/Main.pm6
Xliff: Done specifically to handle when a value is defined, but Falsey (ala 0, False)

vrurg: @Xliff that's exactly what you get with `with`: `my $a = 0; say $_ with $a`

Xliff:
As requested:

Before:

```
Usage:
  test-args.pl6 [--filename[=Any]] [--integer[=Any]] [--boolean[=Any]] [--numeric[=Any]] <positional>
  
    --filename[=Any]    A filename
    --integer[=Any]     An integer value
    --boolean[=Any]     A flag
    --numeric[=Any]     Defaults to a food group!
```

After:

```
Usage:
  test-args.pl6 [--filename[=Any]] [--integer[=Any]] [--boolean[=Any]] [--numeric[=Any]] <positional>
  
    --filename[=Any]    A filename [default: '/local/test.raku']
    --integer[=Any]     An integer value [default: 1]
    --boolean[=Any]     A flag [default: False]
    --numeric[=Any]     Defaults to a food group! [default: 3.141592653589793]
```
Script:

```
sub MAIN ( 
        $positional,
        :$filename = '/local/test.raku',  #= A filename
        :$integer  = 1,                   #= An integer value
        :$boolean  = False,               #= A flag
        :$numeric  = π                    #= Defaults to a food group!
) {
}
```

As you can see, to trigger the new behavior, you have to use trailing comments, otherwise the defaults have nothing to latch on to.

Without trailing comments, the behavior is the same as stock rakudo:

```
Usage:
  test-no-args.pl6 [--filename[=Any]] [--integer[=Any]] [--boolean[=Any]] [--numeric[=Any]] <positional>
```

lizmat:
@jdv Are you ok with merging this before the 2021.12 release?

jnthn:
> Without trailing comments, the behavior is the same as stock rakudo

Finding a way to do the defaults could be useful in the non-doc-comment case too, but certainly I think it's a nice improvement as is. :+1:

Issue #4647, opened 2021-11-19
`Type check failed in binding to parameter` for a named param that has not been passed


patrickbkr:
## The Problem

    sub s(Str :@l, Str :$a) { }; s(:a(1));

## Actual Behavior

    Type check failed in binding to parameter '@l'; expected Positional[Str] but got Array ([])

## Expected Behavior

   Type check failed in binding to parameter '$a'; expected Str but got Int (1)

## Environment
* Compiler version (`perl6 -v` or `raku -v`): v2021.10-91-g4c238bdc9

jnthn:
I believe this needs an matching change in `lower_signature` in `Perl6::Actions`, otherwise it won't fix any of the compiled cases (and thus probably not the original issue).

vrurg:
@jnthn guess, you should've commented on the PR?

jnthn:
@vrurg You are right! Thanks, will add it there.

PR #4648, opened 2021-11-19
Consider Positional/Associative type parameters in optional defaults

Kaiepi:
We need to copy type parameters for this; thankfully, the type of the `@`/`%` must be `Positional`/`Associative`, whose parameters correspond to `Array`/`Hash`'s, so methods for these are already defined.

Allows the following signatures to produce a typechecking error for `$a` instead of `@l` or `%h`:

```raku
:(Str :@l, Str :$a)
:(Str :%h, Str :$a)
:(Str @l?, Str :$a)
:(Str %h?, Str :$a)
```

Preserving:

```raku
:(:@l, Str :$a)
:(:%h, Str :$a)
```

No new tests as of yet.

Fixes https://github.com/rakudo/rakudo/issues/4647.

### DIFF
--- a/src/Perl6/bootstrap.c/BOOTSTRAP.nqp
+++ b/src/Perl6/bootstrap.c/BOOTSTRAP.nqp
@@ -662,14 +662,15 @@ my class Binder {
 
         # Otherwise, go by sigil to pick the correct default type of value.
         else {
+            my $type := nqp::getattr($param, Parameter, '$!type');
             if $flags +& $SIG_ELEM_ARRAY_SIGIL {
-                nqp::create(Array)
+                nqp::create($type =:= Positional ?? Array !! Array.HOW.parameterize(Array, $type.of));
             }
             elsif $flags +& $SIG_ELEM_HASH_SIGIL {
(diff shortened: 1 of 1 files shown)


niner (approved):

jnthn:
I believe this needs an matching change in lower_signature in Perl6::Actions, otherwise it won't fix any of the compiled cases (and thus probably not the original issue).

lizmat (changes requested):
@jnthn suggested that this needs some more tweaks.

lizmat:
@vrurg is this PR now still a WIP or not?

vrurg:
@Kaiepi I would only appreciate if you do it.

Kaiepi:
There already is code for this in `lower_signature`, no?

https://github.com/rakudo/rakudo/blob/d7d8e9d0f41e7316c872d4e33820b4a7e56b0aa2/src/Perl6/Actions.nqp?plain=1#L9728-L9741

 I remember trying to make that consistent with this before my break, but I found that was depended on before `Positional` was defined. Should this match that instead?

PR #4649, opened 2021-11-21
Make class A does B{ } error more awesome

lizmat:
Adds the line "Perhaps you forgot a space before the { ?" to the
otherwise rather obscure error message caused by lack of whitespace
between the role name and the opening curly

    $ raku -e 'role A{ ... }; class B does A{ ... }'
    Autovivifying object closures not yet implemented. Sorry.
    Perhaps you forgot a space before the { ?
    at -e:1
    ------> role A{ ... }; class B does A{ ... }⏏<EOL>

Also adds a "perhaps" attribute to X::NYI.

### DIFF
--- a/src/Perl6/Grammar.nqp
+++ b/src/Perl6/Grammar.nqp
@@ -339,8 +339,8 @@ role STD {
     method missing($what) {
         self.typed_panic('X::Syntax::Missing', :$what);
     }
-    method NYI($feature) {
-        self.typed_panic('X::Comp::NYI', :$feature)
+    method NYI($feature, *%opts) {
+        self.typed_panic('X::Comp::NYI', :$feature, |%opts)
     }
 
     token experimental($feature) {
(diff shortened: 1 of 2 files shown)


MasterDuke17 (approved):

MasterDuke17:
I meant to do the same thing, but forgot. It might be nice to get the eject symbol to point at the opening `{` if possible, but this is a marked improvement as is.

lizmat:
@jnthn Guess I am confused by your reply: is this solution good enough for now or not?

jnthn:
@lizmat The hint is nice, although I suspect the `}` going on the closing curly could render the location rather misleading. I'm not against merging it, but suspect the alternative I suggested would be a similar amount of code (just in a different place), nail the location, and not have to mention auto-vivification closures at all, which also probably are a mostly confusing thing to mention.

lizmat:
Re-imagined with https://github.com/rakudo/rakudo/commit/535e015111 so closing now.

PR #4650, opened 2021-11-23
Don't suggest method of the same name as "not found"

MasterDuke17:
Instead of setting the score to 0, don't even add it to the suggestions.
If there are no other suggestions, before we could still use it, leading
to this example `my Int $a; $a.VAR.abs` giving this silly output:
`No such method 'abs' for invocant of type 'Scalar'.  Did you mean 'abs'?`
Now it just says `No such method 'abs' for invocant of type 'Scalar'`

### DIFF
--- a/src/core.c/Exception.pm6
+++ b/src/core.c/Exception.pm6
@@ -215,15 +215,13 @@ my class X::Method::NotFound is Exception {
 
         my $public_suggested = 0;
         sub find_public_suggestion($before, $after --> Nil) {
-            if $before.fc ne $after.fc {
-                my $dist := StrDistance.new(
-                  before => $before.fc,
-                  after  => $after.fc
-                );
-                if $dist <= $max_length {
-                    $public_suggested = 1;
-                    %suggestions{$after} = $dist.Int;
-                }
+            my $dist := StrDistance.new(
+              before => $before.fc,
+              after  => $after.fc
+            );
+            if $dist <= $max_length {
+                $public_suggested = 1;
+                %suggestions{$after} = $dist.Int;
             }
         }
 
@@ -232,6 +230,7 @@ my class X::Method::NotFound is Exception {
             my $invocant_methods :=
               Set.new: $!invocant.^methods(:local).map: { code-name($_) };
 
+            my $identity-found-already = False;
             for $!invocant.^methods(:all) -> $method_candidate {
                 my $method_name := code-name($method_candidate);
                 # GH#1758 do not suggest a submethod from a parent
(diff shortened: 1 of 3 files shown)


lizmat:
FWIW, I feel like this is shuffling something under the rug.

If the same name occurs in the suggestions, it clearly indicates a containerization problem.  So maybe we should add a special line in the case of identity, that the problem may be caused by containerization or lack thereof?

MasterDuke17:
Now the example says:
`No such method 'abs' for invocant of type 'Scalar'.  Found a method of
the same name on another type, perhaps you called it on a container?`

However, if you call `.log` instead of `.abs`, you now get this:
```
No such method 'log' for invocant of type 'Scalar'.  Possible causes
are:
    - Found a method of the same name on another type, perhaps you
    called it on a container?
    - Did you mean 'log2'?
```
instead of:
```
No such method 'log' for invocant of type 'Scalar'.  Did you mean any
of these: 'log', 'log2'?
```
Neither of those are great, since `log2` shouldn't be suggested at all.

MasterDuke17:
I'm not sure why the invocant is an `Int`, but you can see what happens with this example `my Int $a; try { $a.VAR.log; CATCH { default { dd $_ } } }` which outputs `X::Method::NotFound.new(invocant => Int, method => "log", typename => "Scalar", private => Bool::False, addendum => Any, in-class-call => Bool::False)`. FWIW, before 2021.10, `typename` was `"Int"`. Bisectable is having trouble with the new-disp merge, but it seems like that might be the cause.

lizmat:
> Found a method of the same name on another type, perhaps you
    called it on a container?

I would change `it` to `"(insert name here)"` to emphasize.

lizmat:
> Found a method of the same name on another type

also, isn't the problem that there **is** a method on *that* type, but it cannot bind to it because of a container issue?   So it's not another type at all??

MasterDuke17:
> also, isn't the problem that there **is** a method on _that_ type, but it cannot bind to it because of a container issue? So it's not another type at all??

I don't know. I'd say in this example the method is explicitly being called on the container, but maybe it would still make sense to say something like `"Found '$.method' on $!invocant.^name(), but perhaps you called it on a container?"`?

lizmat:
Perhaps add "explicitly" ?   "... but perhaps you called it explicitly on a container?"  ?

vrurg:
Looking at where the exception is called, it is theoretically possible if not to preserve the original container, but at least to detect the presence of it and add a flag.

MasterDuke17:
> Perhaps add "explicitly" ? "... but perhaps you called it explicitly on a container?" ?

What if they somehow didn't *explicitly* call it on a container, but just happened to? Though I can't come up with an example right now...

vrurg:
At least you'd have a good reason to know if it _wasn't_ invoked on a container. Or, more precisely, you'd know if a container _is_ involved.

MasterDuke17:
> Looking at where the exception is called, it is theoretically possible if not to preserve the original container, but at least to detect the presence of it and add a flag.

You mean here https://github.com/rakudo/rakudo/blob/master/src/vm/moar/dispatchers.nqp#L2839-L2848?

vrurg:
> You mean here

In particular. But even where the exception is used in `Mu`, methods throwing it are using `Mu \SELF:` invocant, also allowing to find out about its containerization status.

MasterDuke17:
After 2708289, `my Int $a; $a.VAR.abs` now prints:
```
No such method 'abs' for invocant of type 'Scalar'.  Possible causes
are:
    - You actually called 'abs' on a container, was that what you
    intended?
    - Found 'abs' on types: 'Int', 'Cool'
  in block <unit> at -e line 1
```

niner (approved):

PR #4651, opened 2021-11-23
Speedup Instant.Num by manual inlining

MasterDuke17:
In the (usual) case that $!tai is actually an Int, this 5.5x faster
because it doesn't involve any Rational math.

`my num $a; $a = now.Num for ^1_000_000; say now - INIT now;`
used to take ~2.3s, now it takes ~0.34s.

### DIFF
--- a/src/core.c/Instant.pm6
+++ b/src/core.c/Instant.pm6
@@ -3,7 +3,7 @@ my class DateTime { ... }
 my class Duration {... }
 
 my class Instant is Cool does Real {
-    has Real $.tai is default(0);
+    has Int $.tai is default(0);
       # A linear count of nanoseconds since 1970-01-01T00:00:00Z, plus
       # Rakudo::Internals.initial-offset. Thus, $.tai matches TAI from 1970
       # to the present.
@@ -19,7 +19,7 @@ my class Instant is Cool does Real {
     }
 
     method to-nanos(--> Int:D) {
-        $!tai.Int
+        $!tai
     }
 
     proto method from-posix(|) {*}
@@ -51,9 +51,9 @@ my class Instant is Cool does Real {
     }
     method Bridge(Instant:   --> Num:D) { self.defined ?? self.tai.Bridge !! self.Real::Bridge }
     method Num   (Instant:D: --> Num:D) { nqp::div_n(self.to-nanos.Num, 1000000000e0)          }
-    method Rat   (Instant:D: --> Rat:D) { self.tai        }
-    method Int   (Instant:D: --> Int:D) { self.tai.Int    }
-    method narrow(Instant:D:          ) { self.tai.narrow }
+    method Rat   (Instant:D: --> Rat:D) { self.tai                                             }
+    method Int   (Instant:D: --> Int:D) { self.to-nanos div 1000000000                         }
+    method narrow(Instant:D:          ) { self.tai.narrow                                      }
 
     method Date(Instant:D:     --> Date:D)     { Date.new(self)     }
     method DateTime(Instant:D: --> DateTime:D) { DateTime.new(self) }
--- a/src/core.c/Instant.pm6
+++ b/src/core.c/Instant.pm6
@@ -50,7 +50,7 @@ my class Instant is Cool does Real {
         'Instant.from-posix(' ~ $posix.raku ~ ($flag ?? ',True)' !! ')')
     }
     method Bridge(Instant:   --> Num:D) { self.defined ?? self.tai.Bridge !! self.Real::Bridge }
-    method Num   (Instant:D: --> Num:D) { self.tai.Num    }
+    method Num   (Instant:D: --> Num:D) { nqp::div_n(self.to-nanos.Num, 1000000000e0)          }
     method Rat   (Instant:D: --> Rat:D) { self.tai        }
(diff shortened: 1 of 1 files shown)


src/core.c/Instant.pm6
```diff
@@ -50,7 +50,7 @@ my class Instant is Cool does Real {
         'Instant.from-posix(' ~ $posix.raku ~ ($flag ?? ',True)' !! ')')
     }
     method Bridge(Instant:   --> Num:D) { self.defined ?? self.tai.Bridge !! self.Real::Bridge }
-    method Num   (Instant:D: --> Num:D) { self.tai.Num    }
+    method Num   (Instant:D: --> Num:D) { nqp::div_n(nqp::decont_n($!tai.Num), 1000000000e0)   }
```
lizmat, line 53: Technically, this loses the subclassibility of `Instant`, because you access the attribute directly.  If we go this way, then there are other places where this could be done.  But maybe we shouldn't, and just use `self.tai` like everywhere else here?

MasterDuke17: Then we can't do this optimization. `self.tai` is what causes the slowdown, since it returns a `Rat` and `.Num` on a `Rat` is slower than `.Num` on an `Int`.

vrurg: Ok, then add a method which would return `$!tai` as-is? It'd be even more correct to replace `$!tai` with `$!nano` and use that.

MasterDuke17: Now implemented via `self.to-nanos`

vrurg: Hopefully, optimizer would eventually be capable of reducing double coercion into one.

lizmat:
Hmmm... maybe we should take a step back:

`$!tai` is an implementation detail
`self.tai` is an API

`self.tai` needs to be a `Rat`, but `$!tai` does not.

Is this a correct assumption?

MasterDuke17:
> Is this a correct assumption?

That's my understanding. See https://github.com/rakudo/rakudo/commit/c1f341cc870e95ea3d492ba948ca5350a9048a36. I tried converting `$!tai` to an `Int` first, but whatever I tried didn't work (iirc it was some date-related code that would store a non-`Int` in `$!tai`).

vrurg:
@lizmat looks like it. That's why I suggested `$.nano` to replace `$.tai` as an attribute, not as an API method. Actually, this replacement would allow future optimizer to reduce `$.nano.Num` to `$!nano.Num` (and then even further into a plain NQP op, perhaps) on objects of type `Instance` or its children with no overriding of method `nano`.

raiph:
> That's why I suggested `$.nano` to replace `$.tai` as an attribute, not as an API method.

Isn't any public method an API method?

Aiui, the original design doc vocabulary was that an attribute is **always entirely private** and, regardless of vocabulary, the point is to keep Raku consistent in this regard with [the Actor model](https://en.wikipedia.org/wiki/Actor_model). This isn't a small thing; sustaining it, or failing to do so, will have truly enormous consequences for the future of Raku.

lizmat:
So, if we change the attribute to `$!nano`, and have a method `tai` that returns a `Rat`, that should preserve API compatibility, does it not?

MasterDuke17:
Wouldn’t it also need a custom .new()? Hm, I typed that and then immediately remembered Instant dies if you try to .new(). But that’s what ISTR when I tried to make $.tai an Int, that somewhere (in the Date code perhaps) Instants were being created by being given a Rat. Maybe it was something messing with internals that could/should be converted to use public api methods though.

lizmat:
If it's in the internals, we adapt the internals.  :-)

MasterDuke17:
After a quick grep through Rakudo and roast, maybe the more problematic one was Duration's `$.tai`.

MasterDuke17:
4985f8a passes `make m-test m-spectest`, but doesn't really make this PR any faster on the example above. But it's probably still a good change to align the type with how it's really used (inside the class).

Kaiepi:
Can `int` help anywhere here?

MasterDuke17:
> Can `int` help anywhere here?

I considered it, but thought that since we're dealing with such large numbers, we would potentially overflow even a `uint`. 2**64 nanoseconds is only ~584 years. We have some time before `now` would overflow, but `Instant`s created from dates could be a problem.

src/core.c/Instant.pm6
```diff
@@ -3,7 +3,7 @@ my class DateTime { ... }
 my class Duration {... }
 
 my class Instant is Cool does Real {
-    has Real $.tai is default(0);
+    has Int $.tai is default(0);
```
niner, line 6: This changes the public API with `$my-instant.tai` now returning an Int instead of Real. I thought the plan was to add a `method tai(--> Real)`?

niner (changes requested):

src/core.c/Instant.pm6
MasterDuke17: `method tai(--> Rat:D)` is already there a couple lines down (hidden in the code fold assuming you're looking on GitHub).

niner: Oh, indeed! Then all is well

niner (approved):

PR #4652, opened 2021-11-24
Make the max number of threads depend on number of cores

lizmat:
Previously, this was set to 64.  Clearly, should be dependent
on actual number of cores available.  For now, the maximum is set
to 8 times the number of cores, based on the previous value an the
more common 4-8 core hardware that was available when that maximum
was originally set.

### DIFF
--- a/src/core.c/ThreadPoolScheduler.pm6
+++ b/src/core.c/ThreadPoolScheduler.pm6
@@ -782,9 +782,9 @@ my class ThreadPoolScheduler does Scheduler {
 
     submethod BUILD(
       Int() :$!initial_threads = 0,
-      Int() :$!max_threads = %*ENV<RAKUDO_MAX_THREADS> // Kernel.cpu-cores * 8;
-        --> Nil
-    ) {
+      Int() :$!max_threads =
+        %*ENV<RAKUDO_MAX_THREADS> // (Kernel.cpu-cores * 8 max 64)
+    --> Nil) {
         die "Initial thread pool threads ($!initial_threads) must be less than or equal to maximum threads ($!max_threads)"
             if $!initial_threads > $!max_threads;
 
--- a/src/core.c/ThreadPoolScheduler.pm6
+++ b/src/core.c/ThreadPoolScheduler.pm6
@@ -781,8 +781,8 @@ my class ThreadPoolScheduler does Scheduler {
     }
 
     submethod BUILD(
-        Int :$!initial_threads = 0,
-        Int :$!max_threads = (%*ENV<RAKUDO_MAX_THREADS> // 64).Int
+      Int() :$!initial_threads = 0,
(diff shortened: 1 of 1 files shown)


jonathanstowe:
I actually thought I'd seen a commit that did this before, obviously a false memory.

lizmat:
Maybe that was for the default for `:degree` in `.race` and `.hyper` ?

jnthn:
This is probably OK, I wondered if memory available should factor into this as well, although perhaps a machine with a lot of cores could also be expected to have a lot of RAM. On a machine that reports 32 cores, the maximum here would be 32 * 8 = 256 threads. Each carries an initial cost of perhaps 1.25MB (around a meg of C stack space and two 128 KB nursery semispaces), those doubling in size if the thread is the one to blame for triggering GC up to 2 4MB semispaces. Thus the worst case is 2 GB of memory usage. Often we'd only end up with so many threads in the case of a ton of blocking operations are done over many threads, though.

src/core.c/ThreadPoolScheduler.pm6
```diff
@@ -781,8 +781,8 @@ my class ThreadPoolScheduler does Scheduler {
     }
 
     submethod BUILD(
-        Int :$!initial_threads = 0,
-        Int :$!max_threads = (%*ENV<RAKUDO_MAX_THREADS> // 64).Int
+      Int() :$!initial_threads = 0,
+      Int() :$!max_threads = %*ENV<RAKUDO_MAX_THREADS> // Kernel.cpu-cores * 8;
```
jnthn, line 785: I wonder if this should maybe be `Kernel.cpu-cores * 8 max 64`; there may be programs out there relying on the current upper bound, that would end up with a lower one here, and could start to deadlock as a result.

vrurg: @jnthn I was deadlocking on insufficient number of threads, until realized how things work. But too many?

Anyway, this parameter is an implementation detail. No one must rely on this number.

jnthn: >  But too many?

`max` would mean "no less than 64", e.g. retaining the current number, so no existing program will run up against a new insufficiency.

> Anyway, this parameter is an implementation detail. No one must rely on this number.

I agree, but what people should do and what people do do is a different matter. It costs us almost nothing to know we won't make something newly deadlock, and is cheaper than dealing with it if we do.

vrurg: @jnthn Sorry, it's my eternal problem with mixing `max`/`min` outcomes, akin to some people mixing left/right.

So, my comment is only relevant to when increasing the number would cause problems.

jnthn: Aha. :-) Yes, increasing the number is much less likely to cause trouble (and if it does, it's higher resource usage trouble, not deadlocked program trouble).

jnthn (approved):

Issue #2676, opened 2019-02-07
Do we need an overhaul of .ACCEPTS methods wrt Any / Mu?


lisprog:
## The Problem

say so (3,0).one ~~ (1 .. 9)

## Expected Behavior

expected to return True

## Actual Behavior

returns False

## Steps to Reproduce

> say (3,5).any ~~ (1 .. 9)
any(True, True)
> say so (3,5).any ~~ (1 .. 9)
True
> say so (3,5).one ~~ (1 .. 9)
False
> say so (3,0).one ~~ (1 .. 9)  # expect True because 0 not in range and 3 is
False
> say so (3,0).any ~~ (1 .. 9)
True
> say so (0, 3).one ~~ (1..9)   # expected True; 0 not in range; exactly one item (3) is in range
False
> so 0 ~~ 1..9                  # as expected;
False
> so 3 ~~ 1..9
True
> say (0, 3).one ~~ (1..9)      # unexpected; 0 not in range;
one(True, True)                 # expected one(False, True)
> say (1..9).elems
9
> say (0, 10).one ~~ (1..9)     # why is it true that 0 ~~ 1..9 ??
one(True, False)
> say so (0, 10).one ~~ (1..9)  # unexpected !!! neither 0 nor 10 in range
True
> say (-1, 3).one ~~ (1..9)     # why -1 in range of 1..9 ??
one(True, True)
> 

## Environment
* Operating system: 

Linux, Xubuntu 18.04

* Compiler version (`perl6 -v`): 

This is Rakudo Star version 2018.10 built on MoarVM version 2018.10
implementing Perl 6.c.

lizmat:
First research indicates that `Range.ACCEPTS` has a `Mu` candidate, which means it will absorb the `Junction`.  This seems like a leftover from way, way in the past as the last change on that line was in 2012, and that did not change the types in the signature.  It makes much more sense to **not** have a `Mu` there, so I removed that.  This changes the result of:

    $ perl6 -e 'dd (1..9).ACCEPTS(0|3)'
    any(Bool::True, Bool::True)

To:

    any(Bool::False, Bool::False)

Even stranger, it then appears that **none** of the other `Range.ACCEPTS` candidates is being called then.  Continuing the search...

lizmat:
[...]

PR #4653, opened 2021-11-27
Many improvements to smartmatching and given/when

vrurg:
- Junctions are now welcome on RHS and as `when` topic
- `when` now uses optimized code unless the topic is a Junction. Falls back to `ACCEPTS`+`Bool` in the latter case.
- smartmatch optimization is now more elaborate about its arguments. It implicitly `Bool`-ifies `ACCEPTS` return value unless its LHS is a regex or negation is to be done.
- more optimizations for literal and type matching

Overall, with this PR it must not be a problem if `ACCEPTS` gets autothreaded and results in a junctionized return. Either it will be collapsed with boolification where smartmatch semantics is expected, or junctionized outcome of a regex match is expected.

### DIFF
--- a/src/Perl6/Actions.nqp
+++ b/src/Perl6/Actions.nqp
@@ -2329,126 +2329,22 @@ class Perl6::Actions is HLL::Actions does STDActions {
         my $pblock := $xblock.shift;
         check_smartmatch($<xblock>,$sm_exp);
 
-        # Handle the smartmatch, but try to gen some more optimized code first.
-        my $match_past;
-
-        my $when_kind;
-        my $Junction := $*W.find_single_symbol('Junction', :setting-only);
-        my $False := $*W.find_single_symbol('False', :setting-only);
-
-        # Fast path for types, e.g., `when Int { ... }`
-        my $sm_type := nqp::null();
-        if nqp::istype($sm_exp, QAST::WVal) && !nqp::isconcrete($sm_type := $sm_exp.value) {
-            $when_kind := 'typematch';
-            $match_past:=
-                QAST::Op.new(
-                    :op('callmethod'),
-                    :name('ACCEPTS'),
-                    $sm_exp,
-                    WANTED(QAST::Var.new( :name('$_'), :scope('lexical') ), 'when')
-                );
-        }
-        # Fast path for things like literals, e.g., `when 5 { ... }`
-        elsif nqp::istype($sm_exp, QAST::Want) {
-            my $op_type;
-            my $method;
-            $when_kind := 'literal';
-            if nqp::istype($sm_exp[2], QAST::IVal) {
-                $op_type := '==';
-                $method  := 'Numeric';
-            }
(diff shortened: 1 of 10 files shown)


vrurg:
PR resolves raku/problem-solving#297. Fixes #2676 and, perhaps, a few more issues (yet to be investigated).

vrurg:
A couple of implementation notes.

To get junctions work as topics with `when` I had to add additional `istype` check into the optimized code @MasterDuke17 produced to fall back to `ACCEPTS` when appropriate. Then I compared this branch with 2021.10 release. And it seems that the branch is even somewhat faster – likely because of all the MoarVM optimizations done since the release.

As I have noticed in the commit comment, `Range` will not work with custom user-provided `cmp` operator. This was always the case and I don't feel ok about trying to fix it as it would require looking for client's context and locating the operator proto in there. Yet, considering that we also always had `Range.ACCEPTS` with `Mu`-typed topic, I decided there is no loss in keeping it this way but adding the code for operator lookup as otherwise it makes no sense. If this approach is considered redundant, not only the corresponding candidate could be removed, but there is a chance for slight optimization of the multi-dispatch by removing `Junction` candidate and providing `Range` with own `proto method ACCEPTS`.

vrurg:
Turning this into draft because there is some more work in progress on optimizing `when` and fixing cases where class defines custom `ACCEPTS`.

vrurg:
I think I'm done with this. The PR passes spectests, including the additions from Raku/roast#774. Aside of the fix for junctions on LHS and as `when` topic, I couldn't resist optimizing some most common cases. The optimization tries to cover cases, where compile time values are known, sometimes replacing a smartmatch with a plain boolean constant. For example, this would happen for:

```raku
sub foo(--> Num) { ... }
if foo() ~~ Numeric { ... }
```

Note that it means no call to `foo()` would ever take place.

Overall, the basic principles of optimization I tried to follow were:

- where possible, get rid of routine calls by replacing their code with nqp ops
- reduce allocations by unwrapping instantiation ASTs and pulling out any constant data available
- calculate op outcomes at compile time where possible

Because the above involves a lot of analyzing of value ASTs, I added a new class `Operand` which does all this work.

Unfortunately, I barely can provide many benchmark results as dispatching is seemingly taking so good care of the cases I tried, that sometimes a test loop happens to be even faster, than a base-line loop with no smartmatch in it. But at least in one case of smartmatching against pairs I did observe a speedup of 49%:

```raku
constant COUNT=1_000_000;
my $base = 0;
my $s = now;
for ^COUNT {
    my $v = $foo.bar1 === True;
}
$base = now - $s;

note "Base: ", $base;

my $total = 0;
$s = now;
for ^COUNT {
    my $v = $foo ~~ :bar1;
}
$total = now - $s;

note "Total: ", $total, ", actual: ", $total - $base;
```

vrurg:
Update. With JVM fixed I've benchmarked it too. 88% for pair, 92% for `$v ~~ Num`.

Issue #4657, opened 2021-11-30
`shell` / `run` isn't working on macOS Monterey


ugexe:
## The Problem

`raku -e 'shell("ls")'` (with and without variations of `:out`,`:err`,`:!out`,`:!err`) appears stuck 90% of the times I run it (no output, never finishes running) when using macOS Monterey. `run` behaves a bit better in that `run("ls")` always seems to work but `my $p = run("ls", :out); $p.out.close` does occasionally freeze up as well. Because precompilation and the build process use `shell`/`run` it also means installing rakudo can become stuck during the later parts of the installation process.

[This callback](https://github.com/rakudo/rakudo/blob/10c3dbb940290c67af33dfc617b2f8ce6a21df28/src/core.c/Proc/Async.pm6#L357) is not getting fired in the problematic cases.

## Expected Behavior

`raku -e 'shell("ls")'` should consistently show the output of the `ls` command

## Actual Behavior

10% of the time I see the output of the `ls` command. 90% of the time I see nothing and the program never finishes running.

## Environment

* Operating system: MacOS Monterey

* Compiler version (`perl6 -v` or `raku -v`): various

ugexe:
Also see https://github.com/rakudo/rakudo/issues/4637

jnthn:
Based on the stack trace posted on #moarvm earlier, I'm curious:

1. Does `raku -e 'use nqp; say nqp::cpucores'` hang?
2. If not, does the patch below help?

```
diff --git a/src/core.c/ThreadPoolScheduler.pm6 b/src/core.c/ThreadPoolScheduler.pm6
index 929652e..4a1aa1f 100644
--- a/src/core.c/ThreadPoolScheduler.pm6
+++ b/src/core.c/ThreadPoolScheduler.pm6
@@ -534,6 +534,7 @@ my class ThreadPoolScheduler does Scheduler {
     my constant NUM_SAMPLES           = 5;
     my constant NUM_SAMPLES_NUM       = 5e0;
     my constant EXHAUSTED_RETRY_AFTER = 100;
+    my int $cpu-cores = max(nqp::cpucores() - 1,1);
     method !maybe-start-supervisor(--> Nil) {
         unless $!supervisor.DEFINITE {
             $!supervisor = Thread.start(:app_lifetime, :name<Supervisor>, {
@@ -579,7 +580,6 @@ my class ThreadPoolScheduler does Scheduler {
[...]

PR #4659, opened 2021-12-01
Make sure that nqp::cpucores is only called once ever

lizmat:
It appears that nqp::cpucores can potentially be a heavy operation
(even like needing to fork() apparently on MacOS Monterey), which
may well have some adverse effects.

Therefore make sure we only call nqp::cpucores only once at startup.
Also introduce an implementation-detail Kernel.cpu-cores-but-one
method that will be one less than the number of CPU cores, with a
minimum of 1.

### DIFF
--- a/src/core.c/Iterable.pm6
+++ b/src/core.c/Iterable.pm6
@@ -10,6 +10,7 @@
 my class HyperSeq { ... }
 my class RaceSeq { ... }
 my class Rakudo::Internals::HyperIteratorBatcher { ... }
+my class Kernel { ... }
 my role Iterable {
     method iterator() { ... }
 
@@ -29,8 +30,8 @@ my role Iterable {
     }
 
     method hyper(
-      Int(Cool) :$batch = 64,
-      Int(Cool) :$degree = max(nqp::cpucores() - 1,1)
+      Int(Cool) :$batch  = 64,
+      Int(Cool) :$degree = Kernel.cpu-cores-but-one;
     ) {
 #?if !js
         HyperSeq.new:
@@ -45,8 +46,8 @@ my role Iterable {
     }
 
     method race(
-      Int(Cool) :$batch = 64,
-      Int(Cool) :$degree = max(nqp::cpucores() - 1,1)
+      Int(Cool) :$batch  = 64,
+      Int(Cool) :$degree = Kernel.cpu-cores-but-one;
     ) {
 #?if !js
         RaceSeq.new:
--- a/src/core.c/Iterable.pm6
+++ b/src/core.c/Iterable.pm6
@@ -31,7 +31,7 @@ my role Iterable {
 
     method hyper(
       Int(Cool) :$batch  = 64,
-      Int(Cool) :$degree = Kernel.cpu-cores-but-one;
+      Int(Cool) :$degree = Kernel.cpu-cores-but-one,
     ) {
 #?if !js
         HyperSeq.new:
@@ -47,7 +47,7 @@ my role Iterable {
 
     method race(
       Int(Cool) :$batch  = 64,
-      Int(Cool) :$degree = Kernel.cpu-cores-but-one;
+      Int(Cool) :$degree = Kernel.cpu-cores-but-one,
     ) {
 #?if !js
         RaceSeq.new:
--- a/src/core.c/Kernel.pm6
+++ b/src/core.c/Kernel.pm6
@@ -164,13 +164,20 @@ class Kernel does Systemic {
     multi method signal(Kernel:D: Signal:D \signal --> Int:D) { signal.value }
     multi method signal(Kernel:D: Int:D    \signal --> Int:D) { signal       }
 
-    method cpu-cores(--> Int) { nqp::cpucores }
+    my int $cpu-cores;
+    proto method cpu-cores(|) {*}
+    multi method cpu-cores(--> Int) { $cpu-cores = nqp::cpucores }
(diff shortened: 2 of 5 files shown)


jnthn:
> (even like needing to fork() apparently on MacOS Monterey)

It's not entirely clear it does fork, but it does load a dynamic library, and that ends up calling `getpwuid_r` (seemingly in library initialization), which goes down a path that ends up in `pthread_atfork`, for reasons that aren't entirely clear to me.

The issue seen in #4657 seems to happen because there's a lock used inside of the pthreads implementation and the thread trying to load the dynamic library and the thread trying to spawn a process end up deadlocked trying to take it. At which point I wonder if neither MoarVM nor libuv are in the wrong, and this is actually a MacOS regression.

lizmat:
Well, regression or not, as @ugexe mentioned: Node on Monterey (also using libuv) does not appear to suffer from this issue.  Perhaps they don't even call that libuv function?

jnthn:
> It's not entirely clear it does fork, but it does load a dynamic library

Went exploring the code in libuv, and it does indeed `dlopen` (and close) on every time we ask for the number of CPU cores, so it's certainly not cheap; that this PR works around a potential OS regression is nice, but it's worth it from a performance perspective too, since it will reduce the resource usage of the supervisor thread potentially quite significantly.

jnthn:
> Perhaps they don't even call that libuv function?

It'd be enough that they don't call it at the same time as trying to spawn a process. (We only do because we are setting up a thread pool to handle events from the spawned process; that'd not be something that happens in Node.js.)

lizmat:
Hmmm... looks like this patch has an adverse effect on spectest (305 -> 314 seconds last night).  Which would also mean an adverse effect on precomp.

Will look at making this lazy for the cases that are needed in core, while the Kernel.cpu-cores functionality would revert to the original state, supporting changes in number of (visible) CPUs

src/core.c/Kernel.pm6
```diff
@@ -164,11 +164,15 @@ class Kernel does Systemic {
...
     multi method signal(Kernel:D: Int:D    \signal --> Int:D) { signal       }
 
-    my int $cpu-cores = nqp::cpucores;
-    my int $cpu-cores-but-one = max $cpu-cores - 1, 1;
+    method cpu-cores(--> Int) { nqp::cpucores }
 
-    method cpu-cores(--> Int) { $cpu-cores }
-    method cpu-cores-but-one() is implementation-detail { $cpu-cores-but-one }
+    my $cpu-cores-but-one := nqp::null;
+    method cpu-cores-but-one() is implementation-detail {
```
atroxaper, line 170: I feel a little inconsistency between .cpu-cores and .cpu-cores-but-one (the first returns dynamic value, the second - static value). implementation-detail trait forgives it of course. What do you think about:
```
method cpu-cores(--> Int) {
  my $cpu-cores := nqp::cpucores;
  $cpu-cores-but-one := max $cpu-cores - 1, 1;
  $cpu-cores
}
```

atroxaper: Or leave only one method
```
my int $cpu-cores := nqp::null;
method cpu-cores(:$last-value = False --> Int) {
  nqp::ifnull(
    $cpu-cores,
    $cpu-cores := nqp::cpucores()
  )
  nqp::if(
    $last-value,
    $cpu-cores,
    $cpu-cores := nqp::cpucores()
  )
}
[...]
max Kernel.cpu-cores(:last-value)-1, 1
```

lizmat: The current API of `Kernel.cpu-cores` is that it will always look up the value.  It was felt that in some situations, making it return a static value, would be incorrect.

On the other hand, it turned out that `nqp::cpucores` is way more expensive than was thought **and** it is causing issues when threading is involved.  So, on the one hand we want to reduce the CPU used, and on the other hand we only want to call `nqp::cpucores` when it is actually necessary.  This is now what the internal `Kernel.cpu-cores-but-one` does, as that is the default value we need to have in all cases internally.

atroxaper: Oh, I missed this moment that `Kernel.cpu-cores` is public API, sorry. I'm fully understand why you introduced `Kernel.cpu-cores-but-one`. Probably it is only naming related inconsistency in my head. Nevermind. Thank you for your time.

src/core.c/Iterable.pm6
```diff
@@ -29,8 +30,8 @@ my role Iterable {
     }
 
     method hyper(
-      Int(Cool) :$batch = 64,
-      Int(Cool) :$degree = max(nqp::cpucores() - 1,1)
+      Int(Cool) :$batch  = 64,
+      Int(Cool) :$degree = Kernel.cpu-cores-but-one;
```
ugexe, line 34: Should this end with a comma instead?

src/core.c/Iterable.pm6
```diff
@@ -45,8 +46,8 @@ my role Iterable {
     }
 
     method race(
-      Int(Cool) :$batch = 64,
-      Int(Cool) :$degree = max(nqp::cpucores() - 1,1)
+      Int(Cool) :$batch  = 64,
+      Int(Cool) :$degree = Kernel.cpu-cores-but-one;
```
ugexe, line 50: Should this end with a comma instead?

lizmat:
It should.  Or nothing.  Not that it matters much, it would just not include any additional parameters in dispatch, and there aren't any.

ugexe (approved):

PR #4671, opened 2021-12-15
Fix the behaviour of Pair ~~ Pair

lizmat:
Apparently, this was borked since d8f7e8b8d6ea25b about 6 years ago.

See discussion at https://logs.liz.nl/moarvm/2021-12-15.html#16:32-0003

This breaks test #177 of t/spec/S02-types/pair.t

### DIFF
--- a/src/core.c/Pair.pm6
+++ b/src/core.c/Pair.pm6
@@ -54,7 +54,8 @@ my class Pair does Associative {
         $!value.ACCEPTS(%h.AT-KEY($!key));
     }
     multi method ACCEPTS(Pair:D: Pair:D $p) {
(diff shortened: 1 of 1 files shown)


lizmat:
To be merged **AFTER** the 2021.12 release.

vrurg (approved):

vrurg:
This also breaks t/spec/S02-types/pair.rakudo.moar, but it's about moving `:3z` from the true-list into the false-list.

And I hope nobody relies on erroneous behavior. But then it's better be fixed on their side.

lizmat:
I guess if this caused Blin fallout, we need to fix this in a language version.

PR #4672, opened 2021-12-16
Warn on potential problem with ternaries and infixes

MasterDuke17:
An infix potentially being used incorrectly (because of precedence problems)
in the condition of a ternary will now cause a warning.

### DIFF
--- a/src/Perl6/Optimizer.nqp
+++ b/src/Perl6/Optimizer.nqp
@@ -1509,6 +1509,23 @@ class Perl6::Optimizer {
             }
         }
 
+        # Warn for something like `my $a = 2.rand.Int; say "a = " ~ $a ?? "true" !! "false";`, which
+        # will just print 'true' because the entire expression `"a = " ~ $a` is being evaluated by the
+        # ternary instead of just the `$a` (the correct form is `"a = " ~ ($a ?? "true" !! "false"`)). 
+        if $optype eq 'if' && $op.name eq '&infix:<??>' &&
+              nqp::istype($op[0], QAST::Op) && ($op[0].op eq 'call' || $op[0].op eq 'callstatic') &&
+              nqp::index($op[0].name, '&infix:') == 0 && $!symbols.is_from_core($op[0].name) &&
+              (nqp::istype($op[0][0], QAST::Want) || nqp::istype($op[0][1], QAST::Want)) {
+            # See if we can staticalize the infix to determine whether it's 'iffy' (produces a boolean value),
+            # which is valid in this construct, so we wouldn't want to warn.
+            try {
+                my $obj := $!symbols.find_lexical($op[0].name);
+                if nqp::can($obj, 'prec') && !$obj.prec<iffy> {
+                    $!problems.add_worry($op, "Potential dead code, the '?? !!' is gobbling up the result of the '" ~ $op[0].name ~ "'");
+                }
+            }
+        }
+
         # Visit the children.
         self.visit_op_children($op);
 


vrurg:
I wouldn't do it. What if this _is_ my intention? And what if I don't want explicit braces?

BTW, correct me if I'm wrong, but it feels to me that even `("a = " ~ $a) ?? "true" !! "false"` won't make the warning go away.

MasterDuke17:
> I wouldn't do it. What if this _is_ my intention? And what if I don't want explicit braces?

Well, it's just a warning, if you intentionally want to perform a no-op it's not stopping you.

> BTW, correct me if I'm wrong, but it feels to me that even `("a = " ~ $a) ?? "true" !! "false"` won't make the warning go away.

Turns out it does, though `quietly` doesn't, which I don't understand.
```raku
# what I intended to write, correct output, no warning
my $a = 3;
say "a: " ~ ($a ?? 1 !! 2);

# OUTPUT:
a: 1


# what I actually wrote, wrong output, now with warning
my $a = 3;
say "a: " ~ $a ?? 1 !! 2;

# OUTPUT:
WARNINGS for -e:
Potential dead code, the '?? !!' is gobbling up the result of the '&infix:<~>' (line 1)
1


# wrong output, no warning
my $a = 3;
say ("a: " ~ $a) ?? 1 !! 2;

# OUTPUT:
1


# wrong output, with warning
my $a = 3;
quietly say "a: " ~ $a ?? 1 !! 2;

# OUTPUT:
WARNINGS for -e:
Potential dead code, the '?? !!' is gobbling up the result of the '&infix:<~>' (line 1)
1


# wrong output, with warning
my $a = 3;
quietly { say "a: " ~ $a ?? 1 !! 2 };

# OUTPUT:
WARNINGS for -e:
Potential dead code, the '?? !!' is gobbling up the result of the '&infix:<~>' (line 1)
1


# wrong output, with warning
my $a = 3;
say quietly "a: " ~ $a ?? 1 !! 2;

# OUTPUT:
WARNINGS for -e:
Potential dead code, the '?? !!' is gobbling up the result of the '&infix:<~>' (line 1)
1


# wrong output, with warning
my $a = 3;
say quietly { "a: " ~ $a ?? 1 !! 2 };

# OUTPUT:
WARNINGS for -e:
Potential dead code, the '?? !!' is gobbling up the result of the '&infix:<~>' (line 1)
1
 ```

**Comment by MasterDuke17** [comment]
> First, will it warn for `say $a ~ $b ?? "yes" !! "no"`?

No, it doesn't warn for this case.

> Another case is `say "" ~ $obj ?? "yes" !! "no"` where `$obj` is of a class with its own stringification which might has unpredictable semantics and could result in an empty string. As far as I understand, `say` is not the only context where the warning can be triggered. So, replace it with `ok` or `nok` – and here comes a test case scenario, where even absurd code sometimes makes sense.

I agree, but that is a bit of a contrived case (what's the point of concatenating with the initial empty string?). And we already have a lot of tests that trigger warnings, which silently get ignored by prove and you only see if you run the files directly.

> From your tests, I'd say that `("a: " ~ $a) ?? 1 !! 2` silencing the warning is a way to go, eventhough it seems rather unexpected outcome even for yourself. :) If you could make it also work for `("a: " ~ $a ?? 1 !! 2)` that would also do.

I would expect `("a: " ~ $a ?? 1 !! 2)` to warn, it still has the precedence problem.

> Otherwise it makes full sense for `quietly` not silencing compile-time warnings. For this a pragme is needed and I don't remember if we have one.

`no worries` https://docs.raku.org/language/pragmas#worries seems like it should do it, but it doesn't.
```raku
no worries;
my $a = 3;
say "a: " ~ $a ?? 1 !! 2;

# OUTPUT:
WARNINGS for -e:
Potential dead code, the '?? !!' is gobbling up the result of the '&infix:<~>' (line 1)
1
```

vrurg:
> I agree, but that is a bit of a contrived case (what's the point of concatenating with the initial empty string?).

As I said, tests may have different kinds of weirdness because bugs are weird sometimes.

> And we already have a lot of tests that trigger warnings

I'm trying to consider if user tests or code are to be harmed somehow. But you're right, unless some really low-level magic is done by the code, `"" ~` would be no different to plain `~` with regard to coercion.

One last thing I've just thought about: custom `&infix:<~>` or whatever other op which can actually produce a falsy value.

MasterDuke17:
> One last thing I've just thought about: custom `&infix:<~>` or whatever other op which can actually produce a falsy value.

I've just pushed a change that only enables the warning for symbols from core (and includes a small simplification). Now,
```raku
my $a = 3;
say "a: " ~ $a ?? 1 !! 2;

# OUTPUT:
WARNINGS for -e:
Potential dead code, the '?? !!' is gobbling up the result of the '&infix:<~>' (line 1)
1
```
but
```raku
multi infix:<~>(Str $s, Int $i) {
    say "custom";
    $i < $s.Int
};
my $a = 3;
say "6" ~ $a ?? 1 !! 2;

# OUTPUT:
custom
1
```
(though also
```raku
multi infix:<~>(Complex $a, Complex $b) {
    $a < $b.Int
};
my $a = 3;
say "a: " ~ $a ?? 1 !! 2;

# OUTPUT:
1
```
even a non-matching multi seems to pollute the core lookup).

vrurg (approved):
> even a non-matching multi seems to pollute the core lookup).

I'm currently trying to finalize a big PR on smartmatch, which includes some optimizations and also depends on user-defined candidates of a multi. But I ended up with the same approach: give up on any non-core candidate.

PR #4679, opened 2021-12-20
Fix segfaults due to concurrent hash access in parallel module loading

niner:
Read access to the %!seen and %!loaded hashes in CompUnit::Repository::
Installation and FileSystem was not protected by a lock. This could lead to us
fetching an only half-setup entry from the hash with a key that was still a
NULL pointer which we then dereferenced.

Fix by extending (or introducing) locks to all access to those hashes.

### DIFF
--- a/src/core.c/CompUnit/Repository/FileSystem.pm6
+++ b/src/core.c/CompUnit/Repository/FileSystem.pm6
@@ -2,6 +2,7 @@ class CompUnit::Repository::FileSystem
   does CompUnit::Repository::Locally
   does CompUnit::Repository
 {
+    has $!lock = Lock.new;
     has %!loaded; # cache compunit lookup for self.need(...)
     has %!seen;   # cache distribution lookup for self!matching-dist(...)
     has $!precomp;
@@ -14,10 +15,14 @@ class CompUnit::Repository::FileSystem
     my constant @extensions = <.rakumod .pm6 .pm>;
 
     method !matching-dist(CompUnit::DependencySpecification $spec) {
-        return $_ with %!seen{~$spec};
+        $!lock.protect: {
+            return $_ with %!seen{~$spec};
+        }
 
         with self.candidates($spec).head {
-            return %!seen{~$spec} //= $_;
+            $!lock.protect: {
+                return %!seen{~$spec} //= $_;
+            }
         }
 
         Nil
@@ -71,7 +76,9 @@ class CompUnit::Repository::FileSystem
 
         --> CompUnit:D)
     {
-        return $_ with %!loaded{~$spec};
+        $!lock.protect: {
+            return $_ with %!loaded{~$spec};
+        }
 
         with self!matching-dist($spec) {
             my $name = $spec.short-name;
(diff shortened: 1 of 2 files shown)


src/core.c/CompUnit/Repository/FileSystem.pm6
```diff
@@ -88,14 +95,16 @@ class CompUnit::Repository::FileSystem
...
-            return %!loaded{~$spec} = CompUnit.new(
-                :short-name($name),
-                :handle($precomp-handle // CompUnit::Loader.load-source($source-handle.open(:bin).slurp(:close))),
-                :repo(self),
-                :repo-id($id.Str),
-                :precompiled($precomp-handle.defined),
-                :distribution($_),
-            );
+            $!lock.protect: {
+                return %!loaded{~$spec} = CompUnit.new(
```
ugexe, line 99: `=` probably should have always been `//=`, but I'm not sure it really matters

ugexe (approved):

MasterDuke17:
I thought read access to a hash was always safe (or at least since MoarVM got its new hash implementation a year ago), as long as the/any writes were protected?

lizmat:
As far as my understanding goes, the chance of a problem has become less, but is still not zero.  @nwc10 comments?

patrickbkr:
@MasterDuke17 AFAIK write <-> write can clash. write <-> read can clash. read <-> read is safe. But since write <-> read is unsafe one needs to guard the reads as well if a concurrent write can't be ruled out. But I'll refer to @nwc10 for the definite answer.

jnthn:
What @patrickbkr said is correct. What has changed in the new hash implementation is that you're more likely to get a useful report giving the location of one of the conflicting writes rather than a SEGV without such information.

nwc10:
I can't add anything beyond what @jnthn confirmed.

MasterDuke17:
There have been a bunch of fixes recently for hashes and arrays in Rakudo being modified concurrently. Would it make sense to bring something like OO::Monitors into core? Or some form of concurrency-safe hashes/arrays?

nwc10:
> Or some form of concurrency-safe hashes/arrays?

I don't know where to even start to write a correct concurrency-safe hash, let alone a correct *and* performant hash, and the thought of trying to do this does not fill me with joy.

(And my vague memory of what I'd read about implementations of concurrency-safe hashes was that they implemented fetch and store, but no-one ever implemented iteration, or even thought about how to do it.)

niner:
FWIW in this case we're totally fine with normal hashes and heavy weight locks because their cost is drowned by the actual module loading - several times over.

The only reason I haven't merged this branch despite the approval is https://github.com/rakudo/rakudo/issues/1920 indicates that there are more issues to be solved. Just haven't found the time to dig deeper yet as I'm still investigating a NativeCall concurrency issue for which a small chance exists that it's a regression. But maybe I just shouldn't let the perfect be the enemy of the good...

MasterDuke17:
> I don't know where to even start to write a correct concurrency-safe hash, let alone a correct _and_ performant hash, and the thought of trying to do this does not fill me with joy.

Yeah, in this case I was thinking we'd start with something implemented in Rakudo and a VM-level implementation would be a possible optimization for the future.

MasterDuke17:
> FWIW in this case we're totally fine with normal hashes and heavy weight locks because their cost is drowned by the actual module loading - several times over.

Like I mentioned in my immediately previous reply, I was mostly thinking of some sugar around normal hashes and locks. So really looking to make correctness easy at first, not necessarily a performance improvement.

So for all the 
```raku
class Something {
    has %!some-cache;

    method update($foo, $bar) {
        %!some-cache{$foo} = $bar;
    }

    method get($foo) {
        %!some-cache{$foo} //= expensive-operation();
    }
}
```
we can instead do something like `has magically-concurrence-safe %!some-cache;` and not need to manually wrap those hash operations.

PR #4686, opened 2021-12-26
Relax cross-language revision role-class boundaries

vrurg:
Prior to this PR 6.c/d classes were not being able to consume roles
originating from 6.e code due to differences in handling submethods.
This commit changes the situation by teaching them this art:

- pre-6.e classes respect 6.e submethod handling and do not transfer
  submethods from 6.e roles
- constructor methods `BUILD` and `TWEAK` on 6.e roles are now included
  into the build plan of pre-6.e classes

From now on the following will work as expected:

```raku
use v6.e.PREVIEW;

my $v = 42 but role {
    submethod TWEAK { say "Tweaking an Int" }
};
```


vrurg:
Kind of a Christmas present to some people. :)

PR #4691, opened 2021-12-30
Add support for $RAKUDO_OPT environment variable

vrurg:
A way to specify default compiler CLI options, overridable by explicitly specified in the command line. For example:

    RAKUDO_OPT="--I. --ll-exception" rakudo --Ilib test.raku

`--I.` would be ignored but `--ll-exeption` will be used.

Simple space split is used making it impossible to use a path with a space in it to `--I`.

First `-e` or argument are ignored as well as any options after them, similarly to what Rakudo compiler does.

### DIFF
--- a/src/main.nqp
+++ b/src/main.nqp
@@ -54,7 +54,18 @@ sub MAIN(@ARGS) {
 sub MAIN(*@ARGS) {
 #?endif
     # Enter the compiler.
-    $comp.command_line(@ARGS, :encoding('utf8'), :transcode('ascii iso-8859-1'));
+    my %defaults;
+    if nqp::existskey(nqp::getenvhash, 'RAKUDO_OPT') {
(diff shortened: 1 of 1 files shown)


vrurg:
The reason I came across this idea is testing/spectesting with various levels of optimization. Which, as matter of fact, fails if the static optimizer is not used at all.

But the feature could prove its usefulness in many other scenarios like easy dumping of stage stats.

MasterDuke17:
> The reason I came across this idea is testing/spectesting with various levels of optimization. Which, as matter of fact, fails if the static optimizer is not used at all.

As I mentioned in #moarvm, I think it's been true for a while that the "optimizer" also does correctness checking. Iirc, part of rakuast is going to be breaking those out into (at least) two parts/stages, so that we can make the optimization stage truly optional and the checking stage non-optional.

PR #4692, opened 2021-12-30
Make Int.rand return an Int rather than a Num

lizmat:
This came up with

  https://logs.liz.nl/raku-beginner/2021-12-30.html#19:54

### DIFF
--- a/src/core.c/Int.pm6
+++ b/src/core.c/Int.pm6
@@ -12,6 +12,8 @@ my class Int does Real { # declared in BOOTSTRAP
     # class Int is Cool
     #     has bigint $!value is box_target;
 
+    method rand() { nqp::rand_I(self,Int) }
+
     multi method WHICH(Int:D: --> ValueObjAt:D) {
         nqp::box_s(
           nqp::concat(
--- a/src/core.c/Rakudo/QuantHash.pm6
+++ b/src/core.c/Rakudo/QuantHash.pm6
@@ -1057,7 +1057,7 @@ my class Rakudo::QuantHash {
     # initialized IterationSet with at least 1 element in Mix format,
     # and the total value of values in the Mix.
     method MIX-ROLL(\elems, \total) {
-        my      $rand := total.rand;
+        my      $rand := total.Num.rand;
         my Real $seen := 0;
         my $iter := nqp::iterator(elems);
 


lizmat:
Spectest clean, fwiw :-)

MasterDuke17 (approved):
You know, I've pretty much always wanted Int.rand to give me an Int, but just assumed it was specced to give a Num.

lizmat:
https://github.com/Raku/old-design-docs/blob/master/S32-setting-library/Numeric.pod#real states:

     multi method rand (Real $x: --> Num ) is export

Still, this PR did not cause any direct spectest breakage.  The test that broke was a `Mix` test, where the test was using `Int` weights.

MasterDuke17:
Might want to ask for a Blin run of this PR: https://gist.github.com/Whateverable/4b6219342e06addcd0af6883c5ea57a5

jnthn:
If I understand correctly, this may break programs obviously but also subtly:

* A `Num` type constraint not being met. At least one would then know the program was broken.
* Something that returned a number between 0 and 1 (so with randomness to the about half the entire range of floating point numbers) would cease to do so. I can imagine folks having relied on this, and suddenly their programs are a lot less random than they were (in the case of `1.rand`, not at all!) While hopefully folks wouldn't be using this for anything needing cryptographic strength, there's quite a lot of things that are blurry (for example, generating a one-time code for a password reset).

MasterDuke17:
Yeah, I approve in principle, but I suspect I'll have to not approve in practice. I wish this had been done a long time ago before it would have caused all the ecosystem fallout, but I guess now we should add a spectest for Int.rand returning a Num so we know we can't do this in the future.

jnthn:
I figured `(^100).pick`  - which a bit longer - was the "standard" way to pick an integer at random from a range. Maybe (if it doesn't already) the `rand` docs could hint at this.

MasterDuke17:
> I figured `(^100).pick` - which a bit longer - was the "standard" way to pick an integer at random from a range. Maybe (if it doesn't already) the `rand` docs could hint at this.

There's at least an error message that mentions it.
```
[dan@alexandria perl6]$ raku -e 'say rand(50)'
Unsupported use of rand(N).  In Raku please use: N.rand for Num or
(^N).pick for Int result.
at -e:1
------> say rand⏏(50)
```

lizmat:
Not worrying too much on `(^100).roll`, because then ideally the `Range` object is probably created at compile time already.  But for `(^$foo).roll`, the `Range` object needs to be created (which isn't *that* cheap), only to find the bounds.

A small benchmark shows that `(^$foo).roll` is about 12x as slow as `$foo.rand` in this PR.

MasterDuke17:
> A small benchmark shows that `(^$foo).roll` is about 12x as slow as `$foo.rand` in this PR.

Maybe the static optimizer (or spesh) can learn to do something about this?

lizmat:
Alternately, we could make `42.roll` not return `42`, but a random value between `0` and `41` ?

MasterDuke17:
> Alternately, we could make `42.roll` not return `42`, but a random value between `0` and `41` ?

And `42.pick`, since `(^N).pick` is what's recommended in the `rand(N)` error message (which should then be changed).

lizmat:
for `42.pick`, yes, but `42.pick(N), that would just pass on to `(^42).pick(N)`.   `42.roll(N)` could also be specially handled  :-)

PR #4695, opened 2022-01-02
Make Int.roll/pick act as (^Int).roll/pick on 6.e

lizmat:
As an alternate solution to https://github.com/rakudo/rakudo/pull/4692

Use a language level to make sure that this new behaviour is opt-in.

### DIFF
--- a/src/core.e/Int.pm6
+++ b/src/core.e/Int.pm6
@@ -1,0 +1,7 @@
+use MONKEY;
+augment class Int {
+    method roll(|c) { (^self).roll(|c) }
+    method pick(|c) { (^self).pick(|c) }
+}
+
+# vim: expandtab shiftwidth=4
--- a/tools/templates/6.e/core_sources
+++ b/tools/templates/6.e/core_sources
@@ -3,6 +3,7 @@ src/core.e/control.pm6
 src/core.e/PseudoStash.pm6
 src/core.e/Grammar.pm6
 src/core.e/EXPORTHOW.pm6
+src/core.e/Int.pm6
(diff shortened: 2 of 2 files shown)


MasterDuke17:
Not going directly to `nqp::rand_I` for performance?

lizmat:
Good point!  That would make sense for `.roll`, for `.pick` not so.

lizmat:
OK, some preliminary benchmarks doesn't show improvement.

So I'll leave it now until we can agree on the PR.

MasterDuke17:
> OK, some preliminary benchmarks doesn't show improvement.
Locally I see
```raku
use MONKEY;
augment class Int {
    method roll(|c) {
        (^self).roll(|c)
    }
};
Int.^compose;
my $a;
my $s = now;
$a = 1_000_000.roll for ^1_000_000;
say now - $s;
say $a
```
take ~1.1s, but
```raku
use nqp;
use MONKEY;
augment class Int {
    method roll(|c) {
        nqp::rand_I(self, Int)
    }
};
Int.^compose;
my $a;
my $s = now;
$a = 1_000_000.roll for ^1_000_000;
say now - $s;
say $a
```
take ~0.06s.

> That would make sense for `.roll`, for `.pick` not so.
I thought the behavior of `Range.roll` was the same as `Range.pick` (assuming no args to either)?

lizmat:
Yeah, `Range.pick()` is the same as `Range.roll`  :-)

I will look more into this when this has been merged.

MasterDuke17 (approved):

Issue #4615, opened 2021-11-02
junctions with given/when behave strangely


doomvox:
Passing an "all" junction to a given/when construct should let you
do a smartmatch against the junction.  This used to work with older 
rakudos, but with a recent build from HEAD it doesn't.

This should say "both int":

    raku -e 'say do given all(3,7) { when Int { "both int" }; default {"differ"} };' 

That used to work, but doesn't with recent builds.

    |                 version | result     | reported by                                  |
    |-------------------------+------------+----------------------------------------------|
    | v2021.10-43-ga8329f6fd. | "differ"   |                                              |
    | v2021.09-229-g38f6f8731 | "differ"   |                                              |
    |                 2021.06 | "differ"   | William Michels                              |
    |                 2021.04 | "differ"   | https://github.com/rakudo/rakudo/issues/4386 |
    |                 2020.11 | "both int" | Bruce Gray                                   |
    |                 2020.10 | "both int" |                                              |
    |             2020.[5..9] | "both int" | Bruce Gray                                  |
    |               2020.05.1 | "both int" | Andy Bach                                    |

I'm on:

    Linux fandango 4.9.0-8-686 #1 SMP Debian 4.9.144-3 (2019-02-02) i686 GNU/Linux

Some discussion of this:

    https://www.nntp.perl.org/group/perl.perl6.users/2021/11/msg10240.html

MasterDuke17:
This bisects to https://github.com/rakudo/rakudo/commit/c080e59404ed1b27e8841249e1c71d7e013c09ca. If we decide the prior behavior is correct, I'm tempted to just revert that change. It was significantly more difficult to implement than expected, makes the code more complex, wasn't all that much faster, and might be made obsolete by RakuAST anyway.

pmichaud:
FWIW, I find the old behavior to be more in line with what I would expect to happen here.  I would support reverting this change.

Pm

doomvox:
I found the current behavior tremendously puzzling.  Consider the fact
that this works:

     given all(3,7) { say so $_ ~~ Numeric; }
     # True

Since the "given" does topicalize the junction, and an explicit
smartmatch against the junction works, it's difficult to see why an
implicit smartmatch with "when" wouldn't work.

raiph:
cf https://github.com/Raku/problem-solving/issues/297

doomvox:
And it turns out fecundf opened up an issue like this one already that I'd forgotten about:

[...]

Issue #4386, opened 2021-05-31
Mismatch between "?(any(...) ~~ value)" and "given (any(...)) {when (value)...}"


fecundf:
## The Problem

A junction on the left side doesn't match a value in "when", even when it matches with a simple ?( ... ~~ ...) check.
Example-

> ? (any(4,3) ~~ 3)
True
> given any(4,3) { when 3 {say '3'}; say 'nope'}
nope

## Expected Behavior

Last line should say '3' matching the '?(any(4,3) ~~ 3)' return value

## Steps to Reproduce
Code snippet at the top of this request, plus environment below, should suffice.

There's a full discussion at https://www.nntp.perl.org/group/perl.perl6.users/2021/05/msg9961.html

Bruce Gray's comment "This looks very version dependent, and I would not have expected it to be version dependent at all" makes me think adding this case to Rakudo's test suite is a good idea.

## Environment
* Operating system: macOS 11.4
* Compiler version (`perl6 -v` or `raku -v`):
Welcome to Rakudo(tm) v2021.04.
Implementing the Raku(tm) programming language v6.d.
Built on MoarVM version 2021.04.

raiph:
cf Raku/problem-solving#297

vrurg:
Addressed by #4698

PR #4698, opened 2022-01-04
Additional fixes and improvements of smartmatch optimizations

### LINKED ISSUE #4693: Negated smartmatching with pair as matcher gives wrong result (due to optimizing)
## The Problem
Given a subdirectory "nqp" in the current working directory, the following two commands both print ```True```.
```
$ ./rakudo-m -e 'say so "nqp".IO ~~ :d'
True
$ ./rakudo-m -e 'say so "nqp".IO !~~ :d'
True
```

## Expected Behavior
The second command should print ```False```.

## Details
The output changes with decreasing the optimizer level to 1 ```--optimize=1```:
```
$ ./rakudo-m --optimize=1 -e 'say so "nqp".IO ~~ :d'
True
$ ./rakudo-m --optimize=1 -e 'say so "nqp".IO !~~ :d'
False
```

The problem also goes away with https://github.com/rakudo/rakudo/commit/7554d73322 reverted.

On the JVM backend the mentioned problem even leads to spectest failures in S16-filehandles/filetest.t
```
S16-filehandles/filetest.rakudo.jvm 14 - ~~:d returns false on non-existent directories
S16-filehandles/filetest.rakudo.jvm 15 - ~~:r returns false on non-existent directories
S16-filehandles/filetest.rakudo.jvm 16 - ~~:w returns false on non-existent directories
S16-filehandles/filetest.rakudo.jvm 17 - ~~:x returns false on non-existent directories
S16-filehandles/filetest.rakudo.jvm 18 - ~~:f returns false on non-existent directories
S16-filehandles/filetest.rakudo.jvm aborted 96 test(s)
```
(The tests for this file are passing on MoarVM.)

## Environment
* Operating system: FreeBSD 12.3-RELEASE
* Compiler version (`perl6 -v` or `raku -v`): Welcome to Rakudo™ v2021.12-27-gd80fb897e.
Built on MoarVM version 2021.12.

vrurg:
- Better detection of invocations, method calls wrapped into `hllize` are considered now.
- Better detection of possible return value type for an AST node.
- As a result of the above two items, smartmatches with method calls can now be optimized too.
- For some instantiations we can now infer the final type of returned object. In some cases it may differ from the invocator's type if method `new` decides it would make sense to return something different. For example, iterators often return an `Empty` iterator if there are no elements in an iterable object.
- On some occasions when smartmatch topic/LHS hapens to be a `Junction` at run-time, instead of falling back to `.ACCEPTS.Bool` on RHS to handle the junction correctly, it is possible to use a new `Junction` method `BOOLIFY-ACCEPTS` with topic/LHS as method's invocant. The method avoids extra iterations over eigenstates when possible; i.e. it shortcuts when conditions allow.
- Fixed typematching sometimes not respecting junctions.
- Tydied up `Regexp` safety check.

Fixes #4693

### DIFF
--- a/src/Perl6/Optimizer.nqp
+++ b/src/Perl6/Optimizer.nqp
@@ -1055,11 +1055,15 @@ my $OPERAND_VALUE_RETURN   := 2;
 
 my class Operand {
     has $!ast;
-    has $!is-want;
     has $!unwrapped;
+    # Invocations can be wrapped into hllize op. Here we will keep the actuall nqp::call* branch of AST
+    has $!invocation-ast;
+    has $!is-want;
     has $!value;
     has $!value-kind;
+    has $!is-instantiation;
     has $!routine;
+    has $!routine-name;
     has $!analyzed;
     has $!optimizer;
     has $!symbols;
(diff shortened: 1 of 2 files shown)


vrurg:
Fixes #4615, #4386

PR #4701, opened 2022-01-05
Remove some no longer needed CALL-MEs

MasterDuke17:
I'm not exactly sure as of when these aren't needed, but likely after
new-disp landed.

### DIFF
--- a/src/core.c/Dateish.pm6
+++ b/src/core.c/Dateish.pm6
@@ -9,8 +9,6 @@ my role Dateish {
         IO::Path.new(~self)
     }
 
-    method CALL-ME(Dateish:U: \dateish) { self.new(dateish) }
-
     # this sub is also used by DAYS-IN-MONTH, which is used by other types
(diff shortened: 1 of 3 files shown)


lizmat:
I assume this is spectest clean.  But where there any tests for that?

MasterDuke17:
> I assume this is spectest clean. But where there any tests for that?

Yep:
https://github.com/Raku/roast/blob/master/S32-temporal/Date.t#L305-L314
https://github.com/Raku/roast/blob/master/S07-slip/slip.t#L14-L16
https://github.com/Raku/roast/blob/master/S02-types/array.t#L340-L343

niner (approved):

jonathanstowe:
> I'm not exactly sure as of when these aren't needed, but likely after
> new-disp landed.

I think it was the coercion rework.

MasterDuke17:
> I think it was the coercion rework.

I'm not 100% sure it was just one or the other. Might have needed both.

PR #4702, opened 2022-01-05
Add CI job that runs a spectest instead of Rakudo tests

MasterDuke17:


vrurg:
_Instead_? It doesn't look right to me.

MasterDuke17:
> _Instead_? It doesn't look right to me.

It's correct. This one new job doesn't run Rakudo tests and uses the time saved to run a spectest. It doesn't mean any changes to the other jobs, they all continue to run Rakudo tests.

niner (approved):
Worth a try!

vrurg (approved):
Ok, let's give it a try. I always wanted specs to be ran too.

PR #4708, opened 2022-01-09
Protect some typechecks with try

vrurg:
If user defined `where` block is involved into type matching there is chance it would throw. Consider such case a failure and don't try optimizing.

See `LibXML` in #4705.

### DIFF
--- a/src/Perl6/Optimizer.nqp
+++ b/src/Perl6/Optimizer.nqp
@@ -1578,7 +1578,9 @@ my class SmartmatchOptimizer {
 
             return nqp::null() if $lhs.value.HOW.archetypes.generic;
 
-            my $matches := nqp::istype($lhs.value, $rhs.value);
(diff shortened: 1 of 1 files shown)


src/Perl6/Optimizer.nqp
```diff
@@ -1578,7 +1578,9 @@ my class SmartmatchOptimizer {
 
             return nqp::null() if $lhs.value.HOW.archetypes.generic;
 
-            my $matches := nqp::istype($lhs.value, $rhs.value);
+            # Wrap into try because for if there a user-defined `where`-block involved into typematching it might throw.
+            # Consider this case a failed typematch and proceed further.
+            my $matches := try { nqp::istype($lhs.value, $rhs.value) };
```
ugexe, line 1583: You shouldn't need a block for either of hese `try`s

vrurg: You're right. I've been overcautious here.

PR #4710, opened 2022-01-11
Make FatRat literals codegen as WVals

lizmat:
A literal as .123.FatRat was codegenned as a Rat WVal and a call to
the .FatRat method at runtime: it is now codegenned as a FatRat WVal.

Also, if a Rat literal exceeds the precision of a Rat, it will issue
a worry that mentions to either make it a Num or a FatRat.

### DIFF
--- a/src/Perl6/Actions.nqp
+++ b/src/Perl6/Actions.nqp
@@ -8394,7 +8394,17 @@ class Perl6::Actions is HLL::Actions does STDActions {
                 $partf := nqp::box_i(1, $Int);
             }
 
-            my $ast := $*W.add_constant('Rat', 'type_new', $parti, $partf, :nocache(1));
+            my $class := 'Rat';
+            if $<fatrattish> {
+                $class := 'FatRat';
+            } elsif nqp::isbig_I($parti) -> $numerator {
+                $/.typed_worry('X::Comp::Rat::Literal', :literal($/.Str),
+                  :$numerator, :denominator(nqp::isbig_I($partf)));
+            } elsif nqp::isbig_I($partf) {
+                $/.typed_worry('X::Comp::Rat::Literal', :literal($/.Str), :denominator);
+            }
+
+            my $ast := $*W.add_constant($class, 'type_new', $parti, $partf, :nocache(1));
             $ast.node($/);
             make $ast;
         }
--- a/src/Perl6/Grammar.nqp
+++ b/src/Perl6/Grammar.nqp
@@ -3295,8 +3295,8 @@ grammar Perl6::Grammar is HLL::Grammar does STD {
     token dec_number {
         :dba('decimal number')
         [
-        | $<coeff> = [               '.' <frac=.decint> ] <escale>?
-        | $<coeff> = [ <int=.decint> '.' <frac=.decint> ] <escale>?
+        | $<coeff> = [               '.' <frac=.decint> ] [ <escale> | <fatrattish> ]?
+        | $<coeff> = [ <int=.decint> '.' <frac=.decint> ] [ <escale> | <fatrattish> ]?
         | $<coeff> = [ <int=.decint>                    ] <escale>
         ]
     }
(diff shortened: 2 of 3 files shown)


codesections:
I have a hesitation with adding the warning (as @lizmat and I [discussed a bit on IRC](https://logs.liz.nl/raku-dev/2022-01-11.html#14:22)).  I'm happy to set this hesitation aside if no one shares it, but I want to mention it in case others feel the same way.

Here's my issue: I believe that this PR turns behavior that was previously added as an intentional feature into behavior that now raises a warning.  Specifically, in work for a [2018 grant](https://news.perlfoundation.org/post/grant_proposal_perl_6_bugfixin#benefits-to-the-perl-community) @zoffixznet added a "new non-infectious high-precision [type of `Rat` that is created] when the user provides high precision in their literals." The initial plan was for it to be a new type, `MidRat`, but the [design evolved](http://blogs.perl.org/users/zoffix_znet/2018/05/perl-6-car-tpf-grant-monthly-report-may-2018.html) and it became a normal `Rat` with higher-than-normal precision.

The motivation for these high-precision `Rat`s was to allow users to defer the decision about precision until the `Rat` was used to perform a calculation: if it was used in a calculation with `FatRat`s, then it would retain its precision, but if it was used in any other calculation, it would degrade into a `Num` and lose that precision.  Or, as the grant proposal put it:

> [This feature] allows us to define `pi` as a high-precision `MidRat` that will function similar current `Num`-based `pi` in normal operations, yet offer extra precision in `FatRat` operations.

Here's a code example that shows this feature in action before the current PR:

```raku
my $pi = 3.1415926535897932384626433; 
say "τ as a Num:    {$pi × 2}";        # OUTPUT: «τ as a Num:    6.283185307179586»
say "τ as a FatRat: {$pi × 2.FatRat}"; # OUTPUT: «τ as a FatRat: 6.2831853071795864769252866»
```

With this PR, the code still works the same (and produces the same output), but it also issues the following worry when `$pi` is first defined:
```
Potential difficulties:
    Numerator and denominator are too large in literal Rat definition and
    will lose precision.  Please consider creating a Num
    (3.1415926535897932384626433e0) for more performance and less precision,
    or to create a FatRat (3.1415926535897932384626433.FatRat) for more
    precision and less performance.
```

So deferring the decision about `$pi`'s precision – previously viewed as a feature worth adding to Raku – now generates a warning and implies that users should _not_ create "alomorphic-precision" `Rat`s such as `$pi`.

I don't have strong feelings about whether letting users have an alomorphic-precision `$pi` is a good idea or is likely to cause too much confusion; I can see design arguments on both sides.  But I do want to highlight that this PR represents a change to an existing feature and to make sure that we're OK with that.

lizmat:
So, is this the only real problem then?
````raku
my $pi =           3.1415926535897932384626433;
say $pi;         # 3.141592653589793230701985792
say $pi.FatRat;  # 3.1415926535897932384626433
````

jnthn:
I certainly think it's reasonable to constant-fold this, but doing so by extending the grammar feels hacky. What about `3.14.FatRat()`, `3.14 . FatRat`, and so forth? Perhaps better moved to the optimizer?

jnthn:
> But I do want to highlight that this PR represents a change to an existing feature and to make sure that we're OK with that.

At first I thought it was good to tell the language user about discarded precision in a literal, but now I understand that we *don't* lose precision if it actually ends up using in a fat context, I'm leaning more towards @codesections hesitation.

lizmat:
This PR consisted of 2 parts:

1. a warning
2. codegenning FatRat directly

1. has been opposed
2. is considered nice, but needs do be done in another way, which I a. do not know how, and b. will most likely not survive in RakuAST anyway.

So closing this PR.

MasterDuke17:
@lizmat were you able to benchmark any performance improvement? And if so, what was the benchmark code you used?

lizmat:
I didn't benchmark it, but technically, it is **removing** a method call at runtime.  So should be noticeable.

PR #4713, opened 2022-01-12
Make add_constant lookup setting symbols only

### LINKED ISSUE #4711: Using symbols (or just `Num`?) as an Enum key is allowed but breaks subsequent code
Declaring an `Enum` that creates a name clash with existing types does not throw an error.  For example, this works (at least for now…):

```raku
enum <Num>;
say Num.raku; # OUTPUT: «::Num»
```

However, the presence of that enum key causes confusing and LTA compile errors in later code (LTA both in that the message doesn't mention the `enum` and because it doesn't contain any line numbers).  Here's a golfed example:

```raku
enum <Num>;
multi f {};
# OUTPUT:
# ===SORRY!===
# This type cannot box a native number: P6opaque,
```

I haven't done enough testing to be sure if this issue occurs with other symbols besides `Num` – the `multi` above runs fine with _some_ types (`Int`, `Sub`, and `num`, for example).  And I'm not sure what the intended behavior is – that is, I don't know if `enum <Num>` is supposed to be allowed (and the bug is that the `multi` (and maybe other code?) doesn't resolve the name correctly) or if `enum <Num>` is _not_ intended to be legal (and the bug is that Rakudo doesn't throw a helpful error at declaration time).

I get this bug on released versions of Rakudo and on the current build (`v2021.12-101-g38063d487`).

vrurg:
The compiler is only uses this method with core types. Limiting lookups to setting only prevents LTA errors in user code.

Fixes #4711

### DIFF
--- a/src/Perl6/World.nqp
+++ b/src/Perl6/World.nqp
@@ -3034,7 +3034,7 @@ class Perl6::World is HLL::World {
         }
 
         # Find type object for the box typed we'll create.
-        my $type_obj := self.find_symbol(nqp::split('::', $type));
+        my $type_obj := self.find_symbol(nqp::split('::', $type), :setting-only);
 
         # Go by the primitive type we're boxing. Need to create
         # the boxed value and also code to produce it.


jnthn (approved):
The `rakuast` branch turns out to already do this. :-) https://github.com/rakudo/rakudo/blob/rakuast/src/Raku/ast/compunit.rakumod#L206

vrurg:
Use of `find_symbol` family across the current compiler is pretty much a mess. But luckily, RakuAST makes the task of cleaning up this mess redundant. :)

PR #4715, opened 2022-01-13
Make $*DISTRO more thread safe

### LINKED ISSUE #4714: $*DISTRO is not thread safe
## The Problem

```
$ raku -e '(1..1000000).hyper.map: { die unless $*DISTRO }'
[1]    15401 segmentation fault  raku -I. -e '(1..1000000).hyper.map: { die unless $*DISTRO }'
```

A less golfed example: https://gist.github.com/ugexe/4f6ec70c0edb7255bd1b99075dfb7cae

## Environment
* Operating system: macOS 12.1 (Monterey)
* Compiler version (`perl6 -v` or `raku -v`): `v2021.12.`

ugexe:
This prevents concurrent write attempts to `PROCESS::<$DISTRO>` when registering `$*DISTRO`.  It also avoids duplicated work in said situations, such as shell calls to `lsb_release -a`.

Fixes #4714

### DIFF
--- a/src/core.c/Distro.pm6
+++ b/src/core.c/Distro.pm6
@@ -20,101 +20,106 @@ class Distro does Systemic {
 
 # set up $*DISTRO
 Rakudo::Internals.REGISTER-DYNAMIC: '$*DISTRO', {
+    my $lock = INIT Lock.new;
+    $lock.protect: {
+        PROCESS::<$DISTRO> //= do {
 #?if jvm
-    my $properties := VM.new.properties;
-    my $name       := $properties<os.name>;
-    my $version    := $properties<os.version>;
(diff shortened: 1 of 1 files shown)


jonathanstowe (approved):

niner (changes requested):
The same problem affects all `PROCESS::` variables. I'd prefer we fix it for all of them, e.g. in `INITIALIZE-DYNAMIC`. We are gonna need to share the lock between those initializers anyway. Otherwise an initialization of e.g. `$*REPO` can still collide with a concurrent initialization of `$*DISTRO`

lizmat:
Fix as suggested by @niner in https://github.com/rakudo/rakudo/commit/768ebea783

PR #4716, opened 2022-01-14
Introduce NYI as a sub

lizmat:
Many places in the core setting use:
````raku
Failure.new: X::NYI.new: feature => "unimplemented feature"
````
in relatively hot code.  Another often occurring idiom is:
````raku
X::NYI.new(feature => "unimplemented feature").throw
````
This sub allows these calls to be simplified to:
````raku
NYI("unimplemented feature")        # saves 136 bytes

NYI("unimplemented feature").throw  # saves 134 bytes
````
Since the feels like this could be useful for module developers
and/or general application developers as well, I decided to *not*
make this an implementation detail, but an actual feature of the
Raku Programming Language.

Please note that technically, the .throw version would unnecessarily
create a Failure object, but by the time we're going to throw the
exception, creating an extra object should hardly be an issue anyway.


jonathanstowe (approved):

vrurg (approved):

moon-chilled:
Making this a user-level feature could add ambiguity as to whether a given feature is unimplemented by rakudo, or unimplemented by some userspace library.

lizmat:
Would that ambiguity be important?  The `X::NYI` exception is spec, so anybody can use it.  Whether or not it is more easily accessible using this sub.

PR #4721, opened 2022-01-16
Implement smartmatch as a dispatcher

vrurg:
Try to optimize some special cases whenever possible by means of utilizing new-disp. This PR introduces:

- `raku-boolify` dispatcher, implementing optimized universal HLL-boolification of both native- and HLL-types
- `raku-smartmatch-topicalized` dispatcher, used in place of the default `.ACCEPTS.Bool` chain
- optimized method `not` on `Match` which doesn't fall back to method `Bool`

Supporting additions which have value on their own:

- extended iterator on `Routine` which allows to (deep-)iterate over all possible candidates
- method `candidates` of `Routine` got two named parameters: `:local`, which is true by default, and `:with-proto`
- method `iterator` of `Routine` also got the above two nameds and additional boolean `:candidates` parameter
- implementation-detail method `IS-SETTING-ONLY` on `Routine` which ensures that all routine candidates are defined by the setting

None of the above additions changes the default `Routine` behaviors.

The meaning of the named parameters are:

- `:candidates` makes `Routine` iterator to iterate over candidates whereas the default behavior is to iterate over the routine itself only
- `:!local` triggers iteration not only over the local candidates (dispatchees), but to consider wrappers too; whereas a wrapper is a multi on its own, it's candidates are considered too. Happens recursively, so no matter how deep the wrapping/multi goes.
- `:with-proto` includes protos into the iteration too

### DIFF
--- a/src/Perl6/Actions.nqp
+++ b/src/Perl6/Actions.nqp
@@ -7408,6 +7408,7 @@ class Perl6::Actions is HLL::Actions does STDActions {
         # regex-ish code like m//, s///, or tr///
         my $boolify := !($negated || $rhs.ann('regex_match_code'));
 
+#?if !moar
         # Call $rhs.ACCEPTS( $_ ), where $_ is $lhs.
         $sm_call := QAST::Op.new(
                         :op('callmethod'), :name('ACCEPTS'),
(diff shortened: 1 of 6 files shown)


vrurg:
Unfortunately, this may remain an unsuccessful experiment. I expected to get some gains on cases like this:

```raku
sub t { Stringy }
sub v { "AAAA" }
for ^COUNT {
    v() ~~ t();
}
```

The point is that LHS is always `Stringy`, RHS is always a string, so the smartmatch itself can shortcut to a constant `True`. But preliminary benchmarking reveals, that the loop itself got about 6 times slower, comparing to 2021.10 release!

vrurg:
Update on the performance. It was incorrect to compare this PR to a previous release. Fixes for correct handling of junctions on LHS introduced performance penalty for cases, where static optimization is not possible. The case of `v() ~~ t()` is one of those.

This PR, thanks to new-disp approach, reclaims some of the lost performance back. In particular, the example above is about 2.5x times faster than it is on the master. 

Therefore I consider the PR as ready for merge, unless there are any suggestions to improve it.

PR #4723, opened 2022-01-16
implement sub hyper/race

gfldex:
This will allow `hyper` and `race` with feed operators.

tests:
```
use v6.*;
use Test;

my @a = 1..10;
my $b;

@a ==> hyper() ==> grep(*.is-prime) ==> sum() ==> $b;
ok $b »==» 17, 'hyper with feed operator';

@a ==> race() ==> grep(*.is-prime) ==> sum() ==> $b;
ok $b »==» 17, 'race with feed operator';

is hyper(:batch(42), @a).configuration.batch, 42, 'sub hyper passes on named args';
is race(:batch(42), @a).configuration.batch, 42, 'sub race passes on named args';
```

usecase:
```
sub is-palindrom(\v) { v eq v.flip }
sub is-boring(\s) { my \c = s.comb; c[0] eq c.all }
my @primes = ^10_000 .grep(*.is-prime);
(@primes X @primes)
==> hyper(:batch(1000))
==> grep({ (.[0] * .[1]).&{ .&is-palindrom & .&is-boring } })
==> map({ [ .[0,1], .[0] * .[1] ] })
==> unique(:as(*[1]))
==> sort(*[1] <=> *[1])
==> tail(20)
==> map({ .[0;0] ~ ' * ' ~ .[0;1] ~ ' = ' ~ .[1] })
==> join($?NL)
==> put();
```


src/core.e/Any-iterable-methods.pm6
```diff
@@ -7,3 +7,10 @@ multi sub rotor(**@cycle-and-thing, *%_) {
     @cycle-and-thing.tail.rotor(@cycle-and-thing.head(*-1), |%_)
 }
 
+multi sub hyper(Iterable:D \thing, *%_) {
```
vrurg, line 10: With new-disp, `\` doesn't bring in any performance gain anymore. And, if memory serves me right, a scalar parameter could even be better.

vrurg (changes requested):

vrurg:
Overall, a good thing to have. But since it's a new feature, spectests are needed too.

src/core.e/Any-iterable-methods.pm6
lizmat: Problem is, that this containerizes: `sub a(Iterable:D $a) { dd $a }; a [1,2,3];  # $[1, 2, 3]`.  Which may or may not be an issue in the receiving sub.  Compare this with `sub a(Iterable:D \a) { dd a }; a [1,2,3]; # [1, 2, 3]` and `sub a(Iterable:D $a is raw) { dd $a }; a [1,2,3];  # [1,2,3]`.

vrurg: I know. But in this particular case there is no difference if the `thing` is containerized, neither `race` nor `hyper` on `Iterable` care about this. And as much as I saw in other comments, a scalar is currently preferable over the raw form.

lizmat: Indeed.  I just wanted the op to be aware of the difference between the various ways.  :-)

jnthn:
The original design plan for `==>` was that it would automatically imply parallel execution, and achieve it by setting up a chain of producer-consumer workers (effectively, each stage in the pipeline is a `start` that consumes an input channel and puts results into an output channel). It's [described here](https://design.raku.org/S06.html#Feed_operators), and it draws a contrast with `hyper`/`race`. We're not beholden to implement things as once speculated, and of course there'd be other ways to end up with a feed encountering a `RaceSeq` or `HyperSeq` even without this `sub`, but if we're introducing it for the sake of working with `==>` and documenting it as such, it pretty much means we're not going to be able to do the speculated behavior in the future.

lizmat:
FWIW, I've implemented the parallel feed once, about 3 years ago: https://gist.github.com/lizmat/cecd280112920719f87cb5c9a8fe0ee7 I'm not sure anymore what that wasn't actually made into a PR

vrurg:
@lizmat you were not alone: #2903 :)

@jnthn perhaps feeds would just fall back to simple chaining upon encountering a `HyperSeq`?

PR #4725, opened 2022-01-18
Add missing precedence info for some infix ops

MasterDuke17:
This fixes the segfault uncovered by 71c62e7. Because `&infix:«+<»`
didn't have its precedence set, therefore the call to `.prec` went
through Code's `method prec`, which simple returns `my %`. For some
reason this is an HLL hash in the optimizer, so the atkey call in MoarVM
has problems. It shouldn't have segfaulted and probably still needs a
fix, but this change in Rakudo should happen regardless.

Fixes the failure in [MessagePack](https://modules.raku.org/dist/MessagePack) found by Blin in https://github.com/rakudo/rakudo/issues/4724.


lizmat:
These are also specified like this in the grammar, so indeed an oversight they weren't in Rakudo.

Fortunately, it will not be needed to keep these in two locations with RakuAST!

Issue #4734, opened 2022-01-23
Getting error 'p6decontrv' expects a QAST::WVal while compiling code.


Xliff:
## The Problem

Getting the following _intermittent_ error when recompiling the code

```
===SORRY!===
    p6decontrv expects a QAST::WVal as its first child
```

An --ll-exception backtrace has been posted in the comments

## Steps to Reproduce
1) Check out [p6-GLib](http://github.com/Xliff/p6-GLib)
2) Follow the build instructions
3) The error will occur somewhere in the build process

###  Linux
```
Linux Infinity4 5.13.0-25-generic #26-Ubuntu SMP Fri Jan 7 15:48:31 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux
```

### Rakudo
```
Welcome to Rakudo™ v2021.12-145-gac009507a.
Implementing the Raku® Programming Language v6.d.
Built on MoarVM version 2021.12-81-gf1101b95d.
```

Xliff:
--ll-exception:

```
p6decontrv expects a QAST::WVal as its first child
   at gen/moar/Ops.nqp:437  (/home/cbwood/Projects/rakudobrew/versions/moar-blead/install/share/perl6/lib/Perl6/Ops.moarvm:)
 from gen/moar/stage2/QAST.nqp:1664  (/home/cbwood/Projects/rakudobrew/versions/moar-blead/install/share/nqp/lib/QAST.moarvm:compile_op)
 from gen/moar/stage2/QAST.nqp:6517  (/home/cbwood/Projects/rakudobrew/versions/moar-blead/install/share/nqp/lib/QAST.moarvm:compile_node)
 from gen/moar/stage2/QAST.nqp:5691  (/home/cbwood/Projects/rakudobrew/versions/moar-blead/install/share/nqp/lib/QAST.moarvm:as_mast)
 from gen/moar/stage2/QAST.nqp:6485  (/home/cbwood/Projects/rakudobrew/versions/moar-blead/install/share/nqp/lib/QAST.moarvm:compile_all_the_stmts)
 from gen/moar/stage2/QAST.nqp:6070  (/home/cbwood/Projects/rakudobrew/versions/moar-blead/install/share/nqp/lib/QAST.moarvm:)
 from gen/moar/stage2/QAST.nqp:6040  (/home/cbwood/Projects/rakudobrew/versions/moar-blead/install/share/nqp/lib/QAST.moarvm:)
[...]

PR #4736, opened 2022-01-25
Eliminate p6decontrv ops caused by `*_u` ops

MasterDuke17:
Otherwise `my uint $a; sub foo { $a = 0 }` gives
`===SORRY!===
p6decontrv expects a QAST::WVal as its first child`.

Fixes https://github.com/rakudo/rakudo/issues/4734 after https://github.com/Raku/nqp/pull/758 is merged and NQP bumped.


niner (approved):
Oh, the optimizer is more important than one would expect... Thanks for the fix!

PR #4743, opened 2022-01-29
Fix method put failing on junctions

### LINKED ISSUE #4742: put method fails to print junctions
## The Problem

`put` as a method call results in an error when taking a Junction.

## Expected Behavior
```raku
(1|2).put;
put 1|2:;
(1|2).&put;
put 1|2;
```
All of the above should be identical, outputting `1` and `2` on separate lines.

## Actual Behavior
The first two lines result in 
```
This type cannot unbox to a native string: P6opaque, Junction
```
However the next two work fine. Interesting that `.&` works even though it supposedly "calls ... with method syntax", though perhaps this is special cased differently?

## Steps to Reproduce
``` raku -e "put 1|2:"```

## Environment
```
Welcome to Rakudo v2021.12.
Implementing the Raku Programming Language v6.d.
Built on MoarVM version 2021.12.
```

A note that this is a regression, since this same code at least worked in 2018.12

vrurg:
Resolves #4742. Aside of plain throwing `(1|2).put` also fix recursive cases when a junction is an eigenstate of another junction.

### DIFF
--- a/src/core.c/Junction.pm6
+++ b/src/core.c/Junction.pm6
@@ -284,6 +284,10 @@ my class Junction { # declared in BOOTSTRAP
         $!type ~ '(' ~ nqp::join(', ',$rakus) ~ ')'
     }
 
+    multi method put() {
+        self.THREAD: { .put }
+    }
+
     method CALL-ME(|c) {
         my \storage     := nqp::getattr(self, Junction, '$!eigenstates');
         my int $elems    = nqp::elems(storage);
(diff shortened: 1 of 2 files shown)


lizmat:
What's so special about `put`?   What not also `say` and `print` ?

vrurg:
Because `say` and `note` do `gist` and it doesn't autothread. `print` works because it falls back to `$*OUT.print`which just autothreads.

But `put` is about `Str` and it uses `nqp::concat`. So, `nqp::concat` fails when it gets a `Junction` instance from `.Str`.

PR #4746, opened 2022-01-30
Add "use show-precompilation" as a pragma

lizmat:
Activates the RAKUDO_PRECOMPILATION_PROGRESS environment variable,
perhaps making this feature a little more rememberable / accessible.

### DIFF
--- a/lib/show-precompilation.rakumod
+++ b/lib/show-precompilation.rakumod
@@ -1,0 +1,6 @@
+# The sole purpose of this module is to make it easier to activate the
+# logic for showing precompilation progress.
+
+%*ENV<RAKUDO_PRECOMPILATION_PROGRESS> = 1;
+
+# vim: expandtab shiftwidth=4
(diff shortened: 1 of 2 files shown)


niner:
When would it be appropriate to use this? What problem does it solve?

lizmat:
In my particular case, whenever I run a script and I start wondering why it takes so long.

I know some people have asked for the `SHOW_PRECOMPILATION_PROGRESS` feature: this just makes it easier to remember and access.

ugexe:
Generally I'd expect helper type code to fix common issues a common raku programmer has. If the reason no one has used `RAKUDO_PRECOMPILATION_PROGRESS` then this is demonstrably because a fair sampling of people can't remember the env name then this would make sense. However I suspect `RAKUDO_PRECOMPILATION_PROGRESS` is not being used enough to (imo) warrant additional exposed api.

PR #4747, opened 2022-01-30
Add "use named-anywhere" as a pragma

lizmat:
In an attempt to make the "use named arguments on the command line
at any location" feature more rememberable and accessible.

Sets named-anywhere key in the %*SUB-MAIN-OPTS has if it already
exists, or creates it in PROCESS:: with the named-anywhere key set
in it.

### DIFF
--- a/lib/named-anywhere.rakumod
+++ b/lib/named-anywhere.rakumod
@@ -1,0 +1,12 @@
+# The sole purpose of this module is to make it easier to activate
+# the possibility of having MAIN accept named arguments anywhere on
+# the command line.
+
+with %*SUB-MAIN-OPTS -> %opts {
+    %opts<named-anywhere> = True;
+}
+else {
+    PROCESS::<%SUB-MAIN-OPTS> = :named-anywhere;
+}
+
+# vim: expandtab shiftwidth=4
--- a/tools/build/install-core-dist.raku
+++ b/tools/build/install-core-dist.raku
@@ -17,6 +17,7 @@ my %provides =
     "snapper"                       => "lib/snapper.rakumod",
     "safe-snapper"                  => "lib/safe-snapper.rakumod",
     "BUILDPLAN"                     => "lib/BUILDPLAN.rakumod",
+    "named-anywhere"                => "lib/named-anywhere.rakumod",
 ;
 
 %provides<NativeCall::Dispatcher> = "lib/NativeCall/Dispatcher.rakumod"


raiph:
Maybe generalize the pragma name: `use MAIN :named-anywhere`?

cf `use experimental :macros;` rather than `use experimental-macros;`.

lizmat:
> Maybe generalize the pragma name: use MAIN :named-anywhere?

This feels more like an ecosystem module to me.  Also, I think this will overload the searchability of MAIN a little bit too much.

raiph:
**TL;DR**

* Either I'm misunderstanding your comments, or you're misunderstanding mine. :)
 
* I think `use MAIN :named-anywhere;` as the name for your pragma would be more discoverable than `use named-anywhere;`;

* I think `use MAIN :named-anywhere;` as the name for your pragma would be more self-explanatory than `use named-anywhere;`;

* I see merit in us having `use experimental :macros;`, `use experimental :pack;`, and `use experimental :cached;` rather than `use experimental-macros;`, `use experimental-pack;`, and `use experimental-cached;`, and think the same principle is even *more* applicable to your pragma.

> This feels more like an ecosystem module to me.

This confuses me. All I was suggesting was a name change.

Your rationale (which I upvoted) was to make using `:named-anywhere` more discoverable. If `use named-anywhere;` is discoverable, `use MAIN :named-anywhere;` is surely *more* discoverable due to addition of `MAIN`. (I disagree with your point about overloading, but will cover that later.)

As another potential benefit, the same pragma would then be available as a one stop shop for reuse/extension for any possible future built in pragmas related to `MAIN`. Though perhaps you're thinking there's little, or at least unknown, likelihood of there ever being any other `MAIN` pragmas core devs want to introduce. So YAGNI applies, which is fair enough. I can't see much harm coming from `use MAIN` ending up having only one use, but if it bugs you, then maybe broaden the "peg" to, say, `use CLI ...`, for covering any built in pragmas related to Raku(do)'s CLI processing, of which `MAIN` is merely a subset, or `use shell ...` for any `MAIN`, CLI or shell related built in pragmas? I'd say `use CORE ...` for any pragmas related to things in `CORE` is *too* broad, but maybe not?)

> Also, I think this will overload the searchability of MAIN a little bit too much.

This has me even more confused than your ecosystem comment!

Your pragma is specifically about `MAIN`, or at least the variable it alters is named `%*SUB-MAIN-OPTS `. How can it be too much for it to be listed when searching for `MAIN` related topics?

Also, my guess would be that having `MAIN` in the pragma name would be a very *good* thing from both a discoverability PoV, and a hint at what the feature is related to if it's encountered by chance.

Conversely, won't `use named-anywhere;` overload the searchability of `named` in a perhaps unfortunate way? My guess is that the first few times someone searches for `named` they will almost certainly not be thinking about named CLI  arguments. And they may well be a bit confused about ordinary named parameter/argument aspects -- it is a bit complicated in Raku! Encountering `named-anywhere`, without an immediate big `MAIN`, might draw them into reading about it, only to discover yet another complication of named arguments. That would be a negative in my book.

The discoverability / self-explanatory considerations were the MAIN reasons I suggested the name change; it seemed someone would be *more* likely to discover the pragma if it included `MAIN` in its name, and they'd be *more* likely to immediately guess it was something to do with named arguments in `MAIN` processing, whereas they'd be much *less* likely to discover it, and/or immediately feel oriented about what it was, if it was just `use named-anywhere;`.

At least, that's my 2c. (Or maybe 2cwt!: ))

raiph:
Hi @lizmat,

If the sole reason you chose not to merge this was my comments, I apologize for my comments and ask that you merge it. My primary reaction was a thumbs up, as it was for Leon and Jonathan, and I don't see any dissent from merging it. If it was for other reasons, I apologize for the noise of this comment.

lizmat:
No worries: I felt the PR had bitrotted conceptually  :-)

PR #4758, opened 2022-02-03
Fix precomp file rename race condition

ugexe:
Previously when two threads were trying to load the same module they could attempt to rename the same temporary file. This adds some uniquifying path extension parts to avoid the aforementioned scenario.

### DIFF
--- a/src/core.c/CompUnit/PrecompilationStore/File.pm6
+++ b/src/core.c/CompUnit/PrecompilationStore/File.pm6
@@ -322,7 +322,8 @@ class CompUnit::PrecompilationStore::File
       CompUnit::PrecompilationId:D $precomp-id,
       CompUnit::PrecompilationUnit:D $unit
     ) {
-        my $precomp-file := self!file($compiler-id, $precomp-id, :extension<.tmp>);
+        my $extension := self!tmp-extension;
+        my $precomp-file := self!file($compiler-id, $precomp-id, :$extension);
         $unit.save-to($precomp-file);
         my &rename-block = {
             $precomp-file.rename(self!file($compiler-id, $precomp-id));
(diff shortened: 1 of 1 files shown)


src/core.c/CompUnit/PrecompilationStore/File.pm6
```diff
@@ -368,6 +370,10 @@ class CompUnit::PrecompilationStore::File
          }
          $compiler-dir.rmdir;
     }
+
+    method !tmp-extension(--> Str:D) {
+        "." ~ now.Num ~ "." ~ nqp::getpid() ~ "." ~ nqp::objectid($*THREAD) ~ ".tmp";
```
niner, line 375: Wouldn't it be easier to just use a random number like `(^2**128).pick.base(36)`?

niner (approved):

src/core.c/CompUnit/PrecompilationStore/File.pm6
vrurg: Random doesn't mean unique. The above combination could happen to be non-unique in the only case when within one thread the method is called twice so fast that `now` is not changed. Well, maybe in a far away future... ;)

niner:
> Random doesn't mean unique. The above combination could happen to be
> non-unique in the only case when within one thread the method is called
> twice so fast that `now` is not changed. Well, maybe in a far away
> future... ;)

Well, yes and no. While of course there is a non-zero chance for a random 128 
bit number to cause collisions, it is practically impossible. There are
340282366920938463463374607431768211456 such numbers after all. That's big 
enough that practically all of cryptography relies on the impossibility of 
having random collisions. This includes our precomp file storage which relies 
on the uniqueness of 128 bit sha1 hashes.

PR #4761, opened 2022-02-06
Enable onlystar proto wrapping

vrurg:
Try to make better dispatching decisions over wrapped routines by adding `$!is-wrapped` attribute to `Routine` and guarding over it. This should work better because this way wrapped routine path is not taken when all wrappers are removed.

### DIFF
--- a/src/vm/moar/dispatchers.nqp
+++ b/src/vm/moar/dispatchers.nqp
@@ -776,7 +776,10 @@ nqp::dispatch('boot-syscall', 'dispatcher-register', 'raku-call', -> $capture {
             $with-name);
     }
     elsif nqp::istype_nd($callee, Routine) {
-        if nqp::can($callee, 'WRAPPERS') {
+        my $track_is-wrapped := nqp::dispatch('boot-syscall', 'dispatcher-track-attr',
+            $track_callee, Routine, '$!is-wrapped');
+        nqp::dispatch('boot-syscall', 'dispatcher-guard-literal', $track_is-wrapped);
+        if $callee.is-wrapped {
             nqp::dispatch('boot-syscall', 'dispatcher-delegate', 'raku-invoke-wrapped', $capture);
         }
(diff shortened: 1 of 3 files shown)


vrurg:
PR makes it possible to wrap an onlystar proto. See related test unfudged by Raku/roast#790

src/vm/moar/dispatchers.nqp
```diff
@@ -776,7 +776,10 @@ nqp::dispatch('boot-syscall', 'dispatcher-register', 'raku-call', -> $capture {
             $with-name);
     }
     elsif nqp::istype_nd($callee, Routine) {
-        if nqp::can($callee, 'WRAPPERS') {
+        my $track_is-wrapped := nqp::dispatch('boot-syscall', 'dispatcher-track-attr',
+            $track_callee, Routine, '$!is-wrapped');
+        nqp::dispatch('boot-syscall', 'dispatcher-guard-literal', $track_is-wrapped);
```
niner, line 781: This makes every call more expensive through the additional guards on `$!is-wrapped`. Guards are pretty cheap, so it may not matter. But there may be a way to avoid the additional cost.

After all, `$!is-wrapped` can only become true, once we have a `WRAPPERS` method, i.e. the Wrapped role is mixed into the callee. Unless we have that we don't have to bother checking (and tracking) `$!is-wrapped`. The existing, necessary type guard will make sure that we re-check once a routine got wrapped (and from then on track the attribute).

There's one caveat however: nqp::can is also not exactly cost free. So getting rid of it would speed up first-time dispatch. So it's quite possible that your change could improve performance overall as it is. Hard to say what's the right trade-off here.

niner (approved):

src/vm/moar/dispatchers.nqp
vrurg: @niner There is another case when a routine gets unwrapped. Guarding on the attribute avoids going the wrapped case path.

Issue #2814, opened 2019-04-05
Autothreading gone wrong?


JJ:
## The Problem

```
<a b c d e f>.classify( * eq 'a'|'c'|'f' ).say; #2813  OUTPUT: «{False => [a b b b c c d d d e e e f f], True => [a c f]}␤» 
(1..5).classify( * == 1|3|5 ).say; # OUTPUT: «{False => [1 2 2 2 3 3 4 4 4 5 5], True => [1 3 5]}␤»  
```

## Expected Behavior

No idea. At least, if they are repeated, they should be repeated the same number of times. If the Junction receives another member, non-junctioned members repeat themselves that number of times

```
07:54:26   jmerelo | m: (1..5).classify( * == 1|3|5|7 ).say # OUTPUT: «{False => [1 2 2 2 2 3 3 3 4 4 4 4 5 5 5], True => [1 3 5]}␤»                                │
```

## Environment

Reproducible in [Camelia](https://gist.github.com/Whateverable/11c9ac8ec81c1274c28d250f6b8dfb7f)

tony-o:
@jj did you mean to use `~~ 'a'|'c'|'f'` ?

JJ:
@tony-o no, I meant to use precisely that, as in the roast test. The classification is shown in the output, and if that's the case, b would be classified as any(False, True, False) and f as any(False, False, True). How come then a appears once classified as True, once as False, and c and f twice as True and once as False?
the fact that the elements of the list appear several times, I can grok that (although it keeps being mildly bewildering). But the fact that the first one of them appears one time less than the rest, that's what strikes me.

FCO:
maybe the returned value should be `.Bool` before used as key if its. a junction?

lizmat:
Golfed to:

    say <a>.classify: * eq "a"|"b";  # {False => [a], True => [a]}

Compare this to:

    say (* eq "a"|"b")("a");  # any(True, False)

And you should see why "a" gets added in both `True` and `False`.

FCO:
```perl6
<a b c d e f>.classify( * eq 'a'|'c'|'f' )
```
I can't see another reason to use that unless I want `{False => [<b d e>], True => [<a c f>]}` as response, but to do what I want, I could just add a `so`...

```perl6
<a b c d e f>.classify( so * eq 'a'|'c'|'f' )
```

I think it would make sense to `classify` do to Junctions the same thing `so` does ...

lizmat:
[...]

PR #4770, opened 2022-02-09
Correctly classify over junctions

vrurg:
_Potentially breaking change!_

Previously `classify` and `categorize` methods were autothreading over junctions internally, resulting in confusing results. In particular, where same value generally must only appear in a single class, use of a junction could result in duplication over different classes (see https://github.com/rakudo/rakudo/issues/2814).

This commit get things done right by putting junctions alongside other data types and allowing them to be the resulting hash keys. In other words:

    <a>.classify( * eq 'a' | 'b' ); # { any(True, False) => ['a']

To achieve this outcome the resulting hash is now parameterized over Mu-typed keys. While this rather extends the API in general, but the change could also be breaking if there is code relying on the resulting hash to autothread over junctional keys.

- resolves Raku/problem-solving#312
- fixes rakudo/rakudo#2814


vrurg:
Unless there're objections to this PR as a resolution of raku/problem-solving#312, even when ready to merge I would postpone it up until after the 2022.02 Rakudo release. If there be any fallout as a result of the merge then it better not make us postpone the release.

PR #4786, opened 2022-02-13
Improve error msg for traits passed invalid args

### LINKED ISSUE #4785: LTA error when traits are passed invalid arguments
For many traits, calling them with an invalid arguments throws an `X::Comp::Trait::Unknown` error that suggests other traits they might have meant.  This is a helpful message when the user has typo'ed the trait name or made a similar error, but is LTA when the problem is that an argument doesn't match the signature.  This creates a WAT when for builtin traits:

```raku
sub infix:<🦋>(|) is equiv«prefix:<i>» {}
# OUTPUT: ===SORRY!=== Error while compiling /home/runner/.code.tio
# Can't use unknown trait 'is equiv' in a sub declaration.
#    expecting any of:
#       rw raw hidden-from-backtrace hidden-from-USAGE pure default
#       DEPRECATED inlinable nodal prec equiv tighter looser assoc
#       leading_docs trailing_docs
```
.oO( "expecting `equiv` but can't use `equiv`...WAT?" )

And with a non-builtin trait, it's a bit less of a WAT but possibly even more confusing:
```raku
multi trait_mod:<is>(Routine:D $r,  :$foo) {}
sub infix:<🦋>(|) is foo<bar> {}
# OUTPUT: ===SORRY!=== Error while compiling /home/runner/.code.tio
# Can't use unknown trait 'is foo' in a sub declaration.
#   expecting any of:
#       rw raw hidden-from-backtrace hidden-from-USAGE pure default
#       DEPRECATED inlinable nodal prec equiv tighter looser assoc
#       leading_docs trailing_docs
```
This error makes it sound like there's a problem with how `is foo` was _declared_ rather than with how it was called, which points the user in the wrong direction entirely.

This seems like it'll be an easy fix, and I plan to submit a PR soon.

codesections:
Resolves  #4785.

Previously, traits called with invalid arguments would throw an `X::Trait::Unknown` error, which was confusing when the trait _was_
known.  With this PR, they'll now throw an `X::Trait::Invalid` error instead.

To correctly distinguish between unknown traits and invalidly called traits, we also now generate a list of known traits via introspection
rather than relying on a list of hard-coded traits.  This has the added benefit of generating more helpful `X::Trait::Unknown` error
messages when non-core traits have been defined (and eliminates the need to update the hard-coded list).

This PR also includes minor refactoring of the `X::Trait::*` code factor out some duplicated attributes into a role.  This refactoring introduces very minor spectest breakage: Two regression tests in [S12-attributes/instance.t](https://github.com/Raku/roast/blob/master/S12-attributes/instance.t#L693-L706) test the exact value of an attribute, and changing that value from `n attribute` to `an attribute` broke those tests.  I've prepared a corresponding Roast PR.

### DIFF
--- a/src/core.c/Exception.pm6
+++ b/src/core.c/Exception.pm6
@@ -1044,7 +1044,7 @@ my role X::Trait is Exception {
 }
 my class X::Trait::Invalid does X::Trait {
     has $.name;       # target of trait, e.g., '$foo' in `$foo is rw`
-    has $.reason;     # reason the the trait was invalid (optional)
+    has $.reason;     # reason the trait was invalid (optional)
     method message () {
         "Cannot use '$.type $.subtype' on $.declaring '$.name'"
          ~($!reason ?? " because:\n$!reason.indent(4)" !! '.');
--- a/src/core.c/Exception.pm6
+++ b/src/core.c/Exception.pm6
@@ -1037,33 +1037,33 @@ my class X::Worry::Precedence::Range is X::Worry {
     }
 }
 
-my class X::Trait::Invalid is Exception {
+my role X::Trait is Exception {
     has $.type;       # is, will, of etc.
     has $.subtype;    # wrong subtype being tried
-    has $.declaring;  # variable, sub, parameter, etc.
-    has $.name;       # '$foo', '@bar', etc.
+    has $.declaring;  # variable, sub, parameter, etc. (optional)
+}
+my class X::Trait::Invalid does X::Trait {
+    has $.name;       # target of trait, e.g., '$foo' in `$foo is rw`
+    has $.reason;     # reason the the trait was invalid (optional)
     method message () {
-        "Cannot use '$.type $.subtype' on $.declaring '$.name'."
+        "Cannot use '$.type $.subtype' on $.declaring '$.name'"
+         ~($!reason ?? " because:\n$!reason.indent(4)" !! '.');
     }
 }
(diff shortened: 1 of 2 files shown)


src/core.c/Exception.pm6
```diff
@@ -1037,33 +1037,33 @@ my class X::Worry::Precedence::Range is X::Worry {
...
-my class X::Trait::Invalid is Exception {
+my role X::Trait is Exception {
     has $.type;       # is, will, of etc.
     has $.subtype;    # wrong subtype being tried
-    has $.declaring;  # variable, sub, parameter, etc.
-    has $.name;       # '$foo', '@bar', etc.
+    has $.declaring;  # variable, sub, parameter, etc. (optional)
+}
+my class X::Trait::Invalid does X::Trait {
+    has $.name;       # target of trait, e.g., '$foo' in `$foo is rw`
```
MasterDuke17, line 1046: “the the”

codesections: Pardon? I don't see a duplicated "the".  Or maybe I don't understand?

MasterDuke17: Oh, this got put one line up. Below, “reason the the trait”.

codesections: Oh, I thought I checked the surrounding lines.  I guess not [well enough](https://en.wikipedia.org/wiki/Paris_in_the_the_Spring)! Thanks; fixed.

PR #4787, opened 2022-02-13
Remove unneeded/confusing can('prec') checks

codesections:
This builds on commits caba0d3 and a7ccfc6 (admittedly not that promptly! @lizmat  made those commits in 2017).  The prior commits moved the `.prec` method to Code to avoid needing to check whether various functions `can('prec')`; this PR removes six such `can('prec')` checks in the `is equiv`, `is tighter` and `is looser` traits.

Three of the six checks were literally unnecessary because they were on a Routine (which is guaranteed to have Code's prec method).  The
other 3 were on variables with a Callable (`&`) type constraint and thus could technically have been necessary in the rare case that someone
passed a non-Code callable.  However, even in that edge case, the check generated a confusing error message that suggested the trait
needed to be given an operator (when, in reality, any Code will do).

To address this edge case, this commit changes the parameter type constraint from Callable to Code, which has the same effect but
avoids the confusing error message.  (The replacement error message is also slightly confusing, but that confusing message is addressed by #4786).

### DIFF
--- a/src/core.c/traits.pm6
+++ b/src/core.c/traits.pm6
@@ -218,26 +218,20 @@ multi sub trait_mod:<is>(Routine:D $r, :prec(%spec)!) {
     0;
 }
 # three other trait_mod sub for equiv/tighter/looser in operators.pm6
-multi sub trait_mod:<is>(Routine:D $r, :&equiv!) {
-    nqp::can(&equiv, 'prec')
-        ?? trait_mod:<is>($r, :prec(&equiv.prec))
-        !! die "Routine given to equiv does not appear to be an operator";
+multi sub trait_mod:<is>(Routine:D $r, Code :$equiv!) {
+    trait_mod:<is>($r, :prec($equiv.prec));
     $r.prec<assoc>:delete;
 }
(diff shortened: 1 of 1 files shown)


MasterDuke17:
Just curious, but do you know of a reason `.prec` shouldn't be moved to `Callable`?

codesections:
> do you know of a reason .prec shouldn't be moved to Callable?

I don't, and I thought about proposing that but didn't for a few reasons: first, Callable is currently [only 7 lines](https://github.com/rakudo/rakudo/blob/master/src/core.c/Callable.pm6) and I figured there might be a reason we've kept it that small.  Second, @lizmat first moved `.prec` to Routine and then to Code, so it looked like some thought had already been put into where it should live.  Finally, ForeignCode does Callable, and I haven't spent any time dealing with ForeignCode (even as a Raku user, much less in Rakudo dev work) – so I really wasn't sure whether there'd be any problem with adding prec methods to ForeignCode.

If you tell me that none of those are problems, then yeah, moving the methods to Callable would probably be cleaner – and I'd be happy to make that change.

MasterDuke17:
Heh, those were all the questions I was hoping you had answers to.

PR #4788, opened 2022-02-14
Fix typo'd missing : in .nominal_type signature

codesections:
The `Parameter.nominal_type method` had a signature that clearly intended to constrain the invocation to be `:D`, but was missing the
trailing `:`.

Fixing this issue causes a TODO'ed Roast test to pass; I'll fix that in a companion PR.


lizmat:
Good spot!

PR #4790, opened 2022-02-14
Allow optional : on LHS of signature binding declaration

### LINKED ISSUE #4789: The `:` in `my :($a, $b)` should be *optional* but is forbidden
Normally, the syntax for declaring a [Signature literal](https://docs.raku.org/type/Signature#Signature_literals) is to use a colon: `my $sig = :($a, $b)`.

[S02 - Signature objects](https://design.raku.org/S02.html#Signature_objects) states that

> Declarators generally make the colon optional:
> 
>     my ($a,$b,$c);      # parsed as signature

However, it currently appears that the colon isn't **optional**, it's forbidden:

```raku
my :($a, $b, $c); 
# OUTPUT: ===SORRY!=== Error while compiling /home/runner/.code.tio
# Malformed my
```

And, [according to bisectable](http://irclogs.raku.org/raku-dev/2022-02-14.html#16:55), it looks like we never allowed the `my :(…)` form.

I can see two benfits in allowing the `:(…)` form: First (and less importantly) it help emphasize that the LHS gets destructuring semantics rather than list assignment semantics, which might be useful for some users/in some contexts.  Maybe this would help with the Raku/problem-solving#307?  More importantly (imo), I think this would be nice for pedagogical/documentation reasons: it would be easier to teach/document `my ($a, $b) := (42, 47)` as a shorthand for `my :($a, $b) := (42, 47)` – but it's only a shorhand if the longhand form works!

(I *think* that allowing this behavior may be as easy as adding a `':'?` in the [`variable_declarator` token](https://github.com/rakudo/rakudo/blob/master/src/Perl6/Grammar.nqp#L2355), though I haven't tested that yet)

codesections:
When using signature binding during variable declaration, the normal syntax is

```raku
my ($a, $b) := (42, 47);
```

The `($a, $b)` on the LHS creates a Signature, which then has 42 and 47 bound to its parameters following the normal signature-binding
process.  But why does `($a, $b)` create a Signature rather than a List?  Normally, the syntax for creating a Signature is `:(...)`, not
`(...)`.

Raku lets us omit the `:` here because we're binding to a list of variables in the context of declaring those variables and, as
S03#Signature-objects says, a declaration context makes the signature-literal's `:` optional (though the `(` and `)` are still
required).  This makes perfect sense, because there's no way Raku can bind to a list of variables on the LHS *other* than by treating the
LHS as a Signature and binding the RHS to it.  So the `:` is a bit redundant in context, and Raku's decision to make it optional is
well-justified.

Currently, however, the `:` is **not** optional – it's forbidden. Prior to this commit, the following incorrectly throws an
X::Syntax::Malformed error:

```raku
my :($a, $b) := (42, 47);
```

This commit fixes that error and gives the line the semantics described in S03 – that is, exactly the same signature-binding
semantics that it would have without the optional `:`.

In addition to being a bugfix that restores apparently intended behavior, allowing `my :($a, $b) := (42, 47)` as a (slightly) more
verbose alternative to `my ($a, $b) := (42, 47)` has a few advantages:

  1. It allows the user to more clearly express their intent that the LHS is a Signature.
  2. Because the user has communicated their intent, Raku can provide helpful error messages rather than silently doing something
     unintended.
  3. It makes Raku's semantics easier to explain to new users – because `my ($a, $b) := (42, 47)` is shorthand for `my :($a, $b) := (42, 47)`, it's confusing that the non-shorthand version isn't allowed.  In particular, the Raku docs have often been confusingly imprecise about variable-declaration signature binding, often calling it "destructuring assignment". I can't be sure, but I believe that normal signature literal syntax not being allowed during variable-declaration signature binding likely contributed to this confusion.

To implement the helpful error messages described in (2), this commit adds two error as shown below.  First, an X::Syntax::Variable::SignatureAssignment error that is thrown when a user tries to *assign* (rather than bind) to a variable-declaration
signature:

```raku
# without `:`
my (@a0, $b0) = ([1, 2], 3); # Unintended list assignment
note $b0; # OUTPUT: «(Any)»

# with `:`
my :(@a, $b) = ([1, 2], 3);
# ===SORRY!=== Error while compiling file.raku
# use assignment when declaring a variable via signature binding.
# Did you mean to use binding?  If so, use `:=` instead of `=`.
# Or did you mean to use list assignment?  If so, don't use `:(...)`
#    (the signature literal syntax) on the left-hand side.
# at file.raku:4
# ------> my :($a, $b)⏏ = (1, 2);
```

Second, an X::Syntax::Variable::SignatureWithoutInitializer error that's thrown when a user tries to use variable-declaration signature
binding without a RHS to bind to the Signature:

```raku
# without `:`
my ($a0, $b0); # Unintended list declaration of mutable variables

# with `:`
my :($a, $b);
# OUTPUT: ===SORRY!=== Error while compiling file.raku
# Variable declaration using a signature literal requires an initializer.
#     Did you mean to declare a list of variables with `(...)` instead of
#     a signature literal with `:(...)`?
# at file.raku:8
# ------> my :($a, $b)⏏;
```

Resolves #4789

### DIFF
--- a/src/Perl6/Grammar.nqp
+++ b/src/Perl6/Grammar.nqp
@@ -2222,6 +2222,7 @@ grammar Perl6::Grammar is HLL::Grammar does STD {
     token declarator {
         :my $*LEFTSIGIL := '';
         :my $*VARIABLE := '';
+        :my $*IN_SIG_DECL := 0;
         [
         # STD.pm6 uses <defterm> here, but we need different
         # action methods
@@ -2232,7 +2233,11 @@ grammar Perl6::Grammar is HLL::Grammar does STD {
           || <?{ $*SCOPE eq 'has' }> <.newpad> [<.ws> <initializer>]? { $*ATTR_INIT_BLOCK := $*W.pop_lexpad() }
           || [<.ws> <initializer>]?
           ]
-        | '(' ~ ')' <signature('variable')> [ <.ws> <trait>+ ]? [ <.ws> <initializer> ]?
+        |  '(' ~ ')' <signature('variable')> [ <.ws> <trait>+ ]? [ <.ws> <initializer> ]?
+        | ':(' ~ ')' <signature('variable')> [ <.ws> <trait>+ ]? [
+            { $*IN_SIG_DECL := 1}
+            [<.ws> <initializer> ]
+            || <.typed_panic: "X::Syntax::Variable::SignatureWithoutInitializer"> ]
         | <routine_declarator>
         | <regex_declarator>
         | <type_declarator>
@@ -3003,6 +3008,7 @@ grammar Perl6::Grammar is HLL::Grammar does STD {
 
     proto token initializer { <...> }
     token initializer:sym<=> {
+            [<!{$*IN_SIG_DECL // 0 }> || <.typed_panic: "X::Syntax::Variable::SignatureAssignment"> ]
         <sym>
         [
             <.ws>
--- a/src/core.c/Exception.pm6
+++ b/src/core.c/Exception.pm6
@@ -1659,6 +1659,23 @@ my class X::Syntax::Variable::Initializer does X::Syntax {
     method message() { "Cannot use variable $!name in declaration to initialize itself" }
 }
 
+my class X::Syntax::Variable::SignatureAssignment does X::Syntax {
+    method message() {
+        "Cannot use assignment when declaring a variable via signature binding.\n"
+        ~"    Did you mean to use binding?  If so, use `:=` instead of `=`.\n"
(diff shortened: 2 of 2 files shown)


raiph:
> `my ($a, $b) := (42, 47)` is shorthand for `my :($a, $b) := (42, 47)`.

I thought it was shorthand for `my :($a, $b) ::= (42, 47)`, no?

Have you pondered how intuition will work out if/when `::=` is implemented?

In particular, stuff like:

```
my ($a) := 1;

:($a) := 2;
:$a) ::= 3;
my $sig = :($a);
$sig ::= 4;

my :($a) := 5;
$sig ::= 6;
my :($a) ::= 7;
```
How many `$a`s are there and what should their values be at the end?

PR #4796, opened 2022-02-17
Make previous values in REPL accessible with `$*N`

lizmat:
Inspired by:
  https://www.reddit.com/r/rakulang/comments/sulhq6/how_to_retrieve_the_last_result_in_the_shell_like/

This will make the following sequence work:
````
[0] > 42
42
[1] > 666
666
[2] > $*0 + $*1
708
[3] > $*2
708
> @*_
[42 666 708 708]
````

### DIFF
--- a/src/core.c/REPL.pm6
+++ b/src/core.c/REPL.pm6
@@ -351,14 +351,13 @@ do {
                 return $!control-not-allowed;
             }
 
-            my $*_ := @*_[0];
             self.compiler.eval(
               $code.subst(/ '$*' \d+ /, { '@*_[' ~ $/.substr(2) ~ ']' }, :g),
               |%adverbs
             )
         }
 
-        method interactive_prompt() { '> ' }
+        method interactive_prompt($index) { "[$index] > " }
 
         method repl-loop(:$no-exit, *%adverbs) {
             my int $stopped;     # did we press CTRL-c just now?
@@ -367,7 +366,7 @@ do {
             signal(SIGINT).tap: {
                 exit if $stopped++;
                 say "Pressed CTRL-c, press CTRL-c again to exit";
-                print self.interactive_prompt;
+                print self.interactive_prompt(@previous-evals.elems);
             }
 #?endif
 
@@ -381,7 +380,7 @@ do {
             my $code;
             sub reset(--> Nil) {
                 $code = '';
-                $prompt = self.interactive_prompt;
+                $prompt = self.interactive_prompt(@previous-evals.elems);
             }
             reset;
 
@@ -428,8 +427,6 @@ do {
                     $!save_ctx := $*MAIN_CTX;
                 }
 
-                reset;
-
                 # Print the result if:
                 # - there wasn't some other output
                 # - the result is an *unhandled* Failure
@@ -440,8 +437,9 @@ do {
                 elsif $initial_out_position == $*OUT.tell
                     or nqp::istype($output, Failure) and not $output.handled {
                     self.repl-print($output);
-                    @previous-evals.unshift: $output;
+                    @previous-evals.push: $output;
                 }
+                reset;
 
                 # Why doesn't the catch-default in repl-eval catch all?
                 CATCH {
--- a/src/core.c/REPL.pm6
+++ b/src/core.c/REPL.pm6
@@ -361,12 +361,12 @@ do {
 
         method repl-loop(:$no-exit, *%adverbs) {
             my int $stopped;     # did we press CTRL-c just now?
-            my @previous-evals;  # the previously calculated values
+            my $previous-evals := IterationBuffer.new;  # previous values
 #?if !jvm
             signal(SIGINT).tap: {
                 exit if $stopped++;
                 say "Pressed CTRL-c, press CTRL-c again to exit";
-                print self.interactive_prompt(@previous-evals.elems);
+                print self.interactive_prompt($previous-evals.elems);
             }
 #?endif
 
@@ -380,7 +380,7 @@ do {
             my $code;
             sub reset(--> Nil) {
                 $code = '';
-                $prompt = self.interactive_prompt(@previous-evals.elems);
+                $prompt = self.interactive_prompt($previous-evals.elems);
             }
             reset;
 
@@ -408,7 +408,7 @@ do {
                 my $output is default(Nil) = self.repl-eval(
                     $code,
                     my $exception,
-                    @previous-evals,
+                    $previous-evals.List,
                     :outer_ctx($!save_ctx),
                     |%adverbs);
 
@@ -437,7 +437,7 @@ do {
                 elsif $initial_out_position == $*OUT.tell
                     or nqp::istype($output, Failure) and not $output.handled {
                     self.repl-print($output);
-                    @previous-evals.push: $output;
+                    $previous-evals.push: $output<>;
                 }
                 reset;
 
--- a/src/core.c/REPL.pm6
+++ b/src/core.c/REPL.pm6
@@ -351,7 +351,7 @@ do {
                 return $!control-not-allowed;
             }
 
-            self.compiler.eval('$_ := $*LAST-OUTPUT;' ~ $code, |%adverbs);
+            self.compiler.eval('$_ = $*LAST-OUTPUT;' ~ $code, |%adverbs);
         }
 
         method interactive_prompt() { '> ' }
--- a/src/core.c/REPL.pm6
+++ b/src/core.c/REPL.pm6
@@ -320,7 +320,7 @@ do {
             self.?teardown-line-editor;
         }
 
-        method repl-eval($code, \exception, *%adverbs) {
+        method repl-eval($code, \exception, $*LAST-OUTPUT, *%adverbs) {
 
             CATCH {
                 when X::Syntax::Missing {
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by Altai-man** [comment]
Will it break explicit setting of `$_`?

**Comment by lizmat** [comment]
Indeed it would.  Fixed with https://github.com/rakudo/rakudo/commit/857b448399  Thanks for the spot!

**Comment by codesections** [comment]
Hmm, this both seems very useful and potentially confusing – I can remember a time or two when I've had a line of code work in the REPL but not in my program because I'd forgotten that I'd previously assigned the global `$_` to some value.  And Raku has some confusing-to-beginners rule about when `$_` is bound (e.g., `if 'foo' { $_ }` vs `if 'foo' -> $_ { $_ }`, so I'm slightly reluctant about the idea of adding a set of REPL-only semantics on top of that?

What would you think about the idea of using `$__` or `$*_` instead?  I'm partial to `$*_`, because the variable really *is* dynamic – it can have different values each time the line is run.

Using a variable other than `$_` would also let us access the result of the previous calculation in an expression that sets `$_` (e.g., `say $_ ~ $*_ with 42`)

(I suppose if we wanted to be *really* cool, we could make it work like shell history/regex matches, and have `$*0`, `$*1`, `$*2` … etc for accessing anything in the REPL history, but that seems like a bigger lift).

**Comment by vrurg** [comment]
> Hmm, this both seems very useful and potentially confusing – I can remember a time or two when I've had a line of code work in the REPL but not in my program because I'd forgotten that I'd previously assigned the global `$_` to some value.

There are already specifics in REPL behavior which makes it necessary to remember about them. Having one more? Why not as long as it makes life easier when experimenting.

The following is just me thinking aloud.

Speaking about the history of results, an array of some kind would be easier to maintain. But then there is a question as to what'd be better way to index it? Making it in the order of statements (direct) allows to re-use any previous statement, which is using the array, from the history without a change. But then referring to the most recent result would look something like `@*_[*-1]`; or `@*_.tail` which is easier to type.

Contrary, the reverse order would mean unshifting the results into the array and then the most recent is referred to as `@*_[0]`. But this may break re-usability of the history.

**Comment by lizmat** [comment]
Whatever we decide on `$*N` or not, I think it makes sense to at least support `$_`, possibly in addition to any other features?

**Comment by codesections** [comment]
> Whatever we decide on `$*N` or not, I think it makes sense to at least support `$_`, possibly in addition to any other features?

I still think using `$_` is/could be confusing, since `$_` already gets set in so many cases.  I personally get a `No such method '$name' for invocation of type 'Any'` error with at least some frequency from times I made a typo in the REPL (e.g., the extra `,` in `(1, 2, 3),.sum`) and would find it very confusing if that instead used the last result.

(Or is the idea that it would just apply to literal `$_`s and not with `.method` calls?  If so, that's even more confusing/inconsistent).

`$*_` solves all of the above at the cost of only one character.  And it emphasizes the variable's dynamic nature: If I ran `$_ + 2` five times, I'd expect to get the same answer each time, but wouldn't if `$_` is the previous result.

**Comment by lizmat** [comment]
I think all comments are addressed with https://github.com/rakudo/rakudo/pull/4796/commits/473b178ed01297c2c7cd92e9b497a9b8739a5416 ?

**Comment by lizmat** [comment]
@japhb Please note that the numbering is backward: `$*0` would always been the most recently evaluated value, and `$*1` would be the one before that.  So I don't see how that prefix would help.

Unless we want to introduce some kind of switch?

**Comment by japhb** [comment]
Or syntax.  `$*-1` kinda visually fits `@a[*-1]` while still being currently unused syntax.  And then you have positive numbers for indexes starting at the session beginning, and negative for moving back into the past from the present.

**Comment by lizmat** [comment]
@Larry decided it was bad to use negative indices to indicate "from the end".  In conjunction with `$*0` syntax in the REPL, `$*-1` might give the wrong impression to the unexperienced.

**Comment by lizmat** [comment]
> how should saved history (from previous sessions) enter in to this? I would argue against such a thing

In total agreement there.

Re the other remarks: I think this is now getting bikeshedded to death  :-(   Piling features upon features whereas the original question was just to be able to use the previous value  :-)

**Comment by FCO** [comment]
I haven’t tried, but will it break the case of someone wanting to make a state nameless variable times a number (not that I think that's very used? Ex:

```raku
say ++$*2 for ^10
```

(https://glot.io/snippets/g79l9xs7o6)

lizmat:
That would indeed not do what you expect it to do.  But adding some whitespace would help: `say ++$ *2 for ^10` or `say ++$* 2 for ^10`

japhb:
Sorry, not my intent to bikeshed to death.

lizmat:
@japhb No worries.

I can take out the `$*N` and `@*_[]` functionality for now, and just leave `$*_` to answer the original op's question, and @codesections comments?

jubilatious1:
We see @lizmat's amazing coding abilities on display--again. 
However, in my honest opinion, this is an unnecessary and potentially unsafe addition to the REPL. 

First of all, let's look at the original reddit thread:
https://www.reddit.com/r/rakulang/comments/sulhq6/how_to_retrieve_the_last_result_in_the_shell_like/

It's not clear to me that the OP understands the `$_` topic variable in Raku. Without understanding the `topic` variable, how can the OP make a request for a `return` variable? It seems easy enough to accomplish what the OP desires with the following code:

```
> $_ = 0.3 + 0.2
0.5
> put $_
0.5
>
```

Easy-peasy. So why introduce another variable? I recall @TimToady stating in a conversation regarding command line flags (paraphrasing), 'some of those flags weren't pulling their weight, so they went away.'  I would gently suggest the same could be said for this `return` variable proposal:  noting how simple it is to code in the REPL already, won't a `return` variable "not be pulling its own weight" eventually, if not sooner (i.e. immediately)?

Before committing this pull-request, I would like someone to do a brief survey of popular REPLs. Anyone? 
My guess is that once we look at other REPLs, we'll see that none of the major REPLs auto-save return values. Which means Raku will have achieved parity with the TI-30X handheld calculator, and not one of the major scripting languages extant today:

https://www.youtube.com/watch?v=4auFYPCZh1U

I know the R-programming language (and its REPL) very well: R does **NOT** auto-save `return` values into a variable. The general course of action is to write/edit/save your command history when working in the REPL, not to expect the last return value to always be present for you. And why **should** you expect that? If you're pulling data from 3 or 4 data sources, it's doubtful that the last return will contain what you need for the next line of code, anyway. So really, having a `return` variable slurp up the last return seems remarkably inefficient, in terms of both storage space and CPU cycles.

Now I know there are modules in R that perform memoization: 
https://cran.r-project.org/web/packages/R.cache/index.html
And recommendations for memoiziation in R:
https://www.r-bloggers.com/2018/10/optimize-your-r-code-using-memoization/
And here's a SO link to a question on memoization in Python: 
https://stackoverflow.com/questions/1988804/what-is-memoization-and-how-can-i-use-it-in-python

So maybe, caching return values can be accomplished in the REPL, by loading a module beforehand?

Finally, security issues: What this pull request will do is alway auto-save return values. So if you return a secret key or hash--it's autosaved. If you accidentally type your password into the REPL--it's autosaved.  If you botch your code with `while` (potentially returning Terabytes of data) --it's autosaved. 

All these caveats suggest that caching return values in the REPL **should not** be the default, but rather a functionality that might be added by loading a module before starting the REPL.

@Util 
@tbrowder 
@doomvox

MasterDuke17:
> My guess is that once we look at other REPLs, we'll see that none of the major REPLs auto-save return values.

FWIW, Node's repl does this (https://www.digitalocean.com/community/tutorials/how-to-use-the-node-js-repl#step-3-mastering-repl-shortcuts "To quickly get the last evaluated value, use the underscore character. At the prompt, type _ and press ENTER:").

lizmat:
> So maybe, caching return values can be accomplished in the REPL, by loading a module beforehand?

That would require either something that would need to be built-in into the REPL, or an API to the REPL.  I'm not sure what would be the point then.

FWIW, it feels to me now that `$*_` should just become `$_`, and that we could leave in the `$*N` and `@*_[]` feature for those willing to know about it and use it.  I don't think in practice this should affect the functioning of the REPL in any way, and it's hardly any overhead.

jubilatious1:
> FWIW, Node's repl does this (https://www.digitalocean.com/community/tutorials/how-to-use-the-node-js-repl#step-3-mastering-repl-shortcuts "To quickly get the last evaluated value, use the underscore character. At the prompt, type _ and press ENTER:").

I've never used it. I was thinking about Ruby's `irb` shell and the Python shell. 

With regards to `node.js`'s REPL, it seems to function somewhat differently from Raku's REPL. Assignments at the command line do **NOT** autoprint values, but rather result in `undefined` being printed out. (Only string literals entered at the command line appear to autoprint the actual string literal value). 

So if you're working in the `node.js` REPL, it _might_ make sense to access previously evaluated entries, since there is NO autoprinting return.

I  don't have `node.js` REPL installed, but frankly I see nothing wrong with using `$_` to solve the Reddit OP's question with direct, readable code (i.e. no additional variables necessary):

```
> $_ = 0.3 + 0.2
0.5
> put $_
0.5
>
> $_ = $_ + 0.2
0.7
> $_ = $_ + 0.2
0.9
> exit
```

(And I'm guessing there are _already_ extant ways to shorten the code above).

Issue #4297, opened 2021-04-04
=defn headers don't render at all in Pod::To::Text


Leont:
## The Problem
```
=defn Happy 
When you're not blue.
 
=defn Blue 
When you're not happy.
```
I get the following output:
```
When you're not blue.


When you're not happy.

```
When I expected the output:
```
Happy
    When you're not blue.

Blue
    When you're not happy.
```

## Environment
```
Welcome to Rakudo(tm) v2020.12.
Implementing the Raku(tm) programming language v6.d.
Built on MoarVM version 2020.12.
```

PR #4803, opened 2022-02-24
Fix/test Pod::To::Text rendering of =defn elements

dwarring:
Pod::To:::Text needed updating, to output Pod::Defn term() attribute.


Leont:
This looks like it fixes #4297

lizmat (approved):

PR #4804, opened 2022-02-28
expose $!file and $!line in Label

gfldex:
To aid introspection and allow self referencing warnings/error messages, expose `$!file` and `$!line`. Make `Label.Str` useful.

Tests:

```
use nqp;
a-label: my $line = $?LINE; my $file = nqp::getlexdyn('$?FILES');

is a-label.line, $line, 'exposed $!line in Label';
is a-label.file, $file, 'exposed $!file in Label';
is a-label.Str, "{a-label.name} $file:$line", 'expose $!file and $!line in Label.Str';
```

### DIFF
--- a/src/core.c/Label.pm6
+++ b/src/core.c/Label.pm6
@@ -15,13 +15,23 @@ my class Label {
         nqp::bindattr($obj, Label, '$!postmatch', nqp::p6box_s($postmatch));
         $obj
     }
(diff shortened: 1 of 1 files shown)


lizmat:
Implemented slightly differently, Wenzel P.P. Peppmeyer++ for the spot

PR #4807, opened 2022-03-02
Some optimizations for regex substitution

MasterDuke17:
Together, these optimizations mean that `my $t = ("errors.md.tmpl" xx 5).join(" foo "); my $page; my $s = now; for ^10_000 { $page = $t; $page ~~ s:g[ \w+ ] = "a" }; say now - $s;` goes from ~0.57 to ~0.47s and `my $t = ("errors.md.tmpl" xx 1_000).join(" foo "); my $page; my $s = now; for ^1_000 { $page = $t; $page ~~ s:g[ "foo" ] = "a" }; say now - $s;` goes from ~2.55 to ~2.00s.

Rakudo builds ok and passes `make m-test m-spectest`. However, I'm not 100% sure about the last commit, maybe @jnthn or @moritz could verify it?

### DIFF
--- a/src/Perl6/Actions.nqp
+++ b/src/Perl6/Actions.nqp
@@ -8862,11 +8862,17 @@ class Perl6::Actions is HLL::Actions does STDActions {
             ));
             $right.push(WANTED($<sibble><right>.ast,'quote:s'));
         }
-        my $rep_block := $*SUBST_RHS_BLOCK;
-        $rep_block.push(QAST::Stmts.new($right, :node($<sibble><right>)));
-        my $closure := block_closure(reference_to_code_object(
-            $*W.create_code_obj_and_add_child($rep_block, 'Code'),
-            $rep_block));
+        my $replacement;
+        if $right.has_compile_time_value {
+            $replacement := QAST::SVal.new( :value($right.compile_time_value) );
+        }
+        else {
+            my $rep_block := $*SUBST_RHS_BLOCK;
+            $rep_block.push(QAST::Stmts.new($right, :node($<sibble><right>)));
+            $replacement := block_closure(reference_to_code_object(
+                $*W.create_code_obj_and_add_child($rep_block, 'Code'),
+                $rep_block));
+        }
 
         # self.match($rx_coderef, |%options);
         my $past := QAST::Op.new( :node($/), :op('callmethod'), :name('match'),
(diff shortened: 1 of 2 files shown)


lizmat:
Re https://github.com/rakudo/rakudo/pull/4807/commits/72f2f40055f4307b209de08435c997fa2b5e4531: Perhaps turn this into `cds = $m if nqp::iscont(cds) && SDS`  ??

lizmat:
Also, if you're looking at performance improvements: most of the parameters to `!APPLY-MATCHES` would be better off being `$` sigilled with a type, as I understand then less code would need to be generated when it can be proven the parameter is not an `Iterable`

MasterDuke17:
> Also, if you're looking at performance improvements: most of the parameters to `!APPLY-MATCHES` would be better off being `$` sigilled with a type, as I understand then less code would need to be generated when it can be proven the parameter is not an `Iterable`

I just tried and it was the same to slightly slower.

lizmat:
@MasterDuke17 Just to be sure, did you just add a sigil, or did you also add a non-`Iterable` type constraint?

MasterDuke17:
> @MasterDuke17 Just to be sure, did you just add a sigil, or did you also add a non-`Iterable` type constraint?

```patch
-    method !APPLY-MATCHES(\matches,$replacement,\cds,\SDS,\word_by_word,\space,\case,\mark) {
+    method !APPLY-MATCHES($matches,$replacement,\cds,int $SDS,int $word_by_word,int $space,int $case,int $mark) {
```

lizmat:
@MasterDuke17 and with `s:g/int/Int:D/` ?

MasterDuke17:
> @MasterDuke17 and with `s:g/int/Int:D/` ?

Doesn't seem to change the timings compared to no sigil. Callgrind shows a small-but-noticeable increase in instructions for the version with sigils. It is a little bit surprising.

lizmat:
Indeed it is  :-(

MasterDuke17:
With the most recent commit (4503faa), the times further drop to ~0.41 and ~1.98s respectively. This commit could also be pulled out as a separate change if this PR is not merged. In that case, the times are ~0.49s and ~2.36s respectively (compared to ~0.57 and ~2.55 for master).

MasterDuke17:
A potential problem case for 72f2f40 of `grammar G { token TOP { <thingy>* }; token thingy { "thingy" } }; class A { method thingy($/) { my $p = "page"; $p ~~ s:g[ \w ] = "a"; $p } }; my $g = G.parse("thingythingy", actions => A.new); say $g` gives the same error as on master (`Cannot assign to a readonly variable or a value`), so going to go ahead and merge this.

PR #4814, opened 2022-03-10
Restore optimization for given/when with Numeric cases

MasterDuke17:
In the case of something like `given "a string" { when 5 { say "five" };
default { say "not five" } }`, the currently-not-working optimization
attempts to turn off `.Numeric` returning an (expensive to create) Failure
via `val` when we just need to know that it didn't succeed (i.e.,
`"a string".Numeric != 5`). However, just moving the `:fail-or-nil` argument
to the correct method call causes `Use of Nil in numeric context`. So
instead change the option to return `NaN` instead of `Nil` and rename
accordingly.

However, this does mean we can't use the same optimization
when smartmatching against a Range (since `NaN` could be a valid thing
we're checking). But I suspect that this is less common of an occurance
so the tradeoff is worth it.

This test code:
```raku
my $a;
$a = do given "hi" {
    when 0 { "zero" };
    when 1 { "one" };
    when 2 { "two" };
    when 3 { "three" };
    when 4 { "four" };
    when 5 { "five" };
    when "hi" { "hello" };
    when * > 40 { "> 40" };
    default { "default" } } for ^10_000;
say now - INIT now;
say $a
```
drops from ~2s to ~0.25s.

### DIFF
--- a/src/Perl6/Optimizer.nqp
+++ b/src/Perl6/Optimizer.nqp
@@ -1728,7 +1728,29 @@ my class SmartmatchOptimizer {
         if $method eq 'Numeric' {
             my $fail-or-nil := QAST::Op.new( :op('hllbool'), QAST::IVal.new( :value(1) ) );
             $fail-or-nil.named('fail-or-nil');
-            $method_call.push($fail-or-nil);
+            $method_call[0].push($fail-or-nil);
+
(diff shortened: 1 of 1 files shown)


vrurg:
I propose a slightly different approach without tradeoffs. Use a private type object as an indicator, not something from the public API. I.e. make it something like:

```
my class NOT-NUM { };

if val($s, :fail-or-sym(NOT-NUM)) === NOT-NUM { ... }
```

The only problem I can think of at the moment is where the symbol is to be declared to make it available at the runtime.

PR #4821, opened 2022-03-19
CURI: pre-compile longest names first

lizmat:
The idea being that the longest names have fewer dependenencies
(in general).  Also make sure that we only sort the entries once.

It would probably be better if we could specify an order in which
modules should be pre-compiled.  But alas, the current META6.json
format is a hash, so without order.  And we need some sort of order
to get reproducible builds.

### DIFF
--- a/src/core.c/CompUnit/Repository/Installation.pm6
+++ b/src/core.c/CompUnit/Repository/Installation.pm6
@@ -283,12 +283,13 @@ sub MAIN(:$name, :$auth, :$ver, *@, *%) {
             my %done;
 
             my $compiler-id = CompUnit::PrecompilationId.new-without-check($*RAKU.compiler.id);
-            for %provides.sort {
+            my @provides = %provides.sort(*.key).reverse;
+            for @provides {
                 my $id = CompUnit::PrecompilationId.new-without-check($_.value.values[0]<file>);
                 $precomp.store.delete($compiler-id, $id);
             }
 
-            for %provides.sort {
+            for @provides {
                 my $id = $_.value.values[0]<file>;
                 my $source = $sources-dir.add($id);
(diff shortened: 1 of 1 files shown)


ugexe:
Unless this is more than an idea I don’t agree with changing this. If there is a measurable difference then sure, but let’s measure that then.

lizmat:
In a lot of distributions (see also `lib/NativeCall.rakumod`), you typically see:
````
Foo
Foo::Bar
Foo::BaZ
````
and then `lib/Foo.rakumod` doing:
````raku
unit class Foo;
use Foo::Bar;
use Foo::Baz;
````
Now, in the current way it is sorted, this will *first* precompile `lib/Foo.rakumod`.  When it does that, it will recurse and precompile `lib/Foo/Bar.rakumod` when it encounters the `use Foo::Bar`.  And then also precompile `lib/Foo/Baz.rakumod` when it encounters the `use Foo::Baz`.

When it is then done with `lib/Foo.rakumod`, it tries to precompile `lib/Foo/Bar.rakumod`, then notices it has already been pre-compiled, and goes on to the next.  Which is (albeit a very little amount of) extra work.  Same for `lib/Foo/Baz.rakumod`.

Ideally, we would really like to build a dependency chain of the modules *inside the distribution*, so that we could potentially hyper pre-compilation.  But alas, we don't actually know the dependency chain until we've pre-compiled everything.  Sorting this way, is in my view the next best thing.

For most distributions, this would not matter.  For some big distributions (like `DateTime::TimeZone`, which takes about **4 minutes** to install), it might.

ugexe:
I know how it works. But I don’t think changing the implementation on assumptions (the assumption here being the new order is better in practice) is good idea.

lizmat:
> I know how it works

@ugexe my explanation was not intended for you alone  :-)

PR #4822, opened 2022-03-19
Move CUR::Staging into core

lizmat:
Yes, it is only needed for very specific applications.  But it is
a very small module, and it simplifies the various install scripts
significantly (most notable "install-core-dist.raku").  And it won't
break code that actually do a 'use CompUnit::Repository::Staging".

And it reduces the core setting compilation with about a second.
Which, on a normal core development day, I'm doing a *lot*.


src/core.c/CompUnit/Repository/Staging.pm6
```diff
@@ -0,0 +1,49 @@
...
+    method resource($dist-id, $key) {
+        # check if the dist is installed here
+        try self.distribution($dist-id)
+          # we have the dist, so it's safe to access the resource the normal way
+          ?? callsame()
+          # lookup failed, so it's probably not installed here
+          !! $!parent.resource($dist-id, $key)
+    }
+
+    sub files($io) {
```
ugexe, line 30: Maybe I’m missing something since the commit message doesn’t mention the refactoring, but what is this for?

ugexe, line 30: Leftover debugging code?

lizmat:
Ah, oops, sorry, something snuck in there that shouldn't have snuck in there.

lizmat:
Closing this PR for now, as it bitrotted

PR #4832, opened 2022-03-27
Add implementation-detail Compiler.cver method

lizmat:
The raw string version of the Compiler.new.version is needed in
several places.  This commit introduces a new Compiler.cver method
that returns that raw string, and removes the private $!cver attribute
from CURI, as it was basically just the raw compiler version.

No functional changes, just less copy-and-paste and a little less
memory use for CURI instances (not that we have many of them).


lizmat:
`CURI.$!cver` turned out to be dead wood.  So this PR no longer makes sense.

PR #4837, opened 2022-03-30
Re-imagine CompUnit::Repository::Spec.from-string

lizmat:
This was using regular expressions and meta ops, for code that is
really pretty simple (apart from the option partsing).  Since this
is being called multiple times for *every* pre-compilation, it seems
to make sense to streamline this code and with fewer regexes and
no meta-ops.

Underlying reason was that another upcoming fix broke because
meta-ops were being used *before* the Rakudo::Metaops mapper had
been initialized.

### DIFF
--- a/src/core.c/CompUnit/Repository/Spec.pm6
+++ b/src/core.c/CompUnit/Repository/Spec.pm6
@@ -1,36 +1,52 @@
 class CompUnit::Repository::Spec {
-    has $.short-id;
+    has Str:D $.short-id is required;
+    has Str:D $.Str      is required;
+    has Str:D $.path = "";
     has %.options;
-    has $.path;
-    has $.Str;
-    method from-string(Str:D $spec, :$default-short-id = 'file') {
-        return unless $spec.chars;
-        # something we understand
-        if $spec.contains('#') {
-            if $spec ~~ /^
-              <before .>
-              [
-                $<type>=[ <.ident>+ % '::' ]
-                $<options>=[ '#' $<option-name>=\w+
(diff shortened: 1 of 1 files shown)


ugexe (approved):
Nice. I wonder if all the examples below (and not just the last one) should work (they don't work in previous releases):
```
❯ raku -e 'say CompUnit::Repository::Spec.from-string("foo#url<https://foo>#/")'
CompUnit::Repository::Spec.new(short-id => "foo", options => {}, path => "url<https:/foo>#", Str => "foo#url<https:/foo>#")

❯ raku -e 'say CompUnit::Repository::Spec.from-string("foo#url<foo.com>#/")'
CompUnit::Repository::Spec.new(short-id => "foo", options => {}, path => "url<foo.com>#", Str => "foo#url<foo.com>#")

❯ raku -e 'say CompUnit::Repository::Spec.from-string("foo#url<fooDOTcom>#/")'
CompUnit::Repository::Spec.new(short-id => "foo", options => {:url("fooDOTcom")}, path => "/", Str => "foo#url<fooDOTcom>#/")
```

lizmat:
These now parse correctly (I think):
````raku
say CompUnit::Repository::Spec.from-string("foo#url<https://foo>#/";
# CompUnit::Repository::Spec.new(short-id => "foo", Str => "foo#url<https://foo>#/", path => "/", options => {:url("https://foo")})
say CompUnit::Repository::Spec.from-string("foo#url<foo.com>#/";
CompUnit::Repository::Spec.new(short-id => "foo", Str => "foo#url<foo.com>#/", path => "/", options => {:url("foo.com")})
````

ugexe (approved):

PR #4840, opened 2022-03-31
Improve typecheck errors on e.g. subsets

lizmat:
Check whether there's a WHY on the expected value, and show the
WHY if there is one in the message.  Example:

    #| Only "foo" is acceptable!
    subset Foo of Str where $_ eq "foo";

    sub a(Foo $a) { dd }
    a "bar";
    # Constraint type check failed in binding to parameter '$a'; expected
    Foo: (Only "foo" is acceptable!) but got Str ("bar")

### DIFF
--- a/src/core.c/Exception.pm6
+++ b/src/core.c/Exception.pm6
@@ -2523,10 +2523,13 @@ my class X::TypeCheck is Exception {
             !! $raku
     }
     method expectedn() {
+        my $why := $!expected.WHY;
         nqp::eqaddr($!got.WHAT, $!expected.WHAT)
           ?? Rakudo::Internals.MAYBE-STRING($!expected, :method<raku>)
           !! nqp::can($!expected.HOW, 'name')
-            ?? $!expected.^name
+            ?? $why
+              ?? $!expected.^name ~ ": ($why)"
+              !! $!expected.^name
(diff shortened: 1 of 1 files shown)


vrurg:
I feel dubious about this particular approach. `WHY` is generally for documenting things. So, one would have to choose wether to document or provide an explanation for an error.

Perhaps an attribute on the meta-class + a trait would do better? Something like `subset Foo is cause("We expect this") ...`?

jnthn:
Think I agree with @vrurg on this. Also wonder whether a `will` trait is nice, in that it can be given when value and decide how to produce that, if appropriate, thus replacing the entire "expected Foo but got Bar" part:

```
# Don't emit the value here, as it can end up in error logs!
subset Password of Str where *.chars >= 8 will complain { "expected a password of at least 8 chars" }
# But this one is fine
subset Percent of Int where 0 <= $_ <= 100 will complain { "$_ is out of range for a percentage" }
```

But maybe I'm over-thinking it. :-)

lizmat:
I like the `will complain` idea, as you could potentially also apply this on `where` clauses?
````raku
sub a($a will complain("needs 'foo'") where $_ eq 'foo') { ... }
````

PR #4842, opened 2022-04-01
Add INSIDE_EMACS environment variable

anquegi:
I experiment some problems when running raku REPL Inside emacs whith the Line Editors activated, I think that is better to disable Line editor  checking the environment variable INSIDE_EMACS

### DIFF
--- a/docs/line-editor.pod
+++ b/docs/line-editor.pod
@@ -40,6 +40,12 @@ This specifies the preferred line editor to use; valid values are C<Readline>,
 C<Linenoise>, C<LineEditor>, and C<none>.  A value of C<none> is useful if you
 want to avoid the recommendation message upon REPL startup.
 
+=item INSIDE_EMACS
+
+This disables line editor while execcuting REPL inside Emacs, see for further
+reference:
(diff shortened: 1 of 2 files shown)


coke:
Should we specifically call out Emacs here, or use something more generic? (is this something that Emacs already has support for?)

anquegi:
Emacs itsel lfprovides this functionality when you execute the REPL inside a comint buffer, which is the case of [raku-repl.el](https://github.com/Raku/raku-mode/blob/master/raku-repl.el), so when you use raku with [Terminal::LineEditor](https://modules.raku.org/dist/Terminal::LineEditor:zef:zef:japhb), also this REPL becomes unusable. 
Emacs also provides the funcionalities that an a line editor provide so it will be no necesssary to load one.
For example in case of the ruby REPL we can have a configuration file that we can disable the line editors while INSIDE emacs [see](https://github.com/nonsequitur/inf-ruby#bugs)
So I think that we can use this to control this, will something similar to use rlwrap when you are inside emacs
[Emacs References](https://www.gnu.org/software/emacs/manual/html_node/emacs/Interactive-Shell.html)

lizmat:
Thank you!

PR #4855, opened 2022-04-07
Put CATCH blocks at the start of the scope

lizmat:
CATCH blocks affect the **entire** scope they're in.  So they should
be as close to the start of the scope they're in, so that you realize
that they can fire well before you think they'd might fire.

This did not change the Rakudo::Internals::JSON class, as that is
being kept in sync with JSON::Fast.

### DIFF
--- a/src/core.c/Backtrace.pm6
+++ b/src/core.c/Backtrace.pm6
@@ -278,6 +278,14 @@ my class Backtrace {
     method nice(Backtrace:D: :$oneline) {
         my $setting = %*ENV<RAKUDO_BACKTRACE_SETTING>;
         try {
+            CATCH {
+                default {
+                    return "<Internal error while creating backtrace: $_.message() $_.backtrace.full().\n"
+                        ~ "Please report this as a bug (mail to rakudobug@perl.org)\n",
+                        ~ "and re-run with the --ll-exception command line option\n"
+                        ~ "to get more information about your error>";
+                }
+            }
             my @frames;
             my Int $i = self.next-interesting-index(-1);
             while $i.defined {
@@ -297,14 +305,6 @@ my class Backtrace {
                 }
                 last if $oneline;
                 $i = self.next-interesting-index($i, :$setting);
-            }
-            CATCH {
-                default {
-                    return "<Internal error while creating backtrace: $_.message() $_.backtrace.full().\n"
-                        ~ "Please report this as a bug (mail to rakudobug@perl.org)\n",
-                        ~ "and re-run with the --ll-exception command line option\n"
-                        ~ "to get more information about your error>";
-                }
             }
             @frames.join;
         }
--- a/src/core.c/CompUnit/PrecompilationRepository.pm6
+++ b/src/core.c/CompUnit/PrecompilationRepository.pm6
@@ -79,18 +79,18 @@ class CompUnit::PrecompilationRepository::Default
     }
 
     method !load-handle-for-path(CompUnit::PrecompilationUnit:D $unit) {
-        $!RMD("Loading precompiled\n$unit")
-          if $!RMD;
-
         my $preserve_global := nqp::ifnull(nqp::gethllsym('Raku','GLOBAL'),Mu);
-        my $handle := CompUnit::Loader.load-precompilation-file($unit.bytecode-handle);
-        nqp::bindhllsym('Raku', 'GLOBAL', $preserve_global);
         CATCH {
             default {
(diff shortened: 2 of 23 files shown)


raiph:
> You say it made sense, but you're not telling me why.

Right. That was deliberate. I was trying to keep things short. Also, I'm not convinced it would help to tell you why.

I found your view compelling. Your abstract point was about maintaining code in general. That made (some) sense to me. Your concrete point was maintaining core code you spend a great deal of energy on. That made (a lot of) sense. You'd created a PR. That was 10X more compelling still. You'd described your view, it seemed plausible other maintainers would agree with it, and created a PR. End of story, for the most part.

> My reasoning ...

As I've hopefully now made clear, I found your original explanation compelling when you introduced it. Your elaboration covers points others may appreciate but fwiw I'd already presumed them (they were obvious to me from your original comment). So I won't address them further.

----

I thought about stopping my comment here. (Is there a way to write "spoiler" style / hidden-till-clicked parts of GH comments?)

Feel free to not read the rest of this comment.

There are oddities in the way I think and communicate that may contribute to misunderstanding. Hopefully the following will steer you toward there being less misunderstanding.

First, I urge anyone who reads what I write to never presume I disagree with anything unless I precisely address it and say I disagree. Similarly, I urge folk to never presume I agree either. Worse, it's best not to presume I think the same now as I did when I wrote something. I continuously change my mind in response to new information. Some might say I do so to a pathological degree, and indeed that sometimes includes me, but it is what it is and it is often a major source of misunderstanding.

Second, more generally, please consider taking all assumptions about normal human behavior off of the table when dealing with me, with, perhaps, the sole exception of knowing that I aspire to be loving. I've reason to think this will greatly simplify dealing with me. (Even if I all too often fail to live up to my aspiration. As you know, life is non-stop learning.)

Third, I think there are folk who are awesome, both in general, and in the particular aspect that they consistently arrive at productive outcomes in their own good time (which is often as quick as something could reasonably have possibly happened). You're one of them.

Fourth, whenever I have a problem with anything anyone says or does, it often comes back to an awareness of "Who am I to say?". While this is so even for the likes of Putin, I find that for many folk I find beautiful and inspiring, including folk I've met, however briefly, like you and Wendy and Larry and Gloria and niner, and others who I at least have had the privilege to be similarly influenced by even if I haven't met them, like jnthn, Audrey, and Damian, I know it's best for me to put aside any differences in view I might have, or rather think I have (in reality we're all seeing the same universe of love, even if sometimes it doesn't seem that way), and let things proceed.

But, fifth, while the previous aspects generally lead me to be silent even if I think I have something worth sharing, I sometimes ignore that inhibition and speak up anyway.

My prior comments in this PR were as a result of me thinking it might at least make sense to write something short, after I reviewed your changes and saw you had changed the last value of many blocks. I thought about, and then decided against, researching to see if any of the changed return values were relied upon by any code. A moments reflection led me to think you would know more or less immediately, and the only issue was whether you had thought about it, so I instead concentrated on writing a succinct comment in case you hand't thought about it.

Then I realized it was probably better to mention my sense that `CATCH` at the end making a difference seemed like a bug to me. (And I see you agree with that. I had already guessed that this would likely be fixed, if viable, as part of RakuAST and see you affirmed that too.)

And then as part of writing *that*, I realized it would just about be worth mentioning that I often place phasers such as `CATCH` and `CONTROL` at the end of blocks, if it seems the best way to express the code.

Perhaps this works for me because I have the kind of mind where, if any given lexical scope isn't completed in a single screenful, my mind is automatically aware there may be a "catch", so to speak. This may well be related to a combination of my perhaps pathological state of never believing I know anything for sure, and, on the other hand, knowing that I can always scroll down and phasers such as `CATCH` will jump out at me. I don't know; but I do know it's plausibly not relevant to mention it as a counter to your point, because I don't maintain core code, and almost certainly not worth writing a book as I have done here; but possibly worth mentioning in case it gave you what I view as healthy food for thought -- yet another perspective, regardless of whether it seems to make sense from one's own.

> if a `CATCH` block only makes sense later in a scope, it is used in an incorrect way, as an exception "before" the `CATCH` block will be caught by that `CATCH` block anyway.

Sure. I did not mean to imply that a phaser block *only* made sense at some particular point in a scope. The whole point of them is their placement doesn't matter (ignoring the issue that some phasers allow multiples of the same phaser kind in a single block and follow FIFO or LIFO handling). I just meant that I've always thought it made sense (to *me*) to put them at the end of their enclosing block (*as a default*).

lizmat:
Wow, that was a lot to take in  :-)

Thank you for your kind words.

PR #4858, opened 2022-04-11
Fix regression from unreviewed commits

ugexe:


ugexe:
I've asked multiple times to create PRs in this area of the code. So much so that I've stopped watching commits to this repo sometime last week. However I noticed some of my code stopped working today so I managed to catch this.

lizmat:
If I run all the tests that I know of, and nothing breaks.  But then you tell me I broke something, and revert what I did **WITHOUT** telling me what broke, that's not very helpful now, is it?

If anything, I would expect a test to be added so that it won't happen again.  Otherwise, what is the point of testing????

ugexe:
It should be obvious when you are breaking a public API. If you had submitted a PR it would have been caught and you would have been informed of what was wrong. However, since you refuse to submit PRs in this area of the code despite me having asked multiple times politely I am basically put between a rock and a hard place.

> Otherwise, what is the point of testing????

What is the point of pull requests?

ugexe:
To add to the above: testing probably won't catch security issues. It should be concerning to developers if the code they are running is not being reviewed at all.

lizmat:
Four of the **seven** Pull Requests made by me in the past 10 days, have not had *any* response.  Most of the PR's that I merged myself in the past months, did not get any review at all.

ugexe:
Is there a number of pull requests that one can have open before they don't need to have their code looked at anymore? Have I not mentioned specifically "this area of the code" every time I've brought up requesting reviews?

vrurg:
> Four of the **seven** Pull Requests made by me in the past 10 days, have not had _any_ response. Most of the PR's that I merged myself in the past months, did not get any review at all.

Merging an un-reviewed PR would have an excuse of nobody willing to check it out. At least. Not to mentions a couple of other safety advantages of PR procedures.

PR #4859, opened 2022-04-11
Rename CompUnit::PrecompilationStore::File -> FileSystem

lizmat:
For a PR by popular demand

### DIFF
--- a/src/core.c/CompUnit/PrecompilationStore/FileSystem.pm6
+++ b/src/core.c/CompUnit/PrecompilationStore/FileSystem.pm6
@@ -1,4 +1,4 @@
-class CompUnit::PrecompilationStore::File
+class CompUnit::PrecompilationStore::FileSystem
   does CompUnit::PrecompilationStore
 {
     has IO::Path:D $.prefix is built(:bind) is required;
--- a/src/core.c/CompUnit/PrecompilationStore/FileSystem.pm6
+++ b/src/core.c/CompUnit/PrecompilationStore/FileSystem.pm6
@@ -241,4 +241,16 @@ class CompUnit::PrecompilationStore::FileSystem
     }
 }
 
+class CompUnit::PrecompilationStore::File
+  is CompUnit::PrecompilationStore::FileSystem {
+
+    method new(|) {
+        DEPRECATED(
+          "the 'CompUnit::PrecompilationStore::FileSystem' class",
+          :what("Use of the 'CompUnit::PrecompilationStore::File' class")
+        );
+        nextsame;
+    }
+}
+
 # vim: expandtab shiftwidth=4
--- a/src/core.c/CompUnit/Repository/FileSystem.pm6
+++ b/src/core.c/CompUnit/Repository/FileSystem.pm6
@@ -328,7 +328,7 @@ class CompUnit::Repository::FileSystem
     }
 
     method precomp-store(--> CompUnit::PrecompilationStore:D) {
-        $!precomp-store //= CompUnit::PrecompilationStore::File.new(
+        $!precomp-store //= CompUnit::PrecompilationStore::FileSystem.new(
             :prefix(self.prefix.add('.precomp')),
         )
     }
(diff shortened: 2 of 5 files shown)


lizmat:
From the original commit message:

- to use the same logic as with CUR::FileSystem
- other modules with ::File at the end are about a specific file
    
This had me stumped for about an hour.  Hope this will not stump any more people in the future.

ugexe:
Renaming things breaks our existing public API

ugexe (changes requested):
Renaming things breaks our existing public API.

Alright, I see now we are adding a deprecation and adding an alias. However I'm still not entirely convinced changing this is worth it. If I went around changing code that *didnt make sense to me* it would probably not make sense to everyone. IMO implying there is a tight coupling with `CUR::FileSystem` is not a good idea.

lizmat:
> Renaming things breaks our existing public API.

AFAICS, `CompUnit::PrecompilationStore::File` is **not** documented.  There *is* one test for it, but it is **not** in roast.  I therefore concluded it was basically an implementation detail.  If code depends on implementation details, it runs the risk of breaking.

> IMO implying there is a tight coupling with CUR::FileSystem is not a good idea.

I'm not implying there is a tight coupling with CURFS.  I'm implying there's a cognitive dissonance because the postfix `::File` is being used in different meanings. 

Naming things is hard.  Being consistent is hard.

There are currently 3 classes that end in `::File`, and one that ends in `::FileSystem`.

- `CompUnit::PrecompilationUnit::File` which represents a single file containing bytecode.

- `CompUnit::PrecompilationDependency::File` which refers to a precompiled dependency in a single file

- `CompUnit::Repository::FileSystem` which maintains information about distributions on a file system.

And now:

- `CompUnit::PrecompilationStore::File`  Seeing the other classes, would you expect this to be about a single file in a precompilation store, or about a precompilation store  in a file system?

And yes, looking back on it, I should have taken the loss of being led into a wrong way of understanding, rather than trying to make sure that the people after me do not wind up in the same alley of thought.

ugexe:
> If code depends on implementation details, it runs the risk of breaking.

This doesn't always excuse breaking code, especially code that would be caught in a review. If you honestly believe your understanding of this part of the code is good enough that no one else needs to look at it first then I'm not sure what else I can do. Ironically some of the changes are because "Production", yet what serious production environments would find unreviewed code changes continually being pushed as anything but a potential risk?

Further, as mentioned earlier, how are security issues supposed to be discovered? This is module loading code, and there are security implications to how these things are handled. For example using `../`, which we unfortunately do for `resources` detection for `-Ilib`. If things like these aren't crossing your mind when you are making these changes then they *definitely* need a review. Maybe that doesn't apply specifically to this scenario, but it very well could and you very well might not know enough to know someone more familiar needs to look at it.

I'm aware you have only good intentions with your changes, but I found my anxiety level increases after watching master break multiple times in the last month and having my request for peer review be generally ignored time after time. I don't want to be in a position where my desire for quality control, security, etc are viewed as undesirable, a waste of time, etc, and at ends with the most prolific committer to the project.

lizmat:
> This doesn't always excuse breaking code

Indeed.  I should have put in a deprecation cycle, which I have now done in the PR.  I could also have added that, had you not reverted my commits.

>  If you honestly believe your understanding of this part of the code is good enough that no one else needs to look at it first

This was a simple rename of a class that I still think is an implementation detail, because it is not tested in roast, and there is no documentation for it.

More generally, indeed I will only push changes if I think I'm understanding the code well enough to know it won't break anything.  I'm not 100% right.

Regarding code review: to make review easier, I'm not creating massive commits, but many smaller ones.  If I'm not sure about the changes, or I know it is going to require more eyes on it (e.g. because it is a code style issue, or an additional feature), I make a PR.  Which I then merge after about two weeks if nobody left any comments on it.

If I would make a PR of every commit I do, I'm not sure I would want to work on the core anymore.  It is bad enough that for **each** cycle I need to wait **95** seconds before the setting is compiled, and about **6 minutes** before testing completes.  Now, multiply that by 2 weeks before I can merge PR's that are without comments.  That is not a workable situation for me.

I *have* worked on projects where all changes needed at least 2 approvals by other team members.  I would gladly work in such a situation for Rakudo, but we simply do not have enough team members that have time enough to do reviews.  Until then, I will depend on my own judgement on whether something should just be a commit or a PR.  And I will get it wrong sometimes.  I think it has gotten us pretty far so far.

lizmat:
So, should I close or merge this PR?  :-)

PR #4861, opened 2022-04-12
Implement `will complain` trait

vrurg:
Works for types and parameters. Need work to get individual variables working.

A followup to #4840.

### DIFF
--- a/lib/experimental.rakumod
+++ b/lib/experimental.rakumod
@@ -192,4 +192,73 @@ package EXPORT::pack {
     OUR::{'&pack'}           := &pack;
 }
 
+package EXPORT::will-complain {
+    multi sub trait_mod:<will>(Mu:U \type, &complainee, :complain($)!) {
+        if type.HOW.archetypes.composable {
+            X::Comp::Trait::Invalid.new(
+                file       => $?FILE,
+                line       => $?LINE,
+                type       => 'will',
+                subtype    => 'complain',
+                declaring  => 'role',
+                name       => type.^name
+            ).throw
+        }
+        type.HOW does Metamodel::Explaining unless nqp::istype(type.HOW, Metamodel::Explaining);
+        type.HOW.SET-COMPLAINEE(&complainee);
+    }
+    multi sub trait_mod:<will>(Parameter:D $param, &complainee, :complain($)!) {
+        # Parameters often doesn't have an associated container descriptor.
+        # Therefore applying directly to the parameter object iyself.
+        $param does Metamodel::Explaining unless nqp::istype(nqp::decont($param), Metamodel::Explaining);
+        $param.SET-COMPLAINEE(&complainee)
+    }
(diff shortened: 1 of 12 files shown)


vrurg:
@jnthn, @lizmat what do you think?

The idea is for the trait to be applicable to anything what could cause typechecking error. Basically, it does already work for parameters, but individual variables requirie something to be done to `X::TypeCheck::Assignment`.

I see no point in changing syntax of the trait until `RakuAST` lands into the release. But I'm considering a possibility of `will complain($complainee)` syntax, where `$complainee` could be a callable, an invocable object via `CALL-ME`, or something stringifyable.

vrurg:
I have gotten it to work as intended. Here is cases where it now does what's intended:

```raku
class Foo will complain { "it's disappointing to get a {.^name} instead of Foo-like" } {
    has Int $.a is rw will complain { "you offer me " ~ .raku ~ ", but with all the respect: an integer, please!" };
}
```

```raku
sub foo(Str:D $s will complain { "expecting a string with an 'a' in it" } where *.contains("a")) {
}
```

```raku
subset IntD of Int:D will complain { "only definite integers, not " ~ .raku };
```

```raku
my Str $s will complain { "gimme a string, not " ~ .raku ~ "!!!" };
my Str %h will complain { "non-string hash values are prohibited" };
my Int @a will complain { "this array must be all integer" };
```

```raku
enum BAR will complain { "I'd like something BAR-ish" } <bar1 bar2 bar3>;
```

There could still be overlooked cases where it would fail, but I can't think of any particular one at the moment.

raiph:
What are multiple dispatch `will complain` semantics?

Presumably bind failure of any given candidate for a call is typically *not* a cause for complaint. So naturally coders will typically not add a `will complain`, or if they refactor an only to a multi they'll typically remove any `will complain`. Or, if they add or keep a `will complain` it'll be because the candidate is a last resort (and they want to add a custom complaint if a bind to that candidate fails).

That said, I can also imagine cases where a coder wants a `will complain` to merely register a *potential* complaint, one that is to only be delivered if the call ultimately fails *all* candidates.

vrurg:
@raiph You seemingly missing the primary point of `will complain`. We cannot use it for bare everything. I don't even see a point in giving extra explanations on candidate match failures – simply because the current error message is detailed enough.

Complaints are to resolve a problem of complex type checks where for one or other reason we cannot infer developer's intentions about a type object. In such cases we give it out to them to explain what was meant by a constraint applied.

So, if you see what `Exception`s are referring to the `Metamodel::Explain` type, these are one way or another related to assignments, bindings, etc.

Another nuance to be noted is that the trait only affects error messages reported by exceptions. No exception – no respect is paid to the trait.

Hopefully, this answers your question.

raiph:
> No exception – no respect is paid to the trait.

.oO ( `will catch` )

Thanks for answering. Happy 2023.

PR #4862, opened 2022-04-12
Add CompUnit::DependencySpecification.lookup-id method

lizmat:
And use it.   I guess technically any CUR can devise its own way of
encoding a shortname lookup into a filename, but it feels that it
being in the CUDS is more appropriate.

### DIFF
--- a/src/core.c/CompUnit/DependencySpecification.pm6
+++ b/src/core.c/CompUnit/DependencySpecification.pm6
@@ -84,6 +84,9 @@ class CompUnit::DependencySpecification {
           ValueObjAt
         )
     }
+
+    # basename inside the short-dir
+    method lookup-id(--> Str:D) { nqp::sha1($!short-name) }
 }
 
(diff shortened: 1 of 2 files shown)


niner:
This feels like exposing an implementation detail.

It also opens a can of worms: since this id is used to construct file paths, you're essentially giving the user a way to influence those paths. They can create a subclass of CUDS with an overridden lookup-id method and we blindly assume that we get something like a serialized SHA hash.

lizmat:
@niner Initially I thought it made a lot of sense to make this commit.  Then I realized we would be making a public API, so turned it into a PR.  You've just convinced me that it would be a bad idea.  So closing this PR.

@ugexe This was one case I wasn't sure.  So I made a PR.

ugexe:
> @ugexe This was one case I wasn't sure. So I made a PR.

Right, its always good to make a PR even if you think you are sure its fine.

PR #4869, opened 2022-04-14
Streamline Distribution::Resource

lizmat:
- this is mostly about maintainability
- introduce private IO attribute set in TWEAK
- use "handles" trait on it to auto-generate many accessors
- use direct attribute access
- Str/gist/raku still have manual code on account of them being multis
  and their behaviour is different when called on type object

### DIFF
--- a/src/core.c/Distribution/Resource.pm6
+++ b/src/core.c/Distribution/Resource.pm6
@@ -4,81 +4,31 @@ class Distribution::Resource {
     has $.dist-id;
     has $.key;
 
-    method IO() {
+    # delegate appropriate IO::Path methods to the resource IO::Path object
+    has IO::Path $.IO is built(False) handles <
+      absolute is-absolute relative is-relative parts volume dirname
+      basename extension open resolve slurp lines comb split words copy
+    >;
+
+    method TWEAK() {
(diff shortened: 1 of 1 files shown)


src/core.c/Distribution/Resource.pm6
```diff
@@ -4,81 +4,31 @@ class Distribution::Resource {
     has $.dist-id;
     has $.key;
 
-    method IO() {
+    # delegate appropriate IO::Path methods to the resource IO::Path object
+    has IO::Path $.IO is built(False) handles <
```
niner, line 8: This is breaking installation via Staging repository. IO() **must** be determined at runtime. The result **must not** make it into a precomp file.
Cf. commit 67906e48dca69ff9af9dbb4d01dd1ff3a1e759b9 and d4d6a9976403c7a8339aaf44d3fb9e9656c28232

niner (changes requested):
This PR should be closed.

src/core.c/Distribution/Resource.pm6
lizmat: FWIW, then we must add some tests somewhere, because this was spectest clean.

PR #4870, opened 2022-04-14
Streamline Distribution::Resources

lizmat:
- mark all attributes as not being built, as BUILD handles them
- leave comment as to why "from-precomp" exists
- tighten up signatures
- made AT-KEY/Str a multi to handle calls with type objects properly
- add ASSIGN/BIND/DELETE-KEY methods with appropriate error messages

### DIFF
--- a/src/core.c/Distribution/Resources.pm6
+++ b/src/core.c/Distribution/Resources.pm6
@@ -1,9 +1,9 @@
 role CompUnit::Repository { ... }
 
 class Distribution::Resources does Associative {
-    has Str $.dist-id;
-    has Str $.repo;
-    has Str $.repo-name;
+    has Str $.dist-id   is built(False);
+    has Str $.repo      is built(False);
+    has Str $.repo-name is built(False);
 
     proto method BUILD(|) {*}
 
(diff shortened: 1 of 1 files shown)


src/core.c/Distribution/Resources.pm6
```diff
@@ -13,26 +13,41 @@ class Distribution::Resources does Associative {
...
     multi method BUILD(:$!dist-id, :$repo, Str :$!repo-name --> Nil) { }
     multi method BUILD(:$!dist-id, Str :$!repo, :$repo-name --> Nil) { }
 
-    method from-precomp() {
-        if %*ENV<RAKUDO_PRECOMP_DIST> -> \dist {
-            my %data := Rakudo::Internals::JSON.from-json: dist;
-            self.new(:repo(%data<repo>), :repo-name(%data<repo-name>), :dist-id(%data<dist-id>))
+    # Alternate instantiator called from Actions.nqp during compilation
+    # of %?RESOURCES
+    method from-precomp(Distribution::Resources:U:) {
```
ugexe, line 21: I wonder if this method, which predates `is-implementation-detail`, should be marked as such

ugexe (approved):

src/core.c/Distribution/Resources.pm6
lizmat: So done with https://github.com/rakudo/rakudo/commit/b24c5d97dd

PR #4872, opened 2022-04-15
Streamline CompUnit::Repository::Distribution

lizmat:
- remove BUILD, binding of attributes handled by "is bind"
- bind the meta to a % instead, set defaults in TWEAK
- tighten up signatures
- remove unneccessary returns
- mark method as implementation-detail and mention why
- don't bother calling .new if you can directly call .bless
- remove scopes from various stringifications

### DIFF
--- a/src/core.c/CompUnit/Repository/Distribution.pm6
+++ b/src/core.c/CompUnit/Repository/Distribution.pm6
@@ -53,7 +53,7 @@ class CompUnit::Repository::Distribution does Distribution {
         }
     }
 
-    method serialize(--> Str:D) {
+    method serialize(--> Str:D) is implementation-detail {
         Rakudo::Internals::JSON.to-json: {:$.repo, :$.repo-name, :$.dist-id}
     }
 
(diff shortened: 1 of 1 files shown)


src/core.c/CompUnit/Repository/Distribution.pm6
```diff
@@ -1,61 +1,67 @@
...
+              :repo-name($name),
+              :dist-id($id);
         }
         else {
             Nil
         }
     }
 
-    method serialize() {
+    method serialize(--> Str:D) {
```
ugexe, line 56: This is probably an implementation detail as well, as its intended usage is as a counterpart of `.from-precomp`

ugexe (approved):

src/core.c/CompUnit/Repository/Distribution.pm6
lizmat: Done in https://github.com/rakudo/rakudo/commit/fdc9608a51

PR #4873, opened 2022-04-15
Streamline CompUnit::Repository::Spec and its usages

lizmat:
- remove "required" check for an essentially implementation-detail class
- bind attributes rather than assigning
- remove default for "path", needs to be specified always
- tighten up signatures
- mark "from-string" as implementation detail
- make optional named arg default-short-id a required positional
- "from-string" now supposed to be called with a non-empty string
- use IterationBuffer for building specs list

### DIFF
--- a/src/core.c/CompUnit/RepositoryRegistry.pm6
+++ b/src/core.c/CompUnit/RepositoryRegistry.pm6
@@ -21,10 +21,21 @@ class CompUnit::RepositoryRegistry {
     my $custom-lib := nqp::hash();
 
     proto method repository-for-spec(|) { * }
-    multi method repository-for-spec(Str $spec, CompUnit::Repository :$next-repo) {
-        self.repository-for-spec(CompUnit::Repository::Spec.from-string($spec), :$next-repo)
+    multi method repository-for-spec(
+      Str:D                 $spec,
+      CompUnit::Repository :$next-repo
+    --> CompUnit::Repository:D) {
+        $spec
+          ?? self.repository-for-spec(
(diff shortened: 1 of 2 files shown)


src/core.c/CompUnit/RepositoryRegistry.pm6
```diff
@@ -21,10 +21,21 @@ class CompUnit::RepositoryRegistry {
     my $custom-lib := nqp::hash();
 
     proto method repository-for-spec(|) { * }
-    multi method repository-for-spec(Str $spec, CompUnit::Repository :$next-repo) {
-        self.repository-for-spec(CompUnit::Repository::Spec.from-string($spec), :$next-repo)
+    multi method repository-for-spec(
+      Str:D                 $spec,
+      CompUnit::Repository :$next-repo
+    --> CompUnit::Repository:D) {
+        $spec
```
ugexe, line 28: I'm not sure if I can grok the reasoning behind this conditional. `$spec` is now `Str:D` (and thus should pass the `str` constraint later on the call to `.from-string`). And it should be pretty rare for `$spec` to be an empty string (so not worth optimizing at the cost of the common case).

lizmat: But that was just the point: `CompUnit::Repository::Spec.from-string` contained a check for empty string.  And the only other place it is being called from (in `parse-include-specS`) already had a check for empty string.  So in that case, it was doing the check twice.  So I removed the check from `from-string` and added it to `repository-for-spec`, so that the check for empty string was only done once in both cases.

ugexe (approved):

PR #4875, opened 2022-04-15
Streamline CompUnit::Repository::Perl5

lizmat:
- tighten signatures
- fewer repeated method calls
- tighten signatures
- move CATCH block to start of applicable scope
- only create Perl 5 dependency specifications once in a process

### DIFF
--- a/src/core.c/CompUnit/Repository/Perl5.pm6
+++ b/src/core.c/CompUnit/Repository/Perl5.pm6
@@ -1,51 +1,56 @@
 class CompUnit::Repository::Perl5 does CompUnit::Repository {
-    method need(
-        CompUnit::DependencySpecification $spec,
-        CompUnit::PrecompilationRepository $precomp = self.precomp-repository(),
-        --> CompUnit:D)
-    {
+    my $perl5-dependency :=
+      CompUnit::DependencySpecification.new(:short-name<Inline::Perl5>);
+
+    method need(CompUnit::Repository::Perl5:D:
+      CompUnit::DependencySpecification $spec,
+      CompUnit::PrecompilationRepository $precomp,
+    --> CompUnit:D) {
         if $spec.from eq 'Perl5' {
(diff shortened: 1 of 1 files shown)


src/core.c/CompUnit/Repository/Perl5.pm6
```diff
@@ -1,51 +1,56 @@
...
-            );
-            return CompUnit.new(
-                :short-name($spec.short-name),
-                :$handle,
-                :repo(self),
-                :repo-id($spec.short-name),
-                :from($spec.from),
+              $short-name,
+              $spec.version-matcher !== True
+                ?? $spec.version-matcher.Numi
```
ugexe, line 29: `.Numi`? Are we sure this code runs and works?

ugexe (changes requested):

src/core.c/CompUnit/Repository/Perl5.pm6
lizmat: Oops.  Fixed with https://github.com/rakudo/rakudo/commit/be98708fd5  Note that the condition was wonky, so I fixed that as well.

niner (changes requested):

src/core.c/CompUnit/Repository/Perl5.pm6
```diff
@@ -1,51 +1,56 @@
 class CompUnit::Repository::Perl5 does CompUnit::Repository {
-    method need(
-        CompUnit::DependencySpecification $spec,
-        CompUnit::PrecompilationRepository $precomp = self.precomp-repository(),
-        --> CompUnit:D)
-    {
+    my $perl5-dependency :=
+      CompUnit::DependencySpecification.new(:short-name<Inline::Perl5>);
```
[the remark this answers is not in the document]
lizmat: Good point.  I had thought about that but somehow thought it could be a bad idea (thinking about the IO issue on a resource recently).  Fixed with https://github.com/rakudo/rakudo/commit/29ba6aa3db

ugexe (approved):

PR #4876, opened 2022-04-16
Add CompUnit.api attribute and add usage where applicable

lizmat:
It appears to have been missed before.

### DIFF
--- a/src/core.c/CompUnit/Repository/Installation.pm6
+++ b/src/core.c/CompUnit/Repository/Installation.pm6
@@ -547,16 +547,17 @@ sub MAIN(:$name, :$auth, :$ver, *@, *%) {
     }
 
     method resolve(CompUnit::DependencySpecification:D $spec --> CompUnit:D) {
-        with self!matching-dist($spec) {
-            my %meta := .meta;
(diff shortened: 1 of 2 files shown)


src/core.c/CompUnit/Repository/Installation.pm6
```diff
@@ -547,16 +547,17 @@ sub MAIN(:$name, :$auth, :$ver, *@, *%) {
...
     method resolve(CompUnit::DependencySpecification:D $spec --> CompUnit:D) {
-        with self!matching-dist($spec) {
-            my %meta := .meta;
+        if self!matching-dist($spec) -> $distribution {
+            my %meta := $distribution.meta;
             CompUnit.new(
               :handle(CompUnit::Handle),
               :short-name($spec.short-name),
               :version(%meta<ver>),
-              :auth(%meta<auth> // Str),
```
ugexe, line 556: I can't tell if this change was intentional or not, but for anyone else following along this change is technically ok because `%meta<auth>` will default to an empty string via `CompUnit::Repository::Distribution`

ugexe (approved):

PR #4877, opened 2022-04-16
Streamline CompUnit::Repository::NQP

lizmat:
- make the %opts hash for loading a constant
- fewer repeated method calls
- don't set up precomp repository unless it's needed
- tighten signatures

### DIFF
--- a/src/core.c/CompUnit/Repository/NQP.pm6
+++ b/src/core.c/CompUnit/Repository/NQP.pm6
@@ -1,38 +1,36 @@
 class CompUnit::Repository::NQP does CompUnit::Repository {
-    method need(
-        CompUnit::DependencySpecification $spec,
-        CompUnit::PrecompilationRepository $precomp = self.precomp-repository(),
-        --> CompUnit:D)
-    {
+    my constant %opts = :from<NQP>,;  # comma needed to make a Map
+
+    method need(CompUnit::Repository::NQP:D:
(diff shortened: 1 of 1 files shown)


ugexe (approved):

src/core.c/CompUnit/Repository/NQP.pm6
```diff
@@ -1,38 +1,36 @@
...
+              :short-name($key),
+              :handle(CompUnit::Handle.new(
+                 nqp::gethllsym('Raku','ModuleLoader').load_module($key, %opts)
+              )),
+              :repo(self),
+              :repo-id($key),
+              :from<NQP>;
+        }
+        elsif self.next-repo -> $repo {
+            $repo.need($spec, self.precomp-repository)
```
ugexe, line 20: This doesn’t use $precomp like it should

ugexe (changes requested):

src/core.c/CompUnit/Repository/NQP.pm6
lizmat: Good catch.  The idea was to not have it need to set up the precomp repos if not needed.  But *if* they're passed, they should be passed  :-)   Fixed with https://github.com/rakudo/rakudo/commit/8a1661bc2f

ugexe (approved):

PR #4881, opened 2022-04-17
Don't instantiate CompUnit::PrecompilationRepository::None

lizmat:
The role does not contain any attributes, so the punned class does
not need to be instantiated.


niner:
I think, I'm against this change. Ostensibly you're entirely correct in that there's no need to create an object when there are no attributes. But if CompUnit::Repository::None is just CompUnit::Repository, then it's basically impossible to distinguish between "I did not get a repository, i.e. it's undefined and I should fall back to some default" and "I got a repository that's just not doing anything, but that is intentional"

For performance there's no difference as this punning is done only once at setting compilation time.

PR #4884, opened 2022-04-18
Micro optimize "self" lookup for attribute binding

lizmat:
Assuming in most cases, the search is successful, it makes more sense
to just do the nqp::atkey and if that fails, then do the error handling.
Instead of first looking up whether it exists at all, and then do another
lookup if the first check is successful.

### DIFF
--- a/src/Perl6/bootstrap.c/BOOTSTRAP.nqp
+++ b/src/Perl6/bootstrap.c/BOOTSTRAP.nqp
@@ -592,10 +592,8 @@ my class Binder {
         # If it's attributive, now we assign it.
         if $is_attributive {
             # Find self.
-            my $self;
-            if nqp::existskey($lexpad, 'self') {
-                $self := nqp::atkey($lexpad, 'self');
-            } else {
+            my $self := nqp::atkey($lexpad, 'self');
+            if nqp::isnull($self) {
                 if nqp::defined($error) {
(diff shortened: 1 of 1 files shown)


src/Perl6/bootstrap.c/BOOTSTRAP.nqp
```diff
@@ -592,10 +592,8 @@ my class Binder {
         # If it's attributive, now we assign it.
         if $is_attributive {
             # Find self.
-            my $self;
-            if nqp::existskey($lexpad, 'self') {
-                $self := nqp::atkey($lexpad, 'self');
-            } else {
+            my $self := nqp::atkey($lexpad, 'self');
```
vrurg, line 595: This will not work as intended because lexpads are not normal hashes. `atkey` on them throws `Lexical with name 'foo' does not exist in this frame` if a key is not found.

vrurg (changes requested):

MasterDuke17:
Another potential micro-optimization would be moving the `if $got_native {` block earlier. It doesn't need `$self`, so it could `return` without needing to do the `(exists|at)key`.

lizmat:
The other micro opt is just that.  The chance of that firing is minimal (in my experience), using an attribute in a scope where there is no `self`, is much more common.

So closing this one.

Issue #4871, opened 2022-04-15
Unicode property methods/subs are not thread-safe


vrurg:
The following simple line causing me some headaches in a production code:

```raku
    unless $amount ~~ s/^<curr-sym=:Sc> { $symbol-at-start = True } || <curr-sym=:Sc>$// {
        X::CurrencyString.new(:$amount).throw
    }
```

The code is being ran by cron and way too often it emails me back with something like:

```
Bad amount (currency) string: "\$75.11"
```

Since the string in the error is produced with `.raku` for my own convenience, the regex acutally fails on `$75.11` or alike. I mean, the actual amount my vary and thus is irrelevant.

The code is rather heavily multi-threaded and that's why concurrency is the primary suspect here. Moreover, the last nail in the coffin of my patience was the following error: 

```
MoarVM oops: insert conflict, titlecase_letter is 477017254, 18 != 0
from SETTING::src/core.c/Str.pm6:1686 (/home/fta/.rakubrew/versions/moar-master/install/share/perl6/runtime/CORE.c.setting.moarvm:match)
```

Full backtrace points back to the above regex again.

Unfortunately, I cannot produce a golf. Event plain concurrent invocation of the code which is using the regex doesn't result in any errors.

niner:
I think inadvertent sharing of the $/ variable between threads is often an 
issue with regexes.

vrurg:
@niner everything is lexical. Unless the sharing happens somewhere at the backstage, but unlikely.

patrickbkr:
This might me related to https://github.com/rakudo/rakudo/issues/4236
I'm pretty sure that's caused by some thread-safety issues in regexes.

vrurg:
This could have something to do with unicode properties, perhaps. While I was trying to find a workaround the following variant was tested and failed too:

```raku
if ! $amount.contains(/^ <:Sc>/) {
    ...
}
else {
    ... # throw
}
```

It did fail on a string starting with _'$'_. 

At the moment I have simplified my code to just `unimatch($amount, "Sc")` and this variant is passing test runs so far.

vrurg:
I have changed the subject because it looks like regxes are not guilty. Changing my code to the following didn't fix the problem:

```raku
    if unimatch($amount, "Sc") {
        ...
    }
    elsif unimatch($amount.substr(* - 1), "Sc") {
        ...
    }
    else {
        FTA::X::CurrencyString.new(:$amount).throw
    }

```

[...]

PR #4886, opened 2022-04-20
Make uniprop/uniprops thread safe

vrurg:
They rely on a local caching hash which could on occasion receive new keys while still being read from. Now access to the hash is lock-protected.

Alongside hash declaraion is changed from `constant` to a bound local since this what it is.


vrurg:
There was two options to get it fixed: either `$lock.protect` or clone+CAS. I tried both and it looks like the lock option is slightly faster.

vrurg:
Could be the cause of #4871.

PR #4891, opened 2022-04-20
Slightly speedup exponentiation by Ints

MasterDuke17:
Only perform the `$b > 0` check once and check for underflow before
creating the Rat for negative powers.

Also, for `$foo ** 2`, re-write into `$foo * $foo`.

(All with `MVM_SPESH_BLOCKING=1`):
`my $a; loop (my Int $i = 0; $i <= 10_000_000_000; $i += 1000) { $a := $i ** 2 }; say now - INIT now` drops from ~4.52s to ~3.15s.
`my $a; loop (my Int $i = 0; $i <= 10_000_000_000; $i += 1000) { $a := $i ** 3 }; say now - INIT now` drops from ~4.76s to ~4.55s.
`my $a; loop (my Int $i = 0; $i <= 1_000_000_000; $i += 1000) { $a := $i ** -2 }; say now - INIT now` drops from ~1.45s to ~1.41s.

### DIFF
--- a/src/Perl6/Optimizer.nqp
+++ b/src/Perl6/Optimizer.nqp
@@ -3335,6 +3335,16 @@ class Perl6::Optimizer {
             && self.optimize-post-pre-inc-dec-ops($op) -> $qast {
                 return $qast;
             }
+
+            # Re-write `$foo ** 2` into `$foo * $foo`. Literals will be constant folded and
+            # we can't do it for non-VARs, since side effects could be duplicated (e.g.,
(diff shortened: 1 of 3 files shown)


MasterDuke17:
This is the Rakudo-only changes from https://github.com/rakudo/rakudo/pull/4320 plus a new static optimization. I'm still contemplating doing the MoarVM change to `nqp::pow_I`, but if so it'll be a separate PR.

lizmat (approved):

MasterDuke17:
Now supports `²` also.

PR #4892, opened 2022-04-20
Re-introduce $?COMPILATION-ID

lizmat:
- but now properly, as the name of the directory in "precomp" where
  this version of Rakudo will look for precompiled files.
- use it in CompUnit::PrecompilationRepository
- since this is compile-time information, make it constants where possible
- add tools/cleanup-precomps.raku script for removing all precomp
  directories that are unreachable for this version of Rakudo


lizmat:
Closing since this could be considered an implementation detail anyway, and much can be gained without actually producing a constant, as `Compiler.id` at compile time will essentially be the same.

PR #4893, opened 2022-04-20
Make compiler config related info more constant

lizmat:
Removed script.


tools/cleanup-precomps.raku
```diff
@@ -0,0 +1,34 @@
...
+    }
+
+    $io.rmdir unless $keep;
+    $bytes
+}
+
+my int $repositories;
+my int $total;
+my @repos := $*REPO.repo-chain;
+for $*REPO.repo-chain.map({ $_ with .?prefix }) {
```
ugexe, line 28: I think this script is heavily coupled to one particular workflow in a way that can lead to confusion. For instance: sould I, someone with multiple raku installations, expect such a script to delete the precomp folders for all but one compiler in `~/.raku`? Or to try to do anything for non-`CURI` repositories that happen to also use a `prefix`?

lizmat: It may lead to confusion in such cases.

Perhaps I should just put this script into module space.

lizmat: Removed with https://github.com/rakudo/rakudo/commit/257e06ab5f

ugexe (approved):

PR #4894, opened 2022-04-20
Introduce (Exception|Cool).Failure coercer

lizmat:
- marked as implementation detail
- used wherever this made sense

The main reason for this, is to reduce the bytecode of scopes where
a Failure needs to be returned.  This reduces the bytecode by 12 bytes
for every change from "Failure.new(foo)" to "foo.Failure", making
these scopes more likely to get inlined.

For reference:
````raku
sub foo() { fail "foo" }              # 86 bytes when inlined
sub foo() { Failure.new("foo") }      # 70 bytes when inlined
sub foo() { "foo".Failure }           # 58 bytes when inlined


jnthn:
fwiw, the `Exception.Failure` method feels like a sensible thing to have in the API rather than as an implementation detail. Not thrilled about the `Cool` one (but I guess in time we should be dying/failing with typed exceptions everywhere in CORE, which is the underlying thing not to be thrilled about :-)).

lizmat:
I see I erred on the side of caution then wrt `is implementation-detail`.  :-)

I will remove the `implementation-detail`s.

PR #4896, opened 2022-04-21
Restore .meta behaviour

lizmat:
on CompUnit::Repository::Distribution and Distribution::Path.

These were recently changed to return an unitemized hash, but
apparently there's code out there that depended on the itemization
behaviour.


lizmat:
See https://github.com/rakudo/rakudo/commit/27f10a6245e4b19628480b3266966108f77dd855#commitcomment-71752730

ugexe (approved):

PR #4897, opened 2022-04-21
Fix repo-chain changes check

lizmat:
It was incorrectly checking the compilation unit ID against the
what is considered the original first REPO.id in the chain.  It
should be checking against the *current* first REPO.id!

### DIFF
--- a/src/core.c/CompUnit/PrecompilationRepository.pm6
+++ b/src/core.c/CompUnit/PrecompilationRepository.pm6
@@ -159,7 +159,7 @@ Need to re-check dependencies.")
 
         if $REPO-id ne $first-repo-id {
             $!RMD("Repo chain changed:
-  $repo-id
+  $REPO-id
   $first-repo-id
 Need to re-check dependencies.")
               if $!RMD;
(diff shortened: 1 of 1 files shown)


lizmat:
This does not actually fix the currently unnecessary dependency checks on `use Test`, but it takes away one of the reasons for that check.

src/core.c/CompUnit/PrecompilationRepository.pm6
```diff
@@ -157,9 +157,9 @@ Need to re-check dependencies.")
             $resolve := True;
         }
 
-        if $unit-id ne $first-repo-id {
+        if $REPO-id ne $first-repo-id {
```
ugexe, line 160: I'm confused, because if we look back at the code before the rechecking started occurring we see it using `$unit-id` (`$repo-id` in the code linked)

https://github.com/rakudo/rakudo/blob/f5ce80e1a6d60a6a51a898da85f3100bb088dfb4/src/core.c/CompUnit/PrecompilationRepository.pm6#L112-L120

ugexe: Oops nevermind, I read `self!load-repo-id` as `self!load-file`

ugexe (approved):

Issue #4900, opened 2022-04-22
Something weird about precompilation, e.g. of modules in "core" repository


lizmat:
There is either something very weird / wrong with loading modules from the "core" repository (such as `Test` and `NativeCall`), which results in `use NativeCall` taking about **270** msecs, as opposed to about **200** msecs it *could* be.

I *think* the issue shows itself with **any** module installed using `CompUnit::Repository::Staging` (aka `CURS`, which e.g. `zef` does **not** use).

The problem shows itself when you are using a module with `RAKUDO_MODULE_DEBUG`.  Whenever one of the modules from the "core" repository is loaded, such as:
````
RAKUDO_MODULE_DEBUG=1 raku -e 'use Test'
````
it will show:
````
 1 RMD: Repo changed:
          84B7E660C038B7618CAE179A45FAFA186E8C271E
          E3F82C70D9120AC7457E5ABBA16F6B2550393E16
        Need to re-check dependencies.
````
So, somehow it thinks that the "core" repo has changed.  Now, checking dependencies of `Test` is a no-op, as it does not have any dependencies.  But for `NativeCall`, they're quite a few to check, recursively.  And *that* causes the extra **70** msecs delay.  And after that check, it will load the precomp files found in `install/share/perl6/core/precomp`.

Oddly enough, there is a very easy way of making the loading of `NativeCall` faster: empty out the `install/share/perl6/core/precomp` directory.  The first time you do a `use NativeCall` after that, the precomp files will be stored in `~/.raku/precomp`, and loading of `NativeCall` after that, will be fast as no dependencies need to be checked.

But it gets weirder: running `make install` in the rakudo checkout, will re-populate the precomp files in `install/share/perl6/core/precomp`.  But loading `NativeCall` still remains fast because it  prefers the precomp files located in `~/.raku/precomp`.

So my conclusion is, that precomp files installed with CURS are somehow installed with a wrong ID, which causes the unneeded dependency checking.  FWIW, this **also** happens with pre-packaged modules, e.g. the ones made by *Claudio Ramirez*.  So this is not just slowing down core devs, this is also slowing down "in production" so to speak.

@niner @ugexe is this idea of CURS installing with a wrong ID the correct hypothesis, or am I missing something essential in that respect?  If it is the correct hypothesis, I will dig deeper in trying to find out how that wrong ID is calculated.  If not, then please tell me as well!

lizmat:
Full output:
````
[...]

PR #4901, opened 2022-04-22
Fix unneeded dependency check

lizmat:
Commit 199888abedfe843996 in March 2020 borked the setting of the
repo-id.  This caused unneeded dependency checking for installed
modules, and a slowdown of e.g. loading NativeCall of about 80 msecs.

Now, the slowdown will only occur the first time after installation,
when the repo-id is now updated correctly so that a subsequent loading
will not have to do dependency checks.

Spotted by nine++ after looking into https://github.com/rakudo/rakudo/issues/4900

### DIFF
--- a/src/core.c/CompUnit/PrecompilationRepository.pm6
+++ b/src/core.c/CompUnit/PrecompilationRepository.pm6
@@ -247,7 +247,7 @@ Need to re-check dependencies.")
                 self.store.store-repo-id(
                   $compiler-id,
                   $precomp-unit.id,
-                  :repo-id($unit-id)
+                  :repo-id($REPO-id)
                 );
                 self.store.unlock;
             }


niner (approved):

ugexe (approved):

lizmat:
So this is nice.  But I can't help but wonder what would happen if we have an installation that does not have any writeable repos at all?  Or is having at least one writeable repo a prerequisite for Raku(do)?

niner:
No, writable repos cannot be a prerequisite. However, writable repos (or at least one such) can be a prerequisite for some performance optimizations (just like on disk caches in other applications). Having said that, I'm sure there is a way to have this particular optimization even without writable repositories.

lizmat:
> Having said that, I'm sure there is a way to have this particular optimization even without writable repositories.

Ok, I think I know why, just to make sure: why is it not possible to already write the correct repo-id at the moment of installation?

niner:
> Ok, I think I know why, just to make sure: why is it not possible to already
> write the correct repo-id at the moment of installation?

Well, actually we do, but that doesn't do us much good in case of the CORE 
dist.

When installing a dist, we use a shortened repo chain, i.e. only the part 
starting with the repo we install to. That's simply because a module must not 
have a dependency futher up in the chain as that part may change (think the 
root user's ~/.raku vs. nine's ~/.raku).

To avoid the dependency check, we must be sure that the whole repo chain is 
exactly the same as at the time we last did that check. Staying with the CORE 
dist example, when loading NativeCall, it's possible to find an updated 
NativeCall::Types in ~/.raku or even in some repo included via -I. Since 
loading the module happens with a longer repo chain than installation, we'll 
always have to re-check dependencies on the first load. And then we are going 
to need to store that result somewhere.

If we're unable to store, we are going to have to do those checks on the next 
run as well.

lizmat:
@niner Ok, but then I think we need to make sure that the not being able to store, should not be lethal.  It is at the moment: https://github.com/rakudo/rakudo/blob/master/src/core.c/CompUnit/PrecompilationStore/File.pm6#L216   Do you agree?

niner:
It already shouldn't be lethal because we shouldn't even try to store that repo-id.

We only ever call store-repo-id if we get a destination to write to from the store:
https://github.com/rakudo/rakudo/blob/master/src/core.c/CompUnit/PrecompilationRepository.pm6#L242
and the store only gives you a writable destination:
https://github.com/rakudo/rakudo/blob/master/src/core.c/CompUnit/PrecompilationStore/File.pm6#L155

PR #4903, opened 2022-04-23
Reinstate CATCH versus LEAVE

lizmat:
Alhough the code with LEAVE is much simpler and clearer, it *does*
impose the overhead of having to execute a phaser on the most common
execution path.  And since both a CATCH block or a LEAVE phaser will
inhibit inlining of a block, from a performance point of view a
CATCH is better, as that will only be executed when something actually
goes wrong (which hopefully is not the most common execution path).

### DIFF
--- a/src/core.c/CompUnit/Loader.pm6
+++ b/src/core.c/CompUnit/Loader.pm6
@@ -10,12 +10,18 @@ class CompUnit::Loader is repr('Uninstantiable') {
 
     # Decode the specified byte buffer as source code, and compile it
     method load-source(Blob:D $bytes --> CompUnit::Handle:D) {
-
         my $original-GLOBAL := nqp::ifnull(nqp::gethllsym('Raku','GLOBAL'),Mu);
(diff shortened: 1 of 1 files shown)


ugexe (approved):

TimToady:
I don't see how you can combine them anyway.  CATCH is supposed to run
before the stack is unwound, while LEAVE is supposed to run during unwind.
You can't combine phasers that run at different times without time travel.

Larry

lizmat:
Hi @TimToady, long time no see!  :-)

> I don't see how you can combine them anyway

This is *not* about combining them.  This is about using a simple `LEAVE` phaser resetting state.  Or always reset state in the normal flow, and using a `CATCH` phaser to reset the state should something have gone wrong.  The latter is faster as it would not involve any special action to be performed on the unwind (which would be the most common flow).

TimToady:
Okay, I see that.  Just need to make sure you think about resumable
exceptions not resetting state it shouldn't then.

Larry

PR #4908, opened 2022-04-26
Fix exponentiation by large negative Ints and speedup exponentiation by any sized negative Ints

MasterDuke17:
and speedup exponentiation by any sized negative Ints. The fix is to
revert the incorrect underflow check added in 6f6fd1f76f and the speedup
comes from bypassing `&infix:</>` when creating the `1 / $power` Rational
since we have some knowledge about the values. Also switch to use
`.Failure` on the exceptions for smaller bytecode (though `&infix:<**>`
is still too large to inline).

Rakudo builds ok and passes `make m-test m-spectest m-stresstest`. @usev6++ for spotting the problem.

### DIFF
--- a/src/core.c/Int.pm6
+++ b/src/core.c/Int.pm6
@@ -389,7 +389,7 @@ multi sub infix:<**>(Int:D $a, Int:D $b --> Real:D) {
         !! X::Numeric::Overflow.new.Failure
       # when a**b is too big nqp::pow_I returns Inf
       !! nqp::istype(($power := nqp::pow_I($a,nqp::neg_I($b,Int),Num,Int)),Num) ||
-         (($power := CREATE_RATIONAL_FROM_INTS(1, $power, Int, Int)) == 0 && nqp::isne_I($a,0))
+         (nqp::istype(($power := CREATE_RATIONAL_FROM_INTS(1, $power, Int, Int)),Num) && nqp::iseq_n($power,0e0) && nqp::isne_I($a,0))
         ?? X::Numeric::Underflow.new.Failure
         !! $power
 }
--- a/src/core.c/Int.pm6
+++ b/src/core.c/Int.pm6
@@ -382,21 +382,16 @@ multi sub infix:<%%>(int $a, int $b --> Bool:D) {
 }
 
 multi sub infix:<**>(Int:D $a, Int:D $b --> Real:D) {
-    my $power;
-    if nqp::isge_I($b, 0) {
-        $power := nqp::pow_I($a, $b, Num, Int);
-        # when a**b is too big nqp::pow_I returns Inf
-        nqp::istype($power, Int)
-            ?? $power
-            !! X::Numeric::Overflow.new.Failure
-    }
-    else {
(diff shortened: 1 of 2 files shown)


MasterDuke17:
With `MVM_SPESH_BLOCKING=1`, `my $a; loop (my Int $i = 0; $i <= 1_000_000_000; $i += 1000) { $a := $i ** -2 }; say now - INIT now` drops from ~1.4s down to ~0.5s.

src/core.c/Int.pm6
```diff
@@ -381,23 +381,21 @@ multi sub infix:<%%>(int $a, int $b --> Bool:D) {
...
-            !! Failure.new(X::Numeric::Overflow.new)
+            !! X::Numeric::Overflow.new.Failure
     }
     else {
         $power := nqp::pow_I($a, nqp::neg_I($b, Int), Num, Int);
         # when a**b is too big nqp::pow_I returns Inf
-        nqp::istype($power, Num) || (nqp::isge_I($power, UINT64_UPPER) && nqp::isne_I($a, 0))
-            ?? Failure.new(X::Numeric::Underflow.new)
-            !! 1 / $power
+        nqp::istype($power, Num) || (($power := CREATE_RATIONAL_FROM_INTS(1, $power, Int, Int)) == 0 && nqp::isne_I($a, 0))
```
lizmat, line 396: I wonder if doing an `nqp::iseq_I($power.numerator,0)` would be faster still, bypassing `&infix:<==>`

MasterDuke17: I had the same thought, but `$power` can be a `Num` after `CREATE_RATIONAL_FROM_INTS`.

lizmat: Ah, duh...  :-)

lizmat (approved):

src/core.c/Int.pm6
```diff
@@ -381,23 +381,21 @@ multi sub infix:<%%>(int $a, int $b --> Bool:D) {
     nqp::hllbool(nqp::iseq_i(nqp::mod_i($a, $b), 0))
 }
 
-my constant UINT64_UPPER = nqp::pow_I(2, 64, Num, Int);
-
 multi sub infix:<**>(Int:D $a, Int:D $b --> Real:D) {
     my $power;
     if nqp::isge_I($b, 0) {
```
lizmat, line 386: Have you tried to make this a single ternary instead of an `if` / `else`?   It probably doesn't matter for inlining, as it becomes too large after inlining `CREATE_RATIONAL_FROM_INTS`

MasterDuke17: I'm currently experimenting with some of the other Rakudo-only changes I tried in https://github.com/rakudo/rakudo/pull/4320 since that was before new-disp, but so far I haven't seen anything worthwhile.

MasterDuke17:
With the latest change, `my $a; loop (my Int $i = 0; $i <= 1_000_000_000; $i += 1000) { $a := $i ** -2 }; say now - INIT now` is down to ~0.44s.

PR #4914, opened 2022-04-29
Micro-optimize the (Code|Block).clone method in the bootstrap

lizmat:
- one variable less
- no use of return for type objects
- postfix unless instead of fully scoped unless


niner:
Is this actually faster? And if yes, would it still be faster than only eliminating that return?
At least for getting rid of a variable, I'm pretty sure that does not actually change performance characteristics. NQP variables map pretty much 1:1 to registers and we are going to need a register for that intermediary anyway.

lizmat:
Hmmm... this gets weirder:  this PR is test and spectest clean, but segfaults on test-t.

lizmat:
To answer your question: in a profile of running test-t, that method is called about 80K times and goes from 35 msecs -> 30 msecs exclusive.  Note that this method is never optimized in any way shape or form (which is why I started looking at it, as this is the highest red one in a profile of test-t).

PR #4917, opened 2022-05-01
Resolve race conditions when working with symbols and repositories

vrurg:
Followup to Raku/problem-solving#323

### DIFF
--- a/src/core.c/CompUnit/Repository/Installation.pm6
+++ b/src/core.c/CompUnit/Repository/Installation.pm6
@@ -37,7 +37,7 @@ sub MAIN(:$name, :$auth, :$ver, *@, *%) {
 
 
     method TWEAK() {
-        $!lock       := Lock.new;
+        $!lock       := Lock::Soft.new;
         $!loaded     := nqp::hash;
         $!seen       := nqp::hash;
         $!dist-metas := nqp::hash;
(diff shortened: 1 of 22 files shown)


vrurg:
`Stash` still needs a solution. The current approach with lock-protected `ASSIGN-KEY` is incomplete as it doesn't solve `BIND-KEY` competing for the `$!storage` hash. Keeping in mind the fact that `ModuleLoader`'s symbol merging is based upon binding, I conclude that the problem is not solved. As a matter of fact, I've seen it failing already, even though chances of hitting the problem are greatly reduced now.

Just in case, here is the code I use for testing:

```raku
#!/usr/bin/env raku -I.

constant MODS = 20;
constant BATCHES = 100;

BEGIN $*SCHEDULER = ThreadPoolScheduler.new: :max_threads(MODS*BATCHES+32);

#`{ Module Gen Section. To be done once to create MODS modules.
BEGIN use lib my $libdir = $?FILE.IO.parent(1).add("lib");


note "Pre-gen modules...";
my @meta_libs;
for 'L1', 'L2' -> $l {
    my $subdir = $libdir.add("Req/$l");
    unless $subdir.d {
        $subdir.mkdir;
    }
    for ^MODS -> $idx {
        my $modname = $idx.fmt('Mod%04d');
        my $modf = $libdir.add("Req/" ~ $l ~ "/" ~ $modname ~ ".rakumod");
        @meta_libs.push: qq«"Req::{$l}::$modname": "lib/Req/{$l}/$modname.rakumod"»;
        my $fullname = "Req::{$l}::$modname";
        unless $modf.e {
            $modf.spurt: qq:to/MODCODE/;
            unit module $fullname;
            our \$ID = $idx;
            MODCODE
        }
    }
}
"META6.json".IO.spurt: qq:to/META6/;
\{
    "name": "require-test",
    "version": "0.0.5",
    "authors": [
        "Vadim Belman"
    ],
    "auth": "zef:vrurg",
    "provides": \{
{ @meta_libs.join(",\n").indent(8) }
    }
}
META6
exit;
#}

my $starter = Promise.new;
my @ready;
my @workers;

sub resolve-package(Str:D $pkg) is raw {
    #::{$pkg}:exists ?? ::($pkg) !! do require ::($pkg) 
    require ::($pkg);
    ::($pkg)
}

my $fmt-lock = Lock.new;
sub modtest($l, $idx, $ready) {
    @ready.push: $ready;
    @workers.push: start {
        $ready.keep;
        await $starter;
        my $modname = $fmt-lock.protect: { "Req::" ~ $l ~ "::" ~ $idx.fmt('Mod%04d') };
        CATCH {
            note "IDX: ", $idx.WHICH if .message.contains('Directive');
            .rethrow
        }
        require ::($modname);
        my $mod := ::($modname);
        #my $mod := resolve-package($modname);
        #note $modname, " -- ", 
        my $id = $mod.WHO<$ID>;
        die "Bad ID " ~ $id.WHICH unless $id.defined && $id == $idx;
    }
}

note "Starting workers...";
for ^BATCHES {
    for 'L1', 'L2' -> $l {
        for ^MODS -> $idx {
            modtest($l, $idx, Promise.new)
        }
    }
}

await @ready;
$starter.keep;
await @workers;

note "Done.";
```

src/core.c/Stash.pm6
```diff
@@ -50,7 +51,7 @@ my class Stash { # declared in BOOTSTRAP
                             nqp::getattr(self, Hash, '$!descriptor')),
                         assignval)
                 );
-                nqp::bindattr(self, Map, '$!storage', $storage);
+                nqp::atomicbindattr(self, Map, '$!storage', $storage);
```
lizmat, line 54: Wouldn't this fail on the JVM backend ?

vrurg: No idea. I picked the op from BOOTSTRAP where it's used in `atomic_store`. The op is not `#?if`-wrapped.

src/core.c/Stash.pm6
```diff
@@ -62,15 +71,31 @@ my class Stash { # declared in BOOTSTRAP
         nextwith(key.Str, assignval)
     }
 
+    proto method BIND-KEY(|) {*}
```
lizmat, line 74: OOC, why the proto?

src/core.c/Stash.pm6
```diff
@@ -34,6 +42,7 @@ my class Stash { # declared in BOOTSTRAP
         )
     }
 
+    proto method ASSIGN-KEY(Stash:D: $, $) {*}
```
lizmat, line 45: OOC, why the proto?

src/core.c/Stash.pm6
lizmat: Hmmm...  that feels like a premature optimization to me...  I mean, if the performance issue were real, there are many other cases in the core we would need to look at.  And I don't think it was ever an issue really, apart from really cold code.  And even less so since new-disp?  @jnthn @niner

vrurg: Could be. That's why I used 'possibly'. But in either case, the two candidates provided by the class are all it needs. So, `proto` makes sense even from this perspective.

I'd rather ask why does it worry you?

lizmat: Well, it is a bit like (lack of) action at a distance.  Any subclass of `Stash` will *not* have access to the `Hash` candidates, like ever.  I think so far, additional `proto`s in the setting, have been used only to fix potential ambiguities.

lizmat: If you use the `proto`s here, I would add a little comment as to *why* they were added, so that our future self / others might find it easier to decide on their need if the situation changes.

vrurg: `Hash` variants are thread-unsafe. Therefore any `Stash`-descendant has to well justify their use. Either way, `Stash` overrides them completely making `self.Hash::BIND-KEY` the only way to address them.

I'll try to remember about the comment. Makes sense.

vrurg:
This PR is ready. All tests are passing, my test script does pass too. What is to be resolved:

1. First of all, PR relies on [a test patch](https://gist.github.com/niner/f1f523c4b146e29323d68cbe067c4496) provided by @niner. Would it be possible to get the fix done in the mainstream MoarVM? 
2. I introduced a new lock class: `Lock::Soft`. The idea is to have a drop-in replacement for `Lock` which would support recursive invocations and condition variable but would not block the thread when cannot be acquired. So, this is `await`/`Promise` based version of `Lock`. I'm not sure about the name of it though. If there are any suggestions I'd gladly consider them.
3. The new lock class would need a new roast test. Since it's a drop-in, it is passing _S17-lowlevel/lock.t_. If the name is agreed upon I would just modify the test to cover both `Lock` and `Lock::Soft`.

vrurg:
Another notable change is `%REQUIRE_SYMBOLS` has been replaced with `%?REQUIRE-SYMBOLS` and it is now a `Stash`. Correspondingly, it gained the newly available thread safety of `Stash` and made it unnecessary to impose any kind of lock on `require` itself. Consequently, overall performance should be better for `require` called in different contexts.

src/core.c/CompUnit/Repository/Installation.pm6
```diff
@@ -227,145 +227,150 @@ sub MAIN(:$name, :$auth, :$ver, *@, *%) {
...
+            $lock.lock;
+
+            my $version = self!repository-version;
+            self.upgrade-repository unless $version == 2;
+
+            my $dist-id = $dist.id;
+            my $dist-dir = self!dist-dir;
+            if not $force and $dist-dir.add($dist-id) ~~ :e {
+                $lock.unlock;
+                $already-installed = True;
```
ugexe, line 244: Is there not a way to keep the original indentation / return style / failure point? I'm guessing its required for the new lock mechanism to work properly, but there are multiple hundred lines of code contained in this lock that anyone else looking here needs to know to e.g. not return early. I find it kind of hard to follow compared to the previous pattern.

vrurg: Unfortunately, not until `LEAVE` handles early `return` correctly. I was very much hesitant about changing this piece and spent hours experimenting with workarounds, but there is none. So, no `return` or `fail` within the protected blocks so far.

ugexe: But what about indirect failures from e.g. some file operation failing?

vrurg: They wouldn't be affected because they'd be handled by the routines where they happen. The problem with `LEAVE` is about blocks only. I guess, it's a kind of optimization for a lexically scoped block to avoid passing through all the stages of exception throwing/catching, which kicks in. As soon as `fail` results in a `Failure` returned it's not a problem anymore.

melezhik:
@vrurg  how can we test that PR? I assume the code above should succeed? ( should not fail ). Is it correct?

vrurg:
@melezhik you can use the script, but it would require a little bit of handwork to pre-create the test modules. Consider the commented out block.

melezhik:
@vrurg , I see, thanks for information. I have added blackbox test for this PR. On older versions or Rakudo it fails with different errors or just hangs. I think should be good enough:

https://gist.github.com/melezhik/12101ab13ea0b5732c461dc5a441769e

melezhik:
> Especially if you start with afresh environment each time, which is likely the case.

yes it does

melezhik:
here is improved output of failing test - http://sparrowhub.io:2222/report/608

melezhik:
so yeah, looks like false negative, sorry

PR #4918, opened 2022-05-03
Cleanup scope exit-handling

lizmat:
An optimization to handle a single LEAVEish phaser block, was broken.
This meant that all block exits that had any phaser, would take the
slow path.

Instead if a block contains *only* a single LEAVE phaser, a super fast
path is taken on exiting the block, without needing any hash lookups.
This e.g. reduces the overhead of Lock.protect by 20%.  And there are
4 of these for *every* startup of rakudo.

This commit changes / fixes the following:
- Block.$!phasers now either is not concrete (no phasers), contains
  a block (a single LEAVE phaser), or contains a hash with phasers (as
  before)
- the Block API for phasers remains the same, handling the new meanings
  of the $!phasers attribute transparently
- The add phaser handling code has been adapted to support the lone
  LEAVE phaser case
- the exit-handler only gets called if $!phasers is concrete: it
  now only has to check whether it is a hash (if not, take the fast
  path, directly executing what is in $!phasers).
- the slow exit-handler path also has a fast path for scopes that
  have leavish phasers, but no KEEP / UNDO phasers
- unnecessary deconting has been remove from the exit handler.
- the PhasersList class has been removed: it predated the IterationBuffer
  class, and it was basically just that.  Changing that, simplified
  HLLizing some Block phaser API calls
- the brittle handling of scopes by the "will" trait on variables, has
  been changed to dynamically search for the variable in question

### DIFF
--- a/src/Perl6/Actions.nqp
+++ b/src/Perl6/Actions.nqp
@@ -4418,7 +4418,7 @@ class Perl6::Actions is HLL::Actions does STDActions {
         # them. No parameters also means no inlining.
         return 0 unless @params;
 
-        my $world   := $*W;
+        my $world  := $*W;
         my $Param  := $world.find_single_symbol_in_setting('Parameter');
         my @p_objs := nqp::getattr($sig, $world.find_single_symbol_in_setting('Signature'), '@!params');
         my %arg_placeholders;
--- a/src/Perl6/Actions.nqp
+++ b/src/Perl6/Actions.nqp
@@ -2170,17 +2170,14 @@ class Perl6::Actions is HLL::Actions does STDActions {
         if $*LABEL {
             $loop.push(QAST::WVal.new( :value($world.find_single_symbol($*LABEL)), :named('label') ));
         }
-        # Handle phasers.
-        my $code := $loop[1].ann('code_object');
-        my $block_type := $world.find_single_symbol_in_setting('Block');
-        my $phasers := nqp::getattr($code, $block_type, '$!phasers');
-        if !nqp::ishash($phasers) {
-            $loop[1] := pblock_immediate($loop[1]);
-        }
-        else {
+        # Handle any loopy phasers.
+        my $code    := $loop[1].ann('code_object');
+        my $Block   := $world.find_single_symbol_in_setting('Block');
+        my $phasers := nqp::getattr($code, $Block, '$!phasers');
+        if nqp::ishash($phasers) {
             my $node := $loop.node;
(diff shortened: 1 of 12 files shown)


src/Perl6/Actions.nqp
```diff
@@ -4416,7 +4418,7 @@ class Perl6::Actions is HLL::Actions does STDActions {
         # them. No parameters also means no inlining.
         return 0 unless @params;
 
-        my $world := $*W;
+        my $world   := $*W;
```
MasterDuke17, line 4421: Looks like one too many spaces added.

src/Perl6/bootstrap.c/BOOTSTRAP.nqp
```diff
@@ -4055,93 +4054,114 @@ nqp::sethllconfig('Raku', nqp::hash(
...
-            my @leaves := nqp::atkey(%phasers, '!LEAVE-ORDER');
-            my @posts  := nqp::atkey(%phasers, 'POST');
-            my @exceptions;
-            unless nqp::isnull(@leaves) {
-
-                # only have a single LEAVEish phaser, so no frills needed
-                if nqp::elems(@leaves) == 1 && nqp::elems(%phasers) == 1 {
+            # when we get here, we assume the $!phasers attribut is concrete.
+            # if it is *not* a hash, it is a lone LEAVE phaser, the most
+            # commenly used phaser (in the core at least).
```
MasterDuke17, line 4059: "commenly" -> "commonly"

src/Perl6/bootstrap.c/BOOTSTRAP.nqp
```diff
@@ -4055,93 +4054,114 @@ nqp::sethllconfig('Raku', nqp::hash(
...
+                    my int $n := nqp::elems(@leaves);
+                    my int $i := -1;
+
+                    # fast leave path
+                    if nqp::isnull(@leaves) && nqp::isnull(@undos) {
+                        while ++$i < $n {
+                            CATCH { nqp::push(@exceptions, $_) }
 #?if jvm
-                    nqp::decont(nqp::atpos(@leaves,0))();
+                            nqp::atpos(@leaves, $i))();
```
usev6, line 4080: There is a superfluous closing paren now (leftover from removing the ```nqp::atpos```). CI is also not happy with that.

lizmat: Thanks for the spot, fixed now!

PR #4923, opened 2022-05-12
Re-imagine DateTime.new(Str) / Str.DateTime

lizmat:
- makes DateTime.new(YYYY-MM-DDTHH:MM:SS) 1.9x as fast
- makes DateTime.new(YYYY-MM-DDTHH:MM:SS+TZ) about 1.3x as fast
- makes DateTime.new(YYYY-MM-DD) possible
- makes DateTime.new(YYYY-MM-DD) 1.15x as fast as Date.new().DateTime
- offloads some error handling to additional private methods to allow
  for better inlining of the fast paths
- uses simple substrings where possible
- avoids using regexes / captures if possible
- uses 3 separate regexes on the slowest path

Inspired by https://www.reddit.com/r/rakulang/comments/unm694/datetime_no_longer_accepts_iso8601_strings/

### DIFF
--- a/src/core.c/DateTime.pm6
+++ b/src/core.c/DateTime.pm6
@@ -308,54 +308,144 @@ my class DateTime does Dateish {
                :$hour,:$minute,:$second,:$timezone,:&formatter,|%_
              )!SET-DAYCOUNT;
     }
-    multi method new(DateTime:
-      Str:D $datetime, :$timezone is copy, :&formatter, *%_
-    --> DateTime:D) {
-        self!tif(
+
+    method !non-iso($datetime) {
+        self!tif:
           $datetime,
           'DateTime',
           'an ISO 8601 timestamp (yyyy-mm-ddThh:mm:ssZ or yyyy-mm-ddThh:mm:ss+01:00)'
-        ) unless $datetime.chars == $datetime.codes and $datetime ~~ /^
-          (<[+-]>? \d**4 \d*)                            # year
-          '-'
-          (\d\d)                                         # month
-          '-'
-          (\d\d)                                         # day
-          <[Tt]>                                         # time separator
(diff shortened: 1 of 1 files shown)


jubilatious1:
Reading the Reddit discussion, all that seems to be required is an error message to use `.Date`  instead of `.DateTime` if the input string has no time component?

And once coerced to `.Date`, following that call with a call to `.DateTime` works just fine.

Not clear there's anything here that needs fixing.

https://www.reddit.com/r/rakulang/comments/unm694/comment/i8937r9/?utm_source=share&utm_medium=web2x&context=3

lizmat:
It seems a bit weird to do checks on YYYY-MM-DD only to tell the user to call `.Date`?

jubilatious1:
> It seems a bit weird to do checks on YYYY-MM-DD only to tell the user to call `.Date`?

I swear I just learned this lesson last week :-)

https://unix.stackexchange.com/a/702103/227738

For the above StackExchange answer, I played with `DateTime` and it didn't work. THANKFULLY. Because I was trying to convert a `mm/dd/yyyy` date into an `ISO 8601` date, and using `DateTime` would have sent be down the "wrong road".

The "correct road" was arrived at because I got an appropriate error message when I tried to use `DateTime`:

```
Invalid DateTime string '1958-06-13'; use an ISO 8601 timestamp (yyyy-mm-ddThh:mm:ssZ or yyyy-mm-ddThh:mm:ss+01:00) instead
  in block <unit> at -e line 1
```

That's a pretty good error message. Maybe it can be improved by mentioning "use `Date` if only trying to convert `yyyy-mm-dd` to `ISO 8601` format". But I don't see how/why checks are necessary, if all you want to do is warn the user about the most common case?

PR #4924, opened 2022-05-15
Fix delegation to pun

### LINKED ISSUE #4916: Inconsistent dispatch to method defined in role
## The Problem

A method may accept a _raw_ invocant, meaning that the invocant parameter is bound to the container if the method is called on a scalar container. Such a method can replace the thing stored inside the container, which can be useful.

A method that accepts a raw invocant can be defined inside a role, and it will work fine if called on a class that `does` the role. However, if called on a pun, the invocant parameter appears to be bound to the thing _inside_ the container, rather than to the container it is stored in, as demonstrated by the following code:

```raku
role Golfy {
    has @.values;
    multi method golf(Golfy:U \SELF: **@values) {
        SELF = SELF.new(:@values);
    }
}

class Golf does Golfy {}

my Golf $var;
try $var.golf(42);
say $! ?? $!.message !! $var.raku;

my Golfy $var2;
try $var2.golf(42);
say $! ?? $!.message !! $var2.raku;
```

## Expected Output

```
Golf.new(values => [42])
Golfy.new(values => [42])
```

## Actual Output

```
Golf.new(values => [42])
Cannot modify an immutable 'Golfy' type object
```

## Environment
* Compiler version (`raku -v`):  
```
Welcome to Rakudo™ v2022.04.
Built on MoarVM version 2022.04.

### PR DESCRIPTION
**Pull request by dumarchie**
If a method is called on a _containerized_ role, the call is now delegated to a `Proxy` that fetches the pun until something is stored in the container.  This resolves #4916.

### DIFF
--- a/src/Perl6/Metamodel/RolePunning.nqp
+++ b/src/Perl6/Metamodel/RolePunning.nqp
@@ -57,6 +57,24 @@ role Perl6::Metamodel::RolePunning {
         self.pun($obj)
     }
 
+    my $proxy_type := nqp::null();
+    method !make_proxy($cont) {
+        if nqp::isnull($proxy_type) {
+            $proxy_type := nqp::gethllsym('Raku', 'Proxy');
+        }
+
+        my $proxy   := nqp::create($proxy_type);
+        my $updated := 0;
+        nqp::bindattr($proxy, $proxy_type, '&!FETCH', -> $var {
+            $updated ?? nqp::decont($cont) !! $!pun;
+        });
+        nqp::bindattr($proxy, $proxy_type, '&!STORE', -> $var, $val {
+            nqp::assign($cont, $val);
+            $updated := 1;
+        });
+        $proxy;
+    }
+
     # Do a pun-based dispatch. If we pun, return a thunk that will delegate.
     method find_method($obj, $name, *%c) {
         if nqp::existskey(%exceptions, $name) {
@@ -72,7 +90,10 @@ role Perl6::Metamodel::RolePunning {
             return nqp::null();
         }
         -> $inv, *@pos, *%named {
-            $meth($!pun, |@pos, |%named);
+            # If called on a container, delegate to a Proxy that fetches the pun
+            # until something is stored in the container
+            my $proxy := nqp::iscont($inv) ?? self.'!make_proxy'($inv) !! $!pun;
+            $meth($proxy, |@pos, |%named);
         }
     }
 
--- a/src/Perl6/Metamodel/RolePunning.nqp
+++ b/src/Perl6/Metamodel/RolePunning.nqp
@@ -85,7 +85,7 @@ role Perl6::Metamodel::RolePunning {
             $!made_pun := 1;
         }
 
-        my $meth := $!pun.HOW.find_method($!pun, $name);
+        my $meth := $!pun.HOW.find_method($!pun, $name, |%c);
         unless nqp::isconcrete($meth) {
             return nqp::null();
         }
--- a/src/Perl6/Metamodel/RolePunning.nqp
+++ b/src/Perl6/Metamodel/RolePunning.nqp
@@ -66,11 +66,13 @@ role Perl6::Metamodel::RolePunning {
             $!pun := self.make_pun($obj);
             $!made_pun := 1;
         }
-        unless nqp::can($!pun, $name) {
+
+        my $meth := $!pun.HOW.find_method($!pun, $name);
+        unless nqp::isconcrete($meth) {
             return nqp::null();
         }
         -> $inv, *@pos, *%named {
-            $!pun."$name"(|@pos, |%named)
+            $meth($!pun, |@pos, |%named);
         }
     }
 
--- a/src/Perl6/bootstrap.c/BOOTSTRAP.nqp
+++ b/src/Perl6/bootstrap.c/BOOTSTRAP.nqp
@@ -4422,7 +4422,10 @@ nqp::bindhllsym('Raku', 'Scalar', Scalar);
 nqp::bindhllsym('Raku', 'ScalarVAR', ScalarVAR);
 nqp::bindhllsym('Raku', 'default_cont_spec',
     Scalar.HOW.cache_get(Scalar, 'default_cont_spec'));
+
+# Stash types that may be instantiated by the Metamodel
 nqp::bindhllsym('Raku', 'Capture', Capture);
+nqp::bindhllsym('Raku', 'Proxy', Proxy);
 
 #?if jvm
 # On JVM, set up JVM interop bits.
--- a/src/vm/moar/dispatchers.nqp
+++ b/src/vm/moar/dispatchers.nqp
@@ -816,22 +816,6 @@ nqp::dispatch('boot-syscall', 'dispatcher-register', 'raku-meth-call', -> $captu
             $capture);
     }
     else {
-        # See if we're making the method lookup on a pun; if so, rewrite the args
-        # to do the call on the pun.
-        if nqp::istype($how, Perl6::Metamodel::RolePunning) &&
-                $how.is_method_call_punned($obj, $name) {
-            nqp::dispatch('boot-syscall', 'dispatcher-guard-type',
-                nqp::dispatch('boot-syscall', 'dispatcher-track-arg', $capture, 0));
-            $obj := $how.pun($obj);
-            $how := $obj.HOW;
-            $capture := nqp::dispatch('boot-syscall', 'dispatcher-insert-arg-literal-obj',
-                nqp::dispatch('boot-syscall', 'dispatcher-drop-arg', $capture, 0),
-                0, $obj);
-            $capture := nqp::dispatch('boot-syscall', 'dispatcher-insert-arg-literal-obj',
-                nqp::dispatch('boot-syscall', 'dispatcher-drop-arg', $capture, 2),
-                2, $obj);
-        }
-
         # Try to resolve the method call.
         # TODO Assorted optimizations are possible here later on to speed up some
         # kinds of dispatch, including:

### CODE REVIEW

File: src/Perl6/Metamodel/RolePunning.nqp [hunk 1]
```diff
@@ -66,11 +66,13 @@ role Perl6::Metamodel::RolePunning {
             $!pun := self.make_pun($obj);
             $!made_pun := 1;
         }
-        unless nqp::can($!pun, $name) {
+
+        my $meth := $!pun.HOW.find_method($!pun, $name);
```
Kaiepi, line 88: This should probably forward any named parameters given.

src/Perl6/Metamodel/RolePunning.nqp
```diff
@@ -57,6 +57,24 @@ role Perl6::Metamodel::RolePunning {
         self.pun($obj)
     }
 
+    my $proxy_type := nqp::null();
+    method !make_proxy($cont) {
+        if nqp::isnull($proxy_type) {
+            $proxy_type := nqp::gethllsym('Raku', 'Proxy');
+        }
+
+        my $proxy   := nqp::create($proxy_type);
```
lizmat, line 66: Why not a construct with `nqp::ifnull($proxy_type,nqp::gethllsym('Raku', 'Proxy'))` ?

dumarchie: Does that bind `$proxy_type` to the `Proxy` type object?

src/Perl6/Metamodel/RolePunning.nqp
dumarchie: I suppose that will make it more consistent. Thanks for the suggestion!

src/Perl6/Metamodel/RolePunning.nqp
lizmat: Good point.  No.  I actually meant:

````raku
nqp::ifnull($proxy_type,$prxy_type := nqp::gethllsym('Raku', 'Proxy'))
````

Comment (dumarchie) [reply src/Perl6/Metamodel/RolePunning.nqp:66]: I guess that would be a good micro-optimization. But not as good as eager initialization, which appears to be possible after all.

Comment (lizmat) [reply src/Perl6/Metamodel/RolePunning.nqp:66]: Indeed...  But I figured that wasn't possible, as this is a precursor to the Bootstrap, is it not?  And `Proxy` only gets defined in the bootstrap I believe?

Comment (dumarchie) [reply src/Perl6/Metamodel/RolePunning.nqp:66]: It appears that the Bootstrap is compiled before RolePunning. Both `nmake` and `nmake test` ran fine. But then `nmake spectest` hung, so I aborted:

```
t\spec\S17-promise\lock-async.t ................................... ok
===(   66393;538  21/28  0/?  0/?  0/?  0/?  0/? )======================NMAKE : fatal error U1058: terminated by user Stop.
```

I've seen that test hanging before, so I don't think it's related. I'll try again.

Comment (dumarchie) [reply src/Perl6/Metamodel/RolePunning.nqp:66]: Eager initialization causes spectests to fail. I'll go for the micro-optimization instead.

Comment (dumarchie) [reply src/Perl6/Metamodel/RolePunning.nqp:66]: On second thought, it doesn't feel right to use `nqp::ifnull` without using the return value. And I don't like to optimize without measuring, which I consider too much work for this case.

Please feel free to apply the optimization yourself if you think it's worth it :)

Comment (vrurg) [reply src/Perl6/Metamodel/RolePunning.nqp:66]: `RolePunning` cannot compile after the bootstrap because it is in the Metamodel.

File: src/Perl6/Metamodel/RolePunning.nqp [hunk 3]
```diff
@@ -57,6 +57,24 @@ role Perl6::Metamodel::RolePunning {
...
+        my $updated := 0;
+        nqp::bindattr($proxy, $proxy_type, '&!FETCH', -> $var {
+            $updated ?? nqp::decont($cont) !! $!pun;
+        });
+        nqp::bindattr($proxy, $proxy_type, '&!STORE', -> $var, $val {
+            nqp::assign($cont, $val);
+            $updated := 1;
```
vrurg, line 73: I feel very much uncertain about the use of `$updated`. What if `$val` is the same role? You get the same container with the same role inside but "updated".

vrurg (commented):
A hacky way to help a trick work. Don't really like it because what really happens is we get a solution for a rare use case at the expense of others, more common.

```raku
role R {
   method check-something { ... }
}

my R $v;

if $v.check-something {
    ....
}
```

Note that I don't even care if `check-something` is invoked on a container or not. And yet I will be operating on an implicit proxy object. I.e. the cost of supporting one particular use case would be:

1. One extra `Proxy` allocation.
2. One extra `Proxy` instance de-containerization (for `self`).
3. And who knows how many times more item 2 repetitions in a loop.

It feels to me that the correct solution must be done at codegen level where we know about `SELF` and that there are chances it would be used as a container.

src/Perl6/Metamodel/RolePunning.nqp
dumarchie: Why would a method store the original role in the invocant container? It would have to extract it from the pun first, so this will hardly be done by accident. And if a method does this for some reason, you just end up with a container that fetches the role, rather than its pun. If you then call another method on the container, that other method's invocant parameter will be bound to a new `Proxy`.

Maybe it helps if you think of `$updated` as `$initialized`. It's not called that way because the latter name would imply that you can only store a concrete thing in the container and the point of the `Proxy` is to allow code to store _anything_ that conforms to the type constraint of the original container. Once something is stored in the container, that's what you will get if you ask for the container's contents. Until something is stored in the container, you'll get its default value which happens to be the pun.

dumarchie:
I guess our disagreement boils down to the question what is more hackish:

* replace a container with a bare pun that cannot be updated
* replace a container with a `Proxy` that can be updated

I think that the original approach is more hackish. The fact that it's easier to optimize does not make it less hackish.

Obviously it would be least hackish to just change the behavior of the original container in the scope of the method call, but I don't know if that is even possible. Do you?

Regarding your use case, I think you're using the wrong idiom. If you're not interested in storing something in `$v`, you better define it like this:

```
my $v := R;
```

Then `$v` is not a container and you don't get any overhead associated with the use of containers.

As to my use case, how do you know it's rare? Maybe I'm just the first Raku developer who figured out what went wrong and complained about it. To be a bit more concrete, I discovered the issue when I tried to implement autovivification when pushing onto [a role I developed](https://raku.land/zef:dumarchie/BinaryHeap). I could not rely on [the `.push` candidate provided by `Any`](https://docs.raku.org/routine/push#(Any)_method_push) because my role is not `Positional` and there's no good reason to make it `Positional`.

Believe me, I don't like it either that this slows down all method calls on roles (on MoarVM). [One of my benchmarks](https://github.com/dumarchie/raku-binaryheap/blob/master/benchmark/binaryheap-new1.raku) now takes twice or thrice the time, though run times became much more stable. But Rakudo is developed on the principle "first make it run, then make it right, then make it fast" and I believe that in this case it was not made right before it was made fast (on MoarVM).

src/Perl6/Metamodel/RolePunning.nqp
vrurg: > Why would a method store the original role in the invocant container?

Say, something like this:

```raku
method foo(::?ROLE \SELF:) {
    SELF = self.are-we-ready ?? SELF.new !! ::?ROLE;
}
```

Without arguing why exactly this approach, the snippet is not prohibited in general. Though I didn't consider this case thoroughly and may miss something, but the result of `find_method` would be cached and the very same thunk would be invoked for the role. But wether I'm right here is not really relevant at the moment due to the bigger issue.

vrurg:
> Then $v is not a container and you don't get any overhead associated with the use of containers.

```raku
sub foo(Mu $v) {
    if $v.check-a-condition { ... }
}
```

Binding is not an option in the above case. `foo(ARole)` would severely suffer from perfrormance downgrade on roles.

> But Rakudo is developed on the principle "first make it run, then make it right, then make it fast"

I don't think this rule is applicable anymore. It was good for the early days when the community desperately needed a working model. Now it is time to focus on production uses. We cannot accept changes which would impact working code unless there are too good reasons to do so.

I'm not against the idea in general. I'm not against hacky approaches, especially if they're well-grounded or well-contained and explained. But in this particular case the solution has to be limited to exactly one use case: containerized method invocant. Unfortunately, neither pun, nor container cannot know _how_ they are being used. It is only known to the compiler. Hence, this is where the fix must be implemented in first place. BTW, thinking of it right now I guess it might be possible to somehow somehow tell `find_method` what is the purpose of its call, so it could accordingly adjust its behavior.

dumarchie:
Oh, I suddenly realized that we can probably restore the old hook in *dispatchers.nqp*, but only take the shortcut if the method has the [`is pure`](https://docs.raku.org/routine/is%20pure) trait. I think that's the most straightforward way for a developer to tell the compiler they don't care about containers.

Of course it remains to be seen if this is an overall performance improvement.

dumarchie:
Note that if the suggested `is decont` parameter trait is implemented, the presence of that trait on the invocant parameter could also trigger a shortcut in *dispatchers.nqp*. That would probably be even more useful than checking the `is pure` trait of the method. Then a role implementation can simply mark the invocant parameter of all methods as `is decont` to get maximum performance on MoarVM.

vrurg:
Let me just sum up my view. There is nothing new, but I can't spent too much on this.

For the purpose of implementing one feature with uncertain rate of use it is proposed to impose performance degradation on other, more common, use cases. In order to compensate the loss a couple of workarounds are proposed. Moreover, a developer would have to know about this specific behavior in order to utilize a workaround.

In neither way it sounds right. Even though I agree that the behavior is not explicitly prohibited and I'd like to see it possible, but the particular approach taken is wrong, up to my view. Thus, I'm back to square one: I would appreciate if proxy gets involved only and only when we deal with invocant. Moreover, it would be possible to make the feature even more explicit by checking the type of the invocant:

```raku
role R {
    method foo(::?ROLE \SELF:) { ... }  # use
    method bar(::?CLASS \SELF:) { ... } # don't
}
```

The optimizer could do even more by introspecting the code and ensuring that `SELF` is not used in a routine call as an argument _and_ that it gets assigned. But having it the proposed way would be sufficient to make the performance cost acceptable.

A couple of notes on the workarounds.

The proposed interpretation of `is pure` is wrong. It's meaning is broader and may not be applicable for cases where you see it. A method might and is likely to have other side effects while not using the containerization of its invocant. So, performance degradation would be inevitable even though the self-replacement feature is not used.

`is decont` sounds like it could be useful. Not sure it would be much faster than `<>` which could easily be optimized into `nqp::decont`, but yet would serve well for declarative purposes.

vrurg:
BTW, it's just crossed my mind that there is another solution where proxy is not getting involved whatsoever. `SELF` might point to the original container as-is, while `self` would be the pun class. After all, the purpose of `method foo(::?ROLE \SELF:)` declaration is exactly that: to point to the original object the method is invoked upon. By adding the word _'unmodified'_ to this definition we get what we want!

lizmat:
Is this PR still alive?

jnthn:
Just glancing over this (was too busy to look at it in detail before, and still am really), but in general, if a dispatch-related thing cannot be made decently fast with the power the generalized dispatch mechanism already offers, then there's probably not that much hope for it getting significantly faster in the future either.

dumarchie:
I also don't think anybody is going to make this solution fast enough, because use of a `Proxy` is basically a workaround.

Currently, auto-punning is implemented purely in terms of delegation. Maybe it could be approached in terms of auto-coercion: bind the invocant parameter to the pun _only_ if the parameter is constrained to a _class_ that implements the role. Then the role's `^find_method` can just return the pun's method, and if that method does _not_ constrain the invocant parameter to a class type, the parameter is bound to the original invocant argument, which may be a container.

I don't know what effect auto-coercion of the invocant parameter will have on the speed of method calls, though, or whether it would be possible to make a naive implementation faster. But if there's an interest in this approach I can try and whip up such an implementation.

dumarchie:
There doesn't appear to be an interest in changing the way auto-punning works.

Kaiepi:
The implementation of role punning has been bugging me for a while now. I'll see if I can do anything about this.

Kaiepi:
After messing with dispatch a bit, I wonder if it's feasible to write a routine (`only term:<groan>() is raw` or `only term:<groan>(&block) is raw`) returning the original invocant of a punned method dispatch or invoking a block under it respectively, in a similar means to how `samewith` and friends are implemented. This would help avoid any confusion or breakage to do with changing the behaviour of the invocant based off its type, and being an `only` invoked when explicitly requested, would allow for this behaviour to be introduced with a more minor performance hit.

Issue #4927, opened 2022-05-18
Regex error message should be detected during the parsing stage


florian-pe:
When executing this command `raku -e '"123456" ~~ m/ ^ $<.digit>=(\d)+/'` we get back the following error message
```
bindkey_o requires a concrete object (got a VMNull type object instead)
  in block <unit> at -e line 1
```
This look like a VM error message while the actual error actually looks like an illegal syntactic or semantic regex. If that is the case, then the exception should be done during the parsing of the regex, and the error message associated should be more useful.

In constrast, here is the error message associated with `raku -e '"123456" ~~ m/ ^ <digit>=(\d)+/'`
```
===SORRY!===
Unrecognized regex metacharacter = (must be quoted to match literally)
at -e:1
------> "123456" ~~ m/ ^ <digit>⏏=(\d)+/
Couldn't find terminator / (corresponding / was at line 1)
at -e:1
------> "123456" ~~ m/ ^ <digit>=⏏(\d)+/
    expecting any of:
        /

```

lizmat:
Thanks for your report.  Fixed with https://github.com/rakudo/rakudo/commit/da65a5f5135f453fbf1bb9013287d1682231e79a

Since it is technically not illegal code, a compilation error would not be appropriate.

Please re-open if you disagree.

PR #4928, opened 2022-05-18
Set empty list/hash on first access

lizmat:
For some reason, the mainline is not run, so the 0 elems branches
would return an nqp::null, which gets upgraded to a Mu, which in
turn causes the exception of #4927

### DIFF
--- a/src/Perl6/bootstrap.c/BOOTSTRAP.nqp
+++ b/src/Perl6/bootstrap.c/BOOTSTRAP.nqp
@@ -3546,8 +3546,8 @@ BEGIN {
 
         # Build a list of positional captures, or return a shared empty list if
         # there are none. This only populates the slots which need an array.
-        my $EMPTY-LIST;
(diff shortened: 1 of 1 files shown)


vrurg:
If replaced with `my %EMPTY-HASH; my @EMPTY-LIST` then wouldn't it be equivalent but more tidy? Besides, no run-time would be involved whatsoever.

lizmat:
I think the whole point was that it should be an NQP list/hash, not a HLL `List` / `Map`.  In any case, I was changing the code as little as possible.

I tried your change, and it killed my Terminal session while doing a spectest.

Now trying to reproduce now

niner:
Alternatively you could just move those variable declarations to the outer 
scope. That one's run for sure as it does all the add_method calls.

lizmat:
I'm pretty sure I tried that.  In any case, @vrurg's  suggestion worked out.

Issue #4935, opened 2022-05-23
Role group typechecking and dispatch to individual role candidates is too loose


Kaiepi:
## The Problem

A role is analogous to a routine. As such, it can be a multi of sorts if more than one role is declared with the same name:

```raku
role Container {
    has $!delegate;
    method item is raw { $!delegate.item }
    method throw-unless-default(--> True) { }
}

role Container[::V] does Positional[V] {
    has @!delegate is built(:bind);
    method AT-POS(+args) is raw { @!delegate.AT-POS: |args }
    method EXISTS-POS(+args) { @!delegate.EXISTS-POS: |args }
    method elems { @!delegate.elems }
}

role Container[::V, ::K] does Associative[V, K] {
    has %!delegate is built(:bind);
    method AT-KEY(+args) is raw { %!delegate.AT-KEY: |args }
    method EXISTS-KEY(+args) { %!delegate.EXISTS-KEY: |args }
    method elems { %!delegate.elems }
}

BEGIN say (
    try all (Container.new, my @ is Container, my % is Container)
        X~~ -> Container $topic is raw { $topic.throw-unless-default }
) // $!;
```

## Expected Behavior
`my @ is Container` and `my % is Container` should **not** be valid.

## Actual Behavior
Outputs `all(True, True, True)`.

## Steps to Reproduce

Save the golf and run it.

## Environment
* Operating system:
```
LAPTOP-7I9RCT0C% uname -a
Linux LAPTOP-7I9RCT0C 5.10.16.3-microsoft-standard-WSL2 #1 SMP Fri Apr 2 22:23:49 UTC 2021 x86_64 x86_64 x86_64 GNU/Linux
```
(OpenSUSE Tumbleweed, WSL2)
* Compiler version (`perl6 -v` or `raku -v`):
```
LAPTOP-7I9RCT0C% raku -v
Welcome to Rakudo™ v2022.04-75-g19c3bd180.
Implementing the Raku® Programming Language v6.d.
Built on MoarVM version 2022.04-3-ga9fcd5a74.
```

Kaiepi:
I wouldn't recommend tightening typechecks for this in v6.c, for the record.

vrurg:
I don't really see what's wrong here. Trait `is` tells that a symbol explicitly _is_ something. I.e. anonymous `@` and `%` literally become instances of `Container`. In order to get the positional or the associative candidates you'd need to parametrize them explicitly like `Container[Int]`.

[...]

Issue #4915, opened 2022-04-29
Smart matching a signature with a type variable fails (`:(::T, T)   ~~ :(::T, T); # No such method 'ACCEPTS' for invocant of type 'T'`)


raiph:
## The Problem

I was trying to require that a routine had a signature that took two `T`s, where `T` was any type.

## Expected Behavior

Simplifying:
```
say :(Int, Int) ~~ :(Int, Int); # True
say :(::T, Int) ~~ :(::T, Int); # True
say :(::T, T)   ~~ :(::T, T);   # True
```

## Actual Behavior

```
say :(Int, Int) ~~ :(Int, Int); # True
say :(::T, Int) ~~ :(::T, Int); # True
say :(::T, T)   ~~ :(::T, T);   # No such method 'ACCEPTS' for invocant of type 'T'
```

## Environment

```
v2022.02
```

Kaiepi:
Generics in signatures are broken in a number of ways. A related problem I'm aware of pertains to generic return types:

```raku
say :(::T --> T) ~~ :(::T --> T); # OUTPUT: False
```

This happens because the generic `T` on the left-hand side is not the exact same type object as the `T` defined by the RHS, which is to say it's backed by `=:=` rather than `ACCEPTS`. This means the following typecheck will not succeed, though it would if the return type were `ACCEPTS`-oriented instead:

```raku
say :(Any --> Any) ~~ :(Mu --> Mu); # OUTPUT: False
```

This typecheck succeeds despite this:

```raku
say :(Any) ~~ :(::T); # OUTPUT: True
```

I suspect this would be due to special handling of generic types in [`Parameter.ACCEPTS`](https://github.com/rakudo/rakudo/blob/master/src/core.c/Parameter.pm6#L664-L668). I don't think it should be concerned with metamodel archetypes to begin with. Typechecks *against* a generic type object should always succeed, but those *given* one should remain invalid in all cases instead. Types for generics can remain enforced by `Parameter.ACCEPTS` if a couple changes to `Metamodel::GenericHOW` are made:

- It needs an `accepts_type` method that always succeeds. Its typecheck mode may need adjustment to account for this.
- It needs to pretend to be `Mu` for the sake of an `ACCEPTS` method dispatch (all objects are supposed to support at least `Mu` anyways).

[...]

PR #4936, opened 2022-05-24
[WIP] Enhance interop of generics, roles, and classes, Raku or nqp

Kaiepi:
If generic role and generic signature signature typechecks are busted, maybe generic typechecks are busted too. In giving it the capacity to perform a typecheck in a more Raku-ish means, however, old errors seen from https://github.com/rakudo/rakudo/pull/4478 reappear. This PR makes a poor starting point because I wind up spinning off into two directions simultaneously past where I left off:

- `accepts_type` leads into problems with generics and enhancements to `Signature.ACCEPTS`/`Parameter.ACCEPTS` and syntax that can be made because of it. nqp needs to mirror Rakudo here, which has its own set of problems to begin with.
- Changing archetypes checks leads into a game of cat and mouse against the MOP as a whole. The gist is Inherit(aliza)bility and compos(aliza)bility can be made more robust if given protocol, but this brings HOWs behaving as composers or children, e.g. enums, into the picture when they wouldn't need to be otherwise (and again, nqp reflects). This is more problem-solving territory.

There's a way to split this in half, taking only the former changes (and thus a portion of the bug fixes) for now.

## Bug Fixes

As before, subtyping of generics is OK:

```raku
role Foo[Any ::T] { }
role Bar[::T] does Foo[T] { }
```

```raku
role Foo[Any ::T] { }
role Bar[::T] is Foo[T] { }
```

But composalizability/inheritalizability is ignored, and thus the following is still **not** valid:

```raku
role OfIndex[Enumeration ::T] does T { }
enum Index <A B C>;
OfIndex[Index]
```

Roles should behave OK as-is, but only because they typecheck similarly enough to their puns and pun to the same metaobject upon method dispatch.

## Typechecks

```raku
enum Index <A B C>;
class Foo does Index { }
Foo ~~ Index
```

This should *not* be a MOP-enforced typecheck. `Index` implies `NumericEnumeration` and thus `Enumeration`, which `Foo` doesn't do because it's an `Index` coercer.

```raku
class Foo is Buf[uint8] is repr<VMArray> { }
Foo ~~ Blob[uint8];
```

Now valid.

```raku
class Foo is Buf is repr<VMArray> { }
Foo ~~ Buf[uint8];
```

This is a bizarre type... I'm not sure how `Buf`  can guarantee `Buf[uint8]` specifically. Remember, `Buf` refers to the role *group*, not the empty parameterization of `Buf`, which remains empty as far as the MOP is concerned because type argument defaults are applied so late.

## Issues

This is prompted by https://github.com/rakudo/rakudo/issues/4915 and the many problems with typechecks I've gotten annoyed by over the years. https://github.com/rakudo/rakudo/issues/4935 is prompted by this, and the noted behaviour is replicated here.

## Tests

I'm aware this breaks two tests in `t/02-rakudo/13-exceptions`  and `t/spec/S02-names-vars/names.t` to do with the following syntax:

```raku
my ::T $
```

Which was presumed to be invalid before. I haven't done anything about this yet.

Only MoarVM is tested thus far.

## Todo

- [ ] Specialization could probably gain a lot from a more prominent curried role by this point.
- [ ] Raku roles and NQP roles should expose more compatible introspection metamethods in order for custom HOWs to behave as expected for parent, roles, methods, and attributes.
- [ ] `my ::T $` needs taking care of.
- [ ] Iffy typecheck/composition hacks in the setting need review.

May add more items as I think of them.

### DIFF
--- a/src/Perl6/Metamodel/GenericHOW.nqp
+++ b/src/Perl6/Metamodel/GenericHOW.nqp
@@ -23,7 +23,8 @@ class Perl6::Metamodel::GenericHOW
         my $meta := self.new();
         my $obj := nqp::settypehll(nqp::newtype($meta, 'Uninstantiable'), 'Raku');
         $meta.set_name($obj, $name);
-        $obj
+        nqp::settypecheckmode($obj,
+            nqp::const::TYPE_CHECK_NEEDS_ACCEPTS)
     }
 
     method instantiate_generic($obj, $type_environment) {
@@ -35,8 +36,8 @@ class Perl6::Metamodel::GenericHOW
     method compose($obj) {
     }
 
-    method type_check($obj, $checkee) {
-        0
+    method accepts_type($obj, $checkee) {
+        1
     }
 }
 
--- a/src/Perl6/Metamodel/GenericHOW.nqp
+++ b/src/Perl6/Metamodel/GenericHOW.nqp
@@ -4,6 +4,7 @@
 # of these confers genericity on the holder.
 class Perl6::Metamodel::GenericHOW
     does Perl6::Metamodel::Naming
+    does Perl6::Metamodel::MethodDelegation
 {
     my $archetypes := Perl6::Metamodel::Archetypes.new( :generic(1) );
     method archetypes() {
@@ -32,10 +33,6 @@ class Perl6::Metamodel::GenericHOW
     }
 
     method compose($obj) {
-    }
-
-    method find_method($obj, $name) {
-        nqp::null()
     }
 
     method type_check($obj, $checkee) {
--- a/tools/build/makeBLOB_ROLES.raku
+++ b/tools/build/makeBLOB_ROLES.raku
@@ -58,8 +58,9 @@ while @lines {
     say Q:to/SOURCE/.subst(/ '#' (\w+) '#' /, -> $/ { %mapper{$0} }, :g).chomp;
 
 my role #name#[::T] is repr('VMArray') is array_type(T) is implementation-detail {
-    method !push-List(str $action, ::?CLASS:D $to, \from) {
+    method !push-List(str $action, ::?CLASS:D \to, \from) {
         my Mu $reified := nqp::getattr(from,List,'$!reified');
+        my Mu $to      := nqp::decont(to);
         if nqp::isconcrete($reified) {
             my int $elems = nqp::elems($reified);
             my int $j     = nqp::elems($to);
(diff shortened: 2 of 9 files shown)


Kaiepi:
I took a quick look at specialization. While binding also instantiates generics, it deals in too different a HOW, too early to help as much as I'd hope. Maybe the instantiation of supertypes can be skipped in some way, but it's not as high a priority as getting nqp to play a little nicer with Raku to me.

src/Perl6/Metamodel/GenericHOW.nqp
```diff
@@ -22,7 +23,8 @@ class Perl6::Metamodel::GenericHOW
         my $meta := self.new();
         my $obj := nqp::settypehll(nqp::newtype($meta, 'Uninstantiable'), 'Raku');
         $meta.set_name($obj, $name);
-        $obj
+        nqp::settypecheckmode($obj,
+            nqp::const::TYPE_CHECK_NEEDS_ACCEPTS)
```
vrurg, line 27: I think `TYPE_CHECK_CACHE_DEFINITIVE` can be used here. The cache can be set to `Mu`. This would be equivalent to `accepts_type` being always true.

BTW, I memory serves me right, I was experimenting with making `GenericHOW` to implement `TypePretense`. Sure enough, I don't remember what's went wrong back then, but it still looks reasonable to me.

tools/build/makeBLOB_ROLES.raku
```diff
@@ -58,8 +58,9 @@ while @lines {
     say Q:to/SOURCE/.subst(/ '#' (\w+) '#' /, -> $/ { %mapper{$0} }, :g).chomp;
 
 my role #name#[::T] is repr('VMArray') is array_type(T) is implementation-detail {
-    method !push-List(str $action, ::?CLASS:D $to, \from) {
+    method !push-List(str $action, ::?CLASS:D \to, \from) {
```
vrurg, line 61: Does the part about `Blob`/`Buf` really belongs to this PR?

src/Perl6/Metamodel/GenericHOW.nqp
Kaiepi: Interesting. That avoids the `accepts_type` and is otherwise equivalent under most circumstances, the exception being prior `type_check`s that fail, e.g. generic-generic smartmatch. I added this because I thought signatures would need that check in particular in order to match other generic signatures' generics, but the new signature tests pass fine without.

tools/build/makeBLOB_ROLES.raku
Kaiepi: It's a commit I've been hoping to get rid of, but `S03-operators/buf.t` wouldn't play nice with the other changes without. Doing the `Mu` type cache allows those to pass.

Kaiepi:
The `Metamodel::MonicMachine` is capable of optimizing the MRO and role typecheck lists, as well as serving as a generic `VMArray` for type objects in the MOP. This helps simplify the work involved in getting nqp more up to Rakudo's speed WRT parents, methods, attributes, etc. I have tried time and time again to optimize `compute_mro` with this, but am left with more of a distaste for C3 and multiple inheritance as a whole than anything else. I have some ideas for an alternative MRO that is mostly compatible, cheaper to produce, but unfortunately, mostly compatible in the ecosystem I guess.

PR #4937, opened 2022-05-25
Reconsider the way `$*PERL` is deprecated

vrurg:
The earlier implementation wasn't taking old 6.c/6.d code into account when it comes to modules. I.e. if some 6.e code uses a 6.c/6.d module which, in turn, references `$*PERL` then the deprecation warning is issued anyway. This would be a violation of the backward compatibility rule.

### DIFF
--- a/src/Perl6/Actions.nqp
+++ b/src/Perl6/Actions.nqp
@@ -3058,10 +3058,27 @@ class Perl6::Actions is HLL::Actions does STDActions {
         make_variable_from_parts($/, @name, ~$<sigil>, ~$<twigil>, ~$<desigilname>);
     }
 
+    # System variables which are considered deprecated. Keys are variable names, values are list of revision which
+    # deprecates the variable, and alternative to be used instead.
+    my %variable_deprecations := nqp::hash(
+        '$*PERL', nqp::list('e', '$*RAKU'),
+    );
+
     sub make_variable_from_parts($/, @name, $sigil, $twigil, $desigilname) {
         my $past := QAST::Var.new( :name(@name[+@name - 1]), :node($/));
         my $name := $past.name();
         my $world := $*W;
+        my @deprecation;
+
+        if nqp::existskey(%variable_deprecations, $name)
+            && nqp::isge_s(
+                nqp::getcomp('Raku').language_revision,
+                %variable_deprecations{$name}[0])
+        {
+            @deprecation := nqp::clone(%variable_deprecations{$name});
+            nqp::push(@deprecation, $world.current_file);
+            nqp::push(@deprecation, $world.current_line($/));
+        }
 
         if $twigil eq '*' {
             if +@name > 1 {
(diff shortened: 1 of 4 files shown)


vrurg:
The idea is to let the deprecation to be issued at the moment of vitalizing a system dynamic variable, so that the following code would still work without a warning:

```raku
my $*PERL = "ok";
say $*PERL;
```

To achieve this, `&DYNAMIC` is invoked with additional `@deprecations` parameter which hold the necessary information for producing the warning when necessary.

lizmat:
Adding extra parameters to `&DYNAMIC` is a **VERY** bad idea from a performance point of view.

`&DYNAMIC` is one of the major bottlenecks in performance, especially those cases where a lot of dynamic variables are used, think `react` for instance.

I think this commit must be reversed until at least after the 2202.05 release

vrurg:
@lizmat I don't think it would be a problem, especially under new-disp governance. The cost is applied when an argument is actually passed in because the binder iterates over arguments. And the only time when the argument is passed is when a deprecation is encountered.

I was considering compile-time warning issuing but, as stated above, it would then complain over custom-declared instances too which is not correct.

vrurg:
Actually, it made me think of making a duplicate of `&DYNAMIC`, `&DYNAMIC-DEPRECATED`. It would make things even more reliable.

vrurg:
I did a bit of straightforward testing and it doesn't look like extra complication is needed. Correct me if the following benchmark is somehow wrong:

```raku
constant TIMEOUT = 5;
my atomicint $ready = 0;
my atomicint $stop = 0;

sub bar {
    $ready ⚛= 1;
    until $stop {
        ++$*TEST;
    }
}

sub foo {
    my $*TEST = 0;
    bar;
    say $*TEST;
}

start {
    until $ready {};
    sleep TIMEOUT;
    $stop ⚛= 1;
}
foo;
```

Apparently, the more the value in `$*TEST` the faster it works. So, release v2022.04 has made it as many as 27491238. The _master_ managed to display 28043030. To make sure the result is not affected by new optimizations somewhere in between the release and the _master_, I made a build out of 19c3bd1809bf2c74da2f6fec69394e6f8b360391, which immediately precedes this PR. It resulted in as high number as 27800214.

With these results I don't think there is a reason to further complicate otherwise rather deprecated compiler code.

lizmat:
Ok, glad to have been proven wrong.  I will update my prejudices now  :-)

jnthn:
> The cost is applied when an argument is actually passed in because the binder iterates over arguments.

This isn't true; the binder very much iterates over parameters, because for named ones then `:(:$foo, :$bar = $foo)` has to bind `$foo` first no matter the order of the `foo` and `bar` named args. However, in the end it doesn't matter so much, since:

* In most cases, the signature is not interpreted, but the binding code is compiled
* Spesh can see that an optional argument is not being passed and eliminate the code handling the optional parameter,folding it to the default
* The default is often a type object, so constant propagation can potentially eliminate any further conditionals inside of the routine

PR #4946, opened 2022-06-07
Add Int/Real coercing versions of `infix:<div|mod>`

MasterDuke17:
This makes them consistent with the other operators, so e.g., `say "6"
div 3` now prints `2`, instead of
`Calling infix:<div>(Str, Int) will never work with any of these multi signatures:
    (Int:D \a, Int:D \b --> Int:D)
    (int $a, int $b --> int)
    (uint $a, uint $b --> uint)`

Fixes https://github.com/Raku/problem-solving/issues/326

### DIFF
--- a/src/core.c/Int.rakumod
+++ b/src/core.c/Int.rakumod
@@ -347,6 +347,12 @@ multi sub infix:<div>(Int:D $a, Int:D $b --> Int:D) {
 multi sub infix:<div>(int  $a, int  $b --> int)  { nqp::div_i($a, $b) }
 multi sub infix:<div>(uint $a, uint $b --> uint) { nqp::div_i($a, $b) }
 
+multi sub infix:<div>($a, $b --> Int:D) {
+    $b.Int
+      ?? nqp::div_I($a.Int,$b.Int,Int)
+      !! X::Numeric::DivideByZero.new(:using<div>, :numerator($a)).Failure
+}
+
 multi sub infix:<%>(Int:D $a, Int:D $b --> Int:D) {
     nqp::isbig_I($a) || nqp::isbig_I($b)
       ?? $b
--- a/src/core.c/Real.rakumod
+++ b/src/core.c/Real.rakumod
@@ -160,6 +160,7 @@ multi sub prefix:<->(Real:D $a) { -$a.Bridge }
 # so fails on differing types."  Thus no casts here.
 proto sub infix:<mod>($, $, *%) is pure {*}
 multi sub infix:<mod>(Real:D $a, Real:D $b) { $a - ($a div $b) * $b }
+multi sub infix:<mod>($a, $b) { $a.Real - ($a.Real div $b.Real) * $b.Real }
 
 multi sub abs(Real:D $a) { $a < 0 ?? -$a !! $a }
 


src/core.c/Real.pm6
```diff
@@ -159,6 +159,7 @@ multi sub prefix:<->(Real:D $a) { -$a.Bridge }
 # so fails on differing types."  Thus no casts here.
 proto sub infix:<mod>($, $, *%) is pure {*}
 multi sub infix:<mod>(Real:D $a, Real:D $b) { $a - ($a div $b) * $b }
+multi sub infix:<mod>($a, $b) { $a.Real - ($a.Real div $b.Real) * $b.Real }
```
lizmat, line 162: Why not `(Real() $a, Real() $b)` as signature?

MasterDuke17: Coercion signatures are quite a bit slower. E.g., `multi sub foo(Int $a, Int $b) { $a + $b - 2 }; multi sub foo($a, $b) { $a.Int + $b.Int - 2 }; my $a; $a = foo(~$_, $_ + 1) for ^1_000_000; say now - INIT now; say $a` reports ~0.5s, but `multi sub foo(Int() $a, Int() $b) { $a + $b - 2 }; my $a; $a = foo(~$_, $_ + 1) for ^1_000_000; say now - INIT now; say $a` reports ~1.0s.

lizmat: Ooh... in that case, that feels like low-hanging fruit ?

MasterDuke17:
Hm, not sure exactly what you mean? Making using coercion signatures
faster? Removing their use from the core setting?

lizmat:
Making coercions faster :-)

MasterDuke17:
Ah. I don't really know anything about how coercions are implemented, but
maybe @vrurg knows if there's any low-hanging fruit.

vrurg:
I see nothing low-hanging in there. I even tried to re-implement the protocol in new-disp and gave up.

As to the signature, it already tries to avoid the protocol when possible. Like, the protocol is not invoked for `Int()` if the value is already `Int`. Further steps would require pulling out some coercion logic out of the protocol code and moving it into the signature binding, but do we really want it? Don't forget that it wouldn't be sufficient to, say, call `.Int` on a non-`Int` if it has the method, but any errors are to be processed and correct exception thrown.

MasterDuke17:
Hm. Now the difference between my two examples above is quite a bit smaller. ~0.48s for the non-coercion signature version and ~0.55s for coercion signature version. Should I switch to the "cleaner" coercion signature version?

vrurg:
Took me a while to refresh memory on this topic...

> Should I switch to the "cleaner" coercion signature version?

I'd say it makes sense.

MasterDuke17:
> > Should I switch to the "cleaner" coercion signature version?
> 
> I'd say it makes sense.

Hm. Now I'm wondering if it should be `Int() $a, Int() $b` or `Int(Cool) $a, Int(Cool) $b`. And if (m)any the existing multis (for not just `infix:<div|mod`>) need to be standardized on one or the other.

vrurg:
> Now I'm wondering if it should be `Int() $a, Int() $b` or `Int(Cool) $a, Int(Cool) $b`

`Int()` (or `Real()`). A class doesn't have to be `Cool` to make sense in an integer context. Say, a price tag would represent a value, a currency, and a name. And, yet, it remains to be a price in first place. Though it makes sense to inherit from `Cool`, but internal logic of application may not benefit from this.

> And if (m)any the existing multis (for not just `infix:<div|mod`>) need to be standardized on one or the other.

If you have time and stamina to do it then the answer is pure "yes" here.

MasterDuke17:
I realized there are already `Int:D $a, Int:D $b` and `Real:D $a, Real:D $b` multis. So it would probably make more sense to just make those coercing instead. But then a more realistic benchmark gets quite a bit worse. So this is what I was just testing:
```raku
multi sub foo($a, $b)                 { $a.Int + $b.Int - 2 };
multi sub foo(Int:D $a, Complex:D $b) { $a + $b.Int - 2 };
multi sub foo(Int:D $a, Int:D $b)     { $a + $b - 2 };
my $a;
for ^1_000_000 -> Int $_ {
    $a = foo(~$_, ~$_)
};
say now - INIT now;
say $a
```
which takes ~0.85s.
vs
```raku
multi sub foo(Int:D $a, Complex:D $b) { $a + $b.Int - 2 };
multi sub foo(Int:D() $a, Int:D() $b) { $a + $b - 2 };
my $a;
for ^1_000_000 -> Int $_ {
    $a = foo(~$_, ~$_)
};
say now - INIT now;
say $a
```
which takes ~2.30s.

So now I'm leaning towards merging this as is, and then doing some more testing with these core multis and seeing what the impact is on them of converting the `Type:D $a` + `$a.Type` multis into a single `Type:D() $a` multi.

MasterDuke17:
Yeah, actually making these coercers is a decent bit slower. With this PR as is, `my $a; for ^1_000_000 -> Int $_ { $a = ~$_ div ~($_ + 1) }; say now - INIT now; say $a` takes ~1.29s, but if I don't add a new multi and instead make the existing `Int:D` version `Int:D()` then it takes ~2.34s.

melezhik:
confirming test passes https://sparky.sparrowhub.io/report/gh-melezhik-r3tool-openbugs/22692

PR #4950, opened 2022-06-08
Consuming class must provide implementation of stubs

lizmat:
This commit changes the way the check for stubbed methods is performed
on classes that consume roles.  Before, if the class could provide a
method for the stub, **even if it was from an inherited class**, it
would be accepted.  This behaviour makes:

    role A { method WALK { ... } }.^pun;

work, while:

    role A { method STAND { ... } }.^pun;
    # Method 'STAND' must be implemented by A because it is required by roles: A

fails.  And that's just because Mu actually defines a WALK method.
Which implies that removing such an internal method such as "WALK" from
Mu, has the danger of breaking people's code *at compile time*.

So I feel this behaviour is inconsistent: the class that is consuming
a role should *itself* provide a stubbed method.  Which could well be
something like:

    role A { method list { ... } }
    class B does A { method list(|) { nextsame } }

Similar to solving collisions.  This semantic is implemented by this
commit.  Furthermore it has improved error reporting on unimplemented
stub methods, by not just giving up on the first, but instead collect
all offenses and format this into a more usable message:

    role A {
        method a {...}
        method b {...}
    }
    role B {
        method c {...}
    }
    class C does A does B { }
    The following methods must be implemented:
      for class 'C':
        'c' required by role 'B'
        'a', 'b' required by role 'A'

This change breaks *one* spectest.  And this can be golfed to:

    QuantHash.does(Associative)
    # The following methods must be implemented:
    #   for class 'QuantHash':
    #     'hash', 'Hash', 'Map' required by role 'QuantHash'

which appears to be improper handling at:

    https://github.com/rakudo/rakudo/blob/master/src/vm/moar/dispatchers.nqp#L819

where it appears it is deciding incorrectly that the invocant should
be punned.

### DIFF
--- a/src/Perl6/Metamodel/ClassHOW.nqp
+++ b/src/Perl6/Metamodel/ClassHOW.nqp
@@ -84,20 +84,6 @@ class Perl6::Metamodel::ClassHOW
         @!fallbacks[+@!fallbacks] := %desc;
     }
 
-    sub has_method($target, $name) {
-        for $target.HOW.mro($target) {
-            my %mt := nqp::hllize($_.HOW.method_table($_));
-            if nqp::existskey(%mt, $name) {
-                return 1;
-            }
-            %mt := nqp::hllize($_.HOW.submethod_table($_));
-            if nqp::existskey(%mt, $name) {
-                return 1;
-            }
-        }
-        return 0;
-    }
-
     method compose($the-obj, :$compiler_services) {
         my $obj := nqp::decont($the-obj);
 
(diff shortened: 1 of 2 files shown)


vrurg:
> So I feel this behaviour is inconsistent: the class that is consuming
> a role should _itself_ provide a stubbed method. Which could well be
> something like:

I disagree from this moment on. If the required method is implemented by a parent class and this is _intentional_ then the only solution would be to re-declare the method on the child class which is clearly sub-optimal approach.

If a role requires a method implementation then it in general means that a class consuming it must have the method. Nothing else.

lizmat:
I understand your point.  But is it realistic?  Having a method implemented by a parent class of a class consuming a role, feels to me like action at a distance.  We also need to provide a method in case of collisions, so why would this be any different?  And the only roast failure on this PR, is from something unrelated, basically uncovering an existing issue.

So I don't see why this would be a big issue.

What I do think is a big issue, is classes consuming roles which have stubs, while the mention of the stub in the role is completely meaningless because `Any` or `Mu` provides **an** implementation of it.

    role A { method iterator { ... } }
    class B does A { }

compiles just because of `Any.iterator`.  Whereas the author of the role probably wanted the consuming class to **really** provide an `iterator` method.

vrurg:
> We also need to provide a method in case of collisions, so why would this be any different?

Because classes are resolved with MRO? Roles are not.

> So I don't see why this would be a big issue.

Neither it resolves any issue, as to my view. Instead it introduces a limitation and breaks the backward compatibility.

> Whereas the author of the role probably wanted the consuming class to **really** provide an `iterator` method.

I'm not really sure why would they want to. If they're not satisfied with the standard implementation somehow then requiring a different method name seems to be more appropriate.

On the other hand, if a role requires `iterator` and I want to mixin it into `Slip` then what's so wrong about `Slip` inheriting the method from `List`?

Nah, I'm certainly not happy about the idea. Though I see the purpose. And what we can do about it is consider what kind of yada-yada is used do declare the stub. `!!!` might have the meaning of "give it to me here and now". It still might break the backward compatibility, but the chances of this are much smaller as people are generally used to use `...`. It could also be implemented as a part of 6.e spec.

ugexe:
> Having a method implemented by a parent class of a class consuming a role, feels to me like action at a distance. 

Do you have any example of other languages that *don't* treat interfaces in the way we do now? To me it would seem inconsistent to do it differently than it currently is.

PR #4953, opened 2022-06-10
Make coercions up to 1.6x as fast

lizmat:
- calling method named after type, e.g. Int() -> .Int: 1.6x as fast
- calling .COERCE: 4% faster
- calling .new:    3% faster

This has been achieved by breaking up the large "coerce" method into
several (private) methods, allowing the fast path (the .Int case) to
have to do a lot less.  The outer "coerce" method has a check for
already the correct type.  Since that is small enough to get inlined,
one wonders whether the current dispatch check on coercing is really
needed, and if removed, would result in smaller bytecode at the call
site, with potentially earlier / better inlining.

Use of intermediate variables has also been done extensively to
reduce repeated attribute and method calls.

All error handling is put into separate private methods to
allow the bytecode for the normal path to be smaller.

All is spectest clean, and error messages should be identical to
before this commit.

### DIFF
--- a/src/Perl6/Metamodel/CoercionHOW.nqp
+++ b/src/Perl6/Metamodel/CoercionHOW.nqp
@@ -113,106 +113,134 @@ class Perl6::Metamodel::CoercionHOW
     # Coercion protocol method.
     method coerce($obj, $value) {
         nqp::istype($value, $!target_type)
-          ?? $value
-          !! self.actually_coerce($obj, $value)
+          ?? $value                             # already done
+          !! self."!coerce_TargetType"($obj, $value)
     }
 
-    method actually_coerce($obj, $value) {
-        # Support nested coercions
-        if nqp::can($!constraint_type.HOW.archetypes, 'coercive')
-            && $!constraint_type.HOW.archetypes.coercive
-        {
-            $value := $!constraint_type.HOW.coerce($!constraint_type, $value);
-        }
+    # Attempt coercion on TargetType
+    method !coerce_TargetType($obj, $value) {
+        my $constraint    := $!constraint_type;
+        my $constraintHOW := $constraint.HOW;
+        $value := $constraintHOW.coerce($constraint, $value)
+          if nqp::can((my $archetypes := $constraintHOW.archetypes), 'coercive')
+          && $archetypes.coercive;
 
-        my $hint;
-        my $coerced_value := nqp::null();
-        my $value_type := nqp::what($value);
-        my $coercion_method;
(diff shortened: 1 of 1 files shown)


vrurg (approved):
This is something I was considering but when attempted to implement a dispatcher. Unfortunately, the only possible implementation is not fully implementable on the call site and required creating a megamorphic location, etc., etc...  So, I never got to it and never thought that inlining alone might work pretty well too.

vrurg:
Speaking about dispatching, I lost too much time on generic implementation before giving up due to lack of time. But as I look into it now, I see one thing that can be done easily. Apparently, the top-level dispatch might bypass calling `coerce` whatsoever if the value is of the target type already, that's simple. 

But at the next level it might fallback to calling a `CoercionHOW` method only when the constraint type is coercive. Otherwise the step of nested coercion is skipped anyway and from this moment on we can dispatch to either `.Type` or `.COERCE` directly. Eventually, we would fallback to `!coerce_new` because it has to handle the exception. But even then there are a chances to do a few moves inside the dispatching.

If there is nobody to try this path soon, I hope to get to it in a couple of weeks.

lizmat:
> So, I never got to it and never thought that inlining alone might work pretty well too.

Only `.^coerce` itself is currently getting inlined.  Unfortunately, all of the other methods are still too big to get inlined  :-(   But they **are** faster :-)

vrurg:
> However, _removing_ that check from `.^coerce`, breaks the setting compilation.

Because it is bypassed in the signature but is necessary otherwise. That's why re-delegating this task to new-disp would be more helpful as it is more universal.

> Where should I look for that, so I can do some testing to see whether it makes a difference?

The check must remain in place because `coerce` is supposed to be part of the MOP API. A user must be able to use it without additional check.

But what could be done, BTW, is bypassing the call to `.coerce` in BOOTSTRAP. Possibly doesn't make much difference in `bind_one_param` which is big and slow anyway. But could pay off for `scalar_cont_spec` on `store` key. And in the `invoke_forwarder`.

> But they **are** faster :-)

This I currently don't get. I'm still living under impression than a method call is costly and better be avoided in most cases. Except for really small inlineable ones.

lizmat:
Well, maybe they're faster because I really worked on preventing multiple calls and attribute accesses.  In which case, since none of the other methods is getting inlined, might make sense to merge all of them into the one (except for the error handling ones).  Will look at that tomorrow, am a little tired right now, after cycling 60+ km  :-)

lizmat:
Merging things into one method, removed all the gains.

PR #4956, opened 2022-06-14
Micro-optimize ADD-PAIRS-TO-BAG ...

MasterDuke17:
by turning an nqp::if(nqp::existskey,nqp::atkey) into just an
nqp::if(nqp::atkey).

My benchmark (`say "big.txt".IO.slurp.lc.words.Bag.elems`, where "big.txt" has 75,161 words and is 6.2mb) drops from ~0.97s to ~0.92s and from ~7.5b instructions to ~7.2b instructions.

### DIFF
--- a/src/core.c/Rakudo/QuantHash.pm6
+++ b/src/core.c/Rakudo/QuantHash.pm6
@@ -1246,18 +1246,12 @@ my class Rakudo::QuantHash {
               )
             ),
             nqp::if(               # not a Pair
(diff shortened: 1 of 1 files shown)


2colours:
Even the code would get simpler, nice.

What's the state of this PR?

MasterDuke17:
I got distracted before I could test the other cases, but I do plan to get
to that, thanks for the reminder.

PR #4957, opened 2022-06-16
Make unsuccessful connect fail rather than die

lizmat:
Spotted in https://stackoverflow.com/questions/72639883/how-to-deal-with-exceptions-in-iosocketinet

There is no documentation or tests covering the behaviour of a failed
connect.  It is actually the nqp::connect that throws (which turns out
also to be not documented BTW).

Having it return a Failure feels more natural indeed.

### DIFF
--- a/src/core.c/IO/Socket/INET.pm6
+++ b/src/core.c/IO/Socket/INET.pm6
@@ -113,6 +113,7 @@ my class IO::Socket::INET does IO::Socket {
 #?endif
         }
         elsif $!type == SOCK_STREAM {
+            CATCH { return .Failure }
             nqp::connect($PIO, nqp::unbox_s($!host), nqp::unbox_i($!port), nqp::decont_i($!family));
         }
 


Kaiepi (approved):

vrurg:
Turns out, this PR requires spec changes. Our new flapper, _S32-io/socket-recv-vs-read.t_, is built around the fact that a new connection throws on error. Same assumption is made about `listen`, BTW.

I'm considering reverting this PR unless there is a very good reason not to do it.

Kaiepi:
I'd like for `IO::Socket::INET`'s `connect`/`listen` to err to fail like its own `IO::Socket` `read`, `recv`, `write`, `close` does in the future, if only to simplify chaining of operations, e.g.:
```raku
with IO::Socket::INET.connect: '127.0.0.1', 23 -> $client {
    LEAVE { $client.close }
    [andthen] gather take $client.print("$_$?NL") for ^100
}
```
But I'd rather not see such a change break networking code.

PR #4959, opened 2022-06-16
Todo failing test on Solaris

vlmarek:
Solaris always returns 0 in max-rss. According to the NOTE in
getrusage(3C):

  The ru_maxrss, ru_ixrss, ru_idrss, and ru_isrss members of  the rusage
  structure are set to 0 in this implementation.


vlmarek:
My understanding is that `getrusage()` retrieves statistics for current process while `sysconf()` reads system wide limits. It's just that Solaris never implemented `getrusage()` fully.

lizmat:
Thanks!

PR #4968, opened 2022-06-25
Make raku installation portable

locriacyber:
- [X] Create portable installation with `zef install --prefix=yourrepohere .`
- [X] ~~Make raku/moarvm portable (use relative path to get dependencies, ship without nqp)~~ Just use https://rakudo.org/downloads

### DIFF
--- a/src/core.c/CompUnit/Repository/Installation.pm6
+++ b/src/core.c/CompUnit/Repository/Installation.pm6
@@ -32,6 +32,9 @@ __END__
 
     my constant $raku-wrapper = '#!/usr/bin/env #raku#
 sub MAIN(:$name, :$auth, :$ver, *@, *%) {
+    my $vendor_dir = $*SPEC.catdir($?FILE.IO.dirname, '..').IO.resolve;
+    my $vendor_repo = CompUnit::Repository::Installation.new(prefix => $vendor_dir);
+    CompUnit::RepositoryRegistry.use-repository($vendor_repo);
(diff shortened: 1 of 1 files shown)


src/core.c/CompUnit/Repository/Installation.pm6
```diff
@@ -32,6 +32,9 @@ __END__
 
     my constant $raku-wrapper = '#!/usr/bin/env #raku#
 sub MAIN(:$name, :$auth, :$ver, *@, *%) {
+    my $vendor_dir = $*SPEC.catdir($?FILE.IO.dirname, '..').IO.resolve;
```
locriacyber, line 35: `.IO.resolve` is technically unnecessary.

ugexe (changes requested):
This PR modifies the repo chain when it should not. fwiw I don't really understand what your issue is (installing to e.g. vendor works fine for me, and I'm not familiar with anyone else ever having such a problem), but this change would be a regression.

locriacyber:
What vendor? I need to make the application to be run anywhere.
AFAIK the current Repository use absolute path.

niner:
That is already supported. Rakudo can be built in a relocatable version with `perl Configure.pl --relocatable`. From the `--help` text:
```
    --relocatable      Dynamically locate NQP and Perl6 home dirs instead of
                       statically compiling them in. (On AIX and OpenBSD Rakudo
                       is always built non-relocatable, since both OSes miss a
                       necessary mechanism.)
```

abraxxa:
Does this help text really still read 'Perl6'?

PR #4972, opened 2022-06-28
Make Array.List with holes respect .default

lizmat:
Spotted by gfldex++ at https://irclogs.raku.org/raku-dev/2022-06-27.html#21:28
this commit will honour any (defined) default value of the Array.

However, after 12d7d5b48add8347eb119 a test for .List **always** returning
Nil was codified in t/spec/S32-array/delete.t .  In retrospect, I think
this is wrong, since a Slip is a specialization of List.  And I also think
that the current implementatation of Array.Slip is actually erroneous,
as it doesn't reify as .List does.

For performance, the ":view" implementation-detail functionality is now
handled by a separate candidate (as as such, properly marked as an
implementation-detail).

### DIFF
--- a/src/core.c/Array.pm6
+++ b/src/core.c/Array.pm6
@@ -429,26 +429,36 @@ my class Array { # declared in BOOTSTRAP
             !! Rakudo::Iterator.Empty
     }
 
-    multi method List(Array:D: :$view --> List:D) {  # :view is implementation-detail
+    multi method List(Array:D: :$view! --> List:D) is implementation-detail {
+        self.is-lazy    # reifies
+          ?? self.throw-iterator-cannot-be-lazy('.List')
+          !! nqp::isconcrete(my $reified := nqp::getattr(self,List,'$!reified'))
+            ?? $reified.List
+            !! nqp::create(List)
+    }
+    multi method List(Array:D: --> List:D) {
         nqp::if(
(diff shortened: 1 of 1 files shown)


lizmat:
I guess before we can make this for 6.e only, we would need to have a way to find out what the caller's language level is, right?

vrurg:
No. There must be a way to provide a "patched" version of class for each language version. I don't see a reason why same core object would behave differently being used in different context? Because if we are inclined to depend on the context then where is the boundary where we stop and say: this is the version assumed by the user! Say, I use a module which is 6.c/d and pass a list into it. I know how the list behaves for me in 6.e, but then I get back the results I don't expect! Why? Because I don't know what language the module is using and I am not supposed to know it.

Instead, if I need 6.c/d behavior then I must be able to create a list object using corresponding class version. I.e. `CORE::v6c::List.new(...)`. This is, again, more of a subject for `problem-solving`, of course. But I was already solving this problem for `PseudoStash` and here is what I came up with:

```raku
my class PseudoStash is CORE::v6c::PseudoStash {
```

The problem with this approach is in accessing private attributes and methods of the parent class. I was thinking about the subject lately and see the proper solution in either allowing `augment` to override the existing methods, or in introducing a new keyword for the purpose of not extending but changing the behavior of a class.

lizmat:
reopened to preserve the idea

PR #4974, opened 2022-06-28
Add method 'inode'

vrurg:
The syscall is already supported by nqp, no extra efforts needed.

### DIFF
--- a/src/core.c/IO/Path.pm6
+++ b/src/core.c/IO/Path.pm6
@@ -847,6 +847,18 @@ my class IO::Path is Cool does IO {
           !! self!does-not-exist("inode")
     }
 
+    method dev(IO::Path:D: --> Int:D) {
+        Rakudo::Internals.FILETEST-E(self.absolute)  # sets $!os-path
+          ?? nqp::stat($!os-path, nqp::const::STAT_PLATFORM_DEV)
+          !! self!does-not-exist("inode")
+    }
+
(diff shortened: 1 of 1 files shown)


lizmat (approved):

jonathanstowe:
While you're in the mood, any chance of adding the `device` too (similarly already implemented in nqp,) I can rid my own implementations in `Linux::Fuser`  then :rofl:

vrurg:
@jonathanstowe Not sure which one do you exactly mean, `st_dev` or `st_rdev`, so added both. :)

jonathanstowe:
> which one do you exactly mean

`st_dev` - both `inode` and `st_dev` need to be compared to determined whether two "files" are actually the same thing.

Cheers!

lizmat:
Hmmm... then maybe also throw in an `infix:<eqv>` candidate for `IO::Path` ?

vrurg:
@lizmat one at a time. :) Though I'm not really sure about the semantics of `eqv`. It's somewhat fuzzy when it comes to options like symlinks, relative paths, etc. Apparently, inode and dev are the ultimate answer, but I'd rather consider options first.

Xliff:
Any idea when these will be merged into master? I already have a use case. :)

vrurg:
@Xliff I have certain doubts about the naming of `dev` method. Would like to give it another day to mull about, perhaps.

PR #4976, opened 2022-07-01
Implement Haskell's "span" as "snip" in 6.e

lizmat:
Inspired by https://stackoverflow.com/questions/72809469/in-raku-how-does-one-write-the-equivalent-of-haskells-span-function

This implements both the method as well as the sub version of "span".

For more flexibility, the condition can also be a List of conditions
(in the sub case), or just have multiple conditions (in the method case).

    .say for span * < 10, 2, 2, 2, 5, 5, 7, 13, 9, 6, 2, 20, 4;
    (2 2 2 5 5 7)
    (13 9 6 2 20 4)

    .say for span (* < 10, * < 20), 2, 2, 2, 5, 5, 7, 13, 9, 6, 2, 20, 4;
    (2 2 2 5 5 7)
    (13 9 6 2)
    (20 4)

    .say for (2, 2, 2, 5, 5, 7, 13, 9, 6, 2, 20, 4).span( * < 5, * < 10 );
    (2 2 2)
    (5 5 7)
    (13 9 6 2 20 4)

Span documentation: https://hackage.haskell.org/package/base-4.16.1.0/docs/Prelude.html#v:span


jnthn:
Functionality wise, it seems like a useful addition. I'm not really sure about the name `span`; none of the meanings of the verb "to span" seem to really describe what this does. Maybe something more like "partition"?

lizmat:
How about "snip" ?

TimToady:
I like it.  It's cute, but I don't think it falls into the category of "You
think that's cute today."

Larry

jnthn:
> How about "snip" ?

It kind of works, but to me at least "snip" has a connotation that the snipped off part is being thrown away, which isn't really the case here.

uzluisf:
@tbrowder If the function is to reflect the behavior of [Haskell's `span`](https://stackoverflow.com/a/59343451/10824322), then `partition`/`parts`/`part` aren't good names since there's already a [Haskell's `partition`](https://stackoverflow.com/a/59343463/10824322) that does something slightly different.

tbrowder:
Well, if the object is to follow Haskell, then we should use the same names; otherwise, choose a name more suitable to the Raku domain.

lizmat:
> Well, if the object is to follow Haskell, then we should use the same names; otherwise, choose a name more suitable to the Raku domain.

Which is exactly why I went for snip.  Also, there is one important difference with the Haskell implementation (afaik): `snip` can take multiple conditions, and snip any list into multiple lists (rather than just two in Haskell).

> It kind of works, but to me at least "snip" has a connotation that the snipped off part is being thrown away, which isn't really the case here.

https://www.reddit.com/r/rakulang/comments/vqlxij/how_can_i_improve_this_for_loop_so_it_doesnt_use/ieq1rnc/ shows an application where `snip` is **exactly** used in that way, in my opinion  :-)

lizmat:
>I suggest lop. It's pretty much a synonym of snip, however it could be arguably a portmanteau of "longest prefix" (emphasis mine):

The thing is that `snip` allows for multiple conditions, snipping off multiple parts of a list.  Afaik, `lop` is only used to cut off a single thing.

TimToady:
I don't like lop because that definitely implies throwing away some of it.
Lopping is what you do when you're pruning a tree.  I don't think snip has
as much throw-away in my mind, because snipping DNA in the middle and then
reattaching it is what CRISPR-CAS9 does.  I would say "snip off" if I meant
the other thing.  Also there's some subtle mnemonic association with SNIPs,
which is what your DNA tests look for, albeit without actually snipping
anything.

Larry

PR #4977, opened 2022-07-03
Implement coercion via new-disp and re-consider return typechecking dispatcher

vrurg:
All cases covered except for `Type1(Type2())` where recursive dispatching is required.

As to the return typechecking, the following is now valid:

```raku
    subset SmallValue of Rat:D where *.abs < 0.5;
    sub small-value($v --> SmallValue()) { $x }
    say small-value("0.2"); # 0.2
    small-value("1.3"); # will throw
```

### DIFF
--- a/src/Perl6/Actions.nqp
+++ b/src/Perl6/Actions.nqp
@@ -9674,6 +9674,14 @@ class Perl6::Actions is HLL::Actions does STDActions {
                             QAST::Op.new(
                                 :op('bind'),
                                 QAST::Var.new(:name($name), :scope('local')),
+#?if moar
+                                QAST::Op.new(
+                                    :op<dispatch>,
+                                    QAST::SVal.new(:value<raku-coercion>),
+                                    QAST::Var.new(:name($low_param_type), :scope<local>),
+                                    QAST::Var.new(:name($name), :scope<local>))))));
+#?endif
+#?if !moar
                                 QAST::Op.new(
                                     :op('callmethod'),
                                     :name('coerce'),
@@ -9682,6 +9690,7 @@ class Perl6::Actions is HLL::Actions does STDActions {
                                         QAST::Var.new(:name($low_param_type), :scope('local'))),
                                     QAST::Var.new(:name($low_param_type), :scope('local')),
                                     QAST::Var.new(:name($name), :scope('local')))))));
+#?endif
             }
             elsif nqp::can($ptype_archetypes, 'coercive') && $ptype_archetypes.coercive {
                 $decont_name_invalid := 1;
@@ -9700,12 +9709,21 @@ class Perl6::Actions is HLL::Actions does STDActions {
                         QAST::Op.new(
                             :op('bind'),
                             QAST::Var.new( :name($name), :scope('local') ),
+#?if moar
+                            QAST::Op.new(
+                                :op<dispatch>,
+                                QAST::SVal.new(:value<raku-coercion>),
+                                QAST::WVal.new(:value($param_type)),
+                                QAST::Var.new(:name($name), :scope<local>)))));
+#?endif
+#?if !moar
                             QAST::Op.new(
                                 :op('callmethod'),
                                 :name('coerce'),
                                 QAST::WVal.new(:value($param_type.HOW)),
                                 QAST::WVal.new(:value($param_type)),
                                 QAST::Var.new( :name($name), :scope('local') )))));
+#?endif
             }
 
             # If it's optional, do any default handling.
--- a/src/Perl6/Metamodel/CoercionHOW.nqp
+++ b/src/Perl6/Metamodel/CoercionHOW.nqp
@@ -112,30 +112,35 @@ class Perl6::Metamodel::CoercionHOW
 
     # Coercion protocol method.
     method coerce($obj, $value) {
+#?if moar
+        nqp::dispatch('raku-coercion', $obj, $value)
+#?endif
+#?if !moar
         nqp::istype($value, $!target_type)
           ?? $value                             # already done
           !! self."!coerce_TargetType"($obj, $value)
+#?endif
     }
 
     # Attempt coercion on TargetType
     method !coerce_TargetType($obj, $value) {
-        my $constraint    := $!constraint_type;
-        my $constraintHOW := $constraint.HOW;
-        $value := $constraintHOW.coerce($constraint, $value)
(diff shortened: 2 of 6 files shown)


vrurg:
# Benchmarks Methodology

The script used for benchmarking is provided below. It outputs CSV-like output which can be later pulled into a spreadsheet. The benchmark first warms up the VM, then does the actual test. Each series includes:

1. Bare loop speed to provide the impression of how much the whole coercion takes minus the loop itself and variable assignment.
2. Coercion via `$obj.Type`
3. Via `COERCE` method
4. Via `new` method

For each of 2,3,4 items two times are measured: the actual timing of the loop, including the assignment and loop times; and the pure time the coercion protocol takes, excluding the loop, the assignment, and even the time corresponding coercion method takes. I.e. it is the pure time of the coercion infrastructure code, user code excluded.

Series are then repeated five times. After importing the data into a spreadsheet I take the minimal time for each measurement based on the idea that anything above that are overheads of the system and the language core like GC kicking in or alike.

```raku
class CNew {
    has $.val;
    multi method new(Int:D $val) {
        samewith :$val;
    }
}

class CCOERCE {
    has $.val;
    multi method COERCE(Int:D $val) {
        self.new :$val;
    }
}

class CType {
    has $.val;
    method Int { $.val }
}

constant WARM-UPS = 10000;
constant REPETITIONS = 100000;

sub bench-type($warmup = False) {
    my $reps = $warmup ?? WARM-UPS !! REPETITIONS;

    my $c = CType.new(:val(11));

    my $dst = now;
    for ^$reps {
        my $int = $c.Int;
    }
    my $dtime = now - $dst;

    my $cst = now;
    for ^$reps {
        my $int = Int($c);
    }
    my $ctime = now - $cst;

    $ctime, ($ctime - $dtime)
}

sub bench-COERCE($warmup = False) {
    my $reps = $warmup ?? WARM-UPS !! REPETITIONS;
    my $dst = now;
    for ^$reps {
        my $c = CCOERCE.COERCE($_);
    }
    my $dtime = now - $dst;

    my $cst = now;
    for ^$reps {
        my $c = CCOERCE($_);
    }
    my $ctime = now - $cst;

    $ctime, ($ctime - $dtime)
}

sub bench-new($warmup = False) {
    my $reps = $warmup ?? WARM-UPS !! REPETITIONS;
    my $dst = now;
    for ^$reps {
        my $c = CNew.new($_);
    }
    my $dtime = now - $dst;

    my $cst = now;
    for ^$reps {
        my $c = CNew($_)
    }
    my $ctime = now - $cst;

    $ctime, ($ctime - $dtime)
}

sub bench-loop($warmup = False) {
    my $reps = $warmup ?? WARM-UPS !! REPETITIONS;
    my $st = now;
    for ^$reps {
        my $v = $_;
    }
    my $ftime = now - $st;
}

my @h = ('Bare Loop', '.Type Loop', '.Type Pure', '.COERCE Loop', '.COERCE Pure', '.new Loop', '.new Pure');
say @h.map(*.raku).join(", ");

for ^5 {
    my @res;
    bench-loop(True);
    @res.push: bench-loop;

    bench-type(True);
    @res.append: bench-type;

    bench-COERCE(True);
    @res.append: bench-COERCE;

    bench-new(True);
    @res.append: bench-new;
    say @res.join(", ");
}
```

vrurg:
# Benchmark Results

`master` here is the latest master branch with @lizmat optimizations. Columns with `Loop` are full timing of loops. `Pure` correspond to the pure timing of the coercion infrastructure code.

What I can say: new-disp is still amazing me!

| Compared To | .Type Loop | .Type Pure | .COERCE Loop | .COERCE Pure | .new Loop | .new Pure |
| --- | --- | --- | --- | --- | --- | --- |
| **2022.06**     |      1.97 | 2.45 | 6.39 | 15.36 | 2.76 | 13.13 |
| **master**      | 1.24 | 1.36 | 4.41 | 9.99 | 2.60 | 11.46 |

Here are results for @lizmat optimizations too, based on the same set of data:

| Compared To | ".Type Loop" | ".Type Pure" | ".COERCE Loop" | ".COERCE Pure" | ".new Loop" | ".new Pure" |
| -- | -- | -- | -- | -- | -- | -- |
| **2022.06** | 1.59 | 1.80 | 1.45 | 1.54 | 1.06 | 1.15 |

The interesting observation here is about `COERCE` and `new`. Since these are the least used coercion methods they're considered later in the original coercion code in metamodel. Correspondingly, it takes most time to invoke them, what is clearly visible from the second table. Hence, the numbers in the first table where _new-disp_ allows to bypass all the other checks and invoke the methods nearly immediately thus sparing us a **lot** of CPU cycles. 

I could expect the timing to be even better for `new` since the coercion code invoking it has to use `try`/`CATCH` in order to intercept `X::Constructor::Positional` and convert it into `no coercion method found` error. Thus, the thunk becomes un-inlinable. Consequently, `new` is still 2.3 times slower than `COERCE` (BTW, it is 2.66 times slower on _master_).

But because `new` is barely used for this purpose, and because `COERCE` is supposed to be the primary method of indirect coercion, I think there is no need in getting rid of the exception handling.

vrurg:
I have it as a draft for now because if time allows I'd like to finish @jnthn work on return values where currently the only allowed method of coercion is `.Type`.

vrurg:
Work done and ready for merge.

vrurg:
The latter changes include special case of a "single" new, I.e. for a case when only one `new` method candidate matches the value we're about to coerce, and it is not from `Mu` (in which case it is "no coercion method found" error). When we know the candidate is the one to be used for the actual coercion, it is possible to avoid  `try`/`CATCH` in the coercer code. With this in mind I have update the benchmarking script ([see this gist](https://gist.github.com/vrurg/711f43e466f0fa3c88e2fef66139fa35)) and the following table are the results:



Compared To | .Type Loop | .Type Pure | .COERCE Loop | .COERCE Pure | .new Multi Loop | .new Multi Pure | .new Single Loop | .new Single Pure | Assign Loop | Assign Pure
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
**2022.06** | 1.91 | 2.40 | 5.95 | 21.13 | 2.65 | 17.26 | 8.59 | 40.09 | 3.21 | 3.63
**master** | 1.15 | 1.27 | 3.80 | 12.68 | 2.31 | 14.57 | 7.33 | 33.66 | 2.22 | 2.46

PR #4978, opened 2022-07-03
Add a help function to the repl

lizmat:
Which is basically a call to a "help" multi sub.  Which can be set
externally with a &*HELP dynamic variable.

This is a concept: the texts need to be fleshed out.  The basic idea
is that you can enter "help" at the REPL prompt to get some help
page.  And "help foo" to get help on the "foo" subject.

Just entering "help" or "help foo" where "foo" is not recognized,
should list all available functions.

This functions by compiling in a "help" sub in any code that is being
evaluated.  Any "help foo" string is converted to "help 'foo'" to allow
it being dispatched correctly.

Suggestions for texts / features are welcome.

### DIFF
--- a/src/core.c/REPL.pm6
+++ b/src/core.c/REPL.pm6
@@ -1,6 +1,32 @@
 class REPL { ... }
 
 do {
+    my $options := q:to/OPTIONS/;
+-------------------------------------------------------------
+edit     information on how to edit current or previous lines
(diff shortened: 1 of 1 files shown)


lizmat:
This has been superseded by the work on [`REPL`](https://raku.land/zef:lizmat/REPL)

PR #4981, opened 2022-07-11
Fix a deep recursion issue in return typecheck dispatcher

### LINKED ISSUE #4980: 'subset' lacking restriction can hang program
## The Problem
Certain <code>subset</code> definitions can cause a program to hang (and consume huge amounts of memory).

## Steps to Reproduce
As written, this code never returns from the subroutine call (add the <code>where</code> restriction and things are fine).
```
subset Boo of Int; # where * > 0;  
sub foo() returns Boo { 42 }
say foo();
```
## Environment
First noticed the issue with build v2022.06-46-g934f9895e, on macOS and Ubuntu.

## Related?
With the same Raku build, Rosetta-code programs using the HTTP::Useragent module started to exhibit the same behavior.

vrurg:
Should have used `wrapee-lookup` on SubsetHOW itself.

Fixes #4980

### DIFF
--- a/src/vm/moar/dispatchers.nqp
+++ b/src/vm/moar/dispatchers.nqp
@@ -3928,9 +3928,9 @@ nqp::dispatch('boot-syscall', 'dispatcher-register', 'raku-isinvokable', -> $cap
         else {
             my sub runtime-only($t) {
                 return 0 if nqp::isnull($t);
-                nqp::isconcrete(nqp::how($t).refinement($t)) 
-                    || nqp::how(my $refinee := nqp::how($t).refinee($t)).archetypes.nominalizable
-                        && runtime-only(nqp::how($refinee).wrappee-lookup($refinee, :subset))
+                nqp::isconcrete(nqp::how_nd($t).refinement($t)) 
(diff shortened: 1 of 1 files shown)


src/vm/moar/dispatchers.nqp
```diff
@@ -3929,7 +3929,8 @@ nqp::dispatch('boot-syscall', 'dispatcher-register', 'raku-isinvokable', -> $cap
             my sub runtime-only($t) {
                 return 0 if nqp::isnull($t);
                 nqp::isconcrete(nqp::how($t).refinement($t)) 
-                    || (nqp::how($t).archetypes.nominalizable && runtime-only(nqp::how($t).wrappee-lookup($t, :subset)))
+                    || nqp::how(my $refinee := nqp::how($t).refinee($t)).archetypes.nominalizable
```
niner, line 3932: I think this would be easier to read if you spell it `(my $refinee := $t.HOW.refinee($t)).HOW.archetypes.nominalizable`

niner (approved):

src/vm/moar/dispatchers.nqp
vrurg: BTW, it makes sense to use `nqp::how_nd` here.

PR #4984, opened 2022-07-15
Introduce "nano" term for 6.e+

lizmat:
Returns nanoseconds since epoch, to allow HLL access to more accurate
timing.

Specifically did not include that for 6.c.  Unfortunately, it's pretty
hard to add support in the grammar for a nano term in 6.e only, so
the older settings create a runtime error.  Suggestions for fixing this
deficiency are welcome.

### DIFF
--- a/src/Perl6/Actions.nqp
+++ b/src/Perl6/Actions.nqp
@@ -6324,6 +6324,10 @@ class Perl6::Actions is HLL::Actions does STDActions {
         make QAST::Op.new( :op('call'), :name('&term:<time>'), :node($/) );
     }
 
+    method term:sym<nano>($/) {
+        make QAST::Op.new( :op('call'), :name('&term:<nano>'), :node($/) );
+    }
+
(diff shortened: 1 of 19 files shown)


vrurg:
You, actually, did it before:

```
        || <?{
              nqp::getcomp('Raku').language_version eq '6.c'
            || $*WHENEVER_COUNT >= 0
          }>
```

Line 1193 in Grammar.nqp. Exactly the way I was about to suggest (except for the `$*WHENEVER_COUNTER` part).

lizmat:
Cool!

eh, but wait?

Ah!  Well, that was part of b787095358 and that was just a `s/perl6/Raku`  I could hardly take credit for that or remember the exact semantics of that!   :-)

vrurg:
For the moment the best is to `<?{ nqp::getcomp('Raku').language_revision ge 'e' }>`. This is if you want `nano` to be totally invisible to 6.c/d and cause a syntax error of a kind.

PR #4987, opened 2022-07-18
Add IO::Path.under-version-control method in 6.e

lizmat:
Returns True if the invocant appears to be under version control.
This is currently defined as it having a sibling ".git" directory,
or any of its parents having a ".git" directory.

Can be handy for CLI scripts before they do massive changes on a
file system: being under version control provides a bigger chance of
reversibility if something went wrong.

### DIFF
--- a/src/core.e/IO/Path.pm6
+++ b/src/core.e/IO/Path.pm6
@@ -1,0 +1,12 @@
+augment class IO::Path {
+    method under-version-control(IO::Path:D: --> Bool:D) {
+        my $io = self.resolve;
+        (my $parent := $io.parent.resolve) eqv $io
+          ?? (return False)
+          !! ($io = $parent)
+          until $io.add(".git").d;
+        True
+    }
+}
+
+# vim: expandtab shiftwidth=4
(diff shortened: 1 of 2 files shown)


Kaiepi:
A Mercurial repo should be inferrable in a similar way to git, but other version control schemes (e.g. SVN, CVS) might be more difficult.

niner:
Honestly, this seems to be too much of a fringe feature to land in core. That - despite git's ubiquitousness - version control systems are a moving target also doesn't fit well with the 100 year language. And even if git is the omega of version control systems, that doesn't mean that it will always work by putting .git directories somewhere. Such things are better left for module space instead of a language specification that will be forever set in stone.

lizmat:
> version control systems are a moving target also doesn't fit well with the 100 year language

So is unicode, but we (try to) follow that as well.  Because it is a selling point for Raku.

There's a reason why it is named "under-version-control" without mentioning git.  It can thus evolve with whatever version control systems might come.  That's why it said: "This is currently defined as" in the commit message.  In that sense, offering a stable API for this functionality belongs very much in a 100 year language in my opinion :-)

In any case, I've also uploaded it to module space: https://raku.land/zef:lizmat/under-version-control (visible in about 30 minutes).

PR #4988, opened 2022-07-19
Set a dynamic variable if exiting program without exception

lizmat:
Inspired by https://stackoverflow.com/questions/73040525/how-to-know-in-a-end-block-if-the-programm-died

This minimal change allows an END block to do special actions if the
program exited because of an exception:

    END {
        unless $*DIED-NATURALLY {
            # do special cleanup actions
        }
    }

### DIFF
--- a/src/main.nqp
+++ b/src/main.nqp
@@ -69,6 +69,7 @@ sub MAIN(*@ARGS) {
     $comp.command_line(@ARGS, :encoding('utf8'), :transcode('ascii iso-8859-1'), |%defaults);
 
     # do all the necessary actions at the end, if any
+    my $*DIED-NATURALLY := 1;
     if nqp::gethllsym('Raku', '&THE_END') -> $THE_END {
         $THE_END()
     }
--- a/src/main.nqp
+++ b/src/main.nqp
@@ -69,7 +69,7 @@ sub MAIN(*@ARGS) {
     $comp.command_line(@ARGS, :encoding('utf8'), :transcode('ascii iso-8859-1'), |%defaults);
 
     # do all the necessary actions at the end, if any
-    my $*DIED-NATURALLY := 1;
+    my $*DIED := 1;
     if nqp::gethllsym('Raku', '&THE_END') -> $THE_END {
         $THE_END()
     }
--- a/src/main.nqp
+++ b/src/main.nqp
@@ -69,7 +69,7 @@ sub MAIN(*@ARGS) {
     $comp.command_line(@ARGS, :encoding('utf8'), :transcode('ascii iso-8859-1'), |%defaults);
 
     # do all the necessary actions at the end, if any
-    my $*DIED := 1;
+    my $*DONE := 1;
     if nqp::gethllsym('Raku', '&THE_END') -> $THE_END {
         $THE_END()
     }


vrurg:
May be `$*EXITED-NATURALLY` or alike? Because when `DIED` I personally expect that `die` is involved somehow.

jnthn:
Could even just be `$*DIED`.

vrurg:
@lizmat I'm still in doubt because not very well familiar with the main workflow. Is it set when exception is the cause or when all is done normally?

lizmat:
Yuck.  Yes, I got it the wrong way around.  Now it is set when it did **NOT** do a `die`.

Perhaps `$*DONE` would be a better name?  Or maybe revert to the `$*DIED-NATURALLY` ?

vrurg:
`$*WELL-DONE` for that matter. ;)

Lets do what I do sometimes to make things clear: be banal. We'd like to tell a user that no exception aborted the code execution. Then `$*NO-EXCEPTION` is, perhaps, the most straightforward?

```raku
END {
    say "Oh, we're good!" if $*NO-EXCEPTION;
}
```

Could be `$*NOT-DIED`, but generally speaking, use of `die` obscures the fact that an exception is thrown for some inexperienced devs.

Altai-man:
>`Perhaps $*DONE would be a better name? Or maybe revert to the $*DIED-NATURALLY ?`

+1 for `$*DONE` or `$*EXITED-something`, `$*DIED-NATURALLY` sounds rather unpleasant for "the program has finished its execution".

vrurg (approved):

jnthn:
But `$*DIED` matches up with the function `die`, whereas there's a function `done` that `$*DONE` has nothing to do with.

lizmat:
But then it would be more troublesome to implement.

Now, if the dynamic variable is not set, it is effectively `False`.  To get the correct meaning with `$*DIED`, it would have to be `False`, which would be indistinctive from not being set.

So that would mean we would always have to set `$*DIED` to `True` for every run?  But that'd be weird?

jnthn:
Ah, I'd assumed from the original naming that it *was* linked to an exceptional exit (with a non-zero exit code not being considered exceptional); I'd not looked at the implementation. I'd also assumed the dynamic variable was being set by the thing that invokes `END` blocks, so it's only visible inside of those (that is, inside the `THE_END` implementation).

If it's about exit code, I'm wondering if a better approach wouldn't be to invoke `THE_END` with the exit code, which can set a `$*EXITCODE` or similar.

lizmat:
@jnthn This is the implementation: https://github.com/rakudo/rakudo/pull/4988/files#diff-77c41e9c8d6874981e5e4ab905cf511c0a51320a5db3ac6fea26093b079f433bR72

It literally is one line.

jnthn:
@lizmat Yes, I read it before writing my latest comment, and wonder if the approach is too simplistic; for example in [the question that I assume triggered this PR](https://stackoverflow.com/questions/73040525/how-to-know-in-a-end-block-whether-the-program-died) then `$?` is set to the exit code:

```
$ perl -e 'END { print $? }; exit 42'
42
```

It'd seem nice to have Raku be equally capable. (A bonus, but probably as a separate PR, would to be able to obtain the exception that the program died with.)

patrickbkr:
> **Upd** Hm, it seems I _do_ like `$*WELL-DONE`! But in the light of terminology discussion in problem solving, that could be an inappropriate choice.

Or a very appropriate choice, depending on the outcome of that discussion. :-)

lizmat:
Closing in favour of https://github.com/rakudo/rakudo/pull/5014

PR #4993, opened 2022-07-21
Implement nominalizable transparency for definites

vrurg:
The following will now produce correct true output:

```raku
    subset Foo of Int();
    say Foo:D.^archetypes.coercive;
```

Also note the use of `.^archetypes` notation.

The target is achieved by parameterizing definites with an additional argument and adding optional `$obj` positional parameter, akin to many other user-land accessible methods invoked via `.^` notation.

The additional parameterization argument is an index built from definite's base type `coercive` and `generic` values by considering them as bits and doiing `($coercive +< 1) +| $generic`. The index is then pulled from the typeobject parameterization arguments and used to choose the right archetypes instance from a list of predefined
`Perl6::Metamodel::Archetype` instances.

In particular, this commit fixes a case like the following:

```raku
    role R[::T, $default] {
        has T:D $.a = $default;
    }
    class C does R[Int, 42] { }
```

Prior this was failing due to `T:D` was not instantiated a was failing type check.

Many other cases of using definite generics are supposedely to be fixed
too.

### DIFF
--- a/src/Perl6/Actions.nqp
+++ b/src/Perl6/Actions.nqp
@@ -3588,7 +3588,7 @@ class Perl6::Actions is HLL::Actions does STDActions {
                     }
                 }
                 elsif $past.ann('metaattr') -> $attr {
-                    if !$attr.required && !$attr.type.HOW.archetypes.generic {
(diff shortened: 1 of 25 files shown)


vrurg:
This PR requires Raku/nqp#775 to work.

PR #4990, opened 2022-07-20
[WIP] Make some enhancements to Semaphore

Kaiepi:
- Summarized in https://github.com/MoarVM/MoarVM/pull/1714.
- A `Semaphore` boxes a `uint` now...
- ...exposed by `Semaphore.Int`/`Semaphore.Numeric`.
- Passes `t/spec/S17-lowlevel/semaphore.t` on WSL2 OpenSUSE Tumbleweed, but needs more tests evidently.

PR #5000, opened 2022-07-24
Make some enhancements to atomic integers

Kaiepi:
- Introduce the nqp-ish, semaphore-ish, serializable `Counter` class for the MOP's sake.
- Consider giving the JVM and JS backends atomicops.
- Tests need adjustment due to the new symbol.
- Takes the place of https://github.com/rakudo/rakudo/pull/4990.


lizmat:
It would be really nice if someone could continue the work of Kaiepi

@patrickbkr @ab5tract @timo @niner

PR #5005, opened 2022-07-26
Introduce :smartcase named argument

lizmat:
In principle for all string related functions that also have :ignorecase.
In this commit, only "starts-with" has been equipped with it.  Basically
it operates as :ignorecase if the needle does **not** have any uppercase
characters.  Otherwise it is ignored.

    say "Foo".starts-with("foo",:smartcase);  # True
    say "Foo".starts-with("Foo",:smartcase);  # True
    say "foo".starts-with("Foo",:smartcase);  # False

If accepted, this PR will be expanded to cover all appropriate methods.

### DIFF
--- a/src/core.c/Str.pm6
+++ b/src/core.c/Str.pm6
@@ -157,6 +157,19 @@ my class Str does Stringy { # declared in BOOTSTRAP
         self.chop($chopping.Int)
     }
 
+    multi method starts-with(Str:D: Str:D $needle, :$smartcase! --> Bool:D) {
+        self.starts-with(
+          $needle,
+          :ignorecase($smartcase && nqp::iseq_i(
+            nqp::chars($needle),
+            nqp::findcclass(                      #?js: NFG
+              nqp::const::CCLASS_UPPERCASE,
+              $needle,0,nqp::chars($needle)
+            )
+          ))
+        )
+    }
+
     multi method starts-with(Str:D:
       Str:D $needle, :i(:$ignorecase)!, :m(:$ignoremark)
     --> Bool:D) {


patrickbkr:
Hm. I know smartcase only as a feature in Vim where it determines if I want case insensitive matching by the very simple heuristic, whether I don't have any upper-case letters in the search word. (I really like that behavior!) But that can't be the reasoning for the feature here, since one still has to type `:smartcase` to enable the heuristic. Can you explain what the point of this feature is?

lizmat:
The point is that you can specify the option `:smartcase` without having to look at the needle yourself?

patrickbkr:
Ah! Understood.

I wonder how common that use case is, though. My guess would be, that smartcase would almost exclusively be useful when the `Str` function in question represents the core functionality of the piece of software (some kind of search interface where the user provides the search keyword). If that guess is right, then I don't think having a function for that is worth the while.

But I'm only mildly opposed. So if you disagree - so be it. :-)

lizmat:
For clarity, I would also want to apply this to `substr-eq`, `ends-with`, `contains`, `indices`, `index`, `rindex` for complete consistency.

thoughtstream:
Personally, I use this feature all the time...in any search-capable app that provides it.
Specifically, I have it permanently enabled in all my editors, and also in `rg`, `ack`, and `rak`.

I'd love to see it available directly in Raku...so that any future search-capable app 
I happen to write in Raku can also have this convenient feature.

Of course, it's already possible to write:
```raku
    $text.contains($target, :ignorecase($target !~~ /<upper>))
```
but that's not nearly as readable, self-explanatory, or DRY as:
```raku
    $text.contains($target, :smartcase)
```

_(By the way, when I first wrote that previous paragraph, I actually enbugged it, by writing:_
```raku
    $text.contains($target, :ignorecase($text !~~ /<upper>))
```
_...which is exactly why we advocate DRY coding in the first place, 
and why `:smartcase` is a worthy addition. :-)_

niner:
> ```
> but that's not nearly as readable, self-explanatory, or DRY as:
> ```raku
>     $text.contains($target, :smartcase)
> ```

FWIW "smartcase" is only self-explanatory, if you happen to know the term 
already. When I duckduckgo that term, I find heaps of results, all relating to 
luggage or smart phone protectives and somewhere in there is one entry about a 
vim feature.

Kaiepi:
If the lowercase-only behaviour can be extended to uppercase-only as well, maybe `:intercase` would describe it? If the case is mixed, there are no DWIMmy alternate cases to consider.

thoughtstream:
@niner observed:

> FWIW "smartcase" is only self-explanatory, if you happen to know the term already.

Yes. But that's kinda how ***all*** words work. ;-)

> When I duckduckgo that term, I find heaps of results, all relating to luggage 
> or smart phone protectives and somewhere in there is one entry about a vim feature.

But if you encountered the term as an argument to a search-related Raku method,
which of those topics would seem most likely to be relevant? Especially given Raku's
_"We steal tech from everyone"_ approach?

And wouldn't you duckduckgo something more like **"Raku smartcase"**, rather than just **"smartcase"**?
In which case DDG would presumably lead you straight to the awesome documentation
that will naturally accompany the PR. :-)

For example, when I do the [equivalent DDG search](https://duckduckgo.com/?t=ffab&q=Raku+ignorecase&ia=web) for an existing adverb like `:ignorecase`, 
the top five results are all pointers into the relevant sections of docs.raku.org.

niner:
> Yes. But that's kinda how ***all*** words work. ;-)

Words yes, but not terms. I don't have to look up what "ignorecase" means, 
because from the context of string matching and the inherent meaning of the 
words "ignore" and "case" I can easily infer what it does. I don't need to 
consult the documentation to understand what the code does here. And indeed, 
if I were to write the docs on this I wouldn't even know what else to write, 
besides putting those words into a whole sentence.

With "smartcase" I would absolutely have to read the documentation. Without 
that I have 0 chance of guessing what it does. What makes case "smart"? No 
idea. None. That's the exact opposite of self-explanatory.

thoughtstream:
> That's the exact opposite of self-explanatory.

Hmmmm. I don't agree with your viewpoint, and I am certainly 
not opposed to Raku adding features that actually require users 
to RTFM. But, in the interest of promoting harmony and agreement,
I am more than happy to withdraw my wanton and unwarranted
characterization of `:smartmatch` as _"self-explanatory"_.

I would, however, maintain that this impenetrably named 
and utterly-incomprehensible-without-documentation facility
...is nevertheless still useful, declarative, DRY, and already available 
(and extensively field-tested) in a significant number 
of other widely used search-related tools and apps.  :-)

vrurg:
This all starts looking like a "war" of vim-users with the rest of the world. ;) BTW, googling on `smartcase matching` provides plenty of resources directly related to the matter.

Speaking of self-explanatory, I can probably find a couple of names in our documentation which barely tell me what are they. `comb`, for example, I cannot use without referring to its description or experimenting first. From that point of view I see no problem in the naming. Especially considering that there is at least one hugely popular piece of software using this term, making it familiar to wast number of devs.

What's more important is that the dispute of usefulness or uselessness of it is seemingly already resolved in favor of the latter. Am I right?

niner:
> This all starts looking like a "war" of vim-users with the rest of the
> world. ;)

Can't be, since every single line of Raku I've ever written was entered into 
vim.

> Speaking of self-explanatory, I can probably find a couple of names in our
> documentation which barely tell me what are they.

Sure. I did neither claim that all other names were nor that this was a 
requirement for new things. I merely contested the description of "smartcase" 
as self-explanatory.

lizmat:
I will expand this to the other applicable string functions and merge then.  Unless someone objects.

2colours:
Seems like nobody objected. :)

For what its worth, I never understood this "smartness" up until a certain point - when "smart" diacritical marks were mentioned that I could suddenly relate to. "futok", "futók", "fűtők", "fűtök" and as some funny nonsense, even "fűtok" could be a meaningful Hungarian word. If you were to ignore all the diacritical marks, you could get horrendous mismatches by accepting any of them as any other - the only sensible overlap is with "futok" which is the one ascii friendly version.  
Sounds like "smartcase" is using exactly the same logic: the default case is the lowercase; if you add an uppercase letter, that's probably an intentional difference from anything except the completely lowercase version, which is kinda like a fallback.

I think the definition is somewhat scary in all software products (iirc I saw it in `diff` for whitespace and grep for casing) but if you find but one situation where it seems useful, you will understand it for good. :)

PR #5008, opened 2022-07-30
Allow for .skip(produce,skip,produce,skip,...) in 6.e

lizmat:
Inspired by https://stackoverflow.com/questions/73175664/whats-an-elegant-way-to-return-a-list-without-the-nth-element

As answer to that question: @b.skip($n,1)

Some examples:

    say (^20).skip(5,3,3,6);     # (0 1 2 3 4 8 9 10 17 18 19)
    say (^20).skip(0,5,3);       # (5 6 7)
    say (^20).skip(|(2,3) xx *); # (0 1 5 6 10 11 15 16)

A * (or no value) on a place for production, means produce the rest.
A * (or no value) on a place for skipping, means stop producing altogether.

### DIFF
--- a/src/core.e/Any-iterable-methods.pm6
+++ b/src/core.e/Any-iterable-methods.pm6
@@ -9,6 +9,13 @@ augment class Any {
     multi method snip(Any:D: *@conditions) {
         Seq.new: Rakudo::Iterator.Snip(@conditions, self.iterator)
     }
+
+    multi method skip(Iterable:D $skips) {
+        Seq.new: Rakudo::Iterator.Skipper: self.iterator, $skips.iterator
+    }
+    multi method skip(*@skips) {
+        self.skip(@skips)
+    }
 }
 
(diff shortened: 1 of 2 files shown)


rassoc:
we are already able to do some of that with `rotor` and `flat`:

```raku
^20 .rotor(5 => 3, 3 => 6, :partial).flat # => (0 1 2 3 4 8 9 10 17 18 19)
^20 .rotor(2 => 3, :partial).flat         # => (0 1 5 6 10 11 15 16)

# can even exclude a number as asked in linked question
^20 .rotor(5 => 1, *).flat                # => (0 1 2 3 4 6 7 8 9 10 11 12 13 14 15 16 17 18 19)
```

lizmat:
@rassoc I don't think the point was that it couldn't be done.  When I look at a question like that, I also think about how it can be done efficiently, both from a programmer's point of view, as well as from an execution point of view.  Hence this PR.

PR #5011, opened 2022-08-01
Allow for --no-foo as alternative to --/foo

lizmat:
By setting the new "allow-no" flag in %*SUB-MAIN-OPTS:

  $ raku -r 'sub MAIN(|c) { say c }; my %*SUB-MAIN-OPTS=:allow-no' --no-foo
  \(:foo(Bool::False))

as opposed to:

  $ raku -r 'sub MAIN(|c) { say c }' --no-foo
  \(:no-foo(Bool::True))

### DIFF
--- a/src/core.c/Main.pm6
+++ b/src/core.c/Main.pm6
@@ -387,6 +387,11 @@ my sub RUN-MAIN(&main, $mainline, :$in-as-argsfiles) {
     my &*ARGS-TO-CAPTURE := &default-args-to-capture;
     my &*GENERATE-USAGE  := &default-generate-usage;
 
+    # Modify args if --no-foo is acceptable as an alternative to --/foo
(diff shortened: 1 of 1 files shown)


jnthn (approved):
Seems safe given it's opt-in, and a common enough convention to offer it out of the box.

Leont:
Over time, the built-in argument parser is getting more and more into line with Getopt::Long 🙂

lizmat:
Is that a documented feature of `Getopt::Long`?  I couldn't find it in the docs, but then again I may have not looked closely enough

Leont:
> Is that a documented feature of `Getopt::Long`? I couldn't find it in the docs, but then again I may have not looked closely enough

All arguments marked as negatable have a `--no-*`. It can be done by giving a Bool a default value or using `is option('!')`

PR #5014, opened 2022-08-02
Introducing $*EXIT and $*EXCEPTION dynvars

lizmat:
Main intent is that these will be used inside an END block to check
whether the program is exiting through an exception or not.

$*EXIT is set to the currently known exit() value, 1 if an exception
occurred, 0 if no exception occurred, N when an exit(N) was executed.

$*EXCEPTION will contain an Exception object if the program is
ending because of an exception.  Will contain the Exception type object
otherwise.

Example:
````
$ raku -e 'END { say $*EXIT; dd $_ with $*EXCEPTION; exit 42 }; die'
Died
  in block <unit> at -e line 1

1
X::AdHoc.new(payload => "Died")
$ echo $?
42

### DIFF
--- a/src/core.c/Exception.pm6
+++ b/src/core.c/Exception.pm6
@@ -522,6 +522,8 @@ do {
     sub print_exception(|) {
         my Mu $ex := nqp::atpos(nqp::p6argvmarray(), 0);
         my $e := EXCEPTION($ex);
+        $*EXIT      = 1;
+        $*EXCEPTION = $e;
 
         if %*ENV<PERL6_EXCEPTIONS_HANDLER> -> $handler {
             my $class := ::("Exceptions::$handler");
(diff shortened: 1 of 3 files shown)


jnthn (approved):
Much prefer this API to the previous proposal in the area. :+1:

vrurg (approved):

FCO:
If you set a new value to $*EXIT, will it change the exit code?

lizmat:
No, if you want to force a new exit code, you will have to call `exit` yourself inside the `END` block.  And that will then only work if you weren't exiting by a call to `exit` already, as only the **first** time you call `exit` in a process, will actually set the value.

patrickbkr:
> No, if you want to force a new exit code, you will have to call `exit` yourself inside the `END` block. And that will then only work if you weren't exiting by a call to `exit` already, as only the **first** time you call `exit` in a process, will actually set the value.

I think this should be documented.

lizmat:
Implemented with https://github.com/rakudo/rakudo/commit/500508664f without needing a noise merge

PR #5016, opened 2022-08-04
[JVM] Tune Java in response to nqp's (de)serialization rewrite

Kaiepi:
i.e. https://github.com/Raku/nqp/pull/777. Untested with a modern JDK, but should be doable with JDK 11.

### DIFF
--- a/tools/templates/jvm/Makefile.in
+++ b/tools/templates/jvm/Makefile.in
@@ -4,7 +4,7 @@ JAR     = jar
 
 NQP_PREFIX  = @nqp_prefix@
 @bpv(NQP)@       = @shquot(@j_nqp@)@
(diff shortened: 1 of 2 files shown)


Kaiepi:
`t/03-jvm/01-interop.t` is broken at the moment, probably related to this.

Kaiepi:
The test breakage is unrelated, so I'll deal with it from https://github.com/rakudo/rakudo/pull/5019. There are a bunch of niceties that Java interop can be given, but that's best dealt with another time.

usev6:
I've tested this on my FreeBSD box and I think I've seen somewhat lower build times than usual.

If no-one objects, I'll merge this PR in a few days.

PR #5017, opened 2022-08-04
Extract heap snapshotting logic into separate module

lizmat:


lizmat:
@timo does this still make sense?

PR #5020, opened 2022-08-12
Don't blindly make a Version object for vN

### LINKED ISSUE #3919: Sigil-less variable v1 is interpreted as version constant
## The Problem

When using `v1`, `v2` etc as sigil-less parameters, they are interpreted as version constants.
This behaviour breaks referential transparency.

## Expected Behavior

    > ->  \v1 { v1 }(42)
    42

## Actual Behavior

    > ->  \v1 { v1 }(42)
    v1

With any other name it works as expected 

    > ->  \x1 { x1 }(42)
    42

## Version:

This is Rakudo version 2020.08.2 built on MoarVM version 2020.08
implementing Raku 6.d.

lizmat:
First look up whether there's a lexical symbol for it, and use that
if we found it.  Should fix #3919

### DIFF
--- a/src/Perl6/Grammar.nqp
+++ b/src/Perl6/Grammar.nqp
@@ -732,7 +732,10 @@ grammar Perl6::Grammar is HLL::Grammar does STD {
     }
 
     token version {
-        <?before v\d+\w*> 'v' $<vstr>=[<vnum>+ % '.' '+'?]
+        <?before v\d+\w*
+            [ '.' \d
+              || <!{ $*W.is_lexical(~$/) }> ]>
+        'v' $<vstr>=[<vnum>+ % '.' '+'?]
         <!before '-'|\'> # cheat because of LTM fail
     }
 


vrurg:
Commenting under a commit is not the best place, after all. :)

I just have realized why it didn't work for you. I'd rather wonder why didn't it throw for you because `$scope` in `is_lexically_visible` must be AST of the current innermost block while `$*SCOPE` is a string.

Anyway, the proper solution is the following:

```
    token version {
        <?before v\d+\w*
            [ '.' \d
              || <!{ $*W.is_lexical(~$/) }> ]>
        'v' $<vstr>=[<vnum>+ % '.' '+'?]
        <!before '-'|\'> # cheat because of LTM fail
    }
```

I don't want to open a new PR on this, so just sharing the code. The point here is to prevent parsing a version earlier if:

1. `v<num>` is not followed by  `.<digit>`
2. and is a lexical

This results in the following:

```raku
my \v1 = 42;
say v1.2; # v1.2
say v1+1; # 43
say v1.2.*; # v1.2.*
say v2; # v2
say v2.*; # v2.*
# Errors:
say v1+; # Missing required term after infix
say v1.aa; # No such method 'aa' for invocant of type 'Int'
say v1.*; # Malformed postfix call
```

I somewhat doubt about the last one. But considering that a version is easily created with `Version.new('1.*')`, it should probably remain this way as a safety measure. I.e. since we consider `v1` as a lexical in first place then wherever there is doubt it must be resolved in favor of the lexical. Then it would mean a compile time error where a run time one could take place otherwise.

But if my reasoning has a flaw then the fix is as simple as using `'.' [\d | '*']`.

PR #5021, opened 2022-08-12
Add `IO::Path.created` returning the `Instant` a path was created

lizmat:
Or returns a `Failure` if the path does not exist.  Why this wssn't
done before, I don't quite understand, as the NQP support appears
to exist on all backends.


Leont:
>  Why this wasn't done before, I don't quite understand, as the NQP support appears to exist on all backends.

Probably because that support has only existed for a few years, and no one noticed libuv adding it.

lizmat:
Ah, that'd be a reason  :-)

PR #5023, opened 2022-08-13
Add :between option to Supply.interval

atroxaper:
With :between option we could emit a value every $interval seconds
*between* the end and the start of a tab calls.

### DIFF
--- a/src/core.c/Supply-factories.pm6
+++ b/src/core.c/Supply-factories.pm6
@@ -76,8 +76,75 @@
         method sane(--> True) { }
         method serial(--> True) { }
     }
-    method interval(Supply:U: $interval, $delay = 0, :$scheduler = $*SCHEDULER) {
-        Supply.new(Interval.new(:$interval, :$delay, :$scheduler));
+
+    class IntervalBetween does Tappable {
+        has $!scheduler;
+        has $!interval;
+        has $!delay;
+
+        submethod BUILD(:$!scheduler, :$!interval, :$!delay --> Nil) { }
+
+        my class IntervalCancellation is Cancellation {
+            has $!delegate;
+            has atomicint $!cancelled;
+            has $!lock;
+
+            submethod BUILD(--> Nil) {
+                $!cancelled ⚛= 0;
+                $!lock = Lock.new;
+            }
+
+            method change-delegate($delegate) {
+                $!lock.protect: { $!delegate = $delegate }
+            }
+
+            method cancelled {
+                ⚛$!cancelled > 0
+            }
+
+            method cancel() {
+                $!lock.protect: {
+                    unless self.cancelled {
+                        $!delegate.cancel with $!delegate;
+                        $!cancelled ⚛= 1;
+                    }
+                }
+            }
+        }
+
+        method tap(&emit, &, &, &tap) {
+            my $i = 0;
+            my $lock = Lock::Async.new;
+            my $cancellation = IntervalCancellation.new;
+            my &code = -> {
+                $lock.protect: { emit $i++ };
+                CATCH { $cancellation.cancel if $cancellation }
+            }
+            $lock.protect: {
+                $cancellation.change-delegate: $!scheduler
+                        .cue({ execute(&code, $cancellation, $!interval, $!scheduler) }, :in($!delay));
+                my $t = Tap.new({ $cancellation.cancel });
+                tap($t);
+                $t
+            }
+        }
+
+        my sub execute(&code, $cancellation, $in, $scheduler) {
+            return if $cancellation.cancelled;
+            code();
+            return if $cancellation.cancelled;
+            $cancellation.change-delegate: $scheduler
+                    .cue({ execute(&code, $cancellation, $in, $scheduler) }, :$in);
+        }
+
+        method live(--> False) { }
+        method sane(--> True) { }
+        method serial(--> True) { }
+    }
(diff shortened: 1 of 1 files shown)


Kaiepi:
Unless I'm missing a detail on `Supply`'s portability, it should be OK to use atomics on the JVM and JS backends in the form of an `Int:D $ is default(0)` given the ops you depend on, though `atomic-assign`/`nqp::atomicstore`'s Unicode sugar is missing there at the moment.

src/core.c/Supply-factories.pm6
```diff
@@ -86,11 +86,16 @@
...
             has atomicint $!cancelled;
+#?endif
+#?if !moar
+            has int $!cancelled;
+#?endif
             has $!lock;
 
             submethod BUILD(--> Nil) {
-                $!cancelled ⚛= 0;
+                $!cancelled = 0;
```
Kaiepi, line 98: `$!cancelled`'s value should already be `0`. That's the default for `atomicint` as a native int.

atroxaper:
I'm going to add tests to roast when we agree with naming and changes at all.

atroxaper:
Tests in PR https://github.com/Raku/roast/pull/817

atroxaper:
Is there anybody here? :)

src/core.c/Supply-factories.pm6
```diff
@@ -76,8 +76,79 @@
         method sane(--> True) { }
         method serial(--> True) { }
     }
-    method interval(Supply:U: $interval, $delay = 0, :$scheduler = $*SCHEDULER) {
-        Supply.new(Interval.new(:$interval, :$delay, :$scheduler));
+
+    class IntervalBetween does Tappable {
+        has $!scheduler;
```
lizmat, line 81: Why not `has $!scheduler is built;` and 2 lines below instead of `submethod BUILD` ??

atroxaper: For consistency with Interval class.

lizmat:
> With :between option we could emit a value every $interval seconds between the end and the start of a tab calls.

I'm still not grokking what this is actually adding feature wise.  Could you elaborate with some examples?

src/core.c/Supply-factories.pm6
```diff
@@ -76,8 +76,79 @@
...
+
+        my class IntervalCancellation is Cancellation {
+            has $!delegate;
+#?if moar
+            has atomicint $!cancelled;
+#?endif
+#?if !moar
+            has Int:D $!cancelled is default(0);
+#?endif
+            has $!lock;
```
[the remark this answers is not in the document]
atroxaper: Agreed.

lizmat:
>  Imagine, the app which receives emails and reply on it. Mail server allow us only 15 outbox mails per minute. We create scheduled task, which takes 15 emails and reply on them. Then we need to wait 1 minute and start the task again.

Isn't this also possible with running an interval of 1 minute?  And only send 15 mails per invocation?

> Or we have task which can take variable amount of time, but still want fixed pause between its execution.

You don't need `:between` for that?

Sorry, I seem to be missing something  :-(

lizmat:
> If sending 15 email took 80 seconds, then next task execution will start immediately after the previous end.

But that's just a matter of setting a "busy" flag and not do anything in that period?
````raku
my atomicint $busy;
whenever $timer {
    unless $busy {
        $busy = 1;
        # send up to 15 emails
        $busy = 0;
    }
}

atroxaper:
We send 10 email in 60 seconds, then 5 more for 20 seconds. Next task starts and we send 15 more for the first 40 seconds - fail.

lizmat:
So, are these mails being sent from separate threads?  Or is there a central place that sends the emails?

atroxaper:
Another example from the real life. After each shower the room must be ventilate for 10 minutes. Scheduled task - take a person and escort to the bathroom. Task ends when the person exit. Between task we need 10 minutes.

lizmat:
> Another example from the real life. After each shower the room must be ventilate for 10 minutes. Scheduled task - take a person and escort to the bathroom. Task ends when the person exit. Between task we need 10 minutes.

This also doesn't feel like a job for `.interval`, more for `Thread.queue` or just a `start { sleep 10 * 60; $ok-to-shower = 1 }`

atroxaper:
`Thread.queue` or `start sleep` is good, but we have `interval` instead of using `$*SCHEDULER.cue`.
I'm not a native English speaker, but I think 'interval between tap' sounds good enough. I propose Promise.interval(1, :between).tap(...) 🤷‍♂️

lizmat:
I'm *still* not convinced this is a feature that is needed because there are no other elegant ways to solve what it is intended for.

I've asked @jnthn for a review, I hope he will be able to spend some time on this soon.

vrurg:
My 5 cents. I do see the use for the feature. It is certainly not useless. But is it gonna have wide use? And if it's not then I barely see a reason why a task couldn't end with something like `Promise.in(60).then({ $queue.emit($task) })`

atroxaper:
@vrurg yes, it could be end with Promise.in. But I see it as coupling a logic of a task itself with a logic of the task management. There are many ways to implement ‘between’ logic, but each time we need to invent it.

vrurg:
> But I see it as coupling a logic of a task itself with a logic of the task management.

The latter could be abstracted out into a task manager class or whatever. This is not the point, the point is wether the feature would be actually used massively enough to make sense to have it included into the core.

atroxaper:
Yes, I agree that less the core size is better.

PR #5026, opened 2022-08-15
Introduce `%SUB-MAIN-OPTS<expand-single-letter-options>`

lizmat:
When activated, allows specification of -abc to be equivalent to
--a --b --c


Leont:
This sounds a lot like the existing bundling option.

lizmat:
?  please elaborate?

lizmat:
Ah...  I had completely missed the existence of `%SUB-MAIN-OPTS = :bundling`

TIL   :-)

Issue #5022, opened 2022-08-13
Add 'ask' as an alias for 'prompt'


tbrowder:
## The Problem

I can never remember the Raku routine name for prompting for user input. I and at least one other Rakuun think 'ask' is easier to remember than 'prompt'.

Altai-man:
>I can never remember the Raku routine name for prompting for user input

But a routine for prompting for user input is literally "prompt".

lizmat:
@Altai-man agree

@tbrowder doesn't `ask` always imply a question to which an answer is needed?  What if we're prompting the user to ask a question?  I think `prompt` is more neutral.

Also, if you really want to use `ask`, then:

````raku
my constant &ask = &prompt;
````
anywhere in your code, will do the job.

If you'd like to have a set of aliases, maybe create a module that should just contain:

````raku
sub EXPORT() {
    Map.new: (
        '&ask'  => &prompt,
        '&find' => &grep,
        # etc etc
    )
}
````

tbrowder:
In my scripts, I’m ‘ask’ing for user input, prompting can meant get one’s attention. I can’t think, in my domain at least, when I’m not ‘ask’ing for something.

Altai-man:
Renaming a CORE thing is a very bad practice, and adding aliases is something that probably shouldn't be done too.

As it's easy tweaked for local projects as shown above, I believe this can be closed as a wontfix. Please, re-open if disagree.

tbrowder:
At least one core dev gave a +1 for the idea. I think it appropriate to ask for a vote on \#raku or \#raku-dev or at least get some other opinions. Remember when we changed several routine names not so many years ago, so it's not like the language is carved in stone. 

Also, note recently how newcomers on IRC discuss confusing Raku naming. I love the the idea of adding appropriate aliases for our 100 year language--especially since the common US English grammmar seems to be changing rapidly.

melezhik:
I second `ask` - it's short and it asks for a user input, what exactly happens during prompt.
From other hand I do understand a certain level of tentativeness of brining new names into a core.

tbrowder:
I am about to release a new version of my Text::Utils, which I use often, and it exports alias 'ask' for routine 'prompt' by default.

lizmat:
A shorter name should only be used for something that is used a lot.

Admittedly, that would be true for *some* users of Raku.  But even then, I'd argue that in a CLI script that prompts a user for input, the input taking part is usually the smallest part of the program.

[...]

PR #5028, opened 2022-08-15
Add "ask" as an alias to "prompt" in 6.e

lizmat:
In response to https://github.com/rakudo/rakudo/issues/5022

FWIW, I'm tending to not support this PR myself.  I just created it so that everybody knows what would be needed to make this happen.  And to make sure it is opt-in for users of 6.e.

### DIFF
--- a/src/core.e/control.pm6
+++ b/src/core.e/control.pm6
@@ -1,4 +1,6 @@
 multi sub next(\x --> Nil) { THROW(nqp::const::CONTROL_NEXT, x) }
 multi sub last(\x --> Nil) { THROW(nqp::const::CONTROL_LAST, x) }
 
+my constant &ask = &prompt;
+
 # vim: expandtab shiftwidth=4


Altai-man:
With all the respect to your time, -1 on actually merging this.

lizmat:
@Altai-man making the PR was about 3 mins, no worries if this doesn't get merged :-)

vrurg:
I don't see a need in this too.

fluca1978:
IMHO this is a bad idea: why having two routines that do the same things without any difference? This can make difficult to document and remember in future applications.

melezhik:
> IMHO this is a bad idea: why having two routines that do the same things without any difference? This can make difficult to document and remember in future applications.

maybe because someone might think that "ask" is better name ...

Altai-man:
>maybe because someone might think that "ask" is a better name ...

Someone might also think that "ask" is a worse name, then what?

Raku is maybe one of the most customizable languages out here, so people are free to modify it how it fits them, but locally. There is just no need to make the change available universally for the whole community, making the CORE fatter, introducing more and more new things unless they fix clear bugs.

Altai-man:
Look at how other languages estimate what should go into core/spec - a thing is introduced as a module and then gets widely adopted, so widely that thousands and thousands of devs are relying on it for years, proving it's actually _useful_ to them.

While the scale of the Raku community is a lot smaller, the same sort of approach is probably making sense here to - one can release a package with "better defaults" and see what's the degree of adoption it'll have. My estimate is most people don't really care about names that much, it's a very tiny nuance, how buggy or stable or fast or easy to adopt is much more important, but given it proves to be true - it should probably be adopted regardless of what I think.

More of that, we're always facing a pile of grains sort of problem - at what point a couple of grains turns into a huge pile if you add them one by one? Sure 1 or 2 or even 10 aliases won't hurt, but at what point will it turn into a huge thing you need a reference to work with on a daily basis?

Given we're adding this, we should probably think of some policy for a process of addition.

2colours:
Is this PR still needed? It seems to me that this isn't actually going to get merged and the supplementary [Ask module](https://raku.land/zef:tbrowder/Ask) can be a long-term solution.

PR #5031, opened 2022-08-16
Add .stem method to IO::Path for 6.e

lizmat:
"stem" takes the number of extension levels to remove.
By default all, * and Inf accepted as arguments.

    say "foo/bar.tar.gz".IO.stem;    # bar
    say "foo/bar.tar.gz".IO.stem(1);  # bar.tar

### DIFF
--- a/src/core.e/IO/Path.pm6
+++ b/src/core.e/IO/Path.pm6
@@ -1,0 +1,14 @@
+augment class IO::Path {
+    method stem(IO::Path:D: --> Str:D) {
+        my str $basename = self.basename;
+        my int $index = nqp::index($basename,'.');
+        $index == -1 ?? $basename !! nqp::substr($basename,0,$index)
+    }
+    method suffix(IO::Path:D: --> Str:D) {
+        my str $basename = self.basename;
+        my int $index = nqp::index($basename,'.');
+        $index == -1 ?? '' !! nqp::substr($basename,$index + 1)
+    }
+}
+
+# vim: expandtab shiftwidth=4
--- a/src/core.e/IO/Path.pm6
+++ b/src/core.e/IO/Path.pm6
@@ -1,13 +1,15 @@
 augment class IO::Path {
-    method stem(IO::Path:D: --> Str:D) {
+    method stem(IO::Path:D: $parts = * --> Str:D) {
         my str $basename = self.basename;
-        my int $index = nqp::index($basename,'.');
-        $index == -1 ?? $basename !! nqp::substr($basename,0,$index)
-    }
-    method suffix(IO::Path:D: --> Str:D) {
-        my str $basename = self.basename;
-        my int $index = nqp::index($basename,'.');
-        $index == -1 ?? '' !! nqp::substr($basename,$index + 1)
+        (my @indices := indices($basename, '.'))
+          ?? nqp::substr(
+               $basename,
+               0,
+               nqp::istype($parts,Whatever) || $parts > @indices
+                ?? @indices[0]
+                !! @indices[@indices - $parts]
+             )
+          !! $basename
     }
 }
 
--- a/tools/templates/6.e/core_sources
+++ b/tools/templates/6.e/core_sources
@@ -6,6 +6,7 @@ src/core.e/PseudoStash.pm6
 src/core.e/Grammar.pm6
 src/core.e/EXPORTHOW.pm6
 src/core.e/Int.pm6
+src/core.e/IO/Path.pm6
 src/core.e/array_multislice.pm6
 src/core.e/hash_multislice.pm6
 src/core.e/Rakudo/Internals/Sprintf.pm6


tbrowder:
As another set of names for the basename’s parts, consider:

    say $basename.IO.barename;
    say $basename.IO.extension; 

Thus only one new method need be added.

lizmat:
Are you suggesting s/stem/barename/ ?

Or are you suggesting that `barename` is the part before the first `.` ?

tbrowder:
> Are you suggesting s/stem/barename/ ?
> 
> Or are you suggesting that `barename` is the part before the first `.` ?

Yes.

vrurg:
Does it mean then that for a _file.tar.gz.1_:

```
.barename: file
.stem: file.tar.gz
.stem.stem: file.tar
```

lizmat:
.stem: file
.stem(1): file.tar
.stem(2): file.tar.gz

FWIW, I don't like "barename" as conceptually, people could think "file.tar" to be the barename of "file.tar.gz".  I think `stem` is clearer in that respect.

vrurg:
I would say, that to me the argument is more about skipping from the end, not about leaving at the start. Normally we strip off extra extensions. Otherwise I may need to know how many extensions are there in order to strip the amount I need.

There is a compromise option of `*-$n` argument. Although in my view `$n` must be the number of extensions to be kept, as with the currently propsed `.stem(1): file.tar`. I mean it would be preferable to be implemented as `.stem(*-1): file.tar`.

tbrowder:
> Does it mean then that for a _file.tar.gz.1_:
> .barename: file
> .stem: file.tar.gz
> .stem.stem: file.tar

No. For file '_file.tar.gz.1_', my recommendation now is:

    basename: ''                    # as in current Raku
    barename: _file.tar.gz
    extension

Leont:
I have a hard time seeing the use-case of these particular functions. What is want is the basename as implemented on the unix command line and in Perl: one that can remove a given suffix. Bonus points if it can take a junction of them but a list is acceptable too. `IO::Path`'s `basename` method not doing what any other basename does is rather frustrating to be honest.

Just give me a `method basename(*@suffixes)` and I'd be happy.

jonathanstowe:
> the basename as implemented on the unix command line

This would be my suggestion too.  I guess it could be awesomified to accept matchers as well as strings for the suffixes to remove, but the basic idea seems good.

lizmat:
I'm not clear in which direction this should go now, so closing this PR.  I'd welcome any new PR with a clear definition and path forward (pun intended).

tbrowder:
I'll make one more observation. The one thing that bothers me about the current situation is the ease of determining the extension of a basename as contrasted with getting the basename without the extension. It just seems like there ought to be a simple, default method for doing so, regardless of what we call it. 

In pseudocode:

        my $extension = $basename.extension;
        my $basename'less'extension = $basename.SOMEMETHOD

Each can have more complexities of use, but the default for getting the two pieces of  the basename defined by the last dot in the name should need no arguments.

lizmat:
I think the issue revolves around the question: "what is *the* extension?".  On "foo.tar.gz", is the extension "tar.gz" or "gz" ?

As long as we don't agree on the name, asking for a method that will give the basename "without the extension", will be impossible to give a good answer, let alone an implementation.

tbrowder:
I'm now arguing that the existing 'extension' method is fine. As a default, the extension default should be the last one as it is now.

The new "thingy" for me is two parts:

1. What to name the part of the 'basename' when the default 'extension' is removed?

2. How to implement it as a method with its default behavior is the return the default 'basename' with the default 'extension' (and the joining dot) removed.

You have solved Part 2. We just need to do Part 1.

jonathanstowe:
I think `stem` is probably a reasonable name for 1 TBH.  And I think what was proposed is good for that on the assumption that the suffix is by default everything after the first '.' and if there are special cases with "stacked" suffixes then that can be dealt with too.  It seems to me that the semantics of any particular "compound suffix" are going to be application specific and would need to be dealt with a program that needs to deal with them individually.  After all 'tar.gz' is merely a convention, some software will do '.tgz' and so forth.

tbrowder:
I think I can live with that since 'extension' gives me the **last** part (after the **last** dot) and 'stem' will give me the first part (before the first dot). That suits my usual use case:

1. use File::Find to find files with a certain 'extension' (usually pdf or csv or ofx) for further processing.
2. remove the stem for a rename of extension

masukomi:
finding the "stem" of a file passed into a script is something i've had to do in every language. Yes, it's simple to combine `basename` and `extention` to get, and then strip the extension, but you could just as easily argue that it's trivial to strip off everything after the last path delimiter in order to get basename. 

If the argument against `.stem` is essentially that "we don't need a method for something so trivial" then  what's the argument for `.basename` ? 

> I think the issue revolves around the question: "what is the extension?". On "foo.tar.gz", is the extension "tar.gz" or "gz" ? - @lizmat 

I disagree. I don't think that's a "valid" question _in this context_. 

We've already got a `.extension` method. So the "extension" is, by definition, whatever the `.extension` method returns. So the "stem" is, by definition, the basename minus whatever `.extension` returns. 

If there's debate about what the "correct" extension is, then that should be taken up in a ticket about the `.extension` method. It doesn't have any real bearing on _this_ PR because _this_ functionality doesn't have to weigh in on what the `extension` is. It can simply use the existing definition of extension, and if that changes, then this will stay in sync with that. 

> As long as we don't agree on the name, asking for a method that will give the basename "without the extension", will be impossible to give a good answer, let alone an implementation.

So, what's the historical resolution here? How has Raku decided on what name to use for things? 

I'm voting for "stem" because i haven't heard anything that's _less_ ambiguous. Personally I don't really care. Every language has method names I disagree with. ;)

PR #5033, opened 2022-08-17
Add method Any.snitch in 6.e

lizmat:
This allows one to easily inspect a stage of a method call chain
on STDERR:

    $ raku -e 'use v6.*; (^10).grep(* %% 2).snitch.map(* ** 2).say'
    (0 2 4 6 8)
    (0 4 16 36 64)

### DIFF
--- a/src/core.e/Any-iterable-methods.pm6
+++ b/src/core.e/Any-iterable-methods.pm6
@@ -16,6 +16,8 @@ augment class Any {
     multi method skip(*@skips) {
         self.skip(@skips)
     }
+
+    method snitch() { note self; self }
 }
 
 proto sub rotor(|) {*}
--- a/src/core.e/Any-iterable-methods.pm6
+++ b/src/core.e/Any-iterable-methods.pm6
@@ -17,7 +17,10 @@ augment class Any {
         self.skip(@skips)
     }
 
-    method snitch() { note self; self }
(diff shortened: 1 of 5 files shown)


vrurg (approved):

vrurg:
I would consider adding `is raw` just for full safety of it. Also, `gist` isn't always the best introspection tool. I'd consider having a candidate for manual stringification:

```raku
multi method snitch(&stringify) { note &stringify(self); self }
```

lizmat:
I was actually thinking the same thing, so you could say `foo.bar.snitch(&dd).baz`, so you would specify the whole `Callable`, so you could do anything inside there.

vrurg:
Except the the best point of the method is it outputs automatically. Meaning if I only need to know a single attribute of an object: `$obj.something.snitch(*.foo).do-more`.

lizmat:
Disagree: if you want to do that, why not `$obj.something.snitch(*.foo.note).do-more` ?

vrurg:
Boilerplate detected! This is why many use `dd`, though I never really got used to it.

Anyway, whatever you choose. I'm not so much in favor of implicit `note` to bikeshed about it. :)

lizmat:
Also, this allows you to make your own debugging subroutine, one that could e.g. log to a file, or emit on a Supply (actually, I think `&emit` as parameter, would already do the trick).

I'll leave this for merging *after* the 2022.08 release.

PR #5040, opened 2022-08-22
Introduce // as a definedness prefix operator in 6.e

lizmat:
As suggested by Leon Timmermans in:
  https://twitter.com/leon_timmermans/status/1560416910101172225

This also disables the "null regex not allowed" error in 6.e, because
otherwise the // prefix operator wouldn't get past parsing.

### DIFF
--- a/src/Perl6/Grammar.nqp
+++ b/src/Perl6/Grammar.nqp
@@ -3510,7 +3510,10 @@ grammar Perl6::Grammar is HLL::Grammar does STD {
         ]
     }
 
-    token quote:sym</null/> { '/' \s* '/' <.typed_panic: "X::Syntax::Regex::NullRegex"> }
+    token quote:sym</null/> {
+        <?{ nqp::getcomp('Raku').language_revision lt 'e' }>
+        '/' \s* '/' <.typed_panic: "X::Syntax::Regex::NullRegex">
+    }
     token quote:sym</ />  {
         :my %*RX;
         :my $*INTERPOLATE := 1;
@@ -4096,6 +4099,10 @@ grammar Perl6::Grammar is HLL::Grammar does STD {
         <?before \d+ <?before \. <.?alpha> > <.worry: "Precedence of ^ is looser than method call; please parenthesize"> >?
     }
     token prefix:sym<⚛>   { <sym>  <O(|%symbolic_unary)> }
+    token prefix:sym<//>  {
+        <?{ nqp::getcomp('Raku').language_revision ge 'e' }>
+        <sym> <O(|%symbolic_unary)>
+    }
 
     token infix:sym<*>    { <sym>  <O(|%multiplicative)> }
     token infix:sym<×>    { <sym>  <O(|%multiplicative)> }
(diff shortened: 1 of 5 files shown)


src/Perl6/Grammar.nqp
```diff
@@ -4096,6 +4099,10 @@ grammar Perl6::Grammar is HLL::Grammar does STD {
         <?before \d+ <?before \. <.?alpha> > <.worry: "Precedence of ^ is looser than method call; please parenthesize"> >?
     }
     token prefix:sym<⚛>   { <sym>  <O(|%symbolic_unary)> }
+    token prefix:sym<//>  {
+        <?{ nqp::getcomp('Raku').language_revision ge 'e' }>
+        <sym> <O(|%symbolic_unary)>
```
vrurg, line 4104: I think we gonna be having a problem with `/  /`-kind of regexes. The above `quote:sym</null/>` should probably be kinda conditional with `\s*` for pre-6.e and `\s+` otherwise.

vrurg (changes requested):

src/Perl6/Grammar.nqp
lizmat: I see what you mean.

Fixed with https://github.com/rakudo/rakudo/pull/5040/commits/b50a0d086adb38e1061c369dd54d813f324cda4d  Or do you have a better solution?

vrurg: No, that's exactly what I had in mind. Thanks!

vrurg (approved):

src/core.e/control.pm6
```diff
@@ -1,4 +1,7 @@
 multi sub next(\x --> Nil) { THROW(nqp::const::CONTROL_NEXT, x) }
 multi sub last(\x --> Nil) { THROW(nqp::const::CONTROL_LAST, x) }
 
+proto sub prefix:<//>($) {*}
+multi sub prefix:<//>(\a) { a.defined }
```
jnthn, line 5: Probably this wants to be marked `is PURE`. That way, some syntax accidents can be detected in the case where the language user thought they were writing an infix `//`, but it's actually a useless prefix `//` in sink context. One particular case that comes to mind is:

```
try { foo() }
    // say "oops";
```

The `}` at line end is a statement terminator, so `//` cannot parse as an infix here. Today there's a (not terribly awesome) syntax error raised mentioning the null regex, so such a mistake won't make it past compilation (although might take an effort to understand what happened). Marking the `proto` here with `is PURE` should at least result in a potential difficulty being reported, although those are easier to ignore.

lizmat: So done with https://github.com/rakudo/rakudo/pull/5040/commits/e3abea7d4dff3faf8a0a594929c1bed1684c3789

jubilatious1:
Can this be used with the reduction metaoperator? Something like:

`[//] @a;`

lizmat:
@jubilatious1 Yes.  But not because of this commit, but because there is already an infix:<//> op:

````raku
my @a;
@a[5] = 42;
@a[7] = 666;
dd [//] @a;  # 42
````

PR #5051, opened 2022-08-30
Make precompilation depend distribution modification time

vrurg:
In the particular case of META6.json based distributions this means depend on the file modification time.

For the moment this is mostly an aid to a developer sparing them from remembering to wipe out `.precomp` each time META6.json modification makes this necessary. The particular case I had on my side was automatic preloading of modules depending on their presence in `provides` key. Adding a new module does not always result in recompilation of the code which is actually responsible for preloading. Since `$?DISTRIBUTION` is a compile-time thing, it stucks with an outdated version of `.meta` where `provides` doesn't have the newly added module. 

This change may also prove useful under scenarios we don't have implemented yet. For example, I'm considering enforcing `ver`, `auth`, and `api` keys for `package` entities which are not explicitly using the adverbs. In this case modification of `META6` must result in their recompilation too.


vrurg:
The downside of the current implementation is broken `zef` and, likely, other modules having their own distribution classes. This happens because of a new required method `modified` they have to implement.

This can be workarounded by making the default `modified` method always returning `Instant.from-posix(0)`. Or, may be, a `Failure` since these distributions are unlikely to be used by Rakudo's CUR* to load modules.

vrurg:
**UPD** Replacing `method modified {...}` with `method modified { fail ... }` does make `zef` compile and work. Since, as I said, `Zef::Distribution` wouldn't make into CUR internals, nothing else is expected to break.

ugexe:
Anything that `does Distribution` would break. I have various `Distribution::` modules for instance. The top level `Distribution` role has so few and simple methods by design -- its supposed to be the most bare representation of a distribution and its data and not a store of e.g. system state.

vrurg:
@ugexe But in order for the task to be solved the modification time is necessary. Yet, unless I add the method to the `Distribution` and implement it for other internal distribution classes, alternative solution turns to be rather cumbersome and less logical.

I'd say that having the default method return a `Failure` is the best compromise then.

ugexe:
Why can't the method be implemented just on the `Distribution` class that is used internally, `CompUnit::Repository::Distribution`?

vrurg:
I consider the distribution in more generic way where its source is not only a file system. Say, it could be a centralized network resource. META6.json may not even exists as such. 

Even though the default precompilation repository assumes a file system, there is few ties which can be cut loose to make it source-agnostic.

vrurg:
BTW, thinking a bit more about it, the default `modification` method can simply return epoch and it should be totally safe.

ugexe:
Yes, distribution is supposed to be super generic and is a good reason to not add methods to the interface. As this PR shows almost every implementation of this method is involves a noop, and hence one has to question if it really warrants specification.

As you mention precompilation is hard coupled to the file system, similarly to certain Distribution objects. This is an area designed around that constraint and thus its perfectly fine to make this assumption in this specific area.

ugexe:
> META6.json may not even exists as such.

There are other options, such as hashing the contents of e.g. `$dist.meta.hash` inside of the `id` method I showed earlier.

ugexe:
It also wouldn't surprise me if the solution can be implemented inside of

https://github.com/rakudo/rakudo/blob/71f5c50bed11dc6eb8aacbdc1fe353bd6aa0ba0e/src/core.c/CompUnit/Repository/Distribution.pm6#L33-L37

vrurg:
CURFS doesn't work for a local development directory (i.e. `raku -I.`). It already includes `<provides>`, but its modification doesn't result in re-compilation.

`Distribution::Path` can have the `modified` method on its own simply because it is a specific file system case. The problem is to actually get hold of it because `$*DISTRIBUTION` in `PrecompilationRepository::Default` is a `CompUnit::Repository::Distribution` whose `$.dist` is CURD again, whose `$.dist` is a `Distribution::Path`. I don't know how this chain is built and how long could it be, but sometimes it is terminated with `Distribution::Hash`. So, I would either have to trace down it manually and make sure that I actually can pull out `META6.json` path somehow, or make the method just available for any distribution instance.

Alternatively, I could add `modification` onto CURD which would then be responsible for finding out the value. And for the moment this seems like the best compromise.

PS. When I mentioned on IRC that you'd be objecting it was a bull-eye shot. ;)

vrurg:
> It also wouldn't surprise me if the solution can be implemented inside of

I don't even know where exactly it is used. Will consider.

Speaking about keeping `Distribution` simple, I have just one point. The reason is to have only the essentials in it. But considering that distributions change, a property indicating that could be considered essential too. Just my opinion.

vrurg:
Ok, I have reduced it to the absolute minimum now. :)

src/core.c/Distribution/Path.pm6
```diff
@@ -52,6 +52,7 @@ class Distribution::Path does Distribution::Locally {
     }
 
     method meta(Distribution::Path:D:) { %!meta.item }
+    method modified(--> Instant:D) { $!meta-file.modified }
```
niner, line 55: We have stopped using file modification times for precomp checks long ago. For one, they don't tell you if a file has actually been modified (could just have been touched or written to with the same content) and there are file systems with a timestamp granularity of 2 whole seconds. Checksums are superior in just about every regard. We should not make a step back here.

src/core.c/CompUnit/Repository/Distribution.pm6
```diff
@@ -30,6 +30,10 @@ class CompUnit::Repository::Distribution does Distribution {
     method id(--> Str:D) { nqp::sha1(self.Str) }
     method meta(CompUnit::Repository::Distribution:D:) { %!meta.item }
 
+    method modified(--> Instant:D) {
+        $!dist.?modified // BEGIN Instant.from-posix(0)
```
niner, line 34: This is another reason, why a file modification timestamp is not the right tool for the job. Not all Distributions are file based. Everything could be stored in memory or one archive, e.g. when someone implements fatpacking of Raku applications. On the other hand, you will always have access to the meta data itself, which is actually what you really care about.

niner (changes requested):

niner:
My initial reaction to this was that it's just one manifestation of the more general "external precomp dependencies" issue discussed in https://github.com/Raku/problem-solving/issues/159 which already has a suggestion for a solution in https://github.com/Raku/problem-solving/issues/159#issuecomment-583522999

While that is still the case, I also realized that the META6.json dependency is a fundamental one. We were just lucky, that our lack of handling it explicitly rarely causes issues. It doesn't causes issues with installed modules as for every change we require an update to the version which results in a change of the distribution's SHA. The dependency between compilation result and META6.json contents will only get stronger if we implement the suggestion to fall back to the distribution's api, auth and version in the absence of explicit values in use statements.

While we don't want to require the user to add the dependency via the mechanism suggested in the linked problem-solving ticket for a dependency as fundamental as this one, Rakudo itself can simply use the same mechanism. This would avoid the architectural issues with this PR: PrecompilationRepository should not know about distributions or how we handle them. It's interface deals with one module at a time.

src/core.c/CompUnit/Repository/Distribution.pm6
vrurg: > Not all Distributions are file based.

See my discussion with Nick about that. While, in general, I agree that mtime is not optimal for the job, yet there will always be a way to know when a distribution has been changed.

Speaking of fatpacks and alike, they are rather immutable things where distribution wouldn't change. For that epoch would work perfectly good.

vrurg:
@niner @ugexe ping for re-review of checksum-based re-implementation.

src/core.c/Distribution/Locally.pm6
```diff
@@ -1,4 +1,4 @@
-role Distribution::Locally does Distribution {
+role Distribution::Locally does Distribution::Checksum {
```
ugexe, line 1: `Distribution::Locally` doesn't require a checksum method

ugexe:
Another thing that might be wise is to make the method an implementation detail, but also the problem of needing a way to tell if something changed doesn't require an API hard coupled (by way of being public) to a specific algorithm that will go out of date. An API that only provides a yes/no answer on if its an identical distribution would be a safer public API to expose because the implementation details of that are not exposed to the user. Theoretically it might be useful to have a method that returns the sha1 checksum, but only in the same vein that theoretically it would be useful to expose the json-ification methods (i.e. `method serialize`) -- exposing such a method would hard couple the json implementation (including bug compatibility) to any consumers of `Distribution`. I don't necessarily have a strong feeling here, but traditionally this area of the code has been very conservative about exposing public apis.

--- 

> architectural issues with this PR: PrecompilationRepository should not know about distributions or how we handle them. It's interface deals with one module at a time.

It seems like `$*DISTRIBUTION` is still being used; did you have a response to the comment above?

src/core.c/Distribution/Locally.pm6
vrurg: Initially, I was lacking information about the exact usages of each kind of `Distribution::` and wasn't sure if there is no way for any of them to end up in `$*DISTRIBUTION`. 

But later on, with wider picture in mind, I got to the conclusion since distribution meta data, generally speaking, affects actions taken upon the distribution itself, there must be a way to know if meta has changed. This consideration makes me think that it has to be a fundamental property of the distribution. In a way `meta` itself can serve the purpose with one huge _but_ when it comes to the current implementation: it is not guaranteed to reflect the original meta, as provided by the source. In particular, `InstallationDistribution` is served with `ver` and `api` fields replaced with `Version` objects. `LazyDistribution` reports a subset of keys. `Distribution::Path` adds `files` key.

Therefore the only way for CURD to get the real checksum of the original data is for these implementations to provide it computed at some point _before_ that change anything.

So, CURD could be the only one to have the method but only when we guarantee that the meta provided is what the low-level object has be initialized with. I'd only be happy if this principle is implemented.

vrurg:
> An API that only provides a yes/no answer on if its an identical distribution would be a safer public API to expose because the implementation details of that are not exposed to the user.

This option I was considering and did want it too. Probably by turning precompilation unit's `is-up-to-date` into a multi with a distribution-accepting candidate. But a working model was the priority. And for now this is part of a bigger issue of `PrecompilationRepository` abstraction level.

> Theoretically it might be useful to have a method that returns the e.g. sha1 checksum

Another matter I was considering and just haven't came to a satisfactory solution. It must be something identical to CURD's `id`. Nobody cares about how `id` is generated as long as it is guaranteed to be unique based upon api, ver, and auth values. Technically, it would be just a name change for `checksum` method, but that would allow approaches different to checksumming if possible and sensible.

> > architectural issues with this PR: PrecompilationRepository should not know about distributions or how we handle them. It's interface deals with one module at a time.
> 
> It seems like `$*DISTRIBUTION` is still being used; did you have a response to the comment above?

Oh, yes. This one. The most problematic.

If a module is part of a distribution then this information is used to create a precomp unit. Because in order to be able to check the unit's _up to date_ status against the distribution we need to have the required information available with the unit. With the current implementation, units are created by `PrecompilationRepository`. This just leaves no other options but to use `$*DISTRIBUTION`. Alternative would be for `CompUnit::PrecompilationStore::FileSystem` in its `new-unit` method or percompilation unit class constructor to do it. Neither option I like.

So, verification of distribution source has been added side-to-side with source code verification.

This brings up another question. `PrecompilationRepository`, as its name suggests, must only be responsible for operating  precomp repo collection and precomp units. But is it true? No, because not only it pays attention to the source file, but it actually initiates precompilation, and it binds itself to the Raku language exclusively at the same time. Keeping in mind that, in theory, we can have another language (implemented, say, on top of NQP) compiling into compatible precomp units and providing its modules through the same repository, that is most certainly not right.

In my view, a lot of `PrecompilationRepository` functionality must be extracted into a language implementation details entity. That entity has to take the responsibility of language-specific work including verification of up-to-date status of a unit. Roughly, `PrecompilationRepository` `try-load` should turn into something like this:

```raku
method try-load($dependency, $lang-handle, :@precomp-stores) {
    my $precomped := 
        self.load($dependency.id, :@precomp-stores)
            or self.store-unit($lang-handle.precompile($dependency, :@precomp-stores));
    }
    $lang-handle.post-process($precomped);
}
```

The biggest problem is that I'm not ready for a refactor of that scale.

PR #5060, opened 2022-09-16
Let `use` get default adverbs from META6.json

vrurg:
Apparently, works for distributions only.

Significant changes:

- Packagy type objects get their default api/ver/auth from META6.json
- `$?DISTRIBUTION` is now installed unconditionally into compunit's lexical space
- Scripts/apps detected as part of a distribution (commonly, these residing under ./bin) now have access to $?DISTRIBUTION`
- CUR::* `files` methods can now return distributions if requested
- CUR::RepositoryRegistry got `candidates` method
- New role `Distribution::Utils`

Resolves Raku/problem-solving#340


niner:
I have only had a quick look, since I'm on vacation. A good rule of thumb is: if your commit message contains a bullet list, your commit is too large and needs to be split into commits for those individual changes. As it is, the single commit changes a lot both of API and implementation with barely any explanation of why these changes are needed. This makes it very hard to review.

One thing I have noticed is that it seems to me like we would use the current distributions version and api for all unqualified use-statements, even if those modules would be loaded from completely different distributions, where version and api would need to be different as well.

src/core.c/CompUnit/Repository/Distribution.pm6
```diff
@@ -54,6 +57,24 @@ class CompUnit::Repository::Distribution does Distribution {
         }
     }
 
+    # Try to locate a distribution by a given file name. Only makes sense for CURFS.
+    # $file is expected to be either absolute or relative to $*CWD.
+    method from-file(::?CLASS:U: $file, :$name, :$ver, :$auth, :$api --> ::?CLASS:D) {
+        my @fs-repos = $*REPO.repo-chain.grep(CompUnit::Repository::FileSystem);
```
ugexe, line 63: If you have to grep for the type of repository then its not generalized enough

src/core.c/CompUnit/Repository/FileSystem.pm6
```diff
@@ -222,7 +222,7 @@ class CompUnit::Repository::FileSystem
     }
 
     proto method files(|) {*}
-    multi method files($file, Str:D :$name!, :$auth, :$ver, :$api) {
+    multi method files($file, Str:D :$name!, :$auth, :$ver, :$api, :$dist) {
```
ugexe, line 225: The dist identity is already declared by the previous parameters. Passing in a dist alongside them doesn't really make fit the api.

vrurg: It's rather about bad naming of the parameter. It's a boolean flag, request for distribution to be returned instead of meta. Your confusion is a clear sign of it needing a rename, but the only option I can come up with is `:$as-dist` and it doesn't appeal to me.

lizmat: :$return-as-dist ?

src/core.c/CompUnit/Repository/Distribution.pm6
vrurg: Agree. I had too many other new concepts to grasp while doing all this, so this one just slipped out of my attention.

Do you have a good idea? I mean, I clearly didn't want the installations to be searched. The best alternative I see is `CompUnit::Repository::Locally && ! CompUnit::Repository::Installable`.

src/core.c/CompUnit/Repository/FileSystem.pm6
vrurg: > :$return-as-dist ?

Would like a short version. After all, I expect the argument to be used more often than omitted. The only reason it can't be the default for now is backward compatibility.

lizmat: Can't we make it default depending on a flag passed to CURFS ?

vrurg:
> A good rule of thumb is: if your commit message contains a bullet list, your commit is too large and needs to be split into commits for those individual changes.

The last thing I tried before committing is to make smaller chunks. Unfortunately, the whole change is so much interlinked that I managed to locate only two hunks which could be made commits on their own. Since there is another rule of thumb: a commit must be bisectable, – I decided that this time a mega-one is justified. :)

> One thing I have noticed is that it seems to me like we would use the current distributions version and api for all unqualified use-statements, even if those modules would be loaded from completely different distributions, where version and api would need to be different as well.

Different distribution -> different precompilation -> different version. That is if you use the same compunit in `provides` of different distributions, though I barely see why this would make sense.

But if by "loaded" you mean `use`, then this is just not gonna happen. Precompilation takes versions from the distribution to which a module belongs. Any other behavior would be an error.

src/core.c/CompUnit/Repository/FileSystem.pm6
vrurg: > Can't we make it default depending on a flag passed to CURFS ?

We can, but that's another complication I would rather avoid.

Instead, I think I have a more generic solution for language-specific behavior of methods. Based on `PseudoStash` experience, the following works well for 6.e:

```
my class PseudoStash is CORE::v6c::PseudoStash {
```

So, any method in 6.c implementation can do `self.^language_revision eq 'e' ?? do { 6e-ish } !! do { 6c-ish }`.

But that's something for later.

ugexe: You can get the dist using the meta data returned from files via something like `my %file-meta = CUR.files(...); CUR.candidates($file-meta<name>, $file-meta<ver>, ...).head`.

src/core.c/CompUnit/Repository/Distribution.pm6
ugexe: At the very least it should be querying (despite the name `from-file` this is indeed effectively querying) a `CompUnit::Repository` object, not a distribution. Furthermore, since it is only ever called with the assumption that its for a precomp file, I would imagine the method (wherever it is) would be more appropriately be called something like `method distribution-from-precomp-file is implementation-detail { ... }` and potentially calling it via `.?distribution-from-precomp`. At least that is my initial thoughts.

src/core.c/CompUnit/Repository/Installation.pm6
```diff
@@ -88,7 +89,13 @@ sub MAIN(:$name, :$auth, :$ver, *@, *%) {
             $!installed-dist;
         }
 
-        method meta(--> Hash:D) {
+        method !repo-dist {
+            ⚛$!repo-dist // cas $!repo-dist, {
+                $_ // CompUnit::Repository::Distribution.new(self)
```
ugexe, line 94: Does this really need an instance?

src/core.c/CompUnit/Repository/Installation.pm6
```diff
@@ -103,9 +110,12 @@ sub MAIN(:$name, :$auth, :$ver, *@, *%) {
 
             %hash;
         }
-        method content($content-id --> IO::Handle:D) { self!dist.content($content-id) }
-        method Str { CompUnit::Repository::Distribution.new(self).Str }
-        method id { $.dist-id }
+        method content(::?CLASS:D: $content-id --> IO::Handle:D) { self!dist.content($content-id) }
+        method module-dependency(::?CLASS:D: Str:D $module_name) is implementation-detail {
```
ugexe, line 114: I wonder why this isn't implemented on `CompUnit::Repository::Distribution` and only called on such objects, because this function should always have the same implementation regardless of the distribution or repository backing it.

ugexe:
fwiw an unfortunate reality is that without precompilation of scripts (and thus the ability to use `$*DISTRIBUTION` in test files and bin scripts) this auto-choosing / loading logic won't be able to work consistently, i.e. it can only work for modules and `use Foo;` will work completely different for the aforementioned scripts.

src/core.c/CompUnit/Repository/Installation.pm6
vrurg: No, but for caching purposes. A trade-off.

vrurg: If you mean the attribute.

src/core.c/CompUnit/Repository/Installation.pm6
vrurg: I don't remember the exact details, but somewhere `LazyDistribution` was used directly. But as you can see, all the method does – it re-delegates to CURD. `content` re-delegates to the backing distribution because that's what CURD does.

I can try removing them and recover what was wrong about them being missing. But I can guarantee that they are there not because I just wanted it be that way.

vrurg:
> fwiw an unfortunate reality is that without precompilation of scripts (and thus the ability to use `$*DISTRIBUTION` in test files and bin scripts) this auto-choosing / loading logic won't be able to work consistently, i.e. it can only work for modules and `use Foo;` will work completely different for the aforementioned scripts.

It is only true for scripts not included into meta. Works perfectly good for those in ./bin. First, because when installed a script is invoked via `CU::RepositoryRegistry` which knows the distribution. Second, because ./bin is automatically included into `files` key of meta.

Tests are problematic _for now_. I don't see why can't we add them to the same key in the same way we do for ./bin? Or find another way of providing the necessary information? It's only a matter of strictly defined rules of the game.

ugexe:
> Works perfectly good for those in ./bin

Really? The reason I haven't added versioned imports into zef is because it doesn't work in bin scripts (last I knew)

> First, because when installed a script is invoked via CU::RepositoryRegistry

What about for scripts that aren't installed i.e. `raku -I./zef-repo bin/zef install ./zef-repo`?

vrurg:
> Really? The reason I haven't added versioned imports into zef is because it doesn't work in bin scripts (last I knew)

> What about for scripts that aren't installed i.e. raku -I./zef-repo ./zef-repo/bin/zef install ./zef-repo?

From the announcement comment:

> - Scripts/apps detected as part of a distribution (commonly, these residing under ./bin) now have access to `$?DISTRIBUTION`

And Raku/roast#822.

Where you commented about grepping being non-generic, that's the part where distribution is getting found by script path.

src/core.c/CompUnit/Repository/FileSystem.pm6
vrurg: Addressed with 672a9b505e2488481343759730b43e224e4a8390. As I still don't see what name would serve better than `:dist` the commit just adds `Bool` type constraint to the parameter.

src/core.c/CompUnit/Repository/Distribution.pm6
vrurg: > since it is only ever called with the assumption that its for a precomp file

The assumption is wrong. The method is to be used for a file, not for a precomp. I.e. its purpose is to locate the distribution to which the file belongs.

src/core.c/CompUnit/Repository/Installation.pm6
vrurg: @ugexe I wonder if `candidates` method is supposed to always return CURD. Because the reason for `module-dependency` method is `LazyDistribution` is returned by `CUR::Installation.candidates`.

src/core.c/CompUnit/Repository/FileSystem.pm6
ugexe: It might be better to create a new method for that, and make this files method deprecated or implementation detail. The new method would be infinitely more useful as a public api than `files`, and the only reason `files` still exists is because its fast pathing with that `<source>` field was needed by `CUR.run-script` to maintain the last overhead.

vrurg: + method `distributions`, - method `files` (deprecate, surely)?

What is the exact meaning of `source`? Can dump it altogether?

src/core.c/CompUnit/Repository/Distribution.pm6
vrurg: Addressed with fc4ee6e73309cefbe5784a48584a18e1f1024ce5

src/core.c/CompUnit/Repository/Installation.pm6
vrurg: Addressed with b92ba9af8ff903da5c6da4a04ab3d40c49d219ef. I decided that it would be right to make all `candidates` method consistently return CURD.

src/core.c/CompUnit/Repository/FileSystem.pm6
vrurg: Addressed with d6b4b5f160dc3e060d2d9beb510a986362bdc88e.

`source` is used in a couple of locations, so for now I don't touch it.

vrurg: And as to deprecation of `.files`, it is currently rather problematic because it is used by roast's S11-repository/cur-candidates.t. It doesn't feel OK to just replace the method in there.

src/core.c/CompUnit/Repository/Installation.pm6
vrurg: After b92ba9af8ff903da5c6da4a04ab3d40c49d219ef where `module-dependency` was removed, the attribute became basically redundant and has been removed in d6b4b5f160dc3e060d2d9beb510a986362bdc88e.

src/core.c/CompUnit/Repository/FileSystem.pm6
ugexe: Not to nit pick the name further, but it seems odd that a method called `candidates` takes all those same parts -- except a path name instead of a name -- and does basically the same thing as a method called `distributions`. Ideally they'd probably both be the same method, but there isn't a way to tell the difference beween a path `foo` and a namespace/provides `foo` unfortunately. And adding a named parameter to `candidates`. Additionally `CURFS` already has a method `distribution`, so it also seems a bit off that the plural is a querying mechanism (and only then a specific kind of query). `name-candidates` + `path-candidates` might have been a more cohesive way of naming these originally. That being said I don't have a better name suggestion, so these are just comments.

src/core.c/CompUnit/Repository/Installation.pm6
ugexe: The idea at least is that the user shouldn't really care what implementation of Distribution they get, only that they get a `Distribution`. Its entirely possible people (maybe even myself) are abusing that by assuming an e.g. `Distribution::Hash` vs `Distribution::Path`, but it would probably be better to just fix any such instances.

src/core.c/CompUnit/Repository/FileSystem.pm6
ugexe: Mostly unrelated to this PR but related to my previous comment is that I always felt `CURFS.distribution` and `CURI.installed` should instead both be called `.distributions` so one can do `$*REPO.repo-chain.map(*.?distributions)` since `.installed` doesn't make sense for a `CURFS` and a singular `.distribution` doesn't make sense for a CURI

vrurg: Why not `.candidates(:$file, ...)`? I like the idea of merging these two into one.

ugexe: As long as both `$name` and `:$file` can't supplied at the same time that could work

vrurg: Done. So, it is now `candidates($name, ...)` and `candidates(:$file!, ...)`. There was an idea of making use of both positional and `:file` an error, but since the named would currently be silently swallowed, I decided not to break this behavior.

niner:
> The last thing I tried before committing is to make smaller chunks.
> Unfortunately, the whole change is so much interlinked that I managed to
> locate only two hunks which could be made commits on their own.

Sorry, that's just not good enough. You can e.g. introduce a new API in one 
commit and switch over to using that in a following one. That gives you a 
whole commit message for explaining the rationale for the new API. As it is, 
this PR is not reviewable.

vrurg:
Now, with all happy CI, I plan to merge this Tue-Wed, unless there are more objections.

tools/build/install-core-dist.raku
```diff
@@ -1,6 +1,7 @@
+BEGIN %*ENV<RAKUDO_INSTALL_STAGING> = 1;
```
ugexe, line 1: We cannot depend on an env var to choose a repo like this — repos are chosen via their repo spec and adding it to the repository chain

src/core.c/CompUnit/Repository/Distribution.pm6
```diff
@@ -54,6 +56,25 @@ class CompUnit::Repository::Distribution does Distribution {
         }
     }
 
+    # Try to locate a distribution by a given file name. Only makes sense for CURFS.
+    # $file is expected to be either absolute or relative to $*CWD.
+    method from-file(::?CLASS:U: $file, :$name, :$ver, :$auth, :$api --> ::?CLASS:D) is implementation-detail {
+        my @fs-repos = $*REPO.repo-chain.grep({ ($^repo ~~ CompUnit::Repository::Locally)
```
ugexe, line 62: Grepping for repo types like this is still a code smell

ugexe (changes requested):

ugexe:
@niner was on vacation recently (and may well still be) so I would suggest waiting for more review comments

src/core.c/CompUnit/Repository/Distribution.pm6
vrurg: What do you propose instead? The purpose of the method is to locate a distribution by a file belonging to it. So, it means

a. we talk about `IO::Path`, which is what `CUR::Locally` is
b. we don't consider installations because for them the distribution is known a priori

tools/build/install-core-dist.raku
vrurg: This is not about choosing a repo but about working around detecting the current distribution at compile time. Since this is a kind of bootstrapping process an exception should not be a problem for it.

vrurg:
> @niner was on vacation recently (and may well still be) so I would suggest waiting for more review comments

@niner noted above that he consider this PR unreviewable. Unfortunately, however much I agree with the rules he expressed, the real situation doesn't let me to conform to them now. I'd better skip all the boring details on why is it so, but I'm not able to redo all the commit history of this branch.

tools/build/install-core-dist.raku
vrurg: I would try to see if there is another solution, but an explicit flag for staging makes things a lot simpler and easier to understand.

src/core.c/CompUnit/Repository/Distribution.pm6
ugexe: You don't know what type of 3rd party repositories might be added externally by users later, hard coupling a bunch of this system when it shouldn't be.

vrurg: `CUR::Locally` is there for the need of `$repo.abspath` attribute.

So, to get rid of the filter there is a constraint is to be put on `:file` argument of `.candidates` (or rather one is to be removed): the path to file is not necessarily normalized into a form relative to the repository prefix. This would offload the task of normalization on repos themselves, but otherwise I would even be in favor of such change.

I'll try it.

tools/build/install-core-dist.raku
vrurg: Addressed with 8006fdab0808c5ec7e73ca4a9c99b668d3c8a5e8

src/core.c/CompUnit/Repository/Distribution.pm6
vrurg: Addressed with 8006fdab0808c5ec7e73ca4a9c99b668d3c8a5e8

ugexe: What I mean is I still querying logic (i.e. calling `.candidates`) doesn't belong on a CUR-agnostic Distribution object. I would expect the call site (around `$Distribution.from-file(self.current_file()`) to make the call to `.candidates` or whatever.

src/core.c/CompUnit/RepositoryRegistry.pm6
```diff
@@ -149,7 +149,7 @@ class CompUnit::RepositoryRegistry {
         my str $prefix = nqp::ifnull(
           nqp::atkey($ENV,'RAKUDO_PREFIX'),
           nqp::gethllsym('default', 'SysConfig').rakudo-home()
-        );
+        ).IO.absolute.IO.resolve.Str;
```
ugexe, line 152: Calling `.IO` multiple times seems redundant. I think `.IO.resolve.absolute` is equivalent

vrurg: No, it isn't, unfortunately. I need to recover when it wasn't interchangeable, but this order was experimentally proved.

I'd rather have `absolute` return `IO` too. Beats me why does it return `Str` together with `relative`.

tools/build/install-core-dist.raku
ugexe: What about debian packagers that use this same CURS workflow for creating os packages?

src/core.c/CompUnit/Repository/Distribution.pm6
ugexe: Yeah path normalization should be something repos do. That is -`:$file` should always be of the relative form that would be found in a META6.json field, and the repository is responsible for creatin

ugexe: Also long ago `CUR::Locally` was used for providing an `absolutify` method, and internally the code was called something like `my $abspath = $repo.?absolutify($path) // $path.absolute`, which seems like it was trying to provide an api for the path normalizaion / absolutification

ugexe:
Its probably worth testing these repository modules to get an idea for possible fallout. I know the ones I authored are current and passing tests on windows, macos, and linux:

https://github.com/ugexe/Raku-CompUnit--Repository--Tar
https://github.com/ugexe/Raku-CompUnit--Repository--Lib
https://github.com/ugexe/Raku-CompUnit--Repository--Github
https://github.com/niner/CompUnit-Repository-Mask

src/core.c/CompUnit/RepositoryRegistry.pm6
ugexe: `.Str` returns the form originally supplied to the constructor so to speak. `.absolute` returns the absolutified string form, and `.relative` the relative form. There isn't really a need for `.absolute` to return an `IO` object because that object is supposed to be aware of the context needed to access the file regardless of it it was supplied a relative or absolute path (since it knows the cwd).

src/core.c/CompUnit/Repository/Distribution.pm6
vrurg: > What I mean is I still think querying logic (i.e. calling `.candidates`) doesn't belong on a CUR-agnostic Distribution object. I would expect the call site (around `$Distribution.from-file(self.current_file()`) to make the call to `.candidates` or whatever. `from-precomp` isn't querying so its not quite the same

I didn't know `Distribution` must be repo-agnostic. And, actually, it's wasn't full conforming to this constraint because it knows about the registry and expects a repository object for `CUR::Distribution`. Event the presence of `Repository` in its name makes it not so agnostic.

The other reason to have `from-file` in CURD was its similarity to `from-precomp`. Makes some sense to have both at the same location.

An alternative was to have the method planted onto `CU::RepositoryRegistry`. But then it breaks the semantic link between the two similar methods and would require longer self-explanatory name like `distribution-for-file`. Not that I would be against moving and renaming it, but the current implementation seemed like a bit more logical to me.

> That is -`:$file` should always be of the relative form that would be found in a META6.json field

In case of something like `raku -I$HOME/src/project $HOME/src/project/bin/script` CURFS should still be capable of resolving the full script name and locating it in meta.

> Also long ago `CUR::Locally` was used for providing an `absolutify` method

Normalization is slightly different as it tries to turn the file name into form usable for matching against meta keys.

To sum up things, I currently see conflicting requirements as to how `:file` must be treated. `run-script` uses the relative form of `"bin/$script"` to locate the installation repository. Yet, `raku -I. bin/script` would result in false positive on a `CURI` if the local directory distribution has been already installed earlier. So, it brings me back to the point where `from-file` does path normalization per-repo.

tools/build/install-core-dist.raku
vrurg: I need to see what they do. Better if there is a test for it.

vrurg:
> Its probably worth testing these repository modules

It most definitely worth. Unfortunately, `.?` doesn't work on missing multi-candidates (and that's for a reason). Could wrap it with `try`, but don't like the cost of it.

Need to think about it. Ideas are welcome too. :)

src/core.c/CompUnit/Repository/Distribution.pm6
ugexe: I'm not sure path normalization such as changing `\` to `/` needs to happens when calling `file` -- the user should always be supplying the paths as they would be in a META6.json.  For example `"foo" : "lib\bar\baz.rakumod"` doesn't need to work, it should be `lib/bar/baz.rakumod`. It already can't be expected normalize things like casing without causing conflicts for itself, so in that regard its reasonable to just expect the user to a) only use prenormalized paths in their META6.json and b) only use those same style of paths when querying for a file. It *doesnt* really make sense for `:$file` to ever be an actual `IO::Path` object -- the more idiomatic name that is used elsewhere is `$name-path` which better describes it isn't meant to be a file object.

ugexe: > raku -I. bin/script

If the first repository, in this case `.` -- or `CURFS` of `$*CWD` -- contains a `bin/script` then it should always be chosen. Indeed it would never reach the point of asking `CURI` if it had that file. So I'm not sure Im understanding the issue.

ugexe: > I didn't know Distribution must be repo-agnostic. And, actually, it's wasn't full conforming to this constraint because it knows about the registry and expects a repository object for CUR::Distribution. Event the presence of Repository in its name makes it not so agnostic.

It used to be called `CU::Distribution if I remember correctly. I wasn't involved in naming that, but I did make the push for defining the `Distribution` interface, what it was meant to accomplish, and how it would do so. With that being said, maybe it doesn't conform but I'm not sure thats a reason to push it further in that direction. Querying repos from a Distribution object just seems antithetical to the `Distribution` `CompUnit::Repository` api and division of responsibility. I can mull over better solutions, but its going to take me more than a few days.

vrurg: > I'm not sure path normalization such as changing `\` to `/` needs to happens when calling `file` -- the user should always be supplying the paths as they would be in a META6.json.

I already have it locally with `candidates` doing nothing to `file` which is type-restricted to `Str:D`. Normalization is offloaded to `from-file` method where it does make more sense.

> So I'm not sure Im understanding the issue.

This is with the default course of events, I.e. when the repo chain is fully built by the core. But what if it has been customized by user and an installation repo of any kind happens to be included before CURFS variants?

> Querying repos from a Distribution object just seems antithetical

Ok, then `CURD.from-file` -> `CURR.distribution-for-file`. Actually, this would allow to make it a public interface, alongside with the rest of `*-for-*` methods.

vrurg: > such as changing `\` to `/`

Speaking of this, I can't quickly find if there is any requirement to always use `/` in META6. So, there gonna be a problem if a Win* user decides to go with `\` they are accustomed with.

But even more importantly, path normalization must, probably,  take care of converting the separator for a Win-platform.

ugexe: > Speaking of this, I can't quickly find if there is any requirement to always use / in META6.

I imagine it might work as-is,  but only coincidently and not in every scenario. The design, at least, is around the concept of the `name-path`, a relative-to-the-distribution-root path. We aren't concerned with how e.g. windows users might think they can write something that looks like a name-path but isn't, similarly to how we don't concern ourselves if users on case-insensitive systems might not capitalize name-paths appropriately -- i.e. even though on their windows system `lib/foo.rakumod` opens `lib/Foo.rakumod` for a CURFS,  only one of `$distribution.content("lib/Foo.rakumod")` or `$distribution.content("lib/foo.rakumod")` should open that file (the one that matches what is in the META6.json file, not all the variants that would happen to open a file on a given persons file system).

ugexe: > This is with the default course of events, I.e. when the repo chain is fully built by the core. But what if it has been customized by user and an installation repo of any kind happens to be included before CURFS variants?

I'm not sure you need to work around this. Lets say I `use Foo:ver<0.1>` in some script which gets loaded from some CURFS, and that `Foo` does a `use Foo::Submodule;`:  This PR wants that to effective be `use Foo::Submodule:ver<0.1>`, which shouldn't concern itself with which repository it is loading from (hence why we don't have load-from-this-repo-only stuff like `use Foo:from-repository<XXX>`).   Loading `Foo::Submodule:ver<0.1>` should be sufficient regardless of what repository it came from. At least theoretically.

ugexe: > I'm not sure you need to work around this.
> shouldn't concern itself with which repository it is loading from

I guess a concern might be that it could be strange to have two slightly different `$?DISTRIBUTION` objects floating about representing the same distribution. I'm not sure of how to describe a scenario where that is actually broken though.

ugexe:
> @ugexe: OOC how would they be different?

Functionally they should be the same -- here is a `Distribution` interface you can use to interact with data. But practically speaking something like quasi code `$?DISTRIBUTION === OUTTERS<$?DISTRIBUTION>` wouldn't be guaranteed (even though to me that seems intuitive). I'm not sure if anyone would actually want to do something like that, but it does show a potential gotcha.

src/core.c/CompUnit/Repository/Distribution.pm6
vrurg: > Loading `Foo::Submodule:ver<0.1>` should be sufficient regardless of what repository it came from. At least theoretically.

You speak of module loading while my concern was about scripts and other kind of loading by path. Module loading logic is unaffected in this respect. 

But when there is a request for script `bin/script` specified on the command line, I barely see a user expecting it to come from an installation repository, no matter how it got into the head of the repo chain. Actually, I would expect the same to be true for `require "lib/Foo.rakumod"` where it is rather be expected a relative path, not a meta key.

And currently it all applies to `distribution-for-file` method only, whereas `candidates` (again) operate on keys only.

vrurg:
> Its probably worth testing these repository modules

31aaad907cb98b6316a8274eacdce236c5961714 covers the old implementations of CUR. BTW, `CompUnit::Repository::Mask` is likely to be long outdated as it doesn't work with master too.

tools/build/install-core-dist.raku
vrurg: @ugexe I see no solution for this. The problem is about conflicting understanding of where `core` repo is located between the script and precompiling `rakudo` sub-process.

For the script it is the staging repo. But when it launches precompilation the repository is reported with its spec as `CompUnit::Repository::Staging#name(core)#<path>` in `RAKUDO_PRECOMP_WITH`. That resolves into `CompUnit::Repository::Unknown` since staging module is not available.

Consequently, the precompilation resolves core repository with the default prefix for staging tools (Rakudo home) which is `$BASE_DIR/gen/build_rakudo_home`. This one remains empty. I'm not sure what is the purpose of it, but @patrickbkr may know better.

The result of the conflict is the compiler failing to locate the actual meta file for the core libraries because the script installs it into the staging repository. This was never a problem because the distribution was not used. But this PR tries to install `$?DISTRIBUTION` as early as possible and reveals the conflict.

Possible solutions would be:

- make the staging repo report itself as `inst#name(core)` for precompilation
- provide a way for `CU::RepositoryRegistry.repository-for-spec` to load custom repository classes by full module file path

Neither approach doesn't appeal to me as I anticipate unforeseen side effects. Though the first one feels as the most reliable because precompilation must not care about staging or not staging, all it needs is access to other precompiled modules, perhaps.

And yet, considering that the core libraries do not need the distribution, I would stick to the controlling environment variable. Just rename it for self-documenting purposes. Say, call it `RAKUDO_NO_PRECOMP_DIST`. The meaning would remain the same: when on, no attempt to locate the distribution would be done at compile time, except for the apparent locations in CU code where it is known as a result of resolution algorithms.

With regard to distro builders, they must not be affected unless they use the same mocking repository, as Rakudo build does.

vrurg: Eventually, after playing on with the problem, I came to the conclusion that explicit blocking of distribution deduction is currently the only solution. Unfortunately, since the precomp parent process provides distribution information to its children via `core` repository, which is then resolved to `$(BASE_DIR)/gen/build_rakudo_home` by them, whereas the required meta is stored under the real installation path, there is no non-hacky solution to get the task done.

Yet, the modules we install via staging process do not need the distribution information. Would they need it we'd hit the problem much earlier. Therefore disabling the deduction would be totally legit.

To be noted that, in my view, any solution to the conflict would be a temporary workaround for a hack, considering [the comment](https://github.com/rakudo/rakudo/blob/90709d5334d59f2e292fd2ec765f1b9a45695b6c/tools/templates/moar/rakudo-m-build.c.in#L40-L43).

Eventually, I ended up renaming the environment variable from rather obscure `RAKUDO_INSTALL_STAGING` to `RAKUDO_NO_PRECOMP_DIST`, as suggested in the previous comment. I consider this the final solution unless the whole staging process finds a non-hacky approach.

@ugexe would you still mind merging?

ugexe: I have two concerns with merging this in the immediate future.

1) There are two blockers that suggest precompilation was broken/deoptimized during the last round of `CompUnit::*` refactorings. As it stands we can't know if this PR would cause similar issues, because the PR is based on branch containing the issue itself. It would probably be wise to wait for that to be fixed / the next release.

2) I sort of wonder if this idea in general needs further design considerations. For instance: how is this supposed to interact with a theoretical `emulates` or `supercedes` etc features (see: https://github.com/Raku/old-design-docs/blob/master/S22-package-format.pod). Similarly, at many of my jobs our codebases would contain something like `lib/DBIx/Class/SomeDBICModule.pm` which was a patched version of the file that would be used by the rest of the original `DBIx-Class` distribution when it does `use DBIx::Class::SomeDBICModule`. My concern is then how is this behavior -- overriding a single module in another distribution -- supposed to be done in a future version of Raku? My mental model has been that `auth` would be the way to handle that (see https://stackoverflow.com/questions/55671684/how-does-raku-decide-which-version-of-a-module-gets-loaded#comment130257025_55671685), i.e. I wouldn't have pinned `auth` if I was the one implementing this. What are your thoughts on this?

vrurg: > 2\. For instance: how is this supposed to interact with a theoretical `emulates` or `supercedes` etc features

I don't see a reason why should it somehow prevent these two from being feasible. Though I may overlook something. But, eventually, It's like having an extra `provides`, just a somewhat more complex one.

> 2\. My mental model has been that `auth` would be the way to handle that

It would be nice to have a test for that. Because yes, I see it the same: without `auth` it would chose based upon `ver` and `api` by both matching and sorting for higher versions. And if it wouldn't then I'd consider it a bug.

> 1. because the PR is based on branch containing the issue itself.

This is the primary reason to postpone merging and, perhaps, any further work on this subject. Knowing the usual course of things, I predict someone would have to come up with an alternative implementation later when things are  settled down.

BTW, there was a note on packaging how that would be affected with staging issues. I was thinking about it and found no reason for them to fail. Would the process be susceptible to the problem then modules using `$?DISTRIBUTION` or `%?RESOURCES` would already be affected.

ugexe: > I don't see a reason why should it somehow prevent these two from being feasible. Though I may overlook something. But, eventually, It's like having an extra provides, just a somewhat more complex one.

There is a closed PR by me implementing emulates, and indeed its fairly simple. However, if distribution `Foo::Dist` contains a `JSON::Fast` and some other dist contains a META6.json with the following, then it seems like `Foo::Dist` maybe (I honestly don't know) should load the emulated `JSON::Fast` and not its own.

```
# Not::Foo::Dist
"emulates" : {
  "JSON::Fast" : "JSON::XS:auth<cpan:MLEHMANN>"
}
```

---

> Because yes, I see it the same: without auth it would chose based upon ver and api by both matching and sorting for higher versions. 

Actually my entire point here was a bit off. I don't think `auth` is useful for solving the issue I mentioned... rather something like: the ability to put a single CURFS `lib/JSON/Fast.rakumod` at the head of the repo chain such that any downstream distributions that try to load a `JSON::Fast` (even if they include their own) would load the CURFS one mentioned earlier. I'm not sure this type of overriding could ever work, though, due to issues like inferring if some file should get its own `$?DISTRIBUTION` or that of whatever its overriding.

vrurg: > then it seems like `Foo::Dist` maybe (I honestly don't know) should load the emulated `JSON::Fast` and not its own.

> the ability to put a single CURFS `lib/JSON/Fast.rakumod` at the head of the repo chain such that any downstream distributions that try to load a `JSON::Fast` (even if they include their own) would load the CURFS one mentioned earlier.

Actually, that's what I thought about. Moreover, having a CURFS at the head of the list is what logically comes to mind first. So far, the approach of using `$*REPO.head.need($dep-spec)` should implemented this logic (see CURR.NEED in this PR).

BTW, I think that if a distro has a module then any other module of the distro must load it's local version in first place. You know, a principle of lexical scoping applied to distributions. It's currently broken, but I would consider this rule as a must have. Including the case of `Foo::Dist` loading its own implementation of `JSON::Fast`.

vrurg: An unrelated consideration: we can have a caching repo at the top of the chain. Provided dependency specification has a unique ID, it can speed up lookups.

niner:
> rather something like: the ability to
> put a single CURFS `lib/JSON/Fast.rakumod` at the head of the repo chain
> such that any downstream distributions that try to load a `JSON::Fast`
> (even if they include their own) would load the CURFS one mentioned
> earlier.

Actually that's an important feature and the reason for why CURFS doesn't care 
about auth, api and ver at all. It's deliberately ignoring those so during 
development, you can use your own, patched version of a module.

niner:
> Not really true. CURFS skips reporting its distro as a match unless any of
> these are unset or set to `'*'`. Otherwise it doesn't care.

When did that change? The design doc is very clear on this and I know that the 
original implementation followed the design:

"A CompUnit::Repository::FileSystem always tries to satisfy a request for a 
module first, and only delegates if it is unable. It also doesn't care for 
versioning or authority."
https://github.com/rakudo/rakudo/blob/master/docs/module_management.md

ugexe:
@niner  During my last major work on CURs years ago I changed CURFS to match exact versions when using e.g. -I. but not -Ilib. When using a CURFS that doesn’t point at a META6 it uses the default * version, but when pointing at a directory with a META6 it treats it similarly to how a CURI would (which is done because until CURS works for non named repos CURFS is the only consistent workflow to test pre install). In other words — the “CURFS doesn’t care about versions” thing doesn’t make as much sense when considering people using multiple CURFS instead of a CURI.

patrickbkr:
@vrurg What is this PR blocking on?

vrurg:
@patrickbkr the question is more to @ugexe than to me. But either way, the PR is unlikely to be revivable as of now. And I barely going to do any major work in CU area anymore.

patrickbkr:
@ugexe What's your call on this PR? Is there any way forward?

ugexe:
I'm conflicted. While I think the majority of this is great, Tthere are couple parts of this that I'm not fond or sure of:

* `Distribution::Utils` uses `system-query` which, while probably fine in zef, makes me wary to be in the core. I've already found one fairly obvious bug in the zef implementation this draws from [here](https://github.com/ugexe/zef/commit/2a9f8854e8c8c54370a5105557c47d0147751018) which, if it was already released in rakudo, would be much harder to "fix" for rakudo users than updating zef.

* `Distribution::Utils` uses a grammar to parse modules names. This is fine in zef where a bug in the grammar can be fixed in zef, but if there is a bug in a rakudo version of such a grammar then certain module names are effectively broken for months/years while the bug fix propigates to e.g.packaged rakudos that aren't bleading edge. 

* `Distribution::Utils` uses `EVAL` in a way that may potentially be injectable -- `try { %adverbs = EVAL('(' ~ $spec ~ ')') } === Nil`

* `CompUnit::RepositoryRegistry` having `.candidates` methods feels kind of wrong, which is sort of supported by the special casing for my current 3rd party CURs.

* `distribution-for-file()` (and ideally any other new public method) should be `implementation-detail` until this new code has had a chance to see some practical use.

 Given the above it is hard for me to make a solo judgement call. I'm not sure I have a great suggestion for how to fix the first 3 bullet points, and the other person best able to review it is not able to. I'm basically trying to convince myself these issues aren't that important practically speaking. I suppose there are two actions, neither of which involve coding, that someone could do to relieve some my uncertanty:

1) Running Blin using this branch and seeing what the fallout is (I'm not sure just merging this and doing Blin afterwards would be better).

2) Measurement of start up times when loading modules before and after this PR.

vrurg:
> * `Distribution::Utils` uses `system-query`

The format is supposed to be supported. As far as I remember, it is not specced but detailed in a synopsis and is used in the wild. Hence we must support it and this requires a kind of `system-query` anyway.

> * `Distribution::Utils` uses a grammar to parse modules names. This is fine in zef where a bug in the grammar can be fixed in zef, but if there is a bug in a rakudo

This is where tests could help a lot. We could also figure out an alternative approach. Grammar is easier and faster to implement, as well as to read afterwards too.

> * `Distribution::Utils` uses `EVAL` in a way that may potentially be injectable -- `try { %adverbs = EVAL('(' ~ $spec ~ ')') } === Nil`

This one I'm truly ashamed of. It sneaked in when adverbs had more strict parsing in the grammar, so nothing malicious could end up in `$spec`. But then I encountered some rather complex cases (beats me, what it was, but shown up as a result of testing) and I loosened the grammar rule for it. 

Apparently, solvable. Perhaps, better solution would be possible with RakuAST.

> * `CompUnit::RepositoryRegistry` having `.candidates` methods feels kind of wrong

CU interfaces is what I'm really bad with. Since there are few to none other sources of information, I tried to follow the sources, but with some limited success, apparently. :)

Anyway, I don't see why CURR cannot be an umbrella object over particular registries? Eventually, when searching for a module candidate one does the same kind of iteration.

patrickbkr:
I've thoroughly reviewed this PR at the core summit. (I invested half a day.) The approach is sensible and the code looks sane. The points ugexe brought up in https://github.com/rakudo/rakudo/pull/5060#issuecomment-1439320174 stand.

@ugexe My recollection is, that we agreed on trying to move this forward. I can give a rebase a try. What do you think?

lizmat:
Yes, please!

patrickbkr:
There is now a rebase in https://github.com/patrickbkr/rakudo/tree/problem-solving-340.

src/core.c/CompUnit/Repository/FileSystem.pm6
```diff
@@ -220,6 +220,44 @@ class CompUnit::Repository::FileSystem
 
         return ($distribution,);
     }
+    multi method candidates(Str:D :$file!, Str:D :$name!, :$auth, :$ver, :$api) {
```
ugexe, line 223: This api feels a bit weird. From a purity standpoint this method should have just been `method candidate(CompUnit::DependencySpecification: $spec)` - the other candidates were added for the convenience of not having to type out the `CompUnit::DependencySpecification` part. i.e. one type of parameters are meant to map directly to the other. In our CUR code `:$file` is sometimes is treated as the short-name, and other times (like when `:$name!`) it is not. It sort of seems like maybe `CompUnit::DependencySpecification` should have an attribute for a file path. I'm not sure if that is a great idea or not, but something like `lib/Foo/Bar.rakumod` does sort of feel like a type of exact dependency specification.

src/core.c/CompUnit/Repository/Locally.pm6
```diff
@@ -22,9 +22,10 @@ role CompUnit::Repository::Locally {
     # from the "prefix" parameter, will be *ignored* any subsequent
     # attempt at creating an object of that type on that prefix.
     method new(CompUnit::Repository::Locally: Any:D :$prefix) {
-        my $abspath = nqp::istype($prefix,IO::Path)
-          ?? $prefix.absolute
-          !! $*SPEC.rel2abs($prefix.Str);
+        my $abspath =
+          (nqp::istype($prefix,IO::Path)
+              ?? $prefix.absolute
+              !! $*SPEC.rel2abs($prefix.Str)).IO.resolve.Str;
```
ugexe, line 28: We know `$prefix` isn't an `IO::Path` at this line; do we really need to call `.Str` on it?

src/core.c/CompUnit/RepositoryRegistry.pm6
```diff
@@ -302,15 +302,53 @@ class CompUnit::RepositoryRegistry {
...
+                }
+                .?candidates(:file(.?normalize-path($file) // $file), :$name, :$auth, :$api).head
+            }).grep(*.defined);
+        +@distros
+            ?? (@distros == 1 ?? @distros !! @distros.sort(*.meta<ver>).sort(*.meta<api>).reverse).head
+            !! Nil
+    }
+
+    # Find all candidate distributions for a given spec
+    proto method candidates(|) {*}
```
ugexe, line 328: I'm still not exactly convinced this `candidates` method (or `distribution-for-file` even) belong in this class. `CompUnit::RepositoryRegistry` deals with repositories, and those repositories may or may not deal with distributions (and the ones that *do* deal with distributions may not deal with *raku* distributions). I understand this logic needs to live somewhere, but it just seems wrong to include a method that "Only makes sense for CURFS" in an object used for managing generic repositories. I imagine it could live somewhere else, and while it might make using it from whatever callsite require a bit more logic, it would be more cohesive imo.

src/core.c/CompUnit/RepositoryRegistry.pm6
```diff
@@ -332,15 +370,36 @@ class CompUnit::RepositoryRegistry {
...
+        my $bin  = $distro.meta<source>;
+        my $*DISTRIBUTION := $distro;
         require "$bin";
     }
 
     method head() { # mostly usefull for access from NQP
         $*REPO
     }
 
+    method NEED(Str:D $module_name, %opts) is implementation-detail {
```
ugexe, line 383: This is another method that seems a bit out of place. For example why is there no `LOAD` / `require` logic? This seems to be further splitting up the loading logic between `need` and `load`

src/core.c/CompUnit/Repository/Locally.pm6
```diff
@@ -64,6 +65,12 @@ role CompUnit::Repository::Locally {
 
     # stubs
     method short-id(CompUnit::Repository::Locally:D: --> Str:D) { ... }
+
+    method normalize-path($path) {
+        (nqp::istype($path, IO::Path)
+            ?? $path.absolute
+            !! $path.Str.IO.absolute).IO.resolve.relative($.abspath)
```
ugexe, line 72: I don't think this method should be exposed - exposing its API doesn't really give users any benefit. It is only used in one spot, so it sort of feels like it should just be some helper private subroutine.

`$path.Str.IO.absolute` should probably just be `$path.IO.absolute`.

src/core.c/Distribution/Utils.pm6
```diff
@@ -0,0 +1,156 @@
+my role Distribution::Utils {
```
ugexe, line 1: Maybe this should just live in the same file as `CompUnit::Repository::Distribution`. That is the only spot it is used, and it is already a private class (so we shouldn't be concerned with the class name and path name matching).

src/core.c/CompUnit/Repository/Installation.pm6
```diff
@@ -532,6 +533,48 @@ sub MAIN(:$name, :$auth, :$ver, *@, *%) {
         }
     }
 
+    # if we have to include :$name then we take the slow path
+    multi method candidates(Str:D :$file!, Str:D :$name!, :$auth, :$ver, :$api) {
```
ugexe, line 537: Something else that is weird about including `:$file` and `:$name` in `method candidates` is that elsewhere (`method files`) `$name` refers to the distribution name, but here it refers to a module name.

src/Perl6/Grammar.nqp
```diff
@@ -1980,23 +1980,37 @@ grammar Perl6::Grammar is HLL::Grammar does STD {
...
+                        my $Version := $*W.find_single_symbol_in_setting('Version');
+                        if $longname {
+                            for $longname.colonpairs_hash('class') -> $adverb {
+                                my str $key := $adverb.key;
+                                if $key eq 'ver' {
+                                    $*VER := $*W.handle-begin-time-exceptions($/,
+                                        'parsing package version',
+                                        -> { $Version.new($adverb.value) });
+                                }
+                                elsif $key eq 'api' {
```
ugexe, line 1993: `api` should be a `Version` as we treat it like a version everywhere else (particularly for its sort logic)

ugexe:
@patrickbkr I agree I would like to see this move forward. I'm not sure about moving forward without addressing at least some of my comments though. My main issues are obviously with the design/api itself - with these changes I don't have a good intuition about this system (and I'm not sure @niner would either), and I'd be worried that we'd then be left without anyone having a mental model of how it should (and how it actually does) work.

ugexe:
I also wonder if there are other changes we could make that would make implementing this easier/cleaner. The one example that comes to mind is if we were to say "`META6.json` should only contain distribution names" (something I'm seriously debating about pushing forward myself) then it would allow simplifying a small part of the code in this PR. I wonder if there are other meta data changes we could spec that would lend themselves to improving this feature.

vrurg:
I would like to leave a note here that I barely be working on anything CU-related any time soon. If somebody could pick this up I'd be all for it and very much appreciate the move!

src/core.c/CompUnit/Repository/Installation.pm6
patrickbkr: In both CURI and CURFS the new `candidates` methods are exact copies of the `files` methods with the only difference that they return the found distributions instead of the contained `%meta`. What makes you think the one is referring to a distribution name while the other to a module name?

PR #5062, opened 2022-09-18
Better cope with laziness in list constructors

Kaiepi:
I ran into an odd problem while working on [`Data::Record`](https://raku.land/zef:Kaiepi/Data::Record); there doesn't appear to be any clean way to reproduce a `List` as a container in its entirety while swapping in a different iterator:

- `new` doesn't copy when the list is lazy.
- `clone` keeps elements.
- `STORE` reifies eagerly.
- `from-iterator` erases container-related information (container descriptors, shapes) and it's typed as a static method.
- `from-slurpy`, `from-slurpy-onearg`, `from-slurpy-flat`, and `from-list` want a wrapper `Seq` or `List` at the least.

This adds an `over-iterator` method to `List` instances, which is a clone-ish counterpart to `from-iterator`. `Array` wants its own version, but this is unimplemented at the moment. I think `new` might be able to steal its arguments' reification target and todo iterator for cheaper construction in certain cases that preserves laziness as well, but I haven't fully tested this approach.


src/core.c/List.pm6
```diff
@@ -190,6 +190,21 @@ my class List does Iterable does Positional { # declared in BOOTSTRAP
         nqp::p6bindattrinvres(result,List,'$!todo',todo)
     }
 
+    method over-iterator(List:D: Iterator $iter --> List:D) {
+        # from-iterator only copies type-level information; here we copy
+        # everything bar the elements (e.g. container descriptors, shapes)
+        nqp::stmts(
```
lizmat, line 196: Why the nqp::stmts??

src/core.c/List.pm6
```diff
@@ -190,6 +190,21 @@ my class List does Iterable does Positional { # declared in BOOTSTRAP
...
+          (my \buffer := nqp::clone(self)),
```
lizmat, line 197: I find the use of the name `buffer` confusing here, as it is in fact a `List` object.  In the sibling methods `buffer` is always an `IterationBuffer`.

Kaiepi: Right, it's `result` elsewhere?

lizmat:
A general comment: we're adding an internal method, but we're not using it anywhere.  So if I understand this right, we're creating an internal method for use by a module?  Shouldn't we rethink this towards a public API then?

src/core.c/List.pm6
Kaiepi: I've seen it produce a simpler QAST before, but I don't think I checked `--optimize`'s output at the time.

src/core.c/List.pm6
lizmat: Looks like  :-)

src/core.c/List.pm6
lizmat: That may be true, but for readability, it would definitely improve if `nqp::stmts` were not used when not really needed for performance.  And I don't think it is needed here.  And I doubt you would be able to show any difference in execution once spesh and inlining done their work.

Yes, there used to be a time when it did make a significant difference.  But that became a lot less when Parrot was not the backend anymore.  And with spesh it became even less.

Kaiepi:
I don't see this, `from-iterator`, or `Iterator` as being that degree of internal, but I'd rather this mirror any adjustments needed in more public API and maybe depend on it there than be its own thing. I add it because, in my case, I wouldn't need a wrapper `Seq`, `List`, or `Slip` like I would with other methods as implemented.

Kaiepi:
Doing `STORE` with `over-iterator` makes the container skips a reification afterwards, While this doesn't break tests, this could affect programs that depend on them forcing one on lazy input with side effects. I'm not sure on this change, but if reasonable, it does seem cheaper.

Kaiepi:
In making one `STORE` multi depend on `over-iterator`, all need to for some degree of coherency. This makes a difficult entry point because I imagine it needs to be able to cope with eager *segments* in an iteration in order for there to be any support for laziness at all in there while protecting its existing feaureset otherwise, which would be the primary benefit of doing this at all. Hopefully `new` is less involved...

Kaiepi:
Before, `List` would, in the case of:
- `STORE($, :INITIALIZE)`, eagerly reify the RHS.
- `STORE($)`, eagerly reify the LHS, then reify a portion of this segment's length from the RHS

This can be defined with a lazier approach. We can, in the case of:
- `STORE($, :INITIALIZE)`, borrow the RHS' iterator, then only reify until that decides it's lazy.
- `STORE($)`, do nothing to the LHS, only doing container assignments if there was any reified to begin with, *then* only reifying until it decides it's lazy, 

I have a difficult time trusting isolated benchmarks so far. Assuming that `Gauge` as a thin `Seq` **DOES NOT** depend on `List.STORE` in any capacity itself:

```raku
use v6;
use Gauge;

my constant @FINISH = eager List.from-iterator: [^10_000].iterator;

my constant %CASES =
    <store-in-lazily>
        => { List.CREATE.STORE: List.from-iterator([^10_000].iterator), :INITIALIZE },
    <store-in-finish>
        => { List.CREATE.STORE: @FINISH, :INITIALIZE },
    <store-re-lazily>
        => { List.from-iterator([^10_000].iterator).STORE: @FINISH },
    <store-re-finish>
        => { @FINISH.STORE: @FINISH };

only MAIN(Str:D $name where %CASES) {
    Gauge(%CASES{$name}).poll(1).throttle(19).produce(&max).map(*.say)
}

only USAGE {
    put 'Usage:';
    put "$*EXECUTABLE-NAME $*PROGRAM-NAME $_" for %CASES.keys;
}
```

Here are some peak iteration per second counts polled with a 19s throttle:

| Benchmark | [master](2070ceebbdc04c2853aca83c392570785c6a0a8d) (it/s) | [list-copying](https://github.com/rakudo/rakudo/pull/5062/commits/01e871b44a2e5d8106c4be11628b7dc0e2088bda) (it/s) | Ratio |
| - | - | - | - |
| store-in-lazily | 170 | 385 | 2.264705 |
| store-in-finish | 947 | 3469 | 3.663147 |
| store-re-lazily | 155 | 2252 | 14.4529032 |
| store-re-finish | 507 | 70 | 0.138067 |

Reassignments of eager lists on eager lists suffer because the prior `STORE` implementation was well optimized for this scenario. Other cases benefit to some degree *I think* because `push-until-lazy` skips some `pull-one`s that aren't always necessary. The JIT seems to enjoy this version of `List.STORE` going off profiling as well.

While 3/4 cases see some improvement, the `store-re-finish` case worries me somewhat because the most eager case should be fairly common, so I trust Roast times a little more: 4580s on `list-copying`, down from 5040s on `master`, a rough 10% improvement in general.

src/core.c/List.pm6
Kaiepi: I don't depend on `nqp::stmts` here anymore, but still do in the context of `StructuredBuffer`'s methods, which are depended on from the context of an `Iterator`. `Gauge`'s iterators are sensitive to it vs. actual statements, so I imagine a list assignment would be too.

Kaiepi:
This adds two new public methods to make custom `STORE` easier to implement from the ecosystem, but makes for questionable immutability semantics:

- `List.make-iterator` completes the `from-iterator`/`over-iterator` trio of constructors. This exposes the mechanism by which `List.STORE` recycles the list as a container, skipping an allocation it would need otherwise, but is arguably too generic a binding to expose.
- `List.imbue` complements `eager` in that it produces a tweaked list from itself. `eager` maps to a `reify-all` call on a list, while `imbue` does `reify-until-lazy` instead.

Using the bug fix commit as a point of reference, we get some more grounded numbers:

| Benchmark | [list-copying]() (it/s) | [list-copying]() (it/s) | Ratio (delta) | Ratio (total) |
| - | - | - | - | - |
| store-in-lazily | 408 | 396 | 0.970588 | 2.329412 |
| store-in-finish | 957 | 955 | 0.99791 | 1.008448 |
| store-re-lazily | 2255 | 2221 | 0.984922 | 14.329032 |
| store-re-finish | 153 | 154 | 1.006536 | 0.303748 |

`store-in-finish` loses its prominent gains against master with the bug fix. Every other case wins regardless, but `store-re-lazily` loses out a little at the same time with the API additions. There's more fiddling to do in the `iterator` methods to make them more extensible from `Array` and to do with `imbue` to make it a true complement to `eager`, but this makes a good checkpoint. Roast times rise to 4798.33s for the sake of correctness.

Kaiepi:
All constructors in `List` and `Array` bar `Array.new` have been touched on:

- `from-iterator` and `over-iterator` can be backed by a common `make-iterator`. These are explicitly lazy; they will never reify, regardless of the eagerness of the iterator provided. This changes the semantics of `Array.from-iterator`, which would reify the eager head of any iterator.
-  `from-list` now exists on `List` as well as `Array`. It is more eager in that it will reify any eager head of an iteration, e.g. a reified portion of a lazy list. This is no longer implicitly typed as List:D, but as Iterable:D. Some performance is lost, but because `imbue` now exists on `Iterable`, arrays can keep their reifying behaviour (lists are cheaper via `iterator`).
- `make-itemized` and `make-iterable` have been introduced to complement `make-iterator`. `from-list` can be backed by `make-iterable`, which carries a looser `Mu` typing, as this only cares about an `iterator`.
-  `List.STORE` with its `StructuredIterator` and `StructuredBuffer` have been scrapped and rewritten to be backed by a leaner `&destructure` routine. This restores most of the performance of a list destructuring if `List`'s `make-itemized` and `make-iterable` are depended on. `Array` has been refactored to match, but is much less complicated with its reification target.
- `List.new` reifies until lazy, like `from-list`, just given a `Slip` in particular. It can now delegate to `STORE`, just it cheats on `List` itself in that it is already provided one as an argument, it just needs to be reified.

In order to benchmark these changes at all, I needed to scrap and rewrite the benchmarks to target less complicated, more specific benchmark paths. This can now be run as a full suite with no arguments, or be passed `--forever` with an individual case name to keep prior behaviour:

```raku
use v6;
use Gauge;

my constant $VALUES = ^10_000;
my constant @VALUES = eager $VALUES;
my constant @SLIPPY = eager Slip.from-iterator: $VALUES.iterator;
my constant @RELIST = eager List.from-iterator: [$VALUES].iterator;

my constant @CASES =
    <imm-from-iter>
        => { List.from-iterator: $VALUES.iterator },
    <mut-from-iter>
        => { Array.from-iterator: $VALUES.iterator },
    <mut-from-list>
        => { Array.from-list: @VALUES },
    <imm-new-create>
        => { List.new },
    <imm-new-finish>
        => { List.new: @SLIPPY },
    <imm-store-in-lazily>
        => { List.CREATE.STORE: $VALUES.lazy, :INITIALIZE },
    <imm-store-in-finish>
        => { List.CREATE.STORE: $VALUES, :INITIALIZE },
    <imm-store-re-finish>
        => { @RELIST.STORE: $VALUES },
    <mut-store-in-lazily>
        => { Array.CREATE.STORE: $VALUES.lazy },
    <mut-store-in-finish>
        => { Array.CREATE.STORE: $VALUES };
my constant %CASES = @CASES;

only USAGE {
    put 'To run the complete suite:';
    put "$*EXECUTABLE-NAME $*PROGRAM-NAME";
    put 'To run individual benchmarks:';
    put "$*EXECUTABLE-NAME $*PROGRAM-NAME $_" for @CASES.map: *.key;
}

proto MAIN(|) {*}
multi MAIN(Str:D $name where %CASES, Bool:D :forever($)! where ?*) {
    Gauge(%CASES{$name}).poll(1).throttle(29).produce(&max).map(*.put)
}
multi MAIN(Str:D $name where %CASES) {
    Gauge(%CASES{$name}).poll(1).throttle(29).produce(&max).rotor(8 => -7).map({ $_[0] if [==] $_ }).head.put
}
multi MAIN {
    for @CASES -> Pair:D (:key($case), :value($)) {
        print "$case: ...";
        put "\r$case: ", await benchmark $case;
        print "cooling...";
        sleep 300;
        print "\r";
    }
}

only benchmark(Str:D $case --> Supply:D) {
    supply given Proc::Async.new: $*EXECUTABLE-NAME, $*PROGRAM-NAME, $case {
        whenever $^process.stdout.lines.head {
            emit val $^line;
        }
        whenever $^process.stderr.lines {
            note $^line;
        }
        whenever $^process.start {
            done;
        }
    }
}
```

Here are some numbers:

| case                  | [`master`](https://github.com/rakudo/rakudo/commit/feb04dbe967e9c5b1c41ac55ebd3b683d6d5bbcb) (it/s) | `list-copying` (it/s) | ratio       |
| --------------------- | -------- | -------------- | ----------- |
| `imm-from-iter`       | 1703918  | 1573686        | 0.9235691   |
| `mut-from-iter`       | 2287     | 1388846        | 607.278531  |
| `mut-from-list`       | 3068     | 2696           | 0.878748    |
| `imm-new-create`      | 2673944  | 3104224        | 1.1609159   |
| `imm-new-finish`      | 1040     | 845            | 0.8125      |
| `imm-store-in-lazily` | 643      | 2211623        | 3439.538103 |
| `imm-store-in-finish` | 4155     | 4005           | 0.963899    |
| `imm-store-re-finish` | 464      | 406            | 0.875       |
| `mut-store-in-lazily` | 2600045  | 2540937        | 0.9772665   |
| `mut-store-in-finish` | 2324     | 791            | 0.340361    |

This particular combination of flavours of reification allows for `mut-from-iter` and `imm-store-in-lazily` to skyrocket. Empty lists are cheaper to construct with its cheap slurpy trickery. Other numbers suffer to varying degrees. Tradeoffs are made in favour of lazy lists over eager lists in general.

I don't trust roast times so much given they've tanked somehow between now and my last test runs. I've updated various components of my system between then and now, just there's enough for me not to know exactly what to blame. `master` completes with 5758.25 *CPU* as per fudged output (not seconds, like I was saying before!), while `list-copying` beats it slightly with 5494.75 CPU.

Kaiepi:
Sorry for the noise, rebased on master.

Kaiepi:
There are a lot of wrinkles in shaped and native arrays to iron out on top of a pretty sizeable ripple from the refactoring done thus far. This gauntlet focuses on `Array::Shaped` primarily, but is forced to extend to typed and native arrays early at times for some degree of coherency between the arrays. This is *still* shaky ground nonetheless, so I don't do a benchmark quite yet, but this is the gist of what's going on:

- `Array::Shaped.reification-target` override's `Array`'s, which is useless to shaped arrays' reifications typically (`Shaped1` being the exception). While we override the `make` constructors these are geared towards anyways, we have `List`'s `from-slurpy` constructors to look into more deeply at some point.

- `Array::Shaped.make-itemized` makes an `X::Assignment::ToShaped` error. This wants an `Array::Shaped1` override at some point.

- `Array::Shaped.make-iterator` is the `Iterator:D` `STORE` candidate. This is flat by type in the eyes the shaped array, so it fills out its reification element-by-element with the reification target.

- `Array::Shaped.make-iterable` is the `Iterable:D` `STORE` candidate. Unlike the `Iterator:D` candidate, we try to infer a shape from `Iterable:D` elements originating from the root iterable's iterator. This is more complex an iteration, but is DWIMMY.

- `Array::Shaped.STORE`, unlike `Array.STORE`, doesn't care about containers. I'm not sure if this is intended, so I leave that aspect of it alone for now. There is missing handling of `str` and `uint` native arrays that can become low-hanging fruit with some refactoring though.

- `Array.STORE` and `List.STORE` now have `Iterator:D` candidates also backed by `make-iterator`, as with shaped and native arrays. In `Array`'s case, it can pipe iterators from `new`, but `List` can't do this as well without Roast breakage.

- `Array.set-shape` is an awkward constructor because it tries to be too magic in performing a mixin and constructing said mixin at the same time. It does need to do both, but the construction can be factored out to a simpler method to make dispatch less complicated. What remains of `set-shape` can serve as a vehicle to eliminating the need to override all of `new` from mixins, just under a less misleading name.

- `List.rub` is the most simplistic constructor of the bunch: it produces a copy of a list or instantiates given its type instead with no reification. In `Array`'s case, this abstracts a common step of copying shapes and types across instances in other constructors, e.g. `clone`. In `List`'s case, this can be used to eliminate the `over-iterator` method, which can then be defined cleanly as a chain of `rub` and `make-iterator`. Likewise, `make-iterable` and `make-iterator` can be chained with `rub` for a more direct path to a particular vein of `STORE`.

- `Array.dim` adds injection of a shape to a `rub` typically. It rename `set-shape` because it's setting the type too. It reminds me of `dim` from Visual Basic somewhat, thus the name (short for "dimension", for the record).

Past this point, *all* shaped and native array constructors are fair game.

PR #5065, opened 2022-09-23
Add a `.warn` method to exceptions

lizmat:
This will allow one to turn an exception into a warning:

    $ raku -e 'CATCH { .warn }; say "before"; die "foo"; say "after"'
    before
    foo
      in block <unit> at -e line 1
    after

Inspired by Xliff: https://irclogs.raku.org/raku/2022-09-23.html#10:45

### DIFF
--- a/src/core.c/Exception.pm6
+++ b/src/core.c/Exception.pm6
@@ -72,6 +72,11 @@ my class Exception {
         nqp::setpayload($!ex, self);
         nqp::rethrow($!ex)
     }
+    method warn(Exception:D:) {
+        nqp::setmessage($!ex, nqp::unbox_s(self.message));
+        nqp::setextype($!ex, nqp::const::CONTROL_WARN);
+        nqp::rethrow($!ex);
+    }
 
     method !maybe-set-control(--> Nil) {
         if nqp::istype(self, X::Control) {


src/core.c/Exception.pm6
```diff
@@ -72,6 +72,11 @@ my class Exception {
         nqp::setpayload($!ex, self);
         nqp::rethrow($!ex)
     }
+    method warn(Exception:D:) {
+        nqp::setmessage($!ex, nqp::unbox_s(self.message));
```
vrurg, line 76: I don't think this is good to mutate the exception instance itself. Why not to, roughly, just `warn self.message`?

vrurg (changes requested):

src/core.c/Exception.pm6
lizmat: Because that would lose the original backtrace?

vrurg: Mostly because it changes the kind of the exception. I make it a `CONTROL` one whereas the wrapping object is not `X::Control`. Besides and more importantly, this may break user expectations about the object.

So, I'd be for `.warn` to throw a new exception, derived from the original one. It could even bear the original exception as an attribute, akin to `Failure`.

vrurg: Oh, I got you all wrong... Well, yes, it would result in lost backtrace. Preserving the original exception in the new warning would be one solution.

Then the original exception could be cloned and modified.

Or `$!bt` could be re-planted onto the new exception instance.

lizmat: I tried that originally.
````
$ raku -e 'use nqp; nqp::clone(nqp::newexception)'
MoarVM panic: MVMException copy_to NYI
````

Comment (vrurg) [reply src/core.c/Exception.pm6:76]: I tried this one too. But either way, it's not my favorite. Here is a mock up I came up with:

```raku
use nqp;

class CX::Warn::Ex is CX::Warn {
    has $.exception handles <message>;
}

class MyEx is Exception {
    method message {
        "test exception"
    }

    method warn is hidden-from-backtrace {
        my $wex := CX::Warn::Ex.new(:exception(self));
        my $ex := nqp::newexception();
        nqp::setmessage($ex, nqp::unbox_s(self.message));
        nqp::setextype($ex, nqp::const::CONTROL_WARN);
        nqp::throw($ex);
        0;
    }
}

sub warningify(&code) {
    CATCH {
        default { .warn }
        # .warn
    }

    &code()
}

sub test {
    MyEx.new.throw
}

say "Pre-test";
warningify &test;
say "Post-test";
```

While playing with this I also noticed that `CATCH { .warn }` semantics of eating up the original exception is wrong. It kind of tries to follow the footsteps of `.fail`, but the difference is that `fail` is a returnish statement whereas `warn` is not. Correspondingly, the above mentioned `CATCH` must result in warning _and_ rethrow the exception for further processing. Hence `CATCH { default { .warn } }` in my snippet.

Xliff: I'm still wondering why everyone is assuming a CATCH is needed?

I've been envisioning it as a non-fatal exception type.

```
class W::Foo is CX::Warn {
  message "Foo!"
}

{
   # This turns a CX::Warn into a fatal exception
  CATCH {
    when W::Foo {
      .rethrow
    }
  }

 {
    my $a = 42;
    W::Foo.new.warn if $a > 30;
    my $a = 5;
    say $a;
 }

Would result in:

```
Foo
5
```

The inner block continues because W::Foo isn't fatal. The outer CATCH is never executed because W::Foo is not rethrown by the inner block. If the inner block HAD rethrown W::Foo, the "5" would have never been emitted.

Comment (vrurg) [reply src/core.c/Exception.pm6:76]: I don't get what you're trying to achieve. `W::Foo` is a control exception and, as such, it has to be handled by `CONTROL`. Hence, the following:

```raku
class W::Foo is CX::Warn {
    method message { "Foo!" }
}

{
    CONTROL {
        note "OK? ", .raku;
        when W::Foo {
            note "It's W::Foo";
            .resume
        }
    }

    my $a = 42;
    W::Foo.new.throw if $a > 30;
    $a = 5;
    say $a;
}
```

So, `W::Foo` isn't fatal. If you try `.rethrow` you'd get a `control exception without handler` error because that's what happens: one `CONTROL` is done and there is no another one to handle `W::Foo`.

@lizmat idea is different. This PR is about taking a normal exception and turning it into a `CX::Warn`-kind one. It's something I wanted for myself for a couple of times, BTW. :)

vrurg: > The outer CATCH is never executed because W::Foo is not rethrown by the inner block.

BTW, it is not executed because there is no reason for it to fire up. :)

Xliff: That kinda was my point. I just wanted that to be clear. :)

I wonder, would it be so hard to turn a Fatal exception back to a Warning and vice versa? I can can rationalize a use-case for such. Would it help to formalize, here?

lizmat: You can already turn a warning into an exception:
````
$ raku -e 'CONTROL { .throw }; warn "foo"'
foo
  in block <unit> at -e line 1
````

Comment (vrurg) [reply src/core.c/Exception.pm6:76]: @lizmat Yes and no.

```
⇒ raku -e 'CATCH { say "CATCH: ", .raku }; CONTROL { say "CONTROL: ", .raku;  .throw }; warn "foo"; say "done"'
CONTROL: CX::Warn.new(message => "foo")
foo
  in block <unit> at -e line 1
```

Comment (lizmat) [reply src/core.c/Exception.pm6:76]: That's just because you put the CONTROL and CATCH in the same scope:
````
$ raku -e 'CATCH { say "CATCH: ", .raku }; { CONTROL { say "CONTROL: ", .raku;  .throw }; warn "foo" }; say "done"'
CONTROL: CX::Warn.new(message => "foo")
CATCH: CX::Warn.new(message => "foo")
foo
  in block  at -e line 1
  in block <unit> at -e line 1
````

Comment (vrurg) [reply src/core.c/Exception.pm6:76]: Heh, not the first time...

Comment (lizmat) [reply src/core.c/Exception.pm6:76]: ?  The `warn` is first caught with the inner `CONTROL` block, and then rethrown to be caught by the outer `CATCH` block.  If the outer had been a `CONTROL` block as well, it would **NOT** catch it again:
````
$ raku -e 'CONTROL { say "CATCH: ", .raku }; { CONTROL { say "CONTROL: ", .raku;  .throw }; warn "foo" }; say "done"'
CONTROL: CX::Warn.new(message => "foo")
foo
  in block  at -e line 1
  in block <unit> at -e line 1

Comment (vrurg) [reply src/core.c/Exception.pm6:76]: Sorry, I meant it's not the first time I make this mistake. It _feels_ like what's thrown from `CONTROL` is to be caught by `CATCH` until you realize that it's not true. :)

Comment (lizmat) [reply src/core.c/Exception.pm6:76]: Hmm... I'm not sure what would make this PR acceptable or not anymore?

Comment (vrurg) [reply src/core.c/Exception.pm6:76]: I would consider it acceptable when it re-throws a new control, as suggested. Not in the current form.

Comment (lizmat) [reply src/core.c/Exception.pm6:76]: But a new CONTROL would lose the backtrace of the original location???

Comment (vrurg) [reply src/core.c/Exception.pm6:76]: I mentioned that earlier. New warning can carry on the original exception. It's not new to us to report double/nested stack traces in some cases.

@Xliff was also mentioning flip-flopping exception type. But that better be done by re-throwing the held exception when necessary.

Comment (lizmat) [reply src/core.c/Exception.pm6:76]: FWIW, I think it may be *technically* correct to do a double/nested stack trace, but this was supposed to be a convenience method, making it easier to recover from an exception, without throwing away the information such as `try` does.

I guess people will have to continue to use:
```raku
CATCH { .note; .resume }
```
instead of:
```raku
CATCH { .warn }
```

niner:
> I guess people will have to continue to use:
```raku
CATCH { .note; .resume }
```

Sometimes it's better to be explicit anyway. A `.warn' to me wouldn't look 
like we're going to resume where the exception was thrown. I'd rather guess 
that `.warn` would be similar to `.note` just with location information (just 
as warn $foo is similar to note $foo just with more info) and that we'd 
continue after the CATCH block.

Issue #4907, opened 2022-04-25
install-dist.raku puts files into $HOME/.raku/precomp instead of the staging directory


voegelas:
## The Problem

The script `tools/install-dist.raku` from Rakudo 2022.04 does not put precompiled files into the staging directory specified with the command line option `--to`. Instead, the files are stored in the home directory. The script from Rakudo 2022.03 works as expected.

## Expected Behavior

Precompiled files are put into `$DESTDIR/usr/share/perl6/vendor/precomp`, for example.

## Actual Behavior

Precompiled files are put into `$HOME/.raku/precomp`.

## Steps to Reproduce

Unpack any Raku distribution, change to the distribution's directory and run commands such as:

```
mkdir /tmp/foo
env RAKUDO_MODULE_DEBUG=1 RAKUDO_LOG_PRECOMP=1 RAKUDO_RERESOLVE_DEPENDENCIES=0 \
raku /path/to/install-dist.raku --to="/tmp/foo" --for=vendor --from=.
find /tmp/foo
find ~/.raku -mmin -15
```

The directory `/tmp/foo/precomp` will be empty. 

## Environment
* Operating system: Fedora 36
* Compiler version: 2022.04

dod38fr:
I have a similar behavior when package JSON::Fast. On the other hand, I've no problem with JSON::Marshall

amano-kenji:
Use a more recent version.

Here's how I make rakudo packages for gentoo linux.

```
env RAKUDO_RERESOLVE_DEPENDENCIES=0 \
/usr/share/perl6/core/tools/install-dist.raku --to="${D}/usr/share/perl6/vendor" --for=vendor --from=.
```

niner:
@voegelas @dod38fr is this problem still present in the most current release?

dod38fr:
Yes. 

Here's the output of install-dist.raku with raku 2022.07 :

```
$ rm -r /tmp/raku-test/
$ raku --version
Welcome to Rakudo™ v2022.07.
Implementing the Raku® Programming Language v6.d.
Built on MoarVM version 2022.07.
$ raku /usr/share/perl6/tools/install-dist.raku --to=/tmp/raku-test --for=vendor --from=.
 1 RMD: Requested for settings CORE.d
 1 RMD: Loading settings CORE.d
 1 RMD: Loading bytecode from CORE.d.setting.moarvm
 1 RMD: Loading bootstrap Perl6::BOOTSTRAP::v6d
[...]

PR #5067, opened 2022-09-24
Keep version file when cleaning an installation

codesections:
This commit changes the CompUnit::Repository::Staging.remove-artifacts method to not remove a repo's `version` file.  This file is required to allow subsequent installations to the same directory; its absence caused `install-dist.raku` to fail in odd ways (see issue #4907).  From searching the codebase, I believe that `install-dist.raku` is the only user of the `remove-artifacts` method, so this change shouldn't have much impact.

(If anyone knows a reason why removing the `version` file is essential, please speak up; AFAIK, it should be harmless at worst)

### DIFF
--- a/lib/CompUnit/Repository/Staging.rakumod
+++ b/lib/CompUnit/Repository/Staging.rakumod
@@ -91,7 +91,6 @@ class CompUnit::Repository::Staging is CompUnit::Repository::Installation {
     }
     method remove-artifacts(CompUnit::Repository::Staging:D: --> Nil) {
         my $io := self.prefix;
-        really-unlink($io.child("version"));
         really-unlink(.IO) for Rakudo::Internals.DIR-RECURSE(
           $io.absolute, :file(*.ends-with(".lock"))
         );


niner (changes requested):
The reason for install-dist to clean up files like `version` when installing via Staging is that the result of the installation is supposed to be droppable into the actual destination repository. I.e. you install into a staging area (on openSUSE that's e.g. /home/abuild/rpmbuild/BUILD in the build VM or chroot env), collect up the result files into some archive (.dep, .rpm or even a .tar or .zip file), ship them and then extract onto the destination repo (e.g. /usr/share/perl6/vendor). Since this archive is supposed to contain only files belonging to the dist installed, it must not contain repository files like `version`. Otherwise every module package would claim to own the same file which should in truth belong to the rakudo package itself (in the case of a standard repository).

So while this fix may point at what's going on, it cannot be the real solution.

codesections:
Interesting.  So maybe this is more of an issue with how `install-dist` is documented?  The "staging" use is different from the `install-dist.raku --to=/usr/share/perl6/vendor --for=vendor` examples provided.

I guess in that use case, the `/usr/share/perl6/vendor/version` file still needs to be created manually (er, manually by the build system, anyway), but only once?

PR #5072, opened 2022-10-08
Reuse $v-how instead of calling HOW again

usev6:
This won't make much of a difference, but I stumbled upon the repeated call to HOW. To me reusing the already computed variable $v-how looks a bit more consistent.


lizmat (approved):

vrurg (approved):

vrurg:
Thanks! Actually, I spotted this once a few weeks ago too, but was busy doing other stuff and forgot to get back to it.

PR #5080, opened 2022-10-11
[JVM] Make generics work with return type in sigs

usev6:
This intends to port the changes to src/vm/moar/Perl6/Ops.nqp from
https://github.com/rakudo/rakudo/pull/5048 to the JVM backend.

### DIFF
--- a/src/vm/jvm/runtime/org/raku/rakudo/RakOps.java
+++ b/src/vm/jvm/runtime/org/raku/rakudo/RakOps.java
@@ -347,6 +347,8 @@ public final class RakOps {
         return cont;
     }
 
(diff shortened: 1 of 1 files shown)


vrurg:
Sorry, I would withdraw myself from reviewing as my relationship with Java are far from ideal. :) But as long as it passes corresponding specs I'd be happy to have it!

usev6:
That's totally fine -- I didn't want to bother you with Java code. ;)

This makes a couple of currently failing (sub-)tests from S06-signature/type-capture.t work -- with no regressions. So, I'm confident that it is an improvement.

lizmat:
Good enough for me!  :-)  Thanks for keeping up the good JVM work!

PR #5084, opened 2022-10-18
Introduce `Dateish.days-in-year` method

lizmat:
Which can be called both as a class method (while specifying the year) as well as an instance method (without specifying a year).

Made the days-in-month method follow the same pattern: callable as a class method, specifying the year and month, or as an instance method (without specifying any arguments).


tbrowder (approved):
Looks good to me. Thanks.

PR #5088, opened 2022-10-28
[JVM] Fix defaults for typed optional @/% params

usev6:
This intends to port 7148da1c61 to the JVM backend.

Please note the workaround for handling @deprecation during compilation of settings. I'm not really sure why this is needed, but I think that the problem arises from "sub DYNAMIC" having  @deprecation as an optional parameter in src/core.c/stubs.pm6. For some reasons the check for gcx.Positional fails in that situation. My wild and vague guess would be that the check happens too early and some important pieces are still missing.

### DIFF
--- a/src/vm/jvm/runtime/org/raku/rakudo/Binder.java
+++ b/src/vm/jvm/runtime/org/raku/rakudo/Binder.java
@@ -819,6 +819,12 @@ public final class Binder {
         return exploder.explodeFlattening(tc.curFrame, new Object[] { list, hash });
     }
 
+    private static final CallSiteDescriptor parameterizeArray = new CallSiteDescriptor(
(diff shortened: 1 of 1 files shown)


lizmat:
Is this ready to merge?

usev6:
From my perspective it could be merged, yes.

If we find a clean solution for the mentioned workaround (https://github.com/rakudo/rakudo/pull/5088/files#diff-791b4be8677299dbadf1289e8b6eaa0896c5c2330bd38f52fa482092fa88777eR869-R877), that could be changed later on.

usev6:
> Unless someone objects I'll hit the merge button in a few days.

And then real life happened. But I'll merge it now.

PR #5094, opened 2022-11-05
Properly compose 'with[out]' with external 'for'

### LINKED ISSUE #5049: Inconsistency using "with" chained with "for"
## The Problem

I noticed this behavior, demonstrated here using the REPL, but it works the same as a regular program:

```
[0] > my @a = <a b c d e>
[a b c d e]
[1] > .say with $_ for @a
()
[2] > .say if .defined for @a
a
b
c
d
e
[3] > (.say with $_) for @a
a
b
c
d
e
[4] > (.say if .defined) for @a
a
b
c
d
e
```

## Expected Behavior

I was expecting that the line marked as [1] would emit the same output as the other statements.

## Actual Behavior

Apparently "with" works only as a statement modifier while "if" works both ways.

## Environment
* Operating system: Debian sid
* Compiler version (`perl6 -v` or `raku -v`): v2022.07

salortiz:
Early returns prevented the proper composition of `with` and `without` with statements modifiers like 'for` or `given`.

This fixes #5049 and now the [respective test](https://github.com/Raku/roast/blob/master/S04-statement-modifiers/with.t#L195) (todo-ed) in roast pass.
This looks like it was low hanging fruit.


niner:
Is the same issue present in rakuast?

src/Perl6/Actions.nqp
```diff
@@ -1675,27 +1675,27 @@ class Perl6::Actions is HLL::Actions does STDActions {
             $past := $<EXPR>.ast;
             if $mc {
                 if ~$mc<sym> eq 'with' {
-                    make thunkity_thunk($/,'.b',QAST::Op.new( :op('call'), :name('&infix:<andthen>')),[$mc,$<EXPR>]);
-                    return;
+                    $past := make thunkity_thunk($/,'.b',QAST::Op.new( :op('call'), :name('&infix:<andthen>')),[$mc,$<EXPR>]);
```
Kaiepi, line 1678: Can these `make` calls be removed alongside the early `return`? There's another at the end of the method.

salortiz: Sure, I forgot these.

salortiz:
@niner
> Is the same issue present in rakuast?

Yes, I just build that branch.
The [code](https://github.com/rakudo/rakudo/blob/rakuast/src/Perl6/Actions.nqp#L1678) in Perl6/Actions.nqp is the same.

niner:
Perl6::Actions is not used anymore in the RakuAST compiler frontend. However the branch still contains both the old and the new frontends. To test on rakuast, build rakudo normally from that branch and then run your test code with RAKUDO_RAKUAST=1 set in the environment.

niner:
Forgot to mention: the RakuAST compiler frontend can be found in src/Raku. With grammar, actions and AST classes.

salortiz:
@niner,
Just tested the branch as instructed and found many problems with these constructs. So, not the same issue but some other related:
```
[0] > my @a=<a b c d e>
[1] > .say with $_ for @a;
Local 'with_tested_1' already declared
[2] > .say if $_ for @a;                      
Cannot map a Array using a Slip
Consider using a block if any of these are necessary for your mapping code.
```
If needed, please let's move this thread elsewhere to avoid distracting this PR.

Kaiepi (approved):
Looks unobtrusive enough a fix for this whereabouts.

lizmat:
I will merge this after the next release.  It feels it could have some ecosystem fallout, and I don't want to make life on the release manager any more difficult right now.

PR #5095, opened 2022-11-05
Implement IO::Path.chown and chown()

lizmat:
Thanks to the work of Daniel Green++, NQP now has a nqp::chown op that can be used to change owner and group of a given path.

Both the subroutine as well as the method version take a :uid and a :gid named argument that should be specified with the integer value of the uid/gid to change.  The default for :uid and :gid are the current setting of the path (aka, omitting will not change the value for that aspect of the file).

IO::Path.chown will throw a new X::IO::Chown exception if failed. chown() will just return the paths of which the chown succeeded, in correspondence with the other IO::Pathy subs.


src/core.c/Exception.pm6
```diff
@@ -796,6 +796,15 @@ my class X::IO::Chmod does X::IO {
     }
 }
 
+my class X::IO::Chown does X::IO {
+    has $.path;
+    has $.uid;
+    has $.gid;
+    method message() {
+        "Failed to owner of '$.path' to $.uid/$.gid: $.os-error"
```
MasterDuke17, line 804: Looks like this is missing a “change”.

src/core.c/IO/Path.pm6
```diff
@@ -490,6 +490,27 @@ my class IO::Path is Cool does IO {
         }}
         nqp::chmod($.absolute, nqp::unbox_i($mode));
     }
+    method chown(IO::Path:D: :$uid is copy, :$gid is copy --> True) {
+
+        my str $path = self.absolute;
+        $uid = $uid.defined
+          ?? $uid.Int
```
ugexe, line 497: These should be `UInt`. see: https://github.com/MoarVM/MoarVM/pull/1725

ugexe (dismissed):

src/core.c/IO/Path.pm6
ugexe: Nevermind, `-1.UInt` gives `-1` so that wouldn't matter

salortiz: The intent is achieved with `?? ($uid +& 0xffffffff)` and in fact safer.

PR #5097, opened 2022-11-06
Introduce RAKUDO_GC_EVERY environment variable

lizmat:
This affects whether a GC run will be forced for every N ThreadPoolScheduler supervisor runs (which is usually at about 100/second).  A value of 1 thus would force GC runs 100x/second, whereas a value of 100 would force a GC run every second.

Note that for this to work, the supervisor must actually be running. The simplest way to force that is, is doing a `start { }`.

The default for RAKUDO_GC_EVERY is effectively to never do an additional GC run.

One could think about starting the supervisor automatically if RAKUDO_GC_EVERY is explicitely specified.

### DIFF
--- a/src/core.c/ThreadPoolScheduler.pm6
+++ b/src/core.c/ThreadPoolScheduler.pm6
@@ -596,6 +596,9 @@ my class ThreadPoolScheduler does Scheduler {
                 my num $per-core;
                 my num $per-core-util;
                 my num $smooth-per-core-util = 0e0;
+                my int $ticks;
+                my int $gc-every =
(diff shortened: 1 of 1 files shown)


niner:
I discussed this with @lizmat on IRC. My argument against this feature is that it's not worth the added maintenance cost as it's trivial and more flexible to do in the affected application itself:
```
# If the app is using the thread pool already:
start { loop { $*VM.request-garbage-collection; sleep 0.01 } }
# If you want to avoid the thread pool:
Thread.run({ loop { $*VM.request-garbage-collection; sleep 0.01 } })
# If you want to avoid threads altogether, add this to your main loop:
$*VM.request-garbage-collection;
```

PR #5098, opened 2022-11-11
Actions: Avoid repeated $<sym> eq 'S'

salortiz:
Promised cleanup

### DIFF
--- a/src/Perl6/Actions.nqp
+++ b/src/Perl6/Actions.nqp
@@ -9098,10 +9098,12 @@ Did you mean a call like '"
             QAST::IVal.new( :value($samemark) ),
         );
 
+        my $Is_S := $<sym> eq 'S';
+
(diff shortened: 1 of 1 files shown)


src/Perl6/Actions.nqp
```diff
@@ -9185,16 +9187,14 @@ Did you mean a call like '"
                 ),
 
                 # The result of this operation.
-                $<sym> eq 's'
-                ?? QAST::Var.new( :name('$/'), :scope('lexical') )
-                !! QAST::Var.new( :name($S_result), :scope('local') ),
+                $Is_S
+                ?? QAST::Var.new( :name($S_result), :scope('local') )
```
lizmat, line 9191: Is this reversal of the condition intentional?

salortiz: Yes, as the original test was for 's', not 'S'

PR #5105, opened 2022-11-21
Fix smartmatch over topic

vrurg:
Previously `ACCEPTS` was caled inside topic temporarization causing `$rhs` being an expression containing a topic to use the temporary one bound to RHS.

Cases not resolved:

    "aaa" ~~ { $_ } given /\w/; # "aaa"

Because Code.ACCEPTS doesn't boolify

The following is currently wrong too:

    "aaa" ~~ {$_}() given rx/\w+/

The topic must pick up the temporary value but `given` one is taken instead.

### DIFF
--- a/src/Raku/ast/expressions.rakumod
+++ b/src/Raku/ast/expressions.rakumod
@@ -233,21 +233,24 @@ class RakuAST::Infix is RakuAST::Infixish is RakuAST::Lookup {
 
     method IMPL-SMARTMATCH-QAST(RakuAST::IMPL::QASTContext $context, Mu $left-qast,
             Mu $right-qast, int $negate) {
+        my $rhs-local := QAST::Node.unique('sm-rhs');
         my $accepts-call := QAST::Op.new(
             :op('callmethod'), :name('ACCEPTS'),
-            $right-qast,
-            QAST::Var.new( :name('$_'), :scope('lexical') )
-        );
+            QAST::Var.new(:name($rhs-local), :scope<local>),
+            QAST::Var.new( :name('$_'), :scope('lexical') ));
         if $negate {
             $accepts-call := QAST::Op.new(
                 :op('hllbool'),
                 QAST::Op.new(
                     :op('not_i'),
-                    QAST::Op.new( :op('istrue'), $accepts-call )
-                )
-            );
+                    QAST::Op.new( :op('istrue'), $accepts-call )));
         }
-        self.IMPL-TEMPORARIZE-TOPIC($left-qast, $accepts-call)
+        QAST::Stmts.new(
+            QAST::Op.new(
+                :op('bind'),
+                QAST::Var.new(:name($rhs-local), :scope<local>, :decl<var>),
+                $right-qast ),
+            self.IMPL-TEMPORARIZE-TOPIC($left-qast, $accepts-call))
     }
 
     method IMPL-LIST-INFIX-QAST(RakuAST::IMPL::QASTContext $context, Mu $operands) {
--- a/src/Raku/ast/expressions.rakumod
+++ b/src/Raku/ast/expressions.rakumod
@@ -85,7 +85,7 @@ class RakuAST::Term is RakuAST::Termish {
 }
 
 # Marker for all kinds of infixish operators.
-class RakuAST::Infixish is RakuAST::Node {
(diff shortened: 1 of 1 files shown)


vrurg:
Needs more work. Looks like the correct implementation cannot avoid run-time checking of the RHS and/or the result.

salortiz:
I haven't looked into `rakuast` branch in detail, but your example is not clear to me since it seems to pretend `~~` to have higher precedence than the statement modifier.
In master the following seem to me to produce what is expected
```raku
"f$_" ~~ 'fo' given 'o'; # True (given topic affects the whole expression)
'fo' ~~ ("f$_" given 'o'); # True (given topic is for RHS)
$_ ~ 'f' ~~ 'of' given 'o'; # True
$_ ~ 'f' ~~ /of/ given 'o'; # ｢of｣
$_ ~ 'f' ~~ ($_ given 'of') given 'o'; # True
$_ ~ 'f' ~~ ($_ given /of/) given 'o'; # ｢of｣
$_ ~ 'f' ~~ ({ $_ } given /of/) given 'o'; # ｢of｣
```
Am I missing something?

vrurg:
@salortiz Apparently, none of your examples are wrong. And no, I'm not assuming higher precedence for smartmatch operator. Here is what must happen:

With `"aaa" ~~ { $_ } given /\w/;` the block is used as RHS object, its ACCEPTS method takes RHS and passes it into the block as the only argument. The RHS becomes the topic within the block and gets returned as-is. The return value is used as smartmatch result. The only problem here: the result has to be boolified and it is currently not in RakuAST.

With `"aaa" ~~ {$_}() given rx/\w+/` things are a little bit different because we force-invoke the block and then use its result as LHS. There are no arguments to the block invocation therefore $_ is lexically resolved to the temporary topic of smartmatch. That's why `given` doesn't affect the block's topic and the expression result overall. 

What is confusing to me is the case of `"aaa" ~~ {note .raku; $_}($_) given rx/\d+/` vs. `"aaa" ~~ $_ given rx/\d+/`. I think `$_` resolving to smartmatch topic in the first case is a bug stemming from the complexity of would be correct implementation.

salortiz:
@vrurg, Thanks for your comments.
What confused me was the `given...` in your examples, since its effect can only be visible in the LHS of `~~`, not its RHS.

And yes, in rakuast it is necessary to boolify Block's ACCEPTS.

On the other hand, I don't see any bug in the  `"aaa" ~~ {note .raku; $_}($_) given rx/\d+/` case since in `'aaa' ~~ foo($_)`   the argument `$_` should be, in any case, the temporary topic, right?

In fact, I can't think of any way that within `foo` you can access the topic, now shadowed, set by any outer `given …`.

vrurg:
Try `say ("aaa" ~~ $_ for rx/\w+/, rx/\d+/)` then. :)

salortiz:
Ok, I see

vrurg:
> But then, how we explain: `say ("aaa" ~~ ($ = $_) for rx/\w+/, rx/\d+/)` ?

As a bug.

vrurg:
> Ok, I see. But then, how we explain: `say ("aaa" ~~ ($ = $_) for rx/\w+/, rx/\d+/)` ?

I just have realized that the behavior is specced. So, despite my view over it, #5106 will cover it. But then, speaking of `"aaa" ~~ $_ for rx/\w+/, rx/\d+/` – it was working accidentally. Since it is not specced, it probably makes sense to make it go away. Or otherwise a special case would be required to implement it which would be a hack.

salortiz:
The composition of topicalizers is a complex issue and is practically ignored by synopses, but the topicalization effect of `~~` is the most localized. That's why I was adamant that outer topicalizers couldn't (shouldn't) modify or have an effect on the RHS of ~~.
But when you showed me the example with `for`, I had no choice but to shut my mouth because it was completely unexpected to me.
In hindsight I now think that is a bug, useful but still a bug. We could probably get the outer topic, when needed, with something like `$OUTER::_`. (I understand that you have a PR on those issues.)
I hope to have the chance soon to collaborate in the rakuast branch.

vrurg:
> We could probably get the outer topic, when needed, with something like `$OUTER::_`.

No, you can't. Smartmatch is replacing the topic altogether and restores it back when done. For `OUTER` to work a new lexical context would be required, apparently, which is costly.

PR #5113, opened 2022-11-29
Avoid loss in precision on FatRat .Str conversion

### LINKED ISSUE #5108: FatRat loss in precision from String
This occurs at the boundary of what a Rat can handle.

```
say '-0.1234567890123456774144'.FatRat;
#-0.123456789012345677   # Incorrect. Lost 4144

say '-0.123456789012345677414400000000000001'.FatRat;
# -0.123456789012345677414400000000000001   # Correct. Even larger value is maintained
```

```
Welcome to Rakudo™ v2022.07.
Built on MoarVM version 2022.07.
```

salortiz:
Get some more digits for the case when denominator length is smaller than expected.
Fixes #5108

In a FatRat it is common for the denominator to be a power of 10 with as many digits as there are significant digits represented in decimal. But in those "lucky" cases where the denominator is significantly smaller, you end up losing some digits during the conversion, as it happens in the issue annotated.

Detailed analysis of the denominator and fractional part would be expensive, so asking for a few more digits, which will be trivially removed by Rational!STRINGIFY, avoids performance loss.

I add four more in the logic used in denominators of length less than six.

### DIFF
--- a/src/core.c/Rational.pm6
+++ b/src/core.c/Rational.pm6
@@ -137,7 +137,7 @@ my role Rational[::NuT = Int, ::DeT = ::("NuT")] does Real {
                          ?? 6
                          !! (nqp::chars(nqp::tostr_I($!denominator))
                               + nqp::chars(nqp::tostr_I(whole))
-                              + 1
+                              + 5
                             )
                    )
               !! nqp::islt_I($!numerator,0)                      # no fract val


lizmat:
@jdv if you're comfortable with merging this before the release, I'm ok with it as well.

PR #5114, opened 2022-11-30
Simplify X::TypeCheck::Assignment for containers

### LINKED ISSUE #5110: Clarity in type errors for List and Map variables
I would like type error messages to have more information when a type error occurs inside a List or Map.

For the type error `my Map %map = foo => "bar";`, instead of "Type check failed in assignment to %map; expected Map but got Str ("bar")", I propose the following possible changes:

- expected Map of Map but got Map of Str ("bar")
- Type check failed in assignment within %map
- expected element Map but got Str ("bar")

Even better, in the case of a type error concerning a Map-typed Map, or a List-typed List, "Map/List types refer to their contents, not the Map/List itself". And then a link to the docs, https://docs.raku.org/language/list#Typing.

## Reasoning

I found myself in a situation where I didn't understand how map typing works (I had searched for type parameterisation but didn't find results). Due to this, I had misled myself into something like the following:

```raku
class X {
    ....
    has Map %map; # supposed to contain many types
    
    submethod BUILD(:%map, *%and-other-stuff) {
        %!map = Map.new(%map);
        ...
    }
}
X.new(map => %(foo => 'bar'));
# Type check failed in assignment to %map; expected Map but got Str ("bar")
```

Given the Raku container type syntax was unknown and alien to me, I came up with all sorts of wild theories to explain this error. Maybe it was interpreting my inner hash as a named argument? Or maybe it was being interpreted as a pair? Was I misusing BUILD? Maybe assigning to an immutable Map class field is not allowed? Can I even assign a new object in this context? On and on.

My point being, if I had just a little extra information hinting at how the type was being interpreted, I may have figured it out.

I think most of my confusion came from 'failed in assignment to %map' part. This seems to imply that my problem was with the map itself, not the contents. This was compounded that when the docs discuss map assignment, it's in the context of assigning a whole map to a variable: https://docs.raku.org/language/hashmap#Hash_assignment.

(Issue description edited for clarity)

michaelruigrok:
For container assignment type errors, change the message from: "Type check failed in assignment to an element of <symbol>; expected <type> but got <type> (<value>)" to 
"Type check failed for an element of <symbol>; expected <type> but got <type> (<value>)".

This removes possible confusion about assignment syntax. Possibly resolves  #5110.

This tweaks #5112 to try resolve any lingering confusion for people unfamiliar with container type declarations in Raku.

### DIFF
--- a/src/core.c/Exception.pm6
+++ b/src/core.c/Exception.pm6
@@ -2724,8 +2724,11 @@ my class X::TypeCheck::Assignment is X::TypeCheck {
     method operation { 'assignment' }
     method message {
         my $symbol := $!symbol // $!desc.name;
-        my $to = $symbol.defined && $symbol ne '$'
-            ?? " to {"an element of " if $symbol.starts-with("@" | "%")}$symbol" !! "";
+        my $location = !$symbol.defined || $symbol eq '$'
+            ?? "in assignment"
+            !! $symbol.starts-with("@" | "%") 
+                ?? "for an element of $symbol"
+                !! "in assignment to $symbol";
         my $is-itself := nqp::eqaddr(self.expected, self.got);
         my $expected = $is-itself
             ?? "expected type $.expectedn cannot be itself"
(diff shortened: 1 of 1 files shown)


lizmat:
Thanks!

PR #5115, opened 2022-11-30
Introduce Least / Most

lizmat:
Work in progress as it unearthed a number of issues.

Introduce the dynamic variables `$*LEAST` and `$*MOST` as identifiers for the smallest and largest possible value (like -Inf and Inf, but without any typing).

In 6.c/d these will be initialized to `-Inf` and `Inf` respectively.

In 6.e, two new constants are made available:  `Least` and `Most`.  These are just instances of `Any` that can be used for dispatching.  In 6.e `$*LEAST` is initialized to `Least`, and `$*MOST` is initialized to `Most`.

Furthermore, candidates for infix operators are added to properly handle the `Least` and `Most` values, so that any comparison with `Least` and/or `Most` will yield the correct `Boolean` or `Order` enum.


src/core.e/LeastMost.pm6
```diff
@@ -0,0 +1,111 @@
+constant Least = Any.new;
```
JJ, line 1: Why not use -Inf here? I get this is a stub, and the real value of $*LEAST is registered somewhere else. Still, it would make it a bit more consistent. Such as it is now, they are the same.

src/core.e/LeastMost.pm6
```diff
@@ -0,0 +1,111 @@
...
+constant Most  = Any.new;
```
JJ, line 2: And +Inf here?

src/core.e/LeastMost.pm6
```diff
@@ -0,0 +1,111 @@
...
+
+$*LEAST = Least;
+$*MOST  = Most;
+
+multi sub infix:<cmp>(Least, Least --> Same) { }
```
JJ, line 7: Bear with me here. This will work as long as these have the same value as the constants. So I don't really see what this is buying you. If you initialize it to -Inf, +Inf you get the same for free...

JJ (approved):
The only nit I have here is that there's no way to guarantee that `$*LEAST < $*MOST` dynamically, but looks good overall.

salortiz:
Since 6.e is design terrain, let me make a few small suggestions on the fly:

- If `Least` and `Most`, visible to the user primarily as a response to `max` and `min`, inherit from `Nil` instead of `Any` they automatically become soft failures, transparent to return type constraints, `undefined` and trivially testable, for example with `with`, `andthen`, etc.

That directly fixes the issue with the code presented by the OP.

- If `Least cmp -Inf === Same` and `Most cmp Inf === Same` they compose well with other uses of `±Inf`
- `Range` creation should, at least, convert `Most` into `Inf`, etc.

I'm mulling also...

PR #5117, opened 2022-12-02
RakuAST origins (provenance)

vrurg:
Implement provenance support for RakuAST nodes.

Here is some concepts behind this implementation.

- Terminology. 'Origin' has been chosen instead of 'Provenance' because it is shorter and shortens even further without creating confusions. But some names used are still leaving me unsatisfied and whish there be better naming options, but I failed finding them.

- I decided to ditch NQP-provided services altogether and make RakuAST self-sufficient. NQP is Rakudo's implementation detail. Therefore RakuAST API must be independent of it.

   From implementation point of view, NQP relies upon World-based compilation process which is not implemented by RakuAST. Thus instead of simulating a World for NQP it is better to have all infra implemented by RakuAST.

- Due to the nested nature of syntax tables may not be sufficient to hold all the information needed. In particular, for a multi-line statement where we report the line number of its start it might be rather complicated in some respects. I've been there, tried it, and ditched. The current design binds origin data to each node and thus preserves the tree structure. It also helps with position search.

- To know what line to be reported in case of an error a 'key node' term is used. So, if a run-time error is caused by a sub-node for a multi-line statement the line reported will be not the one, where the subnode originates from, but the first line of the statement.

   Wether a node is a key one is technically determined by concreteness of `$.nestings` attribute of a `RakuAST::Node` `origin` attribute. Logically it is defined by the grammar. Currently almost any statement within a statement list becomes a key (with some exceptions). Type of a node is irrelevant.

- User-facing API is `RakuAST::Origin::Match` which kind of implements a subset of interfaces of `NQPMatch`/`Match` objects. Instances of the class are to be submitted to the backend compilers via `QAST::Node.node()` interface. They currently lack support for it and rely upon `HLL::Compiler`. The primary problem is about file names and `#line` directives.

   Adding support of `RakuAST::Origin::Match` to NQP is just a couple of lines for MoarVM compiler. Haven't checked out with JVM but it must not be much more complicated. Basically, it winds down to testing for presence of necessary methods on `.node()` object.

- The model chosen allows to integrate AST nodes originating from other source files. Since every node can indirectly point to a `RakuAST::Origin::Source` instance it shouldn't be a problem.

   Thing are a little more complicated for user-facing API because currently the original file name is either picked from `%*COMPILING<%?OPTIONS><source-name>` or from `$?FILES`. Perhaps additional source in form of a dynamic variable or a named argument of `.new` must be introduced. Not a big deal.

Hopefully, I missed nothing.

Related PR for NQP: raku/nqp#786


vrurg:
I made some improvements to the RakuAST compiler part of this PR because I needed/wanted them while was working on it. Hopefully, this is not a problem. :)

src/Raku/ast/base.rakumod
```diff
@@ -217,28 +217,68 @@ class RakuAST::Node {
...
-                    $dump := $dump ~ $child.dump($indent + 4);
-                });
-            });
+        if nqp::isconcrete($!origin) {
+            @markers.push('𝄞') if $!origin.is-key();
         }
+        nqp::join('', @markers)
+    }
+
+    method dump-extras(int $indent) { '' }
```
MasterDuke17, line 230: Shouldn’t this be something like ‘nqp::x(“ “, $indent)’?

src/Raku/Actions.nqp
```diff
@@ -2150,6 +2150,15 @@ class Raku::Actions is HLL::Actions does Raku::CommonActions {
     method pod_block:sym<finish>($/) {
         $*CU.replace-finish-content($*LITERALS.intern-str(~$<finish>));
     }
+
+    method comment:sym<line_directive>($/) {
+        my $origin-source := $*ORIGIN-SOURCE;
+        my $orig-line := $origin-source.original-line($/.from());
+        $origin-source.register-line-directive(
+            $origin-source.original-line($/.from()),
```
MasterDuke17, line 2158: Shouldn't this just be the `$orig-line` declared above?

src/Raku/Grammar.nqp
```diff
@@ -451,6 +451,13 @@ role Raku::Common {
             }
         }
     }
+
+    # Provide parent's rule/token @*ORIGIN-NESTINGS to ease and unitfy creating a stack of key AST nodes.
```
MasterDuke17, line 455: "unitfy" -> "unify"

tools/build/raku-ast-compiler.nqp
```diff
@@ -56,20 +61,25 @@ grammar RakuASTParser {
...
-        | $<variable>=['$' <.identifier>]
+        | $<variable>=[<.sigil> '*'? <.identifier>]
         | <string>
         | ['/' <-[/]>+ '/' || '//' || '/' <?before \s* [\d | '$']>] # regex or // operator
         | $<numeric>=[ \d+ ['.' \d*]? [<[eE]> \d+]? ]
-        | $<paren>='(' <nqp-code> [ ')' || <.panic('Missing )')> ]
-        | $<brace>='{' <nqp-code> [ '}' || <.panic('Missing }')> ]
+        | $<paren>='(' <nqp-code> [ ')' || {} <.panic('Missing ) for opening ( at line ' ~ self.line-of($<paren>))> ]
+        | $<brace>='{' <nqp-code> [ '}' || {} <.panic('Missing } for opening { at line ' ~ self.line-of($<brace>))> ]
+        | $<brckt>='[' <nqp-code> [ ']' || {} <.panic('Missing ] for opening [ at line ' ~ self.line-of($<brckt>))> ]
```
MasterDuke17, line 80: While it's nice that it's just five letters, I'd prefer the full name (i.e., 'bracket'). It just makes it harder to search for if anyone ever wants to find the implementation.

src/Raku/ast/base.rakumod
vrurg: You mean like get the indent as a prepared string? I decided that integer provides more flexibility for custom methods.

src/Raku/Actions.nqp
vrurg: Actually, it's the `$orig-line` which is unnecessary. Fossils of a debug print.

tools/build/raku-ast-compiler.nqp
vrurg: You make me cry in sorrow! :) I'm not a big fan of terrible shortnames, but this time I'm addicted. :)

src/Raku/ast/base.rakumod
MasterDuke17: I mean it isn't using the `$indent` parameter at all.

vrurg: Extras may happen to be a multi-liner. In this case either option must be implemented by the caller:

- require a positional return value
- split on new lines

Not sure if either is better than passing the indent value. Though positional return doesn't look that terrible. Perhaps I would consider this variant.

MasterDuke17:
I don’t quite follow, but if the parameter is intentionally ignored, there should be a comment saying so (or make it anonymous, but I don’t remember if NQP supports that).Sent from my iPhoneOn Dec 2, 2022, at 5:10 PM, Vadim Belman ***@***.***> wrote:﻿
@vrurg commented on this pull request.



In src/Raku/ast/base.rakumod:
>          }
+        nqp::join('', @markers)
+    }
+
+    method dump-extras(int $indent) { '' }

Extras may happen to be a multi-liner. In this case either option must be implemented by the caller:

require a positional return value
split on new lines

Not sure if either is better than passing the indent value. Though positional return doesn't look that terrible. Perhaps I would consider this variant.

src/Raku/ast/base.rakumod
MasterDuke17: Maybe I'm just not seeing the overall picture, but the comment seems to contradict the implementation? "... expected to respect the indenation level provided with $indent", but the implementation doesn't respect (in fact it completely ignores) $indent. Can you show an example of how it would look?
Unrelated, but "indenation" -> "indentation"

vrurg:
Ok, I think in its current state this PR is almost ready. No backtrace is produced for an exception yet, but all frames are in the Backtrace object and are visible with `--ll-exception`.

src/Raku/Grammar.nqp
```diff
@@ -2037,15 +2036,15 @@ grammar Raku::Grammar is HLL::Grammar does Raku::Common {
...
-        <sym> <.end_keyword> <routine_def('sub')>
+        <sym> <.end_keyword> <routine_def=.key-origin('routine_def', 'sub')>
     }
     token routine_declarator:sym<method> {
-        <sym> <.end_keyword> <method_def('method')>
+        <sym> <.end_keyword> <method_def=.key-origin('method_def', 'method')>
     }
     token routine_declarator:sym<submethod> {
-        <sym> <.end_keyword> <method_def('submethod')>
+        <sym> <.end_keyword> <method_def=.key-origin('method_def', 'submethod')>
```
MasterDuke17, line 2047: Should this be `'submethod_def'`?

src/Raku/ast/base.rakumod
vrurg: This one is a stub and produces no output. Look into _expressions.rakumod_. This is where it does some useful work.

The comment describes requirements to more material implementations of the method.

MasterDuke17:
I don't really mind the naming as is, but an alternate suggestion is 'RakuAST::Sourcing'. From google's definition, "find out where (something) can be obtained."

vrurg:
What would be the attribute name then?

`RakuAST::Sourcing::Source` doesn't look appealing either. ;)

PR #5124, opened 2022-12-10
Add Any.decont and decont() sub

lizmat:
Over the past weeks it has become clear that `<>` as a way to decontainerize, is very hard to understand for beginners, and very hard to make findable in the documentation (because you will mostly find it under postcircumfix handling).

This introduces an explicit `.decont` method on objects, and a `decont(value)` sub that will call the `.decont` method on the given value.  It is set up as a multi, so that classes can implement their own deconting logic.

### DIFF
--- a/src/core.e/Any-iterable-methods.pm6
+++ b/src/core.e/Any-iterable-methods.pm6
@@ -21,6 +21,9 @@ augment class Any {
         snitch-on SELF;
         SELF
     }
+
+    proto method decont(|) {*}
+    multi method decont(Any:) { self }
 }
 
 proto sub rotor(|) {*}
@@ -45,5 +48,8 @@ multi sub snitch(&snitch-on, \SELF) is raw {
     snitch-on SELF;
     SELF
 }
+
+proto sub decont(|) {*}
+multi sub decont($value) { $value.decont }
 
 # vim: expandtab shiftwidth=4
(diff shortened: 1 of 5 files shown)


codesections:
A question: If the goal of `.decont` is to provide a longer-but-more-explicit alternative to `.<>`, do you think we should take the de-huffman-ization all the way to `.decontainerize`?

lizmat:
Possibly.

OTOH, `immutable` also comes to mind.  But that would basically be the functionality provided by https://raku.land/zef:lizmat/immutable

Kaiepi:
If I'm working with `Mu`, I'm still stuck with `<>`/`[]` and their dispatch, but if I'm dealing with `Mu` to begin with, I'm probably dealing with metaobjects in general. Could this be defined under `Mu` or given something or other in `Metamodel::Primitives`?

vrurg:
I just've stumbled upon this PR. I'd say that `.decont` must be a `Mu` method. Deconting low-level objects as well as junctions would be much more requested operation, than deconing eigenstates. Besides, the solution for the latter is as simple as:

```
sub decont(Any \v) is raw { v.decont }
decont(a-junction);
```

And this is a kind of thing we often do with junctions.

dumarchie:
Imo, `decont` should be _just_ an "only" operator. Objects should not have to care whether they're containerized and should not be able to redefine the language-level concept of decontainerization. In any case, a method `.decont` would do the same as `.self`, except maybe if called on a `Failure`.

PR #5125, opened 2022-12-11
Add workflow that fails if pushed to master

JJ:
It still allows to push to master, but it shows a helpful error message to steer in the right direction.


vrurg:
It seems that a hook is not needed. I have added a branch rule which locks the master branch. Experimenting with my own fork of the repository resulted in some glitches as if the rule is not getting applied instantly or something like that. But enforcing with a pull request requirement resulted in the desired effect.

PR #5128, opened 2022-12-12
Try to convert `($a, $b) = ($c, $d)` into `$a = $c; $b = $d`

MasterDuke17:
Right now it only works when the two lists are of the same length and none of the vars on the RHS have the same name as any of the vars on the LHS, because `($a, $b) = ($b, $a)` is not the same as `$a = $b; $b = $a`.

I'd greatly appreciate any thoughts/comments/suggestions.

### DIFF
--- a/src/Perl6/Optimizer.nqp
+++ b/src/Perl6/Optimizer.nqp
@@ -2628,6 +2628,41 @@ class Perl6::Optimizer {
             $visit_children_mode := 'sm_chain';
         }
 
+        # Try to re-write list assignment into a list of assignments, e.g., `($a, $b) = ($c, $d);` into `$a = $c; $b = $d;`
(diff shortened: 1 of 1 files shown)


lizmat:
Two notes:

- Will this handle aliases correctly?
- Does it makes sense to still spend time in Optimizer.nqp while the main branch nowadays includes RakuAST support?

vrurg:
Hm, a good question about aliases:

```
sub swp(\a, \b, \c, \d) {
    (a, b) = (c, d);
}
my $a = 1;
my $b = 2;
swp($a, $b, $b, $a); # ???
```

It's unlikely to be solvable without run-time inspection.

lizmat:
Feels to me this could be a RakuAST tree optimization, so closing this in favour of such an approach.

Please re-open if disagree.

PR #5129, opened 2022-12-12
Make use of line directive information for compile time errors

vrurg:
Whereas `#line` – use it to report line numbers. Helpful when compiling compound files made up of several concatenated sources.

Make use of the lates NQP changes and teach `Backtrace` about them.


vrurg:
And with the changes to RakuAST compiler I have all I wanted from this PR.

PR #5135, opened 2022-12-14
Implement a reliable approach to experimental RakuAST

vrurg:
Do it via compiler this time. No modification to the `experimental.rakumod` is needed except that the module must provide the `:rakuast` tag. It is dummy for now, but might be used in the future for extra useful work.

The actual logic now is to install `Failure`-bound `RakuAST` symbol without the `use experimental :rakuast`. And raise a compile-time error whenever used without the statement.

`RakuAST::Deparse` is back into the core.

Changed the logic of `Rakudo::Internals.DEPARSE` from explicitly printing deparse result to returning it as a string. It is up to the caller what to do with it.

### DIFF
--- a/src/Perl6/World.nqp
+++ b/src/Perl6/World.nqp
@@ -1271,6 +1271,25 @@ class Perl6::World is HLL::World {
                 self.throw($/, 'X::Pragma::UnknownArg', :$name, :$arg);
             }
         }
+        elsif $name eq 'experimental' {
+            # Make sure if an experimental feature requires compile-time pre-processing. Currently used for masking
+            # RakuAST symbol unless `use experimental :rakuast` is used.
+            if $on && nqp::defined($arglist) && nqp::elems($arglist) {
+                my $Pair := self.find_single_symbol_in_setting('Pair');
+                my $Bool := self.find_single_symbol_in_setting('Bool');
+                for $arglist -> $arg {
+                    if nqp::istype($arg, $Pair) {
+                        my $value := $arg.value;
+                        if nqp::istype($value, $Bool) && $value && $arg.key eq 'rakuast' {
+                            $*WANT_RAKUAST := 1;
+                            last;
+                        }
+                    }
+                }
+            }
+            # Load module 'experimental' anyway
+            return 0;
+        }
         elsif $name eq 'lib' {
             unless $on {
                 self.throw($/, 'X::Pragma::CannotWhat', :what<no>, :$name);
@@ -4890,6 +4909,7 @@ class Perl6::World is HLL::World {
     # Checks if a given symbol is declared.
     method is_name(@name) {
         my int $is_name := 0;
+        self."!no_experimental_rakuast"(@name);
         if self.is_pseudo_package(@name[0]) {
             $is_name := 1;
         }
@@ -5149,6 +5169,13 @@ class Perl6::World is HLL::World {
         $value
 
     }
+
+    method !no_experimental_rakuast(@name) {
+        if (@name[0] eq 'RakuAST') && !($*COMPILING_CORE_SETTING || $*WANT_RAKUAST) {
+            $*LEAF.typed_panic('X::Experimental', :feature<RakuAST>, :use<rakuast>);
+        }
+    }
+
     method find_symbol(@name, :$setting-only, :$upgrade_to_global, :$cur-package) {
         my $fullname := join("::", @name);
         if $setting-only && nqp::existskey(%!setting_symbols, $fullname) && !$upgrade_to_global {
@@ -5157,6 +5184,8 @@ class Perl6::World is HLL::World {
 
         # Make sure it's not an empty name.
         unless +@name { nqp::die("Cannot look up empty name"); }
+
+        self."!no_experimental_rakuast"(@name);
 
         if +@name == 1 {
             #note("got a single element argument to find_symbol: " ~ @name[0]);
--- a/src/Perl6/World.nqp
+++ b/src/Perl6/World.nqp
@@ -545,6 +545,8 @@ class Perl6::World is HLL::World {
         $!in_unit_parse
     }
 
+    method have_outer() { $!have_outer }
+
     method add_check($check) {
         @!CHECKs := [] unless @!CHECKs;
         @!CHECKs.unshift($check);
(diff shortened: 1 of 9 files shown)


tools/templates/6.c/core_sources
```diff
@@ -250,6 +250,7 @@ src/core.c/REPL.pm6
 src/core.c/WrappedJSObject.pm6)@
 src/core.c/Rakudo/Metaops.pm6
 src/core.c/Rakudo/Internals/PostcircumfixAdverbs.pm6
+src/core.c/RakuAST/Deparse.pm6
```
zhuomingliang, line 253: 1

src/Perl6/World.nqp
```diff
@@ -4890,6 +4909,7 @@ class Perl6::World is HLL::World {
     # Checks if a given symbol is declared.
     method is_name(@name) {
         my int $is_name := 0;
+        self."!no_experimental_rakuast"(@name);
```
lizmat, line 4912: Wouldn't this cause a serious compile time degradation?  Checking for RakuAST *every* time a name is checked ??

tools/templates/6.c/core_sources
patrickbkr: I don't think this is resolved. The duplicate is still present. Did I misunderstand?

lizmat:
FWIW, I don't see why any further work in this area is needed.  On "main" you cannot access any `RakuAST::xxx` classes, and if you do, you get a perfectly good error message:
```
% raku -e 'say RakuAST::Node'
Must do a 'use experimental :rakuast' to access RakuAST::Node
% raku -e 'use experimental :rakuast; dd RakuAST::StrLiteral.new("foo")' 
"foo"
```
Yes, it's not compile time.  But you would only need to make this error once to fix it everywhere.  I don't think this will be an issue in practice.

How is that achieved?  At the end of the core.c epilogue:
1. Stash away the `RakuAST` type object with `nqp::bindhllsym`
2. Lookup where `RakuAST` lives, and replace that by a special sub-class of `Nil`
3. **ONLY** if the lookup of a `Stash` key fails, check if it is that special sub-class of `Nil` and report the nice error message.
4. When compiling `src/core.e/Formatter`, obtain the `RakuAST` type object lexically so that it can compile properly.
5. Add an `EXPORT::rakuast` package in `lib/experimental` that stores the `RakuAST` type (obtained with `nqp::gethllsym`) in `OUR::`, thus making it completely accessible.

I think this is the least intrusive way to handle this, both at compile time as well as at runtime.

The only thing that feels a *little* dirty to me, is step 2.  But only a little.  And maybe the fact that the `RakuAST`type object is `Nil` without the `use experimental :rakuast;`:
```
 % raku -e 'say RakuAST'
Nil
% raku -e 'use experimental :rakuast; say RakuAST'                 
(RakuAST)
```

And mind you, all of this should become moot in the next 3-6 months (I would hope!)

So unless someone can prove that this approach is wrong (as in: it would allow people to use `RakuAST` classes without using `use experimental :rakuast`), I would like to close this PR and this discussion, and focus on actually working on and with RakuAST classes!

tools/templates/6.c/core_sources
```diff
@@ -250,6 +250,7 @@ src/core.c/REPL.pm6
...
 src/core.c/unicodey.pm6
 src/core.c/RakuAST/Deparse.pm6
```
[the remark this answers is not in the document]
vrurg: Thanks! It was an outcome of a last-minute hastish conflict resolution.

src/Perl6/World.nqp
```diff
@@ -5158,6 +5185,8 @@ class Perl6::World is HLL::World {
         # Make sure it's not an empty name.
         unless +@name { nqp::die("Cannot look up empty name"); }
 
+        self."!no_experimental_rakuast"(@name);
```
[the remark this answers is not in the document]
vrurg: I haven't noticed any significant degradation on core compile time. Besides, there are more concerning techniques like further down this method where `try { self.find_symbol }` is used. Exception handling it self is way more costly. Moreover, in one location in `Actions.nqp` a call to `is_name` is immediately followed by `find_symbol` which is, essentially, nothing else but `find_symbol` duplicated. And that is not to mention that `Grammar.nqp` is using it all over the place and often corresponding `Action` methods use `find_symbol` too – i.e. it would be cases of the same duplication.

vrurg:
> And mind you, all of this should become moot in the next 3-6 months (I would hope!)

Unlikely.

Ok, yes, I was wrong about direct access to `RakuAST::Node`. Perhaps it was a glitch on my own early experimenting. Anyway, look at what my approach brings:

1. Less startup time by avoiding the outers scanning loop
2. Direct access to `RakuAST` symbols in the cores without the need for tricks like `my constant RakuAST = nqp::gethllsym()`
3. User code startup times can be further improved by skipping loading the `experimental` module altogether if it is decided that we don't need it; it is currently used more like a stub to allow extra operations at runtime, if necessary.
4. Better error reporting with the message displayed at compile time wherever possible; runtime error happens only when other symbol resolution would take place at runtime – consistency
5. Same error message as for other experimental features
6. Less tricks and hackery, after all

BTW, you said nothing about the change in `Rakudo::Internals.DEPARSE` semantics. My point is that it is a data manipulation method in first place. And as such it must be a black-box behavior, not a reporter of a kind, like `dd`.

tools/templates/6.c/core_sources
vrurg: > I don't think this is resolved. The duplicate is still present. Did I misunderstand?

@patrickbkr the duplicate is now gone.

vrurg:
I would leave my notes here to make myself clear anyway.

> That "trick" is a legitimate way of handling this situation. And it only causes a tiny, tiny bit of overhead in core compilation.

It is still a trick. Why having it if we can avoid it?

> Optimising for user code startup of code that uses RakuAST is a **VERY** premature optimization, when you realize there hasn't been _any_ optimization done in the RakuAST classes themselves.

I'm trying to avoid affecting startup times of the legacy compiler code, not RakuAST, for now.

> Are you sure? :-)

At least for those, which are reported with the help of `X::Experimental` exception.

> As I said, I only consider point 2 a hack. The rest, most definitely not.

Actually, the special handling of the `Nil`-ish class by its name in `Stash` is a hack too. And that is one very easy to overlook when removing the whole thing for good.

Another thing might be my personal quirk, but I don't like augmentation and try to avoid it whenever possible.

PR #5150, opened 2023-01-05
Update to Unicode 15.0

thundergnat:
Basically just adding Unicode 15.0 database and updating test data pointer.

MoarVM pull request https://github.com/MoarVM/MoarVM/pull/1737 will need to be applied before this will pass testing.


MasterDuke17:
There has been idle speculation about adding some sort of `use Unicode v<n>` capability to Rakudo. If that's ever implemented, would it be easy to backfill in Unicode 14 if we skip straight from 13 to 15? Or should we update to 14 and then immediately to 15?

thundergnat:
There is no infrastructure in place to do that right now. If we do go that route, it will be simple to just download the relevant
data from the Unicode consortium site. There is already a property available through .uniprop() that tells you when (in which revision) a particular codepoint was added. That won't help much with changing composition and grapheme break rules though.

thundergnat:
Something I think would be quite useful, and should be fairly straightforward to add would be a $*UNIVER ($*UNICODE-VERSION ?) variable that you could query to see what version of Unicode we are presently using.

lizmat:
If there's some nqp::code to obtain the unicode version, we could expose that as `$?UNICODE-VERSION`.  It would be a compile time variable, as we cannot switch unicode versions at runtime, can we?

thundergnat:
> It would be a compile time variable, as we cannot switch unicode versions at runtime, can we?

Correct. It would need to be compile time (for the near future at least.) Probably as simple as setting up a dynamic variable, and manually updating it whenever we switch versions.

lizmat:
@thundergnat but how to get at that version number?  Check all the codepoints for that unicode version property and pick the highest?  That feels...  hackish   :-)

lizmat:
See also #5153

lizmat (approved):

lizmat:
Manually committed this in https://github.com/rakudo/rakudo/commit/37826798b4 as this PR was based on the now obsolete master branch

PR #5151, opened 2023-01-05
"numeric comparisons" with coercion rather than overload

2colours:
Overloads of "numeric comparison" operators are not only severely undocumented but they also defeat the purpose of having separate operators based on different types: why have specialized and generic operator types if a randomly chosen one will get overloads anyway?

It seemed sane and the least breaking to keep the behavior but actually make it a "numeric comparison", backed up by a reasonable, public and documented integer representation: days relative to Modified Julian Day.

(It could be discussed if a "numeric comparison" is even adequate for Dates, given that they rather look like custom objects - but that would immediately a breaking change and it's not the right time for rallying off in that direction.)

### DIFF
--- a/src/core.c/Date.pm6
+++ b/src/core.c/Date.pm6
@@ -404,15 +404,15 @@ my class Date does Dateish {
           !! nqp::create(Date)!SET-SELF($!year, $!month, $!day, &!formatter)
     }
     
-    method Int() {
+    multi method Int(Date:D:) {
         self.daycount
     }
     
-    method Numeric() {
+    multi method Numeric(Date:D:) {
         self.daycount
     }
     
-    method Real() {
+    multi method Real(Date:D:) {
         self.daycount
     }
     
--- a/src/core.c/Date.pm6
+++ b/src/core.c/Date.pm6
@@ -404,15 +404,15 @@ my class Date does Dateish {
           !! nqp::create(Date)!SET-SELF($!year, $!month, $!day, &!formatter)
     }
     
-    multi method Int(Date:D:) {
+    multi method Int(Date:D: --> Int:D) {
         self.daycount
     }
     
-    multi method Numeric(Date:D:) {
+    multi method Numeric(Date:D: --> Int:D) {
         self.daycount
     }
     
-    multi method Real(Date:D:) {
+    multi method Real(Date:D: --> Int:D) {
         self.daycount
     }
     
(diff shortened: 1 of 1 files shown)


lizmat:
@2colours ping  :-)

2colours:
@lizmat about what?

lizmat:
> These methods need to be a multi methods with a Date:D: in the signature, otherwise calling .Int, .Numeric or .Real on a Date type object, will start causing odd errors.

2colours:
> > These methods need to be a multi methods with a Date:D: in the signature, otherwise calling .Int, .Numeric or .Real on a Date type object, will start causing odd errors.

You make it look like a quote but where do you take it from?

Anyway, I get the Date:D part but why a multi method? Seems like `pred` is doing just fine as a method. What does that change in the bigger picture?

lizmat:
Did you ever look at https://github.com/rakudo/rakudo/pull/5151 ?

Did you see my comment about the Int / Numeric / Real methods?

That's where I quoted from.

lizmat:
The `multi`s are needed to have the `Any:U` candidate for `Int`, `Numeric` and `Real` catch those cases.

2colours:
> Did you ever look at https://github.com/rakudo/rakudo/pull/5151 ?

Wait what? This is 5151, and your comment came out of the blue, as apparent.

> Did you see my comment about the Int / Numeric / Real methods?

No. Moreover, I still don't, only the "quote".

> The `multi`s are needed to have the `Any:U` candidate for `Int`, `Numeric` and `Real` catch those cases.

So to have the "Use of uninitialized value of type Date in numeric context" message over the message of the failed dispatch?

lizmat:
image

So you didn't see this ^^^  ??

2colours:
image

I _still_ don't see that, and I suspect this has something to do with the "pending" label. I think you need to do something like "finish review" or "submit comments" or whatever Github calls this step.

lizmat:
Aha,  TIL  :-)   Thanks for your patience :-)

2colours:
Now it's marked as resolved, although I didn't actually change it to multi with `Date:D` - have you changed your mind or should I change it anyway?

vrurg:
> should I change it anyway?

You should.

lizmat:
@2colours Thank you for the last change.

However, you may have noticed that after that change, the core setting didn't build.  So that makes this PR unmergeable as it is.

Could you look to see what's going on?

2colours:
It "cannot generate a proto-method in the setting", whatever conclusion one can draw from that. I quickly checked the corresponding DateTime conversion methods and well, they also aren't multis: https://github.com/rakudo/rakudo/blob/be0835f5805c1a37eb020bb454cb6723159ef338/src/core.c/DateTime.pm6#L494.

lizmat:
Implemented in https://github.com/rakudo/rakudo/commit/72856df91b

PR #5152, opened 2023-01-05
Dateish "numeric comparisons" pt 2

2colours:
See also https://github.com/rakudo/rakudo/pull/5151

In fact, there already was a Numeric coercion in place that did return a `Real` value - so really, all it takes to eliminate the "numeric comparison" functions is to add a coercion to `Real` directly.

Also, this further justifies the pt 1 change: for times, a Numeric coercion _is already exposed_.


lizmat:
Implemented in https://github.com/rakudo/rakudo/commit/72856df91b

PR #5157, opened 2023-01-15
Reconsider language versioning and implement language version support in RakuAST

vrurg:
RakuAST currently doesn't support `use v6.x;` pragma. This PR should change this. Alongside the way Rakudo handles that is being reconsidered and prepared for future change of the versioning scheme used.


vrurg:
First stage of changing the approach to versioning is done. RakuAST follows.

niner:
I wonder if instead of the mixin, it'd be better to make LanguageRevision a plain class holding a native integer that `is box_target`. This would avoid the global deoptimization caused by doing a mixin and hide the internal representation a bit more than starting with an Int.

vrurg:
> I wonder if instead of the mixin, it'd be better to make LanguageRevision a plain class

I don't want to make it a public as it is an implementation detail. At the same time it seemed preferable to have it implemented in Raku. With these requirements the class has to be a lexical in a block and be registered with `Metamodel::Configuration` to make it available for other COREs. The registration must happen at runtime. But the constant is a compile-time thing. So, with the following in CORE.c:

```
{
   my class LangRev {...};
   Metamodel::Configuration.set_language_revision_type(LangRev);
}

my constant CORE-SETTING-REV = Metamodel::Configuration.language_revision_type.new(1);
```

the constant would be "created" using `VMNull`.

Neither I can register the class from within its body because it'd be compile time too and would trigger another issue where lexicals of methamodel classes are not serialized. Hence the consequent cores wouldn't have access to the class.

If the requirement of Raku implementation is to be dropped then it can be fully constructed within BOOTSTRAP, of course.

> This would avoid the global deoptimization caused by doing a mixin

How much global is it? If it is about the constant alone then it's not a big deal as it is barely used in a hot-path.

Speaking of boxing, I'm not fully familiar with the details of its implementation. How much interchangeable would it be with `Int` in any arithmetic expression when it comes to unboxing? Like what would I get with `++CORE-SETTING-REV`? I mean, coercion would happen as the worse, but if an attribute is a box target of `int` – will the coercion take place?

niner:
The deoptimization happens when creating the mixing, do it's not that bad. It's just that we have to deoptimize the full call stack at that point as since an object changes its type when being mixed into, all bets are off.

To get the same runtime performance as with an Int, you'd have to implement the same operators for your type. Implementation can be the same as for Int. Maybe some can be even better since you can e.g. assume that the values will always be unsigned.

vrurg:
> To get the same runtime performance as with an Int, you'd have to implement the same operators for your type.

I estimated possible usages for it and concluded that it makes no sense whatsoever. So, this part is now final.

Vanilla `use v6.x` was as easy to get working in RakuAST as adding just three lines now. Should get `use v6.*` and `use v6.x+` ready today or else will have no time until the weekend at least.

vrurg:
At this stage it is passing tests for me and `use v6.x` works with both plain versions and globbings using `+` and `*`.

One thing I don't like a lot is use of `nqp::getcomp` which makes use of concurrent compilers impossible and, worse, it would mange language version on the compiler object from sub-compilations. I'd like to get around this by making compiler instance pass itself into the grammar as a named argument. But this requires changes to NQP and it's too late to have them done by now.

Issue #5141, opened 2022-12-17
Deadlock in supply / whenever chain


patrickbkr:
The following code reliably deadlocks. It's a golf of an occasional Cro::HTTP::Client hang during connection setup.

I'm having a difficult time understanding how the BlockAddWheneverAwaiter is meant to work reliably. I haven't succeeded yet with creating a piece of code that deadlocks without the need for a `start {}` block, but my gut feeling tells me it should be possible to guide the program flow in such a way that the `@run-after` list ends up with the first entry blocking on a lock the second entry holds.

@jnthn Can you shed some light on your original thoughts? Are there invariants (which I fail to see) that should guarantee the algorithm to robustly finish without danger of deadlocking? I 'll continue working on / thinking about this issue, but I could use some help understanding the algorithm.

Relevant commits:
https://github.com/rakudo/rakudo/commit/26a9c313297a21c11ac30f02349497822686f507
https://github.com/rakudo/rakudo/commit/547839200a772e26ea164e9d1fd8c9cd4a5c2d9f

In the following code the comments list the order in which the lines are executed and which BAWAs (BlockAddWheneverAwaiters) are active.

```
my $dummy-supply = supply {
    "log".IO.spurt: "=== DUMMY SUPPLY " ~ $*AWAITER.WHICH ~ "\n", :append;
};
my $side-channel = Supplier.new; 
my $supply0 = supply {
    "log".IO.spurt: "=== SUPPLY 0 ENTER " ~ $*AWAITER.WHICH ~ "\n", :append;                          # 3. BlockAddWheneverAwaiter|432 (supply1)
    emit "foo";
    "log".IO.spurt: "=== SUPPLY 0 LEAVE " ~ $*AWAITER.WHICH ~ "\n", :append;
}       
  
my $supply1 = supply {                                                                                
    "log".IO.spurt: "=== SUPPLY 1 ENTER " ~ $*AWAITER.WHICH ~ "\n", :append;                          # 2. BlockAddWheneverAwaiter|232 (supply2)
    whenever $supply0 {                                                                               # supply1 @run-after pos 1, later -> supply2
        "log".IO.spurt: "=== SUPPLY 1 before side-channel " ~ $*AWAITER.WHICH ~ "\n", :append;        # 6. BlockAddWheneverAwaiter|992 (supply2 run-after)
        start {
[...]

PR #5158, opened 2023-01-15
Simplify Supply / whenever processing order

patrickbkr:
Previously whenevers tapped their connected supply instantly. To prevent deadlocks caused by loops back to the supply block the awaiter was overwritten to take a continuation whenever a lock was awaited, snipping off the stack up to the latest whenever. Those continuations were then put in a list and run one after the other. This logic can work around all sorts of locking scenarios, but is difficult to predict and can still manage to deadlock in some cases. In addition this changed awaiter also triggered for locks not part of the supply / whenever setup. So external unrelated locks could affect the order of how the code is executed. That made the code even more difficult to predict.

This patch removes all of that functionality. Instead whenevers now never do their tap instantly, but always do it after the supply (or whenever) code block containing the whenever has finished. That way the supply block lock is released before the tapping happens that could potentially also want the supplies lock, circumventing a lot of deadlock potential.

As the new behavior does not at all try to solve deadlocks by reordering execution order it is possible there are situations where the previous lock solver would have resolved a situation, where the new logic deadlocks. I do not know if there are any such practical cases.

Solves #5141

### DIFF
--- a/src/core.c/Rakudo/Supply.pm6
+++ b/src/core.c/Rakudo/Supply.pm6
@@ -1,56 +1,6 @@
 class Rakudo::Supply {
     my constant ADD_WHENEVER_PROMPT = Mu.new;
 
-    class CachedAwaitHandle does Awaitable {
-        has $.get-await-handle;
-    }
-
-    class BlockAddWheneverAwaiter does Awaiter {
-        has $!continuations;
-
-        method await(Awaitable:D $a) {
-            my $handle := $a.get-await-handle;
-            if $handle.already {
-                $handle.success
-                    ?? $handle.result
-                    !! $handle.cause.rethrow
-            }
-            else {
-                my $reawaitable := CachedAwaitHandle.new(get-await-handle => $handle);
-                $!continuations := nqp::list() unless nqp::isconcrete($!continuations);
-                nqp::continuationcontrol(0, ADD_WHENEVER_PROMPT, nqp::getattr(-> Mu \c {
-                    nqp::push($!continuations, -> $delegate-awaiter {
-                        nqp::continuationinvoke(c, nqp::getattr({
-                            $delegate-awaiter.await($reawaitable);
-                        }, Code, '$!do'));
-                    });
-                }, Code, '$!do'));
-            }
-        }
-
-        method await-all(Iterable:D \i) {
-            $!continuations := nqp::list() unless nqp::isconcrete($!continuations);
-            nqp::continuationcontrol(0, ADD_WHENEVER_PROMPT, nqp::getattr(-> Mu \c {
-                nqp::push($!continuations, -> $delegate-awaiter {
-                    nqp::continuationinvoke(c, nqp::getattr({
-                        $delegate-awaiter.await-all(i);
-                    }, Code, '$!do'));
-                });
-            }, Code, '$!do'));
-        }
-
-        method take-all() {
-            if nqp::isconcrete($!continuations) {
-                my \result = $!continuations;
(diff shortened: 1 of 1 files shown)


patrickbkr:
I have just tried running a `make test` and a `make spectest` with this. There are several legitimately failing tests.
This needs more work. I'll see if I can turn this into somewhere useful.

patrickbkr:
I have now changed `done` to only be run after the whenevers have been processed. That gets a lot more tests to succeed.

Now I'm stumped on the following test from `t/spec/S17-promise/nonblocking-await.t`:

    my $kill = Promise.new;
    my $started = Promise.new;
    my $s-address = '0.0.0.0';
    my $c-address = '127.0.0.1';
    my $port = 4893;
    my $ok = start react {
        whenever IO::Socket::Async.listen($s-address, $port) -> $conn {
            whenever $conn.Supply(:bin) -> $buf {
                await $conn.write: $buf;
                $conn.close;
            }
        }
        whenever $kill { done }
        $started.keep;
    }

    await $started;
    my @responses;
    for ^20 {
        react {
            whenever IO::Socket::Async.connect($c-address, $port) -> $client {
                await $client.write('is this thing on?'.encode('ascii'));
                whenever $client.Supply(:bin) {
                    push @responses, .decode('ascii');
                    $client.close;
                }
            }
        }
    }
    $kill.keep;
    lives-ok { await $ok }, 'Server survived';
    ok @responses == 20, 'Got 20 responses from async socket server that does non-blocking await';
    is @responses[0], 'is this thing on?', 'First response correct';
    ok [eq](@responses), 'Rest of responses also correct';

Note the `$started.keep` at the end of the first react block. With this PR it runs before the whenevers are processed and thus cause the connection to not be set up yet. With the new semantics this can't be made work reliably.

But: This kind of code isn't reliable with the current whenever implementation either, because the BlockAddWheneverAwaiters reorder the code when awaits happen, so it's possible for the same problem to occur, just not reliably.

I'm unsure how to proceed.

vrurg:
I would say if you know a better logic for syncing states then just implement it. As long as the test logic takes care of  non-blocking `await` on sockets implementation details are irrelevant.

patrickbkr:
@vrurg: This issue has nothing to do with sockets. Consider:

```
my $s = supply {
    say "processing whenever"
}
react {
    whenever $s { }
    say "end of react block";
}
```
With this PR that code prints
```
end of react block
processing whenever
```
That's the issue I have. The reason this happens is that with this PR whenevers are always only processed *after* the code in the supply block finished running.

Now consider:

```
my $p = Promise.new;
my $s1 = supply {
    await $p;
    say "processing whenever"
};
react {
    whenever $s1 { }
    say "end of react block";
    $p.keep;
}
```

That code prints, both with *and without* this PR:
```
end of react block
processing whenever
```

So the previous implementation isn't immune to this problem either. It just works out most of the time.

What happens on the previous implementation? When - during the processing of a whenever block - an await happens (be it a `protect`, `acquire`, `lock` or plain `await`) a continuation starting at the root of the `whenever` is taken. Only after the supply block itself finished running the continuation is resumed. So the order in which the code is executed is dependent on whether there is any locking going on in code tapped by the whenever.

I really don't have a good idea how this could be resolved.

I don't like the previous implementation as it's a bit unpredictable and fails in corner cases. (See above and https://github.com/rakudo/rakudo/issues/5141)
I don't like this PR because it's unexpected that whenevers are  processed *after* the supply block finished, the syntax indicates otherwise.

patrickbkr:
Closing in favor of #5202.

PR #5161, opened 2023-01-19
Introduce the IGNORE phaser

lizmat:
This phaser will allow you to have the *runtime* completely ignore a statement or a block.  Any code inside the IGNORE phaser **will** be checked for syntax and legality, but will **not** cause any bytecode to be generated.

This allows a developer to de-activate code (semi-)permanently, while ensuring that the code inside this phaser remains in sync with the rest of the code (which would not be the case if the de-activated code would live on as comments).

### DIFF
--- a/src/Perl6/Grammar.nqp
+++ b/src/Perl6/Grammar.nqp
@@ -1393,6 +1393,8 @@ grammar Perl6::Grammar is HLL::Grammar does STD {
     token statement_prefix:sym<BEGIN>   { :my %*MYSTERY; <sym><.kok> <blorst> <.explain_mystery> <.cry_sorrows> }
     token statement_prefix:sym<COMPOSE> { <sym><.kok> <blorst> }
     token statement_prefix:sym<TEMP>    { <sym><.kok> <blorst> }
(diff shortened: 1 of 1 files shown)


raiph:
Just bikeshedding, but while IGNORE is OK, maybe NOTRUN would be better?

FWIW it took me a few minutes to get used to the notion of this being a phaser. I was thinking things like "Is the compiler ignoring it?". And then "Should folk reading the code ignore it?". And, if the latter was true, maybe it shouldn't SHOUT (be in uppercase)? And then I thought a phase is about things happening (out of phase in some way) but here we're naming a phase that doesn't happen. And so on.

Anyhoo, part of the reason I'm suggesting NOTRUN is because I found that made my train of silly thoughts like the foregoing screech to a halt. Thanks for listening. :)

lizmat:
In that train of thought, maybe it should be called `DONTRUN' ??

gfldex:
The compiler `ONLYCHECK`s the code. Quite frankly, this doesn't sound like a phaser but a RakuAST-macro-problem.

PR #5164, opened 2023-01-22
[JVM] Add support for $?UNICODE-VERSION

usev6:
The op "uniprop" isn't available on the JVM backend yet. But since we rely on Java's Unicode support anyway, we could derive the supported Unicode version from the Java version.

### DIFF
--- a/src/core.e/core_prologue.pm6
+++ b/src/core.e/core_prologue.pm6
@@ -21,7 +21,7 @@ my constant $?UNICODE-VERSION = %(
     '17' => '13.0',
     '18' => '13.0',
     '19' => '14.0',
-).{nqp::jvmgetproperties<java.version>.split('.').first}.Version;
+).{nqp::jvmgetproperties<java.specification.version>}.Version;
 #?endif
 #?if !jvm
 my constant $?UNICODE-VERSION = (
--- a/src/core.e/core_prologue.pm6
+++ b/src/core.e/core_prologue.pm6
@@ -5,6 +5,25 @@ use MONKEY;
 my constant CORE-SETTING-REV = 'e';
 
 # This constant specifies the current Unicode version being supported
+#?if jvm
+my constant $?UNICODE-VERSION = %(
+    # Supported Unicode version per Java (major) version.
+    # (cmp. https://docs.oracle.com/en/java/javase/19/docs/api/java.base/java/lang/Character.html)
+     '8' =>  '6.2',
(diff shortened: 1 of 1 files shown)


raiph:
This is presumably the wrong place to have noticed this new variable and write this comment, but why is `$?UNICODE-VERSION` being introduced instead of `$?UNICODE` and `.version` on whatever it contains?

It seems unlikely to me that the only Unicode information we'll ever want is the version number, and unlikely that the indirection will be so prohibitively expensive when we do want more that we have to have lots of `$?UNICODE-this-and-that` variables.

usev6:
I don't feel qualified to answer this question. Maybe @lizmat could add a thought or two?

Personally, when I saw commit be0835f580 (which introduced ```$?UNICODE-VERSION```), it didn't occur to me that we would need more information about Unicode.

We only have a few compile-time variables and most of them are scalars (https://docs.raku.org/language/variables#Compile-time_variables) -- so adding another scalar looked logical to me. (On a different note, it would probably be good to document the new variable there as well.)

JJ:
Why are semantics different depending on the VM? If I'm correct, `UNICODE-VERSION` will map Java version to Unicode version in Java, but apparently will map Unicode version to... the last character introduced in that version?

Whether you follow @raiph 's advice (which seems sensible enough) or not, don't you think semantics should not change depending on the platform? In that case, maybe introduce simply a new variable (or new attribute to the `UNICODE` variable?)

usev6:
I'm a bit at a loss how to meet the raised concerns. I don't want to fiddle with the implementation for MoarVM, but on the other hand it would be very nice to unbreak the JVM build again.

I've proposed an alternative in https://github.com/rakudo/rakudo/pull/5188, which hides the dirty details in an NQP op (for the JVM backend).

Please note, that I forgot Java version 14.0 in my first commit, so I also force-pushed a new version of this PR.

lizmat:
> This is presumably the wrong place to have noticed this new variable and write this comment, but why is $?UNICODE-VERSION being introduced instead of $?UNICODE and .version on whatever it contains?

Indeed a good point.  I've taken @usev6 's PRs to implement `$?UNICODE-VERSION` on the JVM for now, but will use that code in a PR to remove `$?UNICODE-VERSION` in favour of a `Unicode` class.

lizmat:
New PR: https://github.com/rakudo/rakudo/pull/5191

PR #5166, opened 2023-01-24
[JVM] Avoid failures caused by missing RakuAST

usev6:


niner:
I don't think the nqp::dispatch should be replaced. I have already added a very minimalistic syscall implementation in https://github.com/Raku/nqp/blob/main/src/vm/jvm/runtime/org/raku/nqp/runtime/Ops.java#L2410
We should just add the new syscall.

usev6:
I'll have a look and try to add a similar stub for the version with five arguments. Thanks for the suggestion.

PR #5167, opened 2023-01-24
Make (-1).log mathematically correct

lizmat:
Before, this would return NaN, rather than Complex.new(1.log,pi) (which would be more mathematically correct).

This breaks 2 spectests, checking for (-Inf).log basically, and makes one todo test pass.

Inspired by https://wandering.shop/@colomon/109744612694765045

### DIFF
--- a/src/core.c/Num.pm6
+++ b/src/core.c/Num.pm6
@@ -2,6 +2,16 @@ my class X::Cannot::Capture        { ... }
 my class X::Numeric::DivideByZero  { ... }
 my class X::Numeric::CannotConvert { ... }
 
+my constant tau = 6.28318_53071_79586_476e0;
+my constant pi  = 3.14159_26535_89793_238e0;
+my constant e   = 2.71828_18284_59045_235e0;
+
+my constant π := pi;
+my constant τ := tau;
+#?if !jvm
+my constant 𝑒 := e;
+#?endif
+
 my class Num does Real { # declared in BOOTSTRAP
     # class Num is Cool
     #     has num $!value is box_target;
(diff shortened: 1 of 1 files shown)


colomon:
I'm not sure this approach is a good solution!  In general, it's questionable whether real-valued math functions should return complex numbers.  A canonical example of this is the square root of -1:

    [7] >  (-1).sqrt
    NaN
    [8] >  (-1+0i).sqrt
    0+1i

colomon:
The reason you maybe don't want to do that is that if it's a simple math error that gets you to calling `(-1).sqrt` or `(-1).log`, returning a complex number that the user might not be expecting is just going to make the situation more confusing.

My suggestion was adding something along these lines:

     multi sub infix:<**>(Complex $a, Real $b) {
        $a ** ($b + 0i)
    }

where if the user is already in the Complex domain, we do the entire calculation in the complex domain, instead of blowing up if the Real version of `$b.log` returns `NaN`.

pdcawley:
I'd argue that `-1.log` should return the correct complex answer (and so should `-1.sqrt`) and leave the process of turning that into NaN to the point where it's coerced into a Real. Log, Sqrt and others are only well defined if they map from ℝ→ℂ (strictly from ℂ→ℂ, but the reals are embedded in ℂ). Using the type of the _argument_ to force the type of the _result_ feels like an ugly hack.

2colours:
To even this out, I'm with @pdcawley , including the `(-1).sqrt` case. Actually, even `(-1)**(0.5)` gives `NaN` - that would work even in Python.

> Using the type of the _argument_ to force the type of the _result_ feels like an ugly hack.

Absolutely, and this remainds way too much to integer division in C. It's one of the most advertised aspects of Raku that dividing integers will "just do the right thing" - why ignore all the conceptual framework Raku has been built with and fall back to C-style typing habits? :(

JJ (approved):
LGTM

dakkar:
ok, I'm convinced, we should make `log` return complex numbers when needed.

BUT

I fear this has to be gated by a `use v6.e` or something, otherwise we break currently-working code.

See, `my Real $x = log($something)` currently always works, whatever the value of `$something`

after this change, it will explode at runtime with `Type check failed in assignment to $x; expected Real but got Complex` when `$something` is negative.

lizmat:
Ok, will close this PR in a little while and create a new one, making this a 6.e change (with `sqrt` as well).  Keeping it open until then.

lizmat:
New PR at https://github.com/rakudo/rakudo/pull/5192

Issue #5165, opened 2023-01-24
Parameterised hashes implementation detail sometimes leaks out


cfa:
## The Problem
Parameterised hashes are not stored as expected (under certain circumstances).

## Expected Behaviour
Behaviour consistent with normal hashes.

```raku
my %h1      = :foo, :bar; say %(%h1,);  # OUTPUT: «{bar => True, foo => True}␤»
my %h2{Any} = :foo, :bar; say %(%h2,);  # OUTPUT: «{Str|bar => bar => True, Str|foo => foo => True}␤»
```

In the case of `%h2`, keys are apparently produced via `.WHICH`, with hash values consisting of the source hash pairs.

## Steps to Reproduce
The above examples hopefully provide a minimally reproducible case. The issue originally arose when working with hashes created from sets, e.g.

```raku
say %(<foo bar>.Set.Hash, :baz);        # OUTPUT: «{Str|bar => bar => True, Str|foo => foo => True, baz => True}␤»
```

where `<foo bar>.Set.Hash.WHAT` is `(Hash[Any,Any])`.

[This](https://gist.github.com/Whateverable/83607b0e8760cf60f18b694cf9ef2529) bisection suggests that problem may have been introduced in commit ecfb956bcf92488106a85f9462731095db11dfed.

## Environment
* Operating system: macOS 13.1.
* Compiler version (`perl6 -v` or `raku -v`): Rakudo™ v2022.12 .
* Bisection suggests that the issue has been present since Rakudo 2016.07.01.

2colours:
We have investigated it quite a lot yesterday, however I think the direction wasn't completely right yet.

Most of behavior comes from [here](https://github.com/rakudo/rakudo/blob/main/src/core.c/Hash/Object.pm6) or [here](https://github.com/rakudo/rakudo/blob/main/src/core.c/Hash/Typed.pm6) and these files provide good examples how typed hashes can work. Their underlying `$!storage` is a bit tricky, therefore 
```raku
self.STORE_AT_KEY(iterkey_s($iter),nqp::iterval($iter))
```
is not sufficient for copying an element of the Hash.

2colours:
This seems to be the only location where the internal STORE_AT_KEY method is used with low-level representations of a key-value pair so I think it's a good idea to change this call to something else that takes the $iter itself and different Hash implementation can figure out what they need to do with it.

cfa:
[...]

PR #5169, opened 2023-01-26
Fix for storing parameterised hashes as items

2colours:
Belongs to https://github.com/rakudo/rakudo/issues/5165.

STORE_AT_KEY receives high level key-value arguments, they might not be available from the underlying `$!storage` directly. The classes/roles that implement STORE_AT_KEY (and hence know how their `$!storage` is composed), should also know and provide how to interpret a BOOTIter entry back to a high-level key-value pair. Another solution could be to expose and overload STORE_MAP directly for the subtypes.

### DIFF
--- a/src/core.c/Hash.pm6
+++ b/src/core.c/Hash.pm6
@@ -64,10 +64,12 @@ my class Hash { # declared in BOOTSTRAP
         my $iter := nqp::iterator(nqp::getattr(map,Map,'$!storage'));
         nqp::while(
           $iter,
-          self.STORE_AT_KEY(
-            nqp::iterkey_s(nqp::shift($iter)),nqp::iterval($iter)
-          )
+          map.PUT_FROM_ITER(nqp::shift($iter), self)
         );
+    }
+
+    method PUT_FROM_ITER(Mu \iter, \target --> Nil) is implementation-detail {
+        target.STORE_AT_KEY(nqp::iterkey_s(iter),nqp::iterval(iter))
     }
 
     proto method STORE(|) {*}
--- a/src/core.c/Hash/Object.pm6
+++ b/src/core.c/Hash/Object.pm6
@@ -33,6 +33,11 @@ my role Hash::Object[::TValue, ::TKey] does Associative[TValue] {
         )
     }
 
+    method PUT_FROM_ITER(Mu \iter, \target --> Nil) {
+        my \actual-pair = nqp::iterval(iter);
+        target.STORE_AT_KEY(nqp::getattr(actual-pair, Pair, '$!key'), nqp::getattr(actual-pair, Pair, '$!value'))
+    }
+
     method ASSIGN-KEY(::?CLASS:D: TKey \key, Mu \assignval) is raw {
         my \storage  := nqp::getattr(self, Map, '$!storage');
         my \WHICH    := key.WHICH;


niner (changes requested):
The commit message does not explain what the problem is that this commit solves. Since I don't know what the problem is, I cannot tell if your solution is a good fix for it.

Also please follow commit message formatting conventions, i.e. wrap lines longer than 80 characters.

2colours:
I think both the commit message and the linked issue explain the problem, on two different levels. If it's still not clear after reading both, maybe let's talk about it.

I'm going to reformat the commit message - alas, a vanilla nano editor doesn't give me much sense of aesthetics.

niner:
Please include all relevant information in the commit message. This code base and repository lives for decades. We don't know whether the link will still work in 20 years. Also finding out why something is the way it is, is already hard enough, without having to switch between git log in the terminal and the browser to get that additional information.

jonathanstowe:
> Please include all relevant information in the commit message.

Not entirely related to this ticket but I made exactly the same point to my colleagues yesterday morning: if the _why_ isn't in a commit message some future developer may not have any other way of knowing the purpose of a change.

2colours:
Please be more specific about what you consider "relevant information" so that I don't have to guess. For me, saying that this method assumed a certain simple way of obtaining the key and the value - while this is something that all Hash implementations that have their own STORE, decide for themselves, seemed enough and will continue to seem enough, unless you can elaborate on what you miss.

niner:
I don't know what I miss because I don't understand what's there. What does "high level key-value arguments" mean exactly? What would low level be in comparison?

What is the user-visible problem that this solves? E.g. what Raku code gives which result before and after your change and why was the result before the change wrong and the result after the change correct?

Is the new PUT_FROM_ITER method on Hash::Object meant to be a public API? E.g. do we expect subclasses written by users to provide this method?

2colours:
> What does "high level key-value arguments" mean exactly? What would low level be in comparison?

low-level keys and values are the ones stored in the underlying `$!storage` (which is a `BOOTHash` I think, and only knows string keys) - as opposed to fully-fledged Raku `Hash`es that the users access and deal with, the "high level" key and value.

If you can say how to point out this dichotomy in a clear way, I'll be welcome to change it to that. To be honest, I don't think it's my task as the author of a bugfix to explain the already existing essential architecture, and especially not in a commit message.

> Is the new PUT_FROM_ITER method on Hash::Object meant to be a public API? E.g. do we expect subclasses written by users to provide this method?

I mean, the answer is already there...

> The classes/roles that implement STORE_AT_KEY (and
hence know how their $!storage is composed), should
also know and provide how to interpret a BOOTIter
entry back to a high-level key-value pair.

To be honest, I think this should answer all questions of this sort, and exactly them. For what it's worth, STORE_AT_KEY is not something I added and it's not public API. However, should we have another Hash that needs to provide its own STORE_AT_KEY, it should also provide its own PUT_FROM_ITER - that's what the message says. I wanted to highlight the important bits but it says pretty much exactly this.

> What is the user-visible problem that this solves? E.g. what Raku code gives which result before and after your change and why was the result before the change wrong and the result after the change correct?

I have mixed feelings whether that belongs there but I'd definitely expect that you read the issue and could immediately see the Raku code? Also, I hadn't had time to immediately add Roast tests but now I made [a PR for that as well](https://github.com/Raku/roast/pull/828).

> I don't know what I miss because I don't understand what's there.

Well, sorry if this is mean I something - my impression is that you could have read at least the relevant context that is easily available from the issue and the referenced code I didn't write, you'd have quite surely understood the situation, AND then say exactly what you are missing, with that understanding. Instead, what I kinda see is that you are making an unnecessary point from an unrelatable hypothetical perspective. The questions you asked now are useful for communicating but when I see that it should have been answered already, it's hard to make any improvement.

lizmat:
Taken a different approach that would affect performance in most cases a lot less in https://github.com/rakudo/rakudo/commit/932df924f8

Closing this PR

PR #5173, opened 2023-01-28
[JVM] Fix breakage with "use v6.e.PREVIEW"

usev6:
After a recent rework of language versioning (merge commit a815b5ca12) the JVM backend started to throw an error when it encountered

  use v6.e.PREVIEW;

The underlying reason seemed to be a BOOTStr leaking out in two places when trying to create new version objects.

My change to src/Perl6/World.nqp kind of reverts a change made in e1a7ca6227 for the jvm backend. (I've restrained from putting the call to "find_single_symbol_in_setting" above the for loop, so that the diff between the code for the jvm backend and the other backends is smaller.)


usev6:
I didn't run tests for this yet, but wanted to bring it up before being afk for a week.

vrurg (approved):

PR #5174, opened 2023-01-30
Try to be more explicit when exception reporting dies

vrurg:
Currently when this happens at compile time we end up with useless 'No exception handler found for catch' message providing us with no information as to what the original cause was. With this commit we would try to pull out as much info as possible at report it back with no bells and whistles.

### DIFF
--- a/src/core.c/Exception.pm6
+++ b/src/core.c/Exception.pm6
@@ -544,9 +544,9 @@ do {
             }
         }
 
+        my Mu $err := $*ERR;
         try {
             my $v := $e.vault-backtrace;
-            my Mu $err := $*ERR;
 
             $e.backtrace;  # This is where most backtraces actually happen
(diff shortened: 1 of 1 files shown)


niner:
What if this new code itself throws an exception? Currently with RakuAST `.message` still fails occasionally due to missing line number information.

vrurg:
That's why there are 4 `try` in a row. If there is a chance to pull out a message – OK, we get it. And if no then no. I used to see problems with backtrace too. In the worst case scenario you end up with just exception names and even this alone helped me to make a quick and correct guess about the source of problem in my code.

In the worst of the worst scenarios where `$err.say` itself throws the outcome is no different from what we see now: `No exception  handler` and a backtrace which points back into this location in `Exception`, not into the real source of the original error. But chances for such development are nearly 0.

niner:
Oooh...I missed those trys. ENOCOFFEE apparently.

niner (approved):

PR #5178, opened 2023-02-01
Add RakuAST::Type::Subset

ab5tract:
Note: This is currently totally broken. I'm creating the PR so it is easy to discuss the code while learning how to actually accomplish the PR's objective: Add `RakuAST::Type::Subset`. To that extent, it contains debug logging and many, many issues.


src/Perl6/Metamodel/SubsetHOW.nqp
```diff
@@ -71,6 +71,10 @@ class Perl6::Metamodel::SubsetHOW
         }
     }
 
+    method set_where($obj, $refinement) {
```
MasterDuke17, line 74: Is ‘$obj’ intentionally ignored?

src/Raku/ast/package.rakumod
```diff
@@ -125,7 +125,7 @@ class RakuAST::Package
             # We always install it as a lexical symbol.
             $resolver.current-scope.merge-generated-lexical-declaration:
                 RakuAST::Declaration::LexicalPackage.new:
-                    :lexical-name($name.canonicalize),
+                    :lexical-name($final),
```
MasterDuke17, line 128: ‘$final’ only has a value if ‘$name.is-identifier’, is that correct?

ab5tract: IIUC, no. `$final` is also set in the else block below: `$final := nqp::pop(@parts).name;`.

niner (approved):
If you rename IMPL-INSTALL-INTO-PACKAGE this is good to go!

niner (approved):

PR #5179, opened 2023-02-03
Fix test description for unknown modifier

usev6:
It looks like the test description has been copied from the previous
sub-test in commit 3f25ba9212 (which added these tests).


lizmat:
Thanks!

PR #5181, opened 2023-02-03
Add Complex.sign to v6.e

coke:
Inspired by https://www.reddit.com/r/rakulang/comments/10sm0sm/why_isnt_sign_defined_for_complex_numbers/

### DIFF
--- a/src/core.e/Complex.pm6
+++ b/src/core.e/Complex.pm6
@@ -1,5 +1,8 @@
 augment class Complex {
-    method sign() { self / self.abs }
+    method sign(Complex:D: --> Complex:D) {
+        my $abs = self.abs;
+        $abs == 0 ?? 0i !! self / $abs
+    }
 }
 
 # vim: expandtab shiftwidth=4
--- a/src/core.e/Complex.pm6
+++ b/src/core.e/Complex.pm6
@@ -1,0 +1,5 @@
+augment class Complex {
+    method sign() { self / self.abs }
+}
+
+# vim: expandtab shiftwidth=4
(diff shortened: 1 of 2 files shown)


lizmat (approved):

vrurg:
Any reason not to have it in 6.c? It doesn't break backward compatibility.

2colours:
Don't take this at an instant objection but as a general rule, please don't forget that "breaking backwards compatibility" is essentially subjective. Anything that isn't specified to behave in a certain way may be depended upon at any moment - so even though I agree it would be weird to depend on `$number.sign` breakage for `Complex` values, a stricter sense of "compatibility" might demand that behavior like this might not change with a compiler release.

lizmat:
I think we also need people to incentivize to use v6.e  :-)

p6steve:
this commit will fail

```
ok (0 + 0i).sign == 0,      'sign of 0+0i is 0';
```

src/core.e/Complex.pm6
```diff
@@ -0,0 +1,5 @@
+augment class Complex {
+    method sign() { self / self.abs }
```
markjreed, line 2: You need a check for 0 here. sign(0+0i) should be 0, not trigger a division-by-zero error.

coke:
Fixed sign(0i) and added a signature similar to Int's version.

coke:
> ```
> ok (0 + 0i).sign == 0,      'sign of 0+0i is 0';
> ```

Note that the test given above works even in v6.c, since the imaginary part is 0, it gets coerced to something Cool that works. :)

lizmat:
Thanks!

PR #5185, opened 2023-02-04
Fix non-comment typos

shmup:
I skipped over typos belonging to # comments

### DIFF
--- a/src/vm/jvm/runtime/org/raku/rakudo/Binder.java
+++ b/src/vm/jvm/runtime/org/raku/rakudo/Binder.java
@@ -470,7 +470,7 @@ public final class Binder {
                     /* Try to figure out the most helpful name for the expected. */
                     SixModelObject expectedType = null;
                     SixModelObject postConstraints = param.get_attribute_boxed(tc, gcx.Parameter,
-                            "$!post_contraints", HINT_post_constraints);
+                            "$!post_constraints", HINT_post_constraints);
                     if (postConstraints != null) {
                         SixModelObject consType = postConstraints.at_pos_boxed(tc, 0);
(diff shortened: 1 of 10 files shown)


2colours:
Typos in underlying attribute names? 😱 Creepy.

src/vm/jvm/runtime/org/raku/rakudo/Binder.java
```diff
@@ -470,7 +470,7 @@ else if (Ops.istype_nd(decontValue, gcx.PositionalBindFailover, tc) != 0) {
                     /* Try to figure out the most helpful name for the expected. */
                     SixModelObject expectedType = null;
                     SixModelObject postConstraints = param.get_attribute_boxed(tc, gcx.Parameter,
-                            "$!post_contraints", HINT_post_constraints);
+                            "$!post_constraints", HINT_post_constraints);
```
usev6, line 473: Good catch! This could even fix some problems, because ```postConstraints``` will always be null (same goes for the same typo in line 691).

t/12-rakuast/statement-phaser.rakutest
```diff
@@ -34,7 +34,7 @@ subtest 'BEGIN phaser producing a literal expression works' => {
       for EVAL($ast), EVAL($deparsed);
 }
 
-subtest 'Block with CATCH/default handles exception and evalutes to Nil' => {
+subtest 'Block with CATCH/default handles exception and evaluates to Nil' => {
```
usev6, line 37: There seems to be another ```evalutes``` in line 86. So, if you are on it, you could change that one as well.

usev6 (approved):
Looks good to me, thanks!

PR #5187, opened 2023-02-05
[JVM] Set Xmx4G for building Rakudo

usev6:


lizmat:
Any reason not to merge this?

usev6:
In the current situation this can only help IMHO. And there is always the option to revert.
So, I plan to hit the merge button in a few days.

PR #5191, opened 2023-02-06
Replace `$?UNICODE-VERSION` by `Unicode.version`

lizmat:
- creates a Unicode class to keep info about Unicode versions
- add a Unicode.version class method with former $?UNICODE-VERSION value
- add a Unicode.NFG class method indicating support for NFG
- adapt tests accordingly


vrurg:
This one I would still backport to 6.c. Not all code could easily be converted to the new version, which is anyway is still in PREVIEW state. And yet the feature is way too useful for many, I think. If not crucial sometimes.

lizmat:
Disagree.  If we would backport it, then you wouldn't know which release starts supporting it.  And we have too much of that already.

Also, if it was that crucial sometimes, why wasn't this done earlier?

vrurg:
> Disagree. If we would backport it, then you wouldn't know which release starts supporting it. And we have too much of that already.

For some features/fixes I have compiler version check in my code. It is inevitable sometimes:

```raku
    die "Rakudo of at least v2019.11 required to run this version of " ~ ::?PACKAGE.^name
        unless $*RAKU.compiler.version >= v2019.11;
```

> Also, if it was that crucial sometimes, why wasn't this done earlier?

1. "If not" means "I don't know exactly, but it's possible" :)
2. Not everybody qualified enough to implement it
3. Not everybody is ready to ask for a feature and might just skip implementing something or get around the deficiency in their own way

Anyway, I would insist on having this one at 6.c level. It had to be there from the beginning.

lizmat:
Ok, I tried moving the Unicode class into core.c
```
% git diff  
diff --git a/tools/templates/6.c/core_sources b/tools/templates/6.c/core_sources
index 89667242b..cabc590e4 100644
--- a/tools/templates/6.c/core_sources
+++ b/tools/templates/6.c/core_sources
@@ -250,6 +250,7 @@ src/core.c/REPL.pm6
 src/core.c/WrappedJSObject.pm6)@
 src/core.c/Rakudo/Metaops.pm6
 src/core.c/Rakudo/Internals/PostcircumfixAdverbs.pm6
+src/core.c/Unicode.pm6
 src/core.c/unicodey.pm6
 src/core.c/RakuAST/Deparse.pm6
 src/core.c/core_epilogue.pm6
diff --git a/tools/templates/6.e/core_sources b/tools/templates/6.e/core_sources
index bdd2ca548..56771c3da 100644
--- a/tools/templates/6.e/core_sources
+++ b/tools/templates/6.e/core_sources
@@ -11,4 +11,3 @@ src/core.e/array_multislice.pm6
 src/core.e/hash_multislice.pm6
 src/core.e/Formatter.pm6
 src/core.e/Rakudo/Iterator.pm6
-src/core.e/Unicode.pm6
```
and moving the file from `src/core.e/Unicode.pm6` to `src/core.c/Unicode.pm6`, but got:
```
Stage start      :   0.000
Unhandled exception: getlexrel requires a concrete object with REPR MVMContext, got Null (VMNull)
   at src/vm/moar/dispatchers.nqp:3313  (/Users/liz/Github/rakudo/blib/Perl6/BOOTSTRAP/v6c.moarvm:find-core-symbol)
 from src/vm/moar/dispatchers.nqp:3335  (/Users/liz/Github/rakudo/blib/Perl6/BOOTSTRAP/v6c.moarvm:)
 from SETTING::src/core.c/Exception.pm6:952  (<ephemeral file>:message)
 from src/Perl6/World.nqp:2620  (/Users/liz/Github/rakudo/blib/Perl6/World.moarvm:)
```

I've now spent an hour trying to figure what is tickling this and am now giving up.

vrurg:
I tried it locally, but inserted it after `unicodey.pm6` and it works. Perhaps you got a problem with `uniprop` which is implemented in the file.

lizmat:
Implemented as https://github.com/rakudo/rakudo/commit/db7ca2dd03

PR #5192, opened 2023-02-07
Make log and sqrt handle negative values mathematically correct

lizmat:
- in 6.e only
- make sqrt a multi
- add a new sqrt candidate handling negative values
- add a new log candidate handling negative values

Note that the use of "is default" in these new candidates is needed to resolve dispatch ambiguities.  This is really a stopgap measure for now, hopefully we can dispatch on language version soon!


dakkar:
Nice!

will there be a roast test for this?

lizmat:
@dakkar that would definitely be nice  :-)

PR #5193, opened 2023-02-07
Eradicate knowledge of COMPOSE phaser

lizmat:
There were some speculations, and an attempt at creating a stub
that was supposedly create a NYI error, but didn't.  Meanwhile,
the mainline of a role is run every time it gets composed into
a (new) class, so effectively we already *have* an implicit COMPOSE
phaser.  So there's no need for a COMPOSE phaser, it appears.


vrurg (approved):

Leont:
>  Meanwhile, the mainline of a role is run every time it gets composed into a (new) class

I didn't know that, but that does sound like useful information.

PR #5202, opened 2023-02-11
Fix rare deadlocks during supply setup

patrickbkr:
This fixes rakudo/rakudo#5141 and Raku/problem-solving#364.
`make test` and `make spectest` both pass.

In https://github.com/rakudo/rakudo/commit/26a9c313297a21c11ac30f02349497822686f507
a mechanism to solve deadlocks during supply / whenever setup was
introduced. In that implementation whenever a lock during setup
was hit a continuation would be taken and put into a list for later
processing. That list was then processed in order. If new locks were
encountered they were appended to the list. I was possible to end up
in a situation where the first continuation waited for a lock the
second continuation in the list held.

Later on in https://github.com/rakudo/rakudo/commit/547839200a772e26ea164e9d1fd8c9cd4a5c2d9f
a different mechanism to deal with deadlocks caused by recursion during
supply / whenever processing was introduced. That mechanism is far
simpler and neither involves continuations nor overwriting `$*AWAITER`.

It seems (this is not proven) that a precondition for deadlocks to
appear during supply / whenever setup is recursion locking. So the
second simpler mechanism is able to replace the first. This change does
just that. The first mechanism is removed leaving only the second.

A positive side-effect (apart from solving the rare deadlocks) is
that the whenever setup process is simpler and easier to reason about.
Especially locks unrelated to the supplies do not influence the setup
process anymore.

### DIFF
--- a/src/core.c/Lock/Async.rakumod
+++ b/src/core.c/Lock/Async.rakumod
@@ -150,7 +150,7 @@ my class Lock::Async {
             # Lock is already held on the stack, so we're recursing. Queue.
             $try-acquire.then({
                 self!run-with-updated-recursion-list(&code);
-            });
+            }, :synchronous);
         }
         else {
             # Lock is held but by something else. Await its availability.
--- a/src/core.c/Promise.rakumod
+++ b/src/core.c/Promise.rakumod
@@ -38,6 +38,7 @@ my class Promise does Awaitable {
     has $!lock;
     has $!cond;
     has $!thens;
+    has $!thens-sync;
     has Mu $!dynamic_context;
     has Bool $!report-broken-if-sunk;
 
@@ -48,6 +49,7 @@ my class Promise does Awaitable {
         $!cond                  := $!lock.condition;
         $!status                := Planned;
         $!thens                 := nqp::null;
+        $!thens-sync            := nqp::null;
         self
     }
     submethod new(:$scheduler = $*SCHEDULER, :$report-broken-if-sunk) {
@@ -70,6 +72,7 @@ my class Promise does Awaitable {
         $!cond                  := $!lock.condition();
         $!status                := Planned;
         $!thens                 := nqp::null;
+        $!thens-sync            := nqp::null;
     }
 
     # A Vow is used to enable the right to keep/break a promise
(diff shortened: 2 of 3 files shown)


patrickbkr:
There are three test failures in S17-supplies/flat.t. I do not observe those locally. I am on vacation the next week, so it's unlikely I'll find the time to look into those failures until then.

patrickbkr:
I have implemented a fix for the failures in S17-supplies/flat.t. They turned out to be caused by raciness introduced by `Lock::Async::protect-or-queue-on-recursion`. I have extended `Promise.then` with a flag `:block-when-run` that blocks the `keep` or `break` until all thens have finished. That is action-at-a-distance, so I think its a bit unclean. I'm open for hints on how else to solve this.

patrickbkr:
I just noticed that my latest change now deadlocks again on https://github.com/rakudo/rakudo/issues/5141. Thus I changed it to draft again.

patrickbkr:
The problem with the previous approach of awaiting all queued code blocks was that since the code blocks are executed in
separate continuations the `protect-or-queue-on-recursion` won't trigger in those queued code instances anymore. 
Blocking on them then introduces deadlock potential again. I now changed `protect-or-queue-on-recursion` to not put the
queued code objects on the thread pool anymore, but process them synchronously. That makes `protect-or-queue-on-recursion` trigger reliably.

With https://github.com/rakudo/rakudo/pull/5211 merged this now finally passes `make test`, `make spectest` and the original deadlock scenario.

@lizmat: A part of ffb312f57c7d0b89dfcfbe200a3d59a42a7365e3 is implemented in plain Raku where I'd like to have
optimized NQP. My failed NQP attempts are commented out. Can you have a look? (in Promise.pm6 lines 177-189 and 259)

patrickbkr:
With the help of lizmat I managed to change the hot path bits in Promise back to nqp ops. I consider this PR ready now.

patrickbkr:
I'd like to merge this PR shortly after a release (in contrast to shorty before). The next release is tentatively scheduled for 2023-04-21.

patrickbkr:
Ping @jnthn

patrickbkr:
I've just rebased this onto main.

patrickbkr:
So, I've found and fixed another deadlock. I don't like the solution though.
In src/core.c/Promise.rakumod I now explicitly copy `$*LOCK-ASYNC-RECURSION-LIST` into the `then` threads. This couples `Promise` to `Lock::Async`. That's not good.
Al alternative would be to change the `block-when-run` switch added to `Promise.then` (and used exclusively by `Lock::Async.protect-or-queue-on-recursion`) to not start the code objects in separate threads at all. They'd be serial then and wouldn't lose the call stack, thus no copying of `$*LOCK-ASYNC-RECURSION-LIST` would be necessary.
I'm open for other suggestions.

Apart from this, this PR passes all spec tests including the ones in https://github.com/Raku/roast/pull/833

@jnthn Can you have a look?

PR #5207, opened 2023-02-15
RakuAST: Add RakuAST::Type::Enum

ab5tract:


src/Raku/ast/type.rakumod
```diff
@@ -311,6 +311,226 @@ class RakuAST::Type::Parameterized
...
+        } elsif nqp::istype($evaluated, Str) {
+            # TODO: What do we actually want to do when base-type is defined but they only provide a single Str?
+            #       Base just ignores and uses Int
+            # A single string enum will always have 0, but we use $cur-val to keep it boxed
+            %values{$evaluated} := $cur-val.succ;
+            $base-type := Int;
+        } elsif nqp::istype($evaluated, List) || $is-settings-list {
+            my @items := $is-settings-list
+                            ?? $evaluated.FLATTENABLE_LIST
+                            !! self.IMPL-UNWRAP-LIST($evaluated);
```
niner, line 427: This distinction seems unnecessary as IMPL-UNWRAP-LIST is precicely for getting an NQP list out of a List object.

src/Raku/ast/type.rakumod
```diff
@@ -311,6 +311,226 @@ class RakuAST::Type::Parameterized
...
+  is RakuAST::PackageInstaller
+  is RakuAST::ImplicitLookups
+{
+    has RakuAST::Name           $.name;
+    has RakuAST::Expression     $.term;
+    has RakuAST::Type           $.of;
+    has Mu                      $!current-package;
+    # Note: Not using RakuAST::Type::Derived because we don't always know the base-type ahead of time
+    has Mu                      $!base-type;
+    has Mu                      $!meta;
```
niner, line 329: This attribute is now unused.

niner (changes requested):

src/Raku/ast/type.rakumod
ab5tract: Not sure what's going wrong then, but without this there was no iterating over `$evaluated` when a settings-level List object is returned from BEGIN-TIME-EVALUATE.

ab5tract: You are correct, it works fine without it. Thanks for the clarification.

src/Raku/ast/type.rakumod
```diff
@@ -311,6 +311,240 @@ class RakuAST::Type::Parameterized
...
+
+    method new(RakuAST::Name :$name, List :$term!, RakuAST::Type :$of, str :$scope) {
```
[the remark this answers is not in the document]
ab5tract: Yes, that's a good point. But we should probably standardize expectations around "anonymous-capable" RakuAST nodes, ie should they always require a RakuAST::Name node even if it is just representing `''` underneath? Or should this code (and maybe elsewhere) be smarter and be able to handle an undefined `$name`?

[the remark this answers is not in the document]
lizmat: The latter, as it would simplify creating `RakuAST::Type::Enum` from Raku source

src/Raku/ast/type.rakumod
ab5tract: Fixed.

src/Raku/ast/type.rakumod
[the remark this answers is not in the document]
ab5tract: Changes made, I believe they are acceptable.

niner (approved):

niner:
Rebased and merged manually. Thanks for your patience!

Issue #5170, opened 2023-01-26
Feature: warn on assigning a Block to a Hash


2colours:
It's a typical newbie mistake to use the superfluous brackets when assigning to a `Hash`, and it can easily turn into a `Block` (use of semicolon, `$_` etc.). It is almost certainly a mistake. It would be good to warn the user about it.

It can either be handled in the core as a dispatch for STORE or caught on the RakuAST level.

lizmat:
Could you give an example of code that should warn?

2colours:
Inspired by [this discussion](https://irclogs.raku.org/raku-beginner/2023-01-25.html#14:15) - this is the output on Rakudo v2022.12:

```
Useless use of "=>" in expression "' ' => '<space>'" in sink context (line 4)
Odd number of elements found where hash initializer expected:
Only saw: -> ;; $_? is raw = OUTER::<$_> { #`(Block|2227340887624) ... }
  in sub dispatch at /tmp/demo.raku line 3
  in block <unit> at /tmp/demo.raku line 10
  ```

So one can kind of see something went wrong but you need to know what actually happened in order to understand it.

**Comment by lizmat** [comment]
Ok, so the "odd number of elements" needs to check further if there's only 1 argument and its a `Hash`?

**Comment by vrurg** [comment]
More like it's a block. Though it's clearly not a run-time warning as it may create even more confusion than solve the existing problem. Rather the compiler must check if LHS of an assignment is an `Associative` and RHS is a block and raise a flag.

Apparently, this is more complicated than just an assignment as we have routine parameters, for example. But the overall point remains the same: I wouldn't be happy seeing a warning coming from the depths of object construction and trying to figure out WTH is going on down there.

**Comment by 2colours** [comment]
Either way, it should be better to not dig too deep and handle it on the surface: if it's an assignment to a `%var` and the right handside is a `Block` (or frankly, any `{}` style literal), "you probably don't want that".

I'd have thought that it's possible to catch an assignment like this as a `STORE` with a `Block` argument, on `Hash` at least. That doesn't sound like "a warning coming from the depths of object construction", any more than `Mu.new` having a candidate only to warn you that you shouldn't be passing positionals to it.

**Comment by lizmat** [comment]
As I said, it could be done as a further refinement in the "Odd number" handling

**Comment by vrurg** [comment]
```raku
sub foo($v) { my %h = $v; }
foo({:a; :b}); # Warning
```

[...]

PR #5213, opened 2023-02-23
Provide better error message for my %h = Callable

lizmat:
If it is the *only* value stored in a Map, and it is a Callable, provide an error message that points to the use of ; vs , in the contents of the Callable ({ :a; :b } versus { :a, :b }).

Addresses https://github.com/rakudo/rakudo/issues/5170

This change may actually be a performance improvement, as it reduces the binary size of the STORE methods involved, making it more likely they can be inlined.


vrurg:
`;` is not the only cause of interpreting a supposed to be hash construct as a block, if I'm not mistaken. Use of `$_` and, possibly, a few other elements.

PR #5223, opened 2023-03-01
Introduce Rational.Str(:digits)

lizmat:
Allows one to specify number of digits for stringification of Rat/FatRats.  Inspired by discussion by tadzik++ at

  https://irclogs.raku.org/raku/2023-03-01.html#10:19

### DIFF
--- a/src/core.c/Rational.pm6
+++ b/src/core.c/Rational.pm6
@@ -109,7 +109,7 @@ my role Rational[::NuT = Int, ::DeT = ::("NuT")] does Real {
         $!denominator == 1 && $!numerator.is-prime
     }
 
-    multi method Str(::?CLASS:D: --> Str:D) {
+    multi method Str(::?CLASS:D: :$digits --> Str:D) {
         if $!denominator {
             my \abs   := self.abs;                              # N / D
             my \whole := abs.floor;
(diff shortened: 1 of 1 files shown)


lizmat:
```
$ raku -e 'say (1 / 998.001).Rat.Str(:300digits)'        
0.0010020030040050060070080090100110120130140150160170180190200210220230240250260270280290300310320330340350360370380390400410420430440450460470480490500510520530540550560570580590600610620630640650660670680690700710720730740750760770780790800810820830840850860870880890900910920930940950960970980991
```
Note that the final "100" is shown as "1" because of `0`'s being removed from the tail.  Perhaps a specific `:digits` specification should overrule that.
```
$ raku -e 'say (1 / 998.001).Rat.Str(:303digits)'        
0.001002003004005006007008009010011012013014015016017018019020021022023024025026027028029030031032033034035036037038039040041042043044045046047048049050051052053054055056057058059060061062063064065066067068069070071072073074075076077078079080081082083084085086087088089090091092093094095096097098099100101
```
shows that the correct digits *are* there.

lizmat:
Additionally, if merged, the 6.e Formatter class should take the precision of `%f` formatting into the `.Str` call.

lizmat:
After looking into this some more, I don't think a ~15% performance loss for introducing an optional named argument, is worth the change.  And especially if looking at adding yet another named for keeping trailing zeroes.  And especially when there's a [`Rat::Precise`](https://raku.land/zef:thundergnat/Rat::Precise) module around.

So closing this PR: a lot of this logic will find its way to the new `Formatter` class in time.

PR #5234, opened 2023-03-18
Introduce Any.do(&)

lizmat:
Similar to Kotlin's apply, and basically the same as Supply.do already does, but now for any value.  Calls the given Callable with the invocant as the topic.  Usually executed for its side-effects.

    $ raku -e 'my $a = 42; dd a.do: { .say; .Num }'
    42
    42e0

Inspired by a conversation with jnthn

### DIFF
--- a/src/core.c/Any-iterable-methods.pm6
+++ b/src/core.c/Any-iterable-methods.pm6
@@ -2552,9 +2552,6 @@ Consider using a block if any of these are necessary for your mapping code."
           List,'$!reified',buffer
         )
     }
-
-    proto method do(|) is nodal {*}
-    multi method do(\SELF: &doer) { doer(SELF) }
 }
 
 BEGIN Attribute.^compose;
--- a/src/core.c/Any-iterable-methods.pm6
+++ b/src/core.c/Any-iterable-methods.pm6
@@ -2552,6 +2552,9 @@ Consider using a block if any of these are necessary for your mapping code."
           List,'$!reified',buffer
         )
     }
+
+    proto method do(|) is nodal {*}
+    multi method do(\SELF: &doer) { doer(SELF) }
 }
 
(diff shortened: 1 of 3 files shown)


vrurg:
Transferring a question from [#raku-dev](https://irclogs.raku.org/raku-dev/2023-03-18.html#21:05-0002): what difference does it make with `$foo.&{ ... }`? If we skip the readability point.

lizmat:
The difference is that `.do` will always return the invocant.

2colours:
Is that slight difference really worth the namespace pollution?

lizmat:
1. Adding a method is *not* namespace pollution.  Add a sub *would* be.
2. That's why it's a PR.
3. Some people expressed missing such functionality, that's why I made the PR

vrurg:
I made a note on IRC and would propose it here again. If there is use for the method (looks like there is) then it is better be named `snitch`, to put it in line with similar functionality of `Seq`. Having a `do` method whose semantics would be different from `do {...}` would be confusing.

lizmat:
Ah, duh.  the current implementation of `.do`, *is* `.snitch`!  But without the special handling for `Seq`.  So it is even inferior.  So closing this PR!

2colours:
For what it's worth, adding a method *on `Any` is*, for all intents and purposes, namespace pollution. `Any` is a common base class that user-defined classes shouldn't have to forcefully ignore; and if they don't, they need to take the inherited method into account. Either the author risks an override that contradicts the base class, hence potentially causing problems at type-agnostic consumers, or they have their hands tied, by something they potentially didn't even want.

Mind you, this isn't necessarily a problem for a word like `do` that authors arguably shouldn't use anyway, for their own good. The principle is important nevertheless: public methods of `Any` are a big shared interface for most code out there, and therefore not something to change without sufficient reasoning.

PR #5235, opened 2023-03-20
Fix a wrong protocol in a URL of "git clone"

komainu8:
If we use "git://github.com/rakudo/rakudo.git" when we clone sources of Rakudo, we fail git clone as below.

```
fatal: unable to connect to github.com:
```

I think we should use https:// instead of git:// when we execute "git clone".


patrickbkr:
Good spot. Thanks!

komainu8:
Thank you for merge this PR.

PR #5242, opened 2023-04-03
Copy more grammar to RakuAST for error messages

MasterDuke17:
This gets us the right error messages/suggestions for more things (e.g., a bare `say` now suggests `.say` or `&say`).

Now t/spec/S16-io/bare-say.t and t/spec/S16-io/put.t are passing.


src/Raku/Grammar.nqp
```diff
@@ -1821,6 +1830,36 @@ grammar Raku::Grammar is HLL::Grammar does Raku::Common {
...
+                            }
+                            else {
+                                my $missing := $/.terminator() || $/.infixish();
+                                $/.'!clear_highwater'();  # don't have suppose
+                                my $orry := $missing ?? "sorry" !! "worry";
+                                if $trap == 1 {        # probably misused P5ism
+                                    $<longname>."{$orry}obs"("bare \"$name\"", ".$name if you meant to call it as a method on \$_, or use an explicit invocant or argument, or use &$name to refer to the function as a noun");
+                                }
+                                elsif $trap == 2 {        # probably misused P6ism
+                                    $<longname>."$orry"("Function \"$name\" may not be called without arguments (please use () or whitespace to denote arguments, or &$name to refer to the function as a noun, or use .$name if you meant to call it as a method on \$_)");
```
niner, line 1855: This actually looks like a bug to me even in the old compiler. We never check if the name refers to the sub in the setting:
```
nine@sphinx:~/rakudo (main *=)> rakudo -e 'my sub one { 1 }; say one;'
Function "one" may not be called without arguments (please use () or whitespace to denote arguments, or &one to refer to the function as a noun, or use .one if you meant to call it as a method on $_)
```

MasterDuke17: I've been looking for something like `.is_from_core` from src/Perl/Optimizer.nqp in src/Raku/*, but the closest thing seems to be `.resolve-name-constant-in-setting`, which I don't think does what I want. Am I missing something obvious?

MasterDuke17: Oh hm, maybe I can/should do it the other way round, and use something like `.find-lexical`...

niner: I think you need both. You need to resolve the symbol lexically and also find the equivalent symbol in the setting and check whether they are identical. However you should **not** call find-lexical directly. This method must be called only after the AST is complete, as it will create a cache for the lexicals it finds and never look at the AST again.
So I think this check actually belongs in a PERFORM-CHECK method of a node that does RakuAST::CheckTime rather than in the grammar. That said, I think calling these without arguments is actually allowed if followed by parenthesis, e.g. `say()` - a distinction which might be lost in the AST.

PR #5243, opened 2023-04-12
Add "is-monotonically-increasing" method to Iterators

lizmat:
Returns False by default.  If returning True, then .sort() can pass on the underlying iterator unchanged in a new Seq.

Added many Range iterators to set that, so that now:

    say (1..*).sort;  # (1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14...

will work, rather than throwing because the underlying iterator is lazy.

### DIFF
--- a/src/core.c/Iterator.pm6
+++ b/src/core.c/Iterator.pm6
@@ -55,13 +55,13 @@ my role Iterator {
         )
     }
 
-    # Pushes things until we hit a lazy iterator (one whose is-lazy method returns
-    # True). The default works well for non-composite iterators (that is, those
-    # that don't trigger the evaluation of other iterators): it looks at the
-    # lazy property of itself, and if it's true, does nothing, otherwise it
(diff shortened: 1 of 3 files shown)


grondilu:
"is-monotonically-increasing" is hell of a long name.

Maybe "is-monotonic" or "is-monotone" ([terminology from [the wp article](https://en.wikipedia.org/wiki/Monotonic_function)), would be more suitable.  Whether it is increasing or decreasing should be easily tested with the first values (unless the sequence is constant for many first terms or something).

lizmat:
It is intentionally a long name: you would only need to be confronted by it when you're making an Iterator class yourself.  It will never be something very common, so a long name that **clearly** indicates what it means, trumps brevity, I'd say.

lizmat:
Also, re "Whether it is increasing or decreasing": the whole point of the method is that the iterator is **declared** to be monotonically increasing **without** having to produce any values.

PR #5245, opened 2023-04-16
Introduce `Int.html-entity`

lizmat:
Attempts to return the HTML entity associated with the integer value (assumed to be a codepoint).  Returns Nil if there is no such entity.

Lazily builds the lookup hash from the Str.htmlparse lookup table the first time the Int.html-entity method is invoked, so this won't increase build/startup time significantly, while being able to provide a very nice feature when it is needed.

### DIFF
--- a/src/core.c/htmlparse.pm6
+++ b/src/core.c/htmlparse.pm6
@@ -2156,11 +2156,35 @@ augment class Str {
         }
     }
 
+    method entity2ord(Str:) is implementation-detail { $entity2ord }
 #?endif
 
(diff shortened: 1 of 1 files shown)


niner:
I think Int is the wrong place for this. It has nothing to do with integers. I would never look for it in Int and I would be quite surprised to find it in Int's docs.

lizmat:
Yet there is a precedent, where `Int` is interpreted as a codepoint:
````
$ raku -e 'dd 65.chr'
"A"
````

PR #5262, opened 2023-05-11
Fix handling of =input / =output pod blocks

lizmat:
Before this, =input and =output pod was indistinguishable from =code pod, because they all codegenned to Pod::Block::Code objects.

This introduces two now classes: Pod::Block::Input and Pod::Block::Output. These classes are identical to the Pod::Block::Code classes, apart from their name.  And added their handling to Pod::To::Text, identical to the Pod::Block::Code handling to remain compatible.

This change should allow legacy pod renderers to render =code, =input and =output differently.

Some tests needed adaptation because they expected the Pod::Block::Code object, rather than the new Pod::Block::Input and Pod::Block::Output objects.

### DIFF
--- a/lib/Pod/To/Text.rakumod
+++ b/lib/Pod/To/Text.rakumod
@@ -14,6 +14,8 @@ sub pod2text($pod) is export {
     given $pod {
         when Pod::Heading           { heading2text($pod)             }
         when Pod::Block::Code       { code2text($pod)                }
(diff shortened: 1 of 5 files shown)


lizmat:
Closing because https://github.com/rakudo/rakudo/commit/6f482ca006 turned out to be a simpler, backward compatible solution.

PR #5269, opened 2023-06-04
Add connect-path and listen-path to IO::Socket::Async

### LINKED ISSUE #5024: Support asynchronous unix domain sockets
## The Problem

Currently we support asynchronous IP sockets (`IO::Socket::Async`), but not asynchronous unix domain sockets.

We should add support for this.

This is a Rakudo problem that needs a [MoarVM](https://github.com/MoarVM/MoarVM/issues/1734) solution.

Leont:
This fixes #5024

This should be merged after/with https://github.com/MoarVM/MoarVM/pull/1756 and should bump the MoarVM dependency

### DIFF
--- a/src/core.c/IO/Socket/Async.pm6
+++ b/src/core.c/IO/Socket/Async.pm6
@@ -199,6 +199,36 @@ my class IO::Socket::Async {
         $p
     }
 
+    method connect-path(IO::Socket::Async:U: Str() $path,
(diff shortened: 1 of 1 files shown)


Leont:
I'm not sure if this should take a `Str` or an `IO::Path`, I can see the argument for both.

amano-kenji:
I don't know how to use `listen-path` and `connec-path`. Does `listen-path` create a stream socket or a datagram socket?

PR #5270, opened 2023-06-04
Make IO::Handle.Supply read character-by-character

### LINKED ISSUE #5189: IO::Handle.Supply.lines blocks on buffer.
## The Problem
When turning an `IO::Handle` into a stream using `.Supply`, the resulting supply isn't usable when chain with for example `.lines` or `.words`.

## Expected Behavior
Lines/words are read as they come in.

## Actual Behavior
No lines are processed until the entire buffer (64k by default) is filled (or EOF is reached). This can be worked around, by setting the blocksize to 1 byte, but that is highly unintuitive and performs poorly.

What `.Supply` probably should do instead is read _up to_ `$blocksize` bytes.

## Steps to Reproduce
`react { whenever $*IN.Supply.lines { say $_ } }`

## Environment
```
Welcome to Rakudo™ v2022.12.
Built on MoarVM version 2022.12.
```

Leont:
Otherwise it will only emit any data once 64k bytes are read, which means a construct like $handle.Supply.lines will not DWIM at all.

This fixes #5189. It' not terribly efficient but at least it's correct.

### DIFF
--- a/src/core.c/IO/Handle.pm6
+++ b/src/core.c/IO/Handle.pm6
@@ -560,7 +560,7 @@ my class IO::Handle {
         $result
     }
 
-    multi method Supply(IO::Handle:D: :$size = $*DEFAULT-READ-ELEMS --> Supply:D) {
+    multi method Supply(IO::Handle:D: :$size = 1 --> Supply:D) {
         if $!decoder { # handle is in character mode
             supply {
                 my int $chars = $size;


lizmat:
Feels like **1** is the other end of the spectrum, making `.Supply` potentially useless performance wise?

Leont:
> Feels like **1** is the other end of the spectrum, making `.Supply` potentially useless performance wise?

Any other value would be wrong though.

The better solution would be to add a new read function to nqp/moarvm that has `read(2)` semantics instead of `fread(3)` semantics. That work would be of a different order of magnitude though.

PR #5273, opened 2023-06-05
Make `$(;)` return an empty list as it should be

patrickbkr:
StatementSequence does return an empty list when there are no children, but it failed to treat Statement::Empty as a non-element and thus created a list with one Nil element (that's what Statement::Empty generates).

Fixes one test in `t/spec/S02-types/list.t` but gains no new files:

    is $(;).elems, 0, '$(;) parses, and is empty';

This is roughly equivalent to the change that fixed the same problem on the legacy backend:

https://github.com/rakudo/rakudo/commit/726d2fa

### DIFF
--- a/src/Raku/ast/statements.rakumod
+++ b/src/Raku/ast/statements.rakumod
@@ -416,13 +416,14 @@ class RakuAST::StatementSequence
     method IMPL-TO-QAST(RakuAST::IMPL::QASTContext $context) {
         my @statements := self.code-statements;
         my int $n := nqp::elems(@statements);
-        if $n == 1 {
-            nqp::atpos(@statements, 0).IMPL-TO-QAST($context)
-        }
-        elsif $n == 0 {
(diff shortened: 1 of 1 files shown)


lizmat (approved):

niner:
The patch fixes that spectest, but doesn't fully fix the underlying problem:
```
nine@sunshine:~/rakudo (main *=)> make && RAKUDO_RAKUAST=1 ./rakudo-m -e 'say $(;;;;).elems'
make: Nothing to be done for 'all'.
1
nine@sunshine:~/rakudo (main *=)> ./rakudo-m -e 'say $(;;;;).elems'
0
```
I think it needs generalizing to filter out all Empty statements before checking the number of remaining ones.

lizmat:
Thanks for the PR and the nudge.  Fixed alternately with https://github.com/rakudo/rakudo/commit/a6bec923c5 .

PR #5275, opened 2023-06-05
Make "string" but False eqv "string" but False

Leont:
This doesn't actually fix the general case, only the `but Bool` case. In practice they're by far the most common case.

In particular the following cases are now all true:

say "foo" but False eqv "foo" but False;
say "foo" but False === "foo" but False;
say "foo" but False !eqv "foo" but True;
say "foo" but False !=== "foo" but True;

The only other semi-common case would be `but Str`, which could be implemented in a similar way.


lizmat:
This feels like a hack rather than a generic solution.  @jnthn?

PR #5278, opened 2023-06-10
Deprecate including an ambiguous .pm module file

### LINKED ISSUE #1846: Deprecate the usage of .pm extension for modules
## The Problem
 
One cannot create a perl + perl6 project containing something like:
```
lib/MyPerl6Code.pm6
lib/MyPythonCode.py
lib/MyPerlCode.pm
```

because MyPerlCode.pm will always be treated as perl6 code during module lookup.

## Expected Behavior

```
$ perl6 -Ilib -e 'use MyPythonCode;'
===SORRY!===
Could not find MyPythonCode at line 1 in:
...

$ perl6 -Ilib -e 'use MyPerlCode;'
===SORRY!===
Could not find MyPerlCode at line 1 in:
...
```

## Actual Behavior

```
$ perl6 -Ilib -e 'use MyPythonCode;'
===SORRY!===
Could not find MyPythonCode at line 1 in:
...

$ perl6 -Ilib -e 'use MyPerlCode;'
===SORRY!===
This appears to be Perl 5 code.
...
```

## Why expected behavior is ideal

* Having two extensions make no sense other than for backwards compatibility. This is easy for anyone to submit PRs for, or to automate PRs for.

* Writing code to have to check for multiple extensions sucks, especially because it means it cannot simply be treated as `this-or-that `but `this-or-that-or-this_and_that`. It also means we technically need to document and justify choosing one of `this` or `that` when the result is `this_and_that`.

* `CompUnit::Repository::FileSystem` ( `CURFS` ) does not try to execute python, ruby, etc module files so its possible to have:
    ```
    lib/Implementation/Selector.pm6
    lib/Implementation.pm6
    lib/Implementation.py
    ```
    but not
    ```
    lib/Implementation/Selector.pm6
    lib/Implementation.pm6
    lib/Implementation.py
    lib/Implementation.pm
    ``` 
    meaning `lib/Implementation/Selector.pm6` could do `use Implementation:from<Python>` but not `use Implementation:from<Perl5>;` when invoking perl6 with `-Ilib` or using `use lib 'lib'` in a bin/script.

* Shouldn't break existing installations.

## Why expected behavior is not ideal

* Would break non-installed code relying on `use lib "lib"` finding some `.pm` file ( assuming a non-ecosystem module since PRs can fix those ) after updating rakudo. The deprecation period should probably be based on the seriousness of this point.

* It doesn't map to our packaging format / `provides` -- The expected behavior might lead one to think they can package $other-language module by adding something like `"MyPythonCode" : "lib/MyPythonCode.py"` to `provides` meta data, but that extension would be ignored during installation so any other perl6 modules in that distribution relying on loading that python module would not work the same after installation. ( Note that in the future this might be a perfectly fine way to handle things, but only if perl `.pm` files are treated the same as all other non `.pm6` files. )

* Updating `perl6/roast` will probably suck and will make git blame a bit worse.

ugexe:
We need to know when raku source files have changed for the purposes of precompilation. However, when a user includes a library path (i.e. -Ilib, -I/home/foo) that does not contain a META6.json file we might end up erroneously considering perl module files (which also use .pm) as raku files. While this is mostly harmless, it does waste some cpu and io. This adds a deprecation message when this ambiguous situation occurs so that in the future we might be able to remove support for the .pm extension altogether.

Resolves #1846

### DIFF
--- a/src/core.c/CompUnit/Repository/FileSystem.pm6
+++ b/src/core.c/CompUnit/Repository/FileSystem.pm6
@@ -326,6 +326,18 @@ class CompUnit::Repository::FileSystem
         # make sure .rakumod has precedence over .pm6 and .pm
         $_ = .<rakumod> // .<pm6> // .<pm> for %provides.values;
 
+        # The .pm file extension can also match perl modules. We should encourage users
+        # to stop using it for their raku code so we don't waste cycles sha1ing their
+        # source code or attempt to precompile them.
+        if %provides.values.grep(*.ends-with('.pm')) {
+            DEPRECATED(
+                "the .rakumod extension for raku modules, or include a META6.json file that explicitly declares each raku module file",
+                :what(".pm file extension in raku library path"),
+                :file(self.path-spec),
+                :line(0),
+            );
+        }
+
         Distribution::Hash.new(:$prefix, %(
           name      => ~$!prefix,  # must make up a name when using -Ilib
           ver       => '*',


ugexe:
```
$ tree
.
└── lib
    └── Foo.pm

2 directories, 1 file
```
```
$ raku -Ilib -e 'use Foo;'
Saw 1 occurrence of deprecated code.
================================================================================
.pm file extension in raku library path seen at:
  file#/Users/nlogan/repos/Foo/lib, line 0
Please use the .rakumod extension for raku modules, or include a META6.json file that explicitly declares each raku module file.
--------------------------------------------------------------------------------
Please contact the author to have these occurrences of deprecated code
adapted, so that this message will disappear!
```

ugexe:
Note that I'm not sure if we actually want to do this or not, as it would be annoying for users that both A) use a non-META6.json library path and B) have perl modules alongside their raku files from the aforementioned library path.

vrurg:
I'd rather be pro-deprecation. As to the annoyance, perhaps make the message suppressible? A pragma, maybe?

ugexe:
If we really want a way to opt into the old behavior, we could alternatively require they pass in the option with the path spec. Something like `-I file#allow-pm-ext<1>#/path/to/lib` when they would originally have used `-I /path/to/lib` or `-I file#/path/to/lib`

lizmat:
+1 for deprecation message

vrurg:
> Something like `-I file#allow-pm-ext<1>#/path/to/lib`

Looks about right. We'd need to remember to document it though. :)

ugexe:
So we have a few potential ways to allows users to disable the deprecation message. Now I wonder if we should implement it now, or wait until later once/if anyone complains. I kind of would prefer to implement it later just so we don't have to bother supporting some additional public API that may not actually be needed.

PR #5280, opened 2023-06-12
Include resource files when calculating CURFS repository id

ugexe:
We sha1 the source code of module files for CURFS so we can tell when source code has changed. Since source code can rely on files in resources/ we should also sha1 those files as well.


niner:
I'm against doing this. Resources can potentially be quite large. And there are a lot of use cases where the precompilation result does not depend on them at all. Consider e.g. a web application with templates, JS, images, videos, etc.

A better way is to support explicit registration of external dependencies as described in:
https://github.com/Raku/problem-solving/issues/159#issuecomment-583522999

ugexe:
It was too early: I see from the comment the expectation is that the user would register these in their code. Fair enough.

I have not ruled out the new directory though. Something like that seems like it could also work towards a solution to the "I don't want my resources renamed" issue.

vrurg:
I think it's not the content of the resources what matters. This is where an application itself can do something. But the content of `$?DISTRIBUTION` because it is precompiled. Therefore, checksuming over normalized META6 content would be very much useful.

PR #5282, opened 2023-06-13
RakuAST: add Dutch in foreign language deparsing

lizmat:
I *think* we should have officially supported foreign language translations in core, to ensure that we don't get into a Babylonian knot of language misunderstandings.

OTOH, this could also well live in the ecosystem, to allow updates regardless of Rakudo / Raku language versions.


JJ:
Just a suggestion here. To make implementation easier, could the name of the file/role be one of the standard codes for language? As in NL in this case. That would make easier to map locale settings to these deparsers.

lizmat:
Aahh... good point!   Which standard should we use for that?  I guess ISO-639?

JJ:
> Aahh... good point! Which standard should we use for that? I guess ISO-639?

That seems to be the one used for locale settings, so I guess so.

vrurg:
To be honest, I wouldn't get the module's purpose from its name as NL may stand for too many things. Shouldn't it be `RakuAST::Deparse::L10N::NL`?

lizmat:
@vrurg I think it being a subclass of `::Deparse` already gives that away.

ugexe:
> @vrurg I think it being a subclass of ::Deparse already gives that away.

I don't think so. When I read something like `::Deparse::IO` I don't immediately assume its referencing [some language](https://en.wikipedia.org/wiki/Ido)

PR #5285, opened 2023-06-14
Precomp Pipeline: Fix building on CentOS 7

patrickbkr:
On CentOS 7 latest GCC you can get is 4.8.5. But latest Rakudo 2023.05 contains a libuv that depends on c11 atomics which are first available in GCC 4.9.


patrickbkr:
The failures are unrelated. Merging.

PR #5286, opened 2023-06-17
Remove largely unknown flags from installed raku script shims

ugexe:
A mostly unknown feature of the bin script shims CURI installs is that it allows choosing the distribution to load the script from, which allows loading an e.g. older version of some script. However to do so required using the :$name :$auth :$ver :$api arguments which prevents any installed raku bin script from taking any arguments of those names; the shim scripts breaks things such as providing a bin/foo --ver that outputs version information. This change removes those arguments until a better solution can be implemented.


ugexe:
This is an alternative to #4153

niner (approved):

PR #5287, opened 2023-06-18
Optimize internal directory recursion

ugexe:
This aims to optimize `Rakudo::Internals.DIR-RECURSE` (and thus e.g. `-Ilib` start up time) by reducing the number of `nqp::stat` calls made when doing a recursive directory search for files.

### DIFF
--- a/src/core.c/Rakudo/Internals.pm6
+++ b/src/core.c/Rakudo/Internals.pm6
@@ -1386,12 +1386,12 @@ my class Rakudo::Internals {
               nqp::stmts(
                 (my str $path = nqp::concat($!abspath,$entry)),
                 nqp::if(
-                  (try nqp::stat($path,nqp::const::STAT_ISREG))
-                    && $!file.ACCEPTS($entry),
+                  $!file.ACCEPTS($entry) &&
+                    (try nqp::stat($path,nqp::const::STAT_ISREG)),
                   (return $path),
                   nqp::if(
-                    (try nqp::stat($path,nqp::const::STAT_ISDIR))
-                      && $!dir.ACCEPTS($entry),
+                    $!dir.ACCEPTS($entry) &&
+                      (try nqp::stat($path,nqp::const::STAT_ISDIR)),
                     nqp::stmts(
                       nqp::if(
                         (try nqp::fileislink($path)),
--- a/src/core.c/Rakudo/Internals.pm6
+++ b/src/core.c/Rakudo/Internals.pm6
@@ -1383,21 +1383,18 @@ my class Rakudo::Internals {
         method pull-one() {
             nqp::while(
               nqp::chars(my str $entry = self!next),
-              nqp::if(
-                nqp::stat(
-                  (my str $path = nqp::concat($!abspath,$entry)),
-                  nqp::const::STAT_EXISTS
-                ),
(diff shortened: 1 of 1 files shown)


ugexe:
If we had an `nqp::stat` that returned the full stat data struct then we could further reduce the number of stats by replacing the STAT_ISREG and STAT_ISDIR stat calls with a single stat.

src/core.c/Rakudo/Internals.pm6
```diff
@@ -1383,18 +1383,15 @@ my class Rakudo::Internals {
...
-                  (my str $path = nqp::concat($!abspath,$entry)),
-                  nqp::const::STAT_EXISTS
-                ),
+              nqp::stmts(
+                (my str $path = nqp::concat($!abspath,$entry)),
                 nqp::if(
-                  nqp::stat($path,nqp::const::STAT_ISREG)
-                    && $!file.ACCEPTS($entry),
+                  $!file.ACCEPTS($entry) &&
+                    (try nqp::stat($path,nqp::const::STAT_ISREG)),
```
lizmat, line 1390: Won't this `try` slow things down significantly?

ugexe: not as much as the additional `stat` calls apparently - see the commit message that mentions a 10% speed up

ugexe: The other thing is that it is technically more correct with the `try` because there is a race condition between the old stat on existence and the stat to check if its a file or directory.

lizmat: True.

Wouldn't it make more sense to have a single `try` wrapping the whole thing?

ugexe: No. The desired behavior is not to abort entirely once an exception occurs. If you look at how it is implemented in this PR we would expect the `nqp::stat($path,nqp::const::STAT_ISREG)` to throw an exception yet we still want to continue to the next stat call to test if its a directory.

src/core.c/Rakudo/Internals.pm6
```diff
@@ -1383,18 +1383,15 @@ my class Rakudo::Internals {
...
                   (return $path),
                   nqp::if(
-                    nqp::stat($path,nqp::const::STAT_ISDIR)
-                      && $!dir.ACCEPTS($entry),
+                    $!dir.ACCEPTS($entry) &&
+                      (try nqp::stat($path,nqp::const::STAT_ISDIR)),
                     nqp::stmts(
                       nqp::if(
                         nqp::fileislink($path),
```
ugexe, line 1397: This would need a `try` as well

src/core.c/Rakudo/Internals.pm6
ugexe: Actually maybe it makes sense. I was confused in my previous statement because an exception would not occur when the path exists but it is a directory.

ugexe:
I've been benchmarking with `say Rakudo::Internals.DIR-RECURSE($*CWD.absolute).elems` inside my home directory. Without this PR it takes ~3.7s, and with this PR it takes ~3.17s.

MasterDuke17 (approved):

PR #5288, opened 2023-06-19
Don't lookup $*(SPEC|CWD) for every file in dir...

MasterDuke17:
when recursing through a directory given as a `-I` option. When applied on top of https://github.com/rakudo/rakudo/pull/5287, `time raku -I . -e 'use Test'` in a directory with ~113k subdirectories/files dropped from ~3.7s to ~3.5s.

### DIFF
--- a/src/core.c/CompUnit/Repository/FileSystem.pm6
+++ b/src/core.c/CompUnit/Repository/FileSystem.pm6
@@ -289,7 +289,9 @@ class CompUnit::Repository::FileSystem
     # but probably a poorly formed one.
     method !dist-from-ls {
         my $prefix := self!files-prefix;
-        my &ls := { Rakudo::Internals.DIR-RECURSE($_).map(*.IO) }
+        my $SPEC := $*SPEC;
+        my $CWD := $*CWD;
+        my &ls := { Rakudo::Internals.DIR-RECURSE($_).map({ IO::Path.new($_, :$SPEC, :$CWD) }) }
         my &to-relative := { $_.relative($prefix).subst(:g, '\\', '/') }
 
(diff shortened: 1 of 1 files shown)


ugexe (approved):

src/core.c/CompUnit/Repository/FileSystem.pm6
```diff
@@ -289,7 +289,9 @@ class CompUnit::Repository::FileSystem
     # but probably a poorly formed one.
     method !dist-from-ls {
         my $prefix := self!files-prefix;
-        my &ls := { Rakudo::Internals.DIR-RECURSE($_).map(*.IO) }
+        my $SPEC := $*SPEC;
+        my $CWD := $*CWD;
+        my &ls := { Rakudo::Internals.DIR-RECURSE($_).map({ IO::Path.new($_, :$SPEC, :$CWD) }) }
         my &to-relative := { $_.relative($prefix).subst(:g, '\\', '/') }
```
ugexe, line 295: It occurs to me that we could probably pass in `:$CWD` to `relative` here. 

Also, we could also probably apply a similar `IO::Path.new(...)` replacement to the `'.' ~ .IO.extension` that occurs a little later on (although maybe we should just parse the extension off the string without converting it to an `IO::Path`).

MasterDuke17: Yeah, I thought there would be more places to do this, but hadn’t yet looked exactly where.

ugexe: Ah I was wrong - the positional argument `$prefix` is already setting `$CWD`

PR #5290, opened 2023-06-19
Use nqp::handle instead of try

ugexe:
Previously we added 'try' to ignore errors from the various nqp stat calls. The rest of the code is written in nqp, so this changes that 'try' to use nqp::handle instead for a slight performance benefit.


ugexe:
The performance benefit here is extremely minor - if the rest of the function wasn't already written in nqp I wouldn't bother with this change.

ugexe:
Closing as this won't be needed after https://github.com/rakudo/rakudo/pull/5316

PR #5291, opened 2023-06-20
Make Capture an object type

lizmat:
Rather than a value type, which it is **not** as it can have mutable elements.  See https://github.com/Raku/problem-solving/issues/372 for more information.

This causes:
- t/spec/S14-roles/attributes-6e.t
- t/spec/S14-roles/composition.t

to fail.  Both seem to be related to WALK incorrectly assuming equivalence at some level.

### DIFF
--- a/src/core.c/Capture.pm6
+++ b/src/core.c/Capture.pm6
@@ -15,33 +15,6 @@ my class Capture { # declared in BOOTSTRAP
         nqp::bindattr(self,Capture,'%!hash', $source-hash) if nqp::ishash($source-hash);
     }
 
-    multi method WHICH (Capture:D: --> ValueObjAt:D) {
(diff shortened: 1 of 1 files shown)


lizmat:
@vrurg the failures appear `.WALK` related, which feels up your alley?

vrurg:
No idea what's going on. There is even no any diagnostics to look at.

The only thing I see for now is that neither the `WALK` method, nor the `WalkList` class do any operations on captures. `WALK` itself does nothing at all, `WalkList` actually captures arguments to transfer them to the actual methods it invokes. No equality checks whatsoever. 

So, I think it's something more fundamental.

raiph:
> Both seem to be related to `WALK` incorrectly assuming equivalence at some level.

cf @jnthn:

> it'd make sense for snapshot equality (`eqv`) to work.

Is there a connection between these two comments?

lizmat:
No, not that I know of.

PR #5292, opened 2023-06-22
Don't need to use _I version of nqp::radix...

MasterDuke17:
just to check that the given string is composed of hex characters.


lizmat:
Thanks!

PR #5295, opened 2023-06-25
Add support for RAKUDO_NO_PRECOMPILATION environment variable

ugexe:
This adds a mechanism that allows users to disable precompilation via an environment variable. It prevents rakudo from using precompilation files as well as from generating them.

### DIFF
--- a/src/core.c/CompUnit/PrecompilationRepository.pm6
+++ b/src/core.c/CompUnit/PrecompilationRepository.pm6
@@ -9,8 +9,13 @@ role CompUnit::PrecompilationRepository {
 
     method load(CompUnit::PrecompilationId $id --> Nil) { }
 
-    method may-precomp(--> True) {
-        # would be a good place to check an environment variable
+    has Bool $!may-precomp;
+    method may-precomp(--> Bool:D) {
+      nqp::if(
+        nqp::defined($!may-precomp),
+        $!may-precomp,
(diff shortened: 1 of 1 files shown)


lizmat:
I think the name is ambiguous: it could be interpreted as not *creating* any precomp files, whereas the semantics are of not *using* any precomp files.

Suggest to use `RAKUDO_NO_PRECOMPILED` instead.

ugexe:
I chose the current name because it’s basically the same as the pragma ‘no precompilation’

lizmat:
Hmmm.... if your code has `no precompilation` in it, it will never be able to load any valid precomped files, as the change of adding `no precompilation` to the source, will have invalidated any precomped files.

So I've always seen `no precompilation` as not *creating* any precomped files.

But I see your point.  Still not convinced, though  :-)

lizmat (approved):

ugexe:
> So I've always seen no precompilation as not creating any precomped files.

I see, I had assumed that upon installation there would still be a precomp file getting created despite `no precompilation;` but that isn't the case

ugexe:
I wonder if this should prevent precompilation on install. As in  `$precomp.precompile(...) if $precomp.may-precomp()` [here](https://github.com/rakudo/rakudo/blob/e66bcd670c32de374986862108b19350b7508d40/src/core.c/CompUnit/Repository/Installation.pm6#L357)

ugexe:
https://github.com/rakudo/rakudo/pull/5296

ugexe:
I'm going to change the commit message since this would now (due to https://github.com/rakudo/rakudo/pull/5296) prevent precompilation on installation

PR #5296, opened 2023-06-26
Respect CUPR.may-precomp() in CURI

ugexe:
CompUnit::PrecompilationRepository.may-precomp() can be used to check if something should try to precompile using said repo. It is used in CUPR in its try-load method to determine if it should try to precompile a file when loading, which often happens with a CURFS type repository. However, CURI.install also tries to precompile files, but it does not check may-precomp() before doing so. This means there is no central place to enable or disable precompiling.

This updates CURI to only precompile modules during install if may-precomp() is True.

### DIFF
--- a/src/core.c/CompUnit/Repository/Installation.pm6
+++ b/src/core.c/CompUnit/Repository/Installation.pm6
@@ -326,13 +326,14 @@ sub MAIN(:$name, :$auth, :$ver, *@, *%) {
         # identity changes with every installation of a dist.
         $!id = Any;
 
-        if $precompile {
+        my $precomp := self.precomp-repository;
+
+        if $precompile && $precomp.may-precomp() {
             my $head := $*REPO;
             CATCH { PROCESS::<$REPO> := $head }
             # Precomp files should only depend on downstream repos
             PROCESS::<$REPO> := self;
 
-            my $precomp = self.precomp-repository;
             my $repo-prefix = self!repo-prefix;
             my $*DISTRIBUTION = CompUnit::Repository::Distribution.new($dist, :repo(self), :$dist-id);
             my $*RESOURCES = Distribution::Resources.new(:repo(self), :$dist-id);


niner:
This raises the question of why one would even want to disable precompilation during installation. What's the use-case?

lizmat:
One reason could be that your code will always run with some $*REPO setting that would always invalidate the system installed precomps, so why bother?

niner:
How would it do that? And why?

lizmat:
How?  `RAKULIB=foo`
Why?  to only run dev / production versions

niner:
But how would a mere `RAKULIB=foo` invalidate all existing precomps?
I also don't understand the use-case yet. Say if you have production versions installed and want to only run dev versions, you switch your repo to that containing the dev versions. Why would that require the installed production versions to not be precompiled? If you want to use the installed modules, you'll want the precompiled versions (because why wait for compilation when it doesn't give you any advantage?) and if you don't want to use them having precompiled versions stored doesn't hurt.

ugexe:
For the general user I don't think there is anything useful here. However, it does give power users a new levers to pull (or at least a new way to pull it). For instance: someone might develop a tool that is smarter about precompilation in their specific use case such that it can do it faster than what core raku can. In such a scenario the power user wishes to handle precompilation themselves, and thus the automatic precompilation upon install is a waste of their time. Another scenario that is more plausible is using it to debug... users often seem confused in understanding when a precompilation error has occurred during module installation. When those errors initially come up I'd probably ask them to try their install again with e.g. `RAKUDO_NO_PRECOMPILATION` (instead of trying to explain to them how I can deduce it is a precomp error), which is mostly self explanatory.

ugexe:
I just had a thought - disabling precompilation will at allow people to work around the windows precompilation issue related to path lengths (until that much more difficult issue can be fixed properly). I'm more inclined to agree with merging this unless there are good reasons not to.

PR #5299, opened 2023-06-28
Fix potential concurrent hash access race condition

ugexe:
This fixes a potential race condition in `CompUnit::PrecompilationRepository` where two threads could race each other to read and write to a hash.

### DIFF
--- a/src/core.c/CompUnit/PrecompilationRepository.rakumod
+++ b/src/core.c/CompUnit/PrecompilationRepository.rakumod
@@ -39,6 +39,7 @@ class CompUnit::PrecompilationRepository::Default
     my $loaded        := nqp::hash;
     my $resolved      := nqp::hash;
     my $loaded-lock   := Lock.new;
+    my $resolved-lock := Lock.new;
     my $first-repo-id;
 
     my constant $compiler-id =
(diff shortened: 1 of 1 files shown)


src/core.c/CompUnit/PrecompilationRepository.pm6
```diff
@@ -174,19 +175,20 @@ Need to re-check dependencies.")
...
-                          if $!RMD;
-
-                        $comp-unit and $comp-unit.repo-id eq $dependency.id
-                    },
-                    $loaded-lock.protect({ nqp::bindkey($resolved,$serialized-id, 1) }),
-                    (return False)));
+                $resolved-lock.protect: {
+                    nqp::ifnull(
+                      nqp::atkey($resolved,$serialized-id),
+                      nqp::bindkey($resolved,$serialized-id, do {
```
patrickbkr, line 181: I think this would be less convoluted when changing the `do {}` to `nqp::stmts()` (or a block) and moving the `nqp::bindkey` after the `return False`.
(Or I miss the point. :-P)

niner: Honestly I would be surprised if all the nqp stuff here was worth it. This is not code that runs tens of thousands of times.

PR #5301, opened 2023-06-29
RakuAST: allow → ↔ as unicode version of -> <->

lizmat:
As discussed on https://github.com/rakudo/rakudo/wiki/save-me-from-ASCII


jonathanstowe:
I like these because I can get them with 'compose'....

PR #5302, opened 2023-06-29
RakuAST: allow ⇒ as the unicode version of =>

lizmat:
As discussed on https://github.com/rakudo/rakudo/wiki/save-me-from-ASCII


jonathanstowe:
I actually thought it *did* support that until I tried it, I think it was a flashback to Scala.  Also '→' would be nice when you're in there :-D

lizmat:
That'd be https://github.com/rakudo/rakudo/pull/5301

PR #5304, opened 2023-07-01
Remove nqp-home/lib and rakudo-home/lib from vm search paths

ugexe:
`vm_search_path` used to provide some library paths for MoarVM that are now provided by other methods. This removes these redundant search paths for MoarVM similar to what we do for the JavaScript back end.

### DIFF
--- a/src/vm/moar/ModuleLoaderVMConfig.nqp
+++ b/src/vm/moar/ModuleLoaderVMConfig.nqp
@@ -2,8 +2,6 @@ role Perl6::ModuleLoaderVMConfig {
     method vm_search_paths() {
         my @search_paths;
         @search_paths.push(nqp::gethllsym('default', 'SysConfig').rakudo-home() ~ '/lib');
-        # XXX CHEAT: Goes away when we implement :from<nqp>.
-        @search_paths.push(nqp::gethllsym('default', 'SysConfig').nqp-home() ~ '/lib');
         @search_paths
     }
 
--- a/src/vm/moar/ModuleLoaderVMConfig.nqp
+++ b/src/vm/moar/ModuleLoaderVMConfig.nqp
@@ -1,7 +1,6 @@
 role Perl6::ModuleLoaderVMConfig {
     method vm_search_paths() {
         my @search_paths;
-        @search_paths.push(nqp::gethllsym('default', 'SysConfig').rakudo-home() ~ '/lib');
         @search_paths
     }
 


patrickbkr:
Did you try this outside of CI? I suspect this breaks the installed executables. (The ones generated in the build directory are probably fine.)
Why do I think so: Searching for `nqp-home` in rakudo and nqp sources yields no other place that uses that path. (Except for accessing the `profiler/template.html`).
It seems that the nqp-home path is missing in the NQP module loader: https://github.com/Raku/nqp/blob/492fe69d1aea8f8823fa2a5144185284167bce4f/src/vm/moar/ModuleLoader.nqp#L267 (same for the JS and JVM ones)

ugexe:
It is still early for me so maybe I'm overlooking it, but what actions can I take to demonstrate the executable being broken now?

patrickbkr:
> It is still early for me so maybe I'm overlooking it, but what actions can I take to demonstrate the executable being broken now?

I would expect about nothing to work with the nqp libraries missing. To be extra sure, try doing something that loads a module.

The only idea I have, why it might work is that the `rakudo` executable in the build directory is built differently and doesn't rely on `rakudo-home` or `nqp-home`. So try using the one generated in the installation directory after a `make install`.

Or I'm wrong and this somehow works in a way I don't understand...

ugexe:
I'll test that later today sometime if someone else doesn't beat me to it. I wonder if we should have a CI test that does something like run `make test` but with the installed executable

patrickbkr:
Actually there is: https://github.com/rakudo/rakudo/blob/main/azure-pipelines.yml#L208
And that test did succeed with this PR. But why?

ugexe:
> try doing something that loads a module

```
$ raku -e 'use Test; say 42;'
42

$ raku -e 'use NQPHLL:from<NQP>; say 42;'
42
```

It seems to work 🤔

niner:
Nothing worse than when something works when you think it shouldn't...

ugexe:
Everything works even if I make vm_search_paths return an empty array 🤔

patrickbkr:
Ugh, it's too long since I wrote this... :-(
Doing some research:

I think https://github.com/rakudo/rakudo/blob/main/src/vm/moar/runner/main.c#L428-L446 is the reason these paths are not needed in Moars ModuleLoaderVMConfig. The paths are needed very early on for Moar to find Rakudo itself and in turn the module loader. That's why the paths are added in the runner.

I notice that neither the [JVM](https://github.com/rakudo/rakudo/blob/78dee9569c88e042ea4ac9fcc8403b7fde56ff0c/src/vm/jvm/ModuleLoaderVMConfig.nqp) nor [JS](https://github.com/rakudo/rakudo/blob/78dee9569c88e042ea4ac9fcc8403b7fde56ff0c/src/vm/js/ModuleLoaderVMConfig.nqp) ModuleLoaderVMConfigs return the nqp/rakudo-home lib dirs. Removing those paths in Moars ModuleLoaderVMConfig would be consistent.

I notice that the rakudo-home directory not only contains the `lib/` and `runtime/` directories, but also the `core/`, `site/` and `vendor/` CURIs. Those are added [here](https://github.com/rakudo/rakudo/blob/78dee9569c88e042ea4ac9fcc8403b7fde56ff0c/src/core.c/CompUnit/RepositoryRegistry.pm6#L149-L158). So there still is use for the [rakudo-home retrieval logic](https://github.com/rakudo/rakudo/blob/78dee9569c88e042ea4ac9fcc8403b7fde56ff0c/src/Perl6/SysConfig.nqp#L16-L28) in Rakudo land.

@ugexe It seems you are right. nqp-home and rakudo-home are both not needed and can go.

PR #5308, opened 2023-07-01
Remove unused transcode compiler option

ugexe:
This appears to have only been used by Parrot. Every other backend has their apply_transcodings function in nqp return the original input without any modifications. This change stops passing the transcoding option.


ugexe:
See: https://github.com/Raku/nqp/pull/804

PR #5309, opened 2023-07-04
Increase max heap space to 6G for building jvm rakudo

ugexe:
This allows building rakudo on the jvm to compile the setting once again.


ugexe:
I noticed the JVM builds on azure pipelines is failing due to running out of heap space, so I'm seeing what happens when it is increased.

ugexe:
With this it gets further in the build process. Now it dies with:
> Impossible coercion from BOOTInt into Int: value is of unacceptable type BOOTInt

PR #5310, opened 2023-07-07
Add more Version / Range capabilities

lizmat:
- add support for ^v1.2.3 aka all versions upto/not including v1.2.3
- add smart-matching to allow v1.1 ~~ v1.0 .. v1.2 to work

### DIFF
--- a/src/core.c/Range.pm6
+++ b/src/core.c/Range.pm6
@@ -414,6 +414,18 @@ my class Range is Cool does Iterable does Positional {
     multi method ACCEPTS(Range:D \SELF: Junction:D $topic) {
         $topic.THREAD: { SELF.ACCEPTS($_) }
     }
+    multi method ACCEPTS(Range:D: Version:D $version) {
+        my $first := $version cmp $!min;
+        nqp::if(
+          nqp::eqaddr($first,Order::More)
+            || (nqp::not_i($!excludes-min) && nqp::eqaddr($first,Order::Same)),
+          nqp::hllbool(
+            nqp::eqaddr((my $last := $version cmp $!max),Order::Less)
+              || (nqp::not_i($!excludes-max) && nqp::eqaddr($last,Order::Same))
+          ),
+          False
+        )
+    }
     multi method ACCEPTS(Range:D: Cool:D \got) {
         nqp::if(
             $!is-int && nqp::istype(got, Int),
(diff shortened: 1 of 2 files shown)


src/core.c/Range.pm6
```diff
@@ -786,6 +798,9 @@ multi sub prefix:<^>(Int:D $max) {
     $range
 }
 multi sub prefix:<^>($max) { Range.new(0, $max.Numeric, :excludes-max) }
+multi sub prefix:<^>(Version:D $max) {
```
JJ, line 801: Are you sure about the placement of this sub? Did you consider putting it in the same file as Version, to be consistent with the one above (`Range` as argument, placed in `Range.pm6`)

JJ (approved):
:+1:

src/core.c/Range.pm6
lizmat: Good catch.  Fixed with https://github.com/rakudo/rakudo/commit/8378794657

ugexe:
I think the `^` stuff is going to be confusing.

While raku versions aren't semver, users will often mistakeningly believe it to be the case. This is unfortunate because while it makes sense for e.g. `^1` to be anything less than 1 in raku, in semver it means `$ver >= 1 && $ver < 2`. I'm not sure what to do about this.

Using `^v1.0` is not generally useful enough. When declaring dependencies we do not use the `v`, and thus one would expect to write `Foo:ver<^1.0>`, not `Foo:ver<^v1.0>`. For example:
```
$ raku -e 'say Version.new("v1.0")'
vv.1.0
```
In other words: `Version.new(...)` needs to be able to parse and understand this new mechanism similar to how it handles `+` and `*`. Users should be able to write `Version.new("^1.0")` (and probably `v^1.0` as well).

lizmat:
@thoughtstream ping

Leont:
This think this is useful to have, but versions and version ranges should be different classes because they're really different (and complementary) things.

lizmat:
> versions and version ranges should be different classes because they're really different (and complementary) things.

Agree.  But the only thing I can see happening atm, is to make Version objects that are implicit ranges (aka, with a `*`, `-` or `+`) a subclass of `Version`, as a `VersionRange`class.

Leont:
> Agree. But the only thing I can see happening atm, is to make Version objects that are implicit ranges (aka, with a *, - or +) a subclass of Version, as a VersionRangeclass.

Except the relation is really more the other way around: Any version is also usable as a version range, but version ranges usually aren't usable as a version (e.g. they can't sort).

They should be different classes, with a trivial VersionRange(Version) constructor/coercer

lizmat:
Well, then we get to `Range` with `Version` objects.  And that's another can of worms, with iterable and non-iterable Ranges.

PR #5312, opened 2023-07-13
Implement support for role's COMPOSE submethod in Raku v6.e

vrurg:
The submethod is implementing the funcionality many expect from a phaser of the same name. But the submethod can be considered more fit to the task as it's kind of getting into the line with construction submethods `BUILD` and `TWEAK` with the difference that it is role-specific and it is about type-composition time, not object-composition.

The submethod is invoked on role concretization before it gets applied to a class.

### DIFF
--- a/src/Perl6/Metamodel/RoleToClassApplier.nqp
+++ b/src/Perl6/Metamodel/RoleToClassApplier.nqp
@@ -108,6 +108,21 @@ my class RoleToClassApplier {
                                 && (!nqp::istype($!to_compose_meta, Perl6::Metamodel::LanguageRevision)
                                     || $!to_compose.HOW.language_revision($!to_compose) < 3);
 
+        if @!roles && !$with_submethods && nqp::can($!target.HOW, 'concretizations') {
+            my @concs := $!target.HOW.concretizations($!target, :local);
+            my $idx := nqp::elems(@concs);
+
+            while --$idx >= 0 {
+                my $conc := @concs[$idx];
+                my $how := $conc.HOW;
+                if nqp::can($how, 'submethod_table')
+                    && nqp::existskey( (my %submethod-table := $how.submethod_table($conc)), 'COMPOSE' )
+                {
+                    %submethod-table<COMPOSE>($conc);
+                }
+            }
+        }
+
         # Compose in any methods.
         sub compose_method_table(@methods, @method_names) {
             my $method_iterator := nqp::iterator(@methods);


vrurg:
This is an experimental implementation for my proposal in Raku/problem-solving#376. With the following snippet the submethods would be invoked for R0, R1, R2 in this order:

```raku
use v6.e.PREVIEW;

role R0 { 
    submethod COMPOSE {
        note "COMPOSE on ", self.^name, " of ", self.HOW.^name, " for ", ::?CLASS.^name;
    }
}
role R1 does R0 { 
    submethod COMPOSE {
        note "COMPOSE on ", self.^name;
    }
}
role R2 does R0 { 
    submethod COMPOSE() {
        note "COMPOSE on ", self.^name, " // ", self.HOW.^name, " // ", self ~~ R2;
    }
}

class C does R1 does R2 { }
```

Xliff:
Will proto methods have their candidates available at compose time?

I will grab it and check, but if you happen to test this and find out, please let me know.

Here is the code I am using for testing:

```raku
use Method::Also; 

class A { method a is also<b> { 1 }; }; 
my $a = A.new;
$a.b.say; 
role R { 
  proto method c_d (|) is also<d-e> { * }; 
  multi method c_d { 43 }; 
  multi method c_d (Str $a) { "a" }; 
}; 
$a does R; 
$a does AliasesFromRole[R]; 
$a.c_d.say;
$a.d-e.say;
```

Please note, I am working from my copy of Method-Also at https://github.com/Xliff/Method-Also, branch `role-body`.

Xliff:
So... after a quick compile and a small change, here is my revised test code:

```raku
use v6.e.PREVIEW; 

use Method::Also; 

class A { 
  method a is also<b> { 1 }; 
}; 

my $a = A.new;

$a.b.say; 

role R { 
  also does AliasedMethods; 

  proto method c_d (|) is also<d-e> { * }; 

  multi method c_d { 43 }; 

  multi method c_d (Str $a) { "a" }; 
}; 

$a.c_d.say; 

$a.d-e.say;

**Comment by vrurg** [comment]
@Xliff unfortunately, I wouldn't be able to help with `Method::Also` due to lack of time. Having half a day for this experiment was kind of a miracle. But I have a few notes.

First of all, speaking of the `COMPOSE` itself, there is no reason for it to fire _after_ role application is complete. In my view best time for it is when the target class has done all the primary preparations and is ready to accept objects from roles. Considering that composing multis is a process that requires data from roles, this stage must happen after the application. With this in mind, best time for `COMPOSE` is after all class attributes and their accessors are installed. At this point multis are not incorporated yet.

Next, you operate on the class. This is a mistake. Role HOWs are consumers of `Metamodel::MultiMethodContainer`. I'm not sure what approach would work the best, but in my opinion it must include the step where you operate on multi method candidates. Either by overriding `add_multi_method` to intervene at the early stages; or by iterating over `multi_methods_to_incorporate` at the later stages.

**Comment by Xliff** [comment]
Can a role override `.add_multi_method `?

**Comment by vrurg** [comment]
Continuing our IRC conversation, a role – can't, unless it is a role mixed into HOW. See below:

```raku
role AliasHOW {
    has %!aliases;

    method register-alias(\obj, $from, $to) {
        %!aliases{$from} = $to;
    }

    method add_multi_method(Mu \obj, $name, \code-obj) is raw {
        note "ADDING MULTI ", $name, " ", code-obj.raku;
        with %!aliases{$name} {
            note "  $name -> $_";
            callwith(obj, $_, code-obj);
        }
        callsame
    }
}

multi sub trait_mod:<is>(Method:D \m, Str :$alias!) {
    note "Alias $alias for ", m.raku;
    m.package.HOW does AliasHOW unless m.package ~~ AliasHOW;

    m.package.^register-alias(m.name, $alias);
}

role R {
    proto method foo(|) is alias<bar> {*}
    multi method foo {
    }
}

class C does R { }

C.bar;
```

And don't forget that this snippet only makes sense for the trait applied to a `proto`.

PR #5314, opened 2023-07-15
Implement methods `andthen` and `orelse` on Promise

vrurg:
The methods are intended to simplify conditional processing and chaining of kept or broken promises when used instead of the `then` method.

`andthen` is fired only when its invocant promise is kept. If the invocant is broken then its cause becomes the cause of the promise returned by `andthen`:

```raku
my $p = Promise.new;
my $p2 =
    $p.andthen({ note "must not happen!" }) # will print nothing
      .then({ say "cause: ", ~.cause });    # will print "cause: broken"
$p.break("broken");
try await $p2;
```

`orelse` is fired if its invocant is broken. If the invocant is kept then the promise returned by `orelse` is kept with invocant's result:

```raku
my $p = Promise.new;
my $p2 =
    $p.orelse({ note "this must not be printed!" })
      .then({ say "result: ", ~.result });
$p.keep("all good");
await $p2;
```

If the invocant of `orelse` is broken the purpose of the method is to intercept the cause, handle it, and return a value to be used as the result of `orelse`-returned promise:

```raku
my $p = Promise.new;
my $p2 =
    $p.orelse({ "purified" })
      .then({ say "result: ", ~.result });
$p.break("all bad");
await $p2;
```

### DIFF
--- a/src/core.c/Promise.pm6
+++ b/src/core.c/Promise.pm6
@@ -204,6 +204,17 @@ my class Promise does Awaitable {
         }
     }
 
+    method !PLANNED-THEN(\then-promise, \vow, \then-code) {
+        # Push 2 entries to $!thens: something that starts the then code,
+        # and something that handles its exceptions. They will be sent to the
+        # scheduler when this promise is kept or broken.
+        nqp::bindattr(then-promise, Promise, '$!dynamic_context', nqp::ctx());
+        nqp::push(nqp::ifnull($!thens, ($!thens := nqp::list)), then-code);
+        nqp::push($!thens, -> $ex { vow.break($ex) });
+        nqp::unlock($!lock);
+        then-promise
+    }
+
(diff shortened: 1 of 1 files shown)


vrurg:
A bit of a real life example from my current project of Google Cloud API module, in addition to raku/roast#836:

```raku
    Cro::HTTP::Client.get(
        $api-url,
        headers => [
            $.gcloud.http-auth-header,
            |@headers,
        ],
        |%c )
    .andthen({
        await .result.body-text
    })
    .orelse({
        given .cause {
            when X::Cro::HTTP::Error {
                # If content-type is json then pull in the body, produce a GCloud-specific exception object from 
                # the JSON and throw that instead of the original exception.
            }
        }
        .cause.rethrow
    })
```

In some similar cases the last `.orelse` can be prepended with another `.andthen` performing some additional tasks.

PR #5316, opened 2023-07-19
Use new stat syscalls

MasterDuke17:
See https://github.com/MoarVM/MoarVM/pull/1762 for the MoarVM implementations of the syscalls, and a stack trace of the error I get during a Rakudo install.

### DIFF
--- a/src/core.c/IO/Path.pm6
+++ b/src/core.c/IO/Path.pm6
@@ -704,16 +704,14 @@ my class IO::Path is Cool does IO {
     }
 
     method user(IO::Path:D:) {
-        my str $path = self.absolute;
-        my $stat := nqp::dispatch("boot-syscall", "file-stat", nqp::decont_s($path), 0);
+        my $stat := nqp::dispatch("boot-syscall", "file-stat", nqp::decont_s(self.absolute), 0);
         nqp::dispatch("boot-syscall", "stat-exists", $stat)
           ?? nqp::dispatch("boot-syscall", "stat-uid", $stat)
           !! self!does-not-exist("user")
     }
 
     method group(IO::Path:D:) {
-        my str $path = self.absolute;
-        my $stat := nqp::dispatch("boot-syscall", "file-stat", nqp::decont_s($path), 0);
+        my $stat := nqp::dispatch("boot-syscall", "file-stat", nqp::decont_s(self.absolute), 0);
         nqp::dispatch("boot-syscall", "stat-exists", $stat)
           ?? nqp::dispatch("boot-syscall", "stat-gid", $stat)
           !! self!does-not-exist("group")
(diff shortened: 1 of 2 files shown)


src/core.c/Rakudo/Internals.pm6
```diff
@@ -1384,15 +1384,15 @@ my class Rakudo::Internals {
             nqp::while(
               nqp::chars(my str $entry = self!next),
               nqp::stmts(
+                (next if $entry eq '.' or $entry eq '..'),
                 (my str $path = nqp::concat($!abspath,$entry)),
                 (try
                   nqp::if(
-                    $!file.ACCEPTS($entry) &&
-                      nqp::stat($path,nqp::const::STAT_ISREG),
+                      (my $s := nqp::dispatch('boot-syscall', 'file-stat', nqp::decont_s($path), 0)) && nqp::dispatch('boot-syscall', 'stat-is-reg', $s) && $!file.ACCEPTS($entry),
```
ugexe, line 1391: In the final api I wonder if instead of e.g. `stat-is-reg` we should pass `nqp::const::STAT_ISREG`

src/core.c/IO/Path.pm6
```diff
@@ -682,10 +684,11 @@ my class IO::Path is Cool does IO {
         # check if we need a BOM
         if $enc eq 'utf16' {
 
+           my $stat := nqp::dispatch("boot-syscall", "file-stat", nqp::decont_s($path), 0);
            # add a BOM if writing starts at beginning of file
            if $mode eq 'w'
```
ugexe, line 689: Could probably rewrite this conditional so that the `stat` doesn't occur unless `$mode eq 'w'`

ugexe:
If I update CURI with the following:
```
    method !repository-version(--> Int:D) {
        $!version //= do {
            my $version-file = $.prefix.add('version');
            note "VERSION-FILE: {$version-file} {$version-file.e} {$version-file.f}";
            $version-file.f ?? $version-file.slurp.Int !! 0
        }
    }
```
I get the following error:

> Failed to find '/Users/nlogan/repos/rakudo/install/share/perl6/core/version' while trying to do '.f'

ugexe:
It looks like the existence check returns 0 when the file exists (and `-2` when it doesn't?), which would cause the various ternaries to do the opposite of what they should

```
$ rakudo % ./install/bin/raku -e 'use nqp; my $path = "/Users/nlogan/repos/rakudo/install/share/perl6/core/version"; my $stat := nqp::dispatch("boot-syscall", "file-stat", nqp::decont_s($path), 0); say nqp::dispatch("boot-syscall", "stat-exists", $stat); say nqp::dispatch("boot-syscall", "stat-is-reg", $stat)'
0
1

$ rakudo % ./install/bin/raku -e 'use nqp; my $path = "/Users/nlogan/repos/rakudo/install/share/perl6/core/versionXXX"; my $stat := nqp::dispatch("boot-syscall", "file-stat", nqp::decont_s($path), 0); say nqp::dispatch("boot-syscall", "stat-exists", $stat); say nqp::dispatch("boot-syscall", "stat-is-reg", $stat)'
-2
0
```

MasterDuke17:
Nice catch. I’ll look into that.Sent from my iPhoneOn Jul 22, 2023, at 2:19 PM, Nick Logan ***@***.***> wrote:﻿
It looks like the existence check returns 0 when the file exists (and -2 when it doesn't?), which would cause the various ternaries to do the opposite of what they should
$ rakudo % ./install/bin/raku -e 'use nqp; my $path = "/Users/nlogan/repos/rakudo/install/share/perl6/core/version"; my $stat := nqp::dispatch("boot-syscall", "file-stat", nqp::decont_s($path), 0); say nqp::dispatch("boot-syscall", "stat-exists", $stat); say nqp::dispatch("boot-syscall", "stat-is-reg", $stat)'
0
1

$ rakudo % ./install/bin/raku -e 'use nqp; my $path = "/Users/nlogan/repos/rakudo/install/share/perl6/core/versionXXX"; my $stat := nqp::dispatch("boot-syscall", "file-stat", nqp::decont_s($path), 0); say nqp::dispatch("boot-syscall", "stat-exists", $stat); say nqp::dispatch("boot-syscall", "stat-is-reg", $stat)'
-2
0

src/core.c/Rakudo/Internals.pm6
ugexe: From my earlier example it looks like the file/dir stat calls return `< 1` if the file doesn't exist. Since we don't return any exceptions here to distinguish the difference between a non-existent file and a file that isn't a file or directory, we can probably get by without the existence check at all inside DIR-RECURSE

ugexe: Ah n/m, it is already doing that here

MasterDuke17:
This now passes `make m-test m-spectest` when using a MoarVM with https://github.com/MoarVM/MoarVM/pull/1762 applied and calling `.rwx` on a file a million times drops from 3.7s to 1.5s.

src/core.c/IO/Path.pm6
MasterDuke17: Implemented in [181921e](https://github.com/rakudo/rakudo/pull/5316/commits/181921e0582e327d1bfe089836aff90397d6b985). As is it requires repeating the actual BOM writing logic, but if this is deemed to violate DRY, it could be changed to just set a flag.

src/core.c/Rakudo/Internals.pm6
MasterDuke17: Yeah, I was wondering about that. Should it be more like `nqp::stat` just implemented with syscalls? Or more like it is now? My WAG is that the way it is now is more spesh-able, but since we're dealing with file io, maybe not so much more as to be noticeable?

src/core.c/IO/Path.pm6
```diff
@@ -766,92 +779,122 @@ my class IO::Path is Cool does IO {
         nqp::hllbool(Rakudo::Internals.FILETEST-E(self.absolute))
```
ugexe, line 779: Missed this one

MasterDuke17: Ha. I did look at it, but changing to use syscalls won't reduce the number of `stat`s called, so I left it as-is. `Rakudo::Internals.FILETEST-E` does just one `nqp::stat`. If we want to completely remove all uses of `nqp::stat`, I'll make the change, but will then need to make more changes elsewhere.

src/core.c/Rakudo/Internals.pm6
```diff
@@ -1384,15 +1384,15 @@ my class Rakudo::Internals {
             nqp::while(
               nqp::chars(my str $entry = self!next),
               nqp::stmts(
+                (next if $entry eq '.' or $entry eq '..'),
```
ugexe, line 1387: This does probably need to be removed in the final version as both `$!file` and `$!dir` are configurable (with `$!file` defaulting to matching everything).

src/core.c/Rakudo/Internals.pm6
```diff
@@ -1384,15 +1384,15 @@ my class Rakudo::Internals {
...
                 (my str $path = nqp::concat($!abspath,$entry)),
                 (try
```
ugexe, line 1389: I think this `try` can be removed now

src/core.c/Rakudo/Internals.pm6
ugexe, line 1391: It might be more clear putting the `my $s := nqp::dispatch('boot-syscall', 'file-stat', nqp::decont_s($path), 0))` as its own line in the `nqp::stmts`

src/core.c/IO/Path.pm6
```diff
@@ -682,27 +684,38 @@ my class IO::Path is Cool does IO {
...
+            }
         }
 
         spurt-blob($path, $mode, $blob)
     }
 
     method user(IO::Path:D:) {
-        Rakudo::Internals.FILETEST-E(my str $path = self.absolute)
-          ?? nqp::stat($path,nqp::const::STAT_UID)
+        my str $path = self.absolute;
```
ugexe, line 707: Probably doesn't matter, but I wonder if it would be slightly better to call `self.absolute` in void context (instead of assigning to `$path`) to populate `$!os-path`, and then to just us `$!os-path` instead of `$path`

src/core.c/IO/Path.pm6
```diff
@@ -497,12 +497,13 @@ my class IO::Path is Cool does IO {
               :path($!os-path), :$uid, :$gid, :os-error(.Str) );
         }}
         my str $path = self.absolute;
+        my $stat := nqp::dispatch("boot-syscall", "file-stat", nqp::decont_s($path), 0);
```
ugexe, line 500: The `stat` here isn't technically needed unless `$uid.defined && $gid.defined && $uid ~~ UInt && $gid ~~ Uint`. Dunno if it's worth rewriting though.

src/core.c/IO/Path.pm6
MasterDuke17: Better how/why?

ugexe:
Can you speak to the thread safety of this? For instance it seems like `nqp::dispatch('boot-syscall', 'stat-is-reg', $s)` would be updating `$s`

MasterDuke17:
> Can you speak to the thread safety of this? For instance it seems like `nqp::dispatch('boot-syscall', 'stat-is-reg', $s)` would be updating `$s`

No, none of the `stat-*` syscall modify the passed in MVMStat, they just read from it. E.g., https://github.com/MoarVM/MoarVM/pull/1762/files#diff-d8a00dcd6dd0352fd379e6ed5abbdd85adf5c64fc1c8bddcb758e2c545ff01bfR1341-R1347

src/core.c/IO/Path.pm6
ugexe: No longer need to create $path or assign to it

MasterDuke17: FWIW, given this test script: `MVM_SPESH_BLOCKING=1 raku -e 'my $a = "VERSION".IO; my $b; my $s := now; $b := $a.user for ^1_000_000; say now - $s; say $b'`
* This PR as is reports ~1.25s.
* Changing `my $stat := nqp::dispatch("boot-syscall", "file-stat", nqp::decont_s($path), 0);` to `my $stat := nqp::dispatch("boot-syscall", "file-stat", nqp::decont_s(self.absolute), 0);` gives ~1.13s.
* Changing `my $stat := nqp::dispatch("boot-syscall", "file-stat", nqp::decont_s($path), 0);` instead to `self.absolute; my $stat := nqp::dispatch("boot-syscall", "file-stat", nqp::decont_s($!os-path), 0);` give ~1.14.

I originally didn't make that change (just using `self.absolute` instead of assigning it to a `$path`) to minimize the diff, but seeing the performance improvement, I'll add a new commit doing it everywhere.

lizmat:
So, is this ready for merge now?

MasterDuke17:
> So, is this ready for merge now?

It's ready in the sense that when using a MoarVM with https://github.com/MoarVM/MoarVM/pull/1762 applied, it passes `make m-test m-spectest`. @ugexe has raised an API question (and I agree that the question is something to be thought about, I'm not 100% convinced either way) about whether the syscalls should look more like the current `npq::stat` (i.e., just pass in one or more of the existing `nqp::const::STAT_*` constants to a new syscall instead of effectively a syscall for each constant). It may be a simple enough change that I can raise a new pair of MoarVM/Rakudo PRs soon and we can decide between them.

PR #5326, opened 2023-07-31
Don't mangle with exception backtrace

vrurg:
`.throw` was previously used causing re-written backtrace. Since both fixed locations are not actually producing exceptions but using `CATCH` to do a last-second fix for the `GLOBAL` HLL symbol, trace-preserving `.rethrow` makes better sense.


lizmat (approved):

vrurg:
@ugexe I don't think I missing something here, but I don't like playing on the `CompUnit` field without your participation. :)

ugexe (approved):

PR #5327, opened 2023-07-31
special cases should not be suggested like real methods

gfldex:
Currently, ` class X::Method::NotFound` may suggest introspection macros as if they where real methods and by doing so, suggest the very method that is missing.

```
    say Any."WHAT"();
```

## OUTPUT
```
No such method 'WHAT' for invocant of type 'Any'.  Did you mean any of
these: 'WHAT', 'WHO', 'WHY', 'flat'?
```

As discussed [here](https://irclogs.raku.org/raku/2023-07-22.html#16:44), this is wrong.

## Test
```
use v6.*;
use Test;

my $a = 42;
my @test;

for <HOW WHAT WHO> -> $m {
    try { say $a."$m"(); CATCH { default { @test.push: $m => .message.match(/(<$m>)/).list.elems } } }
}

ok @test».value.all == 1, ‘introspection macros wont be suggested like real methods’;
```

### DIFF
--- a/src/core.c/Exception.pm6
+++ b/src/core.c/Exception.pm6
@@ -283,7 +283,7 @@ my class X::Method::NotFound is Exception {
             }
 
             # handle special unintrospectable cases
-            for <HOW WHAT WHO> -> $method_name {
+            for <HOW WHAT WHO>.grep(* ne $.method) -> $method_name {
                 find_public_suggestion($.method, $method_name);
             }
         }


raiph:
Wouldn't it be better, if the attempted method name matches a "macro" (`WHAT` etc), to either do as the user meant and not generate an error, or, if that can't be done for some reason, tell the user they can't do so (because such-and-such), rather than suggest methods that one already knows almost certainly won't be what they meant?

lizmat:
With 7bc98ec4158dd8ae59d0d9caedef2a9084d3fd72 this is no longer an error, as there are "true" methods by that name which are doing the right thing.

Therefore I'm closing this PR.  Please re-open if you disagree.

PR #5328, opened 2023-07-31
Fix use of CALLERS/CALLER pseudo stashes

vrurg:
There is seemingly a confusion about meaning of `CALLERS` and `CALLER` namespaces. They both are currently bound to dynamic symbols only. Where they used `CALLER::LEXICAL` is more likely to do the job.


vrurg:
This case has also reminded me about raku/problem-solving#294. The point is that `CALLER` is often expected to actually be `CALLER::MY`. And this is what I suggest in the problem solving issue.

lizmat (approved):

PR #5336, opened 2023-08-10
Add sub versions of WHERE WHICH WHO

lizmat:
They seem to have been forgotten, but are actually mentioned in some error messages.  Feels they should be added for completeness sake.  Also adapts tests accordingly.


vrurg:
On the par, can we consider adding methods too? `HOW`, `VAR`, `WHAT`, `WHO`. That'd be a big plus to introspecting. Besides, I personally was once in a situation where indirect invocation of one of these would simplify my code.

vrurg (approved):

lizmat:
> On the par, can we consider adding methods too? HOW, VAR, WHAT, WHO

Added with https://github.com/rakudo/rakudo/commit/7bc98ec415

PR #5338, opened 2023-08-12
Allow default CURIs to be disabled

ugexe:
Previously there was no convenient way to disable the e.g. site, vendor, and core repositories. But when developing with application specific module install locations one might not want for modules to be used outside of that location. This add a new environment variables, RAKU_DISABLE_DEFAULT_CURIS, which can be set to anything to prevent the default repositories from being added to the repo chain itself. Note that it does still register those repositories with the registry.

### DIFF
--- a/src/core.c/CompUnit/RepositoryRegistry.pm6
+++ b/src/core.c/CompUnit/RepositoryRegistry.pm6
@@ -171,38 +171,48 @@ class CompUnit::RepositoryRegistry {
             $home-spec = 'inst#' ~ $home;
         }
 
+        my $disable-default-curis := $ENV<RAKU_DISABLE_DEFAULT_CURIS>:exists;
+
         unless $precomp-specs {
-            $next-repo := self!register-custom-lib-repository(
-              'core', $core,
-              CompUnit::Repository::Installation.new(
-                :prefix("$prefix/core"),
-                :$next-repo
-              )
-            ) unless nqp::existskey($unique,$core);
+            unless nqp::existskey($unique,$core) {
+                my $repo := self!register-custom-lib-repository(
+                  'core', $core,
+                  CompUnit::Repository::Installation.new(
+                    :prefix("$prefix/core"),
+                  )
+                );
+                $repo.next-repo = $next-repo unless $disable-default-curis;
+            }
 
-            $next-repo := self!register-custom-lib-repository(
-              'vendor', $vendor,
-              CompUnit::Repository::Installation.new(
-                :prefix("$prefix/vendor"),
-                :$next-repo
-              )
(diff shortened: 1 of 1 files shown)


ugexe:
### Allow disabling default CURIs

One pattern developers might want to follow is to install all application dependencies into an app-specific location i.e. `./local/inst`. They might want to be sure that their application only runs with the dependencies installed to `./local/inst`, and thus set `RAKULIB="./local/inst"`. However, that only adds the repository to the front of the repo chain. By default there are 4 other `CompUnit::Repository::Installation` behind whatever is set by e.g. `RAKULIB`:

```
inst#/Users/nlogan/.raku
inst#/Users/nlogan/.rakubrew/versions/moar-blead/install/share/perl6/site 
inst#/Users/nlogan/.rakubrew/versions/moar-blead/install/share/perl6/vendor 
inst#/Users/nlogan/.rakubrew/versions/moar-blead/install/share/perl6/core
```

But when a module isn't found by `./local/inst` I don't want it to check these other repositories. I want to be sure that the modules used by the application are self contained to `./local/inst` only. For example:

```
$ RAKU_DISABLE_DEFAULT_CURIS="1" ./install/bin/rakudo -e 'use Test; say 42'
Could not find Test in:
    CompUnit::Repository::AbsolutePath<6433162664288>
    CompUnit::Repository::NQP<6433121507744>
    CompUnit::Repository::Perl5<6433121507784>
at -e:1

$ RAKU_DISABLE_DEFAULT_CURIS="1" RAKULIB="./local/inst" ./install/bin/rakudo -e 'use Test; say 42'
42
```

ugexe:
I haven't thought too much on how to best implement this, but this PR is one way to do it.

hythm7:
Looks like this will disable the `core` repo as well, which includes essential modules, like `NativeCall`, I haven't tested this, but wouldn't that break modules that depend on `NativeCall`?

And if that is the case, I think it will be better to allow disabling indvidual repos (eg. `home` or `home,site`).

Another option may be have a generic env var to specify the allowed repos and their order in the repo chain, for example `site,home,core,nqp` should build the repo chain as, first repo is  `site`, next repo `home`, next `core` and last repo `nqp`.

ugexe:
In a self contained environment as described one would also install the core modules. Just because NativeCall doesn’t exist in an ecosystem doesn’t mean something like zef —contained option can’t obtain it from somewhere else (like the current rakudo installation)

ugexe:
You can accomplish custom repo chain by just disabling them all (as in this PR) and set RAKULIB. I was originally going to write such a custom repo chain mechanism, but this seemed cleaner to me, and far easier to implement.

Issue #5311, opened 2023-07-08
The REPL msg should suggest 'Linenoise' only until 'Readline' is working


tbrowder:
## The Problem

Readline is broken and its author has passed from this Earth. Until it is working again, it should not be mentioned unless there is also a message about needing a volunteer to fix it.

## Environment
* Operating system:
* Compiler version (`perl6 -v` or `raku -v`):

coke:
https://github.com/lathropd/perl6-readline is the current owner, and I don't see any issues there that readline is broken.

coke:
The original repo DOES have issues, but no indication the new owner has them.

tbrowder:
Done.

PR #5339, opened 2023-08-13
Change REPL recommendation order

coke:
Part of #5311

Instead of removing readline, at least mention it last.


coke:
... this doesn't work; installing lineEditor, still get the message.

PR #5350, opened 2023-08-18
Provide a workaround for macOS Security Integrity Protection

vrurg:
A common complain about 'raku is loading libsomething in an unsafe way' doesn't have a single universal solution. Here I try to provide a workaround to it by enforcing DYLD_LIBRARY_PATH variable from a Rakudo-specific RAKUDO_DYLD_LIBRARY_PATH. This, for example, provides a fix for installing OpenSSL module.

### DIFF
--- a/src/vm/moar/runner/main.c
+++ b/src/vm/moar/runner/main.c
@@ -162,7 +162,7 @@ int retrieve_home(
 #if defined(_WIN32) && defined(SUBSYSTEM_WINDOWS)
 int set_std_handle_to_nul(FILE *file, int fd, BOOL read, int std_handle_type) {
     /* Found on https://docs.microsoft.com/en-us/cpp/c-runtime-library/reference/get-osfhandle?view=vs-2019:
-       
+
        "When stdin, stdout, and stderr aren't associated with a stream (for example, in a Windows
        application without a console window), the file descriptor values for these streams are
        returned from _fileno as the special value -2. Similarly, if you use a 0, 1, or 2 as the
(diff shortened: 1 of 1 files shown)


lizmat:
Wouldn't specifying `DYLD_LIBRARY_PATH` have the same effect?  Why the extra indirection?

vrurg:
@lizmat From `man dyld`:

> Note: If System Integrity Protection is enabled, these environment variables are ignored when executing binaries protected by System Integrity Protection.

Up to my understanding, everything in _/bin_ and _/usr/bin_ is protected, including _/usr/bin/env_ and, sure enough _/bin/sh_. Here comes the problem: scripts we get installed with distributions – they're all shebanged with `#!/usr/bin/env`. This is the most apparent issue which causes many tests to fail for modules, depending on 3rd party packages installed either with Homewbrew or MacPorts.

Ok, let's say we find a workaround for that. Next layer of problems is with scripts/apps spawning external processes using `shell` or whichever other way that uses `/usr/bin/sh`. Apparently, the workaround would only work for Rakudo binary, not for anything else.

Hence the primary idea: take an environment variable which is not affected by SIP and use it to re-instate `DYLD_LIBRARY_PATH`.

Actually, it'd be better to have the same code in MoarVM itself, but neither I found a good name for the variable, nor NQP is so heavily reliant on loading native libs to get bothered about it now.

ugexe:
Some writings on the subject - https://hynek.me/articles/macos-dyld-env/

PR #5352, opened 2023-08-19
Use new stat syscalls

MasterDuke17:
A second version of https://github.com/rakudo/rakudo/pull/5316 (performance is the same) that requires https://github.com/MoarVM/MoarVM/pull/1766 for the MoarVM implementations of the syscalls.


ugexe (approved):
Nice!

vrurg:
What about JVM backend?

Leont:
I've been wanting this for `Path::Finder` for years, good work

MasterDuke17:
Arg! Guess I’ll have to wrap the changes in ‘#if moar’.

PR #5361, opened 2023-08-26
Un-nqpify *.combinations() without slowing it down

MasterDuke17:
This reduces the amount of `nqp::*` ops used in the implementation of `*.combinations()`. In my timing of `say (1..100).combinations(4).tail; say now - INIT now` this causes no noticeable slowdown, any further reductions do. There are no logical changes, but it might make it easier to experiment with alternate algorithms or optimizations.

### DIFF
--- a/src/core.c/Rakudo/Iterator.pm6
+++ b/src/core.c/Rakudo/Iterator.pm6
@@ -831,37 +831,25 @@ class Rakudo::Iterator {
             my int $running  = 1;
             my $combination := $!combination;
             my $stack       := $!stack;
+            my int $index;
 
-            nqp::while(
-              ($running && (my int $elems = nqp::elems($stack))),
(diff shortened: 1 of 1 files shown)


lizmat:
Are you sure this produces the same QAST?   And will also not slow down small combinations?

MasterDuke17:
> Are you sure this produces the same QAST? And will also not slow down small combinations?

QAST I don't know about. It maybe slows down `my $a; $a = (1..10).combinations(2).tail for ^10_000; say now - INIT now; say $a` by about 0.1 - 0.05s. The timings are pretty variable, even with `MVM_SPESH_BLOCKING=1`.

lizmat:
Revived with https://github.com/rakudo/rakudo/commit/693f15ead2

Issue #5353, opened 2023-08-20
RakuAST: Role inheritence is missing


ab5tract:
## The Problem

```
role R {}
role S does R {}
S ~~ R
```

## Expected Behavior

`True`

## Actual Behavior

`False`

## Environment
* Operating system: `Ubuntu Linux 22.04.1`
* Compiler version (`rakudo -v` or `raku -v`): `v2023.06`

PR #5363, opened 2023-08-26
RakuAST: Fix role inheritence not working

ab5tract:
This branch fixes issue #5353 .

### DIFF
--- a/src/Raku/ast/package.rakumod
+++ b/src/Raku/ast/package.rakumod
@@ -337,12 +337,16 @@ class RakuAST::Package
         if $!declarator eq 'role' {
             $type.HOW.set_body_block($type, $!body.meta-object);
 
+            # The role needs to be composed before we add the possibility to the group
+            $type.HOW.compose($type);
+
             my $group := $!role-group;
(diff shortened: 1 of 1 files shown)


vrurg (approved):

src/Raku/ast/package.rakumod
```diff
@@ -337,12 +337,16 @@ class RakuAST::Package
         if $!declarator eq 'role' {
             $type.HOW.set_body_block($type, $!body.meta-object);
 
+            # The role needs to be composed before we add the possibility to the group
+            $type.HOW.compose($type);
+
             my $group := $!role-group;
-            $group.HOW.add_possibility($group, $type) unless $group =:= Mu;
+            $group.HOW.add_possibility($group, $type) unless $group =:= Mu || $group =:= $type;
```
niner, line 344: In what case would $group be the same as $type?

ab5tract: Good catch. This was logic ported from `Perl6/Actions`. I just double checked and it passes the same number of spectests without this change.

PR #5366, opened 2023-08-29
RakuAST: Fix t/spec/S14-roles/basic.t 

ab5tract:
This branch has a series of commits that were required to pass `t/spec/S14-roles/basic.t `.

They have been grouped into separate commits based on each required fix.

The result is ~+3~ +4 spectests passing locally.

### DIFF
--- a/src/Raku/ast/variable-declaration.rakumod
+++ b/src/Raku/ast/variable-declaration.rakumod
@@ -726,10 +726,11 @@ class RakuAST::VarDeclaration::Simple
                     }
 
                     my $value := $expression.compile-time-value;
-                    unless nqp::istype($value,$vartype) {
+                    if !nqp::istype($value,$vartype)
+                        && nqp::istype($vartype,$resolver.resolve-name-constant-in-setting(RakuAST::Name.from-identifier('Numeric')).compile-time-value)
+                    {
                         $resolver.add-sorry: $resolver.build-exception:
-                          'X::Syntax::Number::LiteralType',
-                          :varname(self.name), :$vartype, :$value
+                            'X::Syntax::Number::LiteralType', :varname(self.name), :$vartype, :$value
                     }
                 }
             }
--- a/src/Raku/ast/variable-declaration.rakumod
+++ b/src/Raku/ast/variable-declaration.rakumod
@@ -1464,6 +1464,7 @@ class RakuAST::VarDeclaration::Implicit::Constant
   is RakuAST::BeginTime
   is RakuAST::Meta
   is RakuAST::CompileTimeValue
+  is RakuAST::Declaration::Mergeable
 {
     has Mu $.value;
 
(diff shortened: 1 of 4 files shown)


src/Raku/ast/variable-declaration.rakumod
```diff
@@ -649,15 +649,20 @@ class RakuAST::VarDeclaration::Simple
                                 target => RakuAST::ParameterTarget::Var.new('$_')
                             )
                         ])
-                    )),
+                        )),
```
lizmat, line 652: Extra indent?

ab5tract: That's Comma behaving badly. Will fix, thanks!

src/Raku/ast/variable-declaration.rakumod
```diff
@@ -726,10 +731,12 @@ class RakuAST::VarDeclaration::Simple
                     }
 
                     my $value := $expression.compile-time-value;
-                    unless nqp::istype($value,$vartype) {
+                    if !nqp::istype($value,$vartype) &&
+                        # nqp::istype($vartype,Numeric) returns 0 here, so check all relevant types
+                        (nqp::istype($vartype,Int) || nqp::istype($vartype,Num) || nqp::istype($vartype,Rat) || nqp::istype($vartype,FatRat)) {
```
lizmat, line 736: Are these types actually known here?  Or will they just be `NQPMu` at runtime?

src/Raku/ast/variable-declaration.rakumod
```diff
@@ -726,10 +731,12 @@ class RakuAST::VarDeclaration::Simple
...
                         $resolver.add-sorry: $resolver.build-exception:
-                          'X::Syntax::Number::LiteralType',
-                          :varname(self.name), :$vartype, :$value
+                            'X::Syntax::Number::LiteralType',
```
lizmat, line 738: trailing whitespace?

src/Raku/ast/variable-declaration.rakumod
ab5tract: This seems to work as expected with all variants I tried. It even fixes a bug in base where the correct exception was not thrown for `FatRat`.

src/Raku/ast/variable-declaration.rakumod
ab5tract: looks like indent = 2 vs indent = 4 to me

src/Raku/ast/variable-declaration.rakumod
niner: `Int` and `Num` are fine but `Rat` and `FatRat` are not known in the bootstrap and will therefore be silently mapped to NQPMu. If this code works as intended anyway, that suggests that the tests are insufficient to actually trigger this, or there's something else that's fishy. Btw. `Numeric` is also unknown to the bootstrap, which explains why the istype check always fails (it's really `nqp::istype($vartype, NQPMu)`.
If you need these types, you need to resolve them first (see `RakuAST::ImplicitLookups`).

ab5tract: @niner Since I actually have the resolver object in scope and because there are a variable number of implicit lookups, I took the approach of using the resolver directly. I hope that is okay.

lizmat (approved):

ab5tract:
The test regressions are fixed. This should be fully merge-able now.

PR #5377, opened 2023-09-10
Fix of * .. non-number Range scenarios

### LINKED ISSUE #5340: WhateverStar dodges range sanity check
## The Problem
`-Inf .. 'Z'` isn't constructible as a `Range` - except if you write `* .. 'Z'`, in which case it will be created.

## Expected Behavior
An error; either the same one as with literally typing `-Inf .. 'Z'`, or one that states that the WhateverStar cannot be used as the starting point of a `Range` of strings
## Actual Behavior
`-Inf .. 'Z'` is created as a `Range`

## Environment
* Operating system: `Linux ... 4.4.0-19041-Microsoft #2311-Microsoft Tue Nov 08 17:09:00 PST 2022 x86_64 x86_64 x86_64 GNU/Linux`
* Compiler version (`perl6 -v` or `raku -v`): `v2023.06. on MoarVM`

2colours:
Resolves #5340 
The whatever-star is always inferred to mean +/- Inf in Range
constructions, therefore it makes sense to infer the same rules for it as for Inf itself. Since modifying the type constraint changed the order of the resolution of 10 .. ^20 (might be a bug in itself), I rather made a coercion to Real in the function body.

### DIFF
--- a/src/core.c/Range.rakumod
+++ b/src/core.c/Range.rakumod
@@ -58,13 +58,11 @@ my class Range is Cool does Iterable does Positional {
     }
     multi method new(Whatever \min, \max, :$excludes-min, :$excludes-max) {
         use fatal;
-        my \max-real = max.Real;
-        nqp::create(self)!SET-SELF(-Inf,max-real,$excludes-min,$excludes-max,1);
+        nqp::create(self)!SET-SELF(-Inf,max.Real,$excludes-min,$excludes-max,1);
     }
     multi method new(\min, Whatever \max, :$excludes-min, :$excludes-max) {
         use fatal;
-        my \min-real = min.Real;
-        nqp::create(self)!SET-SELF(min-real,Inf,$excludes-min,$excludes-max,1);
+        nqp::create(self)!SET-SELF(min.Real,Inf,$excludes-min,$excludes-max,1);
     }
     multi method new(Real \min, Real(Cool) $max, :$excludes-min, :$excludes-max) {
         nqp::create(self)!SET-SELF(
--- a/src/core.c/Range.rakumod
+++ b/src/core.c/Range.rakumod
@@ -57,10 +57,14 @@ my class Range is Cool does Iterable does Positional {
         nqp::create(self)!SET-SELF(-Inf,Inf,$excludes-min,$excludes-max,1);
     }
     multi method new(Whatever \min, \max, :$excludes-min, :$excludes-max) {
-        nqp::create(self)!SET-SELF(-Inf,max,$excludes-min,$excludes-max,1);
+        use fatal;
+        my \max-real = max.Real;
+        nqp::create(self)!SET-SELF(-Inf,max-real,$excludes-min,$excludes-max,1);
     }
     multi method new(\min, Whatever \max, :$excludes-min, :$excludes-max) {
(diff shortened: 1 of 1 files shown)


src/core.c/Range.rakumod
```diff
@@ -57,10 +57,14 @@ my class Range is Cool does Iterable does Positional {
...
-        nqp::create(self)!SET-SELF(-Inf,max,$excludes-min,$excludes-max,1);
+        use fatal;
+        my \max-real = max.Real;
+        nqp::create(self)!SET-SELF(-Inf,max-real,$excludes-min,$excludes-max,1);
     }
     multi method new(\min, Whatever \max, :$excludes-min, :$excludes-max) {
-        nqp::create(self)!SET-SELF(min,Inf,$excludes-min,$excludes-max,1);
+        use fatal;
+        my \min-real = min.Real;
+        nqp::create(self)!SET-SELF(min-real,Inf,$excludes-min,$excludes-max,1);
```
lizmat, line 67: Why the extra `\min-real`, why not `SET-SELF(min.Real...) ?

2colours: Right, that could probably work. The thing is, downright adding `use fatal` was kind of an afterthought. With that, min.Real would explode in a method call as well, right?

2colours: Yep, changed and tested, this seems to be alright.

src/core.c/Range.rakumod
```diff
@@ -57,10 +57,12 @@ my class Range is Cool does Iterable does Positional {
         nqp::create(self)!SET-SELF(-Inf,Inf,$excludes-min,$excludes-max,1);
     }
     multi method new(Whatever \min, \max, :$excludes-min, :$excludes-max) {
-        nqp::create(self)!SET-SELF(-Inf,max,$excludes-min,$excludes-max,1);
+        use fatal;
```
lizmat, line 60: Hmmm.. but then why the "use fatal".  I seem to be missing something?

2colours: Because otherwise max.Real as a Failure will simply be passed on, and a monstrous Range (that starts/ends with a Failure) will be created.

lizmat: Aha, ok, then maybe SET-SELF would need an `$max.throw if nqp::istype($max,Failure)` ?

2colours: Probably it should check for both $min and $max; I'm not sure of the performance implications. Right now, the other cases only throw because they tried to retrieve the failed value before doing the function call.

2colours:
@lizmat well, it would be good to get this over with. I didn't want to add the check to !SET-SELF because that's common path and I felt that could be heavier on the performance. If I have this wrong, or it's insignificant here, sure I can just change to that. Either is better than leaving it as it is.

lizmat:
Question: did you actually spectest these changes?  I see pretty massive test failures with `make spectest`.

Turns out spectests say that `"a"..*`  and `*..'b'` are considered valid ranges.

Therefore closing this PR.

2colours:
> Turns out spectests say that `"a"..*` and `*..'b'` are considered valid ranges.

That's quite bad given that they literally create ranges that are forbidden otherwise. I'm gonna open an issue for it.

Issue #5180, opened 2023-02-03
[JVM] NullPointerException when handling exception from unsupported revision


usev6:
## The Problem
Nowadays the error reporting from trying to use an unsupported revision leads to a NullPointerException on the JVM backend.
The following example is taken from one of Rakudo's tests -- which now fails on the JVM backend: https://github.com/rakudo/rakudo/blob/c88f50d5e8/t/02-rakudo/14-revisions.t#L32-L36.
```
$ ./rakudo-j -e 'use v6.c.NOMOD'
Unhandled exception: java.lang.NullPointerException
  in DYNAMIC (gen/jvm/CORE.c.setting:849)
  in print_exception (gen/jvm/CORE.c.setting:53658)
  in <anon> (gen/jvm/CORE.c.setting:53758)
  in command_eval (gen/jvm/stage2/NQPHLL.nqp:1394)
  in command_eval (gen/jvm/Compiler.nqp:236)
  in command_line (gen/jvm/stage2/NQPHLL.nqp:1315)
  in MAIN (gen/jvm/rakudo.nqp:151)
  in <mainline> (gen/jvm/rakudo.nqp:130)
  in <anon> (gen/jvm/rakudo.nqp)
```

This used to work in the past, but I didn't try to figure out when it broke.

Interestingly, there is no internal error when using ```--ll-exception```:
```
$ ./rakudo-j --ll-exception -e 'use v6.c.NOMOD'
No compiler available for Raku v6.c.NOMOD
  in throw (gen/jvm/CORE.c.setting:53196)
  in throw (gen/jvm/World.nqp:5483)
  in typed_panic (gen/jvm/Grammar.nqp:296)
  in !check-version-modifier (gen/jvm/World.nqp:557)
  in load-lang-ver (gen/jvm/World.nqp:596)
  in lang-version (gen/jvm/Grammar.nqp:872)
  in comp_unit (gen/jvm/Grammar.nqp:838)
  in TOP (gen/jvm/Grammar.nqp:557)
  in parse (gen/jvm/stage2/QRegex.nqp:2270)
  in parse (gen/jvm/stage2/NQPHLL.nqp:1579)
  in execute_stage (gen/jvm/stage2/NQPHLL.nqp:1491)
  in run (gen/jvm/stage2/NQPHLL.nqp:1527)
  in <anon> (gen/jvm/stage2/NQPHLL.nqp:1530)
  in compile (gen/jvm/stage2/NQPHLL.nqp:1522)
[...]

PR #5379, opened 2023-09-11
[JVM] Add extra check for GLOBAL being null

usev6:
This fixes a NullPointerException on the JVM backend for code like

  use v6.c.UNKNOWN

That code does not blow up on MoarVM. But still, it looks plausible to check for null before calling a method and relying on the backend to handle that gracefully.

### DIFF
--- a/src/core.c/stubs.rakumod
+++ b/src/core.c/stubs.rakumod
@@ -49,12 +49,13 @@ sub DYNAMIC(str $name, @deprecation?) is raw {  # is implementation-detail
           value,
           nqp::stmts(
             (my str $pkgname = nqp::replace($name,1,1,'')),
-            nqp::ifnull(
-              nqp::atkey(GLOBAL.WHO,$pkgname),
+            nqp::if(
+              nqp::isnull(GLOBAL) || nqp::isnull(my \res := nqp::atkey(GLOBAL.WHO,$pkgname)),
               nqp::ifnull(
                 nqp::atkey(PROCESS.WHO,$pkgname),
                 Rakudo::Internals.INITIALIZE-DYNAMIC($name, @deprecation)
-              )
+              ),
+              res
             )
           )
         )


ab5tract:
I think this should be merged already and only hasn't been so due to oversight.

lizmat:
I'm not sure: does the JVM backend return `nqp::null` for non-existing keys with `nqp::atkey(GLOBAL.WHO)`?  If it does not, then *that* would need to be fixed first, as there are many places in NQP and the bootstrap where this semantic is assumed.

As to the nullness of `GLOBAL` itself: doesn't that point to a deeper issue on the JVM backend?

usev6:
> does the JVM backend return nqp::null for non-existing keys with nqp::atkey(GLOBAL.WHO)?

No, AFAICS that is not the case (same output as on MoarVM):
```
$ ./rakudo-j -e 'use nqp; say nqp::isnull(nqp::atkey(GLOBAL.WHO, "foo"))'
1
```

> As to the nullness of GLOBAL itself: doesn't that point to a deeper issue on the JVM backend?

I didn't check again, but last year I noticed that `GLOBAL` is `null` in that place on MoarVM as well: https://github.com/rakudo/rakudo/issues/5180#issuecomment-1712906558
My conclusion was, that the only difference is, that MoarVM doesn't blow up when using `nqp::atkey` on `nqp::null`, while the JVM does.

usev6:
>> As to the nullness of GLOBAL itself: doesn't that point to a deeper issue on the JVM backend?

> I didn't check again, but last year I noticed that GLOBAL is null in that place on MoarVM as well: https://github.com/rakudo/rakudo/issues/5180#issuecomment-1712906558
My conclusion was, that the only difference is, that MoarVM doesn't blow up when using nqp::atkey on nqp::null, while the JVM does.

I took another look and compiled rakudo-m with the following patch:
```
diff --git a/src/core.c/stubs.rakumod b/src/core.c/stubs.rakumod
index 8d44898da1..05375a5116 100644
--- a/src/core.c/stubs.rakumod
+++ b/src/core.c/stubs.rakumod
@@ -48,11 +48,15 @@ sub DYNAMIC(str $name, @deprecation?) is raw {  # is implementation-detail
           value,
           nqp::stmts(
             (my str $pkgname = nqp::replace($name,1,1,'')),
-            nqp::ifnull(
-              nqp::atkey(GLOBAL.WHO,$pkgname),
+            nqp::if(
+              nqp::isnull(GLOBAL),
+              nqp::die('Oops, GLOBAL is null in DYNAMIC'),
               nqp::ifnull(
-                nqp::atkey(PROCESS.WHO,$pkgname),
-                Rakudo::Internals.INITIALIZE-DYNAMIC($name, @deprecation)
+                nqp::atkey(GLOBAL.WHO,$pkgname),
+                nqp::ifnull(
+                  nqp::atkey(PROCESS.WHO,$pkgname),
+                  Rakudo::Internals.INITIALIZE-DYNAMIC($name, @deprecation)
+                )
               )
             )
           )
```

Running the code from the original report https://github.com/rakudo/rakudo/issues/5180 explodes now as well:
```
$ ./rakudo-m -e 'use v6.c.NOMOD'
Unhandled exception: Oops, GLOBAL is null in DYNAMIC
   at SETTING::src/core.c/stubs.rakumod:38  (/path/to/rakudo.moar/blib/CORE.c.setting.moarvm:DYNAMIC)
 from SETTING::src/core.c/Exception.rakumod:571  (/path/to/rakudo.moar/blib/CORE.c.setting.moarvm:print_exception)
 from SETTING::src/core.c/Exception.rakumod:668  (/path/to/rakudo.moar/blib/CORE.c.setting.moarvm:<anon>)
 from NQP::src/HLL/Compiler.nqp:348  (/home/christian/bin/rakudo.moar/share/nqp/lib/NQPHLL.moarvm:command_eval)
 from src/Perl6/Compiler.nqp:234  (/path/to/rakudo.moar/blib/Perl6/Compiler.moarvm:command_eval)
 from NQP::src/HLL/Compiler.nqp:268  (/home/christian/bin/rakudo.moar/share/nqp/lib/NQPHLL.moarvm:command_line)
 from src/main.nqp:81  (/path/to/rakudo.moar/rakudo.moarvm:MAIN)
 from src/main.nqp:63  (/path/to/rakudo.moar/rakudo.moarvm:<mainline>)
 from <unknown>:1  (/path/to/rakudo.moar/rakudo.moarvm:<main>)
 from <unknown>:1  (/path/to/rakudo.moar/rakudo.moarvm:<entry>)
```

So, I still think that the change from this PR makes sense.

usev6:
I've opened https://github.com/rakudo/rakudo/pull/5710 -- which adds a special case for the JVM backend -- as an alternative,

usev6:
May I ask for another review in the light of my above comments?

I'd prefer this patch over https://github.com/rakudo/rakudo/pull/5710, because it doesn't introduce yet another special case. But I have no idea if this is a particular hot code path. In that case it could make total sense to avoid the extra work on MoarVM.

usev6:
Ok, I give up on this one and will go for the special case with #5710.

PR #5384, opened 2023-09-16
RakuAST: Re-formulate VarDeclaration::Implicit::State

ab5tract:
The class needed some re-work to function properly for other use cases like in RakuAST::StatementPrefix::Once.


lizmat:
Thanks!

PR #5395, opened 2023-09-27
Produce fewer of LTA messages, especially when coercing

vrurg:
This is a follow up to [this IRC discussion](https://irclogs.raku.org/raku/2023-09-25.html#18:06-0001). With this PR the following snippet

```raku
role R {
    my Int $v = "aa";
}
my R() $v = 1;
```

would now produce an output like this:

```
Coercion from 'Int' into 'R' died with X::Role::Instantiation while trying to pun the target role; exception details:

    Could not instantiate role 'R' because it is died with X::TypeCheck::Assignment; exception details:

        Type check failed in assignment to $v; expected Int but got Str ("aa")
          in sub  at bad-coercion.raku line 11


  in any  at /Users/vrurg/src/Raku/rakudo/install/share/perl6/lib/Perl6/BOOTSTRAP/v6c.moarvm line 1
```

The output demonstrates two improvements at once: more detailed report of failed role instatiation and the actual message from coercion failing on a problematic role.

### DIFF
--- a/src/core.c/Backtrace.rakumod
+++ b/src/core.c/Backtrace.rakumod
@@ -65,7 +65,10 @@ my class Backtrace::Frame {
           || $!file ~~ / "CORE." \w+ ".setting" $ /
 #?endif
 #?if !jvm
-          || $!file ~~ / "CORE." \w+ ".setting.{ Rakudo::Internals.PRECOMP-EXT }" $ /
+          || $!file ~~ / [ "CORE." \w+ ".setting"
+                           | "NQPCORE.setting"
+                           | "BOOTSTRAP/v6" \w ]
+                         ".{ Rakudo::Internals.PRECOMP-EXT }" $ /
 #?endif
           || $!file.ends-with(".nqp")
     }
@@ -171,9 +174,11 @@ my class Backtrace {
             next if $file.ends-with('BOOTSTRAP.nqp')
                  || $file.ends-with('QRegex.nqp')
                  || $file.ends-with('Perl6/Ops.nqp');
-            if $file.ends-with('NQPHLL.nqp')
+
+            my $name := nqp::p6box_s(nqp::getcodename($do));
+
+            if ($file.starts-with('NQP::') && $file.ends-with('Compiler.nqp') && $name eq 'eval')
                 || $file.ends-with('NQPHLL.moarvm')
-                || $file.starts-with('NQP::')
             {
                 # This could mean we're at the end of the interesting backtrace,
                 # or it could mean that we're in something like sprintf (which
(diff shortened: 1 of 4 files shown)


vrurg:
A note on the IRC discussion. After carefully considering outcomes of disabling role punning for roles as coercion target types I concluded that it'd be a terrible thing to do as some roles might be intentionally designed to be implicitly punned when necessary. Thus, eventually, I came up with this fix for LTA messages.

The only problem I don't know how to fix at the moment is the `in any  at /Users/vrurg/src/Raku/rakudo/install/share/perl6/lib/Perl6/BOOTSTRAP/v6c.moarvm line 1` line of backtrace. It is a result of throwing the exception from dispatcher. I just don't see what it has to be.

lizmat:
I guess "is hidden-from-backtrace" is not an option?

vrurg:
> I guess "is hidden-from-backtrace" is not an option?

dispatchers.nqp, guess it is not. :)

vrurg:
BTW, it's even worse than that because the original line from the source is not even reported. With the sample from the first comment it has to be line 4, in my local test script – line 13 in bad-coercion.raku.

src/core.c/Backtrace.rakumod
```diff
@@ -171,9 +174,11 @@ my class Backtrace {
             next if $file.ends-with('BOOTSTRAP.nqp')
                  || $file.ends-with('QRegex.nqp')
                  || $file.ends-with('Perl6/Ops.nqp');
-            if $file.ends-with('NQPHLL.nqp')
+
+            my $name := nqp::p6box_s(nqp::getcodename($do));
+
+            if ($file.starts-with('NQP::') && $file.ends-with('Compiler.nqp') && $name eq 'eval')
```
ugexe, line 180: Should this have always been been `$file.starts-with('NQP/')`? It seems like a file containing `::` might not be intentional

ugexe: I guess there are other examples like `SETTING::`, so I suppose it must be correct

vrurg: You're correct. `SETTING::` and `NQP::` are special prefixed introduced by composing script via `#line` directive. Very helpful when categorizing files.

vrurg:
Now, with `Backtrace` (hopefully!) fixed, the snippet from the first comment results in way better output:

```
Coercion from 'Int' into 'R' died with X::Role::Instantiation while trying to pun the target role; exception details:

    Could not instantiate role 'R' because it is died with X::TypeCheck::Assignment; exception details:

        Type check failed in assignment to $v; expected Int but got Str ("aa")
          in sub  at bad-coercion.raku line 2
          in block <unit> at bad-coercion.raku line 4

      in block <unit> at bad-coercion.raku line 4

  in block <unit> at bad-coercion.raku line 4
```

PR #5397, opened 2023-10-01
 RakuAST: Migrate to WhateverApplicable

ab5tract:
(tl;dr -- +4 spectests!)

This allows a top-down approach to searching for other
`WhateverApplicable` nodes, building up a list of applicable child
nodes to curry across. This simplifies the actual implementation
of the "currying question" inside of the `Apply*` code significantly
while also covering the creation of deeply nested (or otherwise weird)
`WhateverCode` objects.

This isn't to imply that this implementation is "simple"... anything
as complex as the `*` in Raku is going to be a bit of a mind warp to
comprehend, let alone implement.

### DIFF
--- a/src/Raku/ast/base.rakumod
+++ b/src/Raku/ast/base.rakumod
@@ -174,6 +174,43 @@ class RakuAST::Node {
         self.IMPL-WRAP-LIST(@result)
     }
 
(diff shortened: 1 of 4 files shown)


lizmat:
Cool!

PR #5401, opened 2023-10-06
:sparkles: translation to Italian

JJ:
I'm opening this as draft, since there are a few things that are not entirely clear to me. It says that the identifiers must follow Raku identifier rules, but as a matter of fact most identifiers don't use tildes or circumflexes. Would that be OK here? Also, hyphens seem to be OK, but I haven't seen any in the existing translations, so I'll check and see if they are OK, and don't break any tests or anything like that (are there tests?)

### DIFF
--- a/tools/templates/L10N/IT
+++ b/tools/templates/L10N/IT
@@ -5,21 +5,21 @@
 
 # KEY           TRANSLATION
 block-default   predefinito
-#block-else      else
-#block-elsif     elsif
-#block-for       for
-#block-given     given
-#block-if        if
-#block-loop      loop
-#block-orwith    orwith
-#block-repeat    repeat
-#block-unless    unless
-#block-until     until
-#block-when      when
-#block-whenever  whenever
-#block-while     while
-#block-with      with
-#block-without   without
+block-else      altro
+block-elsif     altro-se
+block-for       per
+block-given     dato
+block-if        se
+block-loop      ciclo
+block-orwith    o-con
+block-repeat    ripeti
+block-unless    salvo-che
+block-until     finché
+block-when      quando
+block-whenever  ogni-volta-che
+block-while     mentre
+block-with      con
+block-without   senza
 
 # KEY             TRANSLATION
 #constraint-where  where
--- a/tools/templates/L10N/IT
+++ b/tools/templates/L10N/IT
@@ -66,29 +66,29 @@ core-comb          pettine
 core-combinations  combinazioni
 
 # KEY         TRANSLATION
-#core-deepmap  deepmap
-#core-defined  defined
-#core-die      die
+core-deepmap  mappa-profonda
+core-defined  definito
+core-die      muori
 #core-dir      dir
-#core-done     done
-#core-duckmap  duckmap
-#
+core-done     fatto
+core-duckmap  mappa-anatra
+
 # KEY        TRANSLATION
 #core-elems   elems
-#core-emit    emit
-#core-end     end
-#core-exit    exit
+core-emit    emettere
+core-end     fine
+core-exit    uscire
 #core-exp     exp
 #core-expmod  expmod
 
 # KEY              TRANSLATION
-#core-fail          fail
+core-fail          fallire
 #core-fc            fc
-#core-first         first
-#core-flat          flat
+core-first         primo
+core-flat          piatto
 #core-flip          flip
-#core-floor         floor
-#core-full-barrier  full-barrier
+core-floor         pavimento
+core-full-barrier  barriera-completa
 
 # KEY      TRANSLATION
 #core-get   get
--- a/tools/templates/L10N/IT
+++ b/tools/templates/L10N/IT
@@ -22,13 +22,13 @@ block-with      con
 block-without   senza
 
 # KEY             TRANSLATION
-#constraint-where  where
+constraint-where  dove
 
 # KEY                  TRANSLATION
 #core-abs               abs
-#core-all               all
-#core-any               any
-#core-append            append
+core-all               tutti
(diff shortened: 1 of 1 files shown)


tools/templates/L10N/IT
```diff
@@ -0,0 +1,411 @@
...
+#stmt-prefix-quietly   quietly
+#stmt-prefix-race      race
+#stmt-prefix-react     react
+#stmt-prefix-sink      sink
+#stmt-prefix-start     start
+#stmt-prefix-supply    supply
+#stmt-prefix-try       try
+
+# KEY      TRANSLATION
+#term-self  self
```
JJ, line 386: this would be `se-estesso`, for instance. `self` would mostly be OK in Italian, though.

tools/templates/L10N/IT
```diff
@@ -0,0 +1,411 @@
...
+#term-nano  nano
+term-now   gia
+#term-rand  rand
+term-time  tempo
+
+# KEY          TRANSLATION
+trait-does     fa
+trait-handles  gestisce
+trait-hides    nasconde
+trait-is       e
```
JJ, line 396: This is, of course, ` è`. Should I go ahead and use it?

tools/templates/L10N/IT
```diff
@@ -0,0 +1,411 @@
...
+trait-of       da
```
JJ, line 397: `of` can be translated in Italian as `da` (origin, procedence) or `di` (possessive). I'm going for origin here.

tools/templates/L10N/IT
```diff
@@ -0,0 +1,411 @@
...
+trait-returns  ritorna
+
+# KEY         TRANSLATION
+#typer-enum    enum
+#typer-subset  subset
+
+# KEY        TRANSLATION
+use-import   importare
```
JJ, line 405: It looks like the Dutch translation is going for infinitive here, so I'm following that.

JJ (commented):
A few online comments here, waiting for clarification. Maybe adding some comments to `CORE` would solve whatever ambiguity there is.
At any rate, translation is an arcane art. Not totally sure why you're needing it here, but as usual, a professional would be the way to go, I think.

tools/templates/L10N/IT
```diff
@@ -0,0 +1,411 @@
+# This file contains the Italian translation of features of the
+# Raku Programming Language.
+#
+# See tools/build/L10N/CORE for more explanation.
+
+# KEY           TRANSLATION
+block-default   predefinito
```
JJ, line 7: Italian, same as Spanish, has gendered adjectives. Going for masculine, here. But it's a judgement call.

JJ:
What's the deal with tests? Should I pay attention to them?

tools/templates/L10N/IT
```diff
@@ -0,0 +1,411 @@
...
+#infix-leg         leg
+#infix-lt          lt
+#infix-max         max
+#infix-min         min
+#infix-minmax      minmax
+#infix-mod         mod
+#infix-ne          ne
+#infix-notandthen  notandthen
+#infix-o           o
+#infix-or          or
```
JJ, line 280: There's a collision here, `or` is `o` in Italian. `od` if it's followed by a bowel, sometimes seen as "oppure" meaning "or simply" Any suggestion about what to do here? By default, just leave it as is.

JJ:
This is done now. There are several words that are repeated, in different blocks. Maybe just writing them one would relieve the burden a bit. When this is accepted I'll go for the Spanish one.

tools/templates/L10N/IT
lizmat: If the original term is ok, then maybe remove the `#` to indicate that it was considered for translation and kept (as opposed to with the `#` to indicate no suitable translation was decided yet).

tools/templates/L10N/IT
lizmat: Yeah, this is a general issue.  That's why I made the "categories" because the same word in english, could have different translations in other languages.

Such as `if` as a block initiator, and `if` as a modifier.

tools/templates/L10N/IT
lizmat: It's referring to:
````raku
default {
    say "default action in 'when' sequence";
}
````

File: tools/templates/L10N/IT (hunk 6 above)
Comment (lizmat) [reply tools/templates/L10N/IT:280]: `oppure` in that case feels more correct to me.  And then maybe translate `o` with some Italian equivalent of "function composition operator" ?

**Comment by lizmat** [comment]
Several words are repeated in different blocks because they *could* have different translations in some languages.

File: tools/templates/L10N/IT (hunk 5 above)
Comment (JJ) [reply tools/templates/L10N/IT:7]: OK, then it's action, "azione", that's female. Changing it.

File: tools/templates/L10N/IT (hunk 2 above)
Comment (lizmat) [reply tools/templates/L10N/IT:396]: Sure, as long as it's a valid identifier, and `è` is  :-)

**Comment by lizmat** [comment]
Ready to merge now?

**Comment by JJ** [comment]
> Ready to merge now?

I've addressed all comments, so if you think it is, please go ahead.

**Comment by lizmat** [comment]
Thanks!  After running the `tools/build/makeL10N.raku` script:
```
$ raku -e 'use RakuAST::Deparse::L10N::IT; say Q|if 42 { say "ciao" }|.AST.DEPARSE("IT")'
se 42 {
    dillo("ciao")
}
```
```
$ raku -e 'say Q|use L10N::IT; se 42 { dillo "ciao" }|.AST.DEPARSE'
if 42 {
    say("ciao")
}
```

JJ:
That's pretty cool.

Issue #5400, opened 2023-10-05
RakuAST: `FIRST` phaser does not appear to be working


ab5tract:
## The Problem

`FIRST` phaser does not appear to be working properly.

## Expected Behavior

```
my $t = True;
while $t {
    FIRST { $t = False };
    say 1;
}
```
Output:
```
1
```

## Actual Behavior

Output:
```
1
1
1
1
1
# .. Inf
```

## Steps to Reproduce
```
RAKUDO_RAKUAST=1 ./rakudo-m --ll-exception -e 'my $t = True; while $t { FIRST { $t = False }; say 1 }'
```

## Environment
* Operating system: WSL2 for Windows (Ubuntu)
* Compiler version (`rakudo -v` or `raku -v`): 2023.09 (blead)

ab5tract:
This is actually fixed by #5406 , #5404 was a misfire.

PR #5404, opened 2023-10-07
RakuAST: Fix FIRST phaser not firing

ab5tract:
This is currently only working with loops, which matches what the base compiler does.

It also recommends (well, demands, really) the more-sensible-for-the-use-case `once` statement prefix. This should be done via a worry but currently that mechanism is not working from this code path.


ab5tract:
Fixes #5400 .

ab5tract:
Seems my fix breaks the working behavior of phasers in for loops (thanks to `t/spec/integration/advent2012-day15.t` for catching this.)

Marking as a draft since it moves backwards on spectests.

PR #5405, opened 2023-10-07
RakuAST: Fix broken test

ab5tract:
I'm using 'can' here because I don't know how many resolvers there may eventually be. The real question is: why doesn't self.add-worry work? (The sorry never sorries).

Until then, just make sure the $resolver actually can do the thing.

### DIFF
--- a/src/Raku/ast/code.rakumod
+++ b/src/Raku/ast/code.rakumod
@@ -797,10 +797,14 @@ class RakuAST::ScopePhaser {
 
     method IMPL-STUB-PHASERS(RakuAST::Resolver $resolver, RakuAST::IMPL::Context $context) {
         if $!FIRST && ! $!is-loop-body {
-            $resolver.add-worry:
-                $resolver.build-exception: 'X::AdHoc',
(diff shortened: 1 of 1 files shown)


raiph (commented):
> why doesn't `self.add-worry` work? (The sorry never sorries).

I've no idea, but am distracted / confused by the switch from mentioning "worry" (which I think of as a *run* phase *warning*) to "sorries" (which I think of as a *compile* phase *error*). I imagine the code behind them may be 90% the same, so I may be missing the point, but thought I'd write this comment anyway.

src/Raku/ast/code.rakumod
```diff
@@ -1133,7 +1137,7 @@ class RakuAST::Block
                                     :scope<local>,
                                     :decl<var>)))),
                     $block,
-                    QAST::Op.new(:op<call>, $var).annotate_self('loop-already-block-first-phaser', $block));
+                    QAST::Op.new(:op<call>, $var));
```
lizmat, line 1140: Why is the annotation lost here?

lizmat:
Moot

PR #5406, opened 2023-10-08
RakuAST: Implement `FIRST` using AST

ab5tract:
The primary commit message is as follows:

```
RakuAST: Implement FIRST phaser as an AST construct

It turns out that there is no need to even dip into QAST in order
to implement the `FIRST` phaser, which was particularly tricky
because it depended on so-called "extops" that are particularly
hard to arrange in pure QAST (`p6setfirstflag` and `p6takefirstflag`,
for the curious).

It turns out (nine++) that we can simply arrange for this construct
at an AST level, by emulating something like:

{
  state $triggered;
  unless $triggered {
    # ... the actual content of the FIRST phaser block
    $triggered = True
  }
}()

This feels like a much more natural implementation and it really
shows off what a difference RakuAST can already make to code generation.
```

Passes one additional spectest:

```
t/spec/S04-phasers/first.t ........................................ ok
```

### DIFF
--- a/src/Raku/ast/expressions.rakumod
+++ b/src/Raku/ast/expressions.rakumod
@@ -3,7 +3,10 @@ class RakuAST::CaptureSource
   is RakuAST::Node { }
 
 # Everything that can appear as an expression does RakuAST::Expression.
+# Furthermore, since every expression can exist as statement, consider
+# us a RakuAST::Blorst since we may easily find ourselves where blorsts go.
 class RakuAST::Expression
+  is RakuAST::Blorst
   is RakuAST::IMPL::ImmediateBlockUser
 {
     # All expressions can be thunked - that is, compiled such that they get
--- a/src/Raku/ast/expressions.rakumod
+++ b/src/Raku/ast/expressions.rakumod
@@ -403,6 +403,10 @@ class RakuAST::Infix
         }
         $op(|@operands)
     }
+
+    method dump-markers() {
+        '【' ~ $!operator ~ '】'
+    }
 }
 
 class RakuAST::Feed
--- a/src/Raku/ast/statementprefixes.rakumod
+++ b/src/Raku/ast/statementprefixes.rakumod
@@ -617,8 +617,56 @@ class RakuAST::StatementPrefix::Phaser::Block
 # The FIRST phaser.
 class RakuAST::StatementPrefix::Phaser::First
   is RakuAST::StatementPrefix::Phaser::Block
+  is RakuAST::BeginTime
 {
     method type() { "FIRST" }
+
+    method is-begin-performed-before-children { True }
+    method is-begin-performed-after-children  { False }
+
+    method PERFORM-BEGIN(Resolver $resolver, RakuAST::IMPL::QASTContext $context) {
+        my $blorst := nqp::getattr(self, RakuAST::StatementPrefix, '$!blorst');
+
+        my $True := RakuAST::Term::Name.new(RakuAST::Name.from-identifier('True'));
+
+        my $attach-block := $resolver.find-attach-target('block');
+        my $trigger-name := QAST::Node.unique('!first_block_triggered');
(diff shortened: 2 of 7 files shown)


src/Raku/ast/expressions.rakumod
```diff
@@ -3,7 +3,10 @@ class RakuAST::CaptureSource
   is RakuAST::Node { }
 
 # Everything that can appear as an expression does RakuAST::Expression.
+# Furthermore, since every expression can exist as statement, consider
```
niner, line 6: That doesn't sound right. While every expression can exist as a statement, those are represented in the AST as RakuAST::Statement::Expression, i.e. the RakuAST::Expression is wrapped in a statement. I think that's an important distinction that we want to retain.

src/Raku/ast/statements.rakumod
```diff
@@ -1,7 +1,25 @@
 # Block or statement, used in statement prefixes which can take either a
-# block or a statement.
+# block or a statement (or an expression, which can always go in a
+# RakuAST::Statement::Expression)
 class RakuAST::Blorst
-  is RakuAST::Node { }
+  is RakuAST::Node {
+    method as-block() {
+        nqp::istype(self, RakuAST::Block)
```
niner, line 7: Instead of doing a type check here, RakuAST::Block could just override the method with an identity version.

src/Raku/ast/statementprefixes.rakumod
```diff
@@ -617,8 +617,56 @@ class RakuAST::StatementPrefix::Phaser::Block
...
+
+    method is-begin-performed-before-children { True }
+    method is-begin-performed-after-children  { False }
+
+    method PERFORM-BEGIN(Resolver $resolver, RakuAST::IMPL::QASTContext $context) {
+        my $blorst := nqp::getattr(self, RakuAST::StatementPrefix, '$!blorst');
+
+        my $True := RakuAST::Term::Name.new(RakuAST::Name.from-identifier('True'));
+
+        my $attach-block := $resolver.find-attach-target('block');
```
niner, line 634: This is wrong. The only place where it's legal to ask for an attach target is the attach method. You don't know when PERFORM-BEGIN is called and what state the resolver will be in at that time. It could be that parsing some descendant is what triggers BEGIN handling and then your attach target may even be one of your children.

src/Raku/ast/statementprefixes.rakumod
```diff
@@ -617,8 +617,56 @@ class RakuAST::StatementPrefix::Phaser::Block
...
 {
     method type() { "FIRST" }
+
+    method is-begin-performed-before-children { True }
+    method is-begin-performed-after-children  { False }
+
+    method PERFORM-BEGIN(Resolver $resolver, RakuAST::IMPL::QASTContext $context) {
+        my $blorst := nqp::getattr(self, RakuAST::StatementPrefix, '$!blorst');
+
+        my $True := RakuAST::Term::Name.new(RakuAST::Name.from-identifier('True'));
```
niner, line 632: This should be an implicit lookup. Alternatively you could just put True in a RakuAST::Declaration::External::Constant. True and False are available in the bootstrap.

src/Raku/ast/expressions.rakumod
ab5tract: Hmm, then we need to be more careful about what ends up in a (supposed-to-be) `RakuAST::Blorst` attribute. But you are probably right, instead the issue is that something is passing a `RakuAST::Expression` instead of a `RakuAST::Statement::Expression`.

If you don't mind, I'll have a look at this in a separate PR since I don't think this alone should result in a revert.

src/Raku/ast/statements.rakumod
ab5tract: Good point!

src/Raku/ast/expressions.rakumod
ab5tract: Looks like the issue was with the original test, actually. I'll fix this up.

PR #5408, opened 2023-10-10
RakuAST: Undo expressions as blorsts thinko

ab5tract:
Rather than waiting to submit this as part of a larger PR, here's the fix for that blorst thinko.


ab5tract:
Opened this too soon, figured out the problem in the larger branch right afterwards.

PR #5411, opened 2023-10-12
RakuAST: Fix binding of $_ on smartmatch RHS

### LINKED ISSUE #5239: `value ~~ $_` does *not* bind the value to $_ on the RHS
## The Problem

Smartmatching is known to behave the following way:
1. the left handside is evaluated
2. its return value is temporarily bound to $_ (for the rest of the operation)
3. the right handside is evaluated
4. on the return value, `.ACCEPTS($_)` is called

Step 2 doesn't seem to happen if the right handside is the topic variable in itself, hence being inconsistent with any other smartmatching that involves `$_` on the right handside as a part of an expression, offering no theoretical foundation for this inconsistency.

## Expected Behavior
A `value ~~ $_`  type of expression always acts as `value ~~ value`, due to the localization of `$_` on the right handside.

## Actual Behavior
`value ~~ $_`  will be executed with the previous value of `$_`

## Steps to Reproduce
```raku
$_ = 12;
say 14 ~~ $_; # False
```

## Environment
* Operating system: `5.14.0-1058-oem #66-Ubuntu SMP Fri Feb 10 09:46:18 UTC 2023 x86_64 x86_64 x86_64 GNU/Linux`
* Compiler version (`perl6 -v` or `raku -v`): Rakudo v2022.12 on MoarVM.

ab5tract:
This fixes #5239 without any changes to spectests (might be good to add a test, or unfudge one if one already existed, as this behavior does not work in the base compiler).

Note: It could also be fixed without breaking tests by removing the guard against `RakuAST::Var` entirely, however I'm operating under the assumption that there are variable related optimizations in the dedicated `infix:<~~>` implementation.

### DIFF
--- a/src/Raku/ast/expressions.rakumod
+++ b/src/Raku/ast/expressions.rakumod
@@ -279,7 +279,9 @@ class RakuAST::Infix
                 nqp::die('Cannot compile bind to ' ~ $left.HOW.name($left));
             }
         }
-        elsif nqp::existskey(OP-SMARTMATCH, $op) && !nqp::istype($right, RakuAST::Var) {
+        elsif nqp::existskey(OP-SMARTMATCH, $op)
+            && (!nqp::istype($right, RakuAST::Var) || (nqp::istype($right, RakuAST::Var::Lexical) && $right.is-topic))
+        {
             self.IMPL-SMARTMATCH-QAST($context, $left, $right, nqp::atkey(OP-SMARTMATCH, $op));
         }
         else {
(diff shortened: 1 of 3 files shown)


niner (approved):
The commit message could use a bit more context. What was broken before? Why can we assume that it's fixed with this change?

ab5tract:
I've just been double-checking this and it looks like this makes

```
    t/12-rakuast/regex-charclass.rakutest
```

fail. Untangling that now, but it looks like this breaks something like `$foo ~~ $_ for @regex-objects` which can't be simply converted into `$_ ~~ $foo for @regex-objects` because it looks like match objects are stringified when on the LHS.

PR #5418, opened 2023-10-14
RakuAST: Improve WhateverCode's .raku

ab5tract:
This change allows `(* * *).raku` to return `* * *`, providing round-tripping through `raku` for WhateverCodes.

### DIFF
--- a/src/Raku/ast/code.rakumod
+++ b/src/Raku/ast/code.rakumod
@@ -2669,10 +2669,12 @@ class RakuAST::CurryThunk
   is RakuAST::ImplicitLookups
 {
     has Mu $!parameters;
+    has Str $!original-expression;
 
-    method new() {
+    method new(Str $original-expression) {
         my $obj := nqp::create(self);
-        nqp::bindattr($obj, RakuAST::CurryThunk,, '$!parameters', []);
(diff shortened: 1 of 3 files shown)


src/Raku/ast/code.rakumod
```diff
@@ -2669,10 +2669,12 @@ class RakuAST::CurryThunk
   is RakuAST::ImplicitLookups
 {
     has Mu $!parameters;
+    has Str $!original-expression;
 
-    method new() {
+    method new(Str $original-expression!) {
```
lizmat, line 2674: ! is not needed here as positionals are mandatory by default

src/Raku/ast/code.rakumod
```diff
@@ -2684,6 +2686,10 @@ class RakuAST::CurryThunk
         '⋐' ~ nqp::x('🔆', self.IMPL-NUM-PARAMS)  ~ '⋑'
     }
 
+    method set-original-expression(Str $original-expression) {
```
lizmat, line 2689: If the "original-expression" is mandatory with `.new`, we don't need this method, do we?  Or do we *change* the original expression at some point?

ab5tract: You're correct, we actually do not need this setter.

PR #5421, opened 2023-10-15
(Re-)align backtraces on JVM with MoarVM

usev6:
With 04a3127569 backtraces on the JVM backend started to include references to NQP's internals. This commit intends to apply the changes from 04a3127569 to the JVM backend, too.


usev6:
Since this should only affect the JVM backend, I plan to merge this in the next days. (Unless someone speaks up against it.)

niner (approved):

Issue #5358, opened 2023-08-23
Supply.throttle should allow `:vent-at(0)`.


amano-kenji:
## The Problem

I want a queue that drops everything while something is processed.

A minimal example is

```
#!/usr/bin/env raku

unlink "/run/user/1000/sock";

my $t = IO::Socket::Async.listen-path("/run/user/1000/sock").throttle: 1, -> $conn {
  sleep 5;
  react {
    whenever $conn.Supply.lines {
      say "\"", $_, "\"";
      done;
    }
  }
}, :vent-at(0), :bleed(Supplier.new);

$t.wait;
```

The documentation says `:bleed` accepts `Supply`, but it actually accepts `Supplier`.

ab5tract:
@AlexDaniel recently mentioned [a real-world example of this in Whateverable](https://github.com/Raku/whateverable/blob/f748c505471565c1e7bcef07514c45f3cbe1bb03/xbin/Buildable.p6#L95-L101)

lizmat:
@amano-kenji @AlexDaniel could you check this out the coming days so that it can make the 2023.10 release due this Thursday?

PR #5424, opened 2023-10-16
Allow a value of 0 with :vent-at

lizmat:
Hopefully fixing #5358

### DIFF
--- a/src/core.c/Supply-coercers.rakumod
+++ b/src/core.c/Supply-coercers.rakumod
@@ -1,4 +1,4 @@
-# continued from src/core.c/Supply-factories.pm6
+# continued from src/core.c/Supply-factories.rakumod
 
     ##
     ## Coercions
--- a/src/core.c/Supply-coercers.rakumod
+++ b/src/core.c/Supply-coercers.rakumod
@@ -1,4 +1,4 @@
-# continued from src/core.c/Supply-factories.rakumod
+# continued from src/core.c/Supply-factories.pm6
 
     ##
     ## Coercions
@@ -869,7 +869,7 @@
     ) {
         my $timer = Supply.interval($seconds,$delay,:$scheduler);
         my int $limit   = $elems;
-        my int $vent = $vent-at if $bleed;
+        my int $vent = $bleed ?? $vent-at !! -1;
         supply {
             my @buffer;
             my int $allowed = $limit;
@@ -903,7 +903,7 @@
                     $emitted = $emitted + 1;
                     $allowed = $allowed - 1;
                 }
-                elsif $vent && +@buffer >= $vent {
+                elsif $vent >= 0 && +@buffer >= $vent {
                     $bleed.emit(val);
                 }
                 else {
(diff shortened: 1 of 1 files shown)


lizmat:
@amano-kenji @AlexDaniel could you check this out the coming days so that it can make the 2023.10 release due this Thursday?

src/core.c/Supply-coercers.rakumod
```diff
@@ -1,4 +1,4 @@
-# continued from src/core.c/Supply-factories.rakumod
+# continued from src/core.c/Supply-factories.pm6
```
2colours, line 1: The extension got reverted here

amano-kenji:
I will test this next morning.

amano-kenji:
I tried to test it. I built it with
```
perl Configure.pl --gen-moar --gen-nqp --backend=moar
```
But, I can't find `rakudo` executable anywhere.

When I call
```
multi method throttle(Supply:D:
      Int()  $elems,
      Callable:D $process,
      Real() $delay = 0,
      :$scheduler   = $*SCHEDULER,
      :$control,
      :$status,
      :$bleed,
      :$vent-at,
    )
```
I want only one process to be executed at a time, and while the process is being executed, I want the supply to drop all incoming messages. Does `vent-at(0)` do that?

lizmat:
That's the idea.  If you also specify `:bleed` of course?

amano-kenji:
If I also specify `:bleed`, I want the supply to send everything to `bleed` which will get rid of excess.

amano-kenji:
If that works properly as expected, then I think it should be okay to merge it without us.

lizmat:
> If that works properly as expected, then I think it should be okay to merge it without us.

That's it: I didn't test it, as I don't really have an easy test case.

Anyway, it is too late for the 2023.10 release, which will be today.

amano-kenji:
How can I test this without installing rakudo?

ab5tract:
@amano-kenji You have to use this git repo checked out to the branch of this PR (`lizmat-vet-0`). First you run `perl Configure.pl --gen-moar --gen-nqp --backends=moar` and _then_ run `make`. This will create executables in the directory. No installation necessary.

amano-kenji:
I got `rakudo-m` executable which seemed to be what I wanted.

I executed this.

```
my $supplier = Supplier.new;

my $chan = $supplier.Supply.throttle(1, -> $a { sleep 5; $a; }, :bleed(Supplier.new), :vent-at(0)).Channel;

start {
  for [1..*] {
    sleep 0.1;
    $supplier.emit($_);
  }
}

react {
  whenever $chan {
    say $_.result;
  }
}
```

The output was

```
1
51
101
151
...
```

ab5tract:
FWIW I was unable to achieve the behavior that @amano-kenji reported when using this branch on macOS 14.

ab5tract:
I've rebased this against current `main` and still unfortunately do not get the expected results:

```
longwalker@mirabilis ~/c/r/z/rakudo ((6435e28b))
> ./rakudo-m ~/code/raku/scripts/ventatzeroraku/vent-at-zero.raku
1
49
97
145
194
242
291
```

PR #5425, opened 2023-10-16
Precompile in parallel

ugexe:
Previously when installing modules we would precompile them one at a time. This adds some basic parallelization to the precompilation loop so we precompile multiple files at the same time.

### DIFF
--- a/src/core.c/CompUnit/Repository/Installation.rakumod
+++ b/src/core.c/CompUnit/Repository/Installation.rakumod
@@ -337,7 +337,6 @@ sub MAIN(:$name, :$auth, :$ver, *@, *%) {
             my $repo-prefix = self!repo-prefix;
             my $*DISTRIBUTION = CompUnit::Repository::Distribution.new($dist, :repo(self), :$dist-id);
             my $*RESOURCES = Distribution::Resources.new(:repo(self), :$dist-id);
-            my %done;
 
             my $compiler-id = CompUnit::PrecompilationId.new-without-check($*RAKU.compiler.id);
             for %provides.sort {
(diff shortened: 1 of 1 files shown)


ugexe:
This brought the install time of `DateTime::TimeZone` (a module with a lot of provides) from 320 seconds to 170 seconds

src/core.c/CompUnit/Repository/Installation.rakumod
```diff
@@ -337,30 +337,26 @@ sub MAIN(:$name, :$auth, :$ver, *@, *%) {
...
             my $*RESOURCES = Distribution::Resources.new(:repo(self), :$dist-id);
-            my %done;
 
             my $compiler-id = CompUnit::PrecompilationId.new-without-check($*RAKU.compiler.id);
             for %provides.sort {
                 my $id = CompUnit::PrecompilationId.new-without-check($_.value.values[0]<file>);
                 $precomp.store.delete($compiler-id, $id);
             }
 
-            for %provides.sort {
```
ugexe, line 348: By doing this in parallel we lose the deterministic file creation order. However, I'm not sure that necessarily goes against the spirit of reproducibility or not.

src/core.c/CompUnit/Repository/Installation.rakumod
```diff
@@ -337,30 +337,26 @@ sub MAIN(:$name, :$auth, :$ver, *@, *%) {
...
-                my $id = $_.value.values[0]<file>;
+            my $to-precompile := %provides.map({ $_.key => $_.value.values[0]<file> }).sort(*.key).unique(:as(*.value));
```
ugexe, line 347: By using `unique` here we should avoid the need for `%done`. And since it is called before we use any parallelization we should always get the same outcome (such as when two different short-names have the same sha1).

src/core.c/CompUnit/Repository/Installation.rakumod
```diff
@@ -337,30 +337,26 @@ sub MAIN(:$name, :$auth, :$ver, *@, *%) {
             my $repo-prefix = self!repo-prefix;
             my $*DISTRIBUTION = CompUnit::Repository::Distribution.new($dist, :repo(self), :$dist-id);
             my $*RESOURCES = Distribution::Resources.new(:repo(self), :$dist-id);
-            my %done;
 
             my $compiler-id = CompUnit::PrecompilationId.new-without-check($*RAKU.compiler.id);
             for %provides.sort {
```
ugexe, line 342: I'm not sure there would be much benefit to parallelizing this loop, so I left it as is.

ugexe:
I think something was wrong with my original test, because I don't see any performance improvement now...

patrickbkr:
Do you have an idea why it's not faster? My intuition is that this is an obvious case for parallelization.

ugexe:
It could be that the module files of `DateTime::TimeZone` are small and the overhead of spawning processes to precompile eat any benefit. That could be tested by installing a distribution with many larger modules. It could also be related to the file locking that gets done during precompilation, although based on my understanding that shouldn't be an issue.

PR #5426, opened 2023-10-17
Sort input file list

bmwiedemann:
Sort input file list so that `Makefile` builds in a reproducible way
in spite of non-deterministic filesystem readdir order.

See https://reproducible-builds.org/ for why this is good.

This patch was done while working on reproducible builds for openSUSE.


bmwiedemann:
Wow. That was fast :+1:

lizmat:
The 2023.10 release is upon us in 2 days, so speed was needed  :-)

PR #5436, opened 2023-10-21
Remove undocumented quotelang extenders

lizmat:
The primary quotelang markers q, qq, and Q allow for a number of extenders according to the documentation: w, ww, and x.  So instead of qw, one could also say q:w, and instead of qqww one could also say qq:ww.

The grammar however **ALSO** allows for: to, s, a, h, f, c, b, o.  Allowing one to say qqto and Qf (to name but two of the possible combinations).

I think this is taking the TIMTOWTDO a little too far.  So this commit removes support for these undocumented quotelang extenders.

This breaks three spectest files:
- t/spec/S02-literals/misc-interpolation.t
- t/spec/S02-literals/quoting.t
- t/spec/integration/advent2014-day16.t

because they check for a, b and to as quotelang extenders respectively. If this change is OKd, then these tests will have to be adapted.  The adaptation is quite simple: add a colon.

### DIFF
--- a/src/Perl6/Grammar.nqp
+++ b/src/Perl6/Grammar.nqp
@@ -3472,13 +3472,13 @@ grammar Perl6::Grammar is HLL::Grammar does STD {
     token quote_mod:sym<w>  { <sym> }
     token quote_mod:sym<ww> { <sym> }
     token quote_mod:sym<x>  { <sym> }
-    token quote_mod:sym<to> { <sym> }
-    token quote_mod:sym<s>  { <sym> }
-    token quote_mod:sym<a>  { <sym> }
-    token quote_mod:sym<h>  { <sym> }
-    token quote_mod:sym<f>  { <sym> }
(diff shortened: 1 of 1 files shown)


raiph:
FWIW I like them. I won't elaborate on why just yet because others might like them too and explain why, or pretty much everyone else thinks it's fine for them to go, and if so, I wouldn't argue against this PR being merged.

lizmat:
Well, if people like them, they should be tested and documented  :-)

raiph:
Notes to self and anyone who wants these to stay:

* There are tests for each of the combinations if one ignores `Q` vs `q` vs `qq`, except, I think, `qv`.

* They need to be documented.

Notes to everyone:

* If these are to go, does that warrant a deprecation process?

TimToady:
My feeling was that it is easier to remember a single rule like "You can
always combine the first one" than it is to remember a list of which ones
do or do not work.

Larry

PR #5437, opened 2023-10-21
Hungarian localisation of core

2colours:


tools/templates/L10N/HU
```diff
@@ -0,0 +1,752 @@
+# This file contains that German translation of features of the
```
MasterDuke17, line 1: s/that German/the Hungarian/ (assuming of course that “German” is not Hungarian for “Hungarian”)

2colours: Good catch... most definitely work in progress.

tools/templates/L10N/HU
```diff
@@ -0,0 +1,752 @@
...
+# Raku Programming Language.
+#
+# CONTRIBUTORS: Elizabeth Mattijsen
```
lizmat, line 4: This also needs updating :-)

2colours:
Don't worry, I haven't forgotten about it but it's not really ready. This week I'll be probably buried with work to finish before a 1-week holiday, plus actually finishing the conference talks. So please hang on, this might have to wait a bit.

2colours:
What command `False`˙? :D

lizmat:
/me hopes to be able to mention a usable Hungarian localization in her advent post  *nudge* *nudge*

2colours:
Sorry I was on a "worcation" in Germany last week. Need to see the priority list now. 😅 Honestly, after this month, it's just nice to find an activity that doesn't cost an egregious amount of money...

2colours:
Thank you for your patience.

This is "as good as I could", meaning that I omitted some (roughly 10, I think) words from the translations, where I really couldn't think of anything. It's really tough with the snip-snitch-spurt-slurp-etc words that are slangy/idiomatic C2 words in English as well, pretty much.

lizmat:
Thanks!  Going to merge after the release!

PR #5442, opened 2023-10-23
Don't drop out empty lines where they a part of the text

vrurg:
Empty lines are likely to be part of document formatting. When RakuAST is used to manipulate the original rakudoc preserving them is highly desirable.

### DIFF
--- a/src/Raku/Grammar.nqp
+++ b/src/Raku/Grammar.nqp
@@ -5200,7 +5200,6 @@ Rakudo significantly on *every* run."
     token doc-TOP {
         { $*PARSING-DOC-BLOCK := 1 }
         :my $*SEEN := {};
-        <.doc-newline>*
         <doc-block>
         { $*PARSING-DOC-BLOCK := 0 }
     }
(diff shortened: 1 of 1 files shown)


lizmat:
It is my understanding that any leading whitespace in pod / rakudoc is insignificant.

I've already made rakudoc preserve trailing whitespace.  But preserving leading whitespace will create significant issues, I'm afraid.

Could you provide an example that is showing the problem?

vrurg:
First of all, there are still problems with extra whitespaces inserted:

```raku
```

vrurg:
BTW, this PR anyway breaks splitting text in paragraphs. It must be possible to fix this, but first I need to know if it worth time spent.

src/Raku/Grammar.nqp
```diff
@@ -5391,7 +5390,7 @@ Rakudo significantly on *every* run."
     }
 
     token doc-block:sym<lines> {
-        [^^ \h* [ <![=]> \N+ | '=' \W \N* ] \n?]+
```
vrurg, line 5394: @lizmat This line worth paying attention anyway because \N+ is clearly an error since it makes single-char lines illegal in a way.

vrurg:
It crossed my mind that skipping of empty lines before and after rakudoc blocks can be related to their incorporation into the code. Because `=begin rakudoc` is a block on its own.

Slightly above the line, I refer to above, the synopsis says: "blank lines within the contents are always treated as part of the contents." The content actually starts with any `=begin`.

PR #5450, opened 2023-10-29
Don't emit warnings when trying to access non-existent file

ugexe:


lizmat (approved):

src/core.c/CompUnit/Repository/Installation.rakumod
```diff
@@ -47,11 +47,10 @@ sub MAIN(*@, *%) {
...
             my $entry = $.meta<provides>.values.first: { $_{$address}:exists };
-            my $file = $entry
-                ?? $.prefix.add('sources').add($entry{$address}<file>)
-                !! $.prefix.add('resources').add($.meta<files>{$address});
-
-            IO::Handle.new(:path($file))
+            my $sources-path = $.prefix.add('sources').add($_) with $entry{$address}<file>;
+            my $resources-path = $.prefix.add('resources').add($_) with $.meta<files>{$address};
+            my $path = $sources-path // $resources-path;
+            IO::Handle.new(:$path);
```
ugexe, line 53: Hmm, actually we should probably throw an exception if the `$path` isn't defined. It doesn't really make much sense to return an `IO::Handle` to nothing

Issue #5448, opened 2023-10-29
multi sub signature for custom infix does not install suitable signature


ab5tract:
## The Problem

```
> RAKUDO_RAKUAST=1 ./rakudo-m -e 'multi sub infix:<quack>($x, $y) { "$x|$y" }; (* quack 5).(3) ==> say()'
Cannot resolve caller infix:<quack>(Int:D, Int:D); none of these signatures matches:
    ()
```

### Subtleties

The above works fine when not declared as a multi:

```
> RAKUDO_RAKUAST=1 ./rakudo-m -e 'sub infix:<quack>($x, $y) { "$x|$y" }; (* quack 5).(3) ==> say()'
3|5
```

and when an explicit proto is defined:

```
> RAKUDO_RAKUAST=1 ./rakudo-m -e 'proto sub infix:<quack>($, $) {*}; multi sub infix:<quack>($x, $y) { "$x|$y" }; (* quack 5).(3) ==> say()'
3|5
```

ab5tract:
Fully addressed with #5460 . Thanks @jnthn ++ for taking the time to review #5451 and suggest changes.

PR #5451, opened 2023-10-30
RakuAST: Re-generate meta-objects on name or signature change

ab5tract:
The previous behavior was only inheriting a blank signature when defining a custom operator.

This change allows the following code to work:

```
    multi sub infix:<quack>($x, $y) { "$x|$y" }; (* quack 5).(3).say
```

The following spectests pass as a result of this change:

```
	t/spec/integration/advent2009-day22.t ............................. ok
	t/spec/integration/advent2012-day19.t ............................. ok
```

### DIFF
--- a/src/Raku/ast/code.rakumod
+++ b/src/Raku/ast/code.rakumod
@@ -1390,11 +1390,13 @@ class RakuAST::Routine
 
     method replace-name(RakuAST::Name $new-name) {
         nqp::bindattr(self, RakuAST::Routine, '$!name', $new-name);
+        self.IMPL-RESET-CACHED-META-OBJECT;
         Nil
     }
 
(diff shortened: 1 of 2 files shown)


ab5tract:
This fixes #5448

jnthn:
This looks wrong. Other places might have references to the meta-object. I'd expect this to be done by producing a stub meta-object (there's already infrastructure for that) if it's needed before the signature is parsed, and then finishing it up at the closing `}`.

ab5tract:
@jnthn Thanks for weighing in on this. I will take a look at how to do this via StubbyMeta.

ab5tract:
Superseded by #5460 , which uses StubbyMeta as suggested.

PR #5452, opened 2023-10-31
Add L10N::PT

FCO:
Adding portuguese to the list of languages on L10N


lizmat:
The `tools/templates/L10N/PT` file appears to be missing.

Issue #5447, opened 2023-10-29
is default with where guard not being parsed in attribute declarations


ab5tract:
## Expected Behavior

```
> RAKUDO_RAKUAST=1 ./rakudo-m -e 'my class { has $.z is default(42) where * == 42 }.new.z ==> say()'
42
```

## Actual Behavior

```
> RAKUDO_RAKUAST=1 ./rakudo-m -e 'my class { has $.z is default(42) where * == 42 }.new.z ==> say()'
Confused
   at SETTING::src/core.c/Exception.rakumod:65  (/Users/longwalker/code/raku/z.others/rakudo/blib/CORE.c.setting.moarvm:throw)
```

Note, however, that is default works fine without the where guard:

```
> RAKUDO_RAKUAST=1 ./rakudo-m -e 'my class { has $.z is default(42) }.new.z ==> say()'
42
```

## Environment
* Operating system: macOS
* Compiler version (`rakudo -v` or `raku -v`): v2023.10-29-g7fcbfd657

ab5tract:
Having looked into it further, this actually affects all variable declarations. They simply aren't working on where clauses.

PR #5453, opened 2023-10-31
RakuAST: Support where blocks on variables

ab5tract:
Previously the grammar was confused by the existence of where blocks in variable declarations.

Now they are fully supported.

This diverges from base behavior in that previously we would output `<anon>` as the name of type that failed in a binding check. Now this is replaced by, for example, `Mu::Int+anon_subset_1`.

This can eventually be cleaned up such that the actual subset definition is displayed in the error message, so leaving it as an implementation detail for now.

[Unfortunately this commit does not introduce any new passing spectests]

### DIFF
--- a/src/Raku/ast/variable-declaration.rakumod
+++ b/src/Raku/ast/variable-declaration.rakumod
@@ -448,6 +448,7 @@ class RakuAST::VarDeclaration::Simple
     has RakuAST::Package     $!attribute-package;
     has RakuAST::Method      $!accessor;
     has RakuAST::Type        $!conflicting-type;
+    has RakuAST::Expression  $.where;
 
     has Mu $!container-initializer;
     has Mu $!package;
@@ -461,6 +462,7 @@ class RakuAST::VarDeclaration::Simple
         RakuAST::Initializer :$initializer,
            RakuAST::SemiList :$shape,
                         Bool :$forced-dynamic,
+         RakuAST::Expression :$where,
     RakuAST::Doc::Declarator :$WHY
     ) {
         my $obj := nqp::create(self);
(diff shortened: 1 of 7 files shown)


ab5tract:
This fixes #5447 .

src/Raku/ast/variable-declaration.rakumod
```diff
@@ -751,7 +766,7 @@ class RakuAST::VarDeclaration::Simple
         my @lookups;
         my str $scope := self.scope;
         # If we have a type, we need to resolve that.  Otherwise assume Mu
-        @lookups.push($!type || nqp::null);
+        @lookups.push($!type || RakuAST::Type::Setting.new(RakuAST::Name.from-identifier('Mu')));
```
niner, line 769: Can just be `@lookups.push($!type || Mu);`
Mu is declared in bootstrap (of which the RakuAST code is part of).

niner (changes requested):

src/Raku/ast/variable-declaration.rakumod
ab5tract: If I want to be able to do things like call methods on it (to get it's name, for example), it pays to have it as a RakuAST::Node. But maybe I can just create such a node without passing it to the produce-lookup?

ab5tract:
This whole branch seems to have gotten broken on its way to PR. Switching to 'Draft' until this is fixed.

src/Raku/ast/variable-declaration.rakumod
ab5tract: I was able to leave 'Mu' out of the implicit lookups without too much trouble.

ab5tract:
@niner Here is an _actually working_ implementation to review.

niner (approved):

PR #5460, opened 2023-11-05
RakuAST: Use stubby meta for Routine

ab5tract:
This is much cleaner approach, as it means that anything that has a reference to the stubbed meta object will still have a reference to it after names/types are reset.

Additionally, using this approach resolves two additional spec tests:

```
	t/spec/S32-list/produce.t ......................................... ok
	t/spec/S32-list/reduce.t .......................................... ok
```


ab5tract:
Address #5448 in a much cleaner, more robust way.

PR #5462, opened 2023-11-05
RakuAST: Make multi subs check for re-declaration of signatures

### LINKED ISSUE #5455: Rakudo should detect multi sub duplicates at compile time
This is the [oldest unsolved issue in the old-issue tracker.](https://github.com/Raku/old-issue-tracker/issues/1455). Won't try to summarize, because there's a lot of context there that could be lost.

ab5tract:
This addresses the oldest existing unsolved ticket in the old issue tracker:

Catch duplicate signatures in a multi sub at compile time, as they will inevitably fail at run-time.

Fixes #5455

### DIFF
--- a/src/Raku/ast/code.rakumod
+++ b/src/Raku/ast/code.rakumod
@@ -1772,6 +1772,7 @@ class RakuAST::Routine
 # A subroutine.
 class RakuAST::Sub
   is RakuAST::Routine
+  is RakuAST::CheckTime
   is RakuAST::SinkBoundary
 {
     has RakuAST::Blockoid $.body;
@@ -1821,10 +1822,62 @@ class RakuAST::Sub
         $signature ?? $signature.provides-return-value !! False
     }
 
+    method PERFORM-CHECK(Resolver $resolver, RakuAST::IMPL::QASTContext $context) {
+        return Nil unless self.multiness eq 'multi';
+
+        self.IMPL-CHECK-FOR-DUPLICATE-MULTI-SIGNATURES($resolver);
+    }
+
     method IMPL-COMPILE-BODY(RakuAST::IMPL::QASTContext $context) {
         self.IMPL-WRAP-RETURN-HANDLER($context,
             self.IMPL-WRAP-SCOPE-HANDLER-QAST($context,
                 self.IMPL-APPEND-SIGNATURE-RETURN($context, $!body.IMPL-TO-QAST($context))))
+    }
+
+    method IMPL-CHECK-FOR-DUPLICATE-MULTI-SIGNATURES(Resolver $resolver) {
+        my $proto := self.meta-object.dispatcher;
+        my $signature := self.signature.compile-time-value;
+        my $meta := self.meta-object;
+
+        # If we ourselves can default, there is no need to check further
+        return Nil if $meta.can("default");
+
+        my int $has-post-constraints;
+        for self.IMPL-UNWRAP-LIST($signature.params) {
+            $has-post-constraints := 1 if $_.constraint_list;
+        }
+
+        my @seen-accepts;
+        for $proto.dispatchees {
+            last if $_.can("default");
+            next if $_ =:= $meta;
+
+            my $other-signature := $_.signature;
+            next unless $signature.arity == $other-signature.arity;
+
+            my $other-has-post-constraints;
+            for self.IMPL-UNWRAP-LIST($other-signature.params) {
+                $other-has-post-constraints := 1 if $_.constraint_list
+            }
+
+            @seen-accepts.push($_)
+                if !($has-post-constraints || $other-has-post-constraints)
+                        && ($other-signature.ACCEPTS($signature) && $signature.ACCEPTS($other-signature));
+
+            if @seen-accepts > 0 {
+                my %args;
+                if my $origin := self.origin {
+                    my $origin-match := self.origin.as-match;
+                    %args<filename>  := $origin-match.file;
+                    %args<line>      := $origin-match.line
+                }
+                self.add-worry: $resolver.build-exception:
+                        'X::Redeclaration::Multi',
+                        :symbol(self.name.canonicalize),
+                        :ambiguous(@seen-accepts),
+                        |%args;
+            }
+        }
     }
 }
 
--- a/src/core.c/Exception.rakumod
+++ b/src/core.c/Exception.rakumod
@@ -1375,6 +1375,15 @@ my class X::Redeclaration::Outer does X::Comp {
     }
 }
 
+my class X::Redeclaration::Multi does X::Comp {
+    has $.symbol;
+    has @.ambiguous;
+    method message() {
+        ("Redeclaration of multi sub '$.symbol' with equivalent signatures would lead to ambiguous dispatch: "
+        ).naive-word-wrapper ~ "\n\t" ~ @!ambiguous.map({ .raku }).join("\n\t")
+    }
+}
+
 my class X::Dynamic::Postdeclaration does X::Comp {
     has $.symbol;
     method message() {


lizmat:
Cool!

Some general remarks.

1. I'm not sure whether this should be a panic, or a worry.  It feels more like a worry to me, because the problem may well not affect the running of a program.  If not a worry, then a sorry.

2. The panic/worry/sorry should be attached to the offending node.

3. The `seen-accepts` should be an array.  If it contains more than one element, it should be that list only that should be mentioned in the error message.

4. The `self.signature.compile-time-value` should probably be taken out of the loop.

ab5tract:
From the discussion in the pre-history of #5455, my impression was that the goal was to make this case a compile-time-error. I used panic because it allows me to specify a single exception. If I simply add a sorry then the exception will show up twice (or more, matching the number of overlapping candidates).

Can you take a look at #5455 and double-check my understanding?

lizmat:
> If I simply add a sorry then the exception will show up twice (or more, matching the number of overlapping candidates)

True that.  Which could be handled by setting a flag on the proto.  But yes.  OK, panic it is then  :-)

ab5tract:
There are test "failures" that really represent changes in behavior:

```
> rt t/spec/integration/advent2011-day24.t
1..8
ok 1 - multi
ok 2 - multi
ok 3 - Yum! /Users/longwalker/code/raku/z.others/rakudo/t/spec/integration/advent2011-day24.t was tasty. Got another one?
ok 4 - unambiguous multi - lives
# Subtest: ambiguous multi - dies
    1..2
    ok 1 - code dies
    not ok 2 - right exception type (X::Multi::Ambiguous)
    # Failed test 'right exception type (X::Multi::Ambiguous)'
    # at /Users/longwalker/code/raku/z.others/rakudo/lib/.precomp/64ECA66FBBADFDAC70EC53A38CB39CFF935D9729/71/71012CB8C080E8C3C2FA5692E6AAF5F974F3DDAA line 1
    # Expected: X::Multi::Ambiguous
    # Got:      X::Redeclaration::Multi
    # Exception message: Redeclaration of multi sub 'Slurp' with equivalent signatures would
    # lead to ambiguous dispatch:
    # 	multi sub Slurp ($filename) { #`(Sub|3671839658496) ... }
    # 	multi sub Slurp ($filename) { #`(Sub|3671839658624) ... }
    # You failed 1 test of 2
not ok 5 - ambiguous multi - dies
# Failed test 'ambiguous multi - dies'
# at /Users/longwalker/code/raku/z.others/rakudo/t/spec/integration/advent2011-day24.t line 35
ok 6 - It's wrapped.
ok 7 - You unwrap the present and find...!
ok 8 - A sock!
# You failed 1 test of 8
```

and failures in `t/12-rakuast/sub.rakutest` and `t/spec/S06-multi/lexical-multis.t` that result from reliance on the existing behavior of compiling same-signature multi subs.

Assuming we want this long-standing misbehavior fixed, it seems that we have to adjust the language spec slightly to do it. Is this the right place for discussion or should I move it to #5455 ?

niner:
But a panic is when the compiler absolutely cannot continue. That's why it only takes one exception. There simply cannot be more. However in this case the compiler could easily continue. It would even succeed and create runnable code. I think a worry is totally fine for this.

lizmat:
#5455 feels like the right place for that discussion.

src/Raku/ast/code.rakumod
```diff
@@ -1826,6 +1827,43 @@ class RakuAST::Sub
             self.IMPL-WRAP-SCOPE-HANDLER-QAST($context,
                 self.IMPL-APPEND-SIGNATURE-RETURN($context, $!body.IMPL-TO-QAST($context))))
     }
+
+    method PERFORM-CHECK(Resolver $resolver, RakuAST::IMPL::QASTContext $context) {
+        return Nil unless self.multiness eq 'multi';
+
+        my $proto := $resolver.resolve-lexical('&' ~ self.name.canonicalize, :current-scope-only).compile-time-value.derive_dispatcher;
```
niner, line 1834: This looks wrong. PERFORM-CHECK is usually called after we are done parsing. That means the resolver will have exited all lexical scopes already. E.g.
```
{
     multi sub foo() {}
     multi sub foo() {} # we're checking this
} 
# <- resolver will look from here
```

The only place where the resolver is guaranteed to be in the right state for resolving lexicals is the attach method.

niner (changes requested):

src/Raku/ast/code.rakumod
ab5tract: But we are dealing with a (likely) generated proto declaration and that happens outside of attach? 

We also need all of the begin time setup to be performed (so for instance 'is default' has been mixed in).

Is it okay to store a reference to the scope in attach and use that reference in PERFORM-CHECK?

niner: Well, sins of the past, what can I say?
While in BEGIN processing we're no longer guaranteed that the resolver is in the correct state, it's still likely as BEGIN will be handled e.g. when we encounter the closing } of a routine. Which will be good enough to find the proto. When using the parser that is. For synthetically generated ASTs it may be quite different and it's likely that we need to re-visit the existing code as well.

That said, there should not be a reason for you to look up the proto in the first place. After all you already have access to one of the multi candidates. A multi candidate knows its proto. `self.meta-object.dispatcher` will give it to you. Also there's no need to derive a dispatcher here. You don't want to add any candidates. You're just checking the existing ones.

ab5tract: Ah, yes, I should have remembered that I can get the dispatcher directly from the mega object.

So doing it that way, I assume there is no issue with the check remaining at check time?

niner: Not at all. Check time is perfect for this.

ab5tract: This is addressed in the newest implementation. As per the discussion in #5455, it sounds like we are going with a sorry for this case.

lizmat: Since there are valid reasons for having multiple candidates matching, as @niner described, shouldn't this be a worry instead then?

patrickbkr: As far as I see there are three options:
- We allow identically named methods. I oppose this. This feature hasn't been explored at all yet. The potential for confusion is large. (You don't notice you have two identical methods, only look at one of the two and wonder what's happening.)
- We worry this. I also oppose this, but slightly less. People will be warned, but we still open the door for such code to appear in the Raku ecosystem. And then we're locked in to support this.
- We sorry this. Should someone want this later on and the use case is legit, we can then start evaluating the pros and cons and take a well thought through decision to add it or not.

ab5tract:
@patrickbkr I'm unable to see any comments you may have left on this PR. An unfortunate side effect of force pushing some changes.

ab5tract:
It now does a worry, as per the conclusion of #5455. Since this doesn't require a spec change, all tests are passing.

JJ:
> It now does a worry, as per the conclusion of #5455. Since this doesn't require a spec change, all tests are passing.

Would it maybe be convenient to write a specific test for this, so that it *always* returns a worry?

ab5tract:
Testing for worries is probably most easily done via is-run but that requires Test::Helpers. So that's a different commit in roast that can be added after this is merged, as part of closing #5455.

src/Raku/ast/code.rakumod
```diff
@@ -1821,11 +1822,61 @@ class RakuAST::Sub
...
+            my $other-has-post-constraints;
+            for self.IMPL-UNWRAP-LIST($other-signature.params) {
+                $other-has-post-constraints := 1 if $_.constraint_list
+            }
+
+            @seen-accepts.push($_)
+                if !($has-post-constraints || $other-has-post-constraints)
+                        && ($other-signature.ACCEPTS($signature) && $signature.ACCEPTS($other-signature));
+
+            if @seen-accepts > 1 {
```
niner, line 1865: You could `next if $_ =:= $meta-object` early in the loop to avoid comparing this method with itself (which should always match anyway) and then change this to `@seen-accepts > 0`

niner (changes requested):
With the one comment addressed, this will be good to go in.

src/Raku/ast/code.rakumod
ab5tract: Can't believe I missed this one! Fixed in latest.

PR #5465, opened 2023-11-08
Fix some aspects of using `is default`

vrurg:
Here is what is taken care of in this PR:

- `(my Int:D %{Str:D} is default(42))<foo>:delete` used to return `Int:D` despite the default. It would now return 42 which puts it in the same league as untyped hashes and arrays, both typed and untyped.
- change the exception thrown when `is default` value type doesn't match attribute's constraint. It used to be `X::TypeCheck::Assignment`, as the default for containers. It is now `X::TypeCheck::Attribute::Default`.
- fix a couple of minor issues with related exceptions

### DIFF
--- a/src/core.c/Exception.rakumod
+++ b/src/core.c/Exception.rakumod
@@ -1461,7 +1461,7 @@ my class X::Parameter::Default::TypeCheck does X::Comp {
     has $.got is default(Nil);
     has $.expected is default(Nil);
     method message() {
-        "Default value '{Rakudo::Internals.MAYBE-STRING: $!got}' will never bind to a parameter of type {$!expected.^name}"
+        "Default value '{Rakudo::Internals.MAYBE-STRING: $!got}' will never bind to a {$!what} of type {$!expected.^name}"
     }
 }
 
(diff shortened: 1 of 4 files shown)


vrurg:
As to changing the exception thrown for attributes, it's a big painful topic, I'm afraid. Even the related test in Roast's _S02-names/is_default.t_ contradicts itself as it was expecting two different exception for the same error. Even though tests for `X::Parameter::Default::TypeCheck` are fudged, they're expected to pass and thus I consider them no less relevant as checks for `X::TypeCheck::Assignment`.

In my view, best if we introduce `X::TypeCheck::Default` for both `is default` and `my Int:D $ = "aa"` cases. It makes sense because both are about default values in first place. The exception could possibly be general enough to be able to cover all cases like variable, attributes, parameters, hash and array values. Then related exceptions would inherit from this one.

`X::Parameter::Default::TypeCheck` could be a painful one as it is really old. But its reincarnation is not really impossible.

Either way, this PR tries to maintain as much of status quo as possible. The change of attribute case exception is revertable and not mandatory for the entire PR.

PR #5466, opened 2023-11-10
RakuAST: change rakufication of `Rat`s

lizmat:
If the Rat cannot be displayed using a decimal notation (such as <22/3>), then rakufy as a vulgar: 7¹/₃ instead of <22/3> if the RakuAST grammar is being used.  The RakuAST grammar allows vulgars such as  7¹/₃ to represent <22/3>, so roundtripping is ensured.

### DIFF
--- a/src/core.c/Rat.rakumod
+++ b/src/core.c/Rat.rakumod
@@ -1,3 +1,5 @@
+my class Raku { … }
+
 # XXX: should be Rational[Int, uint]
 my class Rat is Cool does Rational[Int, Int] {
     method Rat(Rat:D: Real $? --> Rat:D) {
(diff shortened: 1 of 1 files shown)


ab5tract:
I would like to propose not using composed variants even when available so that the output looks consistent.

raiph:
@ab5tract 

I imagine that you could introduce a `:vulgar` adverb. (And if so, use the interface consistency principle, i.e. if someone calls `.raku` on some code, they can pass adverbs and `.raku` will pass those on to calls of `.raku` on particular parts of the code, so an adverb like `:!vulgar` will distribute to the `.raku` of the parts.)

lizmat:
Removed composing feature: it was applicable only for a handful cases anyway.

PR #5471, opened 2023-11-15
Add ! twigil to private attribute

codesections:
Searching the Rakudo codebase shows that Rakudo consistently uses twigils for all attributes with a single exception (outside of test code).  This PR removes that one exception.

See raku/problem-solving#358 for related discussion of whether non-twigiled attributes should be deprecated


lizmat:
Good catch!

PR #5473, opened 2023-11-19
Fix cases where the last element is an undefined

vrurg:
When the last element of a hash initializer list is an undefined it is not reported by the exception. It'd be rather beneficial for diagnosis if it is reported.


vrurg:
Accidental duplicate.

PR #5477, opened 2023-11-27
Move closer to complete generics instantiation

vrurg:
Targetting cases where auto_viv_container is remaining uninstantiated resulting in attributes still using a generic even when installed into classes from a fully specialized role. In particular, the following case is resolved:

```raku
role R[::T] { has @.a is T; }
class C does R[Array[Str:D]] {}
```

Also some cases of scalar containers where default is set to the generic are fixed too:

```raku
role R[::T] { has $.a is default(T); }
```

Key changes:

* Instantiation protocol is introduced. 

  Sometimes generics can make their way inside classes or even objects. For example, an Array could be parameterized with a generic. Currently we cannot properly instantiate it unless some magic is introduced. But magic is magic because it'd be specifically targetting a particular case, leaving, say, user classes in limbo.

  The following bullets are parts of the protocol implementation.
* With this PR method `instantiate_generic` becomes a part of `ClassHOW`. It takes a type environment, creates an instance of Raku wrapper class TypeEnv (a map descenadant) for it, and then submits it to `INSTANTIATE-GENERIC` method, expecting the class itself to do the rest.
* A class or an instance of class can now be a generic. This would be reported as a class archetype, i.e. `MyContainer.^archetypes.generic` would be set.
* All parameterizable classes must be now OK with generics.

### DIFF
--- a/src/Perl6/Actions.nqp
+++ b/src/Perl6/Actions.nqp
@@ -3854,6 +3854,7 @@ class Perl6::Actions is HLL::Actions does STDActions {
 
         my $of_type := nqp::null;
         my $is_type := nqp::null;
+        my $*IS-TYPE-TRAIT := nqp::null;
 
         $world.handle_OFTYPE_for_pragma($/, $*SCOPE eq 'has' ?? 'attributes' !! 'variables');
         if $*OFTYPE {
@@ -3912,6 +3913,7 @@ class Perl6::Actions is HLL::Actions does STDActions {
                                       $type.HOW.parameterize(
                                         $type, |$params.FLATTENABLE_LIST)
                                     );
+                                    $*IS-TYPE-TRAIT := $trait;
                                     next;  # handled the trait now
                                 }
                             }
@@ -6048,6 +6050,7 @@ class Perl6::Actions is HLL::Actions does STDActions {
                 |%!named_args, |%additional);
         }
 
+        method match() { $!match }
         method mod() { $!trait_mod }
         method args() { @!pos_args }
     }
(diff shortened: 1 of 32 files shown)


vrurg:
The instantiation protocol should be part of 6.e, but implementation-wise it takes too much effort do it right. In particular, `is-generic` and `INSTANTIATE-GENERIC` methods are to be added to `Mu`. `Array::Typed` needs to implement them too. This makes augmenting `Mu` in 6.e core problematic because it then overrides whatever is part of 6.c core, meaning shadowing over `Array::Typed` implementation which, as a role, cannot be augmented.

Therefore I currently see it as something that's to be kind of undocumented feature of 6.c which is fully "legal" in 6.e.

vrurg:
The PR would already be ready for accepting unless a very weird case of _t/S12-rakuast/doc-declarator.rakutest_, 'Simple class with declarator pod, one private method' subtest. No idea how is it possible, but `nqp::tryfindmethod` word perfectly fine for the RakuAST-produced class `A` in the subtest body, and yet for the very same class the op returns `null` within method `new` of `Mu`! This doesn't happen without my changes.

My primary culprit are changes to the `Actions.nqp`, where the initializer thunk is given a closure by installing it directly into the mainline. The idea was taken from how `sub`s are installed, but there is a catch which worries me as subs are also recorded in the lexpad whereas the thunk is nameless. But OTOH, it is bound to the attribute object and shouldn't be a subject for GC collection.

vrurg:
Done and ready to merge.

PR #5480, opened 2023-12-01
Minor cleanups for #5477

vrurg:
Resolve a couple of stylistically bad and potentially unsafe pieces.


vrurg:
Not relevant anymore as #5477 had to be reverted.

PR #5485, opened 2023-12-03
 Deprecate IO subs that operate on multiple paths in 6.e

ugexe:
While convienient, being able to pass multiple paths to e.g. &rmdir is problematic because there isn't a blessed way to handle what to do when an exception or failure occurs, i.e. should it try to rmdir every entry even if some things fail, or stop on the first failure?

Previously `rmdir("doesn't exist", "also doesn't exist")` would return an empty array with no indication of what the failure was. Yet `("doesn't exist", "also doesn't exist").grep(*.IO.rmdir) does get failures for these. This difference in behavior is not intuitive.

This changes subroutines that act on multiple paths by adding a single path variant, as well as deprecating the multi file variant. The inclusion of a single path variant is a breaking change, as these subroutines will now return a failure instead of an array.

### DIFF
--- a/src/core.c/io_operators.rakumod
+++ b/src/core.c/io_operators.rakumod
@@ -282,33 +282,78 @@ multi sub indir(IO() $path, &what, :$d = True, :$r, :$w, :$x) {
     }
 }
 
-proto sub chmod($, |) {*}
-multi sub chmod($mode, *@filenames) {
+# NOTE: chmod / chown / unlink / rmdir - these multis currently use two candidates (Str and IO) where
+#       theoretically only one is necessary (IO()). This is because the current revision gating implementation
+#       is not applying to dispatch of "non-trivial" multis -- of which coercers combined with untyped slurpies are
+#       are an example.
+
+proto sub chmod($, |) is revision-gated("6.c") {*}
+multi sub chmod($mode, *@filenames) is revision-gated("6.c") {
     my @ok;
     for @filenames -> $file { @ok.push($file) if $file.IO.chmod($mode) }
     @ok;
 }
+multi sub chmod($mode, *@filenames) is revision-gated("6.e") {
+    if @filenames == 1 {
+        # Technically this should probably error when the type is neither, but for the sake of brevity
+        # we can leave that to a failure against the non-deprecated candidate
+        my $type = @filenames[0] ~~ Str ?? "Str" !! "IO";
+        DEPRECATED(Q:q|chmod(\qq{$type} $path, Int() :$mode)|, :what("deprecated routine signature"));
+    } else {
+        DEPRECATED('@paths.grep(*.IO.chmod)', :what("multi-path operation"));
+    }
+    my @ok;
+    for @filenames -> $file { @ok.push($file) if $file.IO.chmod($mode) }
+    @ok;
+}
+multi sub chmod(Str $path, Int() :$mode!) is revision-gated("6.e") { $path.IO.chmod($mode) }
+multi sub chmod(IO $path, Int() :$mode!) is revision-gated("6.e") { $path.chmod($mode) }
 
-proto sub chown($, |) {*}
-multi sub chown(*@filenames, :$uid, :$gid) {
+proto sub chown($, |) is revision-gated("6.c") {*}
+multi sub chown(*@filenames, :$uid, :$gid) is revision-gated("6.c") {
     @filenames.grep: *.IO.chown(:$uid, :$gid)
 }
+multi sub chown(*@filenames, :$uid, :$gid) is revision-gated("6.e") {
+    my $deprecated-message = Q:c[@paths.grep(*.IO.chown({$uid.defined ?? ':uid(' ~ $uid ~ ')' !! ''}]
+                                ~ Q:c[{$uid.defined && $gid.defined ?? ', ' !! ''}]
+                                ~ Q:c[{$gid.defined ?? ':gid(' ~ $gid ~ ')' !! ''}))];
+    DEPRECATED($deprecated-message, :what("multi-path operation"));
+    @filenames.grep: *.IO.chown(:$uid, :$gid)
+}
+multi sub chown(Str $path, :$uid, :$gid) is revision-gated("6.e") { $path.IO.chown(:$uid, :$gid) }
+multi sub chown(IO $path, :$uid, :$gid) is revision-gated("6.e") { $path.chown(:$uid, :$gid) }
 
-proto sub unlink(|) {*}
+proto sub unlink(|) is revision-gated("6.c") {*}
 multi sub unlink() { "unlink()".no-zero-arg }
-multi sub unlink(*@filenames) {
+multi sub unlink(*@filenames) is revision-gated("6.c") {
     my @ok;
     for @filenames -> $file { @ok.push($file) if $file.IO.unlink }
     @ok;
 }
+multi sub unlink(*@filenames) is revision-gated("6.e") {
+    DEPRECATED('@paths.grep(*.IO.unlink)', :what("multi-path operation"));
+    my @ok;
+    for @filenames -> $file { @ok.push($file) if $file.IO.unlink }
+    @ok;
+}
+multi sub unlink(Str $path) is revision-gated("6.e") { $path.IO.unlink }
+multi sub unlink(IO $path) is revision-gated("6.e") { $path.unlink }
 
-proto sub rmdir(|) {*}
+proto sub rmdir(|) is revision-gated("6.c") {*}
 multi sub rmdir() { "rmdir()".no-zero-arg }
-multi sub rmdir(*@filenames) {
+multi sub rmdir(*@filenames) is revision-gated("6.c") {
     my @ok;
     for @filenames -> $file { @ok.push($file) if $file.IO.rmdir }
     @ok;
 }
+multi sub rmdir(*@filenames) is revision-gated("6.e") {
+    DEPRECATED('@paths.grep(*.IO.rmdir)', :what("multi-path operation"));
+    my @ok;
+    for @filenames -> $file { @ok.push($file) if $file.IO.rmdir }
+    @ok;
+}
+multi sub rmdir(Str $path) is revision-gated("6.e") { $path.IO.rmdir }
+multi sub rmdir(IO $path) is revision-gated("6.e") { $path.rmdir }
 
 proto sub mkdir($, $?, *%) {*}
 multi sub mkdir(IO() $path, Int() $mode = 0o777) { $path.mkdir($mode) }
--- a/src/core.c/Exception.rakumod
+++ b/src/core.c/Exception.rakumod
@@ -3345,6 +3345,15 @@ my class X::Multi::NoMatch is Exception {
     has $.capture;
     method message {
         my @cand = $.dispatcher.dispatchees.map(*.signature.gist);
+
+        if $!dispatcher.can('REQUIRED-REVISION') && $!dispatcher.REQUIRED-REVISION {
+            @cand = @cand Z~ $!dispatcher.dispatchees.map({
+                $^c.can('REQUIRED-REVISION')
+                        ?? Q:c[ is revision-gated("6.{nqp::chr(nqp::ord('c') + ($c.REQUIRED-REVISION - 1))}")]
+                        !! ""
+            })
+        }
+
         my @un-rw-cand;
         if first / 'is rw' /, @cand {
             my $rw-capture = Capture.new(
--- a/src/core.c/IO/Path.rakumod
+++ b/src/core.c/IO/Path.rakumod
@@ -499,6 +499,7 @@ my class IO::Path is Cool does IO {
         my str $path = self.absolute;
 #?if moar
         my $stat := nqp::syscall("file-stat", nqp::decont_s($path), 0);
+        die "Path does not exist" unless nqp::syscall("stat-flags", $stat, nqp::const::STAT_EXISTS);
 #?endif
         $uid = $uid.defined
           ?? $uid.UInt
--- a/src/vm/moar/dispatchers.nqp
+++ b/src/vm/moar/dispatchers.nqp
@@ -2838,6 +2838,7 @@ sub multi-ambiguous-handler($ambig-call, $target, $capture) {
 # Helper sub to filter out revision-gated multi-candidates
 sub multi-filter-revision-gated-candidates($proto, $caller-revision) {
     my @candidates := $proto.dispatch_order;
+    my $proto-revision := $proto.REQUIRED-REVISION;
 
     my int $idx := 0;
     my int $allowed-idx := 0;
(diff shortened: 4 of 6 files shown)


ugexe:
This probably requires some general consensus. Note there will be a bunch of spectest failures because `TestHelpers::Util` is currently using `rmdir(...)` and `unlink(...)` in a way such as to avoid any error messages when it should instead be using `try`.

vrurg:
Can we make it 6.e only? I generally concur, but best if we don't touch 6.c/d.

t/02-rakudo/03-corekeys.t
```diff
@@ -813,6 +813,7 @@ my @allowed =
         Q{&postcircumfix:<[; ]>},
         Q{&postcircumfix:<{; }>},
         Q{&prefix:<//>},
+        Q{&rmdir},
```
ugexe, line 816: I'm not sure why this test only requires the addition of `&rmdir` and not the other subs I adjusted

ab5tract:
Just a thought but you might be able to make good use of the new 'is revision-gated' trait here for ensuring that the new single path candidate is only dispatched in 6.e.

lizmat:
I would be in favour of making all of the `IO` subs return proper info on success / failure in 6.e, using the `is revision-gated` trait.

ugexe:
> Just a thought but you might be able to make good use of the new 'is revision-gated' trait here for ensuring that the new single path candidate is only dispatched in 6.e.

I haven't really thought about it, but what would that look like? Would it be adding the single path candidates to whatever file currently contains e.g. `rmdir()` with that trait and adding the deprecation notice to the multi path candidates, instead of doing those in `core.e/additions.rakumod`?

> I would be in favour of making all of the IO subs return proper info on success / failure in 6.e, using the is revision-gated trait.

If you're suggesting single argument paths should return failures yes I agree, but I don't think that is in scope of this PR and can easily enough be implemented and decided on outside of this discussion.

If it is being suggested we throw errors when calling a IO sub with multiple paths I'd argue only naive solutions exist and thus shouldn't be in the core at all. For example do you return a failure when the first path passed returns a failure? Do you try to e.g. unlink all files even if one fails? These are rhetorical questions as whatever option one might pick I can argue why it isn't ideal. Those type of decisions should be decided in user space where those tradeoffs have to be explicitly understood when making them, not in core where users will equally expect various possible actions.

ab5tract:
> > Just a thought but you might be able to make good use of the new 'is revision-gated' trait here for ensuring that the new single path candidate is only dispatched in 6.e.
> 
> I haven't really thought about it, but what would that look like? Would it be adding the single path candidates to whatever file currently contains e.g. `rmdir()` with that trait and adding the deprecation notice to the multi path candidates, instead of doing those in `core.e/additions.rakumod`? That seems like anytime anyone does e.g. `rmdir()` with any number of arguments but without declaring `use v6.e` a deprecation notice would be emitted.

I'm a bit confused here as revision gating can be used to arrive at literally whatever effects you want.

My suggestion was that we can avoid spending an entire language revision with a deprecation message and move straight to the desired outcome in `6.e` by revision gating only the new, single argument candidate of `rmdir` to `6.e`.

Or if we feel it is important to make a seamless upgrade path, I would suggest a full-on `SORRY` for the old candidate in `6.e`.

In`6.c`/`6.d` the old candidate would just work as before -- or else display a deprecation message. Like I said, you can achieve just about whatever effect you might want.

Considering how long it is taking to get `6.e` out -- and our current core developer horsepower -- I personally don't see much reason to delay tangible improvements to the language by a whole revision cycle  when it's simple for urser to retain any necessary semantics by pinning to a specific language revision until the code can be upgraded.

I could easily be missing something, though!

> > I would be in favour of making all of the IO subs return proper info on success / failure in 6.e, using the is revision-gated trait.
> 
> If you're suggesting single argument paths should return failures yes I agree, but I don't think that is in scope of this PR and can easily enough be implemented and decided on outside of this discussion.

I was just stating that I agree with the approach of returning failures and that this should happen in the new candidate.

ugexe:
> instead move straight to the desired outcome in 6.e by revision gating only the new, single argument candidate of rmdir to 6.e.

The desired outcome includes emitting a deprecation message when passing multiple paths. The current existing candidate handles and number of paths, and so adding a deprecation message there ends up being a bit weird because it is based on parameters of the candidiate (since it has to check the number of paths passed for e.g. 6.c and only emit a deprecation if the number of parameters is greater than 1) instead of the candidiate itself. Maybe that’s fine, but it does seem like a code smell to me personally; it is more clear to have a deprecation apply to an entire candidate when possible (such as documenting deprecated candidates). So to be clear: to use the is revision gated trait in a way that satisfies the things I’ve just mentioned would involve adding an additional 1 path candidiate accessible from eg. 6.c that does not emit a deprecation message. If I’m missing something and there is a simpler way to do it feel free to show an example or just push to this branch as I have no strong opinion on how it is gated as long as the behavior I described is observed.


> I was just stating that I agree with the approach of returning failures and that this should happen in the new candidate.

The quote you are responding to was a reply to Lizmat fwiw

ab5tract:
> > I was just stating that I agree with the approach of returning failures and that this should happen in the new candidate.
> 
> The quote you are responding to was a reply to Lizmat fwiw

Ah, it sounds so similar to something I might have said that I didn't even question it 😅 

I've pushed an edition that I hope matches what your expectations. In this current patch, I've left all calls in earlier versions totally alone, so deprecations are only introduced in `6.e`. Other options are available and I hope this patch highlights that.

I also thought it might be nice to adjust the calling convention for `chmod` to match `chown`, ie switching `$mode` from being the first positional argument to a required named argument. Deprecation messages will be thrown for the old signature starting with 6.e, where the new signature will become available.

Feel free to revise or revert as you see fit, I just wanted to give a demonstrative example of the options.

src/core.c/io_operators.rakumod
```diff
@@ -282,34 +282,77 @@ multi sub indir(IO() $path, &what, :$d = True, :$r, :$w, :$x) {
...
+# We also provide the new signature and make it available to 6.e+
+multi sub chmod($mode, Str $path) is revision-gated("6.e") {
+    DEPRECATED('chmod(Str $path, :$mode)', :what("old-fashioned chmod signature"));
+    $path.IO.chmod($mode)
+}
+multi sub chmod($mode, IO::Path $path) is revision-gated("6.e") {
+    DEPRECATED('chmod(IO::Path $path, :$mode)', :what("old-fashioned chmod signature"));
+    $path.chmod($mode)
+}
+multi sub chmod(Str $path, :$mode!) is revision-gated("6.e") { $path.IO.chmod($mode) }
```
ugexe, line 299: The `Str` and `IO::Path` typing for non 6.e candidates is technically a breaking change and we probably shouldn't change that.

ugexe, line 299: We can use `IO()` instead of having two candidates (one for `Str` and one for `IO::Path`).

src/core.c/io_operators.rakumod
```diff
@@ -282,34 +282,77 @@ multi sub indir(IO() $path, &what, :$d = True, :$r, :$w, :$x) {
...
+multi sub chmod($mode, Str $path) is revision-gated("6.c") { $path.IO.chmod($mode) }
+multi sub chmod($mode, IO::Path $path) is revision-gated("6.c") { $path.chmod($mode) }
+# In 6.e, we adjust the routine arguments to align with how chown is called
+# So we set up 6.e variants of the old signature that include deprecation messages
+# We also provide the new signature and make it available to 6.e+
+multi sub chmod($mode, Str $path) is revision-gated("6.e") {
+    DEPRECATED('chmod(Str $path, :$mode)', :what("old-fashioned chmod signature"));
+    $path.IO.chmod($mode)
+}
+multi sub chmod($mode, IO::Path $path) is revision-gated("6.e") {
```
ugexe, line 295: The `Str` and `IO::Path` typing for non 6.e candidates is technically a breaking change so we probably shouldn't change them.

src/core.c/io_operators.rakumod
```diff
@@ -282,34 +282,77 @@ multi sub indir(IO() $path, &what, :$d = True, :$r, :$w, :$x) {
...
+}
+multi sub unlink(*@filenames) is revision-gated("6.e") {
+    DEPRECATED('@paths.grep(*.IO.unlink)', :what("multi-path operation"));
     my @ok;
     for @filenames -> $file { @ok.push($file) if $file.IO.unlink }
     @ok;
 }
 
-proto sub rmdir(|) {*}
+proto sub rmdir(|) is revision-gated("6.c") {*}
```
ugexe, line 341: Is the prototype supposed to be revision gated?

lizmat: It's my understanding that that is essential.

ab5tract: If you don't put it on the prototype, the dispatcher will have no indication that it needs to engage revision gating amongst candidates.

Automatically adding this trait was not feasible and (imo) not advisable anyway. Sometimes our candidate lists span for hundreds of lines, so being explicit at the prototype declaration makes sense to me even if it were possible to avoid doing so.

src/core.c/io_operators.rakumod
ab5tract: If a non-Str, non-IO::Path argument arrives, it will simply go through the slurpy candidate.

src/core.c/io_operators.rakumod
ab5tract: This occurred to me but I expect IO() to be slower than matching by type across two candidates.

In the end it's probably better, though.

ugexe: Yeah I can see benefits to both ways.

The performance difference would only be non-negligible when using the `IO::Path` candidate and even then it would have to be called a lot. And if someone is that concerned with performance then they can just call the method variants instead of the subroutines. Using `IO()` cuts down on the number of candidates which makes the documentation a bit nicer imo, as well as helps to avoid  progressing further towards a combinatorial explosion  (revision candidates + type candidates + potentially more candidate specific options added in the future).

On the other hand is does feel a bit bad to discourage the functional candidates via the performance reduction, but to be fair that already exists and will exist regardless.

ab5tract: Agreed. I'll change it later today. It's better in the long run anyway because it will "explode" at dispatch instead of inside the routine (when parameter values can't coerce to IO) while also self documenting the solution (provide a coercer or send a value of a type that already has a coercer).

Also, I bet there is an optimization strategy we can implement at some point that can avoid/reduce the costs even for a large number of IO::Path objects.

src/core.c/io_operators.rakumod
ab5tract: I think this should be fine given the IO() candidates?

ab5tract:
Any idea why xx-fixed-in-rakuast.rakutest is failing for this PR?

ugexe:
xx-fixed-in-rakuast.rakutest is fine, the failures are all TODOd. The actual issue is from `t/02-rakudo/03-corekeys.t`
```
$ raku t/02-rakudo/03-corekeys.t
1..3
# Subtest: Symbols in CORE::v6c
    1..2
    ok 1 - No unexpected entries
    ok 2 - No missing entries
ok 1 - Symbols in CORE::v6c
# Subtest: Symbols in CORE::v6d
    1..2
    ok 1 - No unexpected entries
    ok 2 - No missing entries
ok 2 - Symbols in CORE::v6d
# Subtest: Symbols in CORE::v6e
    1..2
    # Missing 4 expected entries: &chmod &chown &rmdir &unlink
    ok 1 - No unexpected entries
    not ok 2 - No missing entries
    # Failed test 'No missing entries'
    # at t/02-rakudo/03-corekeys.t line 873
    # You failed 1 test of 2
not ok 3 - Symbols in CORE::v6e
# Failed test 'Symbols in CORE::v6e'
# at t/02-rakudo/03-corekeys.t line 873
# You failed 1 test of 3
```

src/core.c/io_operators.rakumod
```diff
@@ -283,21 +283,15 @@ multi sub indir(IO() $path, &what, :$d = True, :$r, :$w, :$x) {
...
 proto sub chmod($, |) is revision-gated("6.c") {*}
-multi sub chmod($mode, Str $path) is revision-gated("6.c") { $path.IO.chmod($mode) }
-multi sub chmod($mode, IO::Path $path) is revision-gated("6.c") { $path.chmod($mode) }
+multi sub chmod($mode, IO() $path) is revision-gated("6.c") { $path.IO.chmod($mode) }
 # In 6.e, we adjust the routine arguments to align with how chown is called
 # So we set up 6.e variants of the old signature that include deprecation messages
 # We also provide the new signature and make it available to 6.e+
-multi sub chmod($mode, Str $path) is revision-gated("6.e") {
+multi sub chmod($mode, IO() $path) is revision-gated("6.e") {
     DEPRECATED('chmod(Str $path, :$mode)', :what("old-fashioned chmod signature"));
```
ugexe, line 291: These deprecation message should probably have the types mentioned updated

ugexe:
This is not passing spectests

ugexe:
I rebased to get the CI updates that should show the spectest failing

src/core.c/io_operators.rakumod
```diff
@@ -282,34 +282,68 @@ multi sub indir(IO() $path, &what, :$d = True, :$r, :$w, :$x) {
     }
 }
 
-proto sub chmod($, |) {*}
-multi sub chmod($mode, *@filenames) {
+proto sub chmod($, |) is revision-gated("6.c") {*}
+multi sub chmod($mode, IO() $path) is revision-gated("6.c") { $path.chmod($mode) }
```
ugexe, line 286: Hmm, why is this candidate needed? I expected that I could delete this candidate and that the `multi sub chmod($mode, *@filenames) is revision-gated("6.c")` candidate would get used, but it doesn't:
```
raku t/spec/S16-filehandles/chmod.t
1..19
# Using file temp_1735619232
Cannot resolve caller chmod(Int:D, Str:D); none of these signatures matches:
    ($mode, IO(Any) $path)
    (IO(Any) $path, Int(Any) :$mode!)
    ($mode, *@filenames)
    ($mode, *@filenames)
  in block <unit> at t/spec/S16-filehandles/chmod.t line 38

# You planned 19 tests, but ran 0
```

ab5tract:
Still poking at the revision gating code in dispatchers.nqp

ab5tract:
> Still poking at the revision gating code in dispatchers.nqp

So this turns out to be an issue with how 'non-trivial' dispatching proceeds through it's candidates. The non-triviality of the `IO()`, the `*@filenames`, or maybe both in combination, are triggering the `non-trivial` pathway.

I think it's definitely worth looking into this gap in revision gating's coverage, but for the sake of forward progress on this issue, I suggest that we take the approach of having a slight few more candidates and/or drop the type constraints of the new candidates.

I'll push a revision that matches the above sentiment shortly.

ab5tract:
@ugexe You mentioned in `#raku-dev` that you have some concerns about compatibility related to the change to `IO::Path.chmod`. I added a `die`, but there is a `CATCH` block. So the only thing that should change (besides fixing the silent no-op on Windows) is the precise wording of the exception that gets stored into the `$!os-error` field of an `X::IO::Chown` exception.

I am not sure that I see this as a breaking change because AFAIK we never made assurances that  exception messages are untouchable (sort of the opposite, as we constantly strive to improve LTA exception messages).

I can add a check against `$*DISTRO.is-win` but that feels like overkill.

ab5tract (approved):
I am me and I approve this message.

PR #5501, opened 2023-12-19
Complete generic instantiation 5

vrurg:
In this episode two problems are taken care of:

First, MoarVM was still crashing with `SIGSEGV` on Apple Silicon, despite all the previous attempt to get around the bug; with this PR no crash has been observed so far.

Second, this PR changes the approach to re-parameterization of `Array` and `Hash` classes. Say, we have a generic parameterization of `Array[T]`. Re-parameterizing it an instantiation of `T` would end up with something like `Array[T][Str]`, but the CORE was stripping off any mixins, revealing the original `Array` class and parameterized over it. There is a problem with this approach: if the generic was parameterized over a mixin, giving us something like `Array+{MyRole}[T]` then re-parameterizing resulted in a plain `Array[Str]`. This is barely something user would be happy too see happening.

The new approach is to strip off only _generic_ mixins. This would effectively remove `Array::Typed` (or `Hash` counterparts of it) but preserve user's non-generic mixin. Apparently, those applied to the parameterization (`Array::Typed[Str]+{MyRole}` or alike) would be lost too. `MyRole` must take own care of the instantiation if needs to be there no matter what because there is no single common way to resolve this conflict of interests.

### DIFF
--- a/src/Perl6/Metamodel/ClassHOW.nqp
+++ b/src/Perl6/Metamodel/ClassHOW.nqp
@@ -437,9 +437,9 @@ class Perl6::Metamodel::ClassHOW
                 #   whatever we determine here is type's ultimate archetypes.
                 # * Also, since we've taken care of a concrete object case then here 'is-generic' is invoked on the type
                 #   itself, not an instance of it.
-                $atype := $can-is-generic && $obj.is-generic ?? $archetypes-g !! $archetypes-ng;
                 nqp::getattr($how, Perl6::Metamodel::ClassHOW, '$!archt-lock').protect({
                     nqp::scwbdisable();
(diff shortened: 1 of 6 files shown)


lizmat:
@jdv I assume we want this merged before 2023.12?   If so, I guess we need another Blin run?

vrurg:
We will need the run. As a matter of fact, re-parameterization of arrays and hashes, the way it's been done previously, was also triggering or causing a bug resulting a code object lost and MoarVM attempting to call a `VMNull`.

lizmat:
@vrurg if the VMNull issue is fixed, then that would make me *very* happy!

jdv:
its a little late to be pushing big seeming things but yeah.  i can do a run tonight or tomorrow.

Issue #5504, opened 2023-12-27
Distribution::Resource.Str should be deprecated


ugexe:
## The Problem

`%?RESOURCES<foo>.Str` can often confuse developers by Working As Expected *sometimes*, but not always. For example when installing a module using `CompUnit::Repository::Staging` the target of `%?RESOURCES<foo>` will be one thing when installing into that staging repository and another thing after that staging repository has been copied. If a developer calls `.Str` on `%?RESOURCES<foo>` at compile time then it is possible the path it shows will incorrect after copying the staging repository to a repository such as `site`. This used to not be so apparent prior to zef starting to use `CompUnit::Repository::Staging`, and hence why some instances of using `.Str` on `%?RESOURCES` this way likely exist.

see: https://github.com/ugexe/zef/issues/546

To prevent developers from making this same mistake we should deprecate [Distribution::Resource.Str](https://github.com/rakudo/rakudo/blob/f72c3c019585f00c6bf894d8c5d07367d906c647/src/core.c/Distribution/Resource.rakumod#L25-L27) and maybe suggest they use `%?RESOURCES<foo>` instead of `%?RESOURCES<foo>.Str`. It is probably ok to deprecate it globally and not just for `6.e`, since using `%?RESOURCES<foo>.Str` is problematic for 6.c and 6.d as well. [Here](https://github.com/rakudo/rakudo/blob/f72c3c019585f00c6bf894d8c5d07367d906c647/src/core.c/CompUnit/Repository/FileSystem.rakumod#L340-L347) is an example of how to deprecate such a function.

PR #5505, opened 2023-12-27
Deprecates Distribution::Resource.Str

opoku:
The concern with this method is described in #5504. It may unexpectedly return inconsistent values depending on the compilation stage.

### DIFF
--- a/src/core.c/Distribution/Resource.rakumod
+++ b/src/core.c/Distribution/Resource.rakumod
@@ -22,7 +22,9 @@ class Distribution::Resource {
     }
 
     # delegate appropriate IO::Path methods to the resource IO::Path object
-    multi method Str(::?CLASS:D: |c) {
+    multi method Str(::?CLASS:D: |c)
+      is DEPRECATED('%?RESOURCES<key> directly instead')
+    {
         self.IO.Str(|c)
     }
     multi method gist(::?CLASS:D: |c) {


src/core.c/Distribution/Resource.rakumod
```diff
@@ -22,7 +22,7 @@ class Distribution::Resource {
     }
 
     # delegate appropriate IO::Path methods to the resource IO::Path object
-    multi method Str(::?CLASS:D: |c) {
+    multi method Str(::?CLASS:D: |c) is DEPRECATED('%?RESOURCES<key> directly instead') {
```
ugexe, line 25: A potential improvement might be showing the key being accessed. For example: `is DEPRECATED(sprintf '%?RESOURCES<%s> directly instead', $!key)`

opoku: It looks like this didn't quite work as is. I'm not sure how to access the instances attribute while defining a trait.

ugexe: Thanks for trying my suggestion. After checking this out I think your original commit is better since it (the trait specifically) will show the liner numbers of all the callers, whereas the code I suggested shows the line number of the core rakudo code where `DEPRECATED(...)` is called. If you could `git reset --hard 721e9afc088d714d3ef242c1afd7cde645b0bbd3` or just add a commit going back to your original code I'll accept and merge.

opoku: Done! With some formatting changes. Thanks for the explanation!

ugexe (approved):

PR #5510, opened 2024-01-04
Add ValueList as a 6.e feature in core

lizmat:
Raku needs a value-type object for lists.  This copies the ValueList ecosystem module https://raku.land/zef:lizmat/ValueList verbatim and makes it part of 6.e.  It does *NOT* provide any additional syntactic sugar to easily create ValueLists, so ValueList.new(...) is the way to go now.

### DIFF
--- a/src/core.e/ValueList.rakumod
+++ b/src/core.e/ValueList.rakumod
@@ -1,0 +1,91 @@
+my class ValueList
+  is IterationBuffer   # get some low level functionality for free
+  does Positional      # so we can bind into arrays
+  does Iterable        # so it iterates automagically
+  is repr('VMArray')   # needed to get nqp:: ops to work on self
+{
+
+    multi method WHICH(ValueList:D:) {
+        nqp::box_s(
+          nqp::concat(
+            nqp::if(
+              nqp::eqaddr(self.WHAT,ValueList),
+              'ValueList|',
+              nqp::concat(self.^name,'|')
+            ),
+            nqp::sha1(
+              nqp::join(
+                '|',
+                nqp::stmts(  # cannot use native str arrays early in setting
+                  (my $strings  := nqp::list_s),
+                  (my int $i     = -1),
+                  (my int $elems = nqp::elems(self)),
+                  nqp::while(
+                    nqp::islt_i(($i = nqp::add_i($i,1)),$elems),
+                    nqp::push_s($strings,nqp::atpos(self,$i).Str)
+                  ),
+                  $strings
+                )
+              )
+            )
+          ),
+          ValueObjAt
+        )
+    }
+
+    proto method new(|) {*}
+    multi method new(ValueList: @args) {
+        nqp::create(self)!SET-SELF: @args.iterator
+    }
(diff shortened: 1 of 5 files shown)


lizmat:
As syntactic sugar for `ValueList`s, I would suggest `⁅ ⁆`, basically:

    sub circumfix:<⁅ ⁆>(\a) { ValueList.new(a) }

Allowing `⁅1,2,3⁆` as a value type.

vrurg (approved):

vrurg:
I don't think discussing a new operator here would be a good idea. And I anticipate a dispute because `⁅ ⁆` pair has two major drawbacks.

First, it's a Unicode which is barely easy to use. A no good thing for something that is likely to be used on a very regular basis.

Second, the pair resembles array operator which is, apparently, has quite the opposite semantics.

lizmat:
> I don't think discussing a new operator here would be a good idea

I see it as brain-storming at this point, as any real discussion will be moot without this PR (or a functional equivalent thereof) being merged.

>  anticipate a dispute because ⁅ ⁆ pair has two major drawbacks

Perhaps `((...))` would be good syntactic sugar.  With `⸨...⸩` as the Unicode equivalents.

```raku
use ValueList;
sub circumfix:<(( ))>(\a) { ValueList.new(a) };
dd ((1,2,3)).WHICH;   # ValueObjAt.new("ValueList|C97F62DE1BA14379F6F017F3DFCBF16FB199D041")
dd Map.new(("a",42)); # Map.new((:a(42)))
```

TimToady:
((1 + 2) * 3)

lizmat:
```raku
use ValueList;
sub circumfix:<(( ))>(\a) { ValueList.new(a) };
say ((1 + 2) * 3);
```
```
Unable to parse expression in circumfix:sym<(( ))>; couldn't find final '))' (corresponding starter was at line 6)
at ...:3
------> say ((1+2⏏) * 3);
```
meh  :-)     So `((...))` it won't be.

vrurg:
I can't concentrate on this matter deep enough to come out with a worthwhile proposal, but what I currently see is that the entire field of de-/containerization is not covered by Raku syntax. Basically, all we currently have are ops for itemizing with `$` and deconting with `<>`, but both are targeting individual objects and do not propagate to collection types. Besides, I remember some complains about the weirdness of `<>` operator as its semantics has nothing in common with `<key>` hash access form.

The direction I'm looking at is based on `<>` (until a better replacement is found). What we can do is to use the decont operator (whatever it is) to mark the existing collection creation circumfixes in order to produce value counterparts of hashes and lists. Something like:

- ValueList: `(<> $a, $b, $c)`, unicode: `⦅ $a, $b, $c ⦆`
- ValueMap: `{<> :$a, :$b, :$c }`, `%(<>  :$a, :$b, :$c )`, unicode: `⦃ :$a, :$b, :$c ⦄`

Apparently, `%(<>` is gross, but if decont operator is considered as a modifier then things are looking less horrible.

lizmat:
Closing this now to be discussed again after RakuAST is default

PR #5511, opened 2024-01-06
RakuAST: Fix deparsing of RakuAST::MetaInfix::Negate

Nahuel2998:
Fixes the deparsing of `RakuAST::MetaInfix::Negate` ( by changing the side `!` is being put at ) and adds a test for it.

--- 
( First time contributing so sorry if I failed to follow some protocol )

I was playing with RakuAST and I came across this:
```
[0] > 1 !== 2
True
[1] > '1 !== 2'.AST.EVAL
True
[2] > '1 !== 2'.AST.DEPARSE
1 ==! 2

[3] > '1 !== 2'.AST.DEPARSE.EVAL
False
[4] > say .AST eqv .AST.DEPARSE.AST given '1 !== 2'
False
```

I checked the code and it seems like the `!` was being put on the wrong side on `deparse(RakuAST::MetaInfix::Negate:D $ast --> Str:D)`, I just flipped it.

I checked the `t/12-rakuast/meta-operators.rakutest` tests and found there was none for `RakuAST::MetaInfix::Negate`.  
I added one with a comparison that fails both due to deparsing not being equal and the expressions not being equivalent ( `0 !== 1` and `0 ==! 1` both evaluate to `True`, but `1 !== 2` and `1 ==! 2` differ ).

### DIFF
--- a/src/core.c/RakuAST/Deparse.rakumod
+++ b/src/core.c/RakuAST/Deparse.rakumod
@@ -1007,7 +1007,7 @@ CODE
     }
 
     multi method deparse(RakuAST::MetaInfix::Negate:D $ast --> Str:D) {
-        self.syn-infix(self.deparse($ast.infix) ~ '!')
+        self.syn-infix('!' ~ self.deparse($ast.infix))
     }
 
     multi method deparse(RakuAST::MetaInfix::Reverse:D $ast --> Str:D) {
--- a/t/12-rakuast/meta-operators.rakutest
+++ b/t/12-rakuast/meta-operators.rakutest
@@ -1,7 +1,7 @@
 use v6.e.PREVIEW;
 use Test;
 
-plan 9;
+plan 10;
 
 my $ast;
 my $deparsed;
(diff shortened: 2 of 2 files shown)


src/core.c/RakuAST/Deparse.rakumod
```diff
@@ -1007,7 +1007,7 @@ CODE
     }
 
     multi method deparse(RakuAST::MetaInfix::Negate:D $ast --> Str:D) {
-        self.syn-infix(self.deparse($ast.infix) ~ '!')
+        self.syn-infix('!' ~ self.deparse($ast.infix))
```
lizmat, line 1010: To make localization work correctly, the '!' should be *before* the `self.syn-infix`

Nahuel2998:
```
[0] > '1 !(cont) 2'.AST.DEPARSE: <DE>
1 !(enthält) 2
```
Seems like it works anyway due to `self.deparse($ast.infix)` already handling the localization ( which also seems to be the case in the methods below )
Though since `'!'` isn't localized I suppose it could be `self.hsyn('infix', '!' ~ self.deparse($ast.infix))`
Or rather `self.hsyn('infix', '!' ~ self.xsyn('infix', self.deparse($ast.infix)))` like in the method right above ( But I feel like the call to `xsyn` there isn't really needed? )

In short, isn't `deparse(RakuAST::Infix:D $ast --> Str:D)` (line 963) already calling `self.syn-infix` (and in turn `xsyn`) for the part that should be localized? And if it isn't, isn't it also broken in the methods below?
( Apologies if this is nonsense, I'm trying to understand )

lizmat: No apologies needed.  I'm very grateful for these extra pair of eyes on this code, as there seem to be a number of thinkos embedded here.

Investigating...

lizmat:
Thanks for the nudge in this PR.  As it turned out, there were multiple thinkos in this area, which I believe to be fixed with https://github.com/rakudo/rakudo/commit/3c7057c0f0 .

Closing this PR now, but please don't let that discourage you creating more PRs and/or issues!   Keep up the good work!

lizmat:
Also: I've cherry-picked the test addition into the main branch!

Nahuel2998:
Happy to have helped! :-)

PR #5513, opened 2024-01-14
Add Mu.Callable($method) "coercer"

lizmat:
The idea being that *Raku* should provide a HLL interface to obtain a Callable object for the given method name on the invocant.

    $ raku -e 'say Int.Callable("Str")(687, :subscript)'
    ₆₈₇

In place of the .^find_method logic, which is *not* Raku specific according to the documentation:

  https://docs.raku.org/type/Metamodel/DefiniteHOW#method_find_method

EDIT: change `Sub` to `Callable`

### DIFF
--- a/src/core.c/Mu.rakumod
+++ b/src/core.c/Mu.rakumod
@@ -118,8 +118,13 @@ my class Mu { # declared in BOOTSTRAP
         self.HOW.set_why($why);
     }
 
-    method Callable(str $method) is pure { 
-        nqp::ifnull(nqp::tryfindmethod(self,$method),Nil)
+    method Callable(str $method) { 
+        nqp::ifnull(
+          nqp::tryfindmethod(self,$method),
+          X::Method::NotFound.new(
+            :invocant(self), :typename(self.^name), :$method
+          ).Failure
+        )
     }
 
     proto method Bool() {*}
--- a/src/core.c/Mu.rakumod
+++ b/src/core.c/Mu.rakumod
@@ -118,7 +118,7 @@ my class Mu { # declared in BOOTSTRAP
         self.HOW.set_why($why);
     }
 
-    method Sub(str $method) is pure { 
+    method Callable(str $method) is pure { 
         nqp::ifnull(nqp::tryfindmethod(self,$method),Nil)
     }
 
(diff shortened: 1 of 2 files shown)


jnthn:
I'd count `.^find_method` as a specified part of the language. Probably much of the MOP can be considered to be by this point; anything that's had a consistent interface for the last 10 years can probably be counted on as a durable design!

vrurg:
I would call it `Method` because `Sub` is confusing about what's really happening.

lizmat:
On quick glancing, these appear the meta-model methods called in roast:
```
.^add_attribute
.^add_fallback
.^add_method
.^add_multi_method
.^attributes
.^auth
.^can
.^candidates
.^coerce
.^compose
.^concretization
.^curried_role
.^does
.^enum_value_list
.^find_method
.^isa
.^language
.^lookup
.^methods
.^mro
.^mro_unhidden
.^name
.^num
.^parents
.^pun
.^roles
.^set_name
.^shortname
.^tryit
.^ver

vrurg:
> There are indeed quite of a few occurrences of `find_method` in roast. So I'm starting to wonder whether the documentation is incorrect in stating it is Rakudo specific.

It's not that much the method itself is specific but the interface it provides. For example, `:no_fallback` is better be renamed to `:no-fallback` to follow the general naming rules.

lizmat:
Well, there's that indeed.  Also I think we need to get rid of the "Perl6" in the naming of the MOP.

vrurg:
BTW, did you see my other comments about `is pure` and return value?

raiph:
A cache on `Callable` is slightly surprising to me. Two things come to mind:

1. Are all Raku types such that the only changes that can be made to them (changing attributes, routines, ...) are such that all existing (cached) callables are guaranteed to still exist and be callable? For example, are use of MOP calls and features like `augment` constrained that way?

2. Aiui nqp and/or MoarVM are currently responsible for most or all of the caching related to things like callables. If I'm right about that, is the idea here to eliminate their caches for Raku, or will there be two levels of caching, or...?

vrurg:
It looks like links to review comments are different from plain commits. I've updated the link above to the working one: https://github.com/rakudo/rakudo/pull/5513/files/b23f0e5fe68fbdc5fea8823428445152cc00e6ea

vrurg (approved):

PR #5516, opened 2024-01-22
Compiler.nqp: Add $rakudo-flavor

AntonOks:
see https://github.com/rakudo/star/pull/198 for the related discussion

### DIFF
--- a/src/Perl6/Compiler.nqp
+++ b/src/Perl6/Compiler.nqp
@@ -140,11 +140,13 @@ class Perl6::Compiler is HLL::Compiler {
     method lvs() { LanguageVersionServices }
 
     method version_string(:$shorten-versions, :$no-unicode) {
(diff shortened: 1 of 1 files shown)


AntonOks:
@lizmat @coke

see https://github.com/rakudo/star/pull/198 for the related discussion
please review or assign to someone else

THANK YOU very much for your help & guidance !

lizmat:
Using an environment variable means it requires it to be set in the user's environment at runtime?  Which would affect other rakudo installs as well?

AntonOks:
> Using an environment variable means it requires it to be set in the user's environment at runtime? Which would affect other rakudo installs as well?

Well, yes, an optional new environment variable, which can be set i.e. by various Rakudo installers like with the MSI one under Windows.
This env var could be similar to https://docs.raku.org/programs/03-environment-variables

PR #5519, opened 2024-01-27
Ensure `%!mro` is not modified concurrently

vrurg:
Introduce lock-protection for the attribute and atomize operations we perform with it.

### DIFF
--- a/src/Perl6/Metamodel/C3MRO.nqp
+++ b/src/Perl6/Metamodel/C3MRO.nqp
@@ -1,10 +1,15 @@
 role Perl6::Metamodel::C3MRO {
     # Storage of the MRO.
     has %!mro;
+    has $!mro_lock;
+
+    method setup_mro_engine() {
+        $!mro_lock := NQPLock.new();
+    }
 
     # Computes C3 MRO.
     method compute_mro($class) {
-        %!mro := nqp::hash(
+        my %mro := nqp::hash(
             'all', nqp::hash(
                 'no_roles', nqp::list(),    # MRO with roles excluded
                 'all', nqp::list(),         # MRO with roles as parametric groups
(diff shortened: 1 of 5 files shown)


src/Perl6/Metamodel/C3MRO.nqp
```diff
@@ -1,10 +1,15 @@
 role Perl6::Metamodel::C3MRO {
     # Storage of the MRO.
     has %!mro;
+    has $!mro_lock;
+
+    method setup_mro_engine() {
+        $!mro_lock := NQPLock.new();
+    }
```
lizmat, line 8: If this returns `self`, then all all the `.new` methods would just have to add a call to `setup_mro_engine`, and not need a local `$obj`.

src/Perl6/Metamodel/C3MRO.nqp
```diff
@@ -97,18 +102,20 @@ role Perl6::Metamodel::C3MRO {
...
-                'no_roles', @no_roles,
-            ),
-            'unhidden', nqp::hash(
-                'all', @unhidden_all,
-                'all_conc', @unhidden_all_conc,
-                'no_roles', @unhidden_no_roles,
-            ),
-        );
+        $!mro_lock.protect({
+            %!mro := nqp::hash(
```
lizmat, line 106: Does the hash need to be initialized inside the protected block?   Or could it be initialized outside in a `my %mro` and then just have the attribute binding happen inside the protected block?

src/Perl6/Metamodel/C3MRO.nqp
vrurg: True and I was considering it. But there is better solution, though it involves NQP patching. NQP build plan already supports `BUILD` method. But since `BUILD` replaces the entire default build plan, adding support for `TWEAK` would make it a good solution. In this case a few locations in the Metamodel could be replaced with a series of `self.setup_<something>` calls.

Aside of this, the body of `new` is already a bit hard to read at the first glance. Adding a method call wouldn't be beneficial.

src/Perl6/Metamodel/C3MRO.nqp
vrurg: The entire point of the patch is this protection. It's not about the hash alone but about attribute read/write is not a thread-safe op.

The initialization can be split in two steps, but what would be the benefit of it? Basically, two threads updating the same MRO is a rather rare case. And it's most certainly isn't involved in a hot path. But from the perspective of readability I think it is better this way.

src/Perl6/Metamodel/C3MRO.nqp
MasterDuke17: Why just not `has $!mro_lock := NQPLock.new();` since it looks like you call this method in all the `.new()`s anyway?

vrurg: @MasterDuke17 Maintainability.

MasterDuke17: > @MasterDuke17 Maintainability.

How? Doing it in one place instead of several seems more maintainable to me. Unless you think a bunch of new Perl6::Metamodel::C3MRO subclasses aren't going to want to protect the hash?

vrurg: Oh, I misread your initial comment. It's NQP, it doesn't support attribute initializers. Wish it would, but if we go this way we're going to end up with self-bootstrapping Rakudo. Oh, wait, why not? ;)

PR #5523, opened 2024-02-10
Avoid leaking repo.lock handle 

ugexe:
The repo.lock file would be opened when installing a distribution, but would never get closed. On windows this could manifest as an error saying it can't delete repo.lock because the resource is busy. This updates CURI.install to close the repo.lock file when leaving the function.

### DIFF
--- a/src/core.c/CompUnit/Repository/Installation.rakumod
+++ b/src/core.c/CompUnit/Repository/Installation.rakumod
@@ -231,6 +231,7 @@ sub MAIN(*@, *%) {
         my @*MODULES;
         my $path   = self!writeable-path or die "No writeable path found, $.prefix not writeable";
         my $lock = $.prefix.add('repo.lock').open(:create, :w);
+        LEAVE $lock.close;
         $lock.lock;
 
         my $version = self!repository-version;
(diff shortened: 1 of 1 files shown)


niner (approved):

src/core.c/CompUnit/Repository/Installation.rakumod
```diff
@@ -231,15 +231,16 @@ sub MAIN(*@, *%) {
         my @*MODULES;
         my $path   = self!writeable-path or die "No writeable path found, $.prefix not writeable";
         my $lock = $.prefix.add('repo.lock').open(:create, :w);
+        LEAVE $lock.close;
         $lock.lock;
+        LEAVE $lock.unlock;
```
lizmat, line 236: That's not how `LEAVE` phasers work.  The `LEAVE` phaser will be active for the **entire** block.  So it *will* fire with the `die` in line 232.  And thus cause an execution error because `Any.close` does not work.

Fixed with https://github.com/rakudo/rakudo/commit/7ca96bbd90

ugexe: That commit appears to have broken installation on windows: https://dev.azure.com/Rakudo/rakudo/_build/results?buildId=5249&view=logs&j=70f6b8a1-f3cf-5727-bd79-d08ddb1caf67&t=cd681118-a333-54f7-e221-0bc46b61af45&l=194

PR #5526, opened 2024-02-15
Provide native int coercers to `Int`/`Cool`

lizmat:
Specifically:
- .int, .int64, .int32, .int16, .int8
- .uint, .uint64, .uint32, .uint16, .uint8, .byte

Note that this is just to provide the API.  The implementation is definitely hacky: this should probably be handled in the grammar and in RakuAST it probably will.

This makes the following code work:

    say 256.uint8;  # 0
    say 255.int8;   # -1

And:

    multi a(uint8 $) { dd }
    multi a(Int:D $) { dd }
    a 42.uint8;  # sub a(uint8)

Note however that does **not** make multiple dispatch on different native sizes work:

    multi a(uint8  $) { dd }
    multi a(uint16 $) { dd }
    a 42.uint8;
    # Ambiguous call to 'a(Int)'; these signatures all match:
    #   (uint8 $)
    #   (uint16 $)

This will need additional work in the dispatcher logic.

### DIFF
--- a/src/core.c/Cool.rakumod
+++ b/src/core.c/Cool.rakumod
@@ -467,6 +467,21 @@ my class Cool { # declared in BOOTSTRAP
     }
 
     method Version() { self.Str.Version }
+
+    method  int(-->  int) is raw { my  int $a = self.Int; $a }
+    method uint(--> uint) is raw { my uint $a = self.Int; $a }
+
+    method int64(--> int64) is raw { my int64 $a = self.Int; $a }
+    method int32(--> int32) is raw { my int32 $a = self.Int; $a }
+    method int16(--> int16) is raw { my int16 $a = self.Int; $a }
+    method int8( --> int8 ) is raw { my int8  $a = self.Int; $a }
+
+    method uint64(--> uint64) is raw { my uint64 $a = self.Int; $a }
+    method uint32(--> uint32) is raw { my uint32 $a = self.Int; $a }
+    method uint16(--> uint16) is raw { my uint16 $a = self.Int; $a }
+    method uint8( --> uint8 ) is raw { my uint8  $a = self.Int; $a }
+    method byte(  --> byte  ) is raw { my byte   $a = self.Int; $a }
+
 }
 Metamodel::ClassHOW.exclude_parent(Cool);
 
--- a/src/core.c/Int.rakumod
+++ b/src/core.c/Int.rakumod
@@ -272,6 +272,20 @@ my class Int does Real { # declared in BOOTSTRAP
             }
         }
     }
+
+    method  int(-->  int) is raw { my  int $a = self; $a }
+    method uint(--> uint) is raw { my uint $a = self; $a }
+
+    method int64(--> int64) is raw { my int64 $a = self; $a }
+    method int32(--> int32) is raw { my int32 $a = self; $a }
+    method int16(--> int16) is raw { my int16 $a = self; $a }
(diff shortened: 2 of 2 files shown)


librasteve:
Nice!

I do not think that multis on different native sizes should ever be supported, full raku multi power should only be expected with full types.

ab5tract:
> I do not think that multis on different native sizes should ever be supported, full raku multi power should only be expected with full types.

If it is possible, I don't see why we wouldn't support it.

Am I missing something?

librasteve:
@ab5tract ... 

You raise a good question. I will try to elaborate my perspective, which comes more from a consumer than an implementor of raku.

Sorry about the TL;DR.

----

My impression of the [Numeric](https://docs.raku.org/language/numerics) types in raku is that this is a well thought through design that provides many features:

1. An open & extensible class / role [system](https://docs.raku.org/type/Numeric#typegraphrelations)
In contrast to the number tower patterns of other languages, raku pays attention to the computer science aspects of numerics so that (eg) an Int is not a child of a Num. There is also a well defined Infectiousness for operations.

2. Rats and BigInts are first class citizens
Literals help to promote use of Rats over Nums and there are smooth boundaries (```my $x = 1/0; $x.WHAT;  #(Rat)```, ``` $x ~~ Inf;  #True```), there is no Num32 or Int32 since for the most part modern CPUs render these distinctions moot for numerical operations (it is different for advanced stuff like crypto or bitwise operations).

3. Advanced behaviours
* Numerics can take Nil values and when Nil assigned they offer _is default_ behaviour
* ±Inf semantics is aligned between Numeric, Range and Sequences (and can control lazy/eager behaviour)
* Overflow / underflow is well defined
* Operations are partitioned into rational and irrational
* NaN is part of the high level language
* Range semantics (what is the result of a cats ear ```^10e0``` on a Num?)

----

I am skeptical that this rich range of behaviours can be smoothly extended to native types - imo ```natives can never be Numerics```. My view of the current design is that is it a carefully curated subset of all the possible components and that natives were deliberately excluded from Numerics since they would just add more overhead, but not add any cognitive value for most coders in most situations.

So, since programming raku relies on the language having principles (e.g. an irrational operation like sqrt will give me a Num) I think that there is a principle of ```if its a Numeric it does all of that (^^^), but if its a native it doesn't do any of it```. My case is that we should maintain this as a principle unless the practice makes it too hard.

The implication is that this would keep natives a "specialist subject" and that any code may be machine dependent (as it already is), but that natives give you advantages in size, speed and FFI interop that may be a good trade for your needs. It would also often be a good idea for you to have an abstraction between your native world and the raku Numeric world - such as coercions to/from Int if you are playing with uint256. But maybe you would not want or need this for your crypto lib. Then you can hook in the raku coercion machine and get a smooth interaction with (I don't know) trig functions or whatever you want.

I also get the feeling that keeping Numerics and natives distinct and not taking out more tech debt in trying to make native do the Numeric thing a bit more is the lazy (as in virtuous) path.

If we decide that then it should be possible to document with something like _if your typename start with a lower case letter then smartmatch and multi-despatch will not follow the standard raku typesystem  semantics_

----

There have been some conversations on IRC/Discord about smartmatch to/from natives:
* ```uint ~~ Int; #True```
* ```Int ~~ unit; #False```

But it seems that the smartmatch behaviours are not aligned to the multi despatch behaviours for natives. (This is the tail wagging the dog a bit, but I think it shows some difficulty in implementation in that natives are a MoarVM thing and Numerics are a Raku thing).

PR #5543, opened 2024-03-30
Add multi disambiguation for positional scalars

ab5tract:
Until now there has been no way to capture the use of `$[]`
(as opposed to `[]`) by the caller of a multi routine.

For example:
```
    multi sub($a where *.can("elems") == 2) {...}
    multi sub(@a) {...}

    a $[1,2]
```

In this case, the `@a` candidate always wins (the (ugly) where clause is never
tested).

Or consider:
```
    multi sub(Array $a) {...}
    multi sub(@a) {...}

    a $[1,2]; # goes to first
    a [1,2];  # also goes to first
```

Finally:
```
    multi sub(Positional $a) {...}
    multi sub(@a) {...}

    a $[1,2]; # failure due to ambiguous dispatch
```

This patch addresses the issue by removing the ambiguity from this
third example. By specifying a scalar argument to be of type Postional,
it will "win" against an @ argument in the same place in a competing
multi signature.

By taking this approach, it is hoped that the fallout to the ecosystem
will be minimal as it is unlikely that anyone has intentionally shipped
multis that fail to dispatch due to ambiguity.

Note: This patch does not address signatures where the positional scalar
utilizes a where-guard. Since the where-guard already breaks the tie that
is at the heart of this patch, it is much more likely that such a change
would adversely affect the ecosystem.

### DIFF
--- a/src/vm/moar/dispatchers.nqp
+++ b/src/vm/moar/dispatchers.nqp
@@ -1989,7 +1989,7 @@ nqp::bindhllsym('Raku', 'PROXY-READERS', $PROXY-READERS);
 
 #- raku-multi-core -------------------------------------------------------------
 # The core of multi dispatch. Once we are here, either there was a simple
-# proto that we don't need to inovke, or we already did invoke the proto.
+# proto that we don't need to invoke, or we already did invoke the proto.
 #
 # Multiple dispatch is relatively complex in the most general case. However,
 # the most common case by far consists of:
@@ -2038,7 +2038,7 @@ my class MultiDispatchCall {
     method next() { $!next }
     method trivial() { nqp::istype($!next,MultiDispatchEnd) }
     method debug() {
-        "Candidate " ~ $!candidate.signature.raku ~ "\n" ~ $!next.debug
+        "Candidate " ~ $!candidate.signature.raku ~ ($!next ?? "\n" ~ $!next.debug !! '')
     }
 }
 
@@ -2048,7 +2048,7 @@ my class MultiDispatchTry is MultiDispatchCall {
     method trivial() { 0 }
     method debug() {
         "Try candidate " ~ self.candidate.signature.raku ~ "\n"
-          ~ self.next.debug
+          ~ (self.next ?? self.next.debug !! '')
     }
 }
 
@@ -2172,6 +2172,7 @@ sub raku-multi-plan(
     # references don't count; we know the native types they shall match up with
     # and don't need to dereference them
     my $non-scalar := nqp::list_i();
+    my $capture-pos-scalar := nqp::list_i();
     my int $i;
     while $i < $num_args {
         unless nqp::captureposprimspec($capture, $i) {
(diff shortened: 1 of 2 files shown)


src/core.c/Array.rakumod
```diff
@@ -1178,6 +1229,7 @@ my class Array { # declared in BOOTSTRAP
     multi method splice(Array:D:
       Int:D $offset, Whatever $, @new
     --> Array:D) {
+        say "in the array candidate";
```
lizmat, line 1232: Debug leftover?

src/vm/moar/dispatchers.nqp
```diff
@@ -2574,6 +2599,69 @@ sub raku-multi-plan(
...
+            while $w < $total-capture-scalar-pos-args {
+                $perfect-match := $perfect-match
+                    && nqp::iseq_i(nqp::atpos_i($capture-scalar-pos-args, $w), nqp::atpos_i($cand-scalar-pos-params, $w));
+                ++$w;
+            }
+            return [ %current ] if $perfect-match;
+        }
+        ++$i;
+    }
+    return [ %current ];
```
lizmat, line 2662: There's an ever so slight overhead in using `return`: don't use it in hot code unless it's needed

src/core.c/Array.rakumod
```diff
@@ -1093,6 +1093,57 @@ my class Array { # declared in BOOTSTRAP
                :range("0..^{self.elems - $offset}")
              ).throw
     }
+    #------ splice(offset,size,positional scalar) candidates
```
lizmat, line 1096: would prefer the `Array.splice` changes to be in a separate PR, after it has become clear that the `Positional $foo` feature works and has spectests.  This would allow for better bisectability should problems with `Array.splice` becime clear after a blin run.

src/core.c/Array.rakumod
ab5tract: Good catch. I thought I nuked all those.

src/core.c/Array.rakumod
ab5tract: They are already separate commits, but if you feel a second PR is better I don't have a problem with that.

src/vm/moar/dispatchers.nqp
ab5tract: Fixed.

src/core.c/Array.rakumod
ab5tract: The PR is now adjusted to only address the positional scalar disambiguation.

lizmat:
A more general comment: without identical changes to the non-dispatch case, would this create different semantics on the JVM backend?

ab5tract:
> A more general comment: without identical changes to the non-dispatch case, would this create different semantics on the JVM backend?

Yes, unfortunately. Good pointing this out.

ab5tract:
Legacy dispatch now supports the same proposed semantics.

PR #5544, opened 2024-03-31
Unbreak the JVM build

MasterDuke17:
A partial revert of 196fbee. No idea why it breaks the JVM build and why this unbreaks it.


lizmat:
Possibly because it's too soon in the JVM build to have multis?

What happens on the Moar build if you remove the `multi` part?

niner:
We can easily rename that new multi candidate and turn it into an only. Using multis just felt more Rakuish to me. I don't expect that new candidate to be used by anything but the compiler itself.

Issue #5500, opened 2023-12-18
Array.splice insists on flattening the replacement


DarthGandalf:
## The Problem
I have a 2d array (array of arrays), and I want to insert a new row to it.

```raku
my @map = [[11, 12], [31,32]];
my @newrow = [21, 22];
@map.splice(1, 0, @newrow);
```

## Expected Behavior
I want `@map` to look like this:
```
[
    [11, 12],
    [21, 22],
    [31, 32]
]
```

## Actual Behavior
`@map` will now look like this:
```
[
    [11, 12],
    21,
    22,
    [31, 32]
]
```

I tried to pass `[[[[@newrow]]]]` to .splice(), tried to pass `$@newrow`, without any difference. The only workaround I found is this:
```raku
@map.splice(1, 0, 'dummy');
@map[1] = @newrow;
```

## Environment
* Operating system: Gentoo Linux, amd64
* Compiler version (`rakudo -v` or `raku -v`): v2023.10

ab5tract:
It took a minute (longer than it should have, in retrospect) but I think I have a way to resolve this without breaking any existing `splice` contracts.

```
my @a = [1, 2, 3];
@a.splice(1, 1, [2, 2]) :preserve;
dd @a; # [1, [2, 2], 3
```

ab5tract:
It took a bit longer than expected, but this is finally available in `HEAD`:

   > use v6.e.PREVIEW; # only works in versions 6.e+
   > my @a = 1,2,3;
   > @a.splice(1,1,$[7,8])
   > dd @a
   # [1,[7,8],3]

PR #5549, opened 2024-04-06
Add dispatch disambiguation via 'is item' parameter trait

ab5tract:
This allows for multi routines to distinguish between itemized arguments
via a new `is item` parameter trait.

For the sake of simplicity, the parameter trait does not add any new
ParamTypeChecks[^1] to the parameter. Instead it is only used as a marker
for disambiguation in multiple dispatch:

```
    multi sub a(@a is item) { "itemized!" }
    multi sub a(@a) { "array" }

    a [1,2];  # "array"
    a $[1,2]; # "itemized!"
```

Note that this only makes sense for `@` (positional) and `%` (associative) 
sigilled parameters. Raw, unsigilled parameters are also currently included,
but only if the parameter type is specified to something that does `Positional`
or `Associative`.

[^1]:Taking this approach is complicated by the fact that there appears to be
some decontainerization of arguments via the `callstatic` and `callmethod`
ops: a `ParamTypeCheck` which calls `iscont` on the lowered param always
fails for `$[]` and `${}` arguments. Further research is required.

### DIFF
--- a/src/Perl6/Actions.nqp
+++ b/src/Perl6/Actions.nqp
@@ -5641,6 +5641,23 @@ class Perl6::Actions is HLL::Actions does STDActions {
 
         # Stash any traits.
         %param_info<traits> := $<trait>;
+        if my int $num_traits := nqp::elems(%param_info<traits>) {
+            my int $z;
+            while !%param_info<is_item> && $z < $num_traits {
+                %param_info<is_item> := 1
+                    if nqp::index(%param_info<traits>[$z], 'item') > 0;
+                ++$z;
+            }
+            if %param_info<is_item> && (%param_info<sigil> eq '$' || %param_info<sigil> eq '&') {
+                $/.typed_sorry('X::Comp::Trait::Invalid',
+                    name        => %param_info<variable_name>,
+                    reason      => "only '\@' or '\%' sigiled parameters can be constrained to itemized arguments",
+                    declaring   => 'parameter',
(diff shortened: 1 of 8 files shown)


raiph:
Hi @ab5tract,

Has there been discussion I could read about the use cases and possible solutions?

Some code and related thoughts:
```
multi sub a($a) { $a, VAR $a }
multi sub a(@a) { "array" }
say a VAR $[1,2]; # ([1, 2] $[1, 2])
```
Here I've modified the call site. But I've long wondered why the
semantics of `$` weren't `VAR $`. If that change were made for
6.e then one could use a `$` sigil'd parameter as shown. It's not
the same as what you're suggesting, but perhaps it's better?

ab5tract:
@raiph Thanks for your comments! This is a follow-up/re-implementation of the behavior previously merged in #5543. The use case that has driven the development is to resolve bug report #5500. Discussions were logged over the last 3 or so weeks in `#raku-dev`.

The solution as implemented here is designed to be the least invasive / most agreeable in relation to keeping the base dispatch behavior while allowing `multi` dispatch designs to incorporate a distinction between positionals/associatives and their itemized forms (whether that is actually good API design is a different question).

Having dug around in the dispatcher quite a bit recently, I don't think there is an easy way to implement your proposal. It's also a significant change to the semantics of Raku. That's not a bad thing but it is beyond the scope of this MR's intentions.

PR #5553, opened 2024-04-21
Compiler.nqp: Add $rakudo-flavor

AntonOks:
references:
  https://github.com/rakudo/rakudo/pull/5516
  https://github.com/rakudo/star/pull/198


src/Perl6/Compiler.nqp
```diff
@@ -166,9 +173,23 @@ class Perl6::Compiler is HLL::Compiler {
             $rakudo := "Rakudo™";
         }
 
+        # Support different flavors of Rakudo, i.e. Star.
+        # Rakudo flavor can be either set by an environment variable 'RAKUDO_FLAVOR'
+        # or by the special file '$RAKUDO_PREFIX/share/perl6/site/FLAVOR'
```
ugexe, line 178: The repository paths are not an appropriate location for such a file

AntonOks: @ugexe @lizmat

Well... It has to be somewhere in $RAKUDO_PREFIX. I'm open for any suggestion. Feel free to make a qualified / better one... and make sure you are aware off
- https://github.com/rakudo/star/pull/198#issuecomment-2054111595
- https://github.com/rakudo/star/pull/198#issuecomment-2056327216

Thanks & regards

ugexe: I dunno where a better place is, but the repository paths are coupled to `CompUnit::Repository::Installation` objects and shouldn't contain files unrelated to them. The `version` file mentioned in one of those links is directly coupled to the `CompUnit::Repository::Installation` that path represents, it isn't a general rakudo version.

AntonOks: @ugexe @lizmat

And now? How to continue? Who can clarify and share with me where, within $RAKUDO_PREFIX, to put such a file?

ugexe: What other options (such as creating a new directory) have you considered? I don't exactly have time at the moment to do that type of investigative work, but it only takes me a moment for me to comment on potential options that have already had some thought put into them.

AntonOks: @ugexe @lizmat

Well.... `.config` would be a very prominent example, in Windows and UX environments. Look into your $USERHOME, I'd bet you have it, too ;)
The other option could be `.local`.

If a new directory is really an option, I'd vote for `$RAKUDO_PREFIX/.config/FLAVOR`.

ugexe: That seems fine. It looks like the convention used in `$RAKUDO_PREFIX` currently is UNIX-like e.g. `lib`, `bin`, so `etc` might be a good alternative as well. I don't have strong feelings either way.

AntonOks: @ugexe @lizmat

1.
Which other existing directories and files in `$RAKUDO_PREFIX` are auto-created / managed by other tools, i.e. by `CompUnit::Repository::Installation`?

2.
So, adding new dir's and files outside of what exists today in `$RAKUDO_PREFIX`, is "ok" due to your knowledge, right?

3.
Can one of you clarify, like in the "steering committee" or whatever the right gremium is, and share how to continue?!

Thanks

lizmat: Invoking the wisdom of @niner in these matters: I simply do not know enough to these intricacies yet  :-(

AntonOks: > That seems fine. It looks like the convention used in `$RAKUDO_PREFIX` currently is UNIX-like e.g. `lib`, `bin`, so `etc` might be a good alternative as well. I don't have strong feelings either way.

oki, so changed the lookup path for the new file now to $RAKUDO_PREFIX/etc/FLAVOR`...
hope, this change is woth you approval now....

Thanks for your feedback!

AntonOks:
@niner - seems like @lizmat wants you to review... hope, you are aware...
So, have you had a chance to look into it already? Concerns, suggestions, any other feedback?

thanks!

lizmat:
After discussions after the merge, I've decided to take a little different approach.

In short, you can now **only** set a flavor with the `RAKUDO_FLAVOR` environment variable **when building rakudo**, so:
```
$ RAKUDO_FLAVOR=Star make
````
This will burn in the flavor of Rakudo into the binary.

I've also added the `Compiler.flavor` method that will return the same burnt in flavor, so you *can* adapt any code flow depending on the flavor, should you wish to.

Issue #5464, opened 2023-11-05
Invocant of method 'match' must be a type object of type 'Any'


AlexDaniel:
Originally reported here: https://github.com/Raku/whateverable/issues/388#issuecomment-1737363873

Many of the whateverable bots were unable to run because of this issue:
```
Invocant of method 'match' must be a type object of type 'Any', not an
object instance of type 'IRC::Client::Message::Privmsg::Channel'.  Did
you forget a 'multi'?
```
* https://irclogs.raku.org/raku/2023-09-27.html#12:54-0001
* To me it looks like a Rakudo regression. Why would something be something that is other than “Any”?

Note that in Whateverable I had to apply this workaround to make things work: https://github.com/AlexDaniel/whateverable/commit/633616a3232a54baeb3fb445b8aea44d23537968

ab5tract:
A minimal example, thanks to @dakkar ([source](https://gist.github.com/dakkar/3b5d54bf3830ea1ca1f4ae0e25c326d9))

```
use v6.d;

class Thing {
    has $.text;
    method Str { "thing: $!text" }
}

sub foo($t where { m/^thing:/ }) {
    print $t;
}

sub bar($t where { "$t" ~~ m/^thing:/ }) {
    print $t;
}

# this works
bar(Thing.new(:text<abc>));

# this says `Invocant of method 'match' must be a type object of type
# 'Any', not an object instance of type 'Thing'.  Did you forget a
# 'multi'?`
foo(Thing.new(:text<abc>));
```

ab5tract:
With `RAKUDO_RAKUAST=1`, we get a failure even with the call to `bar`:

```
This type cannot unbox to a native integer: P6opaque, Match
   at /tmp/any.raku:12  (<ephemeral file>:bar)
```

ab5tract:
Unsurprisingly, this also affects subsets:

```
raku -e 'subset S where { m/"aaa"/ }; class C { method Str { "aaa" } }; my S $s = C.new'
Invocant of method 'match' must be a type object of type 'Any', not an
object instance of type 'C'.  Did you forget a 'multi'?
```

ab5tract:
[...]

PR #5559, opened 2024-04-29
Supply an Any.match method candidate for Any:D

ab5tract:
This addresses the use of non-`Str` candidates arriving
in `where` constraints that attempt to match against
a regex. (R#5464 / https://github.com/rakudo/rakudo/issues/5464).

```
subset S where { m/"aaa"/ }
class C { method Str { "aaa" } }
my S $s = C.new
|=> Invocant of method 'match' must be a type object of
|=> type 'Any', not an object instance of type 'C'.
|=> Did you forget a 'multi'?
```

That same error would occur with where clauses on
signature parameters.

Now it works as expected, calling `Str` on the clause
subject and running the match.

This can still fail if you are trying to match against
an instance of an Any class that does not supply a Str
method. However, this seems like a reasonable site for
such a concern to be raised.

The Any:U candidate (formerly the only candidate) is
useful for issuing the standard warning for trying to
stringify a type object. It is worth noting that
removing it causes issues with running a test case
in `t/spec/S05-substitution/subst.t`:

```
$_ = Any;
s[ea] = "rea";
is $_, "", 'can use s[]="" when $_ is not set';
```

This test case is marked as TODO.

### DIFF
--- a/src/core.c/Any.rakumod
+++ b/src/core.c/Any.rakumod
@@ -145,7 +145,9 @@ my class Any { # declared in BOOTSTRAP
 
     multi method iterator(Any:) { self.list.iterator }
 
-    method match(Any:U: |) { self.Str; nqp::getlexcaller('$/') = Nil }
+    proto method match(|)         {*}
+    multi method match(Any:U: |)  { self.Str; nqp::getlexcaller('$/') = Nil }
+    multi method match(Any:D: |c) { self.Str.match(|c) }
 
     proto method classify(|) is nodal {*}
     multi method classify() {


lizmat (approved):

src/core.c/Any.rakumod
```diff
@@ -145,7 +145,9 @@ my class Any { # declared in BOOTSTRAP
 
     multi method iterator(Any:) { self.list.iterator }
 
-    method match(Any:U: |) { self.Str; nqp::getlexcaller('$/') = Nil }
+    proto method match(|)         {*}
+    multi method match(Any:U: |)  { self.Str; nqp::getlexcaller('$/') = Nil }
```
lizmat, line 149: Meh, I should have looked closer, sorry.

I don't think the `nqp::getlexcaller` will actually affect the caller's `$/`, because that is hidden by the proto, at least on the non-MoarVM backends.

See: https://github.com/rakudo/rakudo/blob/main/src/core.c/Str.rakumod#L1495-L1500

lizmat: Also I wonder whether the `proto` in `Str` can now go?

ab5tract: Good catches, I didn't want to rock the boat so I just left the previous method body as is.

ab5tract: Removing the `proto` in `Str` unfortunately causes spectests to stall out.

ab5tract: It happens right after `t/spec/integrations/weird-errors.t` returns `ok`. A handful of tests have also failed by then.

ab5tract: It does look like removing the `nqp::getlexcaller` caller does not have any negative impacts, at least as far as Roast coverage goes.

lizmat:
FWIW, a golf of the issue: `say Int.match(42)`

Issue #5175, opened 2023-01-30
where-clause in subset does not work with Block/Callable


gfldex:
```
subset IntCallabel of Block where *.signature ~~ :(Int);

sub foo(IntCallabel \c) { }

sub works(Int $i) { }

foo(&works);
```
OUTPUT:
```
Can only use dispatcher-track-attr on a P6opaque
  in sub foo at rakudobug/tmp.raku line 3
  in block <unit> at rakudobug/tmp.raku line 7
```

The same happens with a block instead of the Whatever. I didn't check against the RakuAST branch.

## Environment
* Operating system: Linux dexhome 5.18.0-3-amd64 #1 SMP PREEMPT_DYNAMIC Debian 5.18.14-1 (2022-07-23) x86_64 GNU/Linux
* Compiler version (`perl6 -v` or `raku -v`): Welcome to Rakudo™ v2022.12-1-gd52342eb0. Implementing the Raku® Programming Language v6.d. Built on MoarVM version 2022.12.

ab5tract:
Just checked and this is also an issue on the RakuAST branch with the new subset implementation. This suggests the problem/fix is somewhere in the metamodel code itself.

ab5tract:
This appears to be `WhateverCode` specific, as the following version of the subset works fine:

```
subset IntCallable2 of Block where { .signature ~~ :(Int) };
sub foo(IntCallable2 \c) { }
sub works(Int $i) { }
foo(&works);
```

Even though the `WhateverCode` stack in RakuAST has been massively overhauled, the behavior there has not changed (ie, it still fails).

It's failing in MoarVM's `MVM_disp_program_record_track_attr` where it fails the check that ensures `MVM_REPR_ID_P6opaque` because the representation ID is `MVM_REPR_ID_Uninstantiable` (`10`) instead.

ab5tract:
I've created a (non-exploding) golf that demonstrates the issue:

```
> with -> Int $ {} { { .signature ~~ :(Int) } ?? say "yeah" !! say "no" }
yeah
> with -> Int $ {} { *.signature ~~ :(Int) ?? say "yeah" !! say "no" }
no
```

It appears to be an issue with the thunking of blocks into `WhateverCode`.

For example, according to the general rules of `Whatever` thunking, the following should probably work:

```
> with -> Int $i { $i } { *.(5) == 5 ?? say "yeah" !! say "no" }
No such method 'CALL-ME' for invocant of type 'Whatever'
[...]

PR #5564, opened 2024-05-11
Add worry for WhateverCode on LHS of smartmatch

ab5tract:
Smartmatch only curries `Whatever`, so any user that provides a `WhateverCode` expression as an LHS is likely to encounter a surprising failure (when compared to the behavior of most other infix operaators).

In order to be proactive and helpful to such users, we emit a descriptive worry in this scenario.


ab5tract:
This addresses the confusion underlying #5175

lizmat (approved):

Issue #1705, opened 2018-04-07
META_REDUCE_RIGHT doesn't check for 1-arg case like META_REDUCE_LEFT


zoffixznet:
From IRC: https://irclogs.raku.org/perl6/2018-04-07.html#16:15-0001

```
16:15 |   | m: sub infix:<@> ($a, $b) is assoc<left> { [$a, $b] }; say [@] 10
-- | -- | --
16:15 | camelia | rakudo-moar edbbc4426: OUTPUT: «10␤»
16:15 | cfa | m: sub infix:<@> ($a, $b) is assoc<right> { [$a, $b] }; say [@] 10
16:15 | camelia | rakudo-moar edbbc4426: OUTPUT: «Too few positionals passed; expected 2 arguments but got 1␤  in sub infix:<@> at <tmp> line 1␤  in block <unit> at <tmp> line 1␤␤»
16:16 | Zoffix | heh
16:16 | cfa | i'd expect either both to work or both to fail given the lack of supplied identity
16:16 |   | i'm probably missing something obvious
16:16 | Zoffix | Like this being a bug? :P


16:31 | TimToady | well, METAOP_REDUCE_LEFT checks for the 1-arg case, while METAOP_REDUCE_RIGHT doesn't
-- | -- | --
16:32 |   | there's not actually much call for the built-in right-associative operators in real code, so nobody ever noticed that before
16:32 | Zoffix | mhm
16:33 | TimToady | m: say infix:<**>(42)
16:33 | camelia | rakudo-moar edbbc4426: OUTPUT: «42␤»
16:33 |   | wamba joined #perl6
16:33 | TimToady | at least the support is there in that case
16:35 | Zoffix | Basically this https://github.com/rakudo/rakudo/blob/master/src/core/metaops.pm6#L196-L198 needs to be stuck here: https://github.com/rakudo/rakudo/blob/master/src/core/metaops.pm6#L393 (and maybe triangle cases need it too) if anyone wants to do it
```

[Also](https://github.com/perl6/doc/issues/1399) does `METAOP_REDUCE_CHAIN` need something too? There's no 1-arg case for `&infix:<(elem)>` yet the 1-arg case returns `True`. Seems to make sense, but that supposed to be like that?

```
<Zoffix__> m: say &infix:<(elem)>.candidates».signature
[...]

PR #5576, opened 2024-05-22
Give META_REDUCE_RIGHT behavior for 1-arg case

ab5tract:
This aligns the behavior of `META_REDUCE_RIGHT` to that of `META_REDUCE_LEFT` when handling 1-arg cases:

    > sub infix:<@> ($a, $b) is assoc<left> { [$a, $b] }
    > say [@] 10
    10

    > sub infix:<@> ($a, $b) is assoc<right> { [$a, $b] }
    > say [@] 10
    10

This addresses R#1705 (#1705).

My only question is: **_should_** we be doing this?


lizmat:
I'd say **YES**, since:
```raku
sub infix:<@> ($a, $b) is assoc<right> { [$a, $b] }; say [@] 10'
# Too few positionals passed; expected 2 arguments but got 1
```
is an execution error.

Issue #1590, opened 2018-03-05
Procs should not get back with “exit code: 1” if the program does not exist at all


AlexDaniel:
Code:
```perl6
run ‘oeunohunteohuntoahuonathu’
```
Result (`92bd7e4f54a^`):
```
The spawned command 'oeunohunteohuntoahuonathu' exited unsuccessfully (exit code: -1)
  in block <unit> at /tmp/ARHjknJLZK line 1
```

Result (92bd7e4f54a):
```
The spawned command 'oeunohunteohuntoahuonathu' exited unsuccessfully (exit code: 1)
in block <unit> at /tmp/ARHjknJLZK line 1
```

Bisected to: (2017-06-12) https://github.com/rakudo/rakudo/commit/92bd7e4f54a92fa660f99b4d056d33a08fb98bd2
Output on all releases: https://gist.github.com/aa42b6eb50f114c31e4c070a59b26a53

I don't think that exit code `-1` is very meaningful, but at least it was possible to figure out that the program does not exist at all. Now it's indistinguishable from the actual program exiting with 1.

FWIW the `.signal` in this case is now 0 but previously it was `254`.

Context:
* Here's me trying to work around the issue: https://github.com/perl6/whateverable/commit/a401e1da8d
* And here's me asking for help, and then realizing it's a regression (an old one, but still): https://irclog.perlgeek.de/perl6/2018-03-05#i_15884805

AlexDaniel:
For anyone reading by email: the link to the bisect result was wrong. Fixed now.

CIAvash:
Just wanted to mention that `Proc::Async` dies with an exception in this case.

```raku
await Proc::Async.new('abcd').start
```
```
An operation first awaited:
  in block <unit> at <unknown file> line 1

Died with the exception:
    Failed to spawn process abcd: no such file or directory (error code -2)
```

Depending on how you invoke the process, you get different exception names: e.g, `X::OS+{X::Await::Died}`, `X::OS+{X::React::Died}`.
```
X::OS+{X::Await::Died}.new(await-backtrace => Backtrace.new, os-error => "Failed to spawn process abcd: no such file or directory (error code -2)")
```

ab5tract:
With this [roast test @ 3805504](https://github.com/Raku/roast/commit/3805504e59), this can be considered completed.

PR #5580, opened 2024-05-25
Provide deeper context into `run` failures

ab5tract:
After 92bd7e4, we stopped returning an exit code of `-1` for cases when attempting to `run` a nonexistent command.

This patch returns the previous behavior while also surfacing the underlying error from `libuv`, which makes it much more clear to the user what the underlying cause of the `run` failure is:

    > r 'run "e-not-a-command"'
    The spawned command 'e-not-a-command' exited
        unsuccessfully (exit code: -1, signal: 254)
    (OS error = Failed to spawn process e-not-a-command:
        no such file or directory (error code -2))

R#1590 (#1590)

### DIFF
--- a/src/core.c/Proc/Async.rakumod
+++ b/src/core.c/Proc/Async.rakumod
@@ -390,9 +390,8 @@ my class Proc::Async {
             $!ready_vow.keep($pid);
         });
 
-        nqp::bindkey($callbacks, 'error', -> Mu \err {
-            my $error := X::OS.new(os-error => err);
-            $!exit_promise.break($error);
+        nqp::bindkey($callbacks, 'error', -> Mu \err, $code = 0 {
+            $!exit_promise.break(my $error := X::OS.new(:os-error(err), :error-code($code)));
             $!ready_vow.break($error);
         });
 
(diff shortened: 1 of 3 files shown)


src/core.c/Proc/Async.rakumod
```diff
@@ -391,7 +391,10 @@ my class Proc::Async {
         });
 
         nqp::bindkey($callbacks, 'error', -> Mu \err {
-            my $error := X::OS.new(os-error => err);
+            my %details = os-error => err;
+            my $code = try (err ~~ /"(error code " ("-"? \d+) ")"/)[0].Int;
```
niner, line 395: Seems to me that it'd be very easy to pass the error code as a separate argument to the `error` callback which would save us the potentially flaky parsing of the error message.

niner (changes requested):

niner (approved):

lizmat:
Does this require a MoarVM / NQP bump?

niner:
It's the other way round: the MoarVM PR requires this change but this change is backwards compatible with older MoarVMs

Issue #1915, opened 2018-06-12
LTA error message when using ${_}


sjn:
When running the following,

`echo "foo" | perl6 -ne 'say "${_}"'`

we get this error message:

```
===SORRY!=== Error while compiling -e
Unsupported use of ${_}; in Perl 6 please use {$_}
at -e:1
------> say "${_}⏏"
```

...but the example given there doesn't do the same as I originally intended. 

This is Rakudo version 2018.03 built on MoarVM version 2018.03
implementing Perl 6.c.

AlexDaniel:
The problem is not in the error message, it's just that this works:
```bash
echo "foo" | perl6 -ne 'say "$_"' # foo
```
But this doesn't:
```bash
echo "foo" | perl6 -ne 'say "{$_}"' # Use of uninitialized value $_ of type Any in string context.
```

ab5tract:
Unfortunately this is a bit twisty to resolve in base (for me anyway). But it is no longer an issue in RakuAST.

PR #5581, opened 2024-05-26
RakuAST: Add -n and -p compiler options

ab5tract:
Once again RakuAST proves itself by making the implementation of this feature dead-simple while also fixing a long-standing bug.

From R#1915 (#1915):

    raku -ne 'say "{$_}"'
    # Use of uninitialized value $_ of type Any in string context.

No workarounds were required in RakuAST, only the implementation of a helper function to wrap the AST provided to the compiler inside a `for` loop that calls lines, with the optional addition of a call to `say($_)` appended to the provided AST.

### DIFF
--- a/src/Raku/Actions.nqp
+++ b/src/Raku/Actions.nqp
@@ -43,6 +43,28 @@ sub Nodify(*@todo) {
     )
 }
 
+# Used by '-p' to print $_ on each iteration
+sub print-topic() {
+    Nodify('Statement', 'Expression').new(expression =>
+        Nodify('Call', 'Name').new(
+            name => Nodify('Name').from-identifier('say'),
+            args => Nodify('ArgList').new(Nodify('Var', 'Lexical').new('$_'))))
+}
+
+# Provide the functionality of '-n' and '-p'
+sub wrap-in-for-loop($ast) {
+    Nodify('StatementList').new(
+        Nodify('Statement', 'For').new(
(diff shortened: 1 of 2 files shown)


niner (commented):
I think wrap-in-for-loop should be split into `$statement-list.add(print-topic) followed by a $statement-list.wrap-in-for-loop. One function, one action.

niner (approved):

src/Raku/Actions.nqp
```diff
@@ -369,8 +391,15 @@ class Raku::Actions is HLL::Actions does Raku::CommonActions {
         my %OPTIONS  := %*OPTIONS;
         my $RESOLVER := $*R;
 
+        # Do any requested wrapping (-n or -p)
+        my $statement-list := $<statementlist>.ast;
+        if (my $add-print-topic := nqp::existskey(%OPTIONS,'p')) || nqp::existskey(%OPTIONS,'n') {
+            $statement-list.add-statement(print-topic()) if $add-print-topic;
```
MasterDuke17, line 397: Won’t this print with ‘-n’ *or* ‘-p’ instead of just ‘-p’?

niner: I think you missed the postfix-if here

MasterDuke17: Ah no, I missed that the bind to $add-print-topic is tighter than the ||.

Issue #4666, opened 2021-12-12
is cached sometimes fails to cache invocations with hash-like arguments.


nikola-benes:
## The Problem

```Raku
use experimental :cached;
sub foo($s) is cached { say "not cached: $s" }
foo(<1 2 3>.Set) for ^100;
```

## Expected Behavior

One line of output with `not cached:` and an arbitrary ordering of 1, 2, 3.

## Actual Behavior

Usually, six lines of output with all the permutations of 1, 2, 3. If you are really lucky, some of the permutations may be missing.

## Environment
* Operating system: Linux
* Compiler version: Rakudo™ v2021.10

## Comments

I get why this happens (`is cached` stringifies the capture and hashes are randomized); however, it would be nice to have a caching mechanism that works well with hashes (or at least sets).

ab5tract:
We have a few issues going on here:

* `is cached` based on `.gist` doesn't make sense at this point
* `is cached` is not implemented in a thread-safe way (not an issue for this ticket in particular, but since we are poking at the trait, it's worth pointing out)
* even switching to `.raku`, we still have to adjust some of the undordered collections to sort in that method

lizmat:
> however, it would be nice to have a caching mechanism that works well with hashes (or at least sets).

There are no plans on trying to fix "is cached".  Most likely, it will be deprecated from the core.

The ecosystem provides several solutions: https://raku.land/?q=cache

PR #5582, opened 2024-05-26
Provide some fixups to experimental :cached

ab5tract:
Addresses R#4666 (#4666), the discussion of which
raised several issues with the previous `is cached` implementation:

- `.gist` is not appropriate as details may be truncated, leading to potential invalid cache matches

- `is cached` is not thread-safe

This patch moves to address both issues by using `.raku` instead and wrapping it in a `lock`/`unlock` cycle. (`Lock.protect` is not an option as it hijacks the `callsame`).

Additionally, `Setty.raku` and `Baggy.raku` have been adjusted to produce sorted contents.

### DIFF
--- a/lib/experimental.rakumod
+++ b/lib/experimental.rakumod
@@ -7,11 +7,18 @@ package EXPORT::rakuast {
 package EXPORT::cached {
     multi sub trait_mod:<is>(Routine $r, :$cached!) {
         my %cache;
+        my $lock = Lock.new;
         $r.wrap(-> |c {
-            my $key := c.gist;
-            %cache.EXISTS-KEY($key)
-              ?? %cache{$key}
-              !! (%cache{$key} := callsame);
+            my $value;
+            my $key := c.raku;
+            {
+                $lock.lock;
+                $value := %cache.EXISTS-KEY($key)
+                    ?? %cache{$key}
+                    !! (%cache{$key} := callsame);
+                LEAVE $lock.unlock;
+            }
+            $value
         });
     }
 
--- a/lib/experimental.rakumod
+++ b/lib/experimental.rakumod
@@ -29,6 +29,30 @@ package EXPORT::cached {
     OUR::{'&trait_mod:<is>'} := &trait_mod:<is>;
 }
 
+package EXPORT::synchronized-cached {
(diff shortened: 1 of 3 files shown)


src/core.c/Baggy.rakumod
```diff
@@ -318,15 +318,16 @@ my role Baggy does QuantHash {
           $!elems && nqp::elems($!elems),
           nqp::stmts(
             (my \pairs := nqp::join(',',
-              Rakudo::QuantHash.RAW-VALUES-MAP(self, {
+              Rakudo::Sorting.MERGESORT-str(
```
lizmat, line 321: Feels this slipped in?

src/core.c/Setty.rakumod
```diff
@@ -174,9 +174,10 @@ my role Setty does QuantHash {
           nqp::concat(
             nqp::concat(
               nqp::concat(self.^name,'.new('),
-              nqp::join(",",Rakudo::QuantHash.RAW-VALUES-MAP(self, *.raku))
-            ),
-            ')'
+              nqp::join(",", Rakudo::Sorting.MERGESORT-str(
```
lizmat, line 177: Feels this slipped in?

ab5tract: This is intentional: we need these unordered collections to return an ordered accounting via `.raku`.

lib/experimental.rakumod
raiph, line 1: Bikeshed comment... Not sure what you're intending longer term but maybe `:$cached! where 'sync'` instead of `:$synchronized-cached` so one writes `sub foo is cached<sync>`. I get that `where` is slower, but aiui it shouldn't have any run time cost unless an `is cached` trait is _used_, and if it is the run time cost is about a tenth of what it was since the new dispatch work landed (in 2022?).

ab5tract: I like this approach a lot.

However I'm inclined to closing this ticket entirely. Fixing issues with this feature has been essentially sidelined as the raku method is considered to be inappropriate for caching while the scope of caches in general tend to call for different tradeoffs.

This makes the entire enterprise of providing a drop-in cache a bit of a conundrum. I therefore lean towards depreciation instead.

PR #5594, opened 2024-06-11
Improve `.flat` logic

lizmat:
- add positional to indicate number of levels to flatten.  Defaults to Inf / Whatever
- add boolean named arg :hammer.  If specified with a true value, will disregard any containerization on Iterables, and flatten those iterables.  Defaults to False

Inspired by https://github.com/Raku/problem-solving/issues/431

### DIFF
--- a/src/core.c/Array.rakumod
+++ b/src/core.c/Array.rakumod
@@ -410,7 +410,11 @@ my class Array { # declared in BOOTSTRAP
     }
 
     multi method flat(Array:U:) { self }
-    multi method flat(Array:D:) { Seq.new(self.iterator) }
+    multi method flat(Array:D: $levels = Whatever, :$hammer) {
+        $hammer
+          ?? self.Iterable::flat($levels, :hammer)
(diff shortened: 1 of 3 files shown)


lizmat:
Some benchmarks:
```
$ raku -e 'my @a = ^1000; my @b = @a.tree(*.flat xx 10) for ^1000'
real	1.88s
user	1.88s
sys	0.02s
$ raku -e 'my @a = ^1000; my @b = @a.flat(:hammer) for ^1000'
real	0.42s
user	0.41s
sys	0.02s
$ raku -e 'my @a = ^1000; my @b = @a for ^1000'
real	0.42s
user	0.41s
sys	0.02s
$ raku -e 'my @a = ^1000; Nil for ^1000'
real	0.13s
user	0.12s
sys	0.02s
```
So `.flat(:hammer)` is basically the same as storing an array into another array.  If we remove the startup overhead, `.flat(:hammer)` is about 6x as fast as `.tree(*.flat xx *)`

PR #5596, opened 2024-06-13
Introduce a `Block.WhateverCode` coercer

lizmat:
This is really just a quick implementation of an idea wrt performance. Since WhateverCodes cannot have phasers, they can be treated differently in .map / for loops taking a simpler path.  However, from the grammar it is impossible to create slightly more complex WhateverCodes, even though they are actually quite simple, e.g.: { $_ ~ $_ } .

### DIFF
--- a/src/core.c/Block.rakumod
+++ b/src/core.c/Block.rakumod
@@ -30,6 +30,27 @@ my class Block { # declared in BOOTSTRAP
         )
     }
 
+    proto method WhateverCode(|) {*}
+    multi method WhateverCode(Block:U:) { WhateverCode }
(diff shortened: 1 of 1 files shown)


ab5tract:
Would it be an idea to use a "phasers count" flag on blocks and, when zero, send it down the fast path?

lizmat:
There basically is a "phasers count" flag already: `Block.has-phasers`.

niner:
I wonder if the assumption will actually hold. Macros will be able to modify WhateverCodes in arbitrary ways, possibly including the addition of phasers.

PR #5604, opened 2024-07-11
Partially fix mixed bin/no bin reads in IO::Socket

patrickbkr:
When mixing binary and non-binary accesses, the data is put into an
encoder for decoded accesses. It's possible that not all data that was
added to the decoder is read. Subsequent binary reads would miss the
data already added to the decoder. Fix this by first emptying the decoder
in binary accesses should one be set.

This only works within the limitations of the caching of the decoder
itself. If the decoder chooses to keep bytes in it's pipeline (which
happens for example with UTF-8 and combiner chars), then those cases
will suffer the same fate as before: the binary read will not see those
bytes stuck in the pipeline and read the bytes after that. Only the next
decoded read will see those bytes.

### DIFF
--- a/src/core.c/IO/Socket.rakumod
+++ b/src/core.c/IO/Socket.rakumod
@@ -6,12 +6,28 @@ my role IO::Socket {
     has Encoding::Decoder $!decoder;
     has Encoding::Encoder $!encoder;
 
-    method !ensure-coders(--> Nil) {
+    method !ensure-decoder(--> Nil) {
         unless $!decoder.DEFINITE {
             my $encoding = Encoding::Registry.find($!encoding);
             $!decoder := $encoding.decoder();
             $!decoder.set-line-separators($!nl-in.list);
+        }
+    }
+
+    method !ensure-encoder(--> Nil) {
+        unless $!encoder.DEFINITE {
+            my $encoding = Encoding::Registry.find($!encoding);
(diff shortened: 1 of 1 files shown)


ugexe:
@jnthn do you remember if there is a fundamental reason for the existing behavior, or can we actually consider this a bug worth attempting to fix?

jnthn:
@ugexe Best I can recall, the only reason you'd not get chars back is if there could be a combiner that would go onto the last fully read char. Switching mode is rare anyway, and the most obvious practical example - parsing HTTP headers and then reading a binary body - would never end up with bytes lost in the decoder, because the headers end with a control char, and there's no reason to keep those back as they are always grapheme breaks. (While the word "caching" is used in the PR, this isn't a caching mechanism at all, and it's implemented to be as eager at spitting out chars as the Unicode grapehme break rules allow.)

I dunno if "can't imagine any practical situation where you'd be switching not at a grapheme break" is a fundamental enough reason, but I considered it a good enough one to avoid making the decoding code even more complex. :-)

patrickbkr:
Thanks for the insights! I guess this PR is then good enough to merge and probably never care about this ever again. :-)

patrickbkr:
Converted to draft, as I'd like to fix the commit message wrt the wrong use of the word "caching".

PR #5607, opened 2024-07-18
Add `:by` to `.min` / `.max` / `.minmax`

lizmat:
The sub versions of min/max/minmax take a :by named argument to indicate the comperator.  However, the method versions take an optional positional argument and ignore any :by specification (because of the implicit *%_ in the method signature.

This can cause silent breakage when refactoring code from subs to methods as spotted by vendethiel++
  https://irclogs.raku.org/raku/2024-07-18.html#14:27

This adds 3 candidates taking the :by named argument, so that this type of breakage will not happen when refactoring code.

### DIFF
--- a/src/core.c/Order.rakumod
+++ b/src/core.c/Order.rakumod
@@ -246,6 +246,7 @@ augment class Any {
 
         nqp::defined($min) ?? $min !! Inf
     }
+    multi method min(Any:D: :&by!) { self.min(&by, |%_) }
     multi method min(Any:D: &by) {
         my &comparator := aritize22(&by);
 
@@ -302,6 +303,7 @@ augment class Any {
 
         nqp::defined($max) ??  $max !! -Inf
     }
+    multi method max(Any:D: :&by!) { self.max(&by, |%_) }
     multi method max(Any:D: &by) {
         my &comparator := aritize22(&by);
 
@@ -426,7 +428,7 @@ augment class Any {
           ?? Range.new($min,$max,:$excludes-min,:$excludes-max)
           !! Range.Inf-Inf
     }
-
+    multi method minmax(Any:D: :&by!) { self.minmax(&by, |%_) }
     multi method minmax(Any:D: &by) {
         nqp::if(
           (my $iter := self.iterator-and-first(".minmax",my $pulled)),


lizmat:
`.min`, `.max`, `.minmax` use `cmp` semantics, and that's a good thing.  So we don't need different versions.

> I question whether the `:by` parameter unnecessarily complicates the situation.

What's the complication?  Adding the named `:by` argument to the method version, is making it more DWIM.  FWIW, in the future, I would see it quite possible that the positional `&by` argument would be deprecated, for consistency's sake.

jubilatious1:
> `.min`, `.max`, `.minmax` use `cmp` semantics, and that's a good thing. So we don't need different versions.
> 
> > I question whether the `:by` parameter unnecessarily complicates the situation.
> 
> What's the complication? Adding the named `:by` argument to the method version, is making it more DWIM. FWIW, in the future, I would see it quite possible that the positional `&by` argument would be deprecated, for consistency's sake.

_[cmp](https://docs.raku.org/routine/cmp) Compares strings with string semantics, numbers with number semantics, Pair objects first by key and then by value etc._

Except with `.min`, `.max`, `.minmax` , this doesn't appear to be `True`. The default appears to be that elements of a List/Array are  coerce to `Num` and then compared. If you want to compare `Str`, you have to use `:by`. I think I've gotten this correct? If not, why not? Aren't `.min`, `.max`, `.minmax` most commonly used to compare numerics?  So if you need to coerce to `*.Str`, why can't `.min-str`, `.max-str`, `.minmax-str`  be implemented?

Furthermore, AFAIK the docs are silent on a mixed exemplary Array like: `[7,8,9,"10","4"]` :

 - I submit to you that the average Newbie has no idea what happens when you try `[7,8,9,"0","four"].min`. 
 - OTOH,  the average Newbie could predict that  `[7,8,9,"0","four"].minstr` would sort the strings first.

I would have expected by now that pretty much everything in Perl5's `List::Util` would have found its way into Raku's core. As I'm certain you're aware, `List::Util` implements `minstr`, etc.

lizmat:
>> Except with .min, .max, .minmax , this doesn't appear to be True. The default appears to be that elements of a List/Array are coerce to Num and then compared.

What brings you to that conclusion?
```raku
say @a>>.Str.min;  # string semantics
say @a>>.Numeric.min;  # numeric semantics
```
If you want specific semantics, it's not that hard to get them.

0racle:
I'd like to use this platform to state that I was _very_ early on this issue... [8 years early!](https://gist.github.com/0racle/134921795a6d91168f31cd554492d824) :D

Ok, I'm talking about sort, but in the linked IRC chats, I mention that `min`, `max`, and `minmax` have the same gotcha.

> FWIW, in the future, I would see it quite possible that the positional `&by` argument would be deprecated, for consistency's sake.

While breaking a change is never easy, I'm supportive of this. I know Raku doesn't have a "real" [UCS](https://en.wikipedia.org/wiki/Uniform_Function_Call_Syntax) like Nim (`.&sub` syntax notwithstanding)... but it would be nice if the sub and method versions we're more aligned in terms of the args they accept.

ab5tract:
In my opinion, the Raku way is to modify semantics via a named parameter, not by exploding the number of subs and methods.

Barring that, combining method calls to clarify semantic intent is a clear alternative approach.

As to the implementation, I don't quite under why the max method is not implemented as:

~~~
method max(*@a) { max(|@a) }
~~~

Or vice versa.

lizmat:
> I don't quite understand why the max method is not implemented as:
> method max(*@a) { max(|@a) }
> Or vice versa.

It should be vice versa.   That allows one to create multi candidates for custom objects, and have the dispatch in the sub do the right thing.

lizmat:
> you could even cull the &by positional arg using revision gated multi candidates.

I thought that only worked for "revision x or higher" ?

ab5tract:
> > you could even cull the &by positional arg using revision gated multi candidates.
> 
> 
> 
> I thought that only worked for "revision x or higher" ?

I was imagining that you could create new 6.e candidates that don't have &by as a posarg.

But it's true that the trait doesn't currently support a maximal version. So we cannot currently provide disjoint candidates where candidate A is only available for dispatch before v6.e, and candidate B is only available for v6.e and later.

That seems vaguely useful, at least at first glance.

Issue #5611, opened 2024-07-26
Dynamic variables only visible one async level deep


lizmat:
This works as expected:
```raku
my $*FOO := "FOO";
await start { say $*FOO }   # FOO
```
This doesn't:
```raku
my $*FOO := "FOO";
await start { await start { say $*FOO } }
```
```
An operation first awaited:
  in block  at -e line 2

Died with the exception:
    Dynamic variable $*FOO not found
```

It shouldn't matter how many levels deep we are in async processing.

lizmat:
The reason of this behaviour is that at https://github.com/rakudo/rakudo/blob/7ed824904668eea6e6734b29fb8edf4503f9cd6b/src/core.c/Promise.rakumod#L366-L367 the lookup is fixed to the context of **that** call.  That context is then taken in https://github.com/rakudo/rakudo/blob/7ed824904668eea6e6734b29fb8edf4503f9cd6b/src/core.c/stubs.rakumod#L42-L46 to resolve any dynamic variable lookup.

vrurg:
I've been there, tried fixing, until @jnthn pointed out at the fact that maintaining the full chain of `Promise` call "stack" will result in huge memory hog due to closures mounting on top of each other and not being GCed.

Think of it from another point of view too. A `Promise` produce a "child" `Promise` and then gets gone for good. In this case the "child" `Promise` must not be able to see the dynamics that are only visible through the "parent".

lizmat:
I found out about this issue through [`ParaSeq`](https://raku.land/zef:lizmat/ParaSeq).  The batcher runs in its own thread, and that cues the work to be done.

I've fixed this in `ParaSeq` by keeping the outer context, so that any inner async code would just see the most outer context.  It's sufficient for `ParaSeq`, as there's no use of dynamic variables in `ParaSeq` itself (other than the ones found in `PROCESS::`).

Perhaps `start`should use the same approach?  Basically passing on the context (instead of taking a new one) when `start` is being done, and there is already a `$*PROMISE`?

vrurg:
Then `start` would need to know if the context where it's invoked doesn't provide own dynamics. And even if this is implemented then it means one more special case is introduced.

[...]

PR #5612, opened 2024-07-27
Keep outer context when starting inside a start

lizmat:
As a solution to https://github.com/rakudo/rakudo/issues/5611

Instead of keeping a stack of contexts (which jnthn mentioned could become a memory hog), it only keeps the first outer context encountered, rather than taking a new context every time.

This makes:

```raku
my $*FOO := "foo"; await start { await start { say $*FOO } }
```
produce "foo" rather than die.

### DIFF
--- a/src/core.c/Promise.rakumod
+++ b/src/core.c/Promise.rakumod
@@ -364,7 +364,16 @@ my class Promise does Awaitable {
     method start(Promise:U: &code, :&catch, :$scheduler = $*SCHEDULER,
             :$report-broken-if-sunk, |c) {
         my $p := self.new(:$scheduler, :$report-broken-if-sunk);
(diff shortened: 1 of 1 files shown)


vrurg:
I'm afraid, this kind of solution may produce a lot of WAT in a case of recursive production of chained Promises. While visually these grand-grand-...-Promises may be looking like they're second-level, they won't have access to the main code dynamics.

Besides, the choice of two-level nesting looks like arbitrary to me if implementation details are not taken into consideration. While 1-level limit can be well explained by giving user code a chance to re-instantiate an outer dynamic in it's own `my` stash, I currently see no other explanation to the 2-level allowance but as a need to fix a particular code case from a practical application. Otherwise why not 3 or more levels?

PR #5614, opened 2024-08-03
Fix precomp build

patrickbkr:
mirrorlist.centos.org is discontinued.
https://serverfault.com/a/1161847


lizmat:
Is there a reason why this wasn't merged yet?

patrickbkr:
Nope.
The latest release was already done with this patch. It's good to go. Thanks for asking!

lizmat:
Ok, I'll merge after the weekly

PR #5624, opened 2024-08-20
Add a test for interpolating variables into grammars

patrickbkr:
Passes with RakuAST fails with the legacy compiler.


lizmat:
Shouldn't this be in roast?

lizmat:
Aah... missed the file it was being done in...  :-)

Issue #5644, opened 2024-09-17
rakudo optimizer changes behaviour of `$_ ~~ Mu:U`


timo:
## The Problem

I randomly stumbled upon this when looking for a good piece of code to suggest for "does this variable have a type object in it?":

```
rakudo -e '$_ = Int; say $_ ~~ Mu:U; $_ = 99; say $_ ~~ Mu:U'
rakudo --optimize=off -e '$_ = Int; say $_ ~~ Mu:U; $_ = 99; say $_ ~~ Mu:U'
```

## Expected Behavior

```
True
False

True
False
```

## Actual Behavior

```
True
True

True
False
```

## Environment

```
Welcome to Rakudo™ v2024.08-87-g067cd6f3e.
Implementing the Raku® Programming Language v6.d.
Built on MoarVM version 2024.08-10-g571effbd4.

Distributor ID: Fedora
Description:    Fedora Linux 40.20240911.0 (Kinoite)
```

## Bisect

I bisected the commit where with `$_ = 99` the result of `say $_ ~~ Mu:U` switched from False to True:

https://github.com/rakudo/rakudo/commit/91ec1bbf60

This is what I think is the relevant part of the `--target=optimize` output. with `--optimize=` off or 1 it emits a regular call to `&infix:<~~>` instead.
```
- QAST::Op(if)  :smartmatch_optimized<?>
  - QAST::Op(if) 
    - QAST::Op(istype) 
      - QAST::Var(local __lowered_lex_1) <wanted> $_
      - QAST::WVal(Junction) 
    - QAST::Op(isconcrete) 
      - QAST::Var(local __lowered_lex_1) <wanted> $_
  - QAST::Op(callmethod BOOLIFY-ACCEPTS) 
    - QAST::Var(local __lowered_lex_1) <wanted> $_
    - QAST::WVal(Mu:U) <wanted> Mu:U
    - QAST::WVal(Bool) 
  - QAST::WVal(Bool) 
```

lizmat:
This appears fixed (as in perhaps: not yet broken again) in RakuAST

ab5tract:
A fix that works locally is just to reject the optimization when LHS is `$_`.

But I'll look into why its working in the RHS to see if I can come up something less brutal.

vrurg:
> A fix that works locally is just to reject the optimization when LHS is `$_`.
> EDIT: Also curious why it works fine in (all?) other cases on the LHS.

[...]

PR #5645, opened 2024-09-18
Address a small but significant corner case in the optimizer

ab5tract:
(Fixes R#5644 [#5644])

> with 42 { say $_ ~~ Mu:U }
\# True
> my $m = Mu:U; say $m ~~ Mu:U }
\# True

The above was what the optmized QAST was returning, prmpting this patch.

It makes sense, because in almost all cases the LHS of a smart match will return `True` when smartmatched against `Mu`.

The optimizer wisely took this into consideration. There's only one problem, however, and that's that the moment when you are genuinely checking for `Mu:U` is the one moment where the optimization fails.

This commit utilizes `nqp::ifnull` where possible to reduce the overhead, but it's also a bit of a shame that that I couldn't think of a way to accomplish this without adding a conditional statement to every `maybe_typematch` call in the optimizer.

### DIFF
--- a/src/Perl6/Optimizer.nqp
+++ b/src/Perl6/Optimizer.nqp
@@ -28,6 +28,7 @@ my class Symbols {
 
     # Some interesting symbols.
     has $!Mu;
+    has $!Mu'U;
     has $!Junction;
     has $!Any;
     has $!Block;
@@ -66,6 +67,7 @@ my class Symbols {
         }
         nqp::push(@!block_stack, $!UNIT);
         $!Mu          := self.find_in_setting('Mu');
+        $!Mu'U        := nqp::gethllsym('Raku', 'Mu:U');
         $!Junction    := self.find_in_setting('Junction');
         $!Any         := self.find_in_setting('Any');
         $!Block       := self.find_in_setting('Block');
@@ -119,6 +121,7 @@ my class Symbols {
     method GLOBALish()   { $!GLOBALish }
     method UNIT()        { $!UNIT }
     method Mu()          { $!Mu }
+    method Mu'U()        { $!Mu'U }
     method Junction()    { $!Junction }
     method Any()         { $!Any }
     method Block()       { $!Block }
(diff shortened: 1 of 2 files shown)


vrurg:
Are you sure that an nqp op + lexical is faster than attribute access? Besides, I'm not sure that cases of using `Mu:U` are so common in the wild that this kind of micro-optimization would have any observable effect.

Anyway, considering how time is running out for the legacy compiler, I wouldn't be nitpicking. :)

vrurg (approved):

ab5tract:
> Are you sure that an nqp op + lexical is faster than attribute access? Besides, I'm not sure that cases of using `Mu:U` are so common in the wild that this kind of micro-optimization would have any observable effect.
> 
> Anyway, considering how time is running out for the legacy compiler, I wouldn't be nitpicking. :)

It mirrors what we do in the dispatchers.nqp, which is why I went that direction. It probably does only make a difference in the initial type check against the RHS, if it makes a difference at all.

I'm not attached to it, so if you feel like these should be dropped I'm fine with it.

But we do for example stash the result of `nqp::how` in a scalar just above this, so it also seemed reasonable from that angle. (By the way, I couldn't use it in this case, only `.HOW` returned the proper result -- maybe that's expected when dealing with definites? or else it's a subtle bug in the `nqp::how` implementation).

vrurg:
> But we do for example stash the result of `nqp::how` in a scalar just above this, so it also seemed reasonable from that angle. (By the way, I couldn't use it in this case, only `.HOW` returned the proper result -- maybe that's expected when dealing with definites? or else it's a subtle bug in the `nqp::how` implementation).

`HOW` is `nqp::how`. So, I wonder how they could be different?

niner:
nqp::how or nqp::how_nd?

ab5tract:
I must have been wielding it wrong (I suspect I typo'd it and was sending `$sm_type_how` to `archetypes`.

Anyway, it's been adjusted in fdeb87d04bbefb2b5a3ac0e5cb7265706f66416c.

PR #5656, opened 2024-10-08
Remove playful "just yet" in error messages.

coke:
It's not explicitly checked for in test/spectest


lizmat:
OOC, this hasn't been an issue since 27 April 2015 (34f79928e13ba0051) so why now?

coke:
Saw it in IRC chat for 'no scrubs'. The just yet makes sense if it's something we're going to implement, not for a random string.

Issue #3636, opened 2020-04-21
uniname on numerics vs on strs (.chr is now a bit too sensitive?)


AlexDaniel:
These two lines used to give the same output:
```raku
say 0x11FFFF.uniname;
say 0x11FFFF.chr.uniname;
```
Output:
```
<unassigned>
<unassigned>
```

But now there is a difference:
```
<unassigned>
chr codepoint 1179647 (0x11FFFF) is out of bounds
  in block <unit> at /tmp/MbxvOErd4L line 2
```

Output on all releases: https://gist.github.com/9629f58aeb068e1bee4be01f5c5194e4

Bisected: https://github.com/MoarVM/MoarVM/commit/15e94eecb30b850497e847c09d6b31130b775bdc

Uncovered by [these whateverable tests](https://github.com/Raku/whateverable/blob/789d525aad3543bedb549f01bb8785633c629572/xt/unicodable.t#L118-L121).

Ping @samcv @jnthn

AlexDaniel:
To be clear, I don't know what should be the right behavior, I'm just pointing out that there is now a difference. If it's decided that the current behavior is OK I can just change the test in whateverable.

jnthn:
Allowing `chr` to have out-of-range codepoints is just asking for a whole nest of bugs (such as the one that triggered this, where you could overflow the integer to a negative number, and it was then considered a synthetic internally). To be clear, this is only for codepoints that are out of the Unicode range *entirely*, not just those that aren't assigned a character yet but are in the supported range.

AlexDaniel:
@jnthn sure, but then maybe `0x11FFFF.uniname` should also produce an exception?

AlexDaniel:
@jnthn sure, but then maybe `0x11FFFF.uniname` should also produce an exception?

jnthn:
@AlexDaniel Passing in such a thing probably indicates confusion, yes...

samcv:
> @AlexDaniel Passing in such a thing probably indicates confusion, yes...

do you think the check should happen in MoarVM for `.uniname` or in Rakudo? (or both)

AlexDaniel:
Related: https://colabti.org/irclogger/irclogger_log/raku-dev?date=2020-04-26#l101

Altai-man:
There is a behavior change, but it is about handing a buggy situation and we can reach some consensus after the release is done.
An exception we throw now is already a step in the right direction, imo.

jnthn:
[...]

PR #5658, opened 2024-10-13
Make Int.uniname return a Failure on unassigned codepoints

lizmat:
As discussed in https://github.com/rakudo/rakudo/issues/3636

This however breaks tests in t/spec/S15-unicode-information/uniname.t so a decision needs to be made.


ab5tract:
This is just one of many issues with the current uni* infrastructure.

I think it's probably worth more to introduce a comprehensive overhaul here.

lizmat:
babysteps?

lizmat:
Implemented in 6.e with https://github.com/rakudo/rakudo/commit/3b8e50942a

PR #5660, opened 2024-10-15
Make Instant.DateTime coercer take timezone into account

lizmat:
As implicitely suggested in R#4169

No spectest fail, but still feels like it needs more scrutiny before being merged

### DIFF
--- a/src/core.c/Instant.rakumod
+++ b/src/core.c/Instant.rakumod
@@ -55,8 +55,12 @@ my class Instant is Cool does Real {
     method Int   (Instant:D: --> Int:D) { self.to-nanos div 1000000000                         }
     method narrow(Instant:D:          ) { self.tai.narrow                                      }
 
-    method Date(Instant:D:     --> Date:D)     { Date.new(self)     }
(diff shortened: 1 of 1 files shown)


raiph:
@lizmat

Aiui this PR is a relatively simple modification of an existing `DateTime` coercion method declared in the `Instant` class. It will immediately begin changing the result of any extant calls of the method.

Aiui a similar change you made a few weeks ago for a [`Date.DateTime` method](https://github.com/rakudo/rakudo/commit/233b6871fd35e9946353d65f8284be47a0dd56ce) is a significantly more complex change. It's _revision gated_ and is implemented as an `augment` in a `6.e` `Fixups` class.

At a guess this all follows from the two changes having distinct underlying rationales about back compat. Specifically, the change in this PR is considered to be a bug fix that will break back compat, but is on balance worth introducing asap, whereas the `Date.DateTime` coercion is considered to be a bug fix that needs to wait until the `6.e` shift. Is that about right?

lizmat:
Implemented for 6.e only in https://github.com/rakudo/rakudo/commit/e4157a41d8

PR #5661, opened 2024-10-15
[JVM] Unbreak module loading

usev6:
For some reason the check for nqp::null doesn't work and the method returns NQPMu. Maybe TWEAK didn't run correctly?

I didn't want to undo all (or more changes) of efeb9d7742. Hopefully the bandaid can be removed at one point.


usev6:
The underlying reason for this problem seems to be in NQP: https://github.com/Raku/nqp/issues/828

Issue #4273, opened 2021-03-29
LTA message: $a instead of $_


cognominal:
When fooling around, I commited a brain fart. That uncovered a LTA message. The message is about $a instead of $_.
Thx to raydiak on #raku-dev.

Probably the use of $_ in a loop *and* in enclosing code should trigger a warning in a linter.
But that would be too much to ask from a compiler.
On the other hand is it possible to indicate $_ was implicitly bound readonly in the loop ?

The golfed code and the error message
```
$_='toto'; $_++ for 0..1
Cannot resolve caller postfix:<++>(Int:D); the following candidates
match the type but require mutable arguments:
    (Mu:D $a is rw)
    (Int:D $a is rw --> Int:D)

The following do not match for other reasons:
    (Bool:D $a is rw)
    (Bool:U $a is rw --> Bool::False)
    (Mu:U $a is rw)
    (Num:D $a is rw)
    (Num:U $a is rw --> 0e0)
    (int $a is rw --> int)
    (num $a is rw --> num)
  in block <unit> at <unknown file> line 1

```

lizmat:
The message is just introspecting the candidates:

````
.signature.say for &postfix:<++>.candidates;
####
(Mu:D $a is rw)
(Mu:U $a is rw)
(Int:D $a is rw --> Int:D)
(int $a is rw --> int)
(Bool:U $a is rw --> Bool::False)
(Bool:D $a is rw)
(Num:D $a is rw)
(Num:U $a is rw --> 0e0)
(num $a is rw --> num)
````
And if you look at the code, the name of the variable accepting the argument *is* actually `$a`.

Now one could argue a better name could have been used in the code.

jnthn:
In principle the compiler knows that:

* Iterating a range produces immutable `Int` values, at least if `..` is not overridden
* The `Int` candidate for `postfix:<++>` needs something that `is rw` (again, overrides aside)

So this could be reported at compile time.

cognominal:
> The message is just introspecting the candidates:
> 
> ```
> .signature.say for &postfix:<++>.candidates;
> ####
> (Mu:D $a is rw)
> (Mu:U $a is rw)
> (Int:D $a is rw --> Int:D)
> (int $a is rw --> int)
> (Bool:U $a is rw --> Bool::False)
> (Bool:D $a is rw)
> (Num:D $a is rw)
> (Num:U $a is rw --> 0e0)
> (num $a is rw --> num)
> ```
> 
[...]

PR #5663, opened 2024-10-16
Make positional parameters anonymous in candidate list

lizmat:
when throwing an dispatch exception.  Because the actual names ofi positional parameters is not important and can be distracting.

Inspired by https://github.com/rakudo/rakudo/issues/4273

### DIFF
--- a/src/core.c/Exception.rakumod
+++ b/src/core.c/Exception.rakumod
@@ -3340,7 +3340,9 @@ my class X::Multi::NoMatch is Exception {
     has $.dispatcher;
     has $.capture;
     method message {
-        my @cand = $.dispatcher.dispatchees.map(*.signature.gist);
+        my @cand = $.dispatcher.dispatchees.map(
+          *.signature.gist.subst(/ <!after \:> <[$@%&+]> <( <[\w-]>+ /, :g)
+        );
         my @un-rw-cand;
         if first / 'is rw' /, @cand {
             my $rw-capture = Capture.new(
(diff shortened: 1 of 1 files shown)


MasterDuke17:
I'm not sure I agree. I think the names frequently are useful. Here's a completely random example:
```
[dan@alexandria roast]$ re 'say "abc".split(Mu)'
Cannot resolve caller split(Str:D: Mu:U); none of these signatures matches:
    (Str:D $:: Regex:D $regex, $limit = Whatever;; :$v, :$k, :$kv, :$p, :$skip-empty, *%_ --> Seq:D)
    (Str:D $:: Str(Cool) $match;; :$v is copy, :$k, :$kv, :$p, :$skip-empty, *%_ --> Seq:D)
    (Str:D $:: Str(Cool) $match, $limit is copy = Inf;; :$v is copy, :$k, :$kv, :$p, :$skip-empty, *%_)
    (Str:D $:: @needles, $parts is copy = Inf;; :$v is copy, :$k, :$kv, :$p, :$skip-empty, *%_ --> Seq:D)
  in block <unit> at -e line 1
```
Would `@needles` just be `@` with this PR?

lizmat:
> Would @needles just be @ with this PR?
yes

If you would be against this PR, what you're saying is that we should be using more meaningful variable names for positional parameters?

niner:
I also think that parameter names convey important information about those candidates. Maybe not in all cases, but definitely in some.

PR #5666, opened 2024-10-16
[JVM] Flatten without overflow/underflow issues

usev6:
If I'm not mistaken, the code relies on a $level parameter of -1 being coerced to a big positive number (due to the usage of uint). This doesn't work for the JVM backend.

Since the commit that introduced the uint parameter, https://github.com/rakudo/rakudo/commit/74cfe8e373, mentioned some confusion for the expression JIT on Intel, I didn't dare to change the code for MoarVM.

### DIFF
--- a/src/core.c/Rakudo/Iterator.rakumod
+++ b/src/core.c/Rakudo/Iterator.rakumod
@@ -1874,7 +1874,12 @@ class Rakudo::Iterator {
     method Flat(\iterator, $levels, $decont) {
         Flat.new(
           iterator,
+#?if !jvm
           nqp::istype($levels,Whatever) || $levels == Inf ?? -1 !! $levels.Int,
+#?endif
+#?if jvm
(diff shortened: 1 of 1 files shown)


usev6:
This should make `is-run` work again on the JVM backend. Obviously, that would be very helpful to detect other problems.

MasterDuke17:
Since the expression JIT has been disabled, what happens if https://github.com/rakudo/rakudo/commit/74cfe8e373 is reverted?

usev6:
Hmm, it looks like there were other changes later on to the relevant code, so a `git revert 74cfe8e373` didn't apply cleanly. I didn't have the time to look into the resulting conflict. But also, I don't understand what the problem with the expression JIT really was. Maybe @lizmat remembers the details?

lizmat:
What I remember, is that it caused some crashes.

And when run without it, it was found that in most cases the code was executing **faster**.

So it's now opt-in, with the idea to remove the code altogether at some point in the future unless someone comes along to fix the expression JIT.

PR #5674, opened 2024-10-30
when starting a Thread, also set its thread name

timo:
uses the thread ID (rakudo internal) + the given name

Before:
screenshot of gdb's command "info threads" showing 16 threads, most of which are named simply "rakudo-m", one is "spesh optimizer", one is "async io thread"
screenshot of perf report, where the "command" column shows mostly "rakudo", or "spesh optimizer" sometimes

After:
screenshot of gdb's command "info threads" where rakudo threads are identified with a number followed by GeneralWorke, TimerWorker, Supervisor, but also threads without numbers, like async io thread, or spesh thread. the very first one is named simply "rakudo-m".
screenshot of "perf report" showing results from different threads with their names in the "command" column

(note that the length of thread names is limited by the system. for example, on linux the library we use for this (pthreads) writes to `/proc/thread-self/comm` which the kernel limits to 15 bytes + null terminator (see `man 5 proc_pid_comm`)

### DIFF
--- a/src/core.c/Thread.rakumod
+++ b/src/core.c/Thread.rakumod
@@ -53,7 +53,9 @@ my class Thread {
 
         my $entry := anon sub THREAD-ENTRY() {
             my $*THREAD = self;
+#?if !jvm
             nqp::setthreadname(nqp::threadid($!vm_thread) ~ ": " ~ $!name);
+#?endif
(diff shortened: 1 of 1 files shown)


patrickbkr (approved):
I really like this!

Issue #4378, opened 2021-05-25
`Git for Bash` has no TTY, so raku opens in "non-interactive" mode


coke:
## The Problem

running "raku" inside git bash hangs - the REPL never appears

## Environment
* Operating system:
* Compiler version (`perl6 -v` or `raku -v`):

```
$ raku --version
Welcome to Rakudo(tm) v2021.04-22-g87d019f07.
Implementing the Raku(tm) programming language v6.d.
Built on MoarVM version 2021.04.
```

lizmat:
What does `raku -e 'say $*IN.t'` say?

raiph:
Similar problem for `cygwin` on Windows 8. For some further details, see [my comment about this on SO](https://stackoverflow.com/questions/67744623/raku-repl-appears-unresponsive-on-cygwin-under-windows-8#comment119761746_67744623).

leobm:
Hi @coke @lizmat, 
what I found interestingly is that if I use as git bash as terminal in Visual studio code, raku runs without problems. Then even the unicode character set is displayed correctly.


```
{
    "terminal.integrated.shell.windows": "D:\Program Files\Git\bin\bash.exe",
}
```

sumanstats:
Even typing `python` in the git bash won't bring `python` interpreter. And `Ctrl+C` won't exit the repl. We have to explicitly close the git bash. So I think its not a `raku` specific issue.
![raku](https://user-images.githubusercontent.com/8075886/125487169-038ebc87-2bdc-47ca-be15-94097fbee6ef.gif)

lizmat:
Has anybody tried running with `--repl-mode=interactive` ?

ab5tract:
> It seems this is an issue in Git Bash. If it actually is, then I think we shouldn't try to implement bandaids on our side.

Agreed. However, we should probably throw an exception explaining that the repl cannot run in the current environment instead of hanging.

patrickbkr:
Do I understand correctly, that the REPL has a non-interactive mode that is active when we don't have a terminal attached? (I'd be curious what the use case is for that.)
If so and if we want to keep that functionality, then I think we'll have a hard time differentiating the broken Git Bash situation from a legitimate non-interactive REPL situation.

raiph:
@lizmat:

> Has anybody tried running with --repl-mode=interactive ?

[...]

PR #5679, opened 2024-11-11
Fail gracefully (sometimes) when REPL is initialized without TTY

ab5tract:
There is a widespread issue in "Bash for Git for Windows" in that it is not in itself a proper TTY. The proposed solution for docker, node, and others is to run the command again under `winpty`.

We also want this to fail in other scenarios, such as if `repl()` would be called inside of Comma's hand-rolled non-TTY "REPL". In those cases where `winpty` would not be a usable solution, the recommendation is to catch the exception and re-throw with a different (or blank) solution.

Address R#4378 (#4378)


src/core.c/REPL.rakumod
```diff
@@ -315,6 +331,10 @@ do {
         }
 
         method init(Mu \compiler, $multi-line-enabled --> Nil) {
+            if not $*IN.t {
```
lizmat, line 334: s/if not/unless/  ?

src/core.c/REPL.rakumod
```diff
@@ -315,6 +331,10 @@ do {
...
+                die X::REPL::InvalidEnvironment.new: reason => "Unable to initialize REPL outside of a TTY";
```
lizmat, line 335: X::...new(...).throw  ??

ab5tract: Fixed!

src/core.c/REPL.rakumod
ab5tract: I've been in Java/Kotlin land too long!

ab5tract: Fixed. Though I prefer leaving it as a statement+block rather than as a statement modifier in this instance.

ab5tract:
Fixed with [this change in NQP](https://github.com/Raku/nqp/pull/827).

src/core.c/REPL.rakumod
```diff
@@ -315,6 +331,10 @@ do {
         }
 
         method init(Mu \compiler, $multi-line-enabled --> Nil) {
+            if not compiler.is-interactive || $*IN.t {
```
patrickbkr, line 334: Ow man, the 'not' is sneaky without parens.

patrickbkr: Wait no, was this meant to be `compiler.is-interactive && !$*IN.t`?

patrickbkr:
I think I don't understand how this is meant to work. Over in NQP `src/HLL/Compiler.nqp`, when not explicitly requesting a repl-mode, the REPL is activated only when no args are passed and it sees STDIN to be a terminal. If I understand correctly, the code here tries to catch the situation where we want to start a REPL but STDIN is *not* a tty. How would this code ever be triggered?

In general I don't understand how we want to differentiate the broken Git Bash on Windows situation from the non-REPL situation. When only looking at `$*IN.t` the two look identical. (To detect this we would have to check what is connected to the other end of $*IN and whether that thing is an interactive Cygwin/MinGW shell. I believe we should not try to do that.)

ab5tract:
Please look at the code more carefully.

'--repl-mode=interactive' is used by our Test::Helpers to run the repl in a Proc.

My suggestion is to make this much more clear. 

But I don't know if I have the energy to keep fighting tooth and nail about this stuff, so I won't be doing yet more explaining myself.

src/core.c/REPL.rakumod
ab5tract: No, I was avoiding parens. I find 'if not' to be less "sneaky" than 'unless' when paired with other conditional operators.

But since there is controversy, I will change it.

ab5tract:
> When only looking at $*IN.t the two look identical

This is not accurate. Bash for Git is not presenting as a TTY, which is the reason for this problem in the first place.

patrickbkr:
Ah. with non-REPL situation I mean the following: `echo "say 'hi'" | raku`. In this case `$*IN.t` is false, just as in Bash for Git.

src/core.c/REPL.rakumod
ab5tract: I think it's much clearer in the latest iteration too:

```
if compiler.repl-mode eq 'tty' && not $*IN.t {
```

ab5tract:
> Ah. with non-REPL situation I mean the following: `echo "say 'hi'" | raku`. In this case `$*IN.t` is false, just as in Bash for Git.

I see what you mean now. The REPL code-path is not invoked via this channel. Per the docs:

>  -                    read program source from STDIN or start REPL if a TTY

Since, as you point out, `$*IN.t` is False in this case, there is no attempt to load any "REPL machinery".

```
> echo "say 'hi'" | ./rakudo-m
hi
```

I've pushed updates to this PR and the [PR for NQP](https://github.com/Raku/nqp/pull/827) that replace the (IMO horribly named) `interactive|non-interactive` repl-mode options as previously discussed in the commit from NQP's PR.

ab5tract:
I had not seen your other message earlier, sorry that I missed it.

> I'm obviously draining your energy here. I'm sorry for that. It's just that I'm very convinced that with these changes there won't be an error message when running a plain `raku` without arguments in a Git Bash on Windows. I think it will hang, as it did before. I'll give this a try and report back.

I really shouldn't reply when < 5 minutes of waking up (of course I'm tired in that moment!). I did eventually strike out the most emotional reaction.

Anyway, I appreciate that you are willing to take the time to test this. It works as expected in Comma, so I'm curious as to what the outcome in Git for Bash will be. (Just make sure you are using the default install options and not the configuration that fixes this by using a real TTY to begin with :wink: ).

patrickbkr:
I've tried it now. A plain `raku` hangs just as before in Git-Bash on Windows.

ab5tract:
@patrickbkr Thanks for running this! It's very clear now why this patch won't ultimately address the problems with `Bash for Git`.

However, I think it is still useful for detecting at least cases where `repl()` is invoked outside of a TTY (a failure mode that can theoretically be encountered in Comma's REPL, for example).

But IMO, the real main reason to/advantage of adopting this PR is the adjustment to more legible `--repl-mode` options.

patrickbkr:
Let's do this.

patrickbkr (approved):

Issue #5686, opened 2024-11-19
Junctions don't autothread over negative operators


hwayne:
## The Problem

```
[15] > my $m = <1 2 3>.one
one(1, 2, 3)
[16] > $m != 0
True
[17] > $m == 0
one(False, False, False)
```

## Expected Behavior

`$m != 0` should be `one(True, True, True)`, which should `so` to false.

## Actual Behavior

`$m != 0` is just `True`.

Similarly, `>` autothreads and `!>` does not.

## Environment
* Operating system: Windows 11
* Compiler version (`rakudo -v` or `raku -v`):

```
Welcome to Rakudo™ Star v2024.10.
Implementing the Raku® Programming Language v6.d.
Built on MoarVM version 2024.10.
```

timo:
i believe at least part of this is related to the discussion in https://github.com/Raku/problem-solving/issues/319

hwayne:
> Why the result of $m != 0 is True rather than False, I'm not sure.

It makes sense `$m != 0` would be true then, since it's not true that *exactly one* element of that list equals zero

ab5tract:
@hwayne Consider weighing in on that problem-solving ticket, if you have the time/impulse. That might kickstart the discussion and resolution of this.

A few other related items that @timo++ found:

- #1787 (`infix:<!eq> shouldn't differ from infix:<ne> when applied to junctions in Rakudo`)
- #3748 (`any(0,4) != 0 is False`)
- #3874 (`Remove special handling of ne and != wrt Junctions`)

ab5tract:
In #5690 I've adjusted the semantics of both `!=` and `METAOP_NEGATE`:

The implementation of `METAOP_NEGATE` demonstrates this best:

```
# Before
 -> |c { c.elems > 1 ?? not op.(|c) !! True }
# After
 -> |c { c.elems > 1 ?? op.(op.(|c), False) !! True }
```

It's revision gated, so you'll need to have `use v6.e.PREVIEW` going on.

PR #5690, opened 2024-11-22
Address long-standing discrepancy for != and METAOP_NEGATE

ab5tract:
**Note: This represents a significant but (AFAICT) logically equivalent adjustment to the definition of METAOP_NEGATE**

This patch uses revision-gate to migrate v6e to a version of negation that preserves junctions while accomplishing the same underlying functionality as planting `not` in front of `$expr`.

Previously:

    one(0,1) != 0 # False

Now:

    one(0,1) != 0 # one(Bool::False, Bool::True)

Also fixed in this patch is the ability to add a revision-gated method with the same signature as an already existing method. In such cases, the existing method is also revision-gated, but to the revision that it was introduced (in practice this will usually be the same revision as the proto is gated to).

Addresses R#5686 (#5686) / R#3748 (#3748) / R#3874 (#3874)

See also discussion at:
    https://github.com/Raku/problem-solving/issues/319

Note: Adopting this change will require adjustments to roast.

### DIFF
--- a/src/Perl6/Compiler.nqp
+++ b/src/Perl6/Compiler.nqp
@@ -256,8 +256,10 @@ class Perl6::Compiler is HLL::Compiler {
     }
 
     method ast($match, *%adverbs) {
+        self.panic("Unable to obtain AST from parse result (no match object)")
+            unless nqp::isconcrete($match);
         my $ast := $match.ast;
-        self.panic("Unable to obtain AST from parse result")
+        self.panic("Unable to obtain AST from parse result (no AST on match object)")
             unless nqp::isconcrete($ast);
         $ast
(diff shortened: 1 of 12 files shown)


ab5tract:
timo++ noted that `lt` and `gt` are not going to behave well with this current approach.

ab5tract:
`&infix:<ne>` was being very troublesome to apply revision gating to, so it has yet to be addressed.

ab5tract:
`&infix:<ne>` is currently implemented with the new logic (`(a eq b) eq False`) without any revision gating.

coke:
_Moved Draft notice from title to status_

PR #5693, opened 2024-11-24
Don't use Match::chomp on string

usev6:
This avoids the following error on the JVM backend when calling .gist on Match objects:

  Cannot dispatch to method chomp on Match because it is not inherited
  or done by Str

I don't understand how this works on MoarVM, but using an unqualified method call looks reasonable to me.


usev6:
I didn't see any new problems in `make test` and `make stresstest`, so hopefully this doesn't break things.

usev6:
To make progress with the test failures on the JVM backend I plan to merge this in the near future, even though there was no approval/review.

PR #5696, opened 2024-11-25
Add lower-bounding for the revision-gated trait

ab5tract:
This adds a finishing touch to revision gating: the ability to use revision gating to evolve methods that otherwise have the exact same signature.

The is done by adding the trait to the candidate that is to be replaced:

    ```
    # protos always require the minimum version, this is how dispatch
    # knows to look for revision gated candidates
    proto sub foo(|) is revision-gated("6.c")
    # multis with the same signature as their replacement also need
    # the trait
    multi sub foo(Int $x) is revision-gated("6.c") { say "old x: $x" }
    # the replacement is revision gated to the version in which it
    # is introduced
    multi sub foo(Int $x) is revision-gated("6.e") { say "new x: {$x+1}" }
    ```

This leads to the following behavior:

    ```
    foo(555)
    #=> old x: 555
    ```

    ```
    use v6.e.PREVIEW;
    foo(555)
    #=> new x: 556
    ```

### DIFF
--- a/src/Perl6/bootstrap.c/BOOTSTRAP.nqp
+++ b/src/Perl6/bootstrap.c/BOOTSTRAP.nqp
@@ -3889,10 +3889,13 @@ BEGIN {
             nqp::scwbenable;
         }
 
+
+#?if !moar
         my $caller-revision := nqp::getcomp('Raku').language_revision;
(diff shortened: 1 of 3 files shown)


jonathanstowe:
What would be nice is if the plumbing that allows a trait to affect the candidate selection could be generalised so that kind of thing could be set up in user space.

Issue #5703, opened 2024-11-28
[JVM] Spectest failures because types in simple protos are checked strictly


usev6:
## The Problem
Running the test file https://github.com/Raku/roast/blob/3032bcf423/integration/advent2011-day07.t on the JVM backend gives an error:
```
===SORRY!=== Error while compiling /path/to/rakudo.jvm/t/spec/integration/advent2011-day07.t
Type check failed in binding to parameter '<anon>'; expected Any but got ProfiledGrammarHOW (ProfiledGrammarHOW)
at /path/to/rakudo.jvm/t/spec/integration/advent2011-day07.t:5
```

I don't know when this started to fail, but it worked about a year ago.

I think that the above spectest failure can be golfed down to
```
$ ./rakudo-j -e 'proto sub foo ($) {*}; multi sub foo(Mu $a) { say "hi" }; foo(Mu)'
Type check failed in binding to parameter '<anon>'; expected Any but got Mu (Mu)
  in block <unit> at -e line 1
```

On MoarVM this does not error out:
```
$ ./rakudo-m -e 'proto sub foo ($) {*}; multi sub foo(Mu $a) { say "hi" }; foo(Mu)'
hi
```

After [a hint](http://irclogs.raku.org/raku-dev/2024-11-27.html#08:02) from @niner it was clear that MoarVM just ignores the types parameters in the signatures of protos for simple cases like this. The JVM backend on the other hand does the full type check and detects a mismatch in this case, because a plain `$` is interpreted as `Any $` -- which doesn't accept arguments of type `Mu`.

I didn't find a hook where I could add similar code like https://github.com/rakudo/rakudo/blob/ba3b24cff2/src/vm/moar/dispatchers.nqp#L1806-L1887 for the JVM backend. So I'll look which type signatures in the core need to be adjusted to avoid spectest failures. (I think it will be very few -- or only one.) I opened this issue mostly for documentation purposes.

PR #5704, opened 2024-11-28
Type second param of proto for Stash::ASSIGN-KEY

usev6:
As explained in https://github.com/rakudo/rakudo/issues/5703 the current version leads to a spectest failure on the JVM backend. Running https://github.com/Raku/roast/blob/3032bcf423/integration/advent2011-day07.t eventually Stash::ASSIGN-KEY is called with something that derives from Metamodel::GrammarHOW (and doesn't type match with Any) as the second argument.

Typing the second parameter as Mu solves this problem -- and it also matches the types used in the multi variants of Stash::ASSIGN-KEY.


usev6:
To make progress with the test failures on the JVM backend I plan to merge this in the near future, even though there was no approval/review.

PR #5708, opened 2024-12-01
[JVM] Decont Routine object when looking for attribute

usev6:
Without this running running S32-list/skip.t from roast failed with a RuntimeException:

  No such attribute '@!dispatchees' for this object

This is a somewhat golfed version of the failing code, which also blows up on MoarVM.

  use nqp;
  BEGIN my (&plan = do { use Test; (&plan) };
  nqp::getattr(&plan,Routine,'@!dispatchees');


usev6:
This one is now also nearly one week old. The code is only used on the JVM backend, so I'll go on and merge this without an approval. There's always the option to revert ...

PR #5710, opened 2024-12-02
[JVM] Add extra check for GLOBAL being null

usev6:
This fixes a NullPointerException on the JVM backend for code like

  use v6.c.UNKNOWN

That code does not blow up on MoarVM, which handles calling "WHO" on a null object gracefully.


usev6:
This is an alternative to https://github.com/rakudo/rakudo/pull/5379. Actually I'd prefer the old PR, but this is better than nothing. (And it should also fix https://github.com/rakudo/rakudo/issues/5180.)

usev6:
Since this only affects the JVM backend I plan to go ahead and merge it in the near future without a review. (I've already closed my old PR #5379 that added the extra check for all backends.)

PR #5711, opened 2024-12-02
Move test file that dies on JVM to t/09-moar

usev6:
This is really just a bandaid to make "make test" usable again. The test doesn't work on the JVM backend: Even if I replace the usage of atomicint with a plain int, there is a compile time error:

  Type check failed for return value; expected Mu but got BASE (BASE)

As noted in the leading comment of the test file, the final destination of the test should be roast, anyway. So hopefully moving it to t/09-moar is good enough for now.


usev6:
I'm totally open for better solutions, but I've no idea how to fix the underlying problem.

usev6:
If nobody comes up with a better idea, I'll merge this in the near future. Having the test file in a different directory shouldn't be a big problem. And there's always the option to revert ...

PR #5712, opened 2024-12-02
Fix type when binding attribute for RealPolymod

usev6:
On the JMV backend this led to spectest failures in S32-num/polymod.t, like:

  $ ./rakudo-j -e 'say 10e0.polymod(1.5)'
  java.lang.RuntimeException: No such attribute '$!numerator' for this object
    in block <unit> at -e line 1

On MoarVM I didn't notice a problem, but this code lives in the class RealPolymod, so using IntPolymod looks wrong. (Likely it was just a copy-pasto.)


lizmat:
Fixed manually in https://github.com/rakudo/rakudo/commit/22a5b0ec3e thanks for the catch!

Sorry it fell through the cracks initially  :-(

usev6:
No problem, thanks for fixing :)

PR #5713, opened 2024-12-02
Provide adverbial forms of map

ab5tract:
This commit opens a different way of specifying map variants. This is mostly for cosmetic reasons and operates under the impression that named argument adverbs, had they been available, would have been used over the smashedcase versions we currently offer.

* [flatmap](https://github.com/Raku/problem-solving/issues/430)

### DIFF
--- a/src/core.c/Any-iterable-methods.rakumod
+++ b/src/core.c/Any-iterable-methods.rakumod
@@ -100,6 +100,20 @@ Consider using a block if any of these are necessary for your mapping code."
               !! Rakudo::IterateMoreWithoutPhasers.new(&code, $source, $count, $label)
     }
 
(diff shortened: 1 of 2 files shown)


lizmat:
Feels this needs a problem-solving issue to accompany this PR.

ab5tract:
Here's [a comment ](https://github.com/Raku/problem-solving/issues/430#issuecomment-2321874879)from a problem-solving topic already linked.

> I do wonder whether these would all have been adverbial multis for map if adverbs had been implemented when they were introduced. There's something potentially satisfying to me about having them all under one umbrella @a.map: :flat, *.foo. But I expect that this ship has long sailed

This PR was in recognition that maybe this ship had not sailed already, in fact.

lizmat (approved):

PR #5714, opened 2024-12-04
[JVM] Reduce backend-specific code in find_best_dispatchee

usev6:
This is a follow-up to fa853abc7f which reverted 99f52bcd75 for the JVM backend.

From what I understand the cause of the breakage on the JVM was the the removal of the call to nqp::hllizefor in 99f52bcd75. I'm not sure why it isn't needed for MoarVM, esp. since we have such a call in the similar code in https://github.com/rakudo/rakudo/blob/dd45859fea/src/vm/moar/dispatchers.nqp#L2397.

Anyway, this patch shouldn't change anything for MoarVM. For the JVM there are two small changes, which are hard to spot in the diff. Therefore I'll note them explicitly:

1) In line 4032 we use $arg now as the first argument to nqp::istype.
   The old code used $param there.

2) Similarly in line 4051 we call nqp::isconcrete with $arg now, while
   we checked $param before.

I didn't see any new test failures, so these changes are hopefully ok.

### DIFF
--- a/src/Perl6/bootstrap.c/BOOTSTRAP.nqp
+++ b/src/Perl6/bootstrap.c/BOOTSTRAP.nqp
@@ -3962,12 +3962,12 @@ BEGIN {
 
                         # Not a native parameter, but we may getting a native
                         # value, possibly in a container.
-#?if !jvm
                         else {
 
                             # Assume a native value unless proven otherwise
                             my int $primish := 1;
 
+#?if !jvm
                             # Set type to fall back to if no native involved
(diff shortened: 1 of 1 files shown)


usev6:
Since the code for MoarVM doesn't change, I plan to merge this one in the near future.

ab5tract:
> Since the code for MoarVM doesn't change, I plan to merge this one in the near future.

Any idea why all the checks are failing?

usev6:
I thought that was a permissions problem. I noticed the same problem with an earlier PR.

PR #5725, opened 2024-12-08
New script wrappers

patrickbkr:
This is WIP, but it already generates working native executable runners on Windows.

### DIFF
--- a/src/core.c/CompUnit/Repository/Installation.rakumod
+++ b/src/core.c/CompUnit/Repository/Installation.rakumod
@@ -14,24 +14,7 @@ class CompUnit::Repository::Installation does CompUnit::Repository::Locally does
     my constant @all-script-extensions =
         '', '-m', '-j', '-js', '.bat', '-m.bat', '-j.bat', '-js.bat';
 
-    my constant $windows-wrapper = Q/@rem = '--*-Perl-*--
-@echo off
-if "%OS%" == "Windows_NT" goto WinNT
-#raku# "%~dpn0" %1 %2 %3 %4 %5 %6 %7 %8 %9
-goto endofraku
-:WinNT
-#raku# "%~dpn0" %*
-if NOT "%COMSPEC%" == "%SystemRoot%\system32\cmd.exe" goto endofraku
-if %errorlevel% == 9009 echo You do not have Rakudo in your PATH.
-if errorlevel 1 goto script_failed_so_exit_with_non_zero_val 2>nul
-goto endofraku
-@rem ';
-__END__
-:endofraku
-/;
-
-    my constant $raku-wrapper = '#!/usr/bin/env #raku#
-sub MAIN(*@, *%) {
+    my constant $raku-wrapper-code = 'sub MAIN(*@, *%) {
     CompUnit::RepositoryRegistry.run-script("#name#");
 }';
 
@@ -279,20 +262,27 @@ sub MAIN(*@, *%) {
         # bin/ scripts
         for %files.kv -> $name-path, $file is copy {
             next unless $name-path.starts-with('bin/');
-            my $name        = $name-path.subst(/^bin\//, '');
-            my $id          = self!file-id(~$file, $dist-id);
+            my $name         = $name-path.subst(/^bin\//, '');
+            my $id           = self!file-id(~$file, $dist-id);
             # wrappers are put in bin/; originals in resources/
-            my $destination = $resources-dir.add($id);
-            my $withoutext  = $name-path.subst(/\.[exe|bat]$/, '');
+            my $destination  = $resources-dir.add($id);
+            my $withoutext   = $name-path.subst(/\.raku$/, '');
             for @script-postfixes -> $be {
-                $.prefix.add("$withoutext$be").IO.spurt:
-                    $raku-wrapper.subst('#name#', $name, :g).subst('#raku#', "rakudo$be");
+                $.prefix.add("$withoutext$be.raku").IO.spurt:
+                    $raku-wrapper-code.subst('#name#', $name, :g);
+                my $blob = generate-runner-blob(
+                  :program("<plain>rakudo$be.exe"),
+                  :args((
+                  '<abs-slash>' ~ $name.subst(/\.raku$/, '') ~ "$be.raku",
+                  '<cmd-args>',
+                  ))
+                );
                 if $is-win {
-                    $.prefix.add("$withoutext$be.bat").IO.spurt:
-                        $windows-wrapper.subst('#raku#', "rakudo$be", :g);
+                    $.prefix.add("$withoutext$be.exe").IO.spurt: $blob;
                 }
                 else {
-                    $.prefix.add("$withoutext$be").IO.chmod(0o755);
+                    $.prefix.add("$withoutext$be").IO.spurt: $blob;
+                    $.prefix.add("$withoutext$be").IO.chmod: 0o755;
                 }
             }
             self!add-short-name($name-path, $dist, $id);
@@ -743,6 +733,239 @@ and so the packages will not be installed in the correct location.
 Please ask the author to add a \"provides\" section, mapping every exposed namespace to a
 file location in the distribution.
 See http://design.raku.org/S22.html#provides for more information.\n";
+    }
+
+    # Runner generation logic starts here.
+
+    my %arg-op = missing   => 0,
+                 plain     => 1,
+                 cmd-args  => 2,
+                 plat-sep  => 3,
+                 abs       => 4,
+                 abs-slash => 5,
+    ;
+
+    sub validate-env-var($name) {
+        die "$name is not a valid env var name" unless $name ~~ / ^ <[ a..z A..Z _ ]> <[ a..z A..Z 0..9 _ ]>* $ /;
+    }
+
+    sub generate-posix(
+        Str           :$program!,
+        Str           :$cwd,
+                      :@args,
+                      :%env-add,
+                      :@env-remove,
+    ) {
+        my @call-parts = "exec";
+
+        sub sh-escape(Str $text) {
+            S/ ( \$ | \` | \" | \\ | \n ) / \\$1 / given $text;
+        }
+
+        sub process-path(Str() $path) {
+            if $path ~~ /
+                    "<" ( "plain" || "cmd-args" || "plat-sep" || "abs" || "abs-slash" ) ">"
+                    (.*)
+                / {
+                my $op = $0.Str;
+                my $val = $1.Str;
+                given $op {
+                    when "plain" | "plat-sep" {
+                        return '"' ~ sh-escape($val) ~ '"';
+                    }
+                    when "cmd-args" {
+                        return '"$@"';
+                    }
+                    when "abs" | "abs-slash" {
+                        return '"${DIR}/' ~ sh-escape($val) ~ '"';
+                    }
+                }
+            }
+            else {
+                note "Don't understand this arg: $path";
+                exit 1;
+            }
+        }
+
+        # Write program
+        unless $program ~~ / ^ "<" ( "plain" || "plat-sep" || "abs" ) ">" / {
+            note "Program is invalid. Must start with \"<plain>\", \"<plat-sep>\", \"<abs>\".";
+            exit 1;
+        }
+        @call-parts.push: process-path($program);
+
+        # Write args
+        for @args -> $arg {
+            @call-parts.push: process-path($arg);
+        }
+
+        my @lines;
+
+        # Write CWD
+        if $cwd {
+            unless $cwd ~~ / ^ "<" ( "plain" || "abs" ) ">" / {
+                note "CWD is invalid. Must start with \"<plain>\" or \"<abs>\".";
+                exit 1;
+            }
+            @lines.push: 'cd ' ~ process-path($cwd);
+        }
+
+        # Write env vars
+        for %env-add.keys.sort -> $key {
+            validate-env-var $key;
+            my $val = %env-add{$key};
+            @lines.push: "export $key=$val"
+        }
+
+        @env-remove.map: -> $name {
+            validate-env-var $name;
+            @lines.push: "unset $name"
+        }
+
+        @lines.push: @call-parts.join(" ");
+
+        
+        my $text = get-resource('posix-runner-tmpl').slurp;
+        $text ~= @lines.join("\n");
+        $text
+    }
+
+    sub get-resource($name) {
+        nqp::gethllsym('default', 'SysConfig').rakudo-home.IO.add("resources").add($name)
+    }
+
+    sub create-config-blob(
+        Str           :$program!,
+        Str           :$arg0,
+        Str           :$cwd,
+                      :@args,
+                      :%env-add,
+                      :@env-remove,
+    ) {
+        my $char-size = Rakudo::Internals.IS-WIN ?? 2 !! 1;
+        my $encoding = Rakudo::Internals.IS-WIN ?? 'utf16le' !! 'utf8';
+        my @end-str-bytes = Rakudo::Internals.IS-WIN ?? (0x00, 0x00) !! (0x00,);
+
+        my $marker = "EXEC_RUNNER_WRAPPER_CONFIG_MARKER".encode($encoding);
+
+        my Buf $cbuf .= new;
+        my Int $pos = 0;
+
+        sub write-str(Str() $str) {
+            my $bytes = $str.encode($encoding);
+            $cbuf.write-uint16: $pos, ($bytes.elems / $char-size + 1).Int;
+            $pos += 2;
+            $cbuf.append: $bytes;
+            $cbuf.append: @end-str-bytes;
+            $pos += $bytes.elems + $char-size;
+        }
+
+        sub write-path(Str() $path) {
+            if $path ~~ /
+                    "<" ( "plain" || "cmd-args" || "plat-sep" || "abs" || "abs-slash" ) ">"
+                    (.*)
+                / {
+                my $op = %arg-op{$0};
+                my $val = $1;
+
+                $cbuf.write-uint8: $pos, $op;
+                $pos += 1;
+
+                if ($op != %arg-op<cmd-args>) {
+                    write-str($val);
+                }
+            }
+            else {
+                note "Don't understand this arg: $path";
+                exit 1;
+            }
+        }
+
+        # Write marker
+        $cbuf.append: $marker;
+        $cbuf.append: @end-str-bytes;
+        $pos += $marker.elems + $char-size;
+
+        # Write program
+        unless $program ~~ / ^ "<" ( "plain" || "plat-sep" || "abs" ) ">" / {
+            note "Program is invalid. Must start with \"<plain>\", \"<plat-sep>\", \"<abs>\".";
+            exit 1;
+        }
+        write-path($program);
+
+        # Write arg0 set
+        $cbuf.write-uint8: $pos, ($arg0.defined ?? 1 !! 0);
+        $pos += 1;
+
+        # Write arg0
+        if $arg0.defined {
+            write-str($arg0);
+        }
+
+        # Write number of args
+        $cbuf.write-uint16: $pos, @args.elems;
+        $pos += 2;
+
+        # Write args
+        for @args -> $arg {
+            write-path($arg);
+        }
+
+        # Write CWD
+        if $cwd {
+            unless $cwd ~~ / ^ "<" ( "plain" || "abs" ) ">" / {
+                note "CWD is invalid. Must start with \"<plain>\" or \"<abs>\".";
+                exit 1;
+            }
+            write-path($cwd);
+        }
+        else {
+            $cbuf.write-uint8: $pos, %arg-op<missing>;
+            $pos += 1;
+        }
+
+        # Write number of env vars
+        $cbuf.write-uint16: $pos, %env-add.elems;
+        $pos += 2;
+
+        # Write env vars
+        for %env-add.keys.sort -> $key {
+            validate-env-var $key;
+            my $val = %env-add{$key};
+            write-str($key ~ "=" ~ $val);
+        }
+
+        # Write number of ignore env vars
+        $cbuf.write-uint16: $pos, @env-remove.elems;
+        $pos += 2;
+
+        # Write ignore env vars
+        for @env-remove.sort -> $val {
+            validate-env-var $val;
+            write-str($val);
+        }
+
+        $cbuf;
+    }
+
+    sub generate-runner-blob(
+        Str           :$program!,
+        Str           :$arg0,
+        Str           :$cwd,
+                      :@args,
+                      :%env-add,
+                      :@env-remove
+                      --> Blob
+    ) {
+        if Rakudo::Internals.IS-WIN {
+            my Buf $blob = get-resource("win-runner.exe-tmpl").slurp(:bin);
+            $blob.append: create-config-blob(:$program, :$arg0, :$cwd, :@args, :%env-add, :@env-remove);
+            $blob
+        }
+        else {
+            my $text = generate-posix :$program, :$cwd, :@args, :%env-add, :@env-remove;
+            $text.encode
+        }
     }
 }
 
--- a/src/runner/runner.c
+++ b/src/runner/runner.c
@@ -1,0 +1,395 @@
+#include <stdio.h>
+#include <stdlib.h>
+#include <string.h>
+#include <errno.h>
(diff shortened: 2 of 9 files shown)


src/core.c/CompUnit/Repository/Installation.rakumod
```diff
@@ -18,13 +19,13 @@ class CompUnit::Repository::Installation does CompUnit::Repository::Locally does
...
 
-
-    method TWEAK() {
+    method TWEAK(:$wrapper-mode) {
         $!lock       := Lock.new;
         $!loaded     := nqp::hash;
         $!seen       := nqp::hash;
         $!dist-metas := nqp::hash;
         $!precomp-store := $!precomp-stores := nqp::null;
+        self!config(:$wrapper-mode) with $wrapper-mode;
```
ugexe, line 28: Won't this open a file and potentially parse json even when no modules are loaded? I think we would want to avoid that since we have e.g. `LazyMetaReader` which is mostly convoluted logic around not having to parse json unnecessarily.

src/core.c/CompUnit/RepositoryRegistry.rakumod
```diff
@@ -175,10 +175,14 @@ class CompUnit::RepositoryRegistry {
             $home-spec = 'inst#' ~ $home;
         }
 
+        my $reloc = nqp::gethllsym('default', 'SysConfig').rakudo-build-config()<static-rakudo-home> eq '';
+        my $wrapper-mode = $reloc ?? 'relative' !! 'absolute';
```
ugexe, line 179: Should we use an enum for these?

src/core.c/CompUnit/Repository/Installation.rakumod
```diff
@@ -744,6 +808,239 @@ Please ask the author to add a \"provides\" section, mapping every exposed names
...
+        @lines.push: @call-parts.join(" ");
+
+        
+        my $text = get-resource('posix-runner-tmpl').slurp;
+        $text ~= @lines.join("\n");
+        $text
+    }
+
+    sub get-resource($name) {
+        nqp::gethllsym('default', 'SysConfig').rakudo-home.IO.add("resources").add($name)
```
ugexe, line 908: I wonder if there is a better name we could use than `resources`. That path location is already quite closer to the existing `resources` directories that repositories use while not being used for quite the same thing.

src/runner/runner.c
ugexe, line 1: Hmmm, it occurs to me that we would now be requiring a C compiler to install any raku script correct? That doesn't seem ideal, particularly on windows where I would expect precompiled released to be more popular (so more likely to have raku installed without having had e.g. MSVC installed to build it).

ugexe:
Do you know if a wrapper generated on debian would work on another debian system? For example people who package modules on e.g. debian also need to include the wrapper itself, so those packagers need to be able to include the wrapper generated on their build system and for that wrapper to work for users who install that package.

src/core.c/CompUnit/Repository/Installation.rakumod
patrickbkr: I'll have a look.

src/core.c/CompUnit/RepositoryRegistry.rakumod
patrickbkr: I don't know why, but I didn't manage to make enums work in this file. Bootstrap issues? Do you know if there is a way around that?

src/runner/runner.c
patrickbkr: No we don't. The code to handle this is effectively an import of the [Devel-ExecRunnerGenerator](https://sr.ht/~patrickb/Devel-ExecRunnerGenerator/). It compiles the executable once during the Rakudo build. Later on that executable file is primed with a configuration. The logic to do the priming is written in Raku. I'm pretty proud of this piece of software. 😊
There is a catch though: We now require a C compiler during the Rakudo build on Windows with the JVM and JS backends. Previously those didn't need a C compiler.

patrickbkr:
> Do you know if a wrapper generated on debian would work on another debian system? For example people who package modules on e.g. debian also need to include the wrapper itself in their .deb, so those packagers need to be able to include the wrapper generated on their build system and for that wrapper to work for users who install that package.

On non-Windows systems the wrappers are plain POSIX shell scripts. (Only on Windows we need the native executable magic.) They should be portable across systems without issue.

src/core.c/CompUnit/Repository/Installation.rakumod
patrickbkr: I've made it lazy as well.

src/core.c/CompUnit/Repository/Installation.rakumod
patrickbkr: Done. I went with `templates`.

src/core.c/CompUnit/Repository/Installation.rakumod
```diff
@@ -119,34 +120,34 @@ class CompUnit::Repository::Installation does CompUnit::Repository::Locally does
...
-                    wrapper-mode => $wrapper-mode // 'path',
-                ;
+                %!config = %!config-seed;
+                self!write-config if self!prefix-writeable;
             }
         }
         %!config
     }
 
     method change-wrapper-mode($mode) {
```
ugexe, line 147: I think we'd want to mark any publicly exposed methods as an implementation detail

src/core.c/CompUnit/Repository/Installation.rakumod
```diff
@@ -8,40 +8,25 @@ class CompUnit::Repository::Installation does CompUnit::Repository::Locally does
     has Int $!version;
     has $!precomp-store;    # cache for .precomp-store
     has $!precomp-stores;   # cache for !precomp-stores
+    has %!config;
+    has %!config-seed;
```
ugexe, line 12: I think this would be loading the same config json multiple times (once for each repo). We could change these to `my` but even that leaves me feeling like there has to be a better way of handling configuration centralization.

src/core.c/CompUnit/Repository/Installation.rakumod
patrickbkr: Agreed! I still have to learn many things about core hacking. :-)

src/core.c/CompUnit/Repository/Installation.rakumod
patrickbkr: In this PR the unit that is configured and for which we give the guarantee that all the wrappers conform to the config is a CURI. To me this seems to be the right unit.

Maybe it's just an inaccuracy in the language, but just to make sure: It's not the *same* config json that's loaded multiple times. Each of the three standard CURIs (core, site, vendor) and every other CURI there is has it's own config. The three standard CURIs are by default all set to the same wrapper mode (`absolute` in non relocatable Rakudos and `relative` in relocatable Rakudos), other CURIs are set to `path`. In principle even the standard CURIs could be configured differently from each other.

It sounds like there is a desire to have one single configuration for the three standard repos. Can you explain what the motivation for that desire is? Performance? Guaranteeing that they never run out of sync?

Setting this to `my` won't work. That would force all CURIs that happen to be loaded to the same wrapper mode. This makes it really difficult to give the guarantee that all the wrappers in all of those CURIs conform to the config. What would be the unit we configure then? System wide via a `~/.raku/curi_config.json`?

niner: What would be the use-case for this kind of configuration? The defaults sound very sane to me. When would I want to change them?

patrickbkr: For the standard CURIs I'm unsure if anyone would ever want to change the defaults. But custom CURIs (as in not one of the three standard CURIs) might well have different wrapper modes. E.g.

- A distributable installation including a relocatable Rakudo and one or more custom CURIs might want to have their wrappers be set to `relative`.
- A custom CURI in my home dir that I want to use with all Rakudos I happen to use , I might want to set to `path`.
- A custom CURI in my home dir with some app that I want to work reliably irrespective of the Rakudo currently in $PATH, I might want to set to `absolute` and point it at one specific Rakudo installation that I know works for this app.

niner: So configuration is (almost) exclusively relevant for custom CURIs. This means that the user already has to tell rakudo about these via command line switches or environment variables. Both already include a mechanism for passing options to a CUR's constructor:
`rakudo -I 'mycur#wrappers<relative>#/path/to/repo'`

 Is there any reason this is not sufficient for this one configuration switch you need here? Aside from performance considerations or the hurdle to find a place for a config file, using this existing mechanism has the advantage of the everything staying in a single place on the user's side.

patrickbkr:
I just realized two shortcomings in the current implementation.

1. In `relative` mode, the relative path is hard-coded assuming it's a standard CURI. It will fail if the CURI is located anywhere else.

2. Both in `relative` and `absolute` mode, it constructs a path to the Rakudo currently running. If I install dists into the same CURI with different Rakudos, the wrappers will point to different Rakudos as well. That's not ideal. Probably we should write the path to the Rakudo that initially set the mode to the configuration and always use that.

src/core.c/CompUnit/Repository/Installation.rakumod
patrickbkr: I didn't know! And I like that!
One usually only needs to specify the `#wrappers<relative>` part in the `-I 'mycur#wrappers<relative>#/path/to/repo'` thing when installing to the repo, because only then new wrappers are potentially generated. Right?

I still think we'll need a persisted config. I see two reasons:

- I want to guarantee that the wrappers are always uniform and matching the config in a CURI. The only way to do that is re-generating the wrappers when the config changes. So I need to know how they were configured previously to know if it changed.
- To me it seems a bit cumbersome to force users to specify the wrapper mode every time they install to a CURI. I imagine users will forget half of the time and end up trashing their wrappers every time they forget.

ugexe:
> It sounds like there is a desire to have one single configuration for the three standard repos. Can you explain what the motivation for that desire is? Performance? Guaranteeing that they never run out of sync?

It isn't a desire, so don't take my question as a suggestion on what you should implement immediately. However that comment was based on a concern for performance. The theoretical situation I was pondering was adding like 1000 external CURIs and if loading a module would then cause 1000 configurations to be read.  I suppose now that you changed it to be lazy though it shouldn't really be an issue since module loading is generally not thread-safe anyway, nor is it a very common occurrence.

ugexe:
> I just realized two shortcomings in the current implementation.
> 
> 1. In `relative` mode, the relative path is hard-coded assuming it's a standard CURI. It will fail if the CURI is located anywhere else.
> 2. Both in `relative` and `absolute` mode, it constructs a path to the Rakudo currently running. If I install dists into the same CURI with different Rakudos, the wrappers will point to different Rakudos as well. That's not ideal. Probably we should write the path to the Rakudo that initially set the mode to the configuration and always use that.

The "different Rakudo" issue applies to the repository version setting / logic as well. For example if I create a CURI with a rakudo that uses repository v1, then later install a module to that CURI with a newer rakudo that uses repository v2, the older rakudo will no longer be able to use that repository on disk since the newer rakudo will have changed the file layout to v2 and the old rakudo doesn't understand that.

src/core.c/CompUnit/Repository/Installation.rakumod
patrickbkr: I'm pondering the alternative: We give up on the guarantee that all wrappers in a CURI conform to the same wrapper mode. The wrapper mode still defaults to `absolute` or `relative` for standard CURIs and `path` for others. Specifying the mode via a command line switch changes the mode just for this one run / installation. It's then not possible anymore to change the wrappers after an installation, except for reinstalling the respective distros.
Using this mechanism is not obvious. The wish of "Install this one distro with these kinds of wrappers." requires specifying a switch on the *CURI* (not the distro).

I'm undecided which concept I like better. Opinions?

patrickbkr:
> I just realized two shortcomings in the current implementation.
> 
>     1. In `relative` mode, the relative path is hard-coded assuming it's a standard CURI. It will fail if the CURI is located anywhere else.
> 
>     2. Both in `relative` and `absolute` mode, it constructs a path to the Rakudo currently running. If I install dists into the same CURI with different Rakudos, the wrappers will point to different Rakudos as well. That's not ideal. Probably we should write the path to the Rakudo that initially set the mode to the configuration and always use that.

These are now fixed.

src/core.c/CompUnit/Repository/Installation.rakumod
patrickbkr: Done.

patrickbkr: Done.

src/core.c/CompUnit/Repository/Installation.rakumod
patrickbkr: After having slept on this, I still think that having all the wrappers in a CURI being uniform is best. The difference in the different types of wrappers is subtle, allowing those to be mixed can easily result in misbehavior that is really difficult to track down.

So my proposal (and what is implemented in this PR) is to keep the wrappers in a CURI uniform. How they should look is given in a config file with sane defaults (`path` everywhere except for the standard repos). I implemented no additional mechanism to change the configuration, given there will probably be little need. Should the need arise we can still add such a mechanism.

patrickbkr:
It does what I think it should, tests pass, ready for review!

tools/build/upgrade-repository.raku
```diff
@@ -1,7 +1,14 @@
 #!/usr/bin/env raku
+use nqp;
```
ugexe, line 2: Hmm, I wonder if this is the best place for this logic. Will users expect the wrappers to be regenerated when installing rakudo specifically when there is a repository upgrade that needs to occur?

patrickbkr: Once the config is there it won't touch it. So any subsequent runs are basically no-ops.
Currently it doesn't do a regeneration of the wrappers when upgrading. I'd say once we do add a migration for repo version 3 then we can decide if the upgrade makes a regeneration necessary.

Some context:

- To be able to correctly generate a config it's necessary to run on the installed repo, not the temporary one generated in `gen/build_rakudo_home`.
- It's necessary to know the final path of the rakudo the wrappers should point to. That's hacky to come by when trying to do it from within the `RepositoryRegistry`.
- Since the config is only written when it doesn't exist yet, it's good to have the initialization logic in a place that is not repeatedly called. Otherwise it's difficult to maintain, when the init is happening and if it has the right parameters at that point.

I'm open for ideas where else to put the logic, but I've went through several non-working approaches before doing it this way.

lib/CompUnit/Repository/Staging.rakumod
```diff
@@ -64,6 +64,20 @@ class CompUnit::Repository::Staging is CompUnit::Repository::Installation {
             $destination.parent.mkdir;
             $path.IO.copy: $destination;
         }
+
+        # Regenerate the bin wrappers, since those depend on the repository
+        # configuration. So the wrappers generated in this staging repo might
+        # not match the parents configuration.
+        for self.installed() -> $distribution {
```
ugexe, line 71: This does seem a bit unfortunate, and my first thought was it doesn't feel like this should go here. The reason is this `deploy` method isn't a blessed way of moving the repository where it needs to go, it was just a helper method that made some usages of the staging repository more concise. For example I used to just copy the directory myself without calling e.g. `deploy`, and I have to wonder if external packagers may still be doing something similar. Maybe there isn't a much better option though. The only thing that immediately comes to mind is that the configuration for the staging repo should arguably come from the repo it is staging for, and ideally the initially generated wrappers in the staging repo would be sufficient. I suspect I'm overlooking some details as to why that wouldn't work though.

src/core.c/CompUnit/Repository/Installation.rakumod
```diff
@@ -366,6 +382,62 @@ sub MAIN(*@, *%) {
...
+        $lock.lock;
+        LEAVE .unlock, .close with $lock;
+
+        my $version = self!repository-version;
+        self.upgrade-repository unless $version == 2;
+
+        for self.installed() -> $distribution {
+            my $dist  = CompUnit::Repository::Distribution.new($distribution);
+            my @files = $dist.meta<files>.grep(*.defined).map: -> $link {
+                $link ~~ Str ?? $link !! $link.keys[0]
```
ugexe, line 398: I think it would be nice if the logic for this loop was reused in the staging repository since this `$link ~~ Str ?? $link !! $link.keys[0]` logic is a code smell (that I wrote 😬) and it would be best to isolate it as much as possible

lib/CompUnit/Repository/Staging.rakumod
niner: Whatever gets generated during installation into a staging repo is the final version of those generated files. That's the whole point of staging: you get a set of files in a separate space that you can then just package/put into an archive with installation being just a dumb unpack without any post-processing. In other words, `deploy` must not do anything but copy some files.

patrickbkr: Code (tests) potentially runs from the staging repo, right? Can such code in principle try to run the wrappers? If yes, then we need different wrappers in the staging and target repo, because when the wrappers are set to mode relative, then the relative path from the wrapper to the rakudo executable is potentially different in the staging repo and the target repo.

If there is no need for the wrappers to be functional in the staging repo, then we can just generate them so they'll work only in the target repo.

ugexe: Tests are run *using* the staging repo yes, but the tests themselves still exist outside of that repository and thus are run from some other directory. Currently if a test wanted to run a bin/foo that is included in the distribution it'd have to do like `run $*PROGRAM.parent(2).add("bin/foo")` which wouldn't use the existing wrapper.

In other words: I do not believe there is a need for the wrappers to be functional in the staging repo and implementing it that way shouldn't cause any regressions.

src/core.c/CompUnit/Repository/Installation.rakumod
patrickbkr: Done. I've contained in a method in the `CompUnit::Repository::Distribution` class.

lib/CompUnit/Repository/Staging.rakumod
patrickbkr: I've now implemented it so the wrappers can and are again copied over verbatim from the staging repo to the parent.

src/core.c/CompUnit/Repository/Installation.rakumod
```diff
@@ -744,6 +816,239 @@ Please ask the author to add a \"provides\" section, mapping every exposed names
...
+                        return '"$@"';
+                    }
+                    when "abs" | "abs-slash" {
+                        return '"${DIR}/' ~ sh-escape($val) ~ '"';
+                    }
+                }
+            }
+            else {
+                note "Don't understand this arg: $path";
+                exit 1;
```
ugexe, line 868: We should probably be returning a failure or `die`ing instead of `exit`ing since this code can be reached by public methods on CURI

lib/CompUnit/Repository/Staging.rakumod
```diff
@@ -4,6 +4,7 @@ class CompUnit::Repository::Staging is CompUnit::Repository::Installation {
 
     submethod TWEAK(--> Nil) {
         $!parent = CompUnit::RepositoryRegistry.repository-for-name($!name);
+        self.copy-config-from($!parent);
```
niner, line 7: If I understand this correctly, the Staging repo directory will then contain a config file. This will create conflicts when multiple module packages then contain a config file for the same repository. The end result of an installation into a Staging repository should be a directory that can then be packaged as-is. Think: "I can tar this directory and un-tar it on a different machine and the module will then be installed there"

src/core.c/CompUnit/Repository/Installation.rakumod
```diff
@@ -8,40 +8,24 @@ class CompUnit::Repository::Installation does CompUnit::Repository::Locally does
...
-
-    my constant $raku-wrapper = '#!/usr/bin/env #raku#
-sub MAIN(*@, *%) {
+    my constant $raku-wrapper-code = 'sub MAIN(*@, *%) {
     CompUnit::RepositoryRegistry.run-script("#name#");
 }';
 
-
-    method TWEAK() {
+    method TWEAK(:$wrapper-mode, :$wrapper-rakudo-dir) {
```
niner, line 22: Would be nice to have some type constraints on these params.

src/core.c/CompUnit/Repository/Installation.rakumod
```diff
@@ -744,6 +816,239 @@ Please ask the author to add a \"provides\" section, mapping every exposed names
...
+        my @end-str-bytes = Rakudo::Internals.IS-WIN ?? (0x00, 0x00) !! (0x00,);
+
+        my $marker = "EXEC_RUNNER_WRAPPER_CONFIG_MARKER".encode($encoding);
+
+        my Buf $cbuf .= new;
+        my Int $pos = 0;
+
+        sub write-str(Str() $str) {
+            my $bytes = $str.encode($encoding);
+            $cbuf.write-uint16: $pos, ($bytes.elems / $char-size + 1).Int;
```
niner, line 938: I think this should throw if $bytes.elems exceeds 2**16

lib/CompUnit/Repository/Staging.rakumod
patrickbkr: Good point. I didn't realize it would try to copy the config over as well. That should not be.

On a related note, are we guaranteed, that the repo we copy the staged modules into will have the same configuration as the staging repos parent? Or phrased differently, can we accept the limitation that the final module destination repo will have to be identically configured to the staging repos parent? (In the typical case of a linux distro packaging up modules as well as the zef install case, it should work. But if you want to deploy a tared module to an unrelated rakudo installation on a different machine, it will likely break.)

src/core.c/CompUnit/Repository/Installation.rakumod
patrickbkr: Done.

lib/CompUnit/Repository/Staging.rakumod
patrickbkr: I've now made it so the staging repo has no config itself, but relays to the parent repos config. That's a pretty clean solution in my opinion.

src/core.c/CompUnit/Repository/Installation.rakumod
patrickbkr: Done.

src/core.c/CompUnit/Repository/Installation.rakumod
patrickbkr: Done.

zhuomingliang:
```
C:/Strawberry/c/bin/../lib/gcc/x86_64-w64-mingw32/13.2.0/../../../../x86_64-w64-mingw32/bin/ld.exe: unrecognized option '-z'
C:/Strawberry/c/bin/../lib/gcc/x86_64-w64-mingw32/13.2.0/../../../../x86_64-w64-mingw32/bin/ld.exe: use the --help option for usage information
collect2.exe: error: ld returned 1 exit status
gmake: *** [makefile:158: win-runner.exe-tmpl] Error 1
```

patrickbkr:
> ```
> C:/Strawberry/c/bin/../lib/gcc/x86_64-w64-mingw32/13.2.0/../../../../x86_64-w64-mingw32/bin/ld.exe: unrecognized option '-z'
> C:/Strawberry/c/bin/../lib/gcc/x86_64-w64-mingw32/13.2.0/../../../../x86_64-w64-mingw32/bin/ld.exe: use the --help option for usage information
> collect2.exe: error: ld returned 1 exit status
> gmake: *** [makefile:158: win-runner.exe-tmpl] Error 1
> ```

Oof. Obviously the commands that work on Linux don't on Windows. I have now set up a GCC toolchain myself and made it work (on my machine).

@zhuomingliang If you want you could give it (yet another) try.

patrickbkr:
> +++ Compiling inst-rakuw.exe lto1.exe: fatal error: could not open symbol resolution file: No such file or directory compilation terminated. lto-wrapper.exe: fatal error: gcc returned 1 exit status compilation terminated. C:/Strawberry/c/bin/../lib/gcc/x86_64-w64-mingw32/13.2.0/../../../../x86_64-w64-mingw32/bin/ld.exe: error: lto-wrapper failed collect2.exe: error: ld returned 1 exit status gmake: *** [makefile:158: win-runner.exe-tmpl] Error 1

Ow this is annoying. Somehow gcc lto fails on your setup. I've disabled LTO in this PR now.
Try again?

lizmat:
@patrickbkr I am no position to review this PR.  But am willing to merge it, if you say it's ready  :-)

Issue #5729, opened 2024-12-13
[JVM] Internal errors when running S32-str/format.t


usev6:
## The Problem
Running S32-str/format.t on the JVM backend leads to a failure due to internal errors:
```
S32-str/format.t...................................java.lang.NullPointerException
  in block <unit> at t/spec/S32-str/format.t line 28

# You planned 49 tests, but ran 10
  10   39    0    0   49
    39 tests aborted (missing ok/not ok)
```

This specific NullPointerException happens in https://github.com/rakudo/rakudo/blob/4082d2c707/src/Raku/ast/signature.rakumod#L1255 because the call to `nqp::getattr` gives back null when evaluating `my $f1 = Format.new("%5s")`.

But AFAICS there are other problems further down. I tried to find a patch that makes the tests pass, but failed so far. So I decided to create an issue to at least document the problem. I plan to add more information later.

## Environment
* Operating system: FreeBSD 14.1-RELEASE-p5
* Compiler version (`rakudo -v` or `raku -v`): Welcome to Rakudo™ v2024.10-90-g4082d2c707.
Implementing the Raku® Programming Language v6.d.
Built on JVM.

usev6:
With the first NullPointerException out of the way, the next problem happens in https://github.com/Raku/roast/blob/c2d2875756/S32-str/format.t#L39
```
$ ./rakudo-j --ll-exception -e 'use v6.e.PREVIEW; dd Format.new("%5s:%5s")'
java.lang.NullPointerException
  in IMPL-CANONICALIZE-PAIR (gen/jvm/BOOTSTRAP/v6c.nqp:7415)
  in resolve-infix (gen/jvm/BOOTSTRAP/v6c.nqp:7011)
  in PERFORM-PARSE (gen/jvm/BOOTSTRAP/v6c.nqp:15683)
  in ensure-parse-performed (gen/jvm/BOOTSTRAP/v6c.nqp:10679)
  in IMPL-BEGIN (gen/jvm/BOOTSTRAP/v6c.nqp:8869)
  in <anon> (gen/jvm/BOOTSTRAP/v6c.nqp:8862)
  in visit-children (gen/jvm/BOOTSTRAP/v6c.nqp:17933)
  in IMPL-BEGIN (gen/jvm/BOOTSTRAP/v6c.nqp:8862)
  in <anon> (gen/jvm/BOOTSTRAP/v6c.nqp:8862)
[...]

PR #5730, opened 2024-12-13
Only call ensure-sc if there is a descriptor

usev6:
On the JVM backend this avoids a NullPointerException for the following code:

  Format.new("%5s")

The call to nqp::getattr returns a null object in this case (for both backends). On MoarVM this doesn't lead to an error. But it still feels appropriate to avoid the call if there is no container descriptor.

This would be the first step for fixing
https://github.com/rakudo/rakudo/issues/5729.

### DIFF
--- a/src/Raku/ast/signature.rakumod
+++ b/src/Raku/ast/signature.rakumod
@@ -1253,7 +1253,7 @@ class RakuAST::Parameter
         }
 
         my $container_descriptor := $param-obj.container_descriptor;
-        $context.ensure-sc($container_descriptor) if $container_descriptor;
+        $context.ensure-sc($container_descriptor) unless nqp::isnull($container_descriptor);
 
         # Bind parameter into its target.
         if self.invocant {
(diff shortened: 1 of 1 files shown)


src/Raku/ast/signature.rakumod
```diff
@@ -1252,7 +1252,8 @@ class RakuAST::Parameter
                           QAST::Op.new(:op<iseq_s>, $wval, $temp-qast-var));
         }
 
-        $context.ensure-sc(nqp::getattr($param-obj, Parameter, '$!container_descriptor'));
+        my $container_descriptor := $param-obj.container_descriptor;
+        $context.ensure-sc($container_descriptor) if $container_descriptor;
```
lizmat, line 1256: If the object is nqp::null, then a better fix would be `unless nqp::isnull($container_descriptor)`

usev6: Yeah, the problem was a nqp::null. My thought was that testing for truthiness would be good enough and it would match with the other test in line 1299. But your suggestion is more precise, I've pushed a second commit.

PR #5736, opened 2024-12-16
Make mult-arg grep slip its arguments

lizmat:
In response to the discussion at:
  https://www.reddit.com/r/rakulang/comments/1havdtd/why_does_grepping_a_list_in_chunks_group_the/

This breaks 3 tests in roast:
- t/spec/S03-operators/composition.t   Failed tests:  11-12
- t/spec/S32-list/grep.rakudo.moar     Failed test:  50

### DIFF
--- a/src/core.c/Any-iterable-methods.rakumod
+++ b/src/core.c/Any-iterable-methods.rakumod
@@ -334,7 +334,7 @@ Consider using a block if any of these are necessary for your mapping code."
           !! $count == 2
             ?? Rakudo::IterateTwoWithoutPhasers.new(
                  -> \a, \b {
-                     my \params := (a, b);
+                     my \params := Slip.new(a, b);
                      nqp::handle(
                        (my $result := $test(a, b)),
                        'NEXT', nqp::if(
@@ -361,7 +361,7 @@ Consider using a block if any of these are necessary for your mapping code."
                )
             !! Rakudo::IterateMoreWithPhasers.new(
                  -> |c {
-                     my \params := c.list;
+                     my \params := c.list.Slip;
                      nqp::handle(
                        (my $result := $test(|c)),
                        'NEXT', nqp::if(


patrickbkr:
This is an interesting one. I think it has WAT potential in either direction. Given this breaks a few tests, let's let this simmer for a bit longer before taking a decision.

ab5tract:
I think my preferred fix for this would be to return kv to:

{ .keys Z .values }

Then map and grep can stop taking varargs and the output of .kv, .kv.grep, and .kv.map can all have the same shape.

jubilatious1:
I'm really not in favor of flattening by default, because you lose valuable grouping information that is difficult if not impossible to reconstruct. That being said, flattening is usually one of my biggest headaches. 

Reading the `docs` for `grep`, I see invocants that are `List`, `Any`, `RaceSeq`, `HyperSeq`, and `Supply`. All single arguments/invocants, I believe? So it's not quite clear to me why ___multi-arg___ invocants are accepted. 

That being said it seems like you're one `map` away from grepping on both elements of a multi-arg input. If someone forgets to `map`, suggest that they want to grep on `pairs` instead of `kv`? It seems like the right thing to do anyway.

lizmat:
> { .keys Z .values }

That's .pairs, aka :p ?

jubilatious1:
It's a better example to use a real hash, not a sequence:

```raku
~$ raku -e 'my %hash = %("one", 1, "two", 2, "three", 3, "four", 4, "five", 5); say %hash.kv.grep(-> $i, $n { $n %% 2});'
((two 2) (four 4))
~$ raku -e 'my %hash = %("one", 1, "two", 2, "three", 3, "four", 4, "five", 5); say %hash.kv.grep(-> $i, $n { $n %% 2})[0];'
(two 2)
```
Above I like the fact that I get structured data back, because then I can pull out matches by index (e.g. `.[0]`). 

Below, I can get flattened data with `.flat`:
```raku
raku -e 'my %hash = %("one", 1, "two", 2, "three", 3, "four", 4, "five", 5); say %hash.kv.grep(-> $i, $n { $n %% 2}).flat;'
(four 4 two 2)
```

This pull request basically auto-flattens, which will lose the `grep`ed hash-pair info from the outset. Thus it seems this pull request in unnecessary, because calling `.flat` on the OP's example gives the OP's desired output.

ab5tract:
> > { .keys Z .values }
> 
> That's .pairs, aka :p ?

As discussed in `#raku-dev`, the semantics of `{ .keys Z .values }` are different:

```
sub kv(@in) { @in.keys Z @in.values }
dd <x y z>.&kv
# ((0, "x"), (1, "y"), (2, "z")).Seq
```

Unfortunately I feel compelled to retract my preference to return to these semantics, at least until we would have a means to optimize away the substantial performance impact of building the lists and then de-structuring them in the block signature of `map`:

```
sub kv(@in) {
    @in.keys Z @in.values
}

sub zip-kv {
    my @storage;
    my $a := gather for ^Inf {
        state $char //= 'a';
        take $char;
        $char .= succ;
    }

    my $start = now;
    $a[^256_000].&kv.map: -> ($k,$v) {
        say "$k, $v" if $++ %% 10_000;
    }
    say "... zip approach took {now - $start} seconds";
}

sub regular-kv {
    my @storage;
    my $a := gather for ^Inf {
        state $char //= 'a';
        take $char;
        $char .= succ;
    }

    my $start = now;
    $a[^256_000].kv.map: -> $k,$v {
        say "$k, $v" if $++ %% 10_000;
    }
    say "... regular approach took {now - $start} seconds";
}

say "Benchmarking zip-kv...";
kv-zip;

say "Benchmarking regular-kv...";
regular-kv;
```

Output:

```
Benchmarking zip-kv...
... zip approach took 1.856616818 seconds
Benchmarking regular-kv...
... regular approach took 0.833412978 seconds
```

That said, I don't know if flattening `grep` automatically makes sense, especially as anything that isn't "oddly shaped versus `grep/map`" like `.kv` is would lead to unexpected outcomes in the opposite direction.

```
[2] > [|^12].rotor(2).grep( -> ($a,$b) { ($a,$b) if $a %% 3 } )
((0 1) (6 7))
[3] > [|^12].rotor(2).grep( -> ($a,$b) { ($a,$b) if $a %% 3 } ).flat
(0 1 6 7)
```

PR #5739, opened 2024-12-17
Address proc issues caused by repl-mode

ab5tract:
Once [nqp#834](https://github.com/Raku/nqp/pull/834) is merged, this will be updated and draft-status removed.


ab5tract:
I don't think there is any need for this at this point.

PR #5756, opened 2025-01-09
Split core modules into separate distributions

ugexe:


ugexe:
I've pushed this up in case someone else feels like they want to take it further. These commits are a bit old, so there will likely be more modules to moved around, and potentially more tests to update their use of `use lib "lib"`. The commit messages of the two included commits should make things pretty self explanatory.

ab5tract:
@ugexe Would allow for doing things like creating a `Test:api<2>`? If so, I see a lot of utility here.

PR #5757, opened 2025-01-11
Use .raku suffix for install-dist instead of .p6

1-1sam:
The documentation for the install-dist.raku script still refers to itself as install-dist.p6, which is no longer accurate. This commit corrects that and replaces mentions of install-dist.p6 with install-dist.raku.

### DIFF
--- a/tools/install-dist.raku
+++ b/tools/install-dist.raku
@@ -101,9 +101,9 @@ sub meh(Str:D $message --> Nil) {
 
 =begin pod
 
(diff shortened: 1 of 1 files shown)


lizmat:
FWIW, I tried this before, but that caused some ecosystem fallout  :-(

lizmat:
Thanks, I misread initially  :-)

ugexe:
This exact change was reverted in the past. Are either of you sure making this change is correct? Maybe it’s worth looking at the for history of the file

PR #5763, opened 2025-01-18
Pod::To::Text: Sort enum pairs by value

1-1sam:
Currently, Pod::To::Text renders enum pod declarators by passing the pair list returned by .pairs to signature2text. The issue with this is that the pairs will not be in any fixed order, meaning rendered output of enum pod declarators could be inconsistent between different renders, which I find to be undesirable behavior. This PR fixes that by sorting the pair list by value. I chose to sort by value because, in my personal experience, writing enums ordered by enum value seems to be fairly common in code.


lizmat (approved):

lizmat:
Thanks!

PR #5779, opened 2025-02-02
Introduce Hash::Ordered: hashes that keep the order of the keys

lizmat:
This subclass of Hash will keep the order in which keys have been added, with as little overhead as possible.

The theory of operation is that whenever a key is accessed, and the key does **not** exist in the hash, then the key will be added to the list of known keys.  If a key is deleted, that key is **not** removed from the list of known keys (but it *is* deleted from the hash).

Only when some form of iteration is needed over the keys (or values) of the hash, is the list of known keys (which is in order of appearance) checked for existence in the hash, and any double entries or deleted keys are removed, and the list of keys updated.

The current syntax is:
```raku
  my %h is Hash::Ordered = ...;
```
This could probably be improved by either a dedicated trait, such as:
```raku
  my %h is ordered = ...;
```
or maybe some other syntax, such as:
```raku
  my %h{:ordered} = ...;
```
Support for typed ordered hashes is basically automatically available.

Support for object hashes would need to be made separately.  This shouldn't be too hard, but felt only necessary if this change to the Raku Programming Language is approved.

Also, maybe this should become a 6.e feature: but then integration into the legacy, as well as the Raku grammar, would be more troublesome.

### DIFF
--- a/src/core.c/Hash/Ordered.rakumod
+++ b/src/core.c/Hash/Ordered.rakumod
@@ -1,0 +1,102 @@
+my class Hash::Ordered is Hash {
+    has str @.keys;
+
+    method STORE(Hash::Ordered:D: |c) {
+        my str @*KEYS;
+        self.Hash::STORE(|c);
+        @!keys := @*KEYS;
+        self
+    }
+
+    method STORE_AT_KEY(Hash::Ordered:D: Str() $key, Mu \value --> Nil) {
+        nqp::push_s(@*KEYS,$key);
+        self.Hash::STORE_AT_KEY($key, value);
+    }
+
+    method AT-KEY(Hash::Ordered:D: Str() $key) is raw {
+        nqp::ifnull(
+          nqp::atkey(nqp::getattr(self,Map,'$!storage'),$key),
+          nqp::stmts(
+            nqp::push_s(@!keys,$key),
+            nextsame;
+          )
+        )
+    }
+
+    method ASSIGN-KEY(Hash::Ordered:D: Str() $key, Mu \value) is raw {
+        nqp::ifnull(
+          nqp::atkey(nqp::getattr(self,Map,'$!storage'),$key),
+          nqp::stmts(
+            nqp::push_s(@!keys,$key),
+            self.Hash::AT-KEY($key)
+          )
+        ) = value
+    }
+
+    method BIND-KEY(Hash::Ordered:D: Str() $key, Mu \value) is raw {
+        nqp::push_s(@!keys,$key)
+          unless nqp::existskey(nqp::getattr(self,Map,'$!storage'),$key);
+
+        self.Hash::BIND-KEY($key, value)
+    }
+
+    method keys(Hash::Ordered:D:) {
+        my $storage := nqp::getattr(self,Map,'$!storage');
+        my $old     := @!keys;
+
+        # all ok
+        if nqp::elems(@!keys) == nqp::elems($storage) {
+            $old
+        }
+
+        # Something changed, revisit keys
+        else {
+            my str @new;
+            my $seen    := nqp::hash;
+
+            my int $elems = nqp::elems($old);
+            my int $i     = -1;
+
+            nqp::while(
+              nqp::islt_i(++$i,$elems),
+              nqp::if(
+                nqp::existskey($storage,my str $key = nqp::atpos_s($old,$i))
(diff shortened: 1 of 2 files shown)


jubilatious1:
I recognize this is a LOT of work, but it seems to live in Module space already:

https://raku.land/zef:FCO/OrderedHash

At the risk of offending anyone reading, sorting/ordering is a huge subject in Data Science, and I'm not sure that this branch/merged-branch has all the capabilities a DS person might expect from other languages. 

For example, in the R-programming language you can `order` as follows:

https://web.mit.edu/~r/current/lib/R/library/BiocGenerics/html/order.html

> <head></head>
> order {base} | R Documentation

> Ordering Permutation
> 
> Description
> 
> order returns a permutation which rearranges its first argument into ascending or descending order, breaking ties by further arguments. sort.list is the same, using only one argument.
> See the examples for how to use these functions to sort data frames, etc.
> 
> Usage
> 
> `order(..., na.last = TRUE, decreasing = FALSE,
>       method = c("auto", "shell", "radix"))`
> 
> `sort.list(x, partial = NULL, na.last = TRUE, decreasing = FALSE,
>           method = c("auto", "shell", "quick", "radix"))`
> Arguments
> 
> ...	
> a sequence of numeric, complex, character or logical vectors, all of the same length, or a classed R object.
> 
> x	
> an atomic vector.
> 
> partial	
> vector of indices for partial sorting. (Non-NULL values are not implemented.)
> 
> decreasing	
> logical. Should the sort order be increasing or decreasing? For the "radix" method, this can be a vector of length equal to the number of arguments in .... For the other methods, it must be length one.
> 
> na.last	
> for controlling the treatment of NAs. If TRUE, missing values in the data are put last; if FALSE, they are put first; if NA, they are removed (see ‘Note’.)
> 
> method	
> the method to be used: partial matches are allowed. The default ("auto") implies "radix" for short numeric vectors, integer vectors, logical vectors and factors. Otherwise, it implies "shell". For details of methods "shell", "quick", and "radix", see the help for [sort](https://web.mit.edu/r/current/lib/R/library/base/html/sort.html).
>

jubilatious1:
Also, in the R-programming language you can `sort` as follows:

https://web.mit.edu/r/current/lib/R/library/base/html/sort.html

> <head></head>
> sort {base} | R Documentation
> -- | --
> 
> 
> <br class="Apple-interchange-newline" style="caret-color: rgb(0, 0, 0); color: rgb(0, 0, 0); font-style: normal; font-variant-caps: normal; font-weight: 400; letter-spacing: normal; orphans: auto; text-align: start; text-indent: 0px; text-transform: none; white-space: normal; widows: auto; word-spacing: 0px; -webkit-text-stroke-width: 0px; text-decoration: none;">sort {base}	R Documentation
> Sorting or Ordering Vectors
> 
> Description
> 
> Sort (or order) a vector or factor (partially) into ascending or descending order. For ordering along more than one variable, e.g., for sorting data frames, see [order](https://web.mit.edu/r/current/lib/R/library/base/html/order.html).
> 
> Usage
> 
> `sort(x, decreasing = FALSE, ...)`
> 
> ## Default S3 method:
> `sort(x, decreasing = FALSE, na.last = NA, ...)`
> 
> `sort.int(x, partial = NULL, na.last = NA, decreasing = FALSE,
>          method = c("auto", "shell", "quick", "radix"), index.return = FALSE)`
> Arguments
> 
> `x`	
> for sort an R object with a class or a numeric, complex, character or logical vector. For sort.int, a numeric, complex, character or logical vector, or a factor.
> 
> `decreasing`	
> logical. Should the sort be increasing or decreasing? For the "radix" method, this can be a vector of length equal to the number of arguments in .... For the other methods, it must be length one. Not available for partial sorting.
> 
> `...`	
> arguments to be passed to or from methods or (for the default methods and objects without a class) to sort.int.
> 
> `na.last`	
> for controlling the treatment of NAs. If TRUE, missing values in the data are put last; if FALSE, they are put first; if NA, they are removed.
> 
> `partial`	
> NULL or a vector of indices for partial sorting.
> 
> `method	`
> character string specifying the algorithm used. Not available for partial sorting. Can be abbreviated.
> 
> `index.return`	
> logical indicating if the ordering index vector should be returned as well. Supported by method == "radix" for any na.last mode and data type, and the other methods when na.last = NA (the default) and fully sorting non-factors.

PR #5785, opened 2025-02-17
Use a local for a somewhat hot class member access

MasterDuke17:
No need to repeat the `.possible` call for every iteration.


lizmat:
Good catch!

Issue #5788, opened 2025-02-21
Fix: Instant::from-posix-nanos


0rir:
At src/core.c/Instant.rakumod:14 this method could be corrected so:
```
    method from-posix-nanos(Instant:U: Int:D $nanos --> Instant:D) {
        nqp::p6bindattrinvres(nqp::create(Instant),Instant,'$!tai',
          (Rakudo::Internals.tai-from-posix($nanos/10**9,0) * 10**9).Int)
    }
```

I can push this to my existing PR #5786 which is a test for this.  Issue #5595 raised this; I closed it as a mess.

PR #5800, opened 2025-03-05
Test and fix that Instant.from-posix-nanos and Instant.from-posix  match.

0rir:
Test and fix that `Instant.from-posix-nanos` and `Instant.from-posix`  match.  #5788.


lizmat:
@japhb if you need `Instant` that quickly, why not use `nano`?

japhb:
> @japhb if you need `Instant` that quickly, why not use `nano`?

Interesting point.  `nano` doesn't actually get you an Instant (rather just the raw integral nanosecond count) -- and it didn't exist when I wrote a lot of my timing code (and is still "hidden" behind `use v6.e.PREVIEW`) -- but for this use case, it might be a fair replacement.

May be best that I just go and update all my existing timing code to use `nano` instead, since that at least makes *intent* clear (`now` was more of a "close enough").

PR #5804, opened 2025-03-07
Don't try to lower signatures with parametric generic types

timo:
Fixes "inconsistent bind result" for signatures with a type capture followed by @ or % sigiled params using the typevar, or when such a param uses a type capture param of a role.

See discussion starting here: https://irclogs.raku.org/raku-dev/2025-03-07.html#13:49-0001

If someone were so inclined to implement this case by generating code in the lowered signature that creates the concretization on the fly, go ahead. Otherwise, comments from further below in the lowerer seem to imply RakuAST will also make the binding lowering code obsolete anyway.


lizmat:
Thanks!

PR #5817, opened 2025-03-15
improve type check, there still a lot of improvement

zhuomingliang:
improve type check, there still a lot of improvement, such as "my Str @a = 1,2;"

these causes 4 failing test, these test seems it's wrong under rakuast, these are compling error now, not runtime error.
t/spec/6.c/S04-declarations/my-6c.rakudo.moar
t/spec/S04-declarations/my-6e.rakudo.moar
t/spec/S04-declarations/will-complain.t
t/spec/S06-signature/introspection.rakudo.moar

`my-6c` and `my-6e` is a easy fix, I don't know how to rewrite will-complain.t.
 When I run it with `RAKUDO_RAKUAST=1 ./rakudo-m -Ilib t/spec/S06-signature/introspection.rakudo.moar`, seems it's failed at main branch too.


zhuomingliang:
Updated `t/spec/6.c/S04-declarations/my-6c.t` and `t/spec/S04-declarations/my-6e.t` by https://github.com/Raku/roast/commit/dca7e75327

zhuomingliang:
Updated `t/spec/S04-declarations/will-complain.t` by https://github.com/Raku/roast/commit/033efe81bd5a9ed930b3b7a57be4cae148f1084c

src/Raku/ast/expressions.rakumod
```diff
@@ -1982,21 +1982,34 @@ class RakuAST::ApplyInfix
...
+            if (nqp::istype($infix, RakuAST::Assignment) || $infix.operator eq ':=')
+                && !nqp::istype($type.HOW, Perl6::Metamodel::SubsetHOW) {
 
-                    my $value := $right.compile-time-value;
-                    if !nqp::istype($value, $type)
-                      && nqp::istype($type, $resolver.type-from-setting('Numeric')) {
-                        self.add-sorry:
-                          $resolver.build-exception: 'X::Syntax::Number::LiteralType',
-                            :vartype($type), :$value;
+                my $sigil := (try $left.sigil) // '';
```
niner, line 1990: In general I avoid using try in the RakuAST code. Reason is that it doesn't tell the reader in which situations it's expected to fail or succeed. Also it may catch exceptions that you didn't think of. Granted, that's unlikely with such a basic accessor as .sigil. But what if someone adds a method sigil to some subclass that does something completely different and fails?

niner: Or what if a node has a sigil method for some reason (maybe a convenience wrapper) but you did not expect this type of node to be included here?

zhuomingliang: > In general I avoid using try in the RakuAST code. Reason is that it doesn't tell the reader in which situations it's expected to fail or succeed. Also it may catch exceptions that you didn't think of. Granted, that's unlikely with such a basic accessor as .sigil. But what if someone adds a method sigil to some subclass that does something completely different and fails?

agree, I saw three place using try in the RakuAST code, and I copied it from here: https://github.com/Rakudo/Rakudo/commit/a0ed2ae93a3a8d0fa3de5fb079c464bec5f530be#diff-3c8aafcca6a26bfab047f06f101aa7d47de2d605e66ab2c280108063bda8f291L794

zhuomingliang: I think it would be better to have `RAKUAST::TypeCheck` and `RAKUAST::TypeCheck::Parameter` class to reuse these codes?

niner (approved):

PR #5821, opened 2025-03-18
Add variable name if possible to warning about use of uninitialized numeric

MasterDuke17:
So `my $f; ++$f while $f < 3; say $f` now says:
```
Use of uninitialized value $f of type Any in numeric context
  in block <unit> at -e line 1
3
```

### DIFF
--- a/src/core.c/Mu.rakumod
+++ b/src/core.c/Mu.rakumod
@@ -789,30 +789,30 @@ my class Mu { # declared in BOOTSTRAP
     proto method Numeric(|) {*}
     multi method Numeric(Mu:U \v:) {
         my $name = (defined($*VAR_NAME) ?? $*VAR_NAME !! try v.VAR.name) // '';
-        $name   ~= ' ' if $name ne '';
-        warn "Use of uninitialized value {$name}of type {self.^name} in numeric context";
+        $name    = " $name" if $name;
+        warn "Use of uninitialized value$name of type {self.^name} in numeric context";
         0
     }
     proto method Real(|) {*}
     multi method Real(Mu:U \v:) {
         my $name = (defined($*VAR_NAME) ?? $*VAR_NAME !! try v.VAR.name) // '';
-        $name   ~= ' ' if $name ne '';
-        warn "Use of uninitialized value {$name}of type {self.^name} in numeric context";
+        $name    = " $name" if $name;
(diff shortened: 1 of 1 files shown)


src/core.c/Mu.rakumod
```diff
@@ -788,17 +788,23 @@ my class Mu { # declared in BOOTSTRAP
 
     proto method Numeric(|) {*}
     multi method Numeric(Mu:U \v:) {
-        warn "Use of uninitialized value of type {self.^name} in numeric context";
+        my $name = (defined($*VAR_NAME) ?? $*VAR_NAME !! try v.VAR.name) // '';
+        $name   ~= ' ' if $name ne '';
```
lizmat, line 792: `if $name` is sufficient

src/core.c/Mu.rakumod
```diff
@@ -788,17 +788,23 @@ my class Mu { # declared in BOOTSTRAP
...
+        warn "Use of uninitialized value {$name}of type {self.^name} in numeric context";
```
lizmat, line 793: if you put the space *before*, you could do "value$name of" and lose the additional scope

lizmat:
Otherwise, nice catch!

MasterDuke17:
I also wanted to add it to https://github.com/rakudo/rakudo/blob/main/src/core.c/Exception.rakumod#L3240-L3243, but the roast examples that throw that end up traveling through several layers of calls and I don't think it'd be easy (or maybe even possible) to get the right name.

lizmat:
You could try adding "is raw" to `has Numeric $.type` and see what that brings you?

MasterDuke17:
> You could try adding "is raw" to `has Numeric $.type` and see what that brings you?

Unfortunately, it brings `Can't use unknown trait 'is' -> 'raw' in an attribute declaration.`.

lizmat:
Ah, too early in the bootstrap, probably

MasterDuke17:
> Ah, too early in the bootstrap, probably

Did you mean `is rw`? Because `is raw` doesn't work for any attribute declaration. If I do `raku -e 'class A { has $.b is raw; }'` it dies with `===SORRY!=== Error while compiling -e
Can't use unknown trait 'is' -> 'raw' in an attribute declaration.`

PR #5833, opened 2025-03-25
Eliminate magic 37 in Instant class. Modify some comments

0rir:


src/core.c/Rakudo/Internals.rakumod
```diff
@@ -1026,6 +1026,10 @@ my class Rakudo::Internals {
         #END leap-second-posix
     );
 
+    method current-offset-nanos() {
```
lizmat, line 1029: Feels to me this calculation should only be done once at installation?  So prefix with `BEGIN`

src/core.c/Instant.rakumod
```diff
@@ -4,8 +4,8 @@ my class Duration {... }
 
 my class Instant is Cool does Real {
     has Int $.tai is default(0);
-      # A linear count of nanoseconds since 1970-01-01T00:00:00Z, plus
-      # Rakudo::Internals.initial-offset. Thus, $.tai matches TAI from 1970
+      # A count of nanoseconds since 1970-01-01T00:00:00Z with leap seconds
+      # and Rakudo::Internals.initial-offset. Thus, $.tai matches UTC from 1970
```
lizmat, line 8: Why the s/TAI/UTC ??

src/core.c/Instant.rakumod
```diff
@@ -127,14 +127,14 @@ multi sub infix:<->(Instant:D $a, Real:D $b --> Instant:D) {
...
 sub term:<now>(--> Instant:D) {
-    # FIXME: During a leap second, the returned value is one
-    # second greater than it should be.
-    my int constant \tai-offset-nanos = 37 * 1000000000;
-    Instant.from-posix-nanos(nqp::add_i(nqp::time,tai-offset-nanos))
+    # A posix time in the second preceeding a positive leap second,
+    # represents two seconds in UTC time.  This ambiguity is currently
+    # resolved by returning the leap second.
+
+    my int $offset-nanos = Rakudo::Internals.current-offset-nanos;
```
lizmat, line 135: Appears leftover from a previous version?

src/core.c/Instant.rakumod
0rir: Some sites equate TAI as equivalent to UTC, but  TAI  is not corrected from 86400 seconds per day.

src/core.c/Instant.rakumod
0rir: Yes.  I push that out.  Part of my trying to understand the NQP scoping rules.

src/core.c/Instant.rakumod
lizmat: Which is why TAI is mentioned there, instead of UTC?

src/core.c/Rakudo/Internals.rakumod
lizmat: From my vague recollection, is that TimToady wanted astronomers to be able to use what is now Raku

PR #5837, opened 2025-03-30
RakuAST: Throw right exceptions for declaring numeric and match variables

MasterDuke17:
Gets t/spec/S15-literals/identifiers.t passing and two individual tests in S32-exceptions/misc2.t.

### DIFF
--- a/src/Raku/Actions.nqp
+++ b/src/Raku/Actions.nqp
@@ -2494,7 +2494,7 @@ class Raku::Actions is HLL::Actions does Raku::CommonActions {
 
         my str $scope := $*SCOPE;
         my     $type  := $*OFTYPE.ast if $*OFTYPE;
-        my str $sigil := $<sigil>;
+        my str $sigil := $<variable><sigil>;
         my     $where := $*WHERE;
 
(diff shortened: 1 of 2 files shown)


src/Raku/Actions.nqp
```diff
@@ -2522,14 +2522,14 @@ class Raku::Actions is HLL::Actions does Raku::CommonActions {
 
         my $decl;
         my $shape := $<semilist> ?? $<semilist>[0].ast !! Nodify('SemiList');
-        if $<desigilname> {
-            my $desigilname := $<desigilname>;
+        if $<desigilname> || $<variable><desigilname> {
```
niner, line 2525: Isn't it always `$<variable><desigilname>` now? `variable-declarator` doesn't have a desigilname submatch anymore.

MasterDuke17: > Isn't it always `$<variable><desigilname>` now? `variable-declarator` doesn't have a desigilname submatch anymore.

Yep, thanks for catching that.

niner (approved):

PR #5838, opened 2025-04-01
Change := and =:= to ==

zhuomingliang:


niner (approved):

src/Raku/ast/signature.rakumod
```diff
@@ -94,7 +94,7 @@ class RakuAST::Signature
                         self.add-sorry: $resolver.build-exception: 'X::Parameter::WrongOrder', misplaced => 'required', after => 'named', parameter => $_.target.name;
                     }
 
-                } elsif $kind := 2 {           # optional
+                } elsif $kind =:= 2 {          # optional
```
lizmat, line 97: OOC, what's wrong with `==`  ??

niner: That's also something I wonder. I think == would be quite enough here. It is about numeric comparison after all.

zhuomingliang:
Thanks for review, I have force-pushed it, changed it to '=='.

PR #5839, opened 2025-04-03
RakuAST: add some more suggestions to exception messages

MasterDuke17:
Gets 4 more tests passing in t/spec/S32-exceptions/misc2.t


niner:
Thanks! I removed the commented code and fixed a small bug and added type suggestions for Type::Simple (though I failed to come up with a situation where this would actually help, but at least it brought some infrastructure). Pushed the fixed version to main.

PR #5846, opened 2025-04-09
Skip experimental macro implementation tests

ab5tract:


lizmat:
Appears that these tests are already skipped when running under RakuAST.

PR #5856, opened 2025-04-17
RakuAST: Interned int, str and num

zhuomingliang:


lizmat:
Feels like the interning like this, is a possible race condition.  @niner: or are we sure we're always single threaded here, in the same CompUnit?

PR #5867, opened 2025-04-28
IO::Handle.slurp: Read entire file contents at once

ab5tract:
We can easily pre-size the read volume in the call to nqp::readfh for the cases where it makes sense (ie., every path that isn't "-".IO).

We still preserve the older behavior in case the byte length is super low or if there is some other interruption to getting the self.s value at runtime.

@timo++ noticed this opportunity.

### DIFF
--- a/src/core.c/IO/Path.rakumod
+++ b/src/core.c/IO/Path.rakumod
@@ -615,12 +615,13 @@ my class IO::Path is Cool does IO {
     }
 
     # slurp contents of low level handle
-    sub slurp-PIO(Mu \PIO) is raw {
-        constant slurp-size = 0x100000;
+    sub slurp-PIO(Mu \PIO, :$size = 0x100000) is raw {
+        my \slurp-size = $size;
         nqp::readfh(PIO,(my $blob := nqp::create(buf8.^pun)),slurp-size);
 
+        # if the size is set to the default value, then
         # enough to read entire buffer, assume there's more
-        if nqp::iseq_i(nqp::elems($blob),slurp-size) {
+        if slurp-size == 0x100000 && nqp::iseq_i(nqp::elems($blob),slurp-size) {
             nqp::while(
               nqp::elems(
                 nqp::readfh(PIO,(my $part := nqp::create(buf8.^pun)),slurp-size)
@@ -635,9 +636,9 @@ my class IO::Path is Cool does IO {
     sub slurp-stdin-bin() is raw { slurp-PIO(nqp::getstdin) }
 
     # slurp given path in binary mode
-    sub slurp-path-bin(str $path) is raw {
+    sub slurp-path-bin(str $path, int $size) is raw {
         my $PIO  := nqp::open($path,'r');
-        my $blob := slurp-PIO($PIO);
+        my $blob := slurp-PIO($PIO, :$size);
         nqp::closefh($PIO);
         $blob
     }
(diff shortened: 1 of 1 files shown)


niner:
This does another file system access though for the stat() call. Is that really faster than resizing the buffer as needed?

ab5tract:
> This does another file system access though for the stat() call. Is that really faster than resizing the buffer as needed?

For large files I've seen a 2-3x reduction in ingestion time.

But indeed, we would want a final version of such a feature to have some rubrics around only doing this when it's "worth it".

Another option would be to put the onus on the user entirely, for instance with a new multi candidate for `IO::Path.slurp` (eg, `$path.IO.slurp(:presize)`).

ab5tract:
With `AllPrintings.json`, a `512160792` byte file.

```
# r is the alias for the development raku executable
r -Mnqp -e 'my $start = INIT now; say "Loading file..."; my $content = "/home/longwalker/code/raku/y.try/json-decoding/AllPrintings.json".IO.slurp; say "Took {now - $start} seconds to load file into memory.";'
Loading file...
Took 4.564159965 seconds to load file into memory.
```

vs

```
raku -e 'my $start = INIT now; say "Loading file..."; my $content = "/home/longwalker/code/raku/y.try/json-decoding/AllPrintings.json".IO.slurp; say "Took {now - $start} seconds to load file into memory.";'
Loading file...
Took 10.147421301 seconds to load file into memory.
```

Of course, I expect that hardware and software based optimizations around caching frequently accessed data from storage makes this kind of benchmarking (already an inexact science) a bit more tricky.

niner:
As this is about a trade-off we also need numbers for how this affects loading of smaller files - which is arguably the far more common use case. What if the file is on remote storage like NFS where roundtrip times are much larger?

ugexe:
You are referencing `IO::Handle` but all the code is `IO::Path`

ab5tract:
> You are referencing `IO::Handle` but all the code is `IO::Path`

Thanks for catching this. I've adjusted the title.

src/core.c/IO/Path.rakumod
```diff
@@ -650,35 +654,38 @@ my class IO::Path is Cool does IO {
...
 
-        nqp::elems(my $blob := slurp-path-bin($path))
+        nqp::elems(my $blob := slurp-path-bin($path, $size))
           ?? nqp::join("\n",nqp::split("\r\n",nqp::decode($blob,$encoding)))
           !! ""
     }
 
     proto method slurp() {*}
     multi method slurp(IO::Path:D: :$bin!) {
+        my $size = max try self.s, fallback-slurp-size;
```
MasterDuke17, line 667: FYI, infix max (i.e., `self.s max fallback-slurp-size`) is quite a bit faster than sub/method max.

lizmat:
A thought: perhaps we could d a two-tier approach?

First read with the default size.  Then, if that read that amount of bytes, do a file system access to find out the size of the file in question.  And *then* read the rest.

For these mega files, it wouldn't make much difference as the initial size is a very small part of the total size.  And for small files, we wouldn't need to do the extra file system access to find out the size.

lizmat:
So should we close this PR in favour of a new one with the suggestion of https://github.com/rakudo/rakudo/pull/5867#issuecomment-3541439262 ?   Or should we build on this one?

PR #5869, opened 2025-04-29
Draft: Make Rakudo::Internals::JSON 4-6x as fast

ab5tract:
There are likely a number of ways to work around this. My first instinct is to just replace the insides of the `to-json`/`from-json` subs and then remove the warning about their use.

The old, "Raku" versions of those subs can go live in `lib/JSON.rakumod`, where anyone who wants to use a slower but 100% "Raku" version.

#### Beyond speed

By providing a fast JSON parsing option in core, the ecosystem no longer needs to rely on a bunch of NQP code underlying everything.


niner:
I don't like this approach very much. It pushes more code into the VM which is the wrong direction. The VM already does too much. Every complex operation pushed into the VM is a blocker for optimization. Dispatchers are a great example for how pushing functionality into HLL code can actually lead to performance improvements as this makes decisions visible to the specializer which then can often either optimize them away or use the checks to apply other optimizations to the guarded code. I do concede that in this case the performance benefits of having the VM doing the decoding will probably outweigh any penalties due to the opaque implementation for a very long time to come. But that still leaves that this is yet another burden on other backends, most notably the JVM.

I would love to see the effort getting spent on general optimizations that speed up the Raku implementation of JSON parsing. I consider even the NQP code in JSON::Fast a mistake. Ironically some of it is obsolete even now. E.g. a Raku `while` loop without phasers compiles exactly to the same nqp::while op as is used manually there but without having to make the code so hard to read.

lizmat:
> But that still leaves that this is yet another burden on other backends, most notably the JVM.

Hence my question in https://github.com/MoarVM/MoarVM/pull/1936#issuecomment-2834240882

> I would love to see the effort getting spent on general optimizations that speed up the Raku implementation of JSON parsing.

That would be nice indeed.  However, as far as I understand it, the JSON standard is a finite standard that does not allow any non-standard graphemes.  Thus, it can bypass the whole NFG mechanism, and be highly optimized.

If we hadn't chosen JSON for META information in our distributions, then I would be less willing to have this library in MoarVM.  However, we have chosen JSON, so we need a fast JSON decoder to handle anything related to module installation and loading.

> Ironically some of it is obsolete even now. E.g. a Raku while loop without phasers compiles exactly to the same nqp::while op as is used manually there but without having to make the code so hard to read.

I'd call it obsolete if the `nqp::while` would turn out to be *slower* than the HLL `while`.

Let me just say that my work on JSON::Fast (and thus the internal JSON implementation) helped save at least one major Raku project from being unacceptably slow.


In conclusion: I'd welcome a native JSON implementation in *all* backends, so that we can remove the Raku/NQP based JSON implementation from core.

ab5tract:
First, I'll say that it's my understanding that working on the optmizer as-is is a dead end at the moment.

But more broadly, I don't know how to write optimizer code that delivers 4-6x decode speedups. I'd be willing to try to learn but even so, I would *still* recommend having a dedicated, fast, externally maintained JSON parser unless our optimized version was genuinely faster than `yyjson`.

This really feels like a yes-and situation. Having fast JSON parsing may "prevent" (as in, not prod someone with hot coals until they code it from pure hate or spite) optimizer improvements, but it also opens up theoretical tuits for optimizer work that addresses other pain points.

niner:
> The performance benefit is indeed gone. But what is the point of changing it back just for the fact that we can? What happened to "if it ain't broke, don't fix it"?

Where did I demand changing of anything? If it were to be changed then I'd guess it's for the sake of maintainability of the code. Also arguably it is broken as it uses internal interfaces that are not guaranteed.

lizmat:
> Where did I demand changing of anything?

I've interpreted "obsolete" in:  "Ironically some of it is obsolete even now. E.g. a Raku while loop without phasers compiles exactly to the same nqp::while op as is used manually there but without having to make the code so hard to read." as such.

> Also arguably it is broken as it uses internal interfaces that are not guaranteed.

I thought it was ok to use internal interfaces inside the core?

niner:
> I thought it was ok to use internal interfaces inside the core?

I was talking about JSON::Fast

lizmat:
Finding out how much *reading* JSON is going to be needed, I think it *is* worthwhile to pursue this in the near future, even though the JVM would need to keep the old code around for the foreseeable future.

ab5tract:
So I will try to revive this PR against current `main`.

Here is a list of what I understand to be currently missing:

- [ ] Add encode path
- [ ] Add `:immutable`/`:im` flag
- [ ] ???
- [ ] Parse JSON quickly

I'm  was going to suggest a `:mutable` flag instead, defaulting the parser immutable semantics and (hopefully) speed. But then I realized for compatibility purposes, that's probably a no go.

PR #5879, opened 2025-05-12
Enhance feedback around named subset parameters

ab5tract:
Adding a worry here provides users with a bit of
helpful feedback when they define an optional
named parameter with a subset type but fail to
provide a valid default value.

Default values are also checked for conformance
to the provided subset, with the obvious caveat
that an expression without a compile time value
will pass through this protection.

### DIFF
--- a/src/Raku/ast/signature.rakumod
+++ b/src/Raku/ast/signature.rakumod
@@ -1212,16 +1212,29 @@ class RakuAST::Parameter
         }
 
         if $!type {
-            my $archetypes := $!type.compile-time-value.HOW.archetypes;
+            my $param-type := $!type.compile-time-value;
+            my $archetypes := $param-type.HOW.archetypes;
             unless $archetypes.nominalish
                 || $archetypes.generic
                 || $archetypes.definite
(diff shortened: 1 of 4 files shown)


src/Raku/ast/signature.rakumod
```diff
@@ -111,6 +111,27 @@ class RakuAST::Signature
                         }
                         %seen{$name} := 1;
                     }
+
+                    unless $_.is-declared-required || nqp::eqaddr($_.type, RakuAST::Type) {
+                        my $param-type := $_.type.meta-object;
+                        my $param-default := $_.default;
+                        if nqp::istype($param-type.HOW, Perl6::Metamodel::SubsetHOW) {
+                            if ! nqp::eqaddr($param-default, RakuAST::Expression) {
+                                if nqp::can($param-default, 'maybe-compile-time-value')
```
niner, line 120: The proper check is $param-default.has-compile-time-value. If that returns true, the object must support maybe-compile-time-value.

niner (changes requested):

niner (approved):

Issue #5884, opened 2025-05-19
DateTime's built-in methods disregarding timezone


arkiuat:
DateTime's julian-date method seems to be ignoring timezones in Rakudo v2024.07, MoarVM 2024.07, Raku v6.d.  
For example, the following produces 23 identical lines of output.

for -11..11 {DateTime.new("2025-05-19T17:00:00"~"%+03d".sprintf($_)).julian-date.say}

lizmat:
@tbrowder looks like this is up your alley?

arkiuat:
I should have poked around some more: the trouble may be in DateTime's .new rather than in .julian-date

The following produces two sets of 11 identical outputs, with all 23 in UTC (but only the middle specifying the timezone with Z, the others as -00 or +00

for -11..11 {DateTime.new("2025-05-19T17:00:00", timezone => $_).say}

tbrowder:
> [@tbrowder](https://github.com/tbrowder) looks like this is up your alley?

I may have lost my 'tuits by now, but I'll be glad to take a look.

tbrowder:
arkiuat: please show the output you get for the three hours times around Z and what you expect to see.

arkiuat:
@tbrowder Oh, right! It is in seconds, and I knew that but failed to make the connection this time. Sorry about that.

And thank you for the explanation. I'll definitely think about other ways for me to go about things before requesting an addition to core.

And thank you very much for distro DateTime::Julian, which I've already started looking into.

tbrowder:
My problem is the definition is wrapped up with the observer's longitude, not timezone directly. See my comments on the docs PR #4579.

tbrowder:
@lizmat How do you feel about adding :longitude to .julian-date and .modified-julian-date. Then the result might look something like this:

    2025-05-22T04:05:01.005Z at longitude: -90.1

lizmat:
@tbrowder That feels like it doesn't belong there?  Unless it's intended to somehow convert to solar time?

tbrowder:
I agree. Ignore that, please.

PR #5886, opened 2025-05-24
Make DateTime.(modified-)julian-date take timezone into account

lizmat:
In response to https://github.com/rakudo/rakudo/issues/5884

### DIFF
--- a/src/core.c/DateTime.rakumod
+++ b/src/core.c/DateTime.rakumod
@@ -525,7 +525,7 @@ my class DateTime does Dateish {
     }
 
     method modified-julian-date(DateTime:D: --> Real:D) {
-        self.daycount + self.day-fraction
+        .daycount + .day-fraction with self.utc
     }
 
     method julian-date(DateTime:D: --> Real:D) {


lizmat:
FWIW, the reason I made this PR is because I feel (and some other core devs as well) that this is the proper way to fix #5884 .  If your DateTime::Julian produces the same values, than I think we don't need a module, but instead merge this PR and update the documentation accordingly and add tests to roast

tbrowder:
Okay, I will not object and am onboard. And I will help where I can. 

But my module has lots of stuff that is probably too much for core, (unless you add the conversion of julian-date back to DateTime).

ab5tract:
> But my module has lots of stuff that is probably too much for core, (unless you add the conversion of julian-date back to DateTime).

Personally I would expect such parity in core, depending on how much complexity is involved of course.

ab5tract:
> I intend to merge this after the 2025.06 release.

I don't see a corresponding "from-julian-date" method on DateTime (it only needs to be a class method).

I think it makes sense to fill this gap comprehensively.

lizmat:
@ab5tract: perhaps, but I feel that's a different issue altogether.  And thus would need a separate PR

PR #5913, opened 2025-06-20
Don't use two spaces after . or ?

coke:
Per https://www.chicagomanualofstyle.org/qanda/data/faq/topics/OneSpaceorTwo.html et al.


coke:
No tests for this method.

lizmat:
It's marked as " is implementation-detail" at the moment.

Issue #5921, opened 2025-06-29
Error running `raku -V` on MacOS CI


patrickbkr:
The MacOS CI newly fails in `t/02-rakudo/17-cli.t`. There `raku -V` dies and prints a stack trace.

The stack trace:

```
Unhandled exception: No exception handler located for warn
   at SETTING::src/core.c/control.rakumod:267  (/Users/runner/work/1/install/share/perl6/runtime/CORE.c.setting.moarvm:warn)
 from SETTING::src/core.c/Nil.rakumod:70  (/Users/runner/work/1/install/share/perl6/runtime/CORE.c.setting.moarvm:Str)
 from SETTING::src/core.c/Distro.rakumod:93  (/Users/runner/work/1/install/share/perl6/runtime/CORE.c.setting.moarvm:)
 from SETTING::src/core.c/Distro.rakumod:61  (/Users/runner/work/1/install/share/perl6/runtime/CORE.c.setting.moarvm:)
 from SETTING::src/core.c/Rakudo/Internals.rakumod:1176  (/Users/runner/work/1/install/share/perl6/runtime/CORE.c.setting.moarvm:)
 from SETTING::src/core.c/Rakudo/Internals.rakumod:1155  (/Users/runner/work/1/install/share/perl6/runtime/CORE.c.setting.moarvm:)
 from SETTING::src/core.c/Lock.rakumod:30  (/Users/runner/work/1/install/share/perl6/runtime/CORE.c.setting.moarvm:protect)
 from SETTING::src/core.c/Rakudo/Internals.rakumod:1152  (/Users/runner/work/1/install/share/perl6/runtime/CORE.c.setting.moarvm:INITIALIZE-DYNAMIC)
 from SETTING::src/core.c/stubs.rakumod:38  (/Users/runner/work/1/install/share/perl6/runtime/CORE.c.setting.moarvm:DYNAMIC)
 from SETTING::src/core.c/Compiler.rakumod:51  (/Users/runner/work/1/install/share/perl6/runtime/CORE.c.setting.moarvm:verbose-config)
 from <unknown file>:1  (<ephemeral file>:<unit>)
 from <unknown file>:1  (<ephemeral file>:<unit-outer>)
 from NQP::src/HLL/Compiler.nqp:197  (/Users/runner/work/1/install/share/nqp/lib/NQPHLL.moarvm:eval)
[...]

PR #5922, opened 2025-06-29
Fix `raku -V` on strange non-standard MacOS

patrickbkr:
There exist non-standard MacOS versions out there (specifically MacOS-15 on Azure CI) that have this in their license file:

SOFTWARE LICENSE AGREEMENT FOR APPLE CONFIDENTIAL macOS PRE-RELEASE SEED SOFTWARE

instead of the expected

SOFTWARE LICENSE AGREEMENT FOR macOS SomeReleaseName

This breaks filling in the Distro object. Fix by checking whether the respective regex succeeds or not.

@lizmat I'm requesting your review, as I de-NQPized this piece of code and I'm unsure about the perf implications.


patrickbkr:
Fixes https://github.com/rakudo/rakudo/issues/5921

patrickbkr:
> Feels this is a more appropriate solution, could you check?

@lizmat That's why I pinged you. Thanks!

PR #5926, opened 2025-07-07
Optimize unsigned native ++/--

MasterDuke17:
Check for `uint` in addition to `int` and then change the `assign_*` as needed.

Now `my uint $a; $a++ for ^10_000_000; say $a; say now - INIT now` takes ~0.24s, same as the `int` version, where before it was ~1.15s.


lizmat:
Thanks!

MasterDuke17:
> Note this was reverted for release because it broke the blin run. See [451be07](https://github.com/rakudo/rakudo/commit/451be071c02d95e0f2e2cbd06074ae26c14b612b) and #5941 .

Hm. I probably won't be able to look at this for a couple days. I'm surprised by the fails though, so hopefully it's an easy fix....

PR #5936, opened 2025-07-29
Change enum's raku method to return valid code

### LINKED ISSUE #5935: `raku` method returns invalid code for emoji enum members
So, admittedly this is somewhat contrived (and I did stumble upon it by accident) but I think this should work:

```raku
enum Directions <⬆️>; say Directions::<⬆️>.raku.EVAL;
```

However, it fails with:

```
===SORRY!=== Error while compiling /Users/julia/code/energieprijzen/EVAL_0
Bogus postfix
at /tmp/EVAL_0:1
------> Directions::⏏⬆️
    expecting any of:
        infix
        infix stopper
        statement end
        statement modifier
        statement modifier loop
```

This is because the `raku` method returns `Directions::⬆️`, which is not valid syntax. Now, [the documentation for the `raku` method](https://docs.raku.org/routine/raku#___top:~:text=For%20plain%20objects%2C%20it%20will%20conventionally%20return%20a%20representation%20of%20the%20object%20that%20can%20be%20used%20via%20EVAL%20to%20reconstruct%20the%20value%20of%20the%20object.) says (emphasis added)

> [i]t will **conventionally** return a representation of the object that can be used via [`EVAL`](https://docs.raku.org/routine/EVAL) to reconstruct the value of the object.

But I still think `raku` should always work for enum members, even though this is a bit of a weird case.

heyajulia:
Fixes #5935.

### DIFF
--- a/src/core.c/Enumeration.rakumod
+++ b/src/core.c/Enumeration.rakumod
@@ -14,7 +14,7 @@ my role Enumeration {
     method pair(::?CLASS:D:) { $!key => $!value }
 
     multi method gist(::?CLASS:D:) { $!key                     }
-    multi method raku(::?CLASS:D:) { self.^name ~ '::' ~ $!key }
+    multi method raku(::?CLASS:D $self:) { "$self.^name()::<$!key>" }
 
     multi method pick(::?CLASS:U:)       { self.^enum_value_list.pick     }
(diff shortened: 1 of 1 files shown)


lizmat:
Thanks!

PR #5946, opened 2025-09-03
Fix Hash return type error message to match Array behavior and remove confusing gist

### LINKED ISSUE #5845: "You have to pass an explicitly typed..." message missing for typed Hash.
Compare those two cases:

```
$ raku -e 'sub foo ( --> Array of Str ) {  my @a = "foo"; return @a }; foo()'

Type check failed for return value; expected Array[Str] but got Array (["foo"]).  You have to pass an
explicitly typed array, not one that just might happen to contain
elements of the correct type.
```

```
$ raku -e 'sub foo ( --> Hash of Str ) {  my %h = "foo" => "bar"; return %h }; foo()'

Type check failed for return value; expected Hash[Str] but got Hash ({:foo("bar")})
```

Hash message is confusing, because it is missing hint that it has to be explicitly typed, not just contain values of given type.

IMO Both messages shold also not contain gist. When user sees `Array (["foo"])` there is nothing wrong from his perspective, this looks like Array of Str. I think message should only show `of()` type without values, like `expected Array[Str] but got Array[Mu]`.

ttn-ttn:
Fixes #5845.
I'm very new with Raku, so any feedback will be appreciated :)

### DIFF
--- a/src/core.c/Exception.rakumod
+++ b/src/core.c/Exception.rakumod
@@ -2845,7 +2845,7 @@ my class X::TypeCheck is Exception {
         nqp::eqaddr($!got.WHAT, $!expected.WHAT)
           ?? $raku
           !! nqp::can($!got.HOW, 'name')
-            ?? "$!got.^name() ($raku)"
+            ?? "$!got.^name()"
             !! $raku
     }
     method expectedn() {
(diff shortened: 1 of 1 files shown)


lizmat:
Thanks, good catch!

PR #5949, opened 2025-09-07
Add Str.spurt method to 6.e

lizmat:
To easily spurt the contents of a string to a file, with automatic coercing of the path to an IO::Path.

As suggested in https://www.nntp.perl.org/group/perl.perl6.users/2025/09/msg11499.html

### DIFF
--- a/src/core.e/Fixups.rakumod
+++ b/src/core.e/Fixups.rakumod
@@ -404,6 +404,8 @@ augment class Str {
           $needle, $pos, :ignorecase($smartcase && $needle!lowercase-only), |%_
         )
     }
+
+    method spurt(Str:D: IO() $io) { $io.spurt(self) }
 }
 
 #-------------------------------------------------------------------------------


jubilatious1:
Not sure this has gone through enough discussion to implement/approve a Pull Request.

lizmat:
> Not sure this has gone through enough discussion to implement/approve a Pull Request.

For sure: the PR is ready to merge, but is also ready to discuss.  I'll make it a draft to make it clear.

b2gills:
While I agree that the one liner looks cool. There is basically no routines where the arguments change position depending on the types of the argument. Which helps keep things consistent. 

A better idea would be to use a different name for the method.
For example `"Str".spurt-to("filename")`

There may be a better name than that.

lizmat:
I think this is as good a place as any to discuss  :-)

ugexe:
I'm not sure I find the argument given compelling for adding something to the core. Something I personally find myself asking when these type of feature requests come in is what do I think Larry would have done if that same request came in as an RFC for Perl 6. While I personally can't speak for anyone but myself I do find it hard to believe he would have thought something along the lines of "yes need yet another way to spurt to a file". Taken to its logical conclusion one might argue we should have multiple versions of every IO related function, and as that doesn't seem like a good idea it does (to me) suggest this specific case is also not a good idea either.

Issue #4991, opened 2022-07-20
feed operator does not recognise sub-form of method calls


gfldex:
```
 my @a = <a b c d>;
 my $c = class C {
    our sub foo(*@a) { @a };
    method foo(*@a) { @a }
}.new;
@a ==> $c::foo() ==> say();
@a ==> foo($c:) ==> say();
```

output:
```
===SORRY!=== Error while compiling /home/dex/projects/raku/tmp/2021-03-08.raku
Only routine calls or variables that can '.push' may appear on either side of feed operators.
at /home/dex/projects/raku/tmp/2021-03-08.raku:1848
------>     @a ==> ⏏foo($c:) ==> say();
```

exected output:

```
[a b c d]
[a b c d]
```

This is quite surprising to me. I believe the sub-form of method calls was added to support feed operators for OO-folk.

## Environment
* Operating system: Linux dexhome 5.15.0-2-amd64 #1 SMP Debian 5.15.5-1 (2021-11-26) x86_64 GNU/Linux
* Compiler version (`perl6 -v` or `raku -v`): Welcome to Rakudo™ v2022.06-33-g6dc2f92c9. Implementing the Raku® Programming language v6.d. Built on MoarVM version 2022.06-2-gcc49dcf93.

ab5tract:
It's definitely a fair bit of workaround, but the following does work:

```
 my @a = <a b c d>;
 my $c = class C {
    # This sub turns out to be unnecessary for this workaround 
    # our sub foo(*@a) { @a }; 
    method foo(*@a) { @a }
}.new;
@a ==> {$c::foo(@_)}() ==> say();
@a ==> {foo($c: @_)}() ==> say();
# [a b c d]
# [a b c d]
```

That said, I'd like to see this working without the extra hoops in RakuAST, which currently gives the error:

```
> set-rakuast; ./rakudo-m issues/r4991.raku
===SORRY!===
Cannot find method 'PRECURSOR' on object of type QAST::Op
```

ab5tract:
@gfldex Enjoy this functionality! It single-handedly turned me around to the utility of indirect method calls,

PR #5967, opened 2025-10-04
RakuAST - Fix indirect methods with feed operators

ab5tract:
There is an indirect method call syntax, eg:

	foo-method($foo: |@args)

This should work with feeds, such that:

	@a ==> foo-m($foo-o:) ==> say()
	# OUTPUT: [a b c d]

Previously this was a compilation error in base,
despite this being an otherwise valid form for
the feed operator to handle.

Addresses R#4991 (#4991)


lizmat:
I welcome this PR, but feel that it should be noted that QAST manipulation is what RakuAST was supposed to prevent?

lizmat (approved):

Issue #5652, opened 2024-10-01
Can the parser for $] be made a little less agressive?


thundergnat:
## The Problem

The compiler dies with a fatal error if it detects what it _thinks_ is a $] variable, even if there isn't one. Have been bitten by this several times. Very easy to work around, but LTA.

Contrived example:

`my @a = ^9; say @a[++$] for ^5;`

Yeilds (slightly edited): 
```
SORRY!  Error while compiling  code
Unsupported use of $] variable; in Raku please use $*RAKU.version or $*RAKU.compiler.version
```
Just adding a space between the $ and ] "fixes" the problem, but it irritates me.

`my @a = ^9; say @a[++$ ] for ^5;`

> 1
> 2
> 3
> 4
> 5

This was brought up in #1356, but wasn't really addressed. 

 I'm not even asking to remove it completely, just to make the parser a little less aggressive in assuming _any_ presence is an attempt to access legacy special variables.

lizmat:
Well, I think that's one of the cases in Raku where whitespace **is** important.

FWIW, in RakuAST it fails in a different way:
```
% RAKUDO_RAKUAST=1 raku -e 'my @a = ^9; say @a[++$] for ^5;'
===SORRY!=== Error while compiling -e
Unsupported use of $] variable; in Raku please use $*RAKU.version or $*RAKU.compiler.version
```
And then applying `isms`:
```
% RAKUDO_RAKUAST=1 raku -e 'use isms; my @a = ^9; say @a[++$] for ^5;'
===SORRY!=== Error while compiling -e
Unable to parse expression in subscript; couldn't find final ']' (corresponding starter was at -e:1)
```
I'm not sure that's an improvement or not

ab5tract:
I just saw this issue after re-opening and commenting on  #3295.

In RakuAST all we have to do to make this work is move on from these training-wheels-that-break-the-new-bike exceptions:

```
> g diff
diff --git a/src/Raku/Grammar.nqp b/src/Raku/Grammar.nqp
index a4a52970d..f4afb6941 100644
--- a/src/Raku/Grammar.nqp
+++ b/src/Raku/Grammar.nqp
@@ -3558,11 +3562,6 @@ grammar Raku::Grammar is HLL::Grammar does Raku::Common {
         <.obsvar: '$?'>
     }

-    token special-variable:sym<$]> {
-        <.sym> {} <!before \w | '('>
-        <.obsvar: '$]'>
-    }
-
[...]

PR #5969, opened 2025-10-06
RakuAST - Remove special handling of '$]'

ab5tract:
This commit addresses R#5652 (#5652), which noted the following malfunction: using a state variable inside of a positional lookup, eg. `[++$]` caused a SORRY exception related to informing the user of differences between Perl and Raku.

So this commit is also a bit of an instigation to ask ourselves how badly we want these handrails to stay with us.

I don't necessarily think they should all be removed. But... ... it seems like we can safely eject them where doing so enables otherwise valid Raku to compile and run.

### DIFF
--- a/src/Raku/Grammar.nqp
+++ b/src/Raku/Grammar.nqp
@@ -3690,11 +3690,6 @@ grammar Raku::Grammar is HLL::Grammar does Raku::Common {
         <.obsvar: '$?'>
     }
 
-    token special-variable:sym<$]> {
-        <.sym> {} <!before \w | '('>
-        <.obsvar: '$]'>
-    }
-
     regex special-variable:sym<${ }> {
         <sigil>
         '{'


ab5tract:
Another option could be to parse these and only complain about them if they were valid syntax:

```
[++$]] # throws a SORRY
[++$]  # not SORRY
```

I think that should be manageable via relatively clean RakuAST code, right?

niner:
While personally I found those P5ism warnings very useful when learning Raku, I struggle to see the utility of this particular one. `$]` in Perl is the version of the compiler. Its ENGLISH counter part is `$OLD_PERL_VERSION` as it has been made obsolete by `$^V` which is not just a number, but a more useful version object. Thus even in Perl `$]` should not be in use anymore and quite rare. What's the chance of someone being into Raku enough to want to check the actual Raku compiler version but not enough to have the intuition that this info won't be in an obscure punctuation variable? Personally I'd just take this commit as it is and move on with my life.

PR #5975, opened 2025-10-13
[Draft] RakuAST - Address some issues with generics

ab5tract:
Previously the following code:

	role R[::T] { has @.a is Array[T] }
	dd R[Int].new: :a(1,2,3)

 returned:

	Type check failed in assignment to ; expected T but got Int

This now works as expected.

Furthermore, more advanced cases work as expected:

	my \r = my role R[::T] {
	    my package G {
	        class A is Array[T] {}
	    }
	    has @.a is G::A;
	    method aa { G::A:D }
	}

	my class CInt does R[Int] { }
	dd CInt.aa;
	dd CInt.new(:a(1,2,3));

 produces:

	Int  = R::G::A[Int]:D
	CInt.new(a => [1, 2, 3])

There is also a regression in the binding of parameterized definites,
as tested in `xx-fixed-in-rakuast.rakutest`:

	dies-ok {
	    my Positional:D[Int] $x := ("x", "b")
	}

This currently doesn't die, though this might not be a blocker.

### DIFF
--- a/src/Raku/ast/type.rakumod
+++ b/src/Raku/ast/type.rakumod
@@ -474,8 +474,22 @@ class RakuAST::Type::Parameterized
             my $args := $!args.IMPL-COMPILE-TIME-VALUES;
             my @pos := $args[0];
             my %named := $args[1];
-            my $ptype := self.IMPL-BASE-TYPE.compile-time-value;
-            $ptype.HOW.parameterize($ptype, |@pos, |%named)
+            my $ptype := self.base-type.compile-time-value;
+            my $phow  := $ptype.HOW;
+
+            # There are cases where the type is actually incapable of
+            # parameterization, at which point we pass it back unmodified.
+            if nqp::istype($phow, Perl6::Metamodel::ClassHOW) {
+                nqp::can($phow, 'parameterize')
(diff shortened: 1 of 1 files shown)


src/Raku/ast/type.rakumod
```diff
@@ -474,8 +474,22 @@ class RakuAST::Type::Parameterized
...
             my @pos := $args[0];
             my %named := $args[1];
-            my $ptype := self.IMPL-BASE-TYPE.compile-time-value;
-            $ptype.HOW.parameterize($ptype, |@pos, |%named)
+            my $ptype := self.base-type.compile-time-value;
+            my $phow  := $ptype.HOW;
+
+            # There are cases where the type is actually incapable of
+            # parameterization, at which point we pass it back unmodified.
+            if nqp::istype($phow, Perl6::Metamodel::ClassHOW) {
```
MasterDuke17: Can't this whole if/else be simplified to just the else block?

ab5tract: I have it arranged this way due to experimentation / trial and error, but yes, you are correct.

Note also that this commit is not performing as advertised due to a mess up with my environment variables during testing.

Issue #5912, opened 2025-06-19
Using a subset as a type sets the *base* type as default in v6.e.PREVIEW


lizmat:
In 6.d:
```
% raku -e 'subset A of Int; my A $a; say $a.VAR.default'
(A)
% raku -e 'subset A of Int; my A @a; say @a.default'
(A)
% raku -e 'subset A of Int; my A @a; say @a.default'
(A)
```
whereas in 6.e.PREVIEW:
```
% raku -e 'use v6.e.PREVIEW; subset A of Int; my A $a; say $a.VAR.default'
(Int)
% raku -e 'use v6.e.PREVIEW; subset A of Int; my A @a; say @a.default'
(Int)
% raku -e 'use v6.e.PREVIEW; subset A of Int; my A @a; say @a.default'
(Int)
```
This e.g. causes `$foo =  Nil` to reset to the wrong value in 6.e.

ab5tract:
IIUC, the behavior in @lizmat's latest comment iis related to the same change that produced #5915.

Basically we are no longer allowing undefined values to magically pass `where` clauses while being undefined and technically not matching said `where` clause. (I consider the `use of uninitialized` exception firing to be LTA, though, as the real exception becomes buried).

I'd appreciate a sanity check on this understanding though.

Regardless, I expect that the above should not affect subsets that do not use where clauses.

PR #5996, opened 2025-10-26
RakuAST/v6.e - Fix subset type defaults in containers

ab5tract:
The proper place for maybe-nominalization (which in turn also involves maybe-definite-ization, for versions < v6.e.

Only RakuAST needs to be fixed here, thankfully, as I have had way too high a does of metamodel internals for the moment.

Essentially this change is required because subsets are technically nominalizable, but the call to nominalizable returns the refinee ($!of) or the nominalized refinee.

But the base-case (ie, before these nominalizable checks throughout Perl6::Metamodel classes was added) returned the subset type object directly. (Subsets are apparently always nominalizable, as the code attests and my experiments confirmed).

In order to return that state of affairs, we just call IMPL-MAYBE-NOMINALIZE as we generate type metaobjects.

R#5912 (#5912)

### DIFF
--- a/src/Raku/ast/type.rakumod
+++ b/src/Raku/ast/type.rakumod
@@ -55,14 +55,11 @@ class RakuAST::Type
         # If type does LanguageRevision then check what language it was created with. Otherwise base decision on the
         # current compiler.
         my $v-how := $v.HOW;
-        !$v-how.archetypes($v).coercive
-            && (nqp::can($v-how, 'language_revision')
-                    ?? $v-how.language_revision($v) < 3
-                    !! nqp::getcomp('Raku').language_revision < 3)
+        (nqp::can($v-how, 'language_revision')
+                ?? $v-how.language_revision($v) < 3
+                !! nqp::getcomp('Raku').language_revision < 3)
             ?? self.IMPL-MAYBE-DEFINITE-HOW-BASE($v)
-            !! ($v-how.archetypes($v).nominalizable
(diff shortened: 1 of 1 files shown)


ab5tract:
I've marked this as a draft due to the following failure. In base this file also fails but it passes two more tests before dying.

```
> set-rakuast; set -gx WTF 1; r --ll-exception t/spec/S02-types/nominalizables.t 
RAKUDO_RAKUAST set
1..4
# Subtest: Subset of a coercion
    1..7
    ok 1 - variable of subset type has the right type
    ok 2 - the variable got the right value
Type check failed in binding to parameter '$p'; expected Int but got Str ("13")
```

2colours:
EDIT: nevermind, this is about subset types.

<del>This is something I wanted to address but wasn't able to.</del>

<del>Did you check `&variable`s as well, and in particular `my ReturnType &variable`s?</del>

ab5tract:
> EDIT: nevermind, this is about subset types.
> 
> ~This is something I wanted to address but wasn't able to.~
> 
> ~Did you check `&variable`s as well, and in particular `my ReturnType &variable`s?~

Just a note, not sure if it relates to your strikethrough or not.

Technically this affects the timing of nominalizations for all nominalizing types. By pushing it to the exact encapsulation in charge of producing the type, things already seem to be running smoother. There may be a few other major gobstoppers around here. 

@2colours do you know of a ticket for whatever behavior your were describing?

PR #5999, opened 2025-10-29
Prevent issues when multiple threads try to load the same precomp

### LINKED ISSUE #1920: Module loading is not thread safe
## The Problem

```
perl6 -e '\
    class TestMe {\
        has @!modules = <Test NativeCall Telemetry>;\
        method test-method {\
            gather for @!modules -> $module { take (require ::($module)) }\
        }\
    };\
    await Promise.allof: (1..10).map({ start { TestMe.new.test-method.map({ $_ }).sink; } });\
    say "DONE";
'
```

## Observed output
```
Segmentation fault: 11
```

```
moar(5487,0x70000ede8000) malloc: *** error for object 0x7ffb2254e1f0: pointer being freed was not allocated
*** set a breakpoint in malloc_error_break to debug
Abort trap: 6
```

```
Use of uninitialized value of type Any in string context.
Methods .^name, .perl, .gist, or .say can be used to stringify it to something meaningful.
Use of uninitialized value of type Any in string context.
Methods .^name, .perl, .gist, or .say can be used to stringify it to something meaningful.
  in block  at -e line 5
  in block  at -e line 5
Segmentation fault: 11
```

## Environment
* Compiler version (`perl6 -v`): probably affects all releases, but tested on 2017.10 and blead

timo:
Until now, we were only locking the hash of already `$loaded` precomp files for the moment where we check if something is in it for the slot we are interested in, and then again when we want to put something new in.

However, we were not preventing the situation where multiple threads check at the same time, all get the answer "nothing there yet", and all going ahead to load the precomped file, and then setting the result in the hash regardless of a value that may already have appeared in the mean time.

Another approach to this would be simply to not install a result after loading the precomp file if another value has already appeared in the mean time, and throwing the redundant loaded bytecode files away. However, this means that threads will do unnecessary work if we're unlucky.

The approach in this commit installs a Promise in the `$loaded` hash if there was nothing there yet, and awaits the result if there was a Promise at that time. When the precompilation file is loaded, the promise is kept (for any waiting threads) and the slot is replaced with the result (so that we don't pay the overhead of Promises after this). If there was an issue that means the thread that "owns" the Promise can't fulfill it, it breaks the Promise and frees the slot in the hash. Waiting threads will then try again, meaning they will either install a Promise of their own if applicable, or get a "nothing here" result.

This seems to fix #1920 and offers a quite fine locking granularity.

### DIFF
--- a/src/core.c/CompUnit/PrecompilationRepository.rakumod
+++ b/src/core.c/CompUnit/PrecompilationRepository.rakumod
@@ -249,9 +249,17 @@ Need to re-check dependencies.")
         $loaded-lock.protect: {
             for $dependencies.List -> $dependency-precomp {
                 my str $key = $dependency-precomp.id.Str;
-                nqp::bindkey($loaded,$key,
-                  self!load-handle-for-path($dependency-precomp)
-                ) unless nqp::existskey($loaded,$key);
+
+                my $loaded-dep := self!get-already-loaded($key);
+
+                # If we've promised to load this file, load it.
+                # Otherwise the dependency precomp has already been
+                # loaded and we can just go on.
+                if $loaded-dep ~~ Promise {
+                  my $loaded-handle := self!load-handle-for-path($dependency-precomp);
+                  $loaded-dep.keep($loaded-handle);
+                  nqp::bindkey($loaded, $key, $loaded-handle);
+                }
 
                 $dependency-precomp.close;
             }


timo:
I think the handling of the `$loaded` hash in `method !load-dependencies` is not correct, but I think I first have to build an example that makes it actually do anything wrong

timo:
https://gist.github.com/timo/c206a75790f5571912382f133d788c99 - here's a generator for a distro with a configurable amount of modules in it that have dependencies a random amount of levels deep. it seems to run just fine with the rakudo branch applied.

PR #6013, opened 2025-11-08
Don't allow a Range in an array shape specification

lizmat:
Because it probably does *not* do what you think it does.

Before this, my @a[1..4] would create a 4-dimensional array, with dimensions 1,2,3,4.  Instead of defining a 1-dimensional array with indices 1 through 4.


ab5tract:
Looks good to me!

PR #6093, opened 2026-03-22
Make $?SOURCE / $?CHECKSUM available on 6.e and higher

lizmat:
Although this is mostly intended to be used by the new interactive debugger that @patrickb is working on, it feels wrong to introduce these compile time variables in 6.c and 6.d as well, this late in the 6.d lifetime cycle.

Hence this commit, making them *unavailable* on 6.c and 6.d

### DIFF
--- a/src/Perl6/Actions.nqp
+++ b/src/Perl6/Actions.nqp
@@ -1353,11 +1353,13 @@ class Perl6::Actions is HLL::Actions does STDActions {
             statementlist_with_handlers($/)
         );
 
-        $world.install_lexical_symbol($unit, '$?SOURCE', $*OMIT-SOURCE
-            ?? $world.find_single_symbol_in_setting('Nil')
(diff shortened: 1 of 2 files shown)


ab5tract:
I agree with @patrickbkr that revision gating is problematic.

When described, I was imagining that this feature was significantly easier to introduce in RakuAST and so we would just not bother with the old frontend compiler. A sort of revision gating by compiler version, if you will.

But I see right away that there is code in Perl6/Actions.

Am I correct in understanding that both of these variables are entirely new?

I tend to see revision gating as necessary only if there is a behavioral shift of an existing feature that changes between revisions.

Even if that were to be the case here, I think the benefits would be  significant enough to change the behavior without revision gating (presuming that the existing behavior is broken). In such cases I would propose a pragma to provide the old behavior.

tl;dr -- I don't see anything wrong in adding these variables backwards into 6.c/6.d because a good debugger is worth it

PR #6107, opened 2026-04-05
Add `item` to error message stating allowed parameter traits

schultzdavid:
Also a slight re-ordering: move `raw` one to the left, the idea being: 
First list the parameters that affect how the parameter behaves inside the function body (rw, readonly, copy, raw). Then list the parameters that affect dispatch but not parameter behavior in the body (required, item).


ab5tract:
Thanks David!

PR #6110, opened 2026-04-11
Fix SETTING:: prefix having literal single quotes on Windows

### LINKED ISSUE #6109: Junctions not working properly on windows
This code

```
<a b c>.all ~~ Str:D
```

Returns ```True``` on linux/mac, but ```False``` on windows.

ugexe:
The Makefile template used shell-style single quotes around the SETTING:: prefix passed to gen-cat.nqp. On Unix the shell strips these, but on Windows cmd.exe passes them literally, causing .file to return 'SETTING::'src/core.c/Mu.rakumod instead of SETTING::src/core.c/Mu.rakumod.

See: https://irclogs.raku.org/raku/2026-04-11.html#06:20

Resolves #6109


lizmat:
Confirmed this doesn't break the build on MacOS

coke:
Confirmed this fixes the bug on windows!

coke:
Let's also add a test for this in rakudo/t

lizmat:
Thanks!

PR #6114, opened 2026-04-12
Fix type smiley enforcement on anonymous params in list assignment

### LINKED ISSUE #6111: type smilies are ignored in list assigment
As discussed here: https://irclogs.raku.org/raku-dev/2026-04-11.html#16:30

The following does not fail, neither at runtime nor compile-time, but should.

```
my ($foo, Any:D, $buzz) = (42, Any, 'lightyear');
```

As found in `S06-signature/tree-node-parameters.t` line 78, type checks on `$` placeholders are specced. I did not find any test in Roast for the case with type-smilies.

The following workaround can be used and fails as excpected.

```
my ($bar, $ where .defined) = (42, Any);
```

ugexe:
`my ($foo, Any:D, $buzz) = (42, Any, 'lightyear')` should fail because `Any` is undefined, but the `:D` constraint was silently ignored on both frontends. The RakuAST frontend had an additional bug where `$buzz` received `(Any)` instead of 'lightyear'.

The RakuAST issue was that anonymous typed parameters (those with no target variable) were skipped entirely when building the LHS of the assignment, so all subsequent values shifted by one position. The fix creates a placeholder container for these parameters. On the old frontend, `dissect_type_into_parameter` nominalizes `Any:D` to `Any` and stores the definedness constraint in a `defined_only` flag, but the anonymous container path never read that flag. The fix reconstructs the definite type before creating the container descriptor.

After this change, both frontends correctly reject type smiley violations in list assignment:

`Type check failed in assignment to anon; expected Any:D but got Any (Any)`

The RakuAST frontend now also assigns the correct value to `$buzz`:

```
RAKUDO_RAKUAST=1 raku -e 'my ($foo, Int:D, $buzz) = (42, 99, "lightyear"); say "$foo $buzz"'

42 lightyear
```

Resolves #6111

### DIFF
--- a/src/Perl6/Actions.nqp
+++ b/src/Perl6/Actions.nqp
@@ -3664,8 +3664,19 @@ class Perl6::Actions is HLL::Actions does STDActions {
                 else {
                     my $world := $*W;
                     $world.handle_OFTYPE_for_pragma($/,'parameters');
+                    my @value_type := $*OFTYPE ?? [$*OFTYPE.ast] !! [];
+                    if @value_type && $_<defined_only> {
+                        @value_type[0] := $world.create_definite_type(
+                            $world.resolve_mo($/, 'definite'),
+                            @value_type[0], 1);
+                    }
+                    elsif @value_type && $_<undefined_only> {
+                        @value_type[0] := $world.create_definite_type(
+                            $world.resolve_mo($/, 'definite'),
+                            @value_type[0], 0);
+                    }
                     my %cont_info := $world.container_type_info($/, :$post,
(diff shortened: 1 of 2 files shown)


ugexe:
I don't know enough about either frontend to say how appropriate these changes are. I feel like it could be a bit simpler but I haven't been able to put my finger on why I think that nor what I would actually change. That being said it does resolve the referenced issue, so even if the changes aren't ideal they should provide clues as to what actually needs to happen.

lizmat:
Looked all good to me, thanks!

PR #6115, opened 2026-04-12
Use localizable modifier tokens in terminator rules for L10N support

### LINKED ISSUE #6044: RakuAST: statement modifier doesn't localize when called on a sub without ()
```
$ raku -e 'say Q|volgende als 666|.AST("NL")'
===SORRY!=== Error while compiling EVAL_0
Undeclared routine:
    als used at line 1. Did you mean 'abs', 'all'?
```
It appears that "als" is only localized if it is an expression on which the "if" modifier is applied.  If it is an expression, it works ok:
```
$ raku -e 'say Q|42 als 666|.AST("NL")'
RakuAST::StatementList.new(
  RakuAST::Statement::Expression.new(
    expression         => RakuAST::IntLiteral.new(42),
    condition-modifier => RakuAST::StatementModifier::If.new(
      RakuAST::IntLiteral.new(666)
    )
  )
)
```
Spotted by @habere-et-dispertire

ugexe:
The terminator tokens used hardcoded English strings ('if', 'unless', etc.) instead of the localizable modifier-* tokens. This meant that when an L10N role was mixed into the grammar (e.g. L10N::NL mapping if -> als), the EXPR parser didn't recognize localized keywords as terminators. Listop argument parsing would greedily consume the localized keyword as an argument rather than stopping, causing volgende als 666 to fail with "Undeclared routine: als" while 42 als 666 worked fine.

The statement-mod-cond and statement-mod-loop rules already used the localizable tokens correctly.

Fixes #6044

### DIFF
--- a/src/Raku/Grammar.nqp
+++ b/src/Raku/Grammar.nqp
@@ -5353,15 +5353,15 @@ grammar Raku::Grammar is HLL::Grammar does Raku::Common {
     token terminator:sym<]> { <?[\]]> }
     token terminator:sym<}> { <?[}]> }
     token terminator:sym<ang> { <?[>]> <?{ $*IN_REGEX_ASSERTION }> }
-    token terminator:sym<if>     { 'if'     <.kok> }
-    token terminator:sym<unless> { 'unless' <.kok> }
-    token terminator:sym<while>  { 'while'  <.kok> }
-    token terminator:sym<until>  { 'until'  <.kok> }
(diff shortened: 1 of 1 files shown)


lizmat (approved):

lizmat:
Thanks!

habere-et-dispertire:
_Merci à toi !_
🍓

PR #6116, opened 2026-04-12
Fix spurious sigspace warning for rx:s// in RakuAST frontend

### LINKED ISSUE #6006: RakuAST: throws spurious sigspace error
When we use :s with RakuAST, it's not honoured by the error message.  

```
$ raku -e 'my $re = rx:s/asdf this/'
$ RAKUDO_RAKUAST=1 raku -e 'my $re = rx:s/asdf this/'
Potential difficulties:
    Space is not significant here; please use quotes or :s (:sigspace)
    modifier (or, to suppress this warning, omit the space, or otherwise
    change the spacing).
    at -e:1
    ------> my $re = rx:s/asdf⏏ this/
```

HTH,

ugexe:
The `quote:sym<rx>` and `quote:sym</ />` tokens in the RakuAST grammar were missing the `:my $*WHITESPACE-OK := 0` dynamic variable declaration. When the `:s` adverb was parsed, `adverb-rx2str-control` tried to set `$*WHITESPACE-OK := 1` but the assignment silently failed (wrapped in try) because the variable didn't exist in scope. This caused the `atom` token to always emit the "Space is not significant here" warning, even when `:s` was explicitly specified.

Added the missing declaration to both tokens, matching what all other regex quote forms (`m`, `ms`, `s`, `ss`, `S`, `Ss`) already had.

Fixes #6006

### DIFF
--- a/src/Raku/Grammar.nqp
+++ b/src/Raku/Grammar.nqp
@@ -4657,6 +4657,7 @@ grammar Raku::Grammar is HLL::Grammar does Raku::Common {
     token quote:sym</ /> {
         :my %*RX;
         :my $*INTERPOLATE := 1;
+        :my $*WHITESPACE-OK := 0;
         '/'
         <nibble(self.quote-lang(self.Regex, '/', '/'))>
         [ '/' || <.panic: "Unable to parse regex; couldn't find final '/'"> ]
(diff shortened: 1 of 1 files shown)


lizmat:
Cool find!

PR #6117, opened 2026-04-13
Fix coercion in parameterized Quanthashes

### LINKED ISSUE #6031: Parameterized Quanthashes with a coercing parameterization, don't coerce, they just type check
```raku
my %x is Set[Int()] = "42" => 666; dd %x.keys.head;  # "42", should be 42
my %x is Bag[Int()] = "42" => 666; dd %x.keys.head;  # "42", should be 42
my %x is Mix[Int()] = "42" => 666; dd %x.keys.head;  # "42", should be 42
```

ugexe:
Parameterized Quanthashes with coercing type constraints (e.g., Set[Int()]) only type-checked keys via nqp::istype but never performed the actual coercion, so keys were stored in their original type.

Add COERCE-AND-BIND-TO-TYPED-SET/BAG/MIX methods that type-check, coerce, and bind in one step. Callers cache type.^archetypes.coercive once and dispatch to the coercive or original method accordingly, keeping the non-coercive hot path untouched for JIT/specializer compatibility.

For Bag/Mix, coercion is inlined in ADD-PAIRS and ADD-ITERATOR methods since keys must be coerced before WHICH computation for correct deduplication.

Fixes #6031

### DIFF
--- a/src/core.c/BagHash.rakumod
+++ b/src/core.c/BagHash.rakumod
@@ -64,18 +64,29 @@ my class BagHash does Baggy {
                     ),
                     nqp::if(
                       $value > 0,               # new
-                      Rakudo::QuantHash.BIND-TO-TYPED-BAG(
-                        $!elems, $which, k, nqp::decont($value), type
+                      nqp::if(
+                        type.^archetypes.coercive,
+                        Rakudo::QuantHash.COERCE-AND-BIND-TO-TYPED-BAG(
+                          $!elems, $which, k, nqp::decont($value), type),
+                        Rakudo::QuantHash.BIND-TO-TYPED-BAG(
+                          $!elems, $which, k, nqp::decont($value), type)
                       )
                     )
                   ),
                   nqp::if(                      # no hash allocated yet
                     $value > 0,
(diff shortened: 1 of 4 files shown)


ugexe:
I originally had something much simpler but it had significant performance regressions

lizmat:
This is indeed what I had in mind for the fix.

But looking at it, I wonder whether we shouldn't take a deeper look at it.

For instance, if `Int(Any).^coerce("42")` works, why shouldn't `Int.^coerce("42")` work as well?

For ClassHOW, the `coerce` method could involve an `nqp::istype` and the slow path for actual coercion.

Then instead of `nqp::istype($value,type)` in all of this quanthash logic, could be converted to `type.^coerce($value)`, which would quickly inline in the already correct type case.

Does that make sense?

lizmat:
@ugexe This PR doesn't solve some checks for mutable Quanthashes.

Am taking this as inspiration to further work on fixing #6031 .

lizmat:
Closing this in favour of the more deeply involved solution that may not make this release.

PR #6118, opened 2026-04-14
Fix silent no-op for `$.foo = ...` on non-rw attributes

### LINKED ISSUE #5908: $.foo silently ignores assignments if not specified as "is rw"
```raku
class A {
    has $.b = 42;
    method bb() { $.b = 666; }
}
my $A = A.new;
$A.bb;
dd $A.b;  # 42
```
Spotted by Christiaan Kras in https://fosstodon.org/@Htbaa/114698897312673810

Bisect points to https://github.com/rakudo/rakudo/commit/a6608c2121fccb5a1f33ca46127abbc5ffe83e68

The problem is that since then, `.item` is returning a container.  And since `$.foo` codegens to `self.foo.item`, it provides a container that will silently do the arithmetic action, and then be discarded.  While it should be really producing the same error as `self.foo = 666` would do if the attribute is *not* marked with `is rw`.

Solution should be that `.item` should return a read-only container, if the invocant is not a container yet.

### LINKED ISSUE #6113: Assignment to object attribute silently fails (without `is readonly`)
```
class Foo {has $.a; method test {$.a = 5}}; my $f = Foo.new; $f.test; say $f.a
```

Output on recent releases: (Any)
Expected output: an exception or failure like `Cannot modify an immutable Any ((Any))`

Looks like this changed sometime around 2025.02: https://gist.github.com/Whateverable/aff4b8b5c5faef441ef5f7abb1a91ed8

This erorrs as expected when using `$.a is readonly`.

ugexe:
`$.foo` compiles to `self.foo.item` (and `@.foo`/`%.foo` use `.list`/`.hash`). For non-rw accessors, the accessor returns a bare value, and `Mu.item` wraps it in a fresh mutable Scalar via `(my $ = item)`. Assigning to `$.foo` then silently stores into that throwaway container with no effect on the attribute, instead of throwing `X::Assignment::RO` the way `$obj.foo = ...` does.

This regressed in a6608c212 (2025-02), which switched `Mu.item` from `p6bindattrinvres(create(Scalar), ...)` (no rw descriptor, so STORE threw) to `(my $ = item)` (rw descriptor, so STORE silently succeeds) to unbreak a PDF distribution test. A later attempt to make `.item` return a read-only container (f95f85ac3) was reverted in e0d0d960c because it re-broke the same PDF path.

Fix it at the call site instead of in `Mu.item`: strip the sigil contextualizer (`.item`/`.list`/`.hash`) when the `$.foo` / `@.foo` / `%.foo` term is used as an lvalue, so the assignment lands on the bare `self.foo` method call. Non-rw scalar accessors then error via p6store as expected; `@.foo = 1,2,3` still works through `Array.STORE`; `is rw` accessors are unaffected; and explicit `self.foo.item = 5` written by the user is left alone because only the compiler-inserted wrap is stripped.

Fixes #6113. Also fixes #5908 (same root cause), for which t/spec/S12-methods/accessors.t already has a `#?rakudo todo` test; that test now passes but throws `X::Assignment::RO` (matching external `$obj.a = 5`) rather than the `X::AdHoc` the todo assumed, so the roast expectation needs a follow-up update.

### DIFF
--- a/src/Perl6/Actions.nqp
+++ b/src/Perl6/Actions.nqp
@@ -3127,13 +3127,17 @@ class Perl6::Actions is HLL::Actions does STDActions {
             }
             $past.name($desigilname);
             $past.unshift(QAST::Var.new( :name('self'), :scope('lexical') ));
-            # Contextualize based on sigil.
+            # Contextualize based on sigil. Mark so assignment can strip
+            # this wrapper (see assign_op); otherwise .item would wrap a
+            # non-rw accessor result in a throwaway Scalar, silently
+            # swallowing `$.foo = ...`. See rakudo/rakudo#5908 and #6113.
             $past := QAST::Op.new(
                 :op('callmethod'),
                 :name($sigil eq '@' ?? 'list' !!
                       $sigil eq '%' ?? 'hash' !!
                       'item'),
                 $past);
+            $past.annotate('public-attr-contextualizer', 1);
         }
         elsif $twigil eq '^' || $twigil eq ':' {
             $past := add_placeholder_parameter($/, $sigil, $desigilname,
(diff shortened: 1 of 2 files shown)


lizmat:
Thanks!

PR #6120, opened 2026-04-17
RakuAST: Fix module loading with class shadowing a core type

### LINKED ISSUE #6051: RakuAST: module loading fails with a class with the same name as a core class
```
% cat lib/Date.rakumod 
class Date { }

% RAKUDO_RAKUAST=1 raku -Ilib -MDate -e ''
===SORRY!=== Error while compiling 
Redeclaration of symbol 'Date'.

% raku -Ilib -MDate -e ''
```
Looks to me that check is a bit overeager.

ugexe:
When importing a module that defines a class with the same name as a core class (e.g. `class Date { }`), the RakuAST frontend erroneously threw `X::Redeclaration`.

`IMPL-IMPORT-ONE` used `$resolver.resolve-lexical-constant` which searches all scopes including the SETTING. Finding the core `Date` there, it called `merge()` which has no case for a non-stub import replacing a non-stub outer-scope class.

The old frontend only checks `$target.symbol($key)` (the current lexpad), so it never sees the core symbol and just installs the import, correctly shadowing it.

The fix adds `:current-scope-only` to the `resolve-lexical-constant` call in `IMPL-IMPORT-ONE`.

Fixes #6051

### DIFF
--- a/src/Raku/ast/statements.rakumod
+++ b/src/Raku/ast/statements.rakumod
@@ -1798,7 +1798,7 @@ class RakuAST::ModuleLoading {
 
             my $existing := nqp::isconcrete($declarand.compile-time-value)
                 ?? Nil
-                !! $resolver.resolve-lexical-constant($key);
+                !! $resolver.resolve-lexical-constant($key, :current-scope-only);
             if $existing {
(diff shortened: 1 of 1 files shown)


ab5tract:
This one looks great to me

lizmat:
Good catch!

PR #6121, opened 2026-04-17
RakuAST: Implement HyperWhatever currying

ugexe:


ab5tract:
This looks good to me. Thanks @ugexe 🌊

Not for this PR, but this reminds me that we should align the RakuAST class names with the shift from `Curry` to `Prime` in documentation and discourse.

PR #6122, opened 2026-04-17
Fix nested multi-part package name collision with enclosing package

ugexe:


ugexe:

ab5tract:
Another excellent patch that covers a longstanding pain point.

PR #6124, opened 2026-04-17
RakuAST: Fix build-exception crash on uncomposed exception type stubs

ugexe:
During CORE setting compilation (or when user code forward-declares an exception class), build-exception could resolve a stubbed exception type like X::Redeclaration that wasn't composed yet. Attempting to mixin X::Comp into the uncomposed stub would die with "cannot inherit from X::Redeclaration because the parent is not composed yet".

Guard the mixin and constructor with is_composed and repr_composed checks so uncomposed stubs fall through to the existing BOOTException fallback, which was already designed for this scenario during CORE setting compilation.


ugexe:

PR #6125, opened 2026-04-18
RakuAST: avoid pathologically slow Raku dispatch in add-phasers-handling-code

ugexe:
During CORE setting compilation, `add-phasers-handling-code` runs once per Code object being compiled. Two call sites hit Raku level dispatch on Sub meta-objects that carry a trait mixin (e.g. `Sub+{is-implementation-detail}`, `Sub+{is-DEPRECATED}`). Dispatching `.Bool` or a regular method on such a mixed-in type is hundreds of times slower than on a plain Sub, and multiplied over the tens of thousands of `add-phasers-handling-code` invocations in CORE it causes Stage qast to effectively hang.

Replace the two slow sites with nqp primitives:

  1. The entry booleanization `$block ?? ... !! NQPMu` becomes `nqp::isconcrete($block) ?? ... !! NQPMu`.

  2. The ENTER-branch `$block.phasers('ENTER')` method call (and the `.FLATTENABLE_LIST` iteration) becomes a direct `nqp::atkey` lookup plus `nqp::elems`/`nqp::atpos` loop, guarded by an `nqp::ishash` + `nqp::existskey pre-check`.

Both changes must be present together. The ENTER-branch rewrite carries the majority of the benefit by unblocking the first hang point (sub with ENTER phaser, typically installed via the DEPRECATED trait). The entry rewrite by itself has no measurable effect, but once the ENTER hang is unblocked, the entry rewrite is what allows the remaining calls to finish. Removing either causes Stage qast to fail to complete.

Other similar Raku-level boolean checks in this method (`$!PRE`, `$!FIRST`, `$!let`, `$!temp`, etc.) were investigated and shown not to be a bottleneck during CORE compilation and are left alone.

### DIFF
--- a/src/Raku/ast/code.rakumod
+++ b/src/Raku/ast/code.rakumod
@@ -1051,7 +1051,7 @@ class RakuAST::ScopePhaser {
 
     method add-phasers-handling-code(RakuAST::IMPL::Context $context, Mu $qast) {
         my $block := nqp::istype(self, RakuAST::Code) ?? self.meta-object !! NQPMu;
-        my $phasers := $block ?? nqp::getattr($block, Block, '$!phasers') !! NQPMu;
+        my $phasers := nqp::isconcrete($block) ?? nqp::getattr($block, Block, '$!phasers') !! NQPMu;
 
         if $!has-exit-handler || self.needs-result > 1 || $phasers && (nqp::istype($phasers, Code) || nqp::existskey($phasers, 'LEAVE') || nqp::existskey($phasers, 'POST')) {
             $qast.has_exit_handler(1);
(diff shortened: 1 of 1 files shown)


ugexe:
This is the result of 10+ hours of brute forcing my way through the qast stage of `RAKUDO_RAKUAST=1 make install` so of all of my recent RakuAST PRs, this is the one I'm most unsure about. Without it `add-phasers-handling-code` gets called ~50/second, and with it `add-phasers-handling-code` gets called about ~5000/second. This meant `RAKUDO_RAKUAST=1 make install` would have taken hours to finish stage qast.

ugexe:

ab5tract:
Thanks for taking a look at this -- and doubly so for fixing it!

PR #6126, opened 2026-04-18
RakuAST: dedupe IMPL-QAST-DECLS by lexical name, not just objectid

ugexe:
Some CompUnit-level implicit declarations (`$!pod`, `$!data`, `$!finish`, `$!rakudoc`) are reachable from the ast-lexical-declarations DFS walk via two different paths: once when visit-children visits them as direct attributes of CompUnit, and once again when the DFS visits CompUnit itself and folds in its get-implicit-declarations results (via the ImplicitDeclarations branch of the DFS).

This produces a `@declarations` list containing the same underlying `RakuAST::VarDeclaration::Implicit::Doc::Pod` object twice as two separate entries. `IMPL-QAST-DECLS` only dedupes by `nqp::objectid` on its per-invocation `%seen-decl` set, which doesn't catch repeated entries in a single input list. Result: two
`QAST::Var :decl<static> :name<$=pod>` declarations in the same block, tripping "Lexical '$=pod' already declared" in the MAST compiler.

Add a per-invocation `%seen-name` set alongside `%seen-decl` and skip any declaration whose lexical-name has already been emitted in this scope.

### DIFF
--- a/src/Raku/ast/scoping.rakumod
+++ b/src/Raku/ast/scoping.rakumod
@@ -43,15 +43,29 @@ class RakuAST::LexicalScope
     method IMPL-QAST-DECLS(RakuAST::IMPL::QASTContext $context) {
         my $stmts := QAST::Stmts.new();
 
-        # Visit declarations and produce declaration QAST.
+        # Visit declarations and produce declaration QAST. Track both by
+        # objectid (to avoid emitting the same declaration node twice) and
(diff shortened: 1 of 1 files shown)


ugexe:

ugexe:
I'm going to dedupe these inside of `ast-lexical-declarations()` instead which would be cleaner than this

ugexe:
Superseding this with https://github.com/rakudo/rakudo/pull/6154

Issue #5854, opened 2025-04-17
Confusion when declaring nested multi-part-name packages with similar names to the outer package


niner:
`module Foo::Bar { class Foo::Bar { } }` dies with `Redeclaration of symbol 'Foo::Bar'` when RakuAST is used. In this example the class' full name would be `Foo::Bar::Foo::Bar`. However when we try to install it into the current package, we start with the local name `Foo::Bar`. We try to find the first part (Foo) and resolve it to `::Foo`, i.e. the package that contains the module `Foo::Bar`. We then try to find a Bar there and find the module which triggers the Redeclaration error.

Note that the old compiler frontend is broken in almost the same way! The only difference is that instead of throwing the Redeclaration exception, it silently replaces the module with the class:
```
> rakudo -e 'module MoarVM::Profiler { my class MoarVM::Profiler { } }; MoarVM::Profiler.HOW.^name.say'
Perl6::Metamodel::ClassHOW
```

The fix needs to be careful as to not declare a lexically scoped `X` package in e.g. `unit module Test; my class X::SubtestsSkipped is Exception {}`

ab5tract:
So I'm a bit confused as to whether throwing a redeclaration exception is actually the wrong thing to do here.

In other words, what is the desired behavior that would constitute a fix here?

ab5tract:
@MasterDuke17 I self-assigned this one out of pure momentum. However I haven't had a deep look at your branch yet, so please feel free to re-assign back to yourself if you've got the tuits.

PR #6129, opened 2026-04-21
Gate nested-package install on 6.e and deprecate silent replacement of enclosing module

ugexe:
PR #6122 (for issue #5854) fixed a longstanding quirk where declaring a class inside a module with the same name, e.g. `module Foo::Bar { class Foo::Bar { } }`, silently replaced the module with the class in the outer stash. The fix was the correct behavior per the language spec, but it regressed ecosystem modules.

This PR keeps #6122's fix but puts it behind a `use v6.e.PREVIEW` gate. On 6.d and 6.c the traditional grammar continues to silently replace as it always has, and RakuAST now emulates the same behavior rather than throwing X::Redeclaration. On 6.e the new nested semantics apply to both frontends. That gives the ecosystem a migration window for fixes.

Pre-6.e code that hits the legacy silent-replace path additionally emits a compile time worry that suggests how the issue can be fixed.

### 6.d (class silently replaces module, worry fires)

```console
$ raku -e 'use v6.d; module Foo::Bar { class Foo::Bar { method x { 42 } } }; say Foo::Bar.HOW.^name; say Foo::Bar.x'
Potential difficulties:
    Declaring class 'Foo::Bar' inside an enclosing module of the same name
    silently replaces the module in the outer stash. This is legacy behavior
    specific to Raku 6.d and earlier; in Raku 6.e the same pattern installs
    the class as a nested package instead. Rewrite as 'unit class Foo::Bar;'
    in its own file (or otherwise avoid the module+same-named-class
    collision) to work the same way on either revision.
    at -e:1
    ------> use v6.d; module Foo::Bar { class <HERE>Foo::Bar { method x { 42 } } }; say Foo:
Perl6::Metamodel::ClassHOW
42

$ RAKUDO_RAKUAST=1 raku -e 'use v6.d; module Foo::Bar { class Foo::Bar { method x { 42 } } }; say Foo::Bar.HOW.^name; say Foo::Bar.x'
Potential difficulties:
    Declaring class 'Foo::Bar' inside an enclosing module of the same name
    silently replaces the module in the outer stash. This is legacy behavior
    specific to Raku 6.d and earlier; in Raku 6.e the same pattern installs
    the class as a nested package instead. Rewrite as 'unit class Foo::Bar;'
    in its own file (or otherwise avoid the module+same-named-class
    collision) to work the same way on either revision.
    at line
Perl6::Metamodel::ClassHOW
42
```

### 6.e (class installs nested, outer module preserved, no warning)

```console
$ raku -e 'use v6.e.PREVIEW; module Foo::Bar { class Foo::Bar { method x { 42 } } }; say Foo::Bar.HOW.^name; say Foo::Bar::Foo::Bar.x'
Perl6::Metamodel::ModuleHOW
42

$ RAKUDO_RAKUAST=1 raku -e 'use v6.e.PREVIEW; module Foo::Bar { class Foo::Bar { method x { 42 } } }; say Foo::Bar.HOW.^name; say Foo::Bar::Foo::Bar.x'
Perl6::Metamodel::ModuleHOW
42
```

### DIFF
--- a/src/Perl6/World.nqp
+++ b/src/Perl6/World.nqp
@@ -1702,7 +1702,23 @@ class Perl6::World is HLL::World {
         }
         if $create_scope eq 'our' {
             if nqp::existskey($cur_pkg.WHO, $name) {
-                self.steal_WHO($symbol, ($cur_pkg.WHO){$name});
+                my $existing := ($cur_pkg.WHO){$name};
+                # Silently replacing an enclosing module with a class,
+                # role, grammar, etc. is legacy behavior specific to
+                # Raku 6.d and earlier. Under 6.e the same pattern
+                # installs the declaration as a nested package instead,
+                # so warn pre-6.e code to migrate before authors bump
+                # their language version.
+                if nqp::getcomp('Raku').language_revision < 3
+                  && $existing.HOW.HOW.name($existing.HOW) eq 'Perl6::Metamodel::ModuleHOW'
+                  && $symbol.HOW.HOW.name($symbol.HOW) ne 'Perl6::Metamodel::ModuleHOW' {
+                    $/.PRECURSOR.typed_worry(
+                        'X::Package::SameNameAsEnclosingModule',
+                        :kind($pkgdecl),
+                        :name($existing.HOW.name($existing)),
+                    );
+                }
+                self.steal_WHO($symbol, $existing);
             }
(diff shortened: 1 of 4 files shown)


lizmat (approved):

lizmat:
Nice adaptation!

PR #6130, opened 2026-04-21
Extend same-name-as-enclosing deprecation to cover package outers

ugexe:


ugexe:
This came about due to no deprecation messaging being emitted for raku-community-modules/Testo

ab5tract:
I really like the changes to `PackageInstaller` 👍🏻

PR #6131, opened 2026-04-21
MoarVM::Profiler: use unit class to silence deprecation worry

ugexe:
`MoarVM::Profiler.rakumod` was declared as `unit module MoarVM::Profiler` with a nested `class MoarVM::Profiler { ... }` of the same name, which triggers a deprecation worry on 6.d. Collapse the wrapper into `unit class MoarVM::Profiler;` with the attribute declarations at file scope. The nested helper classes and the exported subs keep their public addresses and behavior, so external consumers are unaffected, and the file now works the same way under 6.e where the old pattern would install the class as a nested package.

### DIFF
--- a/lib/MoarVM/Profiler.rakumod
+++ b/lib/MoarVM/Profiler.rakumod
@@ -18,7 +18,7 @@
 
 # Make all class references a lot shorter
 
-unit module MoarVM::Profiler;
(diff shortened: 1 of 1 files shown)


ab5tract:
Looks fully sensible to me

PR #6132, opened 2026-04-22
Fix "Missing infix inside R" when using RZ/RX infix form

### LINKED ISSUE #6128: R and implicit comma
`R` seems to happily borrow an implicit comma from `Z` when distributed:

~~~~raku
say [RZ] (1,2),(3,4) 		
# ((3 1) (4 2))
~~~~

But then gets shy borrowing it in infix form:

~~~~raku
say (1,2) RZ (3,4)
# ===SORRY!=== Error while compiling:
# Missing infix inside R
# ------> say (1,2) RZ⏏ (3,4)
#     expecting any of:
#         infix
#         infix stopper
~~~~

See: raku-beginner logs for [2026-04-19](https://irclogs.raku.org/raku-beginner/2026-04-19.html)

🍓

ugexe:
`infix:sym<Z>` and `infix:sym<X>` carried a negative lookahead `<!before <.sym> <.infixish>>` added in 2010 (commit 1120182ed) so that `Z-` style metaops would still parse. Back then `infixish` tried `<infix>` before `<infix_prefix_meta_operator>`, so without the lookahead `infix:sym<Z>` would greedily consume the `Z` and leave the `-` dangling. The lookahead forced fallthrough to the metaop branch.

The ordering in `infixish` was later flipped so `<infix_prefix_meta_operator>` is now tried first. That makes the lookahead redundant: `Z-`, `Z+`, `X*`, etc. are matched by the metaop branch before the plain `infix:sym<Z>` is ever considered.

The lookahead also caused the bug reported in #6128. When parsing `(1,2) RZ (3,4)`, the enclosing R metaop set `$*IN_META := 'R'` and recursed into `<infixish('R')>`. Inside, the plain-infix branch ran `infix:sym<Z>` whose lookahead called `<.infixish>` without arguments, inheriting `$*IN_META='R'` via `nqp::getlexdyn`. That recursive infixish found nothing parseable at `(3,4)` and tripped the `.missing` panic gated on `$*IN_META ~~ 'R'`, which propagated out of the lookahead as a hard compile error. `[RZ] ...` worked only because `$*IN_REDUCE` suppresses that same panic.

Fix: drop the obsolete lookahead. Metaop disambiguation is already handled by the alternation order in `infixish`.

Fixes #6128

### DIFF
--- a/src/Perl6/Grammar.nqp
+++ b/src/Perl6/Grammar.nqp
@@ -4365,8 +4365,8 @@ grammar Perl6::Grammar is HLL::Grammar does STD {
         { $*INVOCANT_OK := 0; }
     }
 
-    token infix:sym<Z>    { <!before <.sym> <.infixish> > <sym>  <O(|%list_infix)> }
-    token infix:sym<X>    { <!before <.sym> <.infixish> > <sym>  <O(|%list_infix)> }
+    token infix:sym<Z>    { <sym>  <O(|%list_infix)> }
+    token infix:sym<X>    { <sym>  <O(|%list_infix)> }
 
     token infix:sym<...>  { <sym> <O(|%list_infix)> }
     token infix:sym<…>    { <sym> <O(|%list_infix)> }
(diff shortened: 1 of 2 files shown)


lizmat:
Great find!

But would like to lift this over the 2026.04 release, just to make sure this doesn't break anything in the ecosystem.

PR #6135, opened 2026-04-23
Fix RakuAST .EVAL at BEGIN time losing access to setting symbols

### LINKED ISSUE #6134: RakuAST: not able to run an EVAL at compile time
```raku
BEGIN {
    RakuAST::Sub.new(
      name      => RakuAST::Name.from-identifier("foo"),
      signature => RakuAST::Signature.new(),
      body      => RakuAST::Blockoid.new(
        RakuAST::StatementList.new()
      )
    ).EVAL
}
```
Running the above code with RAKUDO_RAKUAST=1 results in the following error:
```
===SORRY!=== Error while compiling foo
An exception X::Comp::AdHoc occurred while evaluating a BEGIN:
Could not find a compile-time-value for lexical Nil
at foo:1
Exception details:
  ===SORRY!=== Error while compiling:
  Could not find a compile-time-value for lexical Nil
  at line 
```
Fixing this error would be a great step towards being able to generate synthetic code in modules and in the setting (think `.assuming` and `printf`).

ugexe:
Calling .EVAL on a synthetic RakuAST node inside a BEGIN block failed
with "Could not find a compile-time-value for lexical Nil" because the
EVAL resolver couldn't reach CORE. At BEGIN time the caller's lexical
context chain terminates at the partially-built CompUnit mainline and
never links to the setting, so IMPL-SETTING-FROM-CONTEXT returns Nil
and the outer-context walk in the resolver never finds built-ins like
Nil, Mu, or &say. Two symptom shapes surface depending on where the
lookup happens:

  * PERFORM-BEGIN on RakuAST::Call::Name calls resolve-lexical('&say')
    and gets Nil, leading to X::Undeclared::Symbols at check time.

  * IMPL-FIXUP-DYNAMICALLY-COMPILED-BLOCK (which statically bakes
    outer-lexical references into dynamically-compiled Code bodies)
    calls resolve-lexical-constant('&say') / ('Nil') and gets Nil,
    throwing "Could not find a compile-time-value for lexical X" or
    "Undeclared routine: &X used at line -1" for synthetic nodes.

Thread the currently-compiling CompUnit's setting through to the EVAL
resolver and give the resolver a setting-aware fallback on both lookup
paths:

1. New narrow CompUnit.setting accessor exposing $!resolver.$!setting
   (the BOOTContext of the outer compile's setting). Guards with
   nqp::isconcrete and uses nqp::getattr directly because the Raku $.
   accessor silently masks VMNull (which $!setting can be while
   compiling CORE.setting itself).

2. RakuAST::Resolver::EVAL.new accepts :$setting and stores it as
   `$!setting`, overriding IMPL-SETTING-FROM-CONTEXT's walk-the-$!outer
   result.

3. ForeignCode's RakuAST EVAL path threads $*CU.setting into the new
   resolver when $*CU is a RakuAST::CompUnit; otherwise (runtime EVAL,
   $*CU undefined) it passes Mu and the resolver falls back to the
   existing IMPL-SETTING-FROM-CONTEXT path, preserving prior behavior.

4. resolve-lexical and resolve-lexical-constant on Resolver::EVAL both
   get a setting fallback: after $!outer fails to find the name, walk
   $!setting directly and return an External::Setting declaration
   (shared helper, matching the External::Setting already returned by
   resolve-lexical-in-outer on setting hits). Both methods need it
   because PERFORM-BEGIN uses resolve-lexical while
   IMPL-FIXUP-DYNAMICALLY-COMPILED-BLOCK (called per-Code-node by
   IMPL-COMPILE-DYNAMICALLY, and per-compunit by IMPL-TO-QAST-COMP-UNIT)
   uses resolve-lexical-constant.

Runtime .EVAL is unchanged: $*CU is undefined at runtime, so the new
setting-threading branch is skipped.

Fixes https://github.com/rakudo/rakudo/issues/6134

### DIFF
--- a/src/Raku/ast/compunit.rakumod
+++ b/src/Raku/ast/compunit.rakumod
@@ -256,6 +256,13 @@ class RakuAST::CompUnit
     # does not declare its own GLOBAL and so forth.
     method is-eval() { $!is-eval ?? True !! False }
 
+    # The setting context the CompUnit's resolver knows about, if any.
+    method setting() {
+        nqp::isconcrete($!resolver)
+          ?? nqp::getattr($!resolver, RakuAST::Resolver, '$!setting')
+          !! Mu
+    }
+
     method attach-target-names() {
         self.IMPL-WRAP-LIST(['compunit'])
     }
--- a/src/Raku/ast/resolver.rakumod
+++ b/src/Raku/ast/resolver.rakumod
@@ -819,11 +819,13 @@ class RakuAST::Resolver::EVAL
     # The stack of scopes we are in (an array of RakuAST::LexicalScope).
     has Mu $!scopes;
 
-    method new(Mu :$global!, Mu :$context!) {
+    method new(Mu :$global!, Mu :$context!, Mu :$setting) {
         my $obj := nqp::create(self);
         nqp::bindattr($obj, RakuAST::Resolver, '$!outer', $context);
         nqp::bindattr($obj, RakuAST::Resolver, '$!setting',
-            self.IMPL-SETTING-FROM-CONTEXT($context));
+            nqp::defined($setting)
+              ?? $setting
+              !! self.IMPL-SETTING-FROM-CONTEXT($context));
         nqp::bindattr($obj, RakuAST::Resolver, '$!global', $global);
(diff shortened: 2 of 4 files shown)


lizmat:
These fixes solve the issue with `say`, so Yay!

PR #6139, opened 2026-04-24
RakuAST: fix 'is T' on generic role attributes

ugexe:
Makes `t/spec/S02-types/is-type.t` pass with `RAKUDO_RAKUAST=1`


ab5tract:
Awesome work! I think this has been a wart for a very long time. Thank you @ugexe for fixing this!

PR #6143, opened 2026-04-24
RakuAST: drop dead compile-time-value read in package-lookup parse

ugexe:
Makes `t/spec/S12-enums/misc.t`  pass with `RAKUDO_RAKUAST=1`


ab5tract:
This looks good to me. Nice catch!

PR #6144, opened 2026-04-24
RakuAST: nominalize sigilless variable invocant for .= initializers

ugexe:
Needed to make `t/spec/MISC/misc.t` pass with `RAKUDO_RAKUAST=1`


ab5tract:
Another great patch

PR #6147, opened 2026-04-24
RakuAST: report stubbed grammars, drop misleading is_composed gate

ugexe:
Makes `t/spec/S32-exceptions/misc2.t` pass with `RAKUDO_RAKUAST=1`


lizmat:
Apart from the fact that it fixes the test, there's not a lot of information about the reasoning of this change?

ugexe:
Are you not reading the commit messages?

lizmat:
Aah.  Indeed I wasn't.

PR #6148, opened 2026-04-25
RakuAST: fix generic Definedness/Coercion in parametric role bodies

ugexe:
Makes tests 1-2 of the (Nominalizable generic subtest) in `t/spec/S02-types/generics.t` pass with `RAKUDO_RAKUAST=1`


ab5tract:
@ugexe this solution is incredible to see. I was banging my head on this pretty hard!

In other words, LGTM 👍🏻

PR #6153, opened 2026-05-02
RakuAST: emit BOOTSTRAP load as pre-deserialize task for CORE.<spec>

ugexe:
Fix RAKUAST built `CORE.<spec>.setting.moarvm` so it loads `Perl6::BOOTSTRAP::v6<spec>` before its own SC deserialization, mirroring what the legacy frontend already emits via `World.load_module_early`.

Before fix:
```
$ RAKUDO_RAKUAST=1 make install   # succeeds

$ RAKUDO_RAKUAST=1 ./install/bin/raku -e 'say 42'
42

$ ./install/bin/raku -e 'say 42'
Missing or wrong version of dependency 'gen/moar/BOOTSTRAP/v6c.nqp'
  (from 'gen/moar/CORE.c.setting')
```

After fix:
```
$ RAKUDO_RAKUAST=1 make install   # succeeds

$ RAKUDO_RAKUAST=1 ./install/bin/raku -e 'say 42'
42

$ ./install/bin/raku -e 'say 42'
42
```

### DIFF
--- a/src/Raku/ast/compunit.rakumod
+++ b/src/Raku/ast/compunit.rakumod
@@ -568,10 +568,12 @@ class RakuAST::CompUnit
             )),
     }
 
-    method IMPL-SETTING-LOADING-QAST(Mu $top-level, Str $name) {
-        # Use the NQP module loader to load Perl6::ModuleLoader, which
-        # is a normal NQP module.
-        my $module-loading := QAST::Stmt.new(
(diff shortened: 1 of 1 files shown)


ugexe:
Note that #6154 is required for the `RAKUDO_RAKUAST=1 make install` command referenced in the reproduction to work

PR #6154, opened 2026-05-02
RakuAST: dedupe @declarations at the source in ast-lexical-declarations

ugexe:
Some CompUnit-level implicit declarations ($!pod, $!data, $!finish, $!rakudoc) are reachable from the ast-lexical-declarations DFS walk via two intentional paths: once when visit-children visits them as direct attributes of CompUnit (so IMPL-BEGIN/IMPL-CHECK reach them to drive PERFORM-BEGIN/PERFORM-CHECK), and once when the DFS visits CompUnit itself and folds in its get-implicit-declarations results (so the resolver registers them as live lexicals via create-implicits).

Both paths are required and cannot be collapsed to one. Removing either side silently breaks a real phase. The bug is that ast-lexical-declarations() was pushing the same underlying RakuAST::Declaration object onto @declarations from both arms of its DFS visitor without deduping, producing duplicate entries in the cached @declarations list. Downstream that surfaces in IMPL-QAST-DECLS as two QAST::Var :decl<static> :name<$=pod> declarations in the same block, tripping "Lexical '$=pod' already declared" in the MAST compiler.

Fix it where the duplicate is introduced: track an objectid-keyed %seen-decl set as @declarations is built, and skip subsequent pushes of an already-seen node. Single source of truth, single dedup key, benefits every consumer of @declarations (IMPL-QAST-DECLS, the generated-lexical clone in lexical-declarations(), and the iterations elsewhere in scoping.rakumod) without per-consumer dedup.

This bug was pre-existing but only surfaced once the Stage qast hangs were fixed and compilation actually reached Stage mast.

### DIFF
--- a/src/Raku/ast/scoping.rakumod
+++ b/src/Raku/ast/scoping.rakumod
@@ -106,13 +106,17 @@ class RakuAST::LexicalScope
     method ast-lexical-declarations() {
         unless nqp::isconcrete($!declarations-cache) {
             my @declarations;
+            my %declarations-seen;
             my @variables;
             my %variables-seen;
             my @not-if-duplicate;
             self.visit-dfs: -> $node {
                 if nqp::istype($node, RakuAST::Declaration) && $node.is-simple-lexical-declaration {
-                    nqp::push(@declarations, $node);
-                    nqp::push(@variables, $node) unless nqp::istype($node, RakuAST::Routine);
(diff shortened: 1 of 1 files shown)


ugexe:
This is the last change required for `RAKUDO_RAKUAST=1 make install` to finish successfully

This supersedes https://github.com/rakudo/rakudo/pull/6126

PR #6161, opened 2026-05-04
Revert "RakuAST: make sure non-blocks are sunk in start"

### LINKED ISSUE #6142: RakuAST: start EXPR discards the expression value
Commit d76960baa (Mar 2026) added `$blorst.apply-sink(True)` in `StatementPrefix::Start.new` for non-block blorsts. That fix addressed issue #5976 (`start loop { ... }` doesn't run) but introduced a regression: `start EXPR` now discards EXPR's value, so `Promise.result` is always `Nil` whenever `start` is followed by a parenthesized expression instead of a block.

This was noticed from a failure running `t/spec/S15-nfg/many-threads.t` and `t/spec/S17-promise/start.t` with `RAKUDO_RAKUAST=1`

ugexe:
This breaks at least three spectests with RakuAST.

Resolves #6142

### DIFF
--- a/src/Raku/ast/statementprefixes.rakumod
+++ b/src/Raku/ast/statementprefixes.rakumod
@@ -446,15 +446,6 @@ class RakuAST::StatementPrefix::Start
 {
     method type() { "start" }
 
-    method new(RakuAST::Blorst $blorst) {
-        $blorst.apply-sink(True)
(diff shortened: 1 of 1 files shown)


ugexe:
@lizmat thoughts? I imagine fixing the sinking WAT was a small win, but given the fallout I'm not sure we should continue making releases with this specific solution.

PR #6163, opened 2026-05-06
Make CORE/SETTING symbol tests pass on RakuAST built rakudo

ugexe:
The seven `t/02-rakudo/03-corekeys*.t` and `04-settingkeys*.t` tests pin the symbols visible at CORE:: and SETTING:: scope per language revision. On a `RAKUDO_RAKUAST=1` built rakudo they diverged from the legacy frontend in a few places, so this PR fixes the underlying RakuAST mismatch that's straightforward to address, and updates the tests for the rest.


ugexe:
With this `RAKUDO_RAKUAST=1 make test` should pass with a `RAKUDO_RAKUAST=1` built rakudo.

Issue #6137, opened 2026-04-23
RakuAST: precomping synthetic subs fails in bytecode generation


lizmat:
This is basically the precomp version of https://github.com/rakudo/rakudo/issues/6134 with https://github.com/rakudo/rakudo/pull/6135 applied:
```raku
my &foo = BEGIN {
  RakuAST::Sub.new(
    name => RakuAST::Name.from-identifier("foo"),
    body => RakuAST::Blockoid.new(
      RakuAST::StatementList.new(
        RakuAST::Statement::Expression.new(
          expression => RakuAST::Call::Name::WithoutParentheses.new(
            name => RakuAST::Name.from-identifier("say"),
            args => RakuAST::ArgList.new(
              RakuAST::QuotedString.new(
                segments   => (
                  RakuAST::StrLiteral.new("foo"),
                )
              )
            )
          )
        )
      )
    )
  ).EVAL
}

foo()
```
Put this code in a `lib/Foo.rakumod`, and then try to load it with:
```
$ RAKUDO_RAKUAST=1 raku --ll-exception -Ilib -MFoo -e ''
Serialization Error: missing static code ref for closure '' (src/Raku/ast/code.rakumod:145)
   at NQP::src/vm/moar/QAST/QASTCompilerMAST.nqp:880  (/Users/liz/Github/rakudo/install/share/nqp/lib/QAST.moarvm:serialize_sc)
 from NQP::src/vm/moar/QAST/QASTCompilerMAST.nqp:885  (/Users/liz/Github/rakudo/install/share/nqp/lib/QAST.moarvm:deserialization_code)
 from NQP::src/vm/moar/QAST/QASTCompilerMAST.nqp:826  (/Users/liz/Github/rakudo/install/share/nqp/lib/QAST.moarvm:compile_node)
 from NQP::src/vm/moar/QAST/QASTCompilerMAST.nqp:664  (/Users/liz/Github/rakudo/install/share/nqp/lib/QAST.moarvm:as_mast)
 from NQP::src/vm/moar/QAST/QASTCompilerMAST.nqp:384  (/Users/liz/Github/rakudo/install/share/nqp/lib/QAST.moarvm:to_mast)
 from NQP::src/vm/moar/QAST/QASTCompilerMAST.nqp:2180  (/Users/liz/Github/rakudo/install/share/nqp/lib/QAST.moarvm:to_mast)
[...]

PR #6170, opened 2026-05-07
Fix .assuming causing serialization error during precomp

### LINKED ISSUE #6169: Using a module that has an .assuming in it causes serialization error
Similar to https://github.com/rakudo/rakudo/issues/5819, but with a twist.  To reproduce:

lib/B.rakumod
```raku
my &max42 = &max.assuming(42);
```

lib/A.rakumod
```raku
use B;
```
then do:
```
$ raku -Ilib -MA -e ''
===SORRY!===
Serialization Error: missing static code ref for closure 'assumed.max' (:1)
```
Doing a direct load of module B is fine:
```
$ raku -Ilib -MB -e ''
```

ugexe:
`.assuming` constructs a RakuAST::Sub via `RakuAST::Sub.new(...).EVAL`. When that synthetic Sub gets bound somewhere a later precomp's serializer can reach (most directly: `my &x = &foo.assuming(...)` in a module that's `use`d by another module being precompiled), the precomp serializer dies with "missing static code ref for closure 'assumed.X'".

The nested branch of ForeignCode.rakumod's EVAL proto already handled this for the BEGIN-time AST EVAL case (#6137, fix in commit 46ce0e3f8). The non-nested branch had identical bookkeeping needs but was never updated, which is what kept `.assuming` tripping the same class of serialization error.

Resolves #6169.

### DIFF
--- a/src/Raku/ast/code.rakumod
+++ b/src/Raku/ast/code.rakumod
@@ -194,6 +194,13 @@ class RakuAST::Code
         $block.code_object($code-obj);
 
         my @compstuff := nqp::getattr($code-obj, Code, '@!compstuff');
+        # @!compstuff is null on a re-compile of a shared AST: the
+        # previous compile's cleanup nulled it and $!begin-performed
+        # keeps IMPL-STUB-CODE from re-running.
+        if nqp::isnull(@compstuff) {
+            @compstuff := nqp::list();
+            nqp::bindattr($code-obj, Code, '@!compstuff', @compstuff);
+        }
         my $cuid := $!cuid;
         $block.set-cuid($!cuid);
 
--- a/src/core.c/ForeignCode.rakumod
+++ b/src/core.c/ForeignCode.rakumod
@@ -103,28 +103,11 @@ $lang = 'Raku' if $lang eq 'perl6';
         my $from := $compiler.exists_stage('optimize') ?? 'optimize' !! 'qast';
         my $qast-cu := $comp-unit.IMPL-TO-QAST-COMP-UNIT;
         my $context := $comp-unit.context;
-        if $context.is-nested {
-            # Nested QAST::CompUnits skip deserialization_code emission, so
-            # their static code refs never get scsetcode'd into the shared
-            # SC. If EVAL returns a Sub the outer compile references,
-            # outer's precomp serialization fails with "missing static
-            # code ref" because the inner code ref's static_code isn't in
-            # any SC. Keep the compunit around via :compunit_ok and run
-            # the shared fixup loop to register each code ref in the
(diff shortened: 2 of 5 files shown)


lizmat (approved):

lizmat:
Confirm this fixes the golfed code case.

Unfortunately, it does **not** fix the issue for `Astro::Utils`:
```
% zef install .         
===> Staging Astro::Utils:ver<0.0.3>:auth<zef:tbrowder>
Potential difficulties:
    Declaring class 'Math::FractionalPart' inside an enclosing module of
    the same name silently replaces the module in the outer stash. This is
    legacy behavior specific to Raku 6.d and earlier; in Raku 6.e the same
    pattern installs the class as a nested package instead. Rewrite as 'unit
    class Math::FractionalPart;' in its own file (or otherwise avoid the
    module+same-named-class collision) to work the same way on either
    revision.
    at /Users/liz/Github/rakudo/install/share/perl6/site/sources/6CDFA487140E7E3D9B786E3B0035B36637E1F502 (Math::FractionalPart):3
    ------> class <HERE>Math::FractionalPart is Num is export {
===> Staging [OK] for Astro::Utils:ver<0.0.3>:auth<zef:tbrowder>
===> Testing: Astro::Utils:ver<0.0.3>:auth<zef:tbrowder>
[Astro::Utils] ===SORRY!=== Error while compiling /Users/liz/Github/tbrowder/Astro-Utils/t/1-montenbruck.rakutest
[Astro::Utils] Missing or wrong version of dependency 'site#sources/91F03A688EB6348D05C4EE01F0E8A12764FA8202 (Astro::Utils)' (from 'site#sources/91F03A688EB6348D05C4EE01F0E8A12764FA8202 (Astro::Utils)')
[Astro::Utils] at /Users/liz/Github/tbrowder/Astro-Utils/t/1-montenbruck.rakutest:2
[Astro::Utils] ===SORRY!=== Error while compiling /Users/liz/Github/tbrowder/Astro-Utils/t/2-nasa.rakutest
[Astro::Utils] Missing or wrong version of dependency 'site#sources/91F03A688EB6348D05C4EE01F0E8A12764FA8202 (Astro::Utils)' (from 'site#sources/91F03A688EB6348D05C4EE01F0E8A12764FA8202 (Astro::Utils)')
[Astro::Utils] at /Users/liz/Github/tbrowder/Astro-Utils/t/2-nasa.rakutest:2
===> Testing [FAIL]: Astro::Utils:ver<0.0.3>:auth<zef:tbrowder>
Aborting due to test failure: Astro::Utils:ver<0.0.3>:auth<zef:tbrowder> (use --force-test to override)
```

PR #6178, opened 2026-05-13
RakuAST: retry .& resolution at check time to handle forward references

ugexe:
Under `RAKUDO_RAKUAST=1`, `.&foo` against a `my sub` declared later in the same scope is reported as undeclared, even though plain `foo()` works fine.

```
RAKUDO_RAKUAST=1 raku -e 'sub MAIN { say 42.&inc }; my sub inc($n) { $n + 1 }'
# ===SORRY!=== Undeclared routine:
#     inc used at line 1
```

`RakuAST::Call::Name` already handles this: it has a `PERFORM-CHECK` that retries `PERFORM-BEGIN` if resolution failed at BEGIN time, which is how forward references to `my sub` work in general. `RakuAST::Call::VarMethod` was missing the same retry. This adds it.

### DIFF
--- a/src/Raku/ast/call.rakumod
+++ b/src/Raku/ast/call.rakumod
@@ -1126,6 +1126,7 @@ class RakuAST::Call::VarMethod
   is RakuAST::Call::Methodish
   is RakuAST::Lookup
   is RakuAST::BeginTime
+  is RakuAST::CheckTime
 {
     has RakuAST::Name $.name;
 
@@ -1176,6 +1177,12 @@ class RakuAST::Call::VarMethod
         }
 
         Nil
+    }
+
+    method PERFORM-CHECK(RakuAST::Resolver $resolver, RakuAST::IMPL::QASTContext $context) {
+        unless self.is-resolved {
+            self.PERFORM-BEGIN($resolver, $context);
+        }
     }
 
     method undeclared-symbol-details() {


coke:
### :white_check_mark: **Snyk checks have passed. No issues have been found so far.**

| Status | Scan Engine |  Critical |  High |  Medium |  Low | Total (0) |
|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/raku.org-default/pr-checks/0d6c309d-c299-4b6f-9483-f2cdcaceaade?refs=60ff486a024973b0654eb753b3d56c12168c93ae&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/raku.org-default/pr-checks/0d6c309d-c299-4b6f-9483-f2cdcaceaade/license?refs=60ff486a024973b0654eb753b3d56c12168c93ae&source=prChecksComment) |

:computer: Catch issues earlier using the plugins for [VS Code](https://marketplace.visualstudio.com/items?itemName=snyk-security.snyk-vulnerability-scanner), [JetBrains IDEs](https://plugins.jetbrains.com/plugin/10972-snyk-security), [Visual Studio](https://marketplace.visualstudio.com/items?itemName=snyk-security.snyk-vulnerability-scanner-vs-2022), and [Eclipse](https://marketplace.eclipse.org/content/snyk-security#external-install-button).

lizmat:
Hmmm... I can't help but think that all of these checks should *always* be done at CHECK time.
```
$ RAKUDO_RAKUAST=1 raku -e 'sub a($a) { dd }; { 42.&a; sub a(|) { dd } }'
sub a(|)
```
So this didn't complain about the missing `&a` because it found one outside, yet at execution time it *does* pick the correct one defined inside the scope.

So is there a reason to do this at BEGIN time at all?

ugexe:
> So is there a reason to do this at BEGIN time at all?

Because `constant X = 42.&inc;` is BEGIN time. The general fix for what you're demonstrating would be a separate PR that re-resolves at CHECK regardless of whether BEGIN found a hit and replaces the bound resolution when the CHECK answer differs, so a later same scope declaration can shadow what BEGIN bound. That one would need to land across all the affected Call:: classes together (at least I assume it affects other classes since this change was modeled after existing code) since they all would share the same BEGIN/CHECK split.

lizmat:
Still, I think that example could be done at `CHECK` time as well: if `&inc` can not be found, execution of the thunk will fail anyway.

ugexe:
If `.&inc` only resolved at CHECK time the constant initializer would have nothing to call when BEGIN evaluates it.

PR #6180, opened 2026-05-13
RakuAST: evaluate non-trivial package colonpair values once, cached

ugexe:
Under `RAKUDO_RAKUAST=1`, expression-valued package adverbs like `module M:ver($expr)` or `module M:ver({...}())` were either dropped (treated as if the version weren't there) or evaluated multiple times, with side effects firing once per consumer. This is what was making `module Zef:ver($?DISTRIBUTION.meta<version> // '*')` produce `Use of Nil in string context` and report `Zef.^ver` as `(Mu)` instead of the real version.

```
RAKUDO_RAKUAST=1 raku -e 'module M:ver({ note "evaluated"; "1.0" }()) { }; say M.^ver'
# evaluated
# evaluated
# evaluated
# (Mu)
```

Expected: one `evaluated` line and `v1.0`.

Legacy resolves this in `Perl6::World.compile_time_evaluate`, called once per colonpair from `colonpairs_hash`, with a thunk-and-call fallback for non-trivial expressions. The fix mirrors that on `RakuAST::ColonPair::Value` by memoising the resolved value on the node and sharing the Failure-rethrow plumbing between the package-decl and use-statement consumers.

Two deliberate divergences from legacy, called out so they're not surprises:

- legacy stringifies a handled `Failure` through `Version.new`, producing garbage like `vHANDLED.<msg>.in.sub...` with backtrace frames as version parts. We rethrow at compile time instead.
- legacy silently drops `:api(Nil)`. We warn and treat it as empty, matching `:ver(Nil)`, since `:api` is a `Version` on the dependency-spec side. `:auth(Nil)` stays silent because `auth` is a string identity.

A subtler thing worth flagging for review: `RakuAST::Role.replace-body` is now called twice. `Package.new` calls it first with a placeholder body before parser state is bound; `package-def` calls it again with the real body. The body-wrapping branch (which calls `self.stubbed-meta-object`) is gated on the caller passing a real body, because `stubbed-meta-object` memoises and calling it with unbound parser state would freeze in a wrong meta-object. Side effect of doing it this way: `role R:ver({...}())` now works at all, where it used to crash with `lang-call cannot invoke object of type 'VMNull'`.

### DIFF
--- a/src/Raku/ast/package.rakumod
+++ b/src/Raku/ast/package.rakumod
@@ -32,6 +32,9 @@ class RakuAST::Package
 
     has RakuAST::CompilerServices $.compiler-services;
 
+    has RakuAST::Resolver $!resolver;
+    has RakuAST::IMPL::QASTContext $!context;
+
     has Mu $!compose-exception;
 
     # Enclosing parametric role captured at BEGIN-time so the package can
@@ -160,6 +163,8 @@ class RakuAST::Package
             }
         }
 
+        nqp::bindattr(self, RakuAST::Package, '$!resolver', $resolver);
+        nqp::bindattr(self, RakuAST::Package, '$!context', $context);
         nqp::bindattr(self, RakuAST::Package, '$!compiler-services', RakuAST::CompilerServices.new(self, $resolver, $context));
     }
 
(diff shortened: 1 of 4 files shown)


coke:
### :white_check_mark: **Snyk checks have passed. No issues have been found so far.**

| Status | Scan Engine |  Critical |  High |  Medium |  Low | Total (0) |
|---|---|---|---|---|---|---|
| :white_check_mark: | Open Source Security| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/raku.org-default/pr-checks/994c39de-3354-463a-8073-f9e657dc4a64?refs=3761996b59a1481638086236a7dfe47b97bdefee&source=prChecksComment) |
| :white_check_mark: | Licenses| 0 | 0 | 0 | 0 | [ 0 issues](https://app.snyk.io/org/raku.org-default/pr-checks/994c39de-3354-463a-8073-f9e657dc4a64/license?refs=3761996b59a1481638086236a7dfe47b97bdefee&source=prChecksComment) |

:computer: Catch issues earlier using the plugins for [VS Code](https://marketplace.visualstudio.com/items?itemName=snyk-security.snyk-vulnerability-scanner), [JetBrains IDEs](https://plugins.jetbrains.com/plugin/10972-snyk-security), [Visual Studio](https://marketplace.visualstudio.com/items?itemName=snyk-security.snyk-vulnerability-scanner-vs-2022), and [Eclipse](https://marketplace.eclipse.org/content/snyk-security#external-install-button).

PR #6190, opened 2026-05-17
RakuAST: fix deparsing of RakuAST::VarDeclaration::Anonymous

lizmat:
Is basically the same as for RakuAST::VarDeclaration::Simple (apart from the name) so abstract the logic into a helper sub

### DIFF
--- a/src/core.c/RakuAST/Deparse.rakumod
+++ b/src/core.c/RakuAST/Deparse.rakumod
@@ -596,7 +596,7 @@ CODE
             @parts.push(self.deparse($initializer));
         }
 
-        self.add-any-docs(@parts.join, $ast.WHY)
+        @parts.join
     }
 
 #- A ---------------------------------------------------------------------------
@@ -2923,7 +2923,7 @@ CODE
     }
 
     multi method deparse(RakuAST::VarDeclaration::Simple:D $ast --> Str:D) {
-        self.var-declaration($ast)
+        self.add-any-docs(self.var-declaration($ast), $ast.WHY)
     }
 
     multi method deparse(RakuAST::VarDeclaration::Term:D $ast --> Str:D) {
--- a/src/core.c/RakuAST/Deparse.rakumod
+++ b/src/core.c/RakuAST/Deparse.rakumod
@@ -558,11 +558,12 @@ CODE
         self.hsyn("typer-$typer", self.xsyn('typer', $typer))
     }
 
-    method var-declaration($ast, str $name = $ast.name) {
-        my str $scope = $ast.scope;
+    method var-declaration(
+      $ast, str $name = $ast.name, str $scope = $ast.scope
+    ) {
         my str @parts;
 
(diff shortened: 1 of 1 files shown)


src/core.c/RakuAST/Deparse.rakumod
```diff
@@ -558,6 +558,47 @@ CODE
         self.hsyn("typer-$typer", self.xsyn('typer', $typer))
     }
 
+    method var-declaration($ast, str $name = $ast.name) {
```
ugexe:
```suggestion
    method var-declaration(RakuAST::VarDeclaration::Simple:D $ast, str $name = $ast.name) {
```

src/core.c/RakuAST/Deparse.rakumod
```diff
@@ -2805,11 +2846,10 @@ CODE
 
     multi method deparse(RakuAST::VarDeclaration::Anonymous:D $ast --> Str:D) {
         my str $sigil = $ast.sigil;
-        my str $scope = $ast.scope;
 
-        $scope eq 'state'
+        $sigil eq '$' && $ast.scope eq 'state'
           ?? $sigil
-          !! self.xsyn('scope', $scope) ~ ' ' ~ $sigil
+          !! self.var-declaration($ast, $sigil).chomp
```
ugexe: This chomp seems wrong. Observe:
```
➜  rakudo git:(e9b57d0a4) ./install/bin/raku -e 'use experimental :rakuast; say Q{my @ = 1; 99}.AST.DEPARSE.raku'
"my \@ = 1;99\n"
➜  rakudo git:(e9b57d0a4) ./install/bin/raku -e 'use experimental :rakuast; say Q{my @x = 1; 99}.AST.DEPARSE.raku'
"my \@x = 1;\n99\n"
```

src/core.c/RakuAST/Deparse.rakumod
ugexe: Probably cleaner to split var-declaration into a body builder + a wrap step (add-any-docs), so `RakuAST::VarDeclaration::Anonymous` calls the body only form for expression context and the full form at statement level.

lizmat: If I would start doing that here, it would need a *lot* of other cases as well.

I felt that with these internal methods, we wouldn't need to do that.  Or is there a performance benefit I don't know about?

ugexe: Is it an internal method? To me it looks like public API since it isn't named e.g. `IMPL-FOO`.

It is useful for understanding what the function intends to accept. Maybe it is obvious to you right now as the person writing the code, but someone else just coming into the code base would have to infer it somehow. Further, we type `$name` already which implies the typing must be useful to this function. Developers used to their tooling making use of these types of things (me) find these useful. If it isn't significantly slower I don't understand why we wouldn't be doing it.

PR #6196, opened 2026-05-18
RakuAST: parens around postfix operand always border currying

ugexe:
`(5 ~~ *).WHAT` returned a curried WhateverCode whose deparsed string was `.WHAT` (and `(*+1).Str`, `(*).foo`, etc. behaved similarly): single parens around a curryable expression were stripped, leaving the inner expression as the bare operand, and the surrounding ApplyPostfix then absorbed the inner currying into a single WhateverCode covering `(expr).postfix`.

Only the double-parens case set IMPL-MUST-NOT-CURRY.  Drop the double-parens gate; any time we unwrap single-expression parens from an ApplyPostfix operand, mark the postfix as a currying border.  The parenthesized operand evaluates to its value first, the postfix then applies to that value, matching the legacy frontend.

Reproducer:

    raku -e 'say (5 ~~ *).WHAT'

Legacy   : (WhateverCode)
RakuAST  : .WHAT  (before)
           (WhateverCode)  (after)

### DIFF
--- a/src/Raku/ast/call.rakumod
+++ b/src/Raku/ast/call.rakumod
@@ -652,6 +652,17 @@ class RakuAST::Call::Method
         SPECIAL-OPS{$name}
     }
 
+    # Special-op postfixes (.WHAT, .HOW, .VAR, .DEFINITE, etc.) compile to
+    # primitive nqp:: ops, not real method calls, so they sit outside the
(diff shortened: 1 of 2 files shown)


ab5tract:
This looks good to me, though I notice it is marked as a draft.

ugexe:
> This looks good to me, though I notice it is marked as a draft.

The CI shows my implementation caused some regressions, so it'll need to be adjusted

PR #6201, opened 2026-05-20
RakuAST: emit proper SORRY exceptions for invalid feed-operator stages

ugexe:
`RakuAST::Feed.IMPL-LIST-INFIX-QAST` validated each non-source stage against the legal QAST shapes and fell through to an error path that called `$_.PRECURSOR.panic($error)`.  That line was lifted from `make_feed` in `src/Perl6/Actions.nqp` where `$_` is a Match (which has `PRECURSOR`); in RakuAST `$_` is a QAST node, which doesn't, so when an invalid stage actually reached the error path the compiler crashed with `Cannot find method 'PRECURSOR' on object of type QAST::Want`.  For some shapes the validation didn't trigger at all and the program just produced wrong output - e.g.
`my $x = grep(...) <== map(...) <== (1..5)` silently returned `()`.

Validation moves to `RakuAST::ApplyListInfix.PERFORM-CHECK`, which has a `Resolver` in scope and the natural `add-sorry` / `build-exception` reporting path.  Acceptance set narrowed to `RakuAST::Call` and `RakuAST::Var` to match legacy.  `&`-sigil detection broadened from `Var::Lexical` to all `Var` subclasses so package-qualified code refs like `&Pkg::foo` get the `did you mean foo()?` hint.  The shaped-arrays wording is restricted to `my`/`our` scope, matching legacy's QAST-shape-based check.

### DIFF
--- a/src/Raku/ast/expressions.rakumod
+++ b/src/Raku/ast/expressions.rakumod
@@ -885,26 +885,9 @@ class RakuAST::Feed
                 $stage := QAST::Op.new( :op('locallifetime'), $stage, $tmp );
             }
             else {
-                my str $error := "Only routine calls or variables that can '.append' may appear on either side
-of feed operators.";
-                if nqp::istype($stage, QAST::Children) && nqp::istype($stage[0], QAST::Var) {
-                    if nqp::istype($stage, QAST::Op) && $stage.op eq 'ifnull'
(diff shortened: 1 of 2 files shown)


ab5tract:
This looks great 👍🏻

PR #6202, opened 2026-05-20
RakuAST: two Code.assuming bugs in synthesised closures

ugexe:
`Code.assuming` synthesises a `RakuAST::Sub` (or `PointyBlock`) at runtime and `.EVAL`s it. Two bugs in that synthesis showed up trying to install `JSON::RPC` and `Pod::Contents`.

`JSON::RPC`'s `BUILD` does `&transport.assuming(uri => $uri)` where `transport` has `URI :$uri`, and died with `'URI' has not been resolved. Type: RakuAST::Type::Simple`: the synthesised `Sub` ran `.EVAL` with `assuming`'s own setting scope as its lexical context, not the caller's, so types defined via `use` never got found. Route `.EVAL` through the caller's scope and replace `RakuAST::Term::Self` / `RakuAST::Var::Lexical("@positionals")` in the synthesised body with literal-embedded values, so nothing else has to be borrowed from the old scope. Embedding the slurpy itself as a literal keeps `is raw` working: indexed slots of an `is raw` slurpy stay the original containers.

`Pod::Contents`'s `&check_pod.assuming: *, $pod_type` hits a separate bug: `check_pod`'s first param is subset-typed (`POD $pod`), and `ParameterAST` was treating any single non-Code constraint as a literal-value match and deleting the parameter's target. Fine for `sub f(42) { }`, wrong for `sub f(Simple $a) { }` where the constraint is the subset type object. Invisible when currying consumes the param, breaks when a Whatever placeholder lets it survive into the synthesised signature with no name. Gate the value-match path on `nqp::isconcrete($head)`; type objects fall through to the where-clause path which keeps the name.

### DIFF
--- a/src/core.c/core_epilogue.rakumod
+++ b/src/core.c/core_epilogue.rakumod
@@ -384,11 +384,18 @@ augment class Code {
         if $parameter.constraint_list -> @constraints {
             my $head := @constraints.head;
             if @constraints.elems == 1
-              && nqp::not_i(nqp::istype($head,Code)) {
-                %args<value> = $head;  # literalize($head)) ??
+              && nqp::not_i(nqp::istype($head,Code))
+              && nqp::isconcrete($head) {
+                # Concrete value constraint, e.g. `sub f(42) { }`:
+                # match by value, parameter is anonymous.
+                %args<value> = $head;
                 %args<target>:delete;
             }
             else {
+                # Type-object constraint (subset type, etc.) or a
+                # `where`-clause Code: keep the parameter's target so
+                # the synthesised body can reference it by name; turn
+                # the constraint into a where-clause on the binding.
                 %args<where> = literalize($head);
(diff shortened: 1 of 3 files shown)


lizmat:
Hi, great news!  Confirmed it fixes also the raw case.

  But I have some code styling nits.

Would you like me to apply them *after* merging this PR, or shall I make a separate PR without the nits?  And then not merge this one?

ugexe:
It would probably be better to apply them after this PR is merged so that the commit messages explaining why things are implemented this way are preserved

lizmat:
Understood.  I'll wait until after the release as not to interfere with @coke's work

coke:
This addresses two of the issues found in the blin run, though, no?

lizmat:
Yes, it does.  Just to be 100% clear :-)

PR #6211, opened 2026-05-28
RakuAST: step one for codegenning POPULATE in every class

lizmat:
This actually compiles and works as expected (as far as I have been able to test in a simulated environment.

The next step will be to actually convert the created AST for the POPULATE method into an actual Method object, and add it to the class.  This appears to not be as straightforward as I hoped.

Note that this approach *completely* bypasses the BUILDPLAN logic, hopefully making that completely unneeded once RakuAST lands as the default implementation.


lizmat:
Closing in favour of a new PR being created now.

PR #6214, opened 2026-06-01
Make RAKUDO_RAKUAST=0 *not* activate the RakUAST logic

lizmat:
Spotted by timo++


timo:
does this prevent rakudo from starting if the value in `RAKUDO_RAKUAST` is something like `yes` or `true` or anything else that can't intify? since this is inside nqp, we don't get a `Failure` from `prefix:<+>`

lizmat:
My mail goals were: simple fix, make `=0` do what's expected.   That `'=foo` doesn't complain I find to be less of an issue.

timo:
turns out intifying something that isn't a valid number in nqp doesn't throw an exception, it just returns 0, same with a number that is followed by invalid characters, so this is actually fine

PR #6215, opened 2026-06-01
TODO both problematic nativecall tests for moar+clang+dyncall

timo:
the code clang generates assumes the upper bytes of an unsigned char or unsigned short are zeroed out, but dyncall only has a function for passing signed char and short, where negative numbers end up sign-extended, messing clang's generated assembly for x == 0xFE up, which tries to comare the whole 32bit register against 0xFE.


coke (approved):
This should clean up the MoarVM CI, thanks!

PR #6216, opened 2026-06-01
RakuAST: allow anonymous declarations across all package kinds

ugexe:
Anonymous package declarations under RAKUDO_RAKUAST=1 had three bugs:

`subset :: of Str where /42/` crashed in
RakuAST::Type::Subset.PERFORM-BEGIN's IMPL-INSTALL-PACKAGE call. This broke JSON::Unmarshal whose t/070-parameterised uses `has Str %.bla{subset :: of Str where ...}`.

`class :: { }`, `role :: { }`, `module :: { }`, `grammar :: { }` compiled but `.^name` returned the empty string because RakuAST::Package::PRODUCE-STUBBED-META-OBJECT passed an empty :name to the HOW.

`role R { subset :: of Int where * > 0; ... }` died with X::Declaration::OurScopeInRole because check-scope fired before the install path could short-circuit.

A new RakuAST::Name.is-installable predicate is the single gate at every site that decides whether to write a name, install a symbol, or pass :name to a HOW. It returns False for any Name with no identifiable symbol (empty Name, single empty part, or the two- Empty-part `::` form).

IMPL-INSTALL-PACKAGE early-returns Nil for non-installable names. Subset and Package PRODUCE-STUBBED-META-OBJECT omit :name so the HOW defaults (`<anon>` for SubsetHOW, `<anon|N>` for ClassHOW/ RoleHOW/ModuleHOW/GrammarHOW) apply. Type::Enum passes :name unconditionally because EnumHOW.new_type requires it, matching legacy. Type::Enum::PERFORM-BEGIN switches its gate to is-installable for consistency. Behavior is unchanged.

check-scope skips the OurScopeInRole sorry for RakuAST::PackageInstaller declarations (Package, Type::Subset, Type::Enum) with a non-installable name. Routines and variables install through other paths where the strict check applies.

RakuAST::Name deparses as `::` when not installable so anonymous AST nodes round-trip through deparse without an empty name slot.

Tests in t/12-rakuast/subset.rakutest and t/12-rakuast/enum.rakutest cover AST construction in one-Empty and two-Empty Name shapes, deparse round-trip, .^name parity across all package kinds, the JSON::Unmarshal hash typed-key case, anonymous subset in a role, and that anonymous our-scoped routines in a role throw X::Declaration::OurScopeInRole.

### DIFF
--- a/src/Raku/ast/name.rakumod
+++ b/src/Raku/ast/name.rakumod
@@ -89,6 +89,12 @@ class RakuAST::Name
         nqp::elems($!parts) == 2 && $!parts[0].is-empty && $!parts[1].is-empty # name is just '::'
     }
 
+    # True when this name denotes a symbol that gets installed in a
+    # scope.
+    method is-installable() {
+        !self.is-empty && !self.is-anonymous
+    }
+
     method is-simple() {
         for $!parts {
             return False unless nqp::istype($_, RakuAST::Name::Part::Simple);
--- a/src/Raku/ast/package.rakumod
+++ b/src/Raku/ast/package.rakumod
@@ -197,7 +197,7 @@ class RakuAST::Package
             my str $scope := self.scope;
             $scope := 'our' if $scope eq 'unit';
             my $name := $!name;
-            if $name && !$name.is-empty && !$name.is-anonymous {
+            if $name && $name.is-installable {
                 my $type-object := self.stubbed-meta-object(:$resolver, :$context);
                 my $full-name := self.IMPL-FULL-NAME($resolver);
                 $type-object.HOW.set_name(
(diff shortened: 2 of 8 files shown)


src/Raku/ast/name.rakumod
```diff
@@ -89,6 +89,12 @@ class RakuAST::Name
         nqp::elems($!parts) == 2 && $!parts[0].is-empty && $!parts[1].is-empty # name is just '::'
     }
 
+    # True when this name denotes a symbol that gets installed in a
+    # scope.
+    method is-installable() {
+        !self.is-empty && !self.is-anonymous
```
lizmat, line 95: Perhaps better as: `!(self.is-empty || self.is-anonymous)`

PR #6218, opened 2026-06-02
RakuAST: emit empty concat for modifier-only alternation branch

ugexe:
When an internal regex modifier (`:i`, `:m`, `:r`, `:s`) appears alone as a top-level branch of a `|` or `||` alternation, its IMPL-REGEX-QAST returns Nil. `RakuAST::Regex::Branching` was pushing that Nil into the QAST::Regex alt, and the NFA builder then died with `elems requires a concrete object` while inspecting the branch.

Legacy emits an empty `QAST::Regex(concat)` for the same shape, which the NFA and capnames helpers handle. Do the same in RakuAST.

Surfaced by `zef install Grammar::HTTP`, whose Grammar::IETF::IOL::RFC5646 contains `token irregular { :i\n| 'en-GB-oed'\n| 'i-ami'\n... }`.

Covers `:i` and `:!i` as the first branch of `|`, a modifier in non-first branch position, `:i` as the first branch of `||` (`SequentialAlternation`), and `:i` as the first branch of `&` (`Conjunction`). Each `Branching` subclass that inherits the patched method has direct coverage.

### DIFF
--- a/t/02-rakudo/regex-modifier-alt-branch.t
+++ b/t/02-rakudo/regex-modifier-alt-branch.t
@@ -1,0 +1,70 @@
+use lib <t/packages/Test-Helpers>;
+use Test;
+use Test::Helpers;
+
+plan 5;
+
+# An internal regex modifier (`:i`, `:m`, `:r`, `:s`) alone as a top-level
+# branch of an alternation or conjunction applies to the surrounding
+# construct and compiles to an empty branch.
+
+is-run q:to/CODE/,
+    grammar G {
+        token irregular {
+            :i
+            | 'en-GB-oed'
+            | 'i-ami'
+        }
+    }
+    say G.parse('EN-GB-OED', :rule<irregular>) // 'no';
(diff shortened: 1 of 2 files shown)


t/02-rakudo/regex-modifier-alt-branch.t
```diff
@@ -0,0 +1,70 @@
...
+use Test;
+use Test::Helpers;
+
+plan 5;
+
+# An internal regex modifier (`:i`, `:m`, `:r`, `:s`) alone as a top-level
+# branch of an alternation or conjunction applies to the surrounding
+# construct and compiles to an empty branch.
+
+is-run q:to/CODE/,
```
lizmat, line 11: Why is it necessary to use `is-run` here?  Surely we can test this without needing to start up another process?

ugexe: How do you propose testing compile time errors?

lizmat: With a string EVAL ?

PR #6223, opened 2026-06-03
RakuAST: declare the anonymous state variable for assign meta-operators

ugexe:
The parser synthesises an anonymous state variable for an assign meta-operator without a left operand, like `=%=`. The synthesised declaration was missing a `declare-lexical` call. Evaluation of `my constant \X =%= %( ... )` at BEGIN time then failed with "Could not find a compile-time-value for lexical %ANON_VAR_1".

### DIFF
--- a/src/Raku/ast/statements.rakumod
+++ b/src/Raku/ast/statements.rakumod
@@ -1917,6 +1917,14 @@ class RakuAST::Statement::Use
     }
 
     method PERFORM-BEGIN(RakuAST::Resolver $resolver, RakuAST::IMPL::QASTContext $context) {
+        CATCH {
+            my $ex := $resolver.convert-exception($_);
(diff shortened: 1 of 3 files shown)


src/Raku/ast/statements.rakumod
```diff
@@ -1917,6 +1917,14 @@ class RakuAST::Statement::Use
     }
 
     method PERFORM-BEGIN(RakuAST::Resolver $resolver, RakuAST::IMPL::QASTContext $context) {
+        CATCH {
```
lizmat, line 1920: Is this supposed to be part of this PR?

ugexe: Yes

ugexe: You can look at the first commit and see the failing spectest run. The second commit addresses that spectest failure caused by the first commit

Issue #5419, opened 2023-10-15
Object hashes are created with `Any` instead of `Mu` as default 'of' type


ab5tract:
## The Problem

When creating an object hash without specifying an 'of type', it defaults to `Any`. But the default 'of type' of a regular hash is `Mu`.

## Expected Behavior

```
> my %h; %h.of.say
(Mu)
> my %h{Int}; %h.of.say
(Mu)
```

## Actual Behavior

```
> my %h; %h.of.say
(Mu)
> my %h{Int}; %h.of.say
(Any)
```

EDIT: This was originally reported with the behaviors flipped.

## Environment
* Operating system: ArchLinux
* Compiler version (`rakudo -v` or `raku -v`): `2023.09-104-gd6cb20606`

2colours:
First, the "expected vs actual behavior" contradicts the informal description - it's implying that it should be `Any` instead of `Mu`.

Second, this is actually a design decision. I think it's strange enough that `Mu` pops up here as a default value at all (although it does pop up for `Array`s as well) - but in any case, these sort of things should be specified.

lizmat:
I believe 8f8b95866f7e292ff3331, 998cde58a70898fe5e1a4554, bcdbbdd8bc21e78af8f normalized the behaviour of object hashes to have `Any` as the default for the containers.  So closing.

ab5tract:
> I believe [8f8b958](https://github.com/rakudo/rakudo/commit/8f8b95866f7e292ff3331a11ab336ec3bf5f8c6d), [998cde5](https://github.com/rakudo/rakudo/commit/998cde58a70898fe5e1a45543094ea21de6c073a), [bcdbbdd](https://github.com/rakudo/rakudo/commit/bcdbbdd8bc21e78af8f72388ce3e7840df8dcc14) normalized the behaviour of object hashes to have `Any` as the default for the containers. So closing.

Did you mean `Mu` as the default?

PR #6226, opened 2026-06-04
RakuAST: default object hash value type to Any when no explicit type is given

ugexe:
`my %h{K}` should parameterise as Hash[Any, K]. The shape-with-no-type branch of IMPL-CALCULATE-TYPES was passing the explicit container base type (Mu when none was supplied) as the value type instead of Any, so the variable came out as Hash[Mu, K]. Surfaced by CBOR::Simple's t/lib/CodecMatches.rakumod comparing decoded maps against their expected shape.

### DIFF
--- a/src/Raku/ast/variable-declaration.rakumod
+++ b/src/Raku/ast/variable-declaration.rakumod
@@ -228,10 +228,13 @@ class RakuAST::ContainerCreator {
                             $bind-constraint, $of, $key-type);
                     }
                     else {
+                        # No explicit value type with a shape: legacy defaults
+                        # the value type to Any so `my %h{K}` parameterises as
+                        # `Hash[Any, K]`, not `Hash[Mu, K]`.
                         $container-type := Hash.HOW.parameterize(
-                            Hash, $explicit-base, $key-type);
+                            Hash, Any, $key-type);
                         $bind-constraint := $bind-constraint.HOW.parameterize(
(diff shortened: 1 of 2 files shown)


ugexe:
@lizmat @ab5tract this PR fixes `CBOR::Simple` but breaks this test:
https://github.com/rakudo/rakudo/blob/f7576048ddcadbc183c8e0eb2011fccda7d713b8/t/12-rakuast/xx-fixed-in-rakuast.rakutest#L36-L40

I'm not sure how to interpret the discussion from https://github.com/rakudo/rakudo/issues/5419 either. Which behavior do we want? Given the example that was given:

```
➜  rakudo git:(ugexe/rakuast-object-hash-default-value-type) RAKUDO_RAKUAST=1 ./install/bin/raku -e 'my %b{Int}; %b{1} = 5|6;'
➜  rakudo git:(ugexe/rakuast-object-hash-default-value-type) ./install/bin/raku -e 'my %b{Int}; %b{1} = 5|6;'
Type check failed for an element of %b; expected Any but got Junction (any(5, 6))
  in block <unit> at -e line 1
```
I think I prefer the `Mu` behavior instead of `Any` (that this PR implements). Given how I've only just now seen this issue it could make sense to update ecosystem modules (and potentially revision gate the change in behavior).

ugexe:
I'm closing this in favor of #6228 which gates the new behavior to 6.e+

PR #6236, opened 2026-06-05
RakuAST: silence some spurious `Useless use` sink worries

ugexe:
`Useless use of X in sink context` worries from the RakuAST frontend cover a slightly wider set than legacy, and the extras leak into ecosystem installs as noise.

`ApplyListInfix::PERFORM-CHECK` calls `add-sunk-worry` whenever the infix is pure, which fires for `,` because `infix:<,>` is pure. `1, 2, 3;` at statement level got a fourth `Useless use of , in sink context` on top of the three constant-integer operand worries. Legacy explicitly skips `,` (and `xx`) for the list-literal case. Add `!IMPL-IS-LIST-LITERAL` to the guard.

`RakuAST::Stub` is a `RakuAST::Term`, so it inherits the unconditional sunk worry from `RakuAST::Expression::PERFORM-CHECK`. Legacy treats `...`, `???`, `!!!` as `&fail` / `&warn` / `&die` calls and the impure-call check keeps them quiet. A stubbed method body like `method foo(--> Nil) { ... }` ended up sunk by `--> Nil` and the inherited worry surfaced. Override `PERFORM-CHECK` on `RakuAST::Stub` (the three concrete stubs inherit it).

`Term::Rand`, `Term::Whatever` and `Term::HyperWhatever` have the same inheritance shape. `rand;`, `*;`, `**;` at statement level got the inherited worry, but `rand` is impure and the two singletons carry meaning even when sunk, so legacy stays silent on all three. Override `PERFORM-CHECK` on each.

Surfaced by `zef install URI` (the comma case at `t/01.rakutest:114`), `zef install Log::Timeline` (the stub case at `Log::Timeline::Output:6`), and any role with a `method ... (--> Nil) { ... }` body.

### DIFF
--- a/src/Raku/ast/call.rakumod
+++ b/src/Raku/ast/call.rakumod
@@ -1329,6 +1329,10 @@ class RakuAST::Stub
         $obj
     }
 
+    method PERFORM-CHECK(RakuAST::Resolver $resolver, RakuAST::IMPL::QASTContext $context) {
+        True
+    }
+
     method PRODUCE-IMPLICIT-LOOKUPS() {
         [
             RakuAST::Type::Simple.new(
--- a/src/Raku/ast/expressions.rakumod
+++ b/src/Raku/ast/expressions.rakumod
@@ -2357,7 +2357,10 @@ class RakuAST::ApplyListInfix
         }
 
         self.add-sunk-worry($resolver, self.origin ?? self.origin.Str !! self.DEPARSE)
-            if self.infix.is-pure && self.sunk && !self.infix.short-circuit;
+            if self.infix.is-pure
+            && self.sunk
+            && !self.infix.short-circuit
+            && !self.IMPL-IS-LIST-LITERAL;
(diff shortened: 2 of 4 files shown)


src/Raku/ast/call.rakumod
```diff
@@ -1329,6 +1329,10 @@ class RakuAST::Stub
         $obj
     }
 
+    method PERFORM-CHECK(RakuAST::Resolver $resolver, RakuAST::IMPL::QASTContext $context) {
```
lizmat, line 1332: This feels repetitive.  I wonder whether making a separate RakuAST::NoUselessUse class with just this would make this more visible and maintainable

ugexe: Probably, but I assume that is a much bigger change which itself will be easier by having these already explicitly discoverable. Right now it is opt-in instead of opt-out.

PR #6237, opened 2026-06-05
RakuAST: handle custom package declarators like legacy does

ugexe:
`stub-package` mapped only the seven built-in declarators (`package`, `role`, `class`, `module`, `grammar`, `knowhow`, `native`) to their `RakuAST::Package` subclasses and fell back to bare `RakuAST::Package` otherwise. Custom declarators registered via `EXPORTHOW::DECLARE` (the way `OO::Monitors` adds `monitor`) became bare `RakuAST::Package`, which is not `Package::Attachable`, so any attribute reference in the body died at BEGIN time with `No such method 'ATTACH-ATTRIBUTE-USAGE' for invocant of type 'RakuAST::Package'`.

Route a custom declarator to `RakuAST::Class` when the HOW can take attributes, otherwise to bare `RakuAST::Package`. The HOW handles compose either way. Matches legacy's `package_def`, which only branches on `$*PKGDECL eq 'role'` and `'grammar'` and parses every other declarator class-like.

`RakuAST::Var::Attribute::PERFORM-BEGIN` raises
`X::Attribute::Package` when the attach target is not Attachable. Both that site and the existing `VarDeclaration::Simple` site read `package-kind` from `nqp::getlexdyn('$*PKGDECL')`, falling back to the attach target's `declarator`. A `mymod Foo { has $!x }` reports `A mymod cannot have attributes` instead of the AST class's hard-coded `package`.

Surfaced by `zef install Terminal::Print`, whose
`Terminal::Print::Grid` is a `unit monitor`.

### DIFF
--- a/src/Raku/ast/variable-declaration.rakumod
+++ b/src/Raku/ast/variable-declaration.rakumod
@@ -849,9 +849,12 @@ class RakuAST::VarDeclaration::Simple
                 nqp::bindattr(self, RakuAST::VarDeclaration::Simple, '$!attribute-package',
                     $attribute-package);
                 if self.is-attribute && !$!attribute-package.can-have-attributes {
+                    my $pkgdecl := nqp::getlexdyn('$*PKGDECL');
                     my $ex := $resolver.build-exception: 'X::Attribute::Package',
                         name         => self.name,
-                        package-kind => $!attribute-package.declarator;
+                        package-kind => nqp::isconcrete($pkgdecl)
+                            ?? ~$pkgdecl
+                            !! $!attribute-package.declarator;
                     if nqp::istype($!attribute-package, RakuAST::Declaration::External::Package) {
                         self.add-worry: $ex;
                     }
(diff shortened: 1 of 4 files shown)


src/Raku/ast/variable-declaration.rakumod
```diff
@@ -849,9 +849,12 @@ class RakuAST::VarDeclaration::Simple
                 nqp::bindattr(self, RakuAST::VarDeclaration::Simple, '$!attribute-package',
                     $attribute-package);
                 if self.is-attribute && !$!attribute-package.can-have-attributes {
+                    my $pkgdecl := nqp::getlexdyn('$*PKGDECL');
```
lizmat, line 852: This feels like a grammar dependency being included in what should probably be grammar agnostic ?

ugexe: Yeah I'm not sure. The original solution I was working with stored the package declaration (e.g. `monitor`) on a private attribute instead, but I noticed the legacy grammar was using this `$*PKGDECL` instead and switched based on that alone.

ugexe: Going to open a new PR

PR #6239, opened 2026-06-05
RakuAST: re-resolve `Call::Name` at CHECK to pick up a forward `sub`

### LINKED ISSUE #6238: RakudoAST failure in Rakudo 2026.05
RakuAST mis-resolves a forward reference to a lexically-declared sub that shadows a CORE routine.
Minimal Repro:

```raku
sub outer() {
  my @x; my @y;
  reduce(@x, @y);            # call resolves to CORE &reduce here…
  sub reduce(@a, @b) { @a }  # …but a lexical `reduce` is declared *after* the call
}
outer();
```

ugexe:
`Call::Name::PERFORM-BEGIN` resolves the call's name against the resolver's lexical chain. A `my`-scoped `sub foo { }` declared later in the same scope is hoisted to the start of that scope, but the hoist happens at parse time of the declaration, which is *after* the call's BEGIN-time pass. The call therefore resolved against an `External` declaration in an outer scope (typically `CORE`) and stuck there because `PERFORM-CHECK` only retried resolution when nothing had been picked up yet.

For `sub outer { reduce(@x, @y); sub reduce(@a, @b) { @a } }` the call's BEGIN-time lookup landed on `CORE::&reduce` and trial-bind reported `Calling reduce(Positional, Positional) will never work`, even though the local `&reduce` shadowed `CORE`.

`Call::Name::PERFORM-CHECK` now re-queries the resolver for the lexical name when the existing resolution is a
`RakuAST::Declaration::External`, and adopts the new resolution if it is not itself `External`. The hoisted local declaration is in scope by CHECK time, so a closer non-`External` match wins; `CORE`-only calls keep their `External::Setting` resolution and pay no churn.

Resolves #6238.

### DIFF
--- a/src/Raku/ast/call.rakumod
+++ b/src/Raku/ast/call.rakumod
@@ -346,6 +346,20 @@ class RakuAST::Call::Name
             self.PERFORM-BEGIN($resolver, $context);
         }
 
+        # A `sub foo { }` declaration is hoisted to the start of its
+        # lexical scope. When the existing resolution is an `External`
+        # (typically `CORE`), re-query and adopt the closer non-`External`
+        # match if one is present.
+        if $!name.is-identifier
+            && self.is-resolved
+            && nqp::istype(self.resolution, RakuAST::Declaration::External)
+        {
+            my $lexical := $resolver.resolve-lexical('&' ~ $name);
+            if $lexical && !nqp::istype($lexical, RakuAST::Declaration::External) {
+                self.set-resolution($lexical);
+            }
+        }
+
         if !self.is-resolved && ($name eq 'pi' || $name eq 'tau' || $name eq 'e') {
             self.add-sorry:
                 $resolver.build-exception: 'X::Undeclared',
--- a/t/02-rakudo/forward-sub-lookup.t
+++ b/t/02-rakudo/forward-sub-lookup.t
@@ -1,0 +1,47 @@
+use Test;
+
+plan 5;
+
+# A `sub foo` declaration is hoisted to the start of its enclosing
+# lexical scope, so a call earlier in the same scope must resolve to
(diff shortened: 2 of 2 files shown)


src/Raku/ast/call.rakumod
```diff
@@ -347,6 +346,20 @@ class RakuAST::Call::Name
             self.PERFORM-BEGIN($resolver, $context);
         }
 
+        # A `sub foo { }` declaration is hoisted to the start of its
+        # lexical scope. When the existing resolution is an `External`
+        # (typically `CORE`), re-query and adopt the closer non-`External`
+        # match if one is present.
+        if $!name.is-identifier
+            && self.is-resolved
```
lizmat, line 354: What if it isn't resolved yet?

Or can that never occur?  In which case, why check for it?

src/Raku/ast/call.rakumod
```diff
@@ -347,6 +346,20 @@ class RakuAST::Call::Name
...
+            && nqp::istype(self.resolution, RakuAST::Declaration::External)
```
lizmat, line 355: If the node isn't resolved yet, then the istype would also fail, would it not?  So maybe the check for resolution isn't needed at all?

ugexe: It would throw, not return false. This is an existing pattern:
https://github.com/rakudo/rakudo/blob/3c8b8c335807b38441d5286c26662d477b3ce351/src/Raku/ast/call.rakumod#L370-L372

src/Raku/ast/call.rakumod
ugexe: PERFORM-CHECK can reach this code with an unresolved node e.g. when PERFORM-BEGIN didn't pick anything up because the lexical name wasn't yet visible and no External fallback exists.

PR #6240, opened 2026-06-05
RakuAST: handle custom package declarators like legacy does

ugexe:
`stub-package` mapped only the seven built-in declarators (`package`, `role`, `class`, `module`, `grammar`, `knowhow`, `native`) to their `RakuAST::Package` subclasses and fell back to bare `RakuAST::Package` otherwise. Custom declarators registered via `EXPORTHOW::DECLARE` (the way `OO::Monitors` adds `monitor`) became bare `RakuAST::Package`, which is not `Package::Attachable`, so any attribute reference in the body died at BEGIN time with `No such method 'ATTACH-ATTRIBUTE-USAGE' for invocant of type 'RakuAST::Package'`.

Route a custom declarator to `RakuAST::Class` when the HOW can take attributes, otherwise to bare `RakuAST::Package`. The HOW handles compose either way. Matches legacy's `package_def`, which only branches on `$*PKGDECL eq 'role'` and `'grammar'` and parses every other declarator class-like.

`RakuAST::Var::Attribute::PERFORM-BEGIN` raises
`X::Attribute::Package` when the attach target is not Attachable. `RakuAST::Package` carries the literal declarator string from the parse site in a new `$!parsed-declarator` attribute exposed via `parsed-declarator`. Both that site and `VarDeclaration::Simple` read `package-kind` from it, so a `mymod Foo { has $!x }` reports `A mymod cannot have attributes` instead of the AST class's hard-coded `package`. Peer diagnostics (`X::Attribute::Undeclared`, `X::Augment::NoSuchType`, `X::Anon::Augment`, `X::Useless::Declaration`) also switch to `parsed-declarator` so a custom declarator renders as the user-typed keyword across error wording rather than landing as `class` for declarations routed to `RakuAST::Class`.

Surfaced by `zef install Terminal::Print`, whose `Terminal::Print::Grid` is a `unit monitor`.

### DIFF
--- a/src/Raku/ast/variable-access.rakumod
+++ b/src/Raku/ast/variable-access.rakumod
@@ -267,14 +267,19 @@ class RakuAST::Var::Attribute
     method PERFORM-BEGIN(RakuAST::Resolver $resolver, RakuAST::IMPL::QASTContext $context) {
         my $package := $resolver.find-attach-target('package');
         if $package {
-            # We can't check attributes exist until we compose the
-            # package, since they may come from roles. Thus we need to
-            # attach them to the package.
-            $package.ATTACH-ATTRIBUTE-USAGE(self);
-            nqp::bindattr(self, RakuAST::Var::Attribute, '$!package', $package);
-        }
-        else {
-            # TODO check-time error
+            if nqp::istype($package, RakuAST::Package::Attachable) {
+                # We can't check attributes exist until we compose the
+                # package, since they may come from roles. Thus we need to
+                # attach them to the package.
+                $package.ATTACH-ATTRIBUTE-USAGE(self);
+                nqp::bindattr(self, RakuAST::Var::Attribute, '$!package', $package);
+            }
+            else {
+                self.add-sorry:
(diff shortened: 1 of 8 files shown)


src/Raku/ast/variable-access.rakumod
```diff
@@ -267,14 +267,19 @@ class RakuAST::Var::Attribute
...
         if $package {
-            # We can't check attributes exist until we compose the
-            # package, since they may come from roles. Thus we need to
-            # attach them to the package.
-            $package.ATTACH-ATTRIBUTE-USAGE(self);
-            nqp::bindattr(self, RakuAST::Var::Attribute, '$!package', $package);
-        }
-        else {
-            # TODO check-time error
+            if nqp::istype($package, RakuAST::Package::Attachable) {
```
lizmat, line 270: I'm not understanding why this `if` is needed.  Could you elaborate?

ugexe: > RakuAST::Var::Attribute::PERFORM-BEGIN raises X::Attribute::Package when the attach target is not Attachable

PR #6242, opened 2026-06-06
Fix Array.shift on holes in arrays

lizmat:
Unlike the needed Array.pop fix in 415fa4b069 to prevent a segfault, Array.shift was already guarded against returning an nqp::null.

However, it was set to return Nil, rather than the default value of the array.  This commit changes that to be more in line with Array.pop.

However, this breaks one spectest in t/spec/S32-array/shift.t which literally expects Nil to be returned.  However the test does not actually want to test that, just that it didn't return a Failure (according to the comment with the test).

It feels to me that the test should be changed, and that Array.shift on a hole in an Array, should return the default value, just like Array.pop does now.

### DIFF
--- a/src/core.c/Array.rakumod
+++ b/src/core.c/Array.rakumod
@@ -1006,7 +1006,7 @@ my class Array { # declared in BOOTSTRAP
           && nqp::elems($reified)
           ?? nqp::ifnull(  # handle holes
                nqp::shift($reified),
-               Nil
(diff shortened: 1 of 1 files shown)


lizmat:
Failed CI tests are caused by the changed behaviour of Array.shift for holes.

lizmat:
Test adapted with https://github.com/Raku/roast/commit/b87c37fc1c

PR #6243, opened 2026-06-06
 RakuAST: route string EVAL through IMPL-FIXUP and wrap nested EVAL output

ugexe:
The string-EVAL branch of ForeignCode skipped IMPL-FIXUP-COMPILED-CODEREFS that the AST-EVAL branch runs. A Sub returned from `BEGIN { &x = EVAL 'sub () { 42 }' }` had a `$!do` coderef whose static_code was not in any SC. Outer-module precomp serialization failed with `missing static code ref for closure`. Route the RakuAST frontend's string EVAL through `:target('ast'), :compunit_ok(1)` and finish through the same helper as the AST-EVAL path: IMPL-TO-QAST-COMP-UNIT, a compile from the qast stage, then IMPL-FIXUP-COMPILED-CODEREFS. The helper binds $*COMPILING_CORE_SETTING to 0, which the bypassed qast stage method otherwise provides, and runs the comp-unit's cleanup from a LEAVE phaser so a failure in a later stage cannot skip it.

When the EVAL'd Sub's outer chain reached BEGIN's frame and BEGIN held non-serializable state (a Lock, a ConditionVariable), the serializer walked the chain into BEGIN's lexpad and tripped on it. RakuAST::CompUnit.IMPL-TO-QAST-COMP-UNIT now interposes an anonymous wrapper QAST::Block when the EVAL is nested. The wrapper declares no symbols and carries the DYN_COMP_WRAPPER annotation, the same shape World.compile_in_context gives the legacy frontend, so lexical lookups from the EVAL'd code stay late-bound and reach the caller's `$_`, `$/`, and `$?PACKAGE` through the forced outer context. It sets the inner mainline's blocktype to `immediate_static` so it auto-invokes and passes its return value through, and has no Code object or SC entry. When the outer serializer walks body.outer from an EVAL'd Sub up to the wrapper frame, closure_to_static_code_ref returns NULL and the walk terminates instead of reaching BEGIN.

### DIFF
--- a/src/Raku/ast/compunit.rakumod
+++ b/src/Raku/ast/compunit.rakumod
@@ -646,10 +646,24 @@ class RakuAST::CompUnit
             $top-level.set_children([$run-main]);
         }
 
+        # Nested EVAL: wrap the inner mainline in an anonymous QAST::Block.
+        # With no Code object and no SC entry, the wrapper's static_code
+        # stays NULL, so the parent's serializer terminates body.outer
+        # walks here instead of reaching the enclosing frame's lexpad and
+        # any non-serializable state it holds (Locks, ConditionVariables,
+        # tap closures). The wrapper declares no symbols of its own and is
+        # marked DYN_COMP_WRAPPER, so lexical lookups from the EVAL'd code
+        # stay late-bound and resolve through the forced outer context to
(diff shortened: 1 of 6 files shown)


ugexe:
This branch requires https://github.com/Raku/nqp/pull/852 hence the RakuAST test failure. After we merge that and bump nqp I will rebase this branch so the tests can be rerun green.

PR #6247, opened 2026-06-10
Add an optimize phase to the RakuAST frontend

ugexe:
The RakuAST frontend goes straight from check to QAST generation, so unlike the legacy frontend there is no place to apply any optimizations (the roadmap calls this out as [step 6](https://github.com/rakudo/rakudo/blob/e78c5a7e672786e5ae658a50cda8515798b04ea9/src/Raku/ast/README.md?plain=1#L29)). This adds that phase along with a few initial optimizations to flesh it out, and is intended to be the starting point the RakuAST optimizer grows from.

The phase is an AST-to-AST pass that runs after check and before QAST generation. It walks the tree post-order and offers every child a node visits to a single dispatch point where the optimizations are registered, replacing the child in place when a rewrite applies. The replacement mechanism locates the slot holding a child via attribute introspection, so coverage is structural: any expression position a node exposes through `visit-children` gets all of the optimizations without needing code of its own. The phase runs for parsed code and for the AST form of `EVAL`, and `--optimize=off` disables it (which also makes `--target=ast` show the tree before any rewrites).

The initial optimizations:

* constant folding: a pure operator (infix, prefix, or list-associative infix) applied to constant operands is evaluated at compile time and replaced with a literal holding the result, e.g. `my $x = 2 ** 10` compiles as `my $x = 1024`
* constant ternary collapse: `my $x = 1 ?? "a" !! "b"` compiles as `my $x = "a"`
* constant short-circuit collapse: `my $x = False || 42` compiles as `my $x = 42`

Each rewrite declines unless it is certain the result preserves the program. Folding only evaluates routines that carry the `is pure` trait (the same signal the legacy optimizer uses) so an unmarked user-defined operator is never run at compile time. Operands must be constants rather than variables, chained comparisons and operators carrying adverbs are left for runtime, and a collapse never drops a branch holding a declaration. Everything also declines during early bootstrap, so the CORE settings build the same way they always have.

On some quick benchmarks the walk costs about 2% at compile time and nothing at runtime. Below are some naive benchmarks with and without these optimizations:

```
# constant folding, 2.6x (most of that is the ** call, which spesh does not eliminate)
time RAKUDO_RAKUAST=1 raku                -e 'my $s = 0; for ^10_000_000 { $s = $s + (2 ** 10 + 7 * 3 - 4) }; say $s'  # ~3.1s
time RAKUDO_RAKUAST=1 raku --optimize=off -e 'my $s = 0; for ^10_000_000 { $s = $s + (2 ** 10 + 7 * 3 - 4) }; say $s'  # ~8.1s

# ternary collapse, ~5%
time RAKUDO_RAKUAST=1 raku                -e 'my $s = 0; for ^50_000_000 { $s = $s + (0 ?? 999 !! 1) }; say $s'        # ~13.3s
time RAKUDO_RAKUAST=1 raku --optimize=off -e 'my $s = 0; for ^50_000_000 { $s = $s + (0 ?? 999 !! 1) }; say $s'        # ~13.9s

# short-circuit collapse, ~18%
time RAKUDO_RAKUAST=1 raku                -e 'my $s = 0; for ^50_000_000 { $s = $s + (False || 1) }; say $s'           # ~13.1s
time RAKUDO_RAKUAST=1 raku --optimize=off -e 'my $s = 0; for ^50_000_000 { $s = $s + (False || 1) }; say $s'           # ~16.0s
```

The two collapses bring their loops down to bare loop overhead. Their numbers are smaller than folding's because spesh already handles constant branches well at runtime.

There is plenty of room to grow from here. Meta operators, defined-or, and folding pure calls the way the legacy optimizer does are potential next steps. Probably easiest to review commit by commit.

### DIFF
--- a/src/Raku/ast/base.rakumod
+++ b/src/Raku/ast/base.rakumod
@@ -672,7 +672,7 @@ class RakuAST::Node {
     method IMPL-OPTIMIZE-EXPRESSION(RakuAST::Resolver $resolver, Mu $expr) {
         return $expr unless nqp::isconcrete($expr);
 
-        my $result := $expr;
+        my $result := self.IMPL-FOLD-CONSTANT($resolver, $expr);
 
         # A replacement stands where the original stood, so it must carry the
         # original's sunk state for any sink-sensitive code generation.
@@ -682,6 +682,20 @@ class RakuAST::Node {
             $result.mark-sunk();
         }
         $result
+    }
+
+    # Whether an expression may serve as an operand for compile-time
+    # evaluation: a literal, an enumeration value such as True, or a quoted
+    # string whose value is known at compile time, which is how a plain string
+    # literal parses. Any of these holds no variable reference, so replacing
+    # it cannot orphan a lowered lexical.
+    method IMPL-FOLDABLE-OPERAND(Mu $operand) {
+        nqp::isconcrete($operand)
+          && (nqp::istype($operand, RakuAST::Literal)
+               || (nqp::istype($operand, RakuAST::QuotedString)
+                    || nqp::istype($operand, RakuAST::Term::Enum))
+                  && $operand.has-compile-time-value)
+            ?? 1 !! 0
     }
 
     # Whether a branch can be removed from the tree. The running program would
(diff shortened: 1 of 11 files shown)


lizmat:
Feels to me that this should be split into two or more PRs:

- one that adds the basic optimizer stage
- one or more that add specific optimizations

That should make bisecting issue easier at a later time.

lizmat:
BTW: @ugexe Great to see you started on this!

ab5tract:
I don't think there is any need for multiple PRs because the branch is already boiled down to 5 distinct and documented commits.

lizmat:
Ok, fair enough.  I tend to think squash too much  :-)

src/Raku/ast/base.rakumod
```diff
@@ -703,6 +717,164 @@ class RakuAST::Node {
         $droppable
     }
 
+    # Constant folding. Given a child expression, if it is a pure operator
+    # applied to constant operands, evaluate it now and return a
+    # RakuAST::Literal holding the result. Otherwise the expression is
+    # returned unchanged. Operands must satisfy IMPL-FOLDABLE-OPERAND (never
+    # variables, even constant ones) so folding never removes a variable
```
MasterDuke17, line 724: I don’t understand this section. If a variable is removed, there won’t be anything to lower to a lexical, right? So nothing would be “orphaned”. Or is this saying it would result in a useless declaration (until we get dead code elimination)?

src/Raku/ast/base.rakumod
```diff
@@ -703,6 +717,164 @@ class RakuAST::Node {
...
+                && $infix.is-resolved
+                && self.IMPL-PURE-ROUTINE($infix)
+                && !$infix.properties.chain
+                && !nqp::isconcrete($expr.args.arg-at-pos(2))
+                && self.IMPL-FOLDABLE-OPERAND($expr.left)
+                && self.IMPL-FOLDABLE-OPERAND($expr.right);
+        }
+        elsif nqp::istype($expr, RakuAST::ApplyPrefix) {
+            my $prefix := $expr.prefix;
+            # An adverb on the operator (zpre 4 :x(5)) is a named argument the
```
MasterDuke17, line 757: What interpreter is being talked about here? The compile-time one? My first thought is that adverbs that get “consumed” by the dispatch (i.e., disambiguate it rather than silently just passed along) shouldn’t disallow constant folding, but I haven’t yet thought about how to tell if/when that’s the case.

src/Raku/ast/base.rakumod
```diff
@@ -703,6 +717,164 @@ class RakuAST::Node {
...
+        my $Failure := self.IMPL-OPTIMIZE-SETTING-TYPE($resolver, 'Failure');
+        if !nqp::isnull($Failure) && nqp::istype($value, $Failure) {
+            $value.Bool;
+            return 0;
+        }
+
+        # An Iterable result is potentially lazy or one-shot, so consuming
+        # it at compile time would not preserve the program.
+        my $Iterable := self.IMPL-OPTIMIZE-SETTING-TYPE($resolver, 'Iterable');
+        return 0 if !nqp::isnull($Iterable) && nqp::istype($value, $Iterable);
```
MasterDuke17, line 855: Could this check for laziness? Or something more specific so e.g., Arrays could be constant folded even if Seqs can’t.

src/Raku/ast/base.rakumod
```diff
@@ -684,6 +688,44 @@ class RakuAST::Node {
...
+        my $keep := $truth ?? $expr.then !! $expr.else;
+        my $drop := $truth ?? $expr.else !! $expr.then;
+        self.IMPL-DROPPABLE($drop) ?? $keep !! $expr
+    }
+
+    # Raku truth value of a node, or -1 when it cannot be determined safely.
+    # Folding only ever evaluates pure operators on foldable operands, while
+    # truthiness has to consider any constant, so this is deliberately narrow:
+    # the value must be a concrete Cool or Bool, whose .Bool is pure and
+    # well-defined. Type objects (not concrete) are declined, since a type used
```
MasterDuke17, line 707: If the type is from the setting, would it be the same at runtime?

src/Raku/ast/base.rakumod
ugexe: Yeah, the wording could be better. The actual reason variables are excluded is that a variable read isn't a compile time value without use-def analysis we don't attempt (legacy has the same rule via its all-args-known check). The lowering bit is describing the invariant that falls out of that, i.e. folded subtrees never contain variable references, so a later pass that maps declarations to their uses (like the lexical to local lowering the legacy optimizer does) never has to account for this phase editing them, even though today you'd just end up with an unused declaration like you said.

src/Raku/ast/base.rakumod
ugexe: It's the RakuAST compile time interpreter (IMPL-INTERPRET), which is what folding uses to evaluate the application. It only assembles the positional operands, e.g. ApplyPrefix interprets as `$op($operand)` without ever looking at the colonpairs, so evaluating an adverbed operator would silently drop the adverb rather than dispatch on it. I don't think you'd even need to tell when an adverb is consumed by dispatch: if its value is also compile time known you could just forward it in the call and dispatch would consume it the same way it does at runtime, with `is pure` covering the rest. Declining was just the cheaper call since the interpreter doesn't forward adverbs today and pure operators taking adverbs are rare, but teaching IMPL-INTERPRET to pass them would lift the restriction naturally.

src/Raku/ast/base.rakumod
ugexe: Laziness is only one of the hazards there, so a naive is-lazy alone wouldn't cover it. A folded result is serialized with the compilation unit as a single object, so every evaluation of that expression would share it: for an Array that means mutations stick and alias across calls, which actually makes Arrays seem like the clearest unsound case rather than the foldable one, and a Seq is one shot even when it isn't lazy. The safe carve outs would be immutable value types that happen to do Iterable, e.g. Range, which the legacy optimizer does fold (`my $r = 1 .. 10` compiles to a WVal(Range) constant there).

This containing IMPL-FOLDABLE-VALUE method was created for these types of refinements though, each restriction or allowance with its own justification, so loosening this for specific types later is straightforward.

src/Raku/ast/base.rakumod
ugexe: For a setting type yeah it would be the very same object at runtime since the condition serializes as a reference into the setting, so my comment is weaker than it should be. The reason type objects stay declined is what .Bool dispatches to: a concrete Cool or Bool hits the CORE candidates, but a type object's truthiness goes through whatever Bool the class declares, and a `method Bool` without a `:D` constraint applies to the type object too (`class C { method Bool() { True } }; C ?? "t" !! "f"` gives "t"). Evaluating that at compile time would mean running user code that never claimed to be pure, the same reason folding refuses to run unmarked user operators. My thinking was a bare type object as a constant condition is degenerate code anyway, so declining costs nothing.

PR #6249, opened 2026-06-10
Another solution to fixing my %h is Set = 1,2,3

### LINKED ISSUE #6246: Syntax for `my %h is QuantHash` dirties the sentinel `set()`/ `bag()` / `mix()` values
The `Great QuantHash Refactor` introduced an issue with some strange 'action at a distance effects'.

In its simplest form, it can be described as:
```raku
my %h is Set = "a";
dd %h;   # set()
dd %h.keys;  # ("a",).Seq
dd set().keys;  # ("a",).Seq
```
Note that `%h` reports itself as the empty set, even though it has keys in it.  But what's worse, the empty set (`set()`) is no longer empty.

This problem was sadly **not** picked up by spectests  :-(

The problem is caused by the fact that `my %h is Set = "a"` codegens to:
```raku
(my %h := Set.new).STORE("a", :INITIALIZE);
```

During the GQR the empty `Set` / `Bag` / `Mix` became sentinels, with the idea that a lot of memory could be saved that way.  And that it would be easier to recognize the empty `Set` / `Bag` / `Mix` (which allows for a lot of shortcuts in set operations).  So `Set.new` without any arguments produces the empty `Set` sentinel, just as `set()`.

What hadn't been realized is the way that `my %h is Set` (and `Bag` and `Mix`) are being codegenned.  And thus causing action at a distance.

And since `MAIN` processing of named arguments uses this syntax, one could see named command line arguments suddenly appear in set operations (which lead to figuring out what was going on).

lizmat:
This effectively reverts e78c5a7e672786e5ae65 and adds logic to the codegenning to make sure "is Set", "is Bag" and "is Mix" are codegenned with `nqp::create(type)` rather then `type.new`.  This ensures that it will always create a new instance.  Then later the special versions of Set|Bag|Mix.STORE can return the sentinel safely after making sure there are indeed no elements created with the .STORE logic.

The RakuAST version is a bit iffy as I've been unable to come up with a way to check the actual types at that moment in QAST generation.  So it checks the name for now.  Fixes welcome!

Fixes #6246

### DIFF
--- a/src/Raku/ast/variable-declaration.rakumod
+++ b/src/Raku/ast/variable-declaration.rakumod
@@ -1365,12 +1365,16 @@ class RakuAST::VarDeclaration::Simple
                     :value($container)
                 );
                 if self.IMPL-HAS-EXPLICIT-CONTAINER-BASE-TYPE {
-                    my $value := self.IMPL-CONTAINER-TYPE($of);
-                    $context.ensure-sc($value);
-                    $qast := QAST::Op.new( :op('bind'), $qast, QAST::Op.new(
-                        :op('callmethod'), :name('new'),
-                        QAST::WVal.new( :$value )
-                    ) );
+                    my $class := self.IMPL-CONTAINER-TYPE($of);
+                    my $wval  := QAST::WVal.new(:value($class));
+                    $context.ensure-sc($class);
+
+                    my str $name := $class.HOW.name($class); # XXX needs lookup solution
+                    $qast := QAST::Op.new( :op('bind'), $qast,
+                      $name eq 'Set' || $name eq 'Bag' || $name eq 'Mix'
+                        ?? QAST::Op.new(:op<create>, $wval)
+                        !! QAST::Op.new(:op<callmethod>, :name<new>, $wval)
+                    );
                 }
                 $qast
             }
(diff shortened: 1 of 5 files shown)


src/Raku/ast/variable-declaration.rakumod
```diff
@@ -1365,12 +1365,16 @@ class RakuAST::VarDeclaration::Simple
...
-                    $context.ensure-sc($value);
-                    $qast := QAST::Op.new( :op('bind'), $qast, QAST::Op.new(
-                        :op('callmethod'), :name('new'),
-                        QAST::WVal.new( :$value )
-                    ) );
+                    my $class := self.IMPL-CONTAINER-TYPE($of);
+                    my $wval  := QAST::WVal.new(:value($class));
+                    $context.ensure-sc($class);
+
+                    my str $name := $class.HOW.name($class); # XXX needs lookup solution
```
lizmat, line 1372: @ugexe any idea how to actually check for `Set`, `Bag` or `Mix` here?

ugexe: I had not found a good way, no. You can see the hacky solution I had resorted to before I moved on [here](https://github.com/ugexe/rakudo/commit/9557dae0095822d598836bc6998c3f705d829aaf#diff-88ceef3931d479805a0b41e5c447907796e966e2451c4876bfdd8d8ee03bec52)

lizmat: Ok, then I'm going to keep this for now.

PR #6250, opened 2026-06-10
RakuAST: streamline sigil -> role mapping

lizmat:
Make the lookup hash a constant, and only look up the sigil once.

Sadly this is too early to make the hash contain RakuAST::Name objects, so we just keep the name and do the instantiation of RakuAST::Name when needed.


lizmat:
Already merged with #6251

PR #6258, opened 2026-06-11
RakuAST: streamline call related RakuAST classes

lizmat:
Mostly code esthetics to improve readability, and some additional comments in places where I was sure I grokked everything.

Slightly changed the interface of the Arglist.IMPL-ADD-QAST-ARGS method to have it return the QAST that it was given.  This allowed for quite significant simplification in most xxxQAST methods.

Also simplified the various forms of Arglist creation by adding a `CREATE` method that does the basic things, and call that from the instantiators
`new`, `from-comma-list` and `from-invocant-list`.


ab5tract:
Looks good to me 👍🏻

Thanks for this cleanup Liz

PR #6261, opened 2026-06-12
Key precompilation stores per compiler frontend

ugexe:
The compiler id was a constant baked while CORE.setting compiled, a sha1 over the setting handle and the source digest, so both frontends of one build shared it, and with it every precompilation store directory. Bytecode precompiled by one frontend then got loaded by processes compiling with the other. The visible damage ranged from "New type Stash for ContainerDescriptor::Untyped is not a mixin type" at startup to subtler staleness: any module that bakes frontend-dependent state into a compile-time constant, like the QAST test helper choosing a compilation target by probing for the optimize stage, serves the wrong answer to the other frontend.

The compiler entry point now binds the active frontend as the COMPILER-FRONTEND hll symbol, and the compiler id is a sha1 over the build stamp and that symbol, computed on first use since the frontend is only known at run time. The constant snapshot of the id in CompUnit::PrecompilationRepository becomes lazy for the same reason. Each frontend thereby precompiles into and loads from its own store subtree; nothing one frontend produces is visible to the other. A frontend that did not build the installation pays a one-time recompilation of distributed modules instead of loading precomp that was never compiled for it.

### DIFF
--- a/src/core.c/CompUnit/PrecompilationRepository.rakumod
+++ b/src/core.c/CompUnit/PrecompilationRepository.rakumod
@@ -42,8 +42,14 @@ class CompUnit::PrecompilationRepository::Default
     my $resolved-lock := Lock.new;
     my $first-repo-id;
 
-    my constant $compiler-id =
-      CompUnit::PrecompilationId.new-without-check(Compiler.id);
+    # The compiler id incorporates the active frontend, which is only
+    # known at run time, so it cannot be captured while the setting is
+    # being compiled.
+    my $compiler-id;
+    sub compiler-id() {
+        $compiler-id //=
+          CompUnit::PrecompilationId.new-without-check(Compiler.id)
+    }
 
     # If $loaded{$key} has a Promise in it, await it. If it's broken, try
     # again, return the result if it's kept. If there's nothing there at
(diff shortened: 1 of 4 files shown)


src/main.nqp
```diff
@@ -16,12 +16,14 @@ nqp::bindhllsym('default', 'SysConfig', Perl6::SysConfig.new(%rakudo-build-confi
 my $comp := Perl6::Compiler.new();
 $comp.language('Raku');
 if +nqp::getenvhash()<RAKUDO_RAKUAST> {
+    nqp::bindhllsym('Raku', 'COMPILER-FRONTEND', 'rakuast');
```
lizmat, line 19: Currently `Compiler.legacy` returns `True` depending on `$*LANG.^name eq 'Perl6::Grammar'`.  Perhaps that should be changed as well?

src/core.c/CompUnit/PrecompilationRepository.rakumod
```diff
@@ -42,8 +42,14 @@ class CompUnit::PrecompilationRepository::Default
...
     my $first-repo-id;
 
-    my constant $compiler-id =
-      CompUnit::PrecompilationId.new-without-check(Compiler.id);
+    # The compiler id incorporates the active frontend, which is only
+    # known at run time, so it cannot be captured while the setting is
+    # being compiled.
+    my $compiler-id;
+    sub compiler-id() {
+        $compiler-id //=
```
lizmat, line 50: Why not assign this in the mainline at the definition of `$compiler-id` ?

lizmat:
Closing in lieu of PR6302

PR #6270, opened 2026-06-14
Allow "unit sub foo;" in 6.e and higher

lizmat:
Turns out the idiom for "unit sub foo is export" is actually quite useful in the case of Air, where generally the logic for each page (part) of a website is stored in a separate Raku .rakumod file.

Since this was technically already possible except for a check on the name of the subroutine (must be "MAIN"), this felt like a straightforward fix to be applied (add feature by removing code!).

However, this showed up quite a few spectest errors.  But even worse, it turned out that the "unit" scope wasn't even checked for!  This means that "sub MAIN($a,$b);" is as valid as "unit sub MAIN($a,$b)".

It felt that adding a check for the "unit" scope now would potentially be a breaking change.  So in order to not break anything in the ecosystem / user code, this approach was not taken.

So this commit really does a number of things:

1. Add typed error X::UnitScope::MustHaveUnit
2. Use this typed error class for package statements lacking "unit" instead of the untyped panic that was there already
3. In 6.e and higher, check for the existence of the "unit" scope with "sub" and throw X::UnitScope::MustHaveUnit error if it is missing
4. In 6.e and higher, allow *any* sub to have "unit" scope, not just MAIN

### DIFF
--- a/src/Perl6/Grammar.nqp
+++ b/src/Perl6/Grammar.nqp
@@ -2209,18 +2209,18 @@ grammar Perl6::Grammar is HLL::Grammar does STD {
                 [
                 || <?{ $*begin_compunit }>
                     {
-                        unless $longname {
-                            $/.panic("Compilation unit cannot be anonymous");
-                        }
-                        unless $*SCOPE eq 'unit' {
-                            if $*PKGDECL eq 'package' {
-                                $/.panic('This appears to be Perl code. If you intended it to be Raku code, please use a Raku style declaration like "unit package Foo;" or "unit module Foo;", or use the block form instead of the semicolon form.');
-                            }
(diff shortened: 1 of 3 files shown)


ab5tract:
My only caveat is that I would like this to support namespacing, à la `unit sub SomePreferModular::foo() {}`.

Unless supporting that is more difficult than it's worth.

lizmat:
What would be the point?  It won't be exported as such?

ab5tract:
Why can't it be installed under the package name provided?

Not all of us want to use comp units directly. You are thinking only about export. I want to support namespacing. 

Why can't we have both?

ab5tract:
Thinking about it a bit further, the containing compunit would still be the thing referenced when use'ing, which does make my request irrelevant in the end.

Please consider my proposal formally withdrawn :)

PR #6282, opened 2026-06-17
RakuAST: give all package installers is-stub and defuse-stub

ugexe:
A compound named `enum` or `subset` such as `enum Foo::Bar` vivifies a stub package for the `Foo` prefix and records the installing node as its owner. Declaring `Foo` afterwards as a class or role merges into that stub and calls `is-stub` and `defuse-stub` on the owner. `RakuAST::Type::Enum` had no `defuse-stub` and `RakuAST::Type::Subset` had neither, so the merge died with a "No such method" error (seen while installing `TAP`).

Only a `RakuAST::Package` can be a forward stub, so move the defaults onto `RakuAST::PackageInstaller` where `Package` overrides them and the other installers inherit.


ab5tract:
Looks much cleaner now!

PR #6299, opened 2026-06-19
RakuAST: Interpolate a non string constant into a regex

ugexe:
A variable interpolated into a regex took a fast path when it resolved to a constant, treating the value as a literal string via `nqp::unbox_s`. That only works for a string. A constant holding a list (`my constant @a = <foo bar>; / @a /`) or any other non-string died with "cannot unbox to a native string".

Take the literal fast path only when the value is a Str. Otherwise fall through to the runtime interpolation path, which already handles a list as an alternation, the same as a non constant variable.

Contact::Name interpolates a constant array of name prefixes into a token.

### DIFF
--- a/src/Raku/ast/regex.rakumod
+++ b/src/Raku/ast/regex.rakumod
@@ -1048,9 +1048,15 @@ class RakuAST::Regex::Interpolation
         if nqp::istype($!var, RakuAST::Lookup) && $!var.is-resolved {
             my $resolution := $!var.resolution;
             if nqp::istype($resolution, RakuAST::VarDeclaration::Constant) {
-                return self.IMPL-APPLY-LITERAL-MODS:
-                    QAST::Regex.new( :rxtype<literal>, nqp::unbox_s($resolution.compile-time-value) ),
-                    %mods
+                my $value := $resolution.compile-time-value;
+                # Only a string constant becomes a regex literal. A constant
+                # that is not a string (e.g. a list) interpolates at runtime
+                # via the slow path below.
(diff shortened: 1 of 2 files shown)


src/Raku/ast/regex.rakumod
```diff
@@ -1048,9 +1048,15 @@ class RakuAST::Regex::Interpolation
...
             my $resolution := $!var.resolution;
             if nqp::istype($resolution, RakuAST::VarDeclaration::Constant) {
-                return self.IMPL-APPLY-LITERAL-MODS:
-                    QAST::Regex.new( :rxtype<literal>, nqp::unbox_s($resolution.compile-time-value) ),
-                    %mods
+                my $value := $resolution.compile-time-value;
+                # Only a string constant becomes a regex literal. A constant
+                # that is not a string (e.g. a list) interpolates at runtime
+                # via the slow path below.
+                if nqp::istype($value, Str) {
```
lizmat, line 1055: Why not call `.Str` on the compile-time value  here and at compile time?  If that fails, it at least fails then, rather than at runtime.

ugexe: Maybe I'm misunderstanding, but the point here is that a list constant like `my constant @a = <foo bar>; / @a /` should interpolate as an alternation the same as a not constant `@a` does. `.Str` would turn it into the literal string `"foo bar"` instead, which never fails but matches the wrong thing

lizmat: Good point, yup!

PR #6301, opened 2026-06-19
Make .return check the routine it actually returns from

ugexe:
.return validates the enclosing routine's return constraint before throwing (added in #6086 to catch a literal constraint the compiler cannot). It found the routine by walking the dynamic call stack, but .return unwinds to the lexically enclosing routine. The two disagree when an internal routine is interposed: a sunk `for` runs its body inside sink-all (--> IterationEnd), so `for @x { .return if ... }` died with "Cannot return ... in the signature of method sink-all". An explicit `.map({ .return }).sink` failed the same way on any frontend.

Walk the lexical outer chain instead, the routine throwpayloadlexcaller reaches. The #6086 literal-constraint check is preserved.

### DIFF
--- a/src/core.c/Mu.rakumod
+++ b/src/core.c/Mu.rakumod
@@ -71,17 +71,18 @@ my class Mu { # declared in BOOTSTRAP
         take self;
     }
 
-    # Check signature of caller Routine for definite value being returned
+    # Check signature of lexically enclosing Routine for definite value being returned
     my sub check-signature(Mu $value, str $method) {
-        my int $index = 2;
+        my $ctx := nqp::ctxcallerskipthunks(nqp::ctxcallerskipthunks(nqp::ctx()));
+        my $callable;
         nqp::until(
-          nqp::isnull(my $cf := callframe($index))
-            || nqp::istype((my $callable := $cf.code),Routine),
-          ++$index
+          nqp::isnull($ctx)
+            || nqp::istype(($callable := nqp::getcodeobj(nqp::ctxcode($ctx))),Routine),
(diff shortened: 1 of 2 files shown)


src/core.c/Mu.rakumod
```diff
@@ -71,17 +71,18 @@ my class Mu { # declared in BOOTSTRAP
...
-            || nqp::istype((my $callable := $cf.code),Routine),
-          ++$index
+          nqp::isnull($ctx)
+            || nqp::istype(($callable := nqp::getcodeobj(nqp::ctxcode($ctx))),Routine),
+          $ctx := nqp::ctxouterskipthunks($ctx)
         );
 
         die "Cannot return $value.raku() with $method when return value $_.raku() is already specified in the signature of $callable.raku.subst(/ "(" .* /)".naive-word-wrapper
-          with $callable.signature.returns;
+          with (nqp::isnull($ctx) ?? Nil !! $callable.signature.returns);
```
lizmat, line 85: This feels bogus: `with Nil` will never fire

ugexe: That is the point

lizmat: Then I suggest:

`with nqp::not_i(nqp::isnull($ctx)) && $callable.signature.returns;`

ugexe: That isn't equivalent
```
$ raku -e 'use nqp; with nqp::not_i(nqp::isnull(nqp::null)) && 2 { say $_ }'
0
```

PR #6302, opened 2026-06-23
Slightly different approach to PR6261

lizmat:
- Stole the test file = Stole the changes to main.nqp
- Put initializations in the mainline, instead of hiding it in a subroutine and assigning on each access.
- Adapted comments accordingly

### DIFF
--- a/src/core.c/CompUnit/PrecompilationRepository.rakumod
+++ b/src/core.c/CompUnit/PrecompilationRepository.rakumod
@@ -42,8 +42,9 @@ class CompUnit::PrecompilationRepository::Default
     my $resolved-lock := Lock.new;
     my $first-repo-id;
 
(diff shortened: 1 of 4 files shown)


lizmat:
@ugexe I think this is a better solution than https://github.com/rakudo/rakudo/pull/6261

ugexe (approved):

ugexe:
In the future it is more useful to use the commit message to describe what/why the change does so someone looking at the blame can understand what a given change was for. Listing things you took from another PR isn't very useful to anyone looking at that commit message... things like that are better communicated via a separate comment on the PR or in the PR description only.

lizmat:
Good point.  Also stole most of the original commit message when squashing the commit.

PR #6306, opened 2026-06-24
RakuAST: build the CORE:: PseudoStash at runtime, not compile time

### LINKED ISSUE #6304: RakuAST: referencing a PseudoStash in a module at compile time produces serialization error
Found in the `Prompt` module, golfed to:
```
# in lib/A.rakumod:
BEGIN &CORE::prompt;
```
then:
```
 % RAKUDO_RAKUAST=1 raku -Ilib -MA -e ''
===SORRY!===
Missing serialize REPR function for REPR MVMContext (BOOTContext)
```
A workaround is simple: don't reference it at compile time, but at runtime.

This is a shorter golf from:
```raku
role A { method a($b) { &CORE::prompt($b) } }
```

ugexe:
Referencing a CORE:: symbol at compile time in a module (for example `&CORE::prompt` in a role method) died "Missing serialize REPR function for REPR MVMContext (BOOTContext)". Dynamically compiled code built the CORE PseudoStash from the setting context and inserted it into the QAST as a literal value. That is fine when the frame is thrown away, but a role body is dynamically compiled and then serialized with its module, and the setting context cannot be serialized.

The special case dates from when the setting was not an outer frame at BEGIN time, so a runtime `PseudoStash.new` could not find CORE. That no longer holds: the runtime lookup resolves CORE:: at BEGIN time too. Drop the special case and always construct the PseudoStash at runtime, which never captures the context.

Fixes #6304.

### DIFF
--- a/src/Raku/ast/name.rakumod
+++ b/src/Raku/ast/name.rakumod
@@ -326,23 +326,11 @@ class RakuAST::Name
         my @parts := self.IMPL-LOOKUP-PARTS;
         my $final := @parts[nqp::elems(@parts) - 1];
         my $PseudoStash-lookup := self.IMPL-UNWRAP-LIST(self.get-implicit-lookups)[1];
-        my $result;
-        #TODO Look at Perl6::World::symbol_lookup. Has a separate implementation for compiling the setting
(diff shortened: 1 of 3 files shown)


lizmat:
Fixing bugs by removing code, the best way!

PR #6317, opened 2026-06-25
Allow resizing the pty of a  non-started Proc::Async

patrickbkr:
Since the the dimensions are passed to the VM when starting, this is safe to do.

### DIFF
--- a/src/core.c/Proc/Async.rakumod
+++ b/src/core.c/Proc/Async.rakumod
@@ -135,7 +135,9 @@ my class Proc::Async {
     method resize-pty(Int :$cols, Int :$rows) {
         $!pty-cols := $cols;
         $!pty-rows := $rows;
-        nqp::syscall('pty-resize', $!process_handle, nqp::unbox_i($!pty-cols), nqp::unbox_i($!pty-rows));
+        if $!started {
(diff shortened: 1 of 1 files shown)


lizmat:
Could you elaborate on what this PR fixes, rather than just saying it's ok?   :-)

patrickbkr:
With this PR, calling `$proc.resize-pty(...)` on a `Proc::Async` object that has not been started yet, succeeds (and does the right thing) instead of dying with "Argument 0 to MoarVM syscall 'pty-resize' has repr P6opaque, but should be MVMOSHandle". It errors because `$!process-handle` is `Any` until the proc is started.

lizmat:
Hmmm... the `if $!started {` sorta indicates the opposite, does it not?  Reading that code snippet gives the impression it's only allowed if it *has* started?  Isn't there a `!` missing?

patrickbkr:
> Hmmm... the `if $!started {` sorta indicates the opposite, does it not?

Correct. That syscall will only succeed if the proc has started. The `if $!started {` ensures, it is only called in that case. The syscall is skipped when the proc hasn't started yet.

So no, the conditional is fine.

lizmat:
OK, rescheduled some failing tests to run again.  If they're ok, (or clearly some external reason for failure) I'll merge this one unless @Coke disagrees.

PR #6328, opened 2026-06-27
RakuAST: Decontainerize a sunk block call result

ugexe:
A routine decontainerizes its return value, stripping the container off a containerized return. An `is rw` routine skips that step, so a returned container passes through unchanged. A bare block does neither. So when a block held in a variable is called via postfix `()` and the result is sunk, a returned Scalar was sunk as-is, and sinking a container is a no-op, so a returned Failure was never thrown. `dies-ok { $failure }` then saw no exception, as in Math::Angle's tests.

Decontainerize a sunk postfix call's result when the callee is a block, so the value is reached and a Failure throws. A routine callee has a `rw` method and is left alone, so its own return handling stands. A bind or other non-sink use is untouched, so a block call that returns a container still preserves it.

### DIFF
--- a/src/Raku/ast/statements.rakumod
+++ b/src/Raku/ast/statements.rakumod
@@ -684,7 +684,36 @@ class RakuAST::Statement::Expression
         else {
             $qast := $!expression.IMPL-TO-QAST($context);
             if self.sunk && $!expression.needs-sink-call {
-                $qast := QAST::Op.new( :op('p6sink'), $qast );
+                # A routine strips the container off a containerized return;
+                # an `is rw` one skips that strip; a bare block does neither.
+                # So a sunk postfix `()` call of a block must decont its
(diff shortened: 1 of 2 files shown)


lizmat:
> A routine decontainerizes its own return value (and an is rw one deliberately keeps its container);

Text nit: An "is rw" does not decontainerize.  If the returned value is not in a container, there's nothing to containerize.  "deliberately keeps its container" could imply there there is always a container.

Does this mean the LLM doesn't "know" about this distinction?

ugexe:
I wouldn't read too much into what could be implied by that wording. I understood it as "on a containerized return, a normal routine strips it and an is rw one preserves it", i.e. in contrast to decontainerizing.

ugexe:
fwiw I just changed the text so one is less likely to think there is any implication that there is always a container

PR #6335, opened 2026-06-28
RakuAST: Run the optimize pass as its own pipeline stage

ugexe:
Previously the RakuAST frontend ran its optimize pass from the comp-unit action, during the parse stage. The legacy frontend runs its optimizer as an `optimize` pipeline stage after `ast`.

Register an `optimize` stage for the RakuAST frontend between `ast` and `qast` and move the pass there. The stage reads the resolver back off the `RakuAST::CompUnit` it is handed through a new `resolver` accessor and runs the pass. `CompUnit.optimize` pushes the unit scope itself, in the batch resolve mode for an already-checked tree, so the stage and the EVAL path drive the pass the same way.

`--target=optimize` now dumps the optimized AST, a new way to inspect the pass. `--target=ast`, and so `Str.AST`, now return the tree as parsed and checked, before optimization, because the ast stage stops short of the optimize stage. `.AST.EVAL` still gives the optimized result because EVAL runs the pass itself.

The optimize tests follow: the tree helper runs the pass explicitly rather than relying on `Str.AST` to do it, and the firing detectors check that the faithful parse keeps the operator while `--target=optimize` shows it folded.

### DIFF
--- a/src/Perl6/Compiler.nqp
+++ b/src/Perl6/Compiler.nqp
@@ -246,6 +246,18 @@ class Perl6::Compiler is HLL::Compiler {
     }
 
     method optimize($past, *%adverbs) {
+        # The RakuAST frontend optimizes the AST in place between the ast and
+        # qast stages. Its input is a RakuAST::CompUnit, not QAST. The pass
+        # manages its own scope, so the stage just hands it the resolver.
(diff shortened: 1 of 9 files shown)


ugexe:
@FCO I think this makes ASTQuery generally useful again

PR #6362, opened 2026-07-01
RakuAST: Decontainerize the value of an `INIT` prefix

ugexe:
`my $x = INIT $y` bound the Scalar the INIT phaser caches its result in, rather than the value in it, so `$x` was a container and a later `$x.method` hit the Scalar instead of the value. It broke Test::Scheduler, which keeps `INIT $*SCHEDULER` and later calls `.cue` on it.

The phaser caches its result in a Scalar and `IMPL-EXPR-QAST` handed that Scalar straight back. Decontainerize it, so the prefix yields the value the block produced, the same as the unphased expression.


ugexe:
closing in favor of https://github.com/rakudo/rakudo/pull/6363

PR #6370, opened 2026-07-02
Make Blob stringification agree across every route

ugexe:
Previously the same Blob could stringify or throw depending on which route the stringification took.

```raku
my $x; $x ~= "abc".encode;  # died: A utf8 is not a Str, so using '~' will not work
say ~"abc".encode;          # died the same way
say "" ~ "abc".encode;      # abc
say "abc".encode.Str;       # abc
say "x{"abc".encode}y";     # xabcy
```

The first commit fixes assigning-concat with an undefined left side. The `METAOP_ASSIGN` fast path for `~` handled that case with prefix `~b`, which refuses a Blob, while the defined path and the generic `METAOP_ASSIGN` fallback both concatenate, which stringifies through `.Stringy`. The same statement behaved three different ways depending on the path taken: the legacy optimizer's inlining works, the generic fallback works, and the fast path dies. The RakuAST frontend always calls the fast path, which broke HTTP::Tiny's test suite, and the legacy frontend reaches it too through `&infix:<~=>` or with the optimizer off. The fast path now uses `'' ~ b`, mirroring the generic case.

The second commit restores prefix `~` on the encoding-carrying Blob types. c94f5db60 gave utf8, utf16, and utf32 a `Str` and `Stringy` of `self.decode`, and `~$utf8` decoded through the general prefix `~`, which defers to `Stringy`. 539e96c26 then added a `prefix:<~>(Blob:D)` candidate so the error for stringifying a Buf names `~` and suggests `.decode`, and being narrower than the general candidate it also captured the encoding-carrying types, turning `~$utf8` into a throw. The candidate is removed, so the type's own `Stringy` decides again, and 539e96c26's goal moves into the message: the `Stringy` error says the `~` operator is what uses it, so every spelling still leads to `.decode`.

After both, every route agrees. utf8, utf16, and utf32 stringify by decoding, and an encoding-less Blob or Buf throws `X::Buf::AsStr` everywhere, with the same `.decode` guidance.

### DIFF
--- a/src/core.c/Buf.rakumod
+++ b/src/core.c/Buf.rakumod
@@ -1454,10 +1454,6 @@ my constant buf16 = Buf[uint16];
 my constant buf32 = Buf[uint32];
 my constant buf64 = Buf[uint64];
 
-multi sub prefix:<~>(Blob:D \a) {
-    X::Buf::AsStr.new(object => a, method => '~' ).throw
-}
-
 multi sub infix:<~>(Blob:D $a) { $a }
 multi sub infix:<~>(Blob:D $a, Blob:D $b) {
     my $res := nqp::create(nqp::eqaddr($a.WHAT,$b.WHAT) ?? $a !! Buf.^pun);
--- a/src/core.c/Rakudo/Internals.rakumod
+++ b/src/core.c/Rakudo/Internals.rakumod
@@ -1711,7 +1711,9 @@ my class Rakudo::Internals {
           &[%], -> Mu \a, Mu \b { a = a.DEFINITE ?? a % b !! "infix:<%>".no-zero-arg },
           &[-], -> Mu \a, Mu \b { a = a.DEFINITE ?? a - b !! -b },
           &[*], -> Mu \a, Mu \b { a = a.DEFINITE ?? a * b !! +b },
-          &[~], -> Mu \a, Mu \b { a = a.DEFINITE ?? a ~ b !! ~b },
+          # '' ~ b rather than ~b, mirroring the generic case, which
+          # concatenates the zero-arg identity with b
+          &[~], -> Mu \a, Mu \b { a = a.DEFINITE ?? a ~ b !! '' ~ b },
         ) -> \op, \metaop {
             metaop.set_name(op.name ~ ' + {assigning}');
             nqp::bindkey($metaop_assign, nqp::objectid(op), metaop);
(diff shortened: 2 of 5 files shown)


src/core.c/Rakudo/Internals.rakumod
```diff
@@ -1711,7 +1711,9 @@ my class Rakudo::Internals {
           &[%], -> Mu \a, Mu \b { a = a.DEFINITE ?? a % b !! "infix:<%>".no-zero-arg },
           &[-], -> Mu \a, Mu \b { a = a.DEFINITE ?? a - b !! -b },
           &[*], -> Mu \a, Mu \b { a = a.DEFINITE ?? a * b !! +b },
-          &[~], -> Mu \a, Mu \b { a = a.DEFINITE ?? a ~ b !! ~b },
+          # '' ~ b rather than ~b, mirroring the generic case, which
+          # concatenates the zero-arg identity with b
+          &[~], -> Mu \a, Mu \b { a = a.DEFINITE ?? a ~ b !! '' ~ b },
```
lizmat, line 1716: Isn't this really an issue with `prefix:<~>` for `Blob`s ?

ugexe: Yes, and the second commit of this PR removes that `prefix:<~>` Blob candidate. This commit is kept because the generic `METAOP_ASSIGN` fallback does `op.(op.(), b)`, i.e. `'' ~ b`. The fast path approximating that with `~b` is what let a prefix candidate change the behavior of `~=` in the first place, so it now uses the same expression as the fallback.

src/core.c/Buf.rakumod
```diff
@@ -1454,10 +1454,6 @@ my constant buf16 = Buf[uint16];
 my constant buf32 = Buf[uint32];
 my constant buf64 = Buf[uint64];
 
-multi sub prefix:<~>(Blob:D \a) {
```
lizmat, line 1457: But if we remove this, wouldn't a bare `~Blob` stop working???

ugexe: Stop working how? A bare `~Blob` has never worked. That candidate is the thing that throws, it doesn't stringify anything. Removing it just moves the throw to `Blob.Stringy` via the general `prefix:<~>(\a) { a.Stringy }`, with the same `X::Buf::AsStr` and a message that still points at `~` and `.decode`. What changes is that `~$utf8` decodes again, like it did before 539e96c26 captured it by accident. The tests in this commit pin both: `~$buf` dies naming `'~'`, `~$utf8` decodes.

lizmat: ok, understood.  Thanks for the elaboration :-)

PR #6395, opened 2026-07-05
RakuAST: let a lexical shadow the now and time terms

ugexe:
`now` and `time` are built-in listop terms, but a lexical of the same name should shadow them, so `my \time = calendar-from-posix ...; time.year` reads the binding rather than calling the built-in `&time`. The grammar matched the built-in term tokens unconditionally, so the bareword always became the built-in and a method call on it hit an Int.

Guard the `now` and `time` term tokens on there being no lexical of that name in scope, the way `nano` already guards on its own resolution. is-identifier-known covers both a sigilless `\time` and a routine `&time` or `sub time`, all of which legacy lets shadow the term, while the built-in lives under `&term:<time>` and so does not block itself.

### DIFF
--- a/src/Raku/Grammar.nqp
+++ b/src/Raku/Grammar.nqp
@@ -3150,8 +3150,8 @@ grammar Raku::Grammar is HLL::Grammar does Raku::Common {
     token term:sym<circumfix> { <circumfix> }
 
     token term:sym<self> { <.term-self> <.end-keyword> }
-    token term:sym<now>  { <.term-now>  <.end-keyword> }
-    token term:sym<time> { <.term-time> <.end-keyword> }
+    token term:sym<now>  { <!{ $*R.is-identifier-known('now')  }> <.term-now>  <.end-keyword> }
+    token term:sym<time> { <!{ $*R.is-identifier-known('time') }> <.term-time> <.end-keyword> }
 
     token term:sym<nano> {
         <?{ self.language-revision >= 3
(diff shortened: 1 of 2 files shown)


lizmat:
I don't understand how this is not a problem in legacy, and feel like the fix is really not adressing an underlying problem with system supplied terms / subs?

ugexe:
> I don't understand how this is not a problem in legacy, and feel like the fix is really not adressing an underlying problem with system supplied terms / subs?

Legacy has the same dedicated grammar tokens for these terms and also solves it the same way

lizmat:
I see the actions in legacy:
```raku
    method term:sym<now>($/) {
        make QAST::Op.new( :op('call'), :name('&term:<now>'), :node($/) );
    }   
        
    method term:sym<time>($/) {
        make QAST::Op.new( :op('call'), :name('&term:<time>'), :node($/) );
    }
```
and I see the tokens in legacy:
```raku
    token term:sym<now> { <sym> <.tok> }

    token term:sym<time> { <sym> <.tok> }
```
So where's the special handling?

ugexe:
Can you describe what `<.tok>` does? Maybe I'm misunderstanding something

lizmat:
It checks whether the current match is **not** defined as a symbol (either with or without `&` prefix), afaik.

UPDATE: I thought it was used a lot more widely, but I confused it with `<.kok>`.  Indeed, `<.tok` is only used for `time`, `now` and `nano` in legacy.

PR #6399, opened 2026-07-06
RakuAST: defer private method resolution when the package is composing

ugexe:
A private method call resolves at compile time by looking the method up on its package. A trait applied at BEGIN wraps methods with closures that are compiled before the class composes, so the lookup found nothing and the call died with "Private method not found" (LibXML's `is reader-raw`). Only resolve at compile time once the package has composed, otherwise emit the runtime private-method dispatch, which finds the method once the class is in place.

### DIFF
--- a/src/Raku/ast/call.rakumod
+++ b/src/Raku/ast/call.rakumod
@@ -1077,7 +1077,7 @@ class RakuAST::Call::PrivateMethod
         if $!name.is-identifier {
             my $name := self.IMPL-UNWRAP-LIST($!name.parts)[0].name;
             my $package := $!package;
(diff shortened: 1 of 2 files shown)


lizmat:
Is the "falling back to runtime lookup" something that legacy does as well?

Feels to me this should be more like the handling that subroutine references get (aka, resolve at compose time?)

ugexe:
> Is the "falling back to runtime lookup" something that legacy does as well?

Runtime lookup isn't something legacy falls back to for private method calls, it's all it ever does

> Feels to me this should be more like the handling that subroutine references get (aka, resolve at compose time?)

Compose time resolution can't work here because the callsite is in a closure the trait compiles at BEGIN, which is already bytecode by the time the class composes

lizmat:
Understood.

PR #6400, opened 2026-07-06
RakuAST: return Nil from a QUIT phaser when its when/default succeeds

ugexe:
Previously a whenever block's QUIT phaser never intercepted the quit under the RakuAST frontend: the react or supply block died with the original exception even though a default clause matched. The supply machinery decides whether a QUIT phaser handled the exception by what the phaser returns, Nil when handled, and treats anything else as unhandled and rethrows. The RakuAST-compiled phaser block ended in the generic succeed handler, which returns the succeed payload, so a matching when/default made the phaser return that payload instead of Nil and the quit was always considered unhandled.

This makes a QUIT phaser block's succeed handler produce Nil, with the payload sunk, the way the legacy frontend rewrites it. A QUIT whose when clauses do not match still returns the exception, so an unhandled quit rethrows as before, and succeed handlers of ordinary blocks are unaffected.

### DIFF
--- a/src/Raku/ast/scoping.rakumod
+++ b/src/Raku/ast/scoping.rakumod
@@ -16,6 +16,7 @@ class RakuAST::LexicalScope
 
     # Handlers related to this scope.
     has int $!need-succeed-handler;
+    has int $!nil-on-succeed;
     has Mu $!catch-handlers;
     has Mu $!control-handlers;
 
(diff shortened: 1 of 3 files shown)


lizmat:
This feels like it is adding yet another flag that's only being used for a single case.  Wouldn't it make more sense to just check for the class name/type (`::QUIT`).

ugexe:
> This feels like it is adding yet another flag that's only being used for a single case. Wouldn't it make more sense to just check for the class name/type (::QUIT).

You can see the flag being set right next to the existing set-needs-result(True) on the blorst. That is the QUIT prefix configuring its block from the outside, because that is the only direction the information flows. The succeed handler itself is emitted over in LexicalScope's QAST generation, where self is just a plain Block, so what exactly would we be checking ::QUIT against there?

lizmat:
Ok, understood.

PR #6405, opened 2026-07-07
RakuAST: box native values produced by begin-time evaluation

ugexe:
Previously a begin-time evaluated expression of a native type, such as the initializer in HarfBuzz's

    constant HB_SET_VALUE_INVALID = (my uint32 $ = -1);

produced a bare VM-level box (a BOOTInt) where the same expression at run time produces an Int. That value has no Raku method table and a null WHO, and it flowed into the constant's lexical value, its package stash entry, and its EXPORT stash entry. Package-qualified access to such a constant then died with "No such method 'defined' for invocant of type 'BOOTInt'", and when the constant reached a consumer's GLOBAL through two dependency paths the GLOBAL merge misclassified the leaf as a package stub, recursed into its null WHO, and died in stash_hash with "Cannot find method 'FLATTENABLE_HASH' on object of type VMNull". This is what prevented HarfBuzz, and everything downstream of it such as Pod::To::PDF, from compiling.

The legacy frontend leaves the same raw BOOTInt in the stash, but its package-qualified lookups go through a Stash.AT-KEY method call and the value gets hllized at that boundary, so the rawness never surfaces there.

This makes IMPL-BEGIN-TIME-EVALUATE box VM-level scalars (BOOTInt, BOOTNum, BOOTStr) into their Raku types, making begin-time evaluation agree with run-time semantics. VM-level aggregates are left alone since constants holding nqp hashes and lists rely on staying unboxed.

### DIFF
--- a/src/Raku/ast/begintime.rakumod
+++ b/src/Raku/ast/begintime.rakumod
@@ -59,8 +59,9 @@ class RakuAST::BeginTime
         }
 
         # Can interprete, so do that
+        my $result;
         if $code.IMPL-CAN-INTERPRET {
-            $code.IMPL-INTERPRET(
+            $result := $code.IMPL-INTERPRET(
               RakuAST::IMPL::InterpContext.new(:$resolver, :$context)
             )
         }
@@ -69,7 +70,7 @@ class RakuAST::BeginTime
         # is whatever its cache holds, which is undefined until it runs. Hand
         # that back rather than the block itself.
         elsif nqp::istype($code, RakuAST::StatementPrefix::Phaser::Init) {
-            nqp::ifnull(nqp::decont($code.container), Mu)
+            $result := nqp::ifnull(nqp::decont($code.container), Mu)
         }
 
         # A code literal's begin-time value is its own code object,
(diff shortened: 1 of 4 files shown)


src/Raku/ast/begintime.rakumod
```diff
@@ -90,11 +91,27 @@ class RakuAST::BeginTime
...
+        # A native-typed expression evaluates to a bare VM-level box (for
+        # example a BOOTInt) where the same expression at run time would
+        # produce an Int. Such a box has no Raku method table and a null
+        # WHO, so letting it escape into a stash breaks stash consumers
+        # like the compunit GLOBAL merge. Box VM-level scalars into their
+        # Raku types; leave VM-level aggregates alone, as constants holding
+        # nqp hashes and lists rely on staying unboxed.
+        unless nqp::isnull($result) {
+            my $type := nqp::what($result);
+            $result := nqp::hllizefor($result, 'Raku')
```
lizmat, line 109: Doesn't `nqp::hllizefor` already do the right thing for all possible native values??

ugexe: > VM-level aggregates are left alone since constants holding nqp hashes and lists rely on staying unboxed.

```
# Raku types; leave VM-level aggregates alone, as constants holding
# nqp hashes and lists rely on staying unboxed.
```

PR #6445, opened 2026-07-12
Implement the appending feed operators <<== and ==>>

ugexe:
Previously both frontends threw X::Comp::NYI for the appending feed operators. This implements them in make_feed and RakuAST::Feed as the rightward and leftward pipeline builders they already half-expected, with one difference from ==> and <==: a variable stage passes on its whole contents after the append rather than just the values fed into it, so `my @c = do { @a ==>> @b }` yields @b's prior contents followed by @a's values, matching S03-feeds/basic.t.

The S03 behavior where ==>> looks ahead to a downstream @(*) stage remains unimplemented along with the rest of the whatever-feed machinery.

### DIFF
--- a/src/Perl6/Actions.nqp
+++ b/src/Perl6/Actions.nqp
@@ -7593,16 +7593,13 @@ class Perl6::Actions is HLL::Actions does STDActions {
     sub make_feed($/) {
         # Assemble into list of AST of each step in the pipeline.
         my @stages;
-        if $/<infix><sym> eq '==>' {
+        my $sym := ~$/<infix><sym>;
+        my int $appending := $sym eq '==>>' || $sym eq '<<==';
+        if $sym eq '==>' || $sym eq '==>>' {
             for @($/) { @stages.push($_); }
         }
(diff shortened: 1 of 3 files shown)


ab5tract:
Honestly I have daydreamed about hijacking `==>>` to allow thunkable/callable arguments that operate per-element, rather than strictly only allowing a Call that takes the whole array:

```
^5 Z 5..9 ==>> *.sum ==>> my @pair-sums;
# @pair-sums = [5, 7, 9, 11, 13]
```

 I always forget what this feed form _actually_ does, and seeing it in action, I can understand why: it feels like a major nothing-burger to me (the design of it, not that the design was implemented). Appending? Really? Why on earth isn't that the default behavior? If you want an empty array, you declare one in your feed. Truncation seems like a really bad default and something that can safely be left behind in previous language versions.
 
 Not only are per-element instructions easier to write and reason about, they should also be simpler to provide auto-threading as elements that are finished with Feed Stage A can be pushed to Feed Stage B, etc.

PR #6448, opened 2026-07-12
Compare revision-gated candidate signatures structurally

ugexe:
Previously the dispatch filter for revision gated protos grouped candidates by their rendered signature, so superseding a candidate at a later language revision only worked when the replacement copied the original's parameter names exactly. With different names both candidates were admitted, and every call from a revision that admits both died at runtime:

```raku
proto sub f(|) is revision-gated("6.c") {*}
multi sub f(Int $x) is revision-gated("6.c") { "old" }
multi sub f(Int $y) is revision-gated("6.e") { "new" }

# use v6.e.PREVIEW callers:
# Ambiguous call to 'f(Int)'; these signatures all match:
#   (Int $x)
#   (Int $y)
```

The first commit keeps the rendered signature as a cheap first comparison and falls back to a structural one: the same parameter shapes and types in the same order, ignoring parameter names. Signatures with named, slurpy, capture or sub-signature parameters, constraints or type captures still require identical text to supersede, since a name-blind comparison cannot vouch for those: a named parameter's name is its interface, and two `where` constraints with identical surroundings cannot be told apart statically. For those shapes the behavior is unchanged. Both filter copies (src/vm/moar/dispatchers.nqp and the BOOTSTRAP one used by other backends) get the same change.

The second commit resolves a TODO in the RakuAST frontend's duplicate multi check. Previously a revision gated proto turned off the equivalent-signature redeclaration worry entirely, because equivalent candidates gated at different revisions are how a routine evolves per language version. That also silenced the worry for candidates clashing within the same revision, which can never both be dispatched to. The check now compares only candidates competing at the same effective revision, with an ungated candidate defaulting to the proto's revision as it does at dispatch time:

```raku
proto sub f(|) is revision-gated("6.c") {*}
multi sub f(Int $x) is revision-gated("6.c") { "first" }
multi sub f(Int $y) is revision-gated("6.c") { "second" }
# now warns: Redeclaration of multi sub 'f' with equivalent signatures ...
```

### DIFF
--- a/src/Perl6/bootstrap.c/BOOTSTRAP.nqp
+++ b/src/Perl6/bootstrap.c/BOOTSTRAP.nqp
@@ -3852,10 +3852,54 @@ BEGIN {
       nqp::getstaticcode(sub ($self, $caller-revision, @candidates) {
         $self := nqp::decont($self);
 
+        # Whether two signatures are interchangeable for revision gating:
+        # the same parameter shapes and types in the same order, differing
+        # at most in parameter names. Anything a name-blind comparison
+        # cannot vouch for (named, slurpy, capture or sub-signature
+        # parameters, constraints, type captures) makes them
+        # non-interchangeable, so such candidates each stand on their own,
+        # as any two distinct signatures do.
+        my sub signatures-equivalent($sig-a, $sig-b) {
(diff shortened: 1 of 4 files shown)


ab5tract:
Looking good :+1:

PR #6449, opened 2026-07-12
RakuAST: compile an .AST tree's own unit when EVALing it

ugexe:
Previously declaring a multi inside a Str.AST tree died at the call site with "lang-call cannot invoke object of type 'VMNull'", and any require died with "bindkey_o requires a concrete object" or "Lexical '&sym' does not exist in this frame". Cool.AST returns the parsed unit's statement list, whose statements were already begun as part of that unit, so begin-time state living on its mainline scope (the implicit proto's generated declaration, require's symbol captures) was lost when EVAL wrapped the statements in a fresh unit, where begin cannot run again.

Cool.AST now records the parsed unit on the statement list it hands out, and EVAL compiles that unit when the statement list is still the unit's own. A statement list transplanted into a user-built tree is unaffected. Since the backend pops the compiling SC after every non-nested compile, a reused unit pushes its SC back beforehand, which also fixes EVALing the same tree, or an .AST(:compunit) result, more than once.

### DIFF
--- a/src/Raku/ast/compunit.rakumod
+++ b/src/Raku/ast/compunit.rakumod
@@ -307,6 +307,15 @@ class RakuAST::CompUnit
     # does not declare its own GLOBAL and so forth.
     method is-eval() { $!is-eval ?? True !! False }
 
+    # Put this unit's SC back on the compiling-SC stack before compiling it
+    # again. Creating the unit pushed the SC, but the backend pops it after
+    # every non-nested compile, so a reused unit (an .AST tree EVALed a
+    # second time) would otherwise compile with no compiling SC. A nested
+    # unit shares the enclosing compilation's SC, which is already current.
+    method IMPL-PUSH-COMPILING-SC() {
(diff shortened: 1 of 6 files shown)


lizmat:
`Cool.AST` has a [:compunit](https://github.com/rakudo/rakudo/blob/main/src/core.c/core_epilogue.rakumod#L788) named argument, that would return the whole compunit.  Is this PR needed if one would have specified `:compunit` with `.AST`?

If the answer is yes, then I suggest changing the default for `:compunit` to `True`.

If the answer is no because of multiple `EVAL`'s on the same tree, then by all means merge  :-)

ugexe:
Yes it's still needed even with :compunit precisely because of multiple EVALs on the same tree. Without this PR:
```
RAKUDO_RAKUAST=1 raku -e 'my $cu = Q|41 + 1|.AST(:compunit); say $cu.EVAL; say $cu.EVAL;'
42
===SORRY!===
No current compiling SC
```

PR #6451, opened 2026-07-14
Remove dead code from Perl6::ModuleLoader

ugexe:
resolve_repossession_conflicts has been dead since 7262b98d3e moved repossession-conflict handling to CompUnit::RepositoryRegistry. Both frontends and the legacy World wire that class's resolver; nothing calls this leftover copy. It also carried a stray-semicolon bug and an "XXX typecheck??". Drop it.

register_absolute_path_func has no callers, so $absolute_path_func is never set and absolute_path is already an identity function. Drop the unused hook and reduce absolute_path to a passthrough.

### DIFF
--- a/src/Perl6/ModuleLoader.nqp
+++ b/src/Perl6/ModuleLoader.nqp
@@ -12,7 +12,6 @@ sub DEBUG(*@strs) {
 class Perl6::ModuleLoader does Perl6::ModuleLoaderVMConfig {
     my %modules_loaded;
     my %settings_loaded;
-    my $absolute_path_func;
 
     my %language_module_loaders := nqp::hash(
         'NQP', nqp::gethllsym('nqp', 'ModuleLoader'),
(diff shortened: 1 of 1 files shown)


src/Perl6/ModuleLoader.nqp
```diff
@@ -406,34 +399,6 @@ class Perl6::ModuleLoader does Perl6::ModuleLoaderVMConfig {
...
-            # get appropriately copied.
-            if $original.HOW.name($original) eq 'Stash' {  # XXX typecheck??
-                my %current := nqp::atpos(@conflicts, $i + 1);
-
-                for $original.FLATTENABLE_HASH {
-                    my str $key := $_.key;
-
-                    nqp::bindkey(%current, $key, $_.value)
-                      if nqp::eqat($key, '&', 0);
-                      || nqp::not_i(nqp::existskey(%current, $key))
```
ugexe, line 427: Note the stray semicolon that was accidentally added to the previous line. This code seems like it has been broken for awhile unnoticed since it also appears to be unused.

PR #6457, opened 2026-07-16
RakuAST: rewrite a square by the power operator to a multiply

ugexe:
Previously $x ** 2 always called the power routine, which handles any exponent and bottoms out in a bignum power for Int and libm pow for Num, where a multiply is a single operation. Measured at 1.4x to 1.7x per operation across Int, Num, and native num operands.

This rewrites a square in the optimize pass when the operand is a plain resolved variable, so no side effect is duplicated, and its type rules out a Junction: a junction squares each eigenstate, but autothreads over both sides of a multiply, which builds a different junction. A native operand can never hold one; a boxed operand qualifies when its declared type and Junction are unrelated. An untyped variable therefore keeps the power routine. The legacy optimizer rewrites any plain variable and autothreads a junction square wrongly, which the type condition avoids rather than inherits.

### DIFF
--- a/src/Raku/ast/base.rakumod
+++ b/src/Raku/ast/base.rakumod
@@ -702,6 +702,10 @@ class RakuAST::Node {
             $result := self.IMPL-COLLAPSE-TYPEMATCH($resolver, $expr);
         }
 
+        if $result =:= $expr {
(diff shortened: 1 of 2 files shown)


ugexe:
just gonna include this in a batch optimization pr

PR #6458, opened 2026-07-17
RakuAST: port six legacy optimizer rewrites to the optimize pass

ugexe:
Previously the RakuAST frontend compiled code without several rewrites the legacy optimizer performs: constant operand folding, reducing a square by the power operator to a multiply, skipping the return decont for container-free results, propagating a callee's return type onto its call, lowering plain array initializations past STORE, and unrolling literal index slices. This ports each one to the RakuAST optimize pass, one commit per optimization.

- **Fold a pure operator on constant operands.** Previously `my constant K = 3; my $y = K * 2` kept the runtime operator. Operands now also accept a name or lexical that resolves to a genuinely constant declaration. The compile-time-value protocol alone cannot vouch for an operand: a plain variable declaration claims its container default as a value, a stash reference claims the resolved package where evaluating it produces that package's stash, and an our-scoped variable imports as its Scalar container whose value a later assignment can replace. All three keep the runtime operator.
- **Rewrite a square by the power operator to a multiply.** `$x ** 2` becomes `$x * $x` when `$x` is a lexical whose type rules out a Junction. This is deliberately narrower than legacy, which rewrites untyped lexicals and mishandles a Junction operand.
- **Skip the return decont for a container-free result.** A routine whose last statement produces a value that cannot be containerized (a native, a literal, `self`, a deconted call) returns without the decont round trip.
- **Copy a bound-once callee's return type onto its call.** A native return is offered both ways through a Want, boxed for object context and raw for native context, so a native consumer of a native returning sub skips the boxing round trip. Dispatchers and rw callees are left alone. A containerized callee also declines: an our-scoped callee imports as its Scalar container, whose value a later assignment can replace, so the return type of the value it happens to hold at compile time is no promise. The legacy optimizer copies it anyway and a call after such a reassignment dies unboxing, which the new test records as a todo there.
- **Lower a plain array initialization past STORE.** `my @a = 1, 2, 3` builds the List, Reifier, and IterationBuffer directly instead of dispatching STORE, following the legacy lowering.
- **Unroll a slice by literal integer indexes to AT-POS calls.** `@a[1, 3]` becomes the AT-POS calls and an infix `,` list instead of a full slice dispatch.

Naive benchmarks against main, same one-liner loop best of three at default optimization: constant folding 5.1x (`K * 2 + 4` in a 3M loop, the body folds to a constant), square rewrite 2.1x (3M), array initialization 2.8x (1M), slice unroll 5.0x (1M). The return decont and return type propagation commits show no difference on a warm microloop, where spesh inlines the callee and erases the round trips anyway. Their effect is on the emitted QAST, which pays in code spesh has not specialized.

Each optimization comes with a `t/08-performance` test file asserting both the emitted QAST shape and runtime behavior under both frontends, including regression coverage for the decline cases above.

### DIFF
--- a/src/Raku/ast/base.rakumod
+++ b/src/Raku/ast/base.rakumod
@@ -702,6 +702,10 @@ class RakuAST::Node {
             $result := self.IMPL-COLLAPSE-TYPEMATCH($resolver, $expr);
         }
 
+        if $result =:= $expr {
+            $result := self.IMPL-REWRITE-SQUARE($resolver, $expr);
+        }
+
         # Lowerings that direct code generation rather than replacing the
         # node register their marks here, gated on the optimize pass running.
         # They each drop a layer of operator dispatch or pin down a routine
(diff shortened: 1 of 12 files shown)


MasterDuke17:
> Naive benchmarks against main, same one-liner loop best of three at default optimization: constant folding 5.1x (`K * 2 + 4` in a 3M loop, the body folds to a constant), square rewrite 2.1x (3M), array initialization 2.8x (1M), slice unroll 5.0x (1M).

Excited to see this, but not sure what these metrics are representing. I assume the `*x` are compared to RakuAST without this PR? What about compared to non-RakuAST?

ugexe:
Those figures are based on the RakuAST frontend with this PR against the RakuAST frontend on main, both at default optimization. With legacy:

| loop | legacy frontend | RakuAST main | RakuAST this PR |
|---|---|---|---|
| constant folding (`K * 2 + 4`, 3M) | 0.079 | 0.534 | 0.105 |
| square (`$x ** 2`, 3M) | 0.429 | 0.928 | 0.434 |
| array init (`my @a = 1, 2, 3`, 1M) | 1.184 | 1.825 | 0.646 |
| slice unroll (`@a[1, 3]`, 1M) | 0.244 | 1.634 | 0.327 |
| return decont (`f()` returning 42, 3M) | 0.081 | 0.108 | 0.109 |
| return type propagation (`my int $x = f() + 1`, 3M) | 0.187 | 0.431 | 0.436 |

and yes, this shows RakuAST is actually faster than legacy in one instance where legacy stopped emitting the optimization at some point in history

PR #6464, opened 2026-07-19
I rename RakuAST currying to priming

### LINKED ISSUE #6199: RakuAST: Rename curry/currying to prime/priming
@ab5tract I remember you saying something about this. If there are any other details that seem relevant please leave them here.

w3lld1:
## Summary

- I rename the RakuAST WhateverCode internals from curry/currying to prime/priming.
- I update the related methods, thunk classes, fields, comments, and the RakuAST TODO entry.
- I leave legacy compiler and metamodel uses of `curry` unchanged because they are outside the RakuAST terminology change.

## Validation

- I completed `perl Configure.pl --gen-moar --gen-nqp --backends=moar`.
- I verified the generated `gen/moar/ast.nqp` contains the renamed classes and methods with no old RakuAST curry symbols.
- I ran a source-reference consistency check across all 37 `src/Raku/ast/*.rakumod` modules; it found both renamed thunk classes, all 16 prime-related methods, no undeclared prime-method calls, and no old curry terminology.
- I verified the complete diff is limited to the requested curry-to-prime terminology mapping.
- I ran `git diff --check`.
- I started `make -j2`; it generated and compiled the Rakudo/NQP stages and reached `blib/CORE.c.setting.moarvm`, but the full build did not complete within the local time limit. I defer the complete build and test suite to CI.

Fixes #6199

### DIFF
--- a/src/Raku/ast/begintime.rakumod
+++ b/src/Raku/ast/begintime.rakumod
@@ -139,9 +139,9 @@ class RakuAST::BeginTime
     ) {
         my $*IMPL-COMPILE-DYNAMICALLY := 1;
 
-        # A curried argument (a WhateverCode) may be interpreted here, as
+        # A primed argument (a WhateverCode) may be interpreted here, as
         # its static block compiles against a real QAST context.
-        my $*IMPL-INTERPRET-CURRIED := 1;
+        my $*IMPL-INTERPRET-PRIMED := 1;
 
         # Ready to call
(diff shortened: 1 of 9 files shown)


ab5tract:
I wasn't expecting this change to be so impactful, but the readability of `IMPL-PRIMES` over `IMPL-CURRIES`, et al, is to my eyes much more viscerally comprehensible.

Thanks!

PR #6467, opened 2026-07-20
RakuAST: stop reporting an orphaned declarator doc as a missing declarand

ugexe:
Legacy never reports a missing declarand. RakuAST did in two cases where legacy stays silent, producing spurious worries on real modules (seen across the PDF::* distributions):

  - A trailing `#=` that could not attach to its preceding declarand became a worry once the next declarand appeared. Such an orphan is now dropped, the way legacy drops it.

  - A leading `#|` before a lexical was reported as a missing declarand when no later documentable declarand claimed it. A lexical is a valid place for the doc to land, so remember the first non-podifiable declarand a leading doc falls through and, if nothing else claims it, ground the doc there instead of warning. A leading doc that reaches no declarand at all is still reported, and one that falls through to a documentable target still attaches there.

### DIFF
--- a/src/Raku/Actions.nqp
+++ b/src/Raku/Actions.nqp
@@ -787,19 +787,31 @@ class Raku::Actions is HLL::Actions does Raku::CommonActions {
         else {
             my $from    := $/.from;
             my $worries := $*DECLARAND-WORRIES;
+            # A trailing `#=` that could not attach to a preceding declarand
+            # is dropped here rather than reported: legacy is silent about an
+            # orphaned trailing doc, so a following declarand must not turn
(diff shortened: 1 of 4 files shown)


lizmat:
It feels to me the errors are valid, and legacy is not doing enough to report it.

So I would suggest it being revision-gated to *have* it report errors in 6.e.

ugexe:
ping @dwarring - this PR would fix many of the PDF distributions in RakuAST but maybe it would be better to fix the issue in those distributions instead. I'll defer to you what you think the correct action should be given your familiarity with that code.

PR #6471, opened 2026-07-20
RakuAST: don't run LAST in a loop that never runs its body

ugexe:
Previously a pre-test loop (while/until, or a C-style loop with a condition) fired its LAST phaser even when the condition was false at once and the body never ran, though FIRST already did not. Gate LAST on the body having run at least once.

An eager (statement-level) while/until loop folds a did-run flag into its condition, set as an iteration starts so an immediate `last` still counts, and runs LAST only when the flag is set. A value-context loop lowers to a lazy from-loop, whose iterator now runs the body's LAST phaser at exhaustion, only if the body ran, gated by a fire-last flag threaded through from-loop.

This is the RakuAST frontend only; the legacy frontend is unchanged.

### DIFF
--- a/src/Raku/ast/statements.rakumod
+++ b/src/Raku/ast/statements.rakumod
@@ -1677,6 +1677,35 @@ class RakuAST::Statement::Loop
         QAST::WVal.new(:value($run-phasers))
     }
 
+    # A pre-test loop (while/until, or a C-style loop with a condition) that
+    # never runs its body must not fire LAST; FIRST already behaves this way. A
+    # repeat loop always runs its body once, so it keeps firing LAST. This is
+    # only for the eager (statement-level) loop; a lazy from-loop fires LAST
+    # from its iterator instead.
+    method IMPL-GUARD-LAST() {
+        my $phasers := nqp::getattr($!body.meta-object, Block, '$!phasers');
+        my @last := nqp::ishash($phasers) && nqp::existskey($phasers, 'LAST')
+          ?? $phasers<LAST> !! [];
+        nqp::elems(@last) && $!condition && !self.repeat
+    }
+
+    # Fold a did-run flag into a pre-test loop condition so LAST can tell
+    # whether the body ran. Binds the flag lexical when an iteration is about to
+    # run (before the body, so an immediate `last` still counts). $while is true
+    # for a while-style loop (body runs on a true condition), false for until.
+    method IMPL-LAST-GUARD-CONDITION(Mu $cond-qast, int $while, str $ran) {
+        my $set := QAST::Op.new(:op<bind>,
+          QAST::Var.new(:name($ran), :scope<lexical>, :returns(int)),
+          QAST::IVal.new(:value(1)));
+        $while
+          ?? QAST::Op.new(:op<if>, $cond-qast,
+               QAST::Stmts.new($set, QAST::IVal.new(:value(1))),
+               QAST::IVal.new(:value(0)))
+          !! QAST::Op.new(:op<if>, $cond-qast,
+               QAST::IVal.new(:value(1)),
+               QAST::Stmts.new($set, QAST::IVal.new(:value(0))))
+    }
+
     method IMPL-TO-QAST(RakuAST::IMPL::QASTContext $context) {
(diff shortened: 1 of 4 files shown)


lizmat:
I think that if we're not doing revision gating on this, it should *also* be fixing the legacy frontend.  Plus this would make the code a lot simpler.

ugexe:
It isn't trivial to fix it on the legacy frontend (i've seen it and rejected it), and it certainly wouldn't simplify anything. What is the value to doing it given how close we are to releasing rakuast outright? What is the point of the xx-fixed-in-rakuast test stuff as well? What would be the point in revision gating the broken behavior in a rakuast only solution if we were also ok with fixing it outright in both frontends with no gating?

lizmat:
Ok, if I understand the issue correctly: the legacy frontend just codegens the call to the LAST phaser after handling the loop.  Part of this PR is to add a flag to tell the iterator to call the LAST phaser if there is one.  The old backend doesn't set this flag, the new one does.

Removing the flag and just always have the iterator run the LAST phaser, would cause the LAST phaser to be run twice on the legacy backend.

Feels to me we should create new iterators that always call the LAST phaser when appropriate (without needing a flag set), rename the old ones to "_legacy" and have the legacy backend call that.

Then once RakuAST is the default, get rid of the "_legacy" iterators.

This at least doesn't add any unneeded overhead on the new backend, and only created dead code if we forget to remove the "_legacy" iterators.

If this is an approach that makes sense, I'll work on that.

ugexe:
Creating a second iterator family doesn't remove the need for the flag, it just moves it. Both frontends codegen the same Seq.from-loop call, so something still has to say which family to build. That means either the legacy codegen changes to call the _legacy ones, or the RakuAST codegen stops calling from-loop and constructs the new iterators itself. It can't be decided at runtime, because a legacy-precompiled module loaded into a RakuAST process would get the always-fire iterators and run LAST twice.

So the trade is a named argument that defaults to off and is checked once when the Seq is constructed, versus four iterators that each exist twice differing by a few lines, kept in sync until the legacy frontend is removed. The flag is the smallest version of the signal the split would need anyway.

With that being said, and as I mentioned previously, I did look at implementations that attempt to solve this on the legacy frontend ([here is one for example](https://github.com/ugexe/rakudo/commit/3e546fd0f84cfe766b01d9bebfa71bcff2fd64ec#diff-ec36479462b7787dc44cbcddb0a5507243cb49adf1280be1eefeaa8a6fabf6e5)) and decided it was not worth it. If you feel strongly about it feel free to take a crack at it.

Regardless, I'm still interested in the answers to these questions:
* What is the value to doing it given how close we are to releasing rakuast outright?
* What is the point of the xx-fixed-in-rakuast test stuff as well?
* What would be the point in revision gating the broken behavior in a rakuast only solution (one of the suggestions made) if we were also ok with fixing it outright in both frontends with no gating?

ab5tract:
xx-fixed-in-rakuast was set up in order to ensure that things that were currently working in RakuAST would stay working. Some of these were working only because there was no optimizer. But no tests failed in roast either way, so one could implement the RakuAST optimizer and make a similar bug and not notice anything wrong.

The intention was that each test would be placed into relevant files in roast upon switching to RakuAST.

ugexe:
> xx-fixed-in-rakuast was set up in order to ensure that things that were currently working in RakuAST would stay working. Some of these were working only because there was no optimizer. But no tests failed in roast either way, so one could implement the RakuAST optimizer and make a similar bug and not notice anything wrong.
> 
> The intention was that each test would be placed into relevant files in roast upon switching to RakuAST.

Maybe a better question is this: does every issue marked with "Fixed in RakuAST" expected to get a fix on the legacy frontend before RakuAST is released? Maybe I misunderstood, but I assumed the answer to this would be no.

ab5tract:
I do not expect Fixed in RakuAST to be back ported. So for me I don't see a strong need for the legacy compiler.

@ugexe raises some good points in terms of questioning the utility of fixing this for both frontends vs the same impact of leaving the fix without a revision gate.

I vote for bringing this in as is. Probably whoever encountered this changed their code to avoid the broken base behavior anyway.

PR #6474, opened 2026-07-21
RakuAST: fold constant conditions and drop the code they make dead

ugexe:
Previously the RakuAST optimize pass compiled code that can never run. A conditional with a compile-time condition still emitted both branches and tested the constant at runtime, and statements after an unconditional `return` were compiled though nothing reaches them. This teaches the pass to drop such code, and extends constant folding to the chaining comparisons that make many conditions decidable in the first place.

**Fold a chaining comparison on constant operands.** Chaining infixes were excluded from constant folding, because replacing a link of a longer chain with its answer would hand the next comparison a boolean in place of the shared middle operand. A decided comparison in an argument position now records its answer on the operator, where an enclosing chain withdraws it, rather than replacing the node. So a constant comparison folds like any other pure operator, and `2 > 4` becomes decidable at compile time.

**Keep only the branch a constant condition selects.** An `if` or `unless`, statement or statement modifier, whose condition is known at compile time now keeps only the code that could run. The taken branch stands as a bare block, which the lowering analysis may flatten into the enclosing frame, and a false conditional with no other branch leaves its `Empty`. A branch that takes arguments, a pointy or one with placeholders, receives the condition's value, so it keeps its conditional, as does a block statement under a condition modifier. A `with` tests definedness and topicalizes, so only plain `if` and `unless` collapse, and a constant whose `Bool` throws keeps the throw at runtime. A dropped branch confines its declarations, so no name is lost.

**Drop statements nothing after a return can reach.** Statements following an unconditional `return`, `return-rw`, or `exit` are now dropped. Only those three end reachability. A thrown exception does not, since a handler can `.resume` into the statements right after the throw, so `die` and the loop controls leave their followers alone, and the call must carry no condition or loop modifier. Statements that still register something are kept: anything that declares a name, since a routine declared after a return is callable through its hoisted name, plus phasers and `CATCH` and `CONTROL` blocks, which attach to the scope wherever they sit.

### DIFF
--- a/src/Raku/ast/base.rakumod
+++ b/src/Raku/ast/base.rakumod
@@ -748,6 +748,7 @@ class RakuAST::Node {
             self.IMPL-MARK-WHEN-TYPEMATCH($resolver, $expr);
             self.IMPL-MARK-JUNCTION-FOLD($resolver, $expr);
             self.IMPL-MARK-CHAIN-LINKS($resolver, $expr);
+            self.IMPL-DROP-UNREACHABLE($resolver, $expr);
             self.IMPL-MARK-PARAM-WHERE-JUNCTION($resolver, $expr);
         }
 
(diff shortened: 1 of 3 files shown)


ugexe:
@MasterDuke17 a couple of dead code eliminations are included in this PR

MasterDuke17:
> @MasterDuke17 a couple of dead code eliminations are included in this PR

Very cool. It's fun to see the code just disappear. However, it's definitely not catching everything it could. So `sub f($a) { if $a > 3 { return 4 }; return 9 ; say "here?" }; say f(4)` doesn't have the `say "here?"` in the output of `--target=optimize`, but `sub f($a) { if $a > 3 { return 4 } else { return 9 }; say "here?" }; say f(4)` does.

PR #6478, opened 2026-07-23
RakuAST: look up an indirect ::("EXPORT") at run time

ugexe:
Reading a required module's exports through an indirect lookup, as in `require ::($bundle); ::("EXPORT").WHO<DEFAULT>`, returned nothing under the RakuAST frontend. The constant string "EXPORT" resolved at compile time to this compilation unit's own export-package, which is empty, instead of running the symbolic lookup that would see the exports a `require` merged into the caller's %?REQUIRE-SYMBOLS.

An indirect lookup is a run-time symbolic lookup, so three places now treat it as one rather than folding the constant string:

- The resolver no longer applies the EXPORT special-case to an indirect root, so `::("EXPORT")` stays unresolved and reaches code generation as a runtime lookup. A direct EXPORT::<...> reference is unaffected.
- is-name-type reports an indirect lookup as a term, not a type, so an unresolved one is not routed into the undeclared-type check.
- Code generation for an unresolved indirect pseudo-package lookup uses the lookup that consults %?REQUIRE-SYMBOLS instead of the pseudo-stash lookup.

A constant indirect lookup that resolves to a stable symbol (a type, a setting sub) still folds, so `my constant T = ::("Int")` and the reduce metaops are unchanged.

This surfaced from Test::Async, whose bundles declare exports through an explicit EXPORT::DEFAULT package that its aggregator re-exports.

### DIFF
--- a/src/Raku/ast/resolver.rakumod
+++ b/src/Raku/ast/resolver.rakumod
@@ -354,6 +354,14 @@ class RakuAST::Resolver {
 
         # Resolve the root part.
         my $name     := $root.name;
+
+        # EXPORT resolves to this unit's own export-package. That suits a
+        # written EXPORT::<...> reference, but an indirect lookup is a runtime
+        # symbolic lookup and must not fold to a compile-time package. A
+        # constant-string indirect lookup still arrives as an Expression part.
+        return Nil
+          if $name eq 'EXPORT'
+          && nqp::istype($root, RakuAST::Name::Part::Expression);
         my str $setting-rev;
         if ($name eq 'CORE') {
             return Nil unless @parts;  # bare CORE has nothing to resolve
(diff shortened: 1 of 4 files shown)


lizmat:
Isn't this an issue with *all* lookups that could happen at compile time, instead of waiting until runtime?

ugexe:
Yes, the underlying rule is that folding a constant indirect lookup is only sound when the folded result is provably identical to what the runtime symbolic lookup would find, and since the runtime lookup checks `%?REQUIRE-SYMBOLS` first, any require injected name that also resolves at compile time diverges. This PR fixes the one case where divergence is unconditional (every unit has its own EXPORT). Whether RakuAST should stop folding indirect lookups entirely, or fold only setting/lexical resolutions when no require is in scope, is a valid question. But it changes behavior for constant T = ::("Int") style code, so it arguably deserves its own decision rather than riding along with this fix.

ugexe:
Superseded by https://github.com/rakudo/rakudo/pull/6528

PR #6480, opened 2026-07-23
Explain the implicit default when an unpassed optional parameter fails to bind

ugexe:
Previously the RakuAST frontend attached a worry to every optional named parameter that has a subset type and no default value:

```
$ RAKUDO_RAKUAST=1 raku -e 'sub f(UInt :$flags) { }'
Potential difficulties:
    Optional named parameter 'flags' with subset type UInt needs a valid default value
    otherwise it will throw exceptions when a value is absent at the callsite
    at -e:1
```

The worry came from #5879 after a discussion about the confusing runtime error you get when such an argument is actually omitted (https://irclogs.raku.org/raku-dev/2025-05-12.html#12:13). The condition discussed there was a missing conformant default, but whether the implicit default conforms cannot be known at compile time without running the where clause, so the check warned on every defaultless case instead. That makes it fire on parameters that bind cleanly when omitted, like UInt, whose refinement accepts its base type object by design. It warns on idiomatic code the legacy frontend compiles silently, it showed up in 23 of 139 module logs in a recent RakuAST blin run, and under `use fatal`, which promotes worries to sorries, it turns into an outright compile failure.

This replaces the worry with feedback at the moment the problem actually occurs. When an optional parameter is not passed and its implicit default fails a post constraint, the error now says where the failing value came from:

```
$ raku -e 'subset X where /"x"/; sub foo(X :$x) { }; foo()'
Constraint type check failed in binding to parameter '$x'; expected X but got Any (Any)
The parameter is optional and was not passed an argument, so its
implicit default value was checked against the constraint. Give the
parameter a default value that satisfies the constraint, mark it as
required, or make the constraint accept the implicit default.
```

Previously that error stopped after the first line, which names a value the caller never passed and gives no hint that a default was involved.

The binder now passes along whether the value being bound is the implicit default of an unpassed optional parameter, and X::TypeCheck::Binding::Parameter exposes it as a new `omitted` attribute. The first line of the message is unchanged, so existing code matching on it keeps working. Both frontends produce the new message since they share the binder. The explanation is not added when a value was passed, even if that value happens to be the bare type object, or when an explicit default fails the constraint.

The worry, its X::Parameter::Named::SubsetTypeWithoutDefault exception class, and its test are removed. Tests for the new message and its negative cases are added to t/05-messages/03-errors.t.


ugexe:
ping @ab5tract

ab5tract:
Looks good to me

PR #6492, opened 2026-07-25
RakuAST: install lowered array init after its operands evaluate

ugexe:
Previously the lowered initialization of a plain array from a comma list bound the fresh reification buffer into the variable before the operand expressions evaluated and reified. For the declaration form that order cannot be observed, but the assignment form covers an operand that reads the target, and @a = |@a, 3 slipped the already-emptied array.

This makes the lowered build follow the STORE method's order: fill the detached buffer from the operands while the variable still holds its old content, then install the buffer and reifier. The reification target wraps whatever buffer the variable holds, so the fresh buffer is swapped in just to mint it and the old one put right back.

### DIFF
--- a/src/Raku/ast/base.rakumod
+++ b/src/Raku/ast/base.rakumod
@@ -1037,9 +1037,13 @@ class RakuAST::Node {
         1
     }
 
-    # The lowered initialization of a plain array from a comma list: bind a
-    # fresh reification buffer and a reifier whose future is the elements,
-    # then reify, the same layout the STORE method builds through dispatch.
+    # The lowered initialization of a plain array from a comma list: fill a
+    # fresh reification buffer from a reifier whose future is the elements,
+    # then install the buffer and reifier, the same layout and order the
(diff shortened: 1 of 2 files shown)


lizmat:
Wouldn't it make more sense to codegen a call to `.append` ?

ugexe:
Maybe... someone could certainly look into it and do a follow up if they wish.  Presumably the tests + Blin can confirm if any follow up changes break anything. This PR fixes `Pretty::Table` (I think... it fixes some module anyway) so that can also be used to test the behavior. Although it'll also need to be benchmarked to ensure it doesn't regress the performance for whatever reason.

I'm going to be away for a week, and when I get back I will presumably still have things that need to be fixed (rather than improved), so it's probably going to be awhile before I personally could get around to addressing e.g. improvements that don't otherwise block the release of RakuAST itself. Hence my aforementioned suggestion of a follow up. Although if there is something functionally incorrect about this PR then we certainly should not take it as-is.

PR #6498, opened 2026-07-30
RakuAST: put "use if" functionality in core

lizmat:
The "if" module https://raku.land/zef:raku-community-modules/if is not really easily portable to RakuAST.  But it *is* quite easy to implement the functionality into RakuAST.  Which is what this commit does:

1. make "if" a pragma
2. in the module loading action, check if the "if" pragma is set
3. if so, check if there are any adverbs
4. if there are, check if the last one is "if"
5. if it is, check if there is a compile time value
6. if there is, and the value is truish, remove that adverb and continue as if nothing has happend
7. if the value is not trueish, codegen an empty statement
8. if there is no compile time value, panic

This leaves the question on what to do if an ":if" adverb is specified, and the "if" pragma is not: currently all unknown adverbs are ignored.  OTOH, *always* checking for adverbs also feel icky from a performance point of view.

With this change, the "if" module can become a no-op for RakuAST.

  use if;
  use Test:if(1);
  ok 1;   # works

  use if;
  use Test:if(0);
  ok 1;   # compilation error, "ok" not known

### DIFF
--- a/src/Raku/Actions.nqp
+++ b/src/Raku/Actions.nqp
@@ -1008,7 +1008,11 @@ class Raku::Actions is HLL::Actions does Raku::CommonActions {
                     my @cp   := $module-name.IMPL-UNWRAP-LIST($colonpairs);
                     my $last := @cp.pop;
                     if $last.key eq 'if' {
-                        my $value := $last.value.literalize;
+                        my $value := Nodify(
+                          "BeginTime"
+                        ).IMPL-BEGIN-TIME-EVALUATE(
+                          $last.value, $*R, $*CU.context
+                        );
                         if nqp::isconcrete($value) {
 
                             # Module should be loaded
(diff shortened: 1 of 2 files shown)


src/Raku/Actions.nqp
```diff
@@ -1000,16 +1000,44 @@ class Raku::Actions is HLL::Actions does Raku::CommonActions {
...
-            );
+            my $LANG        := $*LANG;
+            my $module-name := $<module-name>.ast;
+
+            if $LANG.pragma("if") {
+                if $module-name.colonpairs -> $colonpairs {
+                    my @cp   := $module-name.IMPL-UNWRAP-LIST($colonpairs);
+                    my $last := @cp.pop;
+                    if $last.key eq 'if' {
+                        my $value := $last.value.literalize;
```
lizmat: @ugexe ok, the `literalize` here is not enough, it also needs to be able to execute an expression such as `($*RAKU.version ~~ v6.*` ( which is what the `if` module uses in its tests).

A simple `.EVAL` doesn't  cut it (which is not suprising, but I thought I'd try that anyway).

I seem to hit a dead end here.  Suggestions?

lizmat:
Fixed the issue in module land.

PR #6502, opened 2026-08-02
RakuAST: sink when/default bodies whose succeed payload is unused

### LINKED ISSUE #6074: last statements in CATCH phasers are *silently* discarded
```raku
sub x4 {
  CATCH {
    default {
      42;
      43
    }
  }
  die 'Boom';
}

say x4;

# output:
# Useless use of constant integer 42 in sink context (line 4)
# Nil
```

This will produce a worry about the discarded value `42` but not about `43` which is also discarded (`CATCH` phasers don't produce a value by themselves). It would be great to have a worry for the last statement as well, if it's clearly just a discarded value.

ugexe:
Previously the body of a when or default statement was always compiled as wanted, since its value escapes as the succeed payload and normally becomes the outcome of the block the statement attaches to. That left the last statement of a CATCH or CONTROL handler unwarned even though handlers discard their result, so `CATCH { default { 42; 43 } }` warned about 42 but stayed silent about 43.

The nil-on-succeed flag already marks a scope whose succeed handler sinks the payload and produces Nil, which QUIT phasers set explicitly. This sets it on any scope whose own outcome is discarded: blocks that fully sink their bodies, and sunk sink boundaries such as routines returning Nil and non-EVAL mainlines. when and default now remember the scope they attach to and fully sink their bodies when that scope has no use for the payload, so the trailing statement warns in sink context and gets the usual runtime sink treatment.

Fixes #6074

### DIFF
--- a/src/Raku/ast/code.rakumod
+++ b/src/Raku/ast/code.rakumod
@@ -1624,7 +1624,9 @@ class RakuAST::Block
     }
 
     method propagate-sink(Bool $is-sunk) {
-        $!body.apply-sink($is-sunk && !self.needs-result);
+        my $body-sunk := $is-sunk && !self.needs-result;
+        self.set-nil-on-succeed() if $body-sunk;
+        $!body.apply-sink($body-sunk);
     }
 
     method PRODUCE-IMPLICIT-DECLARATIONS() {
@@ -1929,7 +1931,9 @@ class RakuAST::PointyBlock
     method bare-block() { False }
 
     method propagate-sink(Bool $is-sunk) {
-        self.body.apply-sink($is-sunk && !self.needs-result);
+        my $body-sunk := $is-sunk && !self.needs-result;
+        self.set-nil-on-succeed() if $body-sunk;
+        self.body.apply-sink($body-sunk);
         $!signature.apply-sink(True);
     }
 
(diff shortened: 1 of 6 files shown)


src/Raku/ast/code.rakumod
```diff
@@ -1624,7 +1624,9 @@ class RakuAST::Block
     }
 
     method propagate-sink(Bool $is-sunk) {
-        $!body.apply-sink($is-sunk && !self.needs-result);
+        my $body-sunk := $is-sunk && !self.needs-result;
```
lizmat, line 1627: Wouldn't this be better as a named argument `:set-nil-on-succeed` to `.apply-sink` ?

ugexe: The flag can't ride along on that `apply-sink` call. The receiver there is the Blockoid but nil-on-succeed is LexicalScope state on the Block (since that's where the succeed handler gets emitted) and nodes have no parent pointers to climb back up. Moving it to the outer callers doesn't work either because the decision depends on `needs-result`, which is private to the block (a CATCH body with a KEEP phaser must keep its outcome, so it must not set the flag). It also wouldn't unify the setters because sunk boundaries set the flag in calculate-sink without going through apply-sink at all, and QUIT sets it at BEGIN time on a block that keeps its outcome so the flag and body sinking are independent properties.

PR #4582, opened 2021-10-19
Fixes `require STUB:file<>` bug #4488

tony-o:

PR #6508, opened 2026-08-03
RakuAST: worry about unknown `use` adverbs in 6.e+

lizmat:
As a result of discussion about 'use' adverbs in
  https://github.com/Raku/problem-solving/issues/525

  use Test:foo:ver:bar; # no worries

  use v6.e.PREVIEW; use Test:foo:ver:bar;
  Potential difficulties:
    Detected unsupported adverbs in 'use' statement: :foo:bar

### DIFF
--- a/src/Raku/Actions.nqp
+++ b/src/Raku/Actions.nqp
@@ -1000,10 +1000,29 @@ class Raku::Actions is HLL::Actions does Raku::CommonActions {
 
         # proper module loading
         else {
+            my $module-name := $<module-name>.ast;
+
+            # 6.e or higher and adverbs spotted
+            if nqp::getcomp('Raku').language_revision >= 3
+              && $module-name.colonpairs -> $colonpairs {
+                my constant known-adverbs := nqp::hash(
+                  'ver', 1, 'auth', 1, 'api', 1, 'from', 1
+                );
+                my @unknown;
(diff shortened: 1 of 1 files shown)


ugexe:
We could consider also add a similar check for `require` and `need`.

fwiw adverbs don't work for `require` at all on the legacy frontend, although that also accepts `:name` and `:file`. Having just quickly looked I think the RakuAST logic for adverbs on require probably missed `:name` and `:file` though.

lizmat:
What are `:name` and `:file` supposed to do?  Doesn't appear to have been documented anywhere?

ugexe:
I believe it was to let you name the namespace for plugin type functionality. For example imagine a plugin interface that requires a module to provide a specific function, you could do `require MY-STUB:file($some-runtime-determined-file);` and then you could do `MY-STUB::some_function()` i.e. use `some_function()` without having to know the namespace declared inside of `$some-runtime-determined-file`. It was added in https://github.com/rakudo/rakudo/pull/4582 and references an issue it fixes.

I don't remember anything about `:name` other than it exists, but I assume it was meant to do something similar... perhaps aliasing as another name or something? Or maybe it was intended to give a name if someone does like `require "some/path.rakumod":name<...>` or something (obviously that syntax wouldn't work though)

PR #6509, opened 2026-08-03
Remove return constraints from CompUnit::Repos

lizmat:
If a `require ::($foo)` failed, then the initial execution error would trigger additional execution errors if .resume was called on it.  One for each recursion level in $*REPO.

  my $s; require ::("Foo"); CATCH { ++$s; .resume }; say $s;  # 9

This commit reduces that to 2.  Not sure yet where the last additional execution error is coming from.


lizmat:
Decided to close this, as using .resume in the golf *is* the best thing to use, but not in a loop.  Then `next` is better:
```raku
my @failed;
for <foo bar> -> $m {
    CATCH {
        @failed.push($m);
        next;
    }
    require ::($m);
}
say "Failed: @failed" if @failed;  # Failed: foo bar
```

PR #6513, opened 2026-08-04
Make MAIN recognize enum values that are not in scope

### LINKED ISSUE #5091: imported MAIN can't use enum arguments
Suppose the following module:
kép
with the following file contents
https://gist.github.com/2colours/d85b593f70a0b6e557c61f733f00d9c6

When this module is installed, `demo.raku` can be executed - however, no matter which enum value one is passed, the dispatch will fail apparently, generating the default usage message. The problem persists even when the enum is explicitly exported.

## Environment
* Operating system: `5.4.0-47-generic #51-Ubuntu SMP Fri Sep 4 19:50:52 UTC 2020 x86_64 x86_64 x86_64 GNU/Linux`
* Compiler version (`perl6 -v` or `raku -v`): `Rakudo™ v2022.07 with MoarVM 2022.07`

ugexe:
Previously the command line argument parsing in RUN-MAIN resolved a potential enum value by looking the word up in the caller's lexical scope, GLOBAL::, and CORE::. When MAIN is imported from a module that declares the enum inside its own package none of those lookups can see the enum values, so dispatch would fail and the usage message would be shown instead.

This makes the argument parsing also consult the enum types found in the signatures of the MAIN candidates, which are reachable no matter where the enum was declared. The by-name lookups still take precedence, so behavior for enums that are visible in the caller's scope is unchanged.

Fixes #5091

### DIFF
--- a/src/core.c/Main.rakumod
+++ b/src/core.c/Main.rakumod
@@ -71,6 +71,23 @@ my sub RUN-MAIN(&main, $mainline, :$in-as-argsfiles) {
             }
         }
 
+        # Enum values can only be resolved by name if they are visible in
+        # the caller's scope, which they are not when MAIN was imported
+        # from a module that does not also export the enum.  The enum types
+        # in the signatures of the MAIN candidates are always available, so
+        # collect their values by name for direct lookup.  The first enum
+        # seen wins if candidates use different enums that share a key.
+        my %signature-enum-values;
+        for &main.candidates -> &candidate {
+            for &candidate.signature.params -> $param {
+                my \type := $param.type;
+                if Metamodel::EnumHOW.ACCEPTS(type.HOW) {
+                    %signature-enum-values{.key} //= $_
+                      for type.^enum_value_list;
+                }
+            }
+        }
+
         sub thevalue(\a) {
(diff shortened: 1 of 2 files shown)


src/core.c/Main.rakumod
```diff
@@ -71,6 +71,24 @@ my sub RUN-MAIN(&main, $mainline, :$in-as-argsfiles) {
             }
         }
 
+        # Enum values can only be resolved by name if they are visible in
+        # the caller's scope, which they are not when MAIN was imported
+        # from a module that does not also export the enum.  The enum types
+        # in the signatures of the MAIN candidates are always available, so
+        # values of those can be found through the type instead.
+        my @signature-enums = &main.candidates.map({
```
lizmat: Why is this not set op as a hash, so one would only need to do a hash lookup?

ugexe: Mostly because it isn't something that needs optimization and I was too lazy to look into if there was going to be some unexpected interaction with keying on a hash with an enum this way. Given you bothered to comment at all though makes me think it would probably be fine. I'll make the suggested change

PR #6518, opened 2026-08-05
Fix lookbehind for patterns that cannot be flipped

### LINKED ISSUE #6087: `after` incorrectly handles (flips) "sub-rules"
```
my regex ab { ab }
say .pos given 'ab ba' ~~ / <?after    ab   > /; # 2
say .pos given 'ab ba' ~~ / <?after   <ab>  > /; # 5 <- WRONG!
say .pos given 'ab ba' ~~ / <?after <{'ab'}>> /; # 5 <- WRONG!
say .pos given 'ab ba' ~~ / <?after <{'ba'}>> /; # 2 <- WRONG!
```

[A 2012 Rakudo commit "Implement lookbehind" ](https://github.com/Raku/nqp/commit/3442c181da4df6b57ffa8e90d8f049ddeafe2d7b) involves conditional flipping of qast. That's got to be related inasmuch as something related to that flipping has gone wrong, either from the getgo (though jnthn wrote it, so I doubt it's that), or at some point since.

(I arrived at this golf, and my hint about what seems to be the problem, as an evolution of my initial answer to SO [Why doesn't this look-behind assertion with a predefined regular expression work as expected?](https://stackoverflow.com/a/79906200/1077672).)

ugexe:
The <?after ...> assertion compiles by reversing the pattern syntax tree and matching it against a reversed copy of the target string. That only works for nodes the compiler knows how to reverse. Subrule calls (named rules, interpolated patterns, backreferences), embedded code, and zero-width checks other than anchors were left matching left-to-right against the reversed target, giving wrong results:

    my regex ab { ab }
    say .pos given 'ab ba' ~~ / <?after    ab   > /;  # 2
    say .pos given 'ab ba' ~~ / <?after   <ab>  > /;  # 5, should be 2
    say .pos given 'ab ba' ~~ / <?after <{'ab'}>> /;  # 5, should be 2

Both frontends now only reverse when every node in the lookbehind pattern is reversible, and otherwise emit a call to the new NQP-side after_scan method, which matches the unmodified pattern forward from each position at or before the current one and succeeds if any match ends exactly at the current position.

This also fixes goal matching, character class subtraction, and zero-width peeks such as <?[x]> inside lookbehinds, and the capturing <after ...> form no longer produces a negative-width match object.

Needs the corresponding NQP change to provide after_scan.

Resolves #6087

### DIFF
--- a/src/Perl6/Actions.nqp
+++ b/src/Perl6/Actions.nqp
@@ -11964,9 +11964,16 @@ class Perl6::RegexActions is QRegex::P6Regex::Actions does STDActions {
                     for $<arglist>.ast.list { $qast[0].push(wanted($_, 'assertname')) }
                 }
                 elsif $<nibbler> {
-                    my $nibbled := $name eq 'after'
-                        ?? self.flip_ast($<nibbler>.ast)
-                        !! $<nibbler>.ast;
+                    my $nibbled := $<nibbler>.ast;
+                    if $name eq 'after' {
+                        if self.can_flip_ast($nibbled)
+                          || !nqp::istype($qast[0][0], QAST::SVal) {
+                            $nibbled := self.flip_ast($nibbled);
+                        }
+                        else {
+                            $qast[0][0].value('after_scan');
+                        }
+                    }
                     my $sub := $/.slang_actions('Regex').qbuildsub($nibbled, :anon(1), :addself(1));
                     $qast[0].push($sub);
                 }
--- a/src/Raku/ast/regex.rakumod
+++ b/src/Raku/ast/regex.rakumod
@@ -1464,18 +1464,55 @@ class RakuAST::Regex::Assertion::Named::RegexArg
     }
 
     method IMPL-REGEX-QAST(RakuAST::IMPL::QASTContext $context, %mods) {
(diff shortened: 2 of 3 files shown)


ugexe:
Requires https://github.com/Raku/nqp/pull/856

raiph:
Sorry for spam, but in case my emojis didn't quite express my actual reaction, I felt I needed to add a prose comment (despite it being technically redundant).

> Both frontends now only reverse when every node in the lookbehind pattern is reversible, and otherwise emit a call to the new NQP-side `after_scan` method, which matches the unmodified pattern forward from each position at or before the current one and succeeds if any match ends exactly at the current position.
>
> This also fixes goal matching, character class subtraction, and zero-width peeks such as `<?[x]>` inside lookbehinds, and the capturing `<after ...>` form no longer produces a negative-width match object.

Wow.

patrickbkr:
An alternative could be to error out when unsupported constructs are found. That has the disadvantage of well not supporting those cases, but the advantage, of being easier to reason about. Then there is no "Why is my regex suddenly so slow?"

I'm not entirely opposed, but have mixed feelings.

PR #6524, opened 2026-08-06
Add C99 hexadecimal float literals

dankogai:
Accept C99 hexadecimal floating point literals as Num literals, in both the legacy and the RakuAST frontends:

```raku
say 0x1.8p+1;                  # 3 (a Num)
say -0x1.999999999999ap-4;     # -0.1
say 0x.8p+0;                   # 0.5
```

The binary exponent (`p`/`P`, optionally signed) is mandatory, as in C99, which keeps existing syntax unambiguous: `0x1.abs` remains a method call, `0x10` remains an `Int`, `0x1.8` remains an error. Values are constructed exactly, with a single correctly rounded conversion.

The sprintf `%a`/`%A` counterpart for the 6.e Formatter is in #6525.

### DIFF
--- a/src/Perl6/Actions.nqp
+++ b/src/Perl6/Actions.nqp
@@ -8686,6 +8686,7 @@ Did you mean a call like '"
         if $<integer> {
             make $*W.add_numeric_constant($/, 'Int', $<integer>.ast);
         }
+        elsif $<hex_float>      { make $<hex_float>.ast; }
         elsif $<dec_number>     { make $<dec_number>.ast; }
         elsif $<rad_number>     { make $<rad_number>.ast; }
         elsif $<rat_number>     { make $<rat_number>.ast; }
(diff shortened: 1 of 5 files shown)


ugexe:
Just a couple of suggestions to suppress some claude-isms for potential future PRs:

1) for multi commit PRs the PR description doesn't need to explain each commit individually. that is what the commit messages are for. and indeed listing each commit individually makes it more likely that the PR description will become stale if any of the commits need to be reworked or if new commits are added

2) is mentioning "NQP_REVISION bump" in the PR description useful? not really

3) mentioning things that are the same isn't useful. for example things like "Output is byte-identical ..." should be implied... it wouldn't be useful to mention "this doesn't cause regressions" in every PR because if your PR causes regressions it probably shouldn't be a PR.

4) mentioning the number of tests that were added isn't needed as we can see the diff ourself and can count. Similarly mentioning `make test` works isn't worth mentioning as it is both implied/expected to pass and we can see the CI test for ourselves

To be clear I'm not requesting that you fix these things in this PR, but it would be helpful if they were addressed/enforced in future PRs.

lizmat:
@dankogai: thank you very much!

But could you please create separate PRs for commit 1 and 2?  Commit 3 is not needed.

I feel hexfloats and %a/A formats in sprintf are two different things, and should be judged separately.

dankogai:
@lizmat Done — this PR is now just the hexfloat literals (single commit), the Formatter %a/%A directives moved to their own PR #6525, and the NQP_REVISION bump is dropped. The literal test that round-tripped through sprintf '%a' went with it, since that needs an nqp with %a.

@ugexe thanks for the pointers — applied to both descriptions here, and noted for future PRs.

lizmat:
Thanks!

PR #6525, opened 2026-08-06
Add %a/%A (C99 hex float) directives to the 6.e Formatter

dankogai:
Implement the C99 hexadecimal floating point directives in the RakuAST-based Formatter, as suggested by @lizmat on Raku/nqp#859:

```raku
use v6.e.PREVIEW;
say sprintf('%a', 0.1e0);  # 0x1.999999999999ap-4
```

Semantics match the nqp implementation merged in Raku/nqp#859, except that NaN/Inf follow the Formatter's existing conventions for float directives.

### DIFF
--- a/src/core.e/Formatter.rakumod
+++ b/src/core.e/Formatter.rakumod
@@ -250,6 +250,108 @@ our class Formatter {
         }
     }
 
+    # Set up value in C99 hexadecimal floating point notation, e.g.
+    # 0x1.8p+1 for 3e0.  At this point it is guaranteed that the value
+    # is *not* Inf, -Inf or NaN.  $positions is the number of hex digits
+    # to render after the radix point, with -1 indicating "no precision
+    # given": render exactly, without trailing zeroes.  $point forces a
+    # radix point even without fraction digits (the "#" flag).
+    #
+    # The mantissa hex digits form an integer H with some number of bits
+    # $f shifted into the fraction, so the value is H * 2 ** ($exp - $f).
+    # H is exact in (big)integer arithmetic and rounding to a given
+    # precision is done there as well, ties-to-even, so the rendering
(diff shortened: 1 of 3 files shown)


lizmat:
Marked as draft for now.  Need a clearer mind to look at this PR.  Looks good so far, but on cursory inspection feels like it is over-complicating things.

Will look at it tomorrow again.

lizmat:
@dankogai: I've decided to merge this PR in the knowledge that it is functionally correct afaics, but definitely does not meet *my* standards wrt performance and code clarity.

I will make those changes in followup commits.

One question:
```
$ perl -E 'say sprintf("%f",2/3)'
0.666667
$ perl -E 'say sprintf("%a",2/3)'
0x1.5555555555555p-1
```
I sorta expected the last digit of the fraction to be `6` instead of 5 because of rounding.  Or is that the "even" behaviour you described in other PRs?

dankogai:
@lizmat Thanks for the merge — follow-up commits are more than welcome.

No rounding is involved: those 5s are the double's stored bits, verbatim. That's exactly why bare `%a` exists next to `%.6a` in the spec — with the precision omitted, the output is defined to be the *exact* representation (13 hex digits = all 52 mantissa bits), so nothing is ever rounded at print time. And even `%.6a` stays all-5s here, since the dropped `0101…` tail is always below the halfway point; the ties-to-even rule only decides the exact-half case.

That exactness is the strongest reason to support hexfloat in the first place: a bit-to-bit copy of a double in textual form. We started exchanging doubles and floats across computers — decimal output rounds and drifts, while `%a` round-trips exactly, on every platform.

lizmat:
@dankogai: in case you're interested, these were my tweaks to make it 15% faster: https://github.com/rakudo/rakudo/commit/63de34645e

PR #6531, opened 2026-08-09
Don't count handler calls to test an operator wrap

ugexe:
Previously the wrap test recorded each handler call into an array and compared that array against a single expected entry. A wrap replaces the operator routine for every caller in the process however, so any core code that dispatched `postfix:<++>` between the wrap and the unwrap appended to the array as well. Whether a given site dispatches the routine or runs a raw op depends on precomp state, so the test failed sporadically with hundreds of entries.

This checks that the variable was left alone instead. That is what `use soft` is for. Without it the increment lowers to a raw op and never reaches the wrap. Other callers dispatching the operator during the window no longer affect the result. The handler body is empty now, which also keeps an array push out of the window in which the operator is broken.


ugexe:
Ideally this makes the targeted test less flakey

PR #6534, opened 2026-08-10
RakuAST: Make using a Perl 5 regex a compile time error

ugexe:
Previously the :P5 and :Perl5 regex adverbs selected the Perl 5 regex slang, whose actions stubbed out the entire regex body as an always failing assertion. That let constructs such as m:P5/o+/ and s:P5/o+/i/ compile but silently never match. Perl 5 regexes are not going to be supported by the new grammar.

This makes a truthy :P5 or :Perl5 adverb a compile time X::Syntax::Regex::P5 error on all of the quoting constructs that can carry it. Directly constructed RakuAST nodes that carry a truthy :P5 adverb get the same error when checked.

### DIFF
--- a/src/core.c/Exception.rakumod
+++ b/src/core.c/Exception.rakumod
@@ -2259,6 +2259,13 @@ my class X::Syntax::Regex::Adverb does X::Syntax {
     method message() { "Adverb $.adverb not allowed on $.construct" }
 }
 
+my class X::Syntax::Regex::P5 does X::Syntax {
+    has $.adverb = 'P5';
+    method message() {
+        "Perl 5 regexes are not supported. Please remove the :$.adverb adverb and write a Raku regex"
+    }
+}
+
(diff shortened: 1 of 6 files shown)


ugexe:
This would break the `t/spec/S05-modifier/Perl_*.t` tests. They already skip the tests if it detects they don't work, but the tests are then effectively pointless. Since this adds a compile time error it occurs before that check though.

src/core.c/Exception.rakumod
```diff
@@ -2259,6 +2259,13 @@ my class X::Syntax::Regex::Adverb does X::Syntax {
     method message() { "Adverb $.adverb not allowed on $.construct" }
 }
 
+my class X::Syntax::Regex::P5 does X::Syntax {
```
ugexe, line 2262: We could also get rid of this and throw the same exception we do for any other unknown adverb

ugexe:
This should probably be considered superseded by https://github.com/rakudo/rakudo/pull/6708

PR #6542, opened 2026-08-11
Make the 6.e sprintf produce C compatible output

ugexe:
Previously `sprintf("%f", 1e-5)` gave `0.100000`, `sprintf("%g", 999999.5)` gave `9.99:e+05` with a literal colon, a negative zero lost its sign, `%s` of a type object returned the type object itself, and an explicit parameter index broke the argument count. This makes the output of the float directives match a C printf (at any magnitude and precision) and fixes the `%s` and index behavior to match the sprintf used outside 6.e.

Review commit by commit as each one carries specific details, but to summarize the changes:

- Make the 6.e sprintf render a type object as an empty string. `%s` of a type object now renders as the empty string with the standard warning.
- Make a 6.e sprintf explicit index leave the implicit sequence alone. `sprintf('%2$d %d %d', 1, 2, 3)` now gives `2 1 2`, and a format with explicit indices accepts arguments beyond those its directives consume.
- Make the 6.e sprintf %e produce the exact digits of its value. The digits now come from the exact rational value of the double, so they match C at any precision. This also fixes garbage output for values whose decimal exponent reaches the native integer range, and for values just below a power of ten, which rendered ten times too large with a digit missing.
- Make the 6.e sprintf %f produce the exact digits of its value. The fill zeroes go on the correct side of the digits, a minus sign is never counted as a digit, and ties round to even as C rounds the exact value.
- Rebuild the 6.e sprintf %g on the exponential renderer. The precision counts significant digits, the notation follows the C rule, the alternate form keeps its insignificant zeroes, and a zero value renders through the same runtime path as any other value.

### DIFF
--- a/src/core.e/Format.rakumod
+++ b/src/core.e/Format.rakumod
@@ -116,10 +116,16 @@ my class Format is Str is Callable {
     method render($format, @args) is implementation-detail {
         my &render := nqp::atpos(Formatter.fetch($format),0);
 
-        &render.count == @args.elems
+        # A format with explicit indices takes any number of arguments
+        # beyond the highest index it needs, giving it an infinite count
+        my $count       := &render.count;
+        my int $indexed  = $count == Inf;
+        @args.elems == $count
+          || ($indexed && @args.elems >= &render.arity)
           ?? render(|@args)
           !! X::Str::Sprintf::Directives::Count.new(
-               :$format, :args-have(@args.elems), :args-used(&render.count)
+               :$format, :args-have(@args.elems),
+               :args-used($indexed ?? &render.arity !! $count)
              ).throw
     }
 }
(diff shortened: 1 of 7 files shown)


lizmat:
This is going to require a thorough examination, because I implemented this according to the standard afaik (which the old sprintf in both Perl and Raku definitely were **not**).

So please don't merge this without an explicit ok from me.  Which is also why I changed it to draft.

ugexe:
> This is going to require a thorough examination

I can empathize with wanting to go over this thoroughly

> because I implemented this according to the standard afaik

...however looking at the first sentence of the PR I would hope we could agree that (at a minimum) the existing 6.e formatter does not match any known standard for those examples. For instance:

```
$ raku -e 'use v6.e.PREVIEW; say sprintf("%g", 999999.5)'
9.99:e+05
```

ugexe:
Should we consider merging this yet? rakupp has also used this PR to improve its own sprintf implementation

lizmat:
Thanks for the nudge.  Will look at it tonight and merge if I'm ok with it.

PR #6547, opened 2026-08-12
Make loop controls work in hyper and race iteration

### LINKED ISSUE #2208: `last` with parallel loops NYI, and loop labels disable parallelization
## The Problem

`last` does not work in `hyper for`/`race for` (silently ignored, probably NYI). Furthermore, adding a loop label to such a loop disables parallelization.

## Expected Behavior

* `last` works in `hyper`/`race` `for` loops
* Loop labels on such loops don't disable parallelization

## Actual Behavior

* `last` ignored
* Removing label is the difference between a program using 130% CPU and 930% CPU (and so running in appreciably less wallclock time :-))

## Environment
* Operating system: Linux
* Compiler version (`perl6 -v`): This is Rakudo version 2018.06-430-g6587bf7 built on MoarVM version 2018.06-419-g829b8ee implementing Perl 6.c.

ugexe:
Previously a `last` inside a hyper or race operation was swallowed by the batch processing loop. It truncated the batch it happened in while the iteration itself carried on, and a labeled loop control escaped its worker thread as a "labeled last without loop construct" error. Passing a loop label to such a map also forced the sequential fallback, so labeling a `hyper for` or `race for` loop quietly disabled its parallelization. Under RakuAST a label on a prefixed for loop additionally ended up on the wrapping expression statement instead of the loop, so `last FOO` died outright.

This teaches the parallel mapper and grepper the same control exception handling their serial counterparts use, including the loop label for the mapper and the control exception payload semantics of both, such as `last 42` in a map and `last True` verdicts in a 6.e grep. A `next` skips the value, a `redo` reruns the block, and a `last` truncates the batch at the control point and marks the iteration stopped. The pipeline reacts to the stop by no longer producing batches and by emptying already produced batches past the stop. The joiners complete once every batch up to the stop has been consumed and drop later in-flight batches.

An ordered hyper map or grep ended by `last` produces exactly the values its serial counterpart would. A race also emits values from batches that were already delivered when the stop arrived, which race semantics permit. An error a worker raises in speculative work after a stop completed the iteration is dropped, as the serial equivalent would never have run that work. An error that arrives before completion surfaces normally.

The RakuAST frontend now attaches a label on a prefixed for loop to the loop itself, matching the plain `FOO: for @a { }` form. The legacy frontend already passed the label into the map call and only needed the runtime side to accept it.

Fixes #2208

### DIFF
--- a/src/Raku/Actions.nqp
+++ b/src/Raku/Actions.nqp
@@ -671,7 +671,16 @@ class Raku::Actions is HLL::Actions does Raku::CommonActions {
         # Setting label on already created statement
         if $<label> {
             my $ast := $<statement>.ast;
-            $ast.add-label($<label>.ast);
+
+            # A for loop with a statement prefix such as `hyper` parses as
+            # an expression statement wrapping the for loop; the label
+            # belongs on the loop so that loop controls can target it.
+            my $target := $ast;
+            if nqp::istype($ast, Nodify('Statement::Expression'))
+                && nqp::istype($ast.expression, Nodify('Statement::For')) {
+                $target := $ast.expression;
+            }
+            $target.add-label($<label>.ast);
             make $ast;
             return;       # nothing left to do here
         }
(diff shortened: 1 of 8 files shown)


lizmat:
FWIW, I think the whole current hyper / race implementation needs to be revisited, and possibly ripped out, as it does not allow for any backpressure.

So I'm ok with merging this without any further review.

PR #6559, opened 2026-08-14
RakuAST: generate the POPULATE method from the build plan

ugexe:
Previously the RakuAST frontend did not implement the metamodel's generate_buildplan_executor hook, so every class it compiled constructed objects through Mu.POPULATE interpreting the BUILDALLPLAN on each call, while the legacy frontend has generated a per class POPULATE submethod at class composition for years. This implements the hook on RakuAST::CompilerServices, compiling the plan into a POPULATE submethod the same way. That takes object construction on the RakuAST frontend from about 4x slower than the legacy frontend to parity or near parity, for user classes and for most classes in the core setting. Classes composed before the needed setting symbols have compiled, as during the early setting bootstrap, and classes composed at runtime through the MOP keep the interpreting fallback. Benchmarks and the reasoning behind the approach follow.

**Relation to the plan to get rid of BUILDPLAN**

There has been talk of eventually making the BUILDPLAN obsolete by giving each class a POPULATE method built from RakuAST at composition time. This change doesn't conflict with that. It implements the same metamodel hook the legacy frontend has implemented for years, so if that plan ever lands both implementations get removed at the same time. In the meantime the BUILDPLAN stays the single source of truth for construction semantics, which is what keeps both frontends constructing objects the same way. Classes composed at runtime through the MOP keep the interpreting fallback, which they need under either approach.

**Why not just get rid of the BUILDPLAN instead**

The BUILDPLAN is what the metamodel computes from the MRO, composed roles, defaults, typed and native attributes, and BUILD/TWEAK submethods. Generating POPULATE from anything else means deriving all of that a second time, and anywhere the two derivations disagree is a silent construction bug. There is also no answer yet for classes composed at runtime through the MOP, which have no compiler available at composition time. Until there is one the BUILDPLAN and its interpreter have to stay anyway, so removing it isn't currently possible. Compiling the plan gets the construction cost down now and removes cleanly if that ever changes.

**Prior discussion**

nine went over this on #raku-dev back in March 2025 when getting rid of the BUILDPLAN came up [here](https://irclogs.raku.org/raku-dev/2025-03-19.html#14:03): "It only works on the old frontend because of generate_buildplan_executor. The bug is in Mu.POPULATE", "There's still gonna be the bootstrapping issue and the case where a class is purely generated via MOP at runtime", and "Feeding from generated meta-object is not enough. The MOP would still need access to the compiler. After all, the whole point of CompilerServices is to give the MOP access to the compiler". That's the approach taken here: the MOP gets the compiler through CompilerServices when there is one, and falls back to interpreting the plan when there isn't.

### DIFF
--- a/src/Raku/ast/code.rakumod
+++ b/src/Raku/ast/code.rakumod
@@ -3661,6 +3661,439 @@ class RakuAST::Method::AttributeAccessor
     }
 }
 
+# The generated POPULATE method: a flattened compile of the class's
+# BUILDALLPLAN, which Mu.POPULATE otherwise interprets per object
+# construction. Created by RakuAST::CompilerServices when the
+# metamodel composes a class, mirroring what the legacy frontend's
+# generate_buildplan_executor installs. The setting values the plan
+# needs arrive resolved through the constructor, since this file
+# cannot name them and composition may run before they exist.
+class RakuAST::Submethod::BuildPlanExecutor
+  is RakuAST::Submethod
+{
+    has Mu $!package-type;
+    has Mu $!build-plan;
+    has Mu $!True;
+    has Mu $!Failure;
+    has Mu $!X-Attribute-Required;
+    has Mu $!return-routine;
+
+    method new(Mu :$package-type, Mu :$build-plan, Mu :$True,
+            Mu :$Failure, Mu :$X-Attribute-Required, Mu :$return-routine) {
+        my $obj := nqp::create(self);
+        nqp::bindattr_s($obj, RakuAST::Declaration, '$!scope', 'has');
+        nqp::bindattr_s($obj, RakuAST::Routine, '$!multiness', '');
+        nqp::bindattr($obj, RakuAST::Method, '$!private', False);
+        nqp::bindattr($obj, RakuAST::Method, '$!meta', False);
+        nqp::bindattr($obj, RakuAST::Routine, '$!need-routine-variable', False);
+        nqp::bindattr($obj, RakuAST::Method, '$!body', RakuAST::Blockoid.new);
+        nqp::bindattr($obj, RakuAST::Routine, '$!signature', RakuAST::Signature.new);
+        nqp::bindattr($obj, RakuAST::Routine, '$!name',
+            RakuAST::Name.from-identifier('POPULATE'));
+        nqp::bindattr($obj, RakuAST::Submethod::BuildPlanExecutor, '$!package-type', $package-type);
+        nqp::bindattr($obj, RakuAST::Submethod::BuildPlanExecutor, '$!build-plan', $build-plan);
+        nqp::bindattr($obj, RakuAST::Submethod::BuildPlanExecutor, '$!True', $True);
+        nqp::bindattr($obj, RakuAST::Submethod::BuildPlanExecutor, '$!Failure', $Failure);
+        nqp::bindattr($obj, RakuAST::Submethod::BuildPlanExecutor, '$!X-Attribute-Required', $X-Attribute-Required);
+        nqp::bindattr($obj, RakuAST::Submethod::BuildPlanExecutor, '$!return-routine', $return-routine);
+        $obj
+    }
+
+    # The body is fully synthetic: nothing in it can reach the special
+    # variables, so only self is declared.
+    method PRODUCE-IMPLICIT-DECLARATIONS() {
+        [ RakuAST::VarDeclaration::Implicit::Self.new() ]
+    }
+
+    # The block binds the invocant and the initialization hash as raw
+    # locals with no binder run, the shape the legacy frontend
+    # installs. The definite invocant type on the signature informs
+    # introspection, not a run time check. The lexical %_ stays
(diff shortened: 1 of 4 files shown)


ugexe:
# Generated POPULATE benchmarks

Object construction timings for the branch that makes the RakuAST frontend generate a per-class `POPULATE` method from the build plan at class composition time, instead of falling back to `Mu.POPULATE` interpreting the build plan on every object construction. The legacy frontend has generated this method for years; this closes that gap.

## Configurations

| Column | User code compiled by | CORE setting compiled by |
|---|---|---|
| branch RakuAST | RakuAST frontend | RakuAST frontend |
| branch legacy | legacy frontend | RakuAST frontend |
| 2026.07 RakuAST | RakuAST frontend | legacy frontend |
| 2026.07 legacy | legacy frontend | legacy frontend |

The branch build compiles its setting with `RAKUDO_RAKUAST=1 make install`, so both of its columns run on a RakuAST-built CORE. The 2026.07 release ships a legacy-built CORE. That matters when comparing legacy columns across builds: differences there come from the setting build frontend, not from this change.

To reproduce, save a snippet as `bench.raku` and run:

```sh
# branch RakuAST
RAKUDO_RAKUAST=1 raku bench.raku
# branch legacy
raku bench.raku
# 2026.07 via rakubrew
RAKUDO_RAKUAST=1 rakubrew exec --with moar-2026.07 raku bench.raku
rakubrew exec --with moar-2026.07 raku bench.raku
```

## Results

Times in seconds, single run on an otherwise idle machine.

### 1. Plain attributes, 2M constructions

```raku
class P { has $.x; has $.y = 2; has @.t }
my $t = now; my $o; $o := P.new(x => $_) for ^2_000_000; say now - $t
```

| 2026.07 RakuAST | branch RakuAST | branch legacy | 2026.07 legacy |
|---|---|---|---|
| 2.58 | 0.60 | 0.54 | 0.48 |

### 2. Defaults including a closure reading self, 1M constructions

```raku
class D { has $.a = 1; has $.b = 2; has $.c = self.a + self.b; has @.l = 1, 2 }
my $t = now; my $o; $o := D.new for ^1_000_000; say now - $t
```

| 2026.07 RakuAST | branch RakuAST | branch legacy | 2026.07 legacy |
|---|---|---|---|
| 2.75 | 1.85 | 1.47 | 0.76 |

### 3. Native attributes, 2M constructions

```raku
class N { has int $.i = 3; has num $.n = 2e0; has str $.s = "d" }
my $t = now; my $o; $o := N.new(i => $_) for ^2_000_000; say now - $t
```

| 2026.07 RakuAST | branch RakuAST | branch legacy | 2026.07 legacy |
|---|---|---|---|
| 2.16 | 0.56 | 0.38 | 0.34 |

### 4. Attributes plus TWEAK, 1M constructions

```raku
class T { has $.x; has $.double; submethod TWEAK { $!double = ($!x // 0) * 2 } }
my $t = now; my $o; $o := T.new(x => $_) for ^1_000_000; say now - $t
```

| 2026.07 RakuAST | branch RakuAST | branch legacy | 2026.07 legacy |
|---|---|---|---|
| 2.14 | 1.40 | 1.22 | 1.15 |

### 5. Pair.new control, 2M constructions

```raku
my $t = now; my $o; $o := Pair.new(:key("k"), :value($_)) for ^2_000_000; say now - $t
```

| 2026.07 RakuAST | branch RakuAST | branch legacy | 2026.07 legacy |
|---|---|---|---|
| 0.27 | 0.29 | 0.30 | 0.27 |

`Pair` has a hand written constructor that never reaches `POPULATE`, so this row is a no regression control and stays flat by design.

### 6. Empty class, 2M constructions

```raku
class E { }
my $t = now; my $o; $o := E.new for ^2_000_000; say now - $t
```

| 2026.07 RakuAST | branch RakuAST | branch legacy | 2026.07 legacy |
|---|---|---|---|
| 0.35 | 0.11 | 0.10 | 0.05 |

## Reading the results

The RakuAST frontend went from about 4x slower than legacy to parity or near parity on the construction paths `POPULATE` owns: plain attributes 2.58 to 0.60, natives 2.16 to 0.56, empty classes 0.35 to 0.11, each landing next to the legacy frontend on the same setting build. The defaults and TWEAK shapes improved about 1.5x and sit close to their same setting legacy numbers. Where a branch legacy number trails its 2026.07 legacy counterpart, both frontends on the branch build show the same distance, which points at the shared RakuAST-built setting rather than at this change.

MasterDuke17:
Any idea why 2026.07 seems quite a bit faster for legacy?

ugexe:
> Where a branch legacy number trails its 2026.07 legacy counterpart, both frontends on the branch build show the same distance, which points at the shared RakuAST-built setting rather than at this change.

MasterDuke17:
> > Where a branch legacy number trails its 2026.07 legacy counterpart, both frontends on the branch build show the same distance, which points at the shared RakuAST-built setting rather than at this change.

Ah, I see. This is where I wish we had a nice set of (micro)benchmarks to point out specific areas that are slower.

PR #6567, opened 2026-08-15
Intern curried roles created by generic instantiation

### LINKED ISSUE #4097: Dispatch with native arrays unnecessary slow
$ raku -e 'sub a(int @i) { }; my int @j = ^10; a(@j) for ^500000'
    real	0m0.795s
    user	0m0.813s
    sys	0m0.029s

    $ raku -e 'sub a(@i) { }; my int @j = ^10; a(@j) for ^500000'
    real	0m0.133s
    user	0m0.151s
    sys	0m0.022s

The `sub a(int @i)` version is so much slower because **80%** of the time is spent in https://github.com/rakudo/rakudo/blob/master/src/Perl6/Metamodel/CurriedRoleHOW.nqp#L197

Am not sure if this is a recent performance regression or not.

ugexe:
Previously CurriedRoleHOW.instantiate_generic built its instantiated curry with CurriedRoleHOW.new_type. Instantiating the Positional[TValue] that Array::Typed does thus produced a Positional[Int] distinct from the interned one that writing Positional[Int] in source returns. Type-check caches match on object identity, so a typed array never got a cache hit when checked against Positional[Int] and every such check fell back to CurriedRoleHOW.accepts_type. That made calls to a routine with a typed array parameter some 40x slower than the untyped equivalent.

This delegates to the curried role's own parameterize method, which for role groups interns through the 6model parameterization cache. The instantiated curry is then the same object as any directly written parameterization. Type-check caches hit and dispatch with typed arrays runs at the same speed as with untyped ones. The pun repr of the role group now also carries over to such curries.

Fixes #4097

### DIFF
--- a/src/Perl6/Metamodel/CurriedRoleHOW.nqp
+++ b/src/Perl6/Metamodel/CurriedRoleHOW.nqp
@@ -234,11 +234,11 @@ class Perl6::Metamodel::CurriedRoleHOW
                   $value.HOW.instantiate_generic($value, $type_env)
                 ) if $value.HOW.archetypes($value).generic;
             }
-            self.new_type($!curried_role, |@new_pos, |%new_named)
+            $!curried_role.HOW.parameterize($!curried_role, |@new_pos, |%new_named)
         }
 
         else {
-            self.new_type($!curried_role, |@new_pos)
+            $!curried_role.HOW.parameterize($!curried_role, |@new_pos)
         }
     }
 
--- a/t/02-rakudo/curried-role-instantiation.t
+++ b/t/02-rakudo/curried-role-instantiation.t
@@ -1,0 +1,37 @@
+use Test;
+
+plan 5;
+
+# Instantiating a generic curried role must preserve everything the curry
+# was declared with: named and value arguments have to survive into the
+# composed class, and the pun repr of the role group has to reach the
+# instantiated curry.
+
+{
+    role Flagged[::T, :$flag] { method flag() { $flag } }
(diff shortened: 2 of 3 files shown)


ugexe:
```
$ raku -e 'sub f(@i) { }; my @j = ^10; f(@j); my $s = now; f(@j) for ^2_000_000; printf "%.1f ns/call\n", (now - $s) / 2e6 * 1e9'
legacy:  20.8 ns/call
rakuast: 32.0 ns/call
```

```
$ raku -e 'sub f(Int @i) { }; my Int @j = ^10; f(@j); my $s = now; f(@j) for ^2_000_000; printf "%.1f ns/call\n", (now - $s) / 2e6 * 1e9'
legacy:  21.4 ns/call   (was 931.6 on main)
rakuast: 32.0 ns/call   (was 964.8 on main)
```

```
$ raku -e 'sub f(Str @i) { }; my Str @j = "a".."j"; f(@j); my $s = now; f(@j) for ^2_000_000; printf "%.1f ns/call\n", (now - $s) / 2e6 * 1e9'
legacy:  21.2 ns/call   (was 932.3 on main)
rakuast: 32.9 ns/call   (was 943.5 on main)
```

```
$ raku -e 'sub f(int @i) { }; my int @j = ^10; f(@j); my $s = now; f(@j) for ^2_000_000; printf "%.1f ns/call\n", (now - $s) / 2e6 * 1e9'
legacy:  21.1 ns/call   (was 949.7 on main)
rakuast: 32.4 ns/call   (was 948.0 on main)
```

PR #6573, opened 2026-08-16
Share the revision gate mixin across gated routines

ugexe:
Previously every `is revision-gated` application parameterized the is-revision-gated role with a freshly computed Int. Parameterizations intern by argument identity, and during the setting build each site's revision Int is a distinct object, so every gated routine minted its own curried role, mixin type, and mixin cache type, all serialized into the setting and deserialized again on every startup.

This keeps one parameterization per language revision in a constant table so every gated routine with the same revision shares its mixin type, falling back to direct parameterization for revisions beyond the table.


ab5tract:
Nice catch!

PR #6574, opened 2026-08-19
Report (at compile time) trying to store a wrong value in a typed variable

ugexe:
Previously a store of a value the variable's declared type can never accept, e.g. `my Int $x = "foo"` or `my Str @a = 1,2,3`, compiled fine and only failed at run time, and only if that code was reached. The compiler already knows both the value and the type in these cases, so this makes it a compile time error on both frontends. That covers item and list variables, declaration initializers and later assignments, constants, enums, natives, and parameters that get a container of their own (`is copy`).

The check only reports when it is sure. A bind replaces the container a declaration made, so a variable a bind targets is left alone regardless of where in the unit the bind is written, and a bind through a pseudo package naming a lexical scope (`MY::<$x> := ...`, `$CALLER::x := ...`) stands the whole unit down. Subsets, roles, generics, coercion and definite types, a container with its own `STORE`, hashes, and anything whose value is not known at compile time are left to the run time check as before. There is one knowingly imprecise corner: an `is rw` parameter stores into the caller's container, which may be wider than the declared type, however `sub f(Str $x is rw) { $x = 43 }` is reported anyway since that is a bug in practically every case.

`X::Syntax::Number::LiteralType` was previously only for numeric literals into numeric natives. It is now the compile time counterpart of `X::TypeCheck::Assignment` (a subclass of it), answers `.got`, `.expected` and `.symbol`, and names a native as declared along with the type it unboxes from. The last commit removes a native subscript rule the optimize pass had of its own, since the bind mark the check leaves already covers it.

### DIFF
--- a/src/Perl6/Actions.nqp
+++ b/src/Perl6/Actions.nqp
@@ -7817,6 +7817,21 @@ Did you mean a call like '"
         ).annotate_self('smartmatch', 1);
     }
 
+    # A bind through a pseudo package naming a lexical scope replaces a
+    # container this analysis never resolves, so no store in the unit can be
+    # judged. Whether the leading part names such a scope rides on the lookup
+    # symbol_lookup builds, on the package the stash is taken from for a
+    # subscripted bind and on the whole lookup for a sigil qualified one.
+    sub note_pseudo_package_bind($/, $subscript) {
+        $*W.mark_pseudo_package_bind
+            if $subscript.ann('lexical_pseudo_package')
+            || nqp::elems($subscript)
+            && nqp::istype($subscript[0], QAST::Op)
+            && $subscript[0].op eq 'who'
+            && nqp::elems($subscript[0])
+            && $subscript[0][0].ann('lexical_pseudo_package');
+    }
+
     sub bind_op($/, $target, $source, $sigish) {
         my $world := $*W;
         # Check we know how to bind to the thing on the LHS.
@@ -7875,6 +7890,10 @@ Did you mean a call like '"
                 if $world.is_lexical_marked_ro($target.name) || !$was_lexical {
                     $world.throw($/, ['X', 'Bind', 'Rebind'], target => $target.name);
                 }
+
+                # A bind replaces the container the declaration made, so no
+                # store into this variable has to satisfy the type it named.
+                $world.mark_lexical_bind_targeted($target.name);
             }
 
             # Finally, just need to make a bind.
@@ -7885,6 +7904,7 @@ Did you mean a call like '"
                 ((my $target_0_name := $target[0].name) eq '&postcircumfix:<[ ]>' ||
                  $target_0_name eq '&postcircumfix:<{ }>' ||
                  $target_0_name eq '&postcircumfix:<[; ]>') {
+            note_pseudo_package_bind($/, $target[0]);
             $source.named('BIND');
             $target[0].push($source);
             $target.nosink(1);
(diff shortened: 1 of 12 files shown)


ugexe:
Some before and after to show what this catches. Before is `v2026.07-152-g94ae304e51`, after is this branch.

An item store. The obvious case, but note the error now points at the value and suggests what to do about it:

```
$ raku -e 'my Int $x = "foo"'
Type check failed in assignment to $x; expected Int but got Str ("foo")
  in block <unit> at -e line 1

$ raku -e 'my Int $x = "foo"'
Cannot assign a literal of type Str ("foo") to a variable ($x) of type
Int. You can declare the variable to be of type Cool, or try to coerce
the value with "foo".Int or Int("foo").
at -e:1
------> my Int $x = "<HERE>foo"
```

A list store, which is judged element by element:

```
$ raku -e 'my Str @a = 1,2,3'
Type check failed for an element of @a; expected Str but got Int (1)
  in block <unit> at -e line 1

$ raku -e 'my Str @a = 1,2,3'
Cannot assign a literal of type Int (1) to a variable (@a) of type Str.
You can declare the variable to be of type Cool, or try to coerce the
value with 1.Str or Str(1), or just write the value as "1".
at -e:1
------> my Str @a = <HERE>1,2,3
```

The real value is code that never ran. Previously this was a bug waiting for the first caller:

```
$ raku -e 'sub f(Str $x is copy) { $x = 43 }; say "f was never called"'
f was never called

$ raku -e 'sub f(Str $x is copy) { $x = 43 }; say "f was never called"'
Cannot assign a literal of type Int (43) to a variable ($x) of type
Str. You can declare the variable to be of type Cool, or try to coerce
the value with 43.Str or Str(43), or just write the value as "43".
at -e:1
------> sub f(Str $x is copy) { $x = <HERE>43 }; say "f was never called"
```

Same for a branch that did not happen to be taken:

```
$ raku -e 'if False { my Int $x = "foo" }; say "dead branch never ran"'
dead branch never ran

$ raku -e 'if False { my Int $x = "foo" }; say "dead branch never ran"'
Cannot assign a literal of type Str ("foo") to a variable ($x) of type
Int. You can declare the variable to be of type Cool, or try to coerce
the value with "foo".Int or Int("foo").
at -e:1
------> if False { my Int $x = "<HERE>foo" }; say "dead branch never ran"
```

A native used to give an unbox error that named neither the variable nor the value:

```
$ raku -e 'my int $x = "foo"'
This type cannot unbox to a native integer: P6opaque, Str
  in block <unit> at -e line 1

$ raku -e 'my int $x = "foo"'
Cannot assign a literal of type Str ("foo") to a native variable ($x)
of type int. You can declare the variable to be of type Cool, or try to
coerce the value with "foo".Int or Int("foo").
at -e:1
------> my int $x = "<HERE>foo"
```

A constant is a value the compiler already holds, and an enum is a type like any other:

```
$ raku -e 'constant Half = 1/2; my Str $x = Half'
Type check failed in assignment to $x; expected Str but got Rat (0.5)
  in block <unit> at -e line 1

$ raku -e 'constant Half = 1/2; my Str $x = Half'
Cannot assign a literal of type Rat (0.5) to a variable ($x) of type
Str. You can declare the variable to be of type Cool, or try to coerce
the value with 0.5.Str or Str(0.5), or just write the value as "0.5".
at -e:1
------> constant Half = 1/2; my Str $x = <HERE>Half

$ raku -e 'enum Colour <white>; my Colour $c = "nope"'
Type check failed in assignment to $c; expected Colour but got Str ("nope")
  in block <unit> at -e line 1

$ raku -e 'enum Colour <white>; my Colour $c = "nope"'
Cannot assign a literal of type Str ("nope") to a variable ($c) of type
Colour. You can declare the variable to be of type Cool.
at -e:1
------> enum Colour <white>; my Colour $c = "<HERE>nope"
```

PR #6575, opened 2026-08-19
Clean up all generated build artifacts on make clean

ugexe:
Previously make clean removed the bytecode under blib/Perl6 but left blib/Raku in place, so Grammar and Actions survived a clean and could be picked up by a later build. CLEANUPS_ALL now covers BLIB_RAKUDO_R alongside BLIB_RAKUDO, and the js backend cleanup list gains the matching source map entry.

gen/build_rakudo_home is created by the all target and holds the RAKUDO_HOME used during the build, including its core, site and vendor precomp stores. It was only removed by realclean. It moves to clean, which realclean already runs.


ab5tract:
Looks good to me

PR #6576, opened 2026-08-20
Answer reductions over endless sequences

### LINKED ISSUE #1210: [max] doesn't like big Iterables
m: say [max] (1..100000)
rakudo-moar 97b11edd6: OUTPUT: «Too many arguments in flattening array.␤  in block <unit> at <tmp> line 1␤␤»

ugexe:
Previously a reduction folded its values one at a time until they ran out, so a reduction over values that never run out never returned. Some reductions have their answer well before the end, and this teaches them to take it.

    say [+] 0..Inf;                  # Inf   (previously never returned)
    say [min] 0..Inf;                # 0
    say [max] 0..Inf;                # Inf
    say [||] flat(0, 0, 5, 1..Inf);  # 5
    say [min] 1..1000000;            # 1     (previously "argument lists are limited to 65535")

Two independent mechanisms do the work.

**Short-circuit operators settle.** `[||]` and `[or]` stop folding once the result is true, `[&&]` and `[and]` once it is false, `[//]` once it is defined. This works on any sequence, not just Ranges. The settle test runs only after a further value has been pulled, which is exactly when the operator itself would have inspected the result, so a Failure the fold would have left unhandled stays unhandled and a Junction result comes through untouched.

**Endless Ranges answer from their endpoints.** `[+]` takes the answer `Range.sum` gives. `[min]` takes the first value produced. `[max]` is `Inf` for a Range climbing towards `Inf`. These answers are taken only where the values are known to back them. The Range must be exactly `Range`, so a subclass or mixin producing its own values folds. The start must be a type whose stepping is known: an `Int` or `Rational` climbs by one without bound, while a `Num` stops climbing once a step of one is below its precision and a `Str` rolls over past `z` to `aa`, producing values below its start, so those fold like before.

Keeping the endless check off the common path led to a direct route for the most common argument, a sole fully reified list, and two things fell out of it.

Performance, against the same tree without this change: `[min] $a, $b` about 2.2x faster, `[+] $a, $b` about 3.6x, `[+]` over ten values about 5x, `[min]`/`[max]` over a hundred values about 20% faster, `[min] 1..100` about 50% faster.

The argument list limit. `[min]`, `[max]` and `[+]` over more values than an argument list holds previously died with the 65535 limit error. A fully reified list is now handed to the operator whole, so they answer. A `Slip` among the values still stands for what it holds, so a list holding one keeps spreading, and values still being produced keep pulling, exactly as before.

The first commit is a prerequisite. Summing an endless Range asks `int-bounds` about endpoints with no integer value, and the Failure that produced was never marked handled, so the garbage collector reported it at exit. `Inf` and `NaN` endpoints are now ruled out before `.Int` is asked, and anything else that fails there is marked handled.

`[*]`, `[~]`, `[,]` and the list-associative short-circuits like `[orelse]` still fold forever over endless values. They have no answer to take before the end.

Fixes #1210

### DIFF
--- a/src/core.c/Range.rakumod
+++ b/src/core.c/Range.rakumod
@@ -415,14 +415,28 @@ my class Range is Cool does Iterable does Positional {
             True
           ),
           nqp::if(
+            # .Int hands back a Failure for an endpoint with no integer
+            # value, such as a NaN, an Inf, or a Rational with a zero
+            # denominator.  Ruling the first two out keeps that Failure from
+            # being made, and marking any other one handled keeps it off the
+            # garbage collector's report.  The bounds themselves come from
+            # .floor, which rounds down where .Int truncates towards zero.
             nqp::istype($!min,Real)
               && $!min.floor == $!min
               && nqp::istype($!max,Real)
-              && nqp::istype($!min.Int, Int)  # exclude NaN and Infs, who will fail() here
-              && nqp::istype($!max.Int, Int),
+              && nqp::isfalse(nqp::istype($!min,Num) && nqp::isnanorinf($!min))
+              && nqp::isfalse(nqp::istype($!max,Num) && nqp::isnanorinf($!max))
+              && nqp::if(
+                   nqp::isconcrete(my \low := $!min.Int) && nqp::istype(low,Failure),
+                   nqp::stmts((low.Failure::handled = True), 0),
+                   nqp::istype(low,Int))
+              && nqp::if(
+                   nqp::isconcrete(my \high := $!max.Int) && nqp::istype(high,Failure),
+                   nqp::stmts((high.Failure::handled = True), 0),
+                   nqp::istype(high,Int)),
             nqp::stmts(
               ($from = $!min.floor + $!excludes-min),
-              ($to   = $!max.floor - ($!excludes-max && $!max.Int == $!max)),
+              ($to   = $!max.floor - ($!excludes-max && high == $!max)),
               True
             ),
             False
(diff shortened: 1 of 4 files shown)


japhb:
Technically `[*]` could end at first zero value, since the result will be zero thereafter.

ugexe:
> Technically [*] could end at first zero value, since the result will be zero thereafter.

`[*] 0, Inf` is NaN, not 0, so a 0 can't stop the reduction early. Only the operators that never evaluate their right side (`||` `&&` `//`) can skip values without changing the answer.

same applies to `+` fwiw. Inf looks absorbing but `[+] Inf, -Inf` is NaN, and while NaN really does absorb every numeric, `+` still evaluates its right side so `[+] NaN, "abc"` has to die. that's why the `[+] 0..Inf` case doesn't watch the accumulator at all, it takes the answer from the Range's endpoints where every value is known up front.

japhb:
Oh subtle.  I was thinking "Well then what about stopping at NaN", but you answered that pretty solidly.

MasterDuke17 (commented):
Hm, shouldn't something like `say [+] 2..1000000` trigger the `.sum` optimization? I see ~1.67s for that form, but ~0.086s for `say (2..1000000).sum`.

MasterDuke17:
> Hm, shouldn't something like `say [+] 2..1000000` trigger the `.sum` optimization? I see ~1.67s for that form, but ~0.086s for `say (2..1000000).sum`.

Same after rebuilding after https://github.com/rakudo/rakudo/pull/6580 was merged.

ugexe:
> Hm, shouldn't something like say [+] 2..1000000 trigger the .sum optimization? I see ~1.67s for that form, but ~0.086s for say (2..1000000).sum.

I suppose it could be optimized further by special casing the integer case. the `.sum` optimiztation can't be used in the general case though:

```
➜  rakudo git:(main) raku -e 'say (2e0..5e0).sum.raku'
14
➜  rakudo git:(main) raku -e 'say ([+] 2e0..5e0).raku'
14e0
```

MasterDuke17:
> I suppose it could be optimized further by special casing the integer case. the `.sum` optimiztation can't be used in the general case though:
> 
> ```
> ➜  rakudo git:(main) raku -e 'say (2e0..5e0).sum.raku'
> 14
> ➜  rakudo git:(main) raku -e 'say ([+] 2e0..5e0).raku'
> 14e0
> ```

Ugh. Should that difference really exist?

ugexe:
another thing that prevents it being applied more generally is `[op] a, b, c` means `(a op b) op c` whereas with e.g. `.sum` i assume we just use whatever the best algorithm is

MasterDuke17:
> another thing that prevents it being applied more generally is `[op] a, b, c` means `(a op b) op c` whereas with e.g. `.sum` i assume we just use whatever the best algorithm is

Floats are obviously trickier since addition on them isn't commutative, but yeah, for int ranges we just directly compute the value (https://github.com/rakudo/rakudo/blob/main/src/core.c/Range.rakumod#L737).

PR #6578, opened 2026-08-23
RakuAST: re-form BEGIN-compiled code with the optimize walk's marks

ugexe:
A routine invoked at BEGIN time compiles ahead of the unit's optimize walk and caches its QAST block. Previously the unit emission reused that cache, so such routines ran at runtime with none of the marks the walk decides. The setting machinery the compiler itself executes while building CORE is exactly such code, so a hot iterator method paid an operator call through an attribute reference where the same method compiled outside a BEGIN use runs a native increment.

The code node now records that its block was formed early. Once the optimize walk has run, the first request for the block re-forms it from the marked AST and grafts the result into the same QAST::Block object, so every registration made against the block keeps resolving. Closures serialized by the early compilation pin the frame's lexical shape, so the unused elisions stand down during the re-formation and the lowering walk does not approve flatten for frames under a code object that will re-form. Emission that appends to a cached block guards against the second pass, and a code object whose formation has side effects a re-formation must not repeat declines it: a role body appends its lexical fixup nodes, and a regex body accumulates alternation NFAs. The re-formation runs only between a unit's optimize walk and that unit's emission, never inside a dynamic compilation.

Iterating a ten element array five hundred thousand times drops from 0.95s to 0.54s with a RakuAST-compiled CORE.

### DIFF
--- a/src/Raku/ast/code.rakumod
+++ b/src/Raku/ast/code.rakumod
@@ -102,6 +102,14 @@ class RakuAST::Code
     has Mu $!qast-block;
     has str $!cuid;
 
+    # A BEGIN-time use forces compilation ahead of the unit's optimize
+    # walk, and the QAST block that compilation forms is cached. These
+    # record that the cache was formed early, and with what arguments,
+    # so the block can be re-formed once the marks are known.
+    has int $!begin-time-cached;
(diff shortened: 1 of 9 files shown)


ab5tract:
Loving the benchmark result 💪🏻

PR #6583, opened 2026-08-25
Key precompilation stores to the setting they bind to

ugexe:
Previously the compilation id keying a precompilation store derived from the compiler's sources and the frontend compiling the modules, while the setting a precompilation binds to also varies by the frontend that compiled it and by the version string it bakes in. Rebuilding the setting with the other frontend, or after a commit that moves the described version without touching a source, replaced the setting while every store stayed keyed as before, and loading from one fails with errors naming types that have nothing to do with the module asked for.

The compilation id now folds in the frontend that compiled the setting, baked in at setting compile time, and the version string the setting already bakes in beside it, so the stores move whenever the setting they bind to does.

### DIFF
--- a/src/core.c/Compiler.rakumod
+++ b/src/core.c/Compiler.rakumod
@@ -5,12 +5,18 @@ class Compiler does Systemic {
     my constant $backend = $compiler.backend;
     my constant $name    = $backend.name;
 
+    # The frontends serialize a setting differently from the same sources
+    my constant $core-frontend =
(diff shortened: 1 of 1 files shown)


lizmat:
Hmmm..  I'd hoped that this would fix the issue I found with `Identity::Utils`'s `bytecoe-io` subroutine (which didn't work on RakuAST).  With this change, it doesn't work on legacy either  :-(

PR #6585, opened 2026-08-25
RakuAST: fix issue with m:p/foo/ codegenning

lizmat:
It appears that is was conflating the functionality of
  m:p/foo/
with
  m:p(2)/foo/

Fix this by only taking the special "to" codegenning path if the specified adverb represents True.


ugexe:
RakuAST::Substitution has an identical bug

PR #6588, opened 2026-08-25
RakuAST: attach doc blocks to the statement that follows them

ugexe:
Previously a parsed doc block was claimed by the first statement action to run after it. Statement actions run innermost first, so a doc block directly before a `role` or `class` was claimed by the first statement completed inside the package body rather than the declaration itself. For a role whose first member was a stub method the doc block landed in the stub's body statement list, the method no longer read as a stub, and it lost its yada bit. That silently turned off required method enforcement:

```raku
=begin pod
docs
=end pod
role Sink {
    method dispatch($event) { ... }
}
Sink.new.dispatch("x") # "Stub code executed" instead of "Method 'dispatch' must be implemented..."
```

(found via blin, JobQueue's test suite expects the must be implemented error)

This stores a statement id mark with each collected doc block so a block is claimed by the statement that follows it in the same statement list. Statements parsed outside any statement list, i.e. the argument of a statement prefix such as `do` or a regex `:my` declaration, hand claimed blocks back for the enclosing statement. Backtracked parses refresh the mark, since a config value such as `:kind("Type")` advances the statement counter on every pass. Anything still unclaimed is added to the outermost statement list when the compilation unit is assembled, so a doc block can never be silently dropped from `$=pod`. A second commit makes stub detection itself ignore doc blocks, so a pod block inside a stub body does not defeat the yada bit either.

### DIFF
--- a/src/Raku/Actions.nqp
+++ b/src/Raku/Actions.nqp
@@ -197,6 +197,24 @@ role Raku::CommonActions {
     # this was done with the "r" method.  Since it is apparently impossible
     # to reliably export Nodify, this interface is kept alive.
     method r(str $class) { Nodify($class) }
+
+    # A statement parsed outside of any statement list, as in the argument
+    # of a statement prefix or a regex :my declaration, is a dead end for
+    # doc blocks it claimed, since it is never added to a statement list.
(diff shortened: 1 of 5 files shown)


lizmat:
Wow!

PR #6589, opened 2026-08-26
RakuAST: don't set the caller's $/ in 6.e

### LINKED ISSUE #1235: Caller-$/-setting is too invasive
This started with [RT#126721](https://rt.perl.org/Ticket/Display.html?id=126721#txn-1376978) that points out `.subst`'s replacement block has wrong `$/` if caller's `$/` is read-only. The bug was originally [fixed](https://github.com/rakudo/rakudo/commit/738908be4d) by having `.subst` throw in such circumstances, just like `.match` and `.trans` do (now [reverted](https://github.com/rakudo/rakudo/commit/1c387153ea)).

This fix caused breakage in about 20 modules and when I went to fix it I ended up staring at [diffs like this](https://gist.github.com/zoffixznet/c68579b7cad03b286d735dbe91e529f1), with a dozen changes, all just to rename a variable, all just because a `.subst` call was used. 

This is what the users would regularly have to do in their grammar actions, since `$/` is the common name for the parameter. For that reason, I propose we go into the opposite direction and instead **don't set caller's `$/`** in method forms. Off the top of my head, that'd affect `.parse`, `.subparse`, `.match`, `.trans` (currently throw on read-only `$/`) and `.subst` and .`subst-mutate` (currently silently fail). `.comb` with regex might be also in the list, except it already appears it does not touch `$/`.

The operator forms (`/.../`, `s/...//`, `S/...//`, and `tr///`) on the other hand *would* set `$/`. Unlike operators, methods are often chained, and at least to me feel like they shouldn't be messing with variables in my current scope.

So `"foo" ~~ /(f)(.)(.)/` would set `$/` and the user can obtain captures from it. `"foo".match: /(f)(.)(.)/` would not and it'd be assumed the user would be chaining it with something and using the return value.

One obvious implementation would be to add something like `:$RAKUDO-INTERNAL-SET-DOLLAR-SLASH` parameter to the methods that operators would set, since they're implemented in terms of those methods. The arg could also be made public API (e.g. `:$set-match`), so users could set it in method forms to affect the `$/` . Finally, the back-compat way to do this could be to add a reverse arg: one that makes method forms NOT touch `$/` but IMO that feels like a bandaid over the issue and users would still get confusing errors when using these methods in grammar actions, except instead of telling them to rename the parameter we'd be telling them to set an arg.

This change would break 6.c tests, so it's 6.d material.

ugexe:
Previously the method forms of matching (`.match`, `.subst`, `.trans`, `Grammar.parse` and friends) reached into their caller's frame and wrote the match into its `$/`. That is the behavior #1235 describes as too invasive. It breaks when the caller's `$/` is bound or read-only, which is the normal state of affairs inside grammar actions, and it makes matching in threaded code unsafe since every block shares the enclosing scope's one container:

```raku
say (^1000).race.map({ S/ (.*) /$0/ }).unique.elems;  # ~915 instead of 1000

$/ := "meow";
say "42".subst(/(\d)/, { $0 * 2 }, :g);  # dies trying to numify "meow"
```

As jnthn noted in the issue, simply not setting `$/` would break the very common `$str.subst(/(\d+)/, { $0 * 2 })` idiom, since the replacement block reads its captures through the caller's `$/`. However, bare blocks already receive the match as their topic. As such, under `v6.e.PREVIEW` every block now declares its own `$/`, initialized at block entry from the topic when the topic is a Match and from the enclosing `$/` value otherwise. A regex operator inside a block writes the block's own container, a subst callback reads the captures of the match it was invoked with, `"ab".match(/(\w)/, :g).map({ ~$0 })` gives per item captures, and a block with no match of its own still sees the enclosing one.

With that in place the method forms no longer need to write the caller's `$/` at all. An untyped `$/` compiled by the RakuAST frontend in a 6.e scope carries a marker container descriptor (`ContainerDescriptor::IsolatedMatch`) and the method side write points leave such a container alone. The operator forms (`~~`, `m//`, `s///`, `S///`, `tr///`) still set `$/` in their own scope. The visible migration item is that 6.e code calling `.match` and then reading `$0` should use the return value instead, e.g. `with $str.match(/re/) { ~$0 }`. Callers on 6.d and earlier keep the old behavior in full, as their containers carry no marker, and the legacy frontend is unchanged.

The first commit stands somewhat apart and applies to every language revision. `make` now attaches to the match of the nearest enclosing regex frame through its `$¢` rather than through the caller's `$/`, which user code can clobber. Previously a `~~` inside a token's code block would redirect `make` to the wrong Match and `.made` on the parse result came back `Nil`. Two corner cases change with this in older revisions, however: a closure that escapes a regex frame and calls `make` later now attaches to that frame's cursor instead of throwing, and an explicit `$/ = $other` inside a regex code block is no longer honored.

Fixes #1235

### DIFF
--- a/src/Perl6/bootstrap.c/BOOTSTRAP.nqp
+++ b/src/Perl6/bootstrap.c/BOOTSTRAP.nqp
@@ -1550,6 +1550,13 @@ class ContainerDescriptor does Perl6::Metamodel::Explaining {
 # descriptor is used as a marker
 class ContainerDescriptor::Untyped is ContainerDescriptor { }
 
+#- ContainerDescriptor::IsolatedMatch ------------------------------------------
+# Container descriptor for the implicit $/ of a 6.e scope. The type is used
+# as a marker: method forms such as .match and .subst leave a container with
+# this descriptor alone instead of writing the match into their caller's $/,
+# while operator forms assign through it as with any other container.
+class ContainerDescriptor::IsolatedMatch is ContainerDescriptor::Untyped { }
+
 #- ContainerDescriptor::Whence -------------------------------------------------
 # Role for container descriptors that need to bind the container to some place
 # (hence the "whence") on first assignment.  Most commonly used for elements
@@ -2449,7 +2456,8 @@ BEGIN {
                         }
                         my $what := $desc.WHAT;
                         unless nqp::eqaddr($what, ContainerDescriptor) ||
-                               nqp::eqaddr($what, ContainerDescriptor::Untyped) {
+                               nqp::eqaddr($what, ContainerDescriptor::Untyped) ||
+                               nqp::eqaddr($what, ContainerDescriptor::IsolatedMatch) {
                             $desc.assigned($cont)
                                 unless nqp::eqaddr($what, ContainerDescriptor::UninitializedAttribute);
                             nqp::bindattr($cont, Scalar, '$!descriptor', $desc.next);
@@ -2475,7 +2483,8 @@ BEGIN {
                 my $desc := nqp::getattr($cont, Scalar, '$!descriptor');
                 my $what := $desc.WHAT;
                 unless nqp::eqaddr($what, ContainerDescriptor) ||
-                       nqp::eqaddr($what, ContainerDescriptor::Untyped) {
+                       nqp::eqaddr($what, ContainerDescriptor::Untyped) ||
+                       nqp::eqaddr($what, ContainerDescriptor::IsolatedMatch) {
                     $desc.assigned($cont)
                         unless nqp::eqaddr($what, ContainerDescriptor::UninitializedAttribute);
                     nqp::bindattr($cont, Scalar, '$!descriptor', $desc.next);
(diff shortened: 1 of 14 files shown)


ab5tract:
:rocket: xx Inf

niner:
Epic! I'm glad to see this design flaw finally fixed.

jubilatious1:
@ugexe Don't know if I agree with this, since I have literally hundreds of one-liners posted on StackExchange, and many use the idiom you describe in your opening sentence:

https://unix.stackexchange.com/search?q=user%3A227738+raku

I certainly think there has to be a way to preserve `$/` loading for single threaded use, while enabling Raku to "roll-over" to a different variable for multi-threaded use. 

For example, a synonym for `$/` was (for many years) the visually pleasing `$<>`, which can **still** be resurrected for use in multi-threaded code (for example, by creating a hash with thread-number as key, and `$<>` as variable.

jubilatious1:
@ugexe 

> `with $str.match(/re/) { ~$0 }`

No parens, so `$0` isn't set.

Possibly you mean the following?

> `with $str.match(/(re)/) { ~$0 }`

##OR

> `with $str.match(/re/) { ~$/ }`

Thank you.

PR #6591, opened 2026-08-26
RakuAST: implement FOREIGN-LANG for Raku level slang grammars

ugexe:
Previously FOREIGN-LANG died with an NYI error when the requested slang grammar was a Raku grammar rather than an NQP one, which prevented slang modules such as Slang::BNF from parsing a foreign syntax inside their package declarator. This ports the legacy frontend's non-NQP branch: run the Raku grammar's regex with the slang's actions and copy the result back into an NQP match that the calling grammar can use like any other subrule result. Positional and named arguments are passed through to the foreign regex, an unregistered language name dies with a message naming it, and the extent copy honors capture markers in the foreign regex.


lizmat:
Cool!

PR #6592, opened 2026-08-27
Derive typed hash storage keys the way Hash does

ugexe:
Previously Hash::Typed derived a storage key by calling .Str on it and assigned at Raku level. The AT-KEY, EXISTS-KEY and DELETE-KEY it inherits take a concrete Str key by its string value, so a Str subclass overriding .Str was written to one slot and read from another:

    my $key = "a" but role { method Str { "b" } };
    my Int %h;
    %h{$key} = 1;
    say %h{$key};   # (Int)

ASSIGN-KEY and BIND-KEY now derive the key the way those inherited lookups do, which also means a concrete Str never has its .Str called. ASSIGN-KEY assigns through nqp::p6assign, which is what sent every store into the container store fallback, and keeps the Raku level store for a key holding something other than a container so that its STORE runs. That store also names the value it will not modify, where an untyped hash reports only that the value is immutable.

A store into `my Int %h` runs 0.27s per million against 0.61s before it, with an untyped hash at 0.17s on the same build.

### DIFF
--- a/src/core.c/Hash/Typed.rakumod
+++ b/src/core.c/Hash/Typed.rakumod
@@ -10,26 +10,33 @@ my role Hash::Typed[::TValue, ::TKey, ::TDefault = TValue] does Associative[TVal
 
     method ASSIGN-KEY(::?CLASS:D: Mu \key, Mu \assignval) is raw {
         my \storage  := nqp::getattr(self, Map, '$!storage');
-        my \which    := key.Str;
-        my \existing := nqp::atkey(storage,which);
+        my \dkey     := nqp::decont(key);
+        my str $which = nqp::istype(dkey,Str) && nqp::isconcrete(dkey)
+          ?? nqp::unbox_s(dkey)
+          !! dkey.Str;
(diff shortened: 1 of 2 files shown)


lizmat:
Hmm... I understand the logic and like the optimization result.

But to *not* have it call `.Str` feels wrong to me **precisely** because it would allow some type of code injection into the process.

Also, the whole idea of it calling `.Str` was that it would get optimized away very quickly as a no-op if called on a `Str`: if that didn't happen, I'd say we would need to look at that?

lizmat:
```raku
    multi method AT-KEY(Hash:D: \key) is raw {
        nqp::ifnull(
          nqp::atkey(nqp::getattr(self,Map,'$!storage'),key.Str),
          self!AT_KEY_CONTAINER(key.Str)
        )
    }
```
Looks like that's what `Hash` is doing??

ugexe:
That's the \key candidate, right above it is the Str:D one that actually handles Str keys. Hash::Typed overrides only the writes, so that candidate is what reads my Int %h, which is why a key written through .Str never came back.

PR #6598, opened 2026-08-28
Initialize %*ENV, $*CWD and $*SCHEDULER lazily

ugexe:
Previously the core setting built these at every startup: `%*ENV `by iterating the whole environment and running val() on every value, `$*CWD` as an IO::Path, and `$*SCHEDULER` as a full ThreadPoolScheduler. A process that never touches them still paid for all three, and the `%*ENV` cost grows with the size of the environment.

They now register with the dynamic variable initializer registry the way `$*TMPDIR` and `$*PROGRAM` do, and are built on first access. The deprecations END phaser checks its report before consulting `%*ENV`, so a process that saw no deprecations does not build the environment hash at exit. The scheduler debug flags read the raw environment because they are evaluated while the setting loads. The REPL reaches the scheduler through the dynamic lookup, which initializes it on demand.

Removes about 5ms of work from every startup on my personal environment.


lizmat:
Weird.  I recall making these eager as they would be always needed.

lizmat:
I think this is going to need a lot of more thought.

Yes, it does make the empty startup a lot faster.  But as soon as you `use` a module, the advantage becomes a disadvantage: spectest gains at least 4 seconds for me with this change.

I think we need an other lazy approach to vivifying `%*ENV` as clearly the `val` processing is the factor that is slowing things down.

Going to mark this PR as a draft for now while I do some testing.

lizmat:
Thinking about this some more, I think (as the original inventor of the `REGISTER/INITIALIZE-DYNAMIC` mechanism more than a decade ago) a different approach is needed.

A vague idea that I have developing now is, is that we need a new `ContainerDescriptor`: one that will execute a bit of code on first *access*.  A bit of a `Proxy` light.  The return value of that call to be used in the container.  In the case of `%*ENV` that would be initialized with:

```raku
{
    nqp::if(
      nqp::isnull(my $raw := nqp::atkey(nqp::getenvhash,$key)),
      Failure.new...,
      val($raw)
    )
}
```
 This would also allow getting rid of the `INITIALIZE-DYNAMIC` call in `&DYNAMIC`.   

In the case of other `$*`variables, they would be essentially using the code that they have with `INITIALIZE-DYNAMIC`.

And I feel this could actually be a very valuable generic language feature:
```raku
my $a is lazy = complicated calculation thunk;
```
wouldn't actually run the thunk until the variable is actually accessed.

See also: [Object::Delayed](https://raku.land/zef:lizmat/Object::Delayed)

lizmat:
The deprecations part I liked, so implemented in https://github.com/rakudo/rakudo/commit/df41eb1977

lizmat:
Q: does this mean I should expect a new PR?   or that you're abandoning the idea altogether?

PR #6605, opened 2026-08-29
Bump NQP to get precomputed grammar NFAs serialized in their VM form

ugexe:


ugexe:
hmm, I'll need to revert my nqp change or fix it as I see some potential difficulty messages in the output I didn't see before

PR #6608, opened 2026-08-29
RakuAST: honour a user operator wherever the setting's was assumed

ugexe:
Previously the optimize walk and the meta-op emission took an operator for the setting's in more places than a later or extending user declaration allowed. The core operator check judged by the routine's SETTING:: file, which a user multi's derived proto keeps, so the native lowerings and the range loop ran the setting's routine past a user candidate. Compile time dispatch decided against the routine the node resolved to at parse time, so a user multi candidate declared after the use was not seen. And every meta operator, and a reduction, formed its meta-op at compile time from the setting's routine at any optimize level, so `$x += 2`, `[+] 1, 2, 3`, `(1, 2) Z+ (3, 4)` and `1 !eq 2` ignored a user operator declared later in the unit where the legacy frontend calls it:

```
$ RAKUDO_RAKUAST=1 raku -e 'say [+] 1, 2, 3; sub infix:<+>($a, $b) { "user" }'
6
$ RAKUDO_RAKUAST=1 raku -e 'multi sub infix:<..>(Int $a, Int $b) is default { (100,) }; my $s = 0; for 1..3 { $s = $s + $_ }; say $s'
6
```

The first commit compares the operator's routine with the one the setting binds to the name, and decides the dispatch against the declaration the scope holds, so a user candidate that cannot take the operands leaves the native splice in place wherever it is declared and one that can is what the dispatch lands on. The second forms a meta-op constant only when the optimize walk found the operator's name still resolving to the setting routine, recorded as a mark on the operator node, and looks the operator up at run time otherwise, as the legacy frontend does. Both one-liners above now print what the legacy frontend prints.

The third commit turned up along the way: a string EVAL compiled to the AST and went straight on to QAST, skipping the optimize stage that sits between them, which an EVAL of an AST already ran itself. Nothing EVALled was optimized, a native counting loop inside an EVAL took thirty times longer than the same loop in a script, and with the meta-op constant now depending on the walk it would have lost that too. The stage runs for a string EVAL now, honouring the optimize option as it does elsewhere.

### DIFF
--- a/src/Raku/ast/base.rakumod
+++ b/src/Raku/ast/base.rakumod
@@ -1754,7 +1754,14 @@ class RakuAST::Node {
             return Nil;
         }
 
+        # A routine declared after the use shadows the setting's the node
+        # resolved to and dispatches every use in its scope, so the
+        # dispatch is decided against the declaration the scope holds.
         my $resolution := $target.resolution;
+        if nqp::istype($resolution, RakuAST::Declaration::External::Setting) {
+            my $current := $resolver.resolve-lexical($lexname);
+            $resolution := $current if nqp::isconcrete($current);
+        }
         return Nil unless self.IMPL-RESOLUTION-BOUND-ONCE($resolver, $resolution, $lexname);
         my $routine := self.IMPL-DECLARATION-VALUE($resolution);
         return Nil unless nqp::isconcrete($routine)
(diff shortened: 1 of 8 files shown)


lizmat:
Does this fix `Slang::Predicate` ?

ugexe:
I don't see how it would. The problem for Slang::Predicate is the unicode variants of `=>` and `:=` in rakudo were just added to the grammar which isn't a sufficient way of adding aliases (and indeed none of the other unicode variants that work fine just added a different spelling to the grammar).

lizmat:
Ack

PR #6614, opened 2026-08-30
Sort, slip and split without work the answer does not need

ugexe:
Three unrelated changes in the setting, each removing work that was being done to reach the same answer:

`.sort` compared every pair through the `infix:<cmp>` multi dispatch, which hands back an `Order` the merge then tests by identity. For a list that is all `Int` or all `Str` that whole step is one VM op, so a walk decides whether it is and the merge runs against `nqp::cmp_I` or `nqp::cmp_s` directly. Sorting by an extracted key does the same for the keys. Both pick a merge once rather than testing the kind inside the comparison, so a sort that cannot use one pays nothing for it. Only the exact types qualify, since an `IntStr` breaks a numeric tie on its string where `nqp::cmp_I` calls the pair equal, and the walk only happens from sixteen elements up, which is where it starts paying for itself. A million `Int` goes 3.57s to 2.75s and six hundred thousand strings by an extracted key 2.33s to 1.33s, while sorting `Num`, `Rat`, `Version` or `Instant`, by value or by key, stays where it was.

`SlippyIterator` recognizes `Empty` with `nqp::eqaddr`, but a `$` parameter binds the Slip through a container, so reached the way an iterator reaches it, with a map or grep block's result, `slip-all` compared the container and never matched. Every element such a block rejected built an iterator over `Empty` and pushed from it to push nothing. A grep over two million rejected elements goes 0.63s to 0.49s.

```
use nqp; sub f(Slip:D $s) { nqp::eqaddr($s, Empty) }; say f(Empty)   # False
```

`Str.split` checked its `:v`/`:k`/`:kv`/`:p` adverbs by adding four boolifications together, on a call that almost never passes one. An adverb nobody passed still holds `Any`, which boolifies false, so four of those sum to zero without asking.

### DIFF
--- a/src/core.c/Rakudo/SlippyIterator.rakumod
+++ b/src/core.c/Rakudo/SlippyIterator.rakumod
@@ -4,11 +4,17 @@
 my role Rakudo::SlippyIterator does Iterator is implementation-detail {
     has Mu $!slipper;  # iterator of the Slip we're iterating, null if none
 
+    # Every multi here takes the Slip raw and decontainerizes it in the
+    # body.  Bound to a `$` parameter, a Slip that arrived as a map or
+    # grep block's result reaches slip-all inside a container, and the
+    # nqp::eqaddr that recognizes Empty compares that container instead.
+
     proto method start-slip(|) {*}
-    multi method start-slip(Slip:U $slip) {
-        $slip
+    multi method start-slip(Slip:U \slip) {
+        nqp::decont(slip)
     }
-    multi method start-slip(Slip:D $slip) {
+    multi method start-slip(Slip:D \slip) {
+        my $slip := nqp::decont(slip);
         nqp::if(
           nqp::eqaddr($slip,Empty),
           IterationEnd,                  # we know there's nothing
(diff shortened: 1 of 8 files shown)


lizmat:
I find the [4e974d0](https://github.com/rakudo/rakudo/pull/6614/commits/4e974d0a9dceab7b1357aa4faa0fa14562355765) commit suspect in the sense that it appears to duplicate the functionality of `are`.  Shouldn't a `are(Int)` or `are(Str` be a quick way to see if they're Int or Str?  Or maybe just do an `are` on the list and check the resulting type?

Also, it potentially feels like this optimization would backfire on large lists.  Have you benchmarked that?

lizmat:
I've cherry-picked the other two commits.

ugexe:
> Also, it potentially feels like this optimization would backfire on large lists. Have you benchmarked that?

Yes, it is literally mentioned in the commit message. It gets *better* with bigger lists even.

ugexe:
> Shouldn't a are(Int) or are(Str be a quick way to see if they're Int or Str? Or maybe just do an are on the list and check the resulting type?

`are` answers a different question. `are(Int)` is an `nqp::istype`, so it is True for `IntStr` and for any `Int` subclass, and those are the lists the merge should not take given `nqp::cmp_I` calls a numeric tie equal where `infix:<cmp>` breaks it on the string.

`.are` and a type check is closer, but it hands back the narrowest common type, so a mixed list says `Int` and sorts wrong:

```
my @m = 1, IntStr.new(1,"b"), IntStr.new(1,"a"), 0;
say @m.are.^name;            # Int
say @m.sort.map(*.Str);      # (0 1 a b), which a cmp_I merge cannot produce
```

On cost, `.are` walks the whole list through its iterator: 170ms on a million elements, and the same 170ms on a million `Num`, where the walk in the commit reads element 0, finds neither `Int` nor `Str` and stops. Keying on `are` is what would make this backfire on a large list.

PR #6617, opened 2026-08-31
RakuAST: fix package naming and version parsing for Inline::Perl5

ugexe:
Getting Inline::Perl5 working under the RakuAST frontend turned up two unrelated compiler bugs. Both are RakuAST only, the legacy frontend does the right thing in each case, and each commit carries a test that runs on both frontends.

**Naming a package hands the meta-object a raw VM string.** Previously a package's meta-object was created with the name it was declared with and then renamed with its fully qualified name, always handing `set_name` a raw VM string. The legacy frontend passes the fully qualified name to `new_type` once and never renames. That matters for a HOW written in Raku rather than in NQP, e.g. `Inline::Perl5::ClassHOW`, which receives the name through a Raku signature, which boxes it, and then reads `$!name` back directly, which does not. After the rename it is holding a string with no methods on it, and the first use of the name dies with `No such method 'Stringy' for string 'Foo'`. An NQP HOW wants the raw string and does its own string ops on it, so we cannot just box it either way. This passes the rename the same kind of string the name already holds.

**A version adverb eats a module name.** Previously the optional `+` or `-` that may follow a version's numbers would take the hyphen out of a module name such as `v5-inline`. `vnum` is `\w+` and stops at the hyphen, so `v5-inline` matched as version `5-`, and the `<!before '-'>` guard after it cannot help because the hyphen has already been consumed by then. A compilation unit whose first statement is `use v5-inline` would fail with `No compiler available for Raku v5-`, as that is the slot `lang-setup` parses a `use v6` in. Inside a block LTM picks the longname instead, so this only shows up at the very start of a file. This only takes the hyphen when what follows it cannot continue an identifier. The `+` is left as it was, as is `v1.2-`, which a04661369b added and the legacy frontend does not have.

With both of these plus https://github.com/niner/Inline-Perl5/pull/194 Inline::Perl5's test suite passes in full under RAKUDO_RAKUAST=1, the same as it does on the legacy frontend.

### DIFF
--- a/src/Raku/Grammar.nqp
+++ b/src/Raku/Grammar.nqp
@@ -4643,7 +4643,7 @@ grammar Raku::Grammar is HLL::Grammar does Raku::Common {
         <?before v\d+\w*
             [ '.' \d
               || <!{ $*R.resolve-lexical(~$/) }> ]>
-        'v' $<vstr>=[<vnum>+ % '.' ['+' | '-']? ]
+        'v' $<vstr>=[<vnum>+ % '.' ['+' | '-' <!before \w>]? ]
         <!before '-'|\'> # cheat because of LTM fail
     }
 
--- a/src/Raku/ast/package.rakumod
+++ b/src/Raku/ast/package.rakumod
@@ -218,9 +218,18 @@ class RakuAST::Package
                     self.stubbed-meta-object(:$resolver, :$context)
                 );
                 my $full-name := self.IMPL-FULL-NAME($resolver);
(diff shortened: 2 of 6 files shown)


lizmat:
That is good news!   So does that imply that Inline::Perl5 will work unchanged under RakuAST?  I think you said it does, but I wanted to double-check  :-)

lizmat:
Ah, I now see the Inline::Perl5 PRs...   so scratch my question :-)

lizmat (approved):

PR #6619, opened 2026-09-01
Simplify initialization of $*REPO

lizmat:
Did not find a reason why CALLERS:: should be inspected.  Commit 081e4ba0666a7900b44276f2 could not tell me either.

So Remove that and prepare the code to get rid of the legacy backend handling.

### DIFF
--- a/src/core.c/Process.rakumod
+++ b/src/core.c/Process.rakumod
@@ -66,13 +66,23 @@ Rakudo::Internals.REGISTER-DYNAMIC: '$*TOLERANCE', {
 
 Rakudo::Internals.REGISTER-DYNAMIC: '$*REPO', {
     my $repo := PROCESS::<$REPO> := CompUnit::RepositoryRegistry.setup-repositories;
-    my $world := $*W;
-    my $cu    := CALLERS::.EXISTS-KEY('$*CU') ?? $*CU !! Nil;
-    $world.suspend_recording_precompilation_dependencies if $world;
-    $cu.suspend-recording-precompilation-dependencies    if $cu;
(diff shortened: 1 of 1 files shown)


ugexe:
Checking `CALLERS::` is a preexisting idiom. For example see https://github.com/rakudo/rakudo/blob/4fb9c0dcb86a5907fd4e66ff8466c08bd39d5424/src/core.c/CompUnit/PrecompilationRepository.rakumod#L103

The git blame describes the reasoning. See https://github.com/rakudo/rakudo/commit/7194b1e6b39ecc2141051bf87094cd0453f618b6

src/core.c/Process.rakumod
```diff
@@ -66,13 +66,23 @@ Rakudo::Internals.REGISTER-DYNAMIC: '$*TOLERANCE', {
...
+        $cu.suspend-recording-precompilation-dependencies;
+        CompUnit::RepositoryRegistry.resolve-unknown-repos($repo);
+        $cu.resume-recording-precompilation-dependencies;
+    }
+    # Can be removed as soon as 6.e is default
+    elsif $*W -> $world {
+        $world.suspend_recording_precompilation_dependencies;
+        CompUnit::RepositoryRegistry.resolve-unknown-repos($repo);
+        $world.resume_recording_precompilation_dependencies;
+    }
```
ugexe, line 79: To be clear this new change isn't really correct until legacy support is removed entirely. This conditional assumes the frontends of a process are mutually exclusive, but that isn't the case.

lizmat: Aaahh... ok, understood.

PR #6623, opened 2026-09-01
Bump NQP to give a block reference the frame of its own compilation

ugexe:


ugexe:
Gonna make another nqp change shortly so closing this for now

PR #6627, opened 2026-09-02
Streamline setting core dynamic vars, part 1

lizmat:
The ideas behind this work are:
- remove setting up code for lazy setting of core dynamic vars using REGISTER-DYNAMIC, but instead do all of this at once by initializing the lookup hash once, and thus removing some overhead of code that is run at every startup
- move this logic as late as possible in the core setting, so that all Raku features can be used
- improving maintainability of this logic that was previously shattered all around the core

"Process" now augments Rakudo::Internals and there contains and runs all the code for setting up lazy initialization of core dynamic variables.  This also meant moving quite significant chunks of code from other files in the core.  This made the "ArgFiles" and "Env" files basically empty, so they have been removed.

Did not see any noticeable improvement in startup time just yet, but maybe this is already more noticeable on slower machines.


lizmat:
Merging this even though CI claims there are errors, the Azure pipeline doesn't actually show any

PR #6638, opened 2026-09-04
Let make attach to a capture's own match in its code block

ugexe:
Previously make attached to the match of the cursor in the nearest enclosing regex frame. In a capture group's code block the legacy frontend resolves that cursor to the enclosing regex, while `$/` is the capture's own match, so a make there landed on the enclosing match and the capture's made stayed empty. This attaches to the caller's `$/` when it is a match of the parse the cursor belongs to, and falls back to the cursor's match when `$/` has been overwritten by a match of another parse.

### DIFF
--- a/src/core.c/Match.rakumod
+++ b/src/core.c/Match.rakumod
@@ -443,12 +443,16 @@ multi sub infix:<eqv>(Match:D $a, Match:D $b) {
 }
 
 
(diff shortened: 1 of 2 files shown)


coke:
_made small edit to your comment above to escape the `$/`_

PR #6639, opened 2026-09-04
Various nqp bumps

ugexe:


coke:
Previously, each NQP bump was entirely separate, not part of merge. not sure how this will impact bisectability.

ugexe:
git bisects operates on commits so it should be fine

PR #6642, opened 2026-09-05
RakuAST: Fold character class entries after assembling them

### LINKED ISSUE #2962: A Synthetic in a character range, interferes with :ignoremark
$ perl6 -e 'say "a\x[300]" ~~ / :ignoremark <[ a b ]> /'
    ｢à｣

    $ perl6 -e 'say "a\x[300]" ~~ / :ignoremark <[ a \r\n ]> /'
    Nil

Expected to see `｢à｣` in both cases.

ugexe:
Previously a character class folded each plain character for :ignorecase and :ignoremark before appending it to the entries, and escaped characters were appended without any folding at all. As such <[ \x[E0] ]> under :ignoremark did not match "a", and an entry that adjacent characters form under NFG, such as the "à" in <[ a \x[300] ]>, was never reduced to its base character. The case folding also replaced a character by its case forms even when a form was not a single character, so <[ ß ]> under :ignorecase matched "s" but not "ß".

This assembles the entries first and then folds each grapheme of the result, keeping the character itself and only adding case forms that are a single character. A folded entry that would form one grapheme with the entry before it, such as a carriage return followed by a newline, becomes a list of its own.

Fixes #2962

### DIFF
--- a/src/Raku/ast/regex.rakumod
+++ b/src/Raku/ast/regex.rakumod
@@ -2097,10 +2097,8 @@ class RakuAST::Regex::CharClassElement::Enumeration
         # If we collected characters, add the enumeration to the alternation
         # parts we'll compile into.
         if $enum {
-            @alts.push: QAST::Regex.new:
-                :rxtype<enumcharlist>, :negate(self.negated),
-                :subtype(%mods<m> ?? 'ignoremark' !! ''),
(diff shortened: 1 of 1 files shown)


ugexe:
Requires https://github.com/Raku/nqp/pull/870

PR #6647, opened 2026-09-06
RakuAST: split long decimal numbers before converting them to Int

ugexe:
Previously the literal builder handed a whole decimal digit string to the VM bigint parser, which is quadratic in the digit count. A capture variable, integer literal, or regex quantifier bound with a million digits took around a minute to compile, while the legacy frontend parsed the same capture index into a native int in under a second.

This halves the digit string recursively down to runs that fit a native int, parses those with the native radix op, and recombines the halves with bigint multiplication. An underscore landing on a split is dropped so both halves keep the shape the native op validates. The same million digits now convert in a fraction of a second, and signs, underscores between digits, and malformed input are handled as before.

### DIFF
--- a/src/Raku/ast/compunit.rakumod
+++ b/src/Raku/ast/compunit.rakumod
@@ -961,38 +961,106 @@ class RakuAST::LiteralBuilder {
     # Build a decimal Int constant and intern it
     method intern-Int(str $source) {
         my $lookup := $!interned-int;
-
-        # Logic to reliably convert a string in a base to n Int
-        my sub build-Int() {
-            my $res := nqp::radix_I(10,$source,0,2,Int);
-            nqp::atpos($res,2) == nqp::chars($source)
-              ?? nqp::atpos($res, 0)
-              !! nqp::die("'$source' is not a valid number")
-        }
-
         nqp::ifnull(
           nqp::atkey($lookup,$source),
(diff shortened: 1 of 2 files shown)


ugexe:
For `raku -e 'use MONKEY-SEE-NO-EVAL; my $t = now; EVAL("1" x 200000); say now - $t'`:

2026.08 legacy: 14.8s
2026.08 rakuast: 2.0s

this branch legacy: 14.8s (not covered in this PR)
this branch rakuast: 0.14s

I noticed that `S05-capture/dot.t` was taking 50s for rakuast and 1s for legacy, and after this PR rakuast finishes it in 1.2s. The legacy front end appears equally as fast here which might be a bit confusing given the earlier benchmark numbers where legacy has no improvement, but it turned out that legacy was returning a wrapped native int so it wasn't actually Doing The Right Thing:

```
$ raku -e 'my sub postcircumfix:<[ ]>($m, $i) { say $i }; "a" ~~ /a/; use MONKEY-SEE-NO-EVAL; EVAL q[$99999999999999999999]'

7766279631452241919


$ RAKUDO_RAKUAST=1 raku -e 'my sub postcircumfix:<[ ]>($m, $i) { say $i }; "a" ~~ /a/; use MONKEY-SEE-NO-EVAL; EVAL q[$99999999999999999999]'

99999999999999999999
```

PR #6651, opened 2026-09-06
Match the initializer symbol before rejecting a signature declaration

ugexe:
Previously initializer:sym<=> checked for a signature declaration before matching its symbol, so that group led the candidate and set its declarative prefix. Once a sequential alternation ends the prefix, that candidate has an empty prefix and gets tried on any input. A my :(...) declaration with no initializer at all then died with the assignment error instead of the missing initializer one.

This matches the symbol first, so the candidate is only tried when an assignment is actually there.


ugexe:
Fixing this legacy issue so that some nqp changes can pass their tests

Issue #1924, opened 2018-06-15
LTM semantics of frugal quantifiers


jnthn:
## The Problem

The semantics of frugal quantifiers in LTM are currently unspecified (or at least I couldn't find that they are anywhere in roast). S05 speculates that they should be LTM terminators and not themselves feature in the longest token, which makes sense, and we should probably implement that.

I discovered this while investigating [this SO post](https://stackoverflow.com/questions/50879757/is-a-nongreedy-pattern-part-of-perl-6s-longest-token-matching).

## Expected Behavior

````
say 'aaaaaaaaa' ~~ m/
    | a+?
    | a+
    /; # ｢aaaaaaaaa｣
```

## Actual Behavior

````
say 'aaaaaaaaa' ~~ m/
    | a+?
    | a+
    /; # ｢a｣
```

W4anD0eR96:
Related #1222

PR #6654, opened 2026-09-07
Match the lookahead greedily in the LTM before-assertion test

ugexe:
Previously the test grammar looked ahead with `.+?`, which only contributed to the declarative prefix because the NFA treated a frugal quantifier like a greedy one. S05 ends the longest token at a frugal quantifier, so the test would then assert nothing about the lookahead being inlined, which is what it was written for.

This looks ahead with `.+` instead. Inside a lookahead nothing is consumed, so the assertion succeeds on the same input either way.


ugexe:
The purpose of this change is so if this test gets moved to the roast that we don't encode behavior that would prevent #1924

PR #6656, opened 2026-09-07
Look ahead greedily for a bare rational or complex literal

ugexe:
Previously `bare_rat_number` and `bare_complex_number`, and their RakuAST counterparts, looked ahead with `+?` to tell `<1/2>` and `<1+2i>` apart from quote words. Inside a lookahead nothing is consumed, and neither `/` nor `i` is in the character class being scanned, so a frugal scan and a greedy scan stop at the same place and the assertion succeeds on the same input either way.

This looks ahead with `+` instead. S05 ends the longest token at a frugal quantifier, and a greedy scan declares the same prefix whether or not the NFA follows that rule, so the literal stays ranked ahead of quote words either way instead of tying with them on a single `<`.

### DIFF
--- a/src/Perl6/Grammar.nqp
+++ b/src/Perl6/Grammar.nqp
@@ -3426,10 +3426,10 @@ sub, perhaps you accidentally placed a semicolon after routine's definition?"
     token sign { '+' | '-' | '−' | '' }
 
     token rat_number { '<' <bare_rat_number> '>' }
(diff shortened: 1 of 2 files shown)


ugexe:
I created this in preparation for https://github.com/Raku/nqp/pull/872 which (assuming we actually merge it) would require this PR to be merged. There shouldn't be any harm in merging this PR regardless.

PR #6658, opened 2026-09-08
Apply :ratchet to an atom followed by sigspace whitespace

### LINKED ISSUE #3890: A quantified non-word character incorrectly allows backtracking with :sigspace:ratchet (or rules)
## The Problem

Consider the following Regexes that match against a series of non-word characters:

```raku
my $s = '!@$'

$s ~~ rx/.*<.ws>'$'/    # 1) RESULT: ｢!@$｣  -- correct
$s ~~ rx:r/.*<.ws>'$'/  # 2) RESULT Nil     -- correct
$s ~~ rx:r:s/.* '$'/    # 3) RESULT ｢!@$｣   -- wrong!
$s ~~ rx:r:s/.*: '$'    # 4) RESULT Nil     -- correct
```
in case 3, the regex incorrectly allows for backtracking, despite the presence of the `:r` adverb.  This can be seen even more directly with `'!@$' ~~ rx:r:s/.* {say $¢}'$';` which shows the backtracking occurring: 

```raku
｢!@$｣
｢!@｣
```

ugexe:
Previously the `<.ws>` that `:sigspace` inserts after an atom was wrapped together with the atom, and the ratchet was applied to that wrapper. The regex compiler ignores a ratchet on a concat, so in a `rule` (or any `:r:s` regex) a quantifier, alternation, subrule call or capture group followed by whitespace was free to backtrack as if `:ratchet` were not there. The same regex with the whitespace spelled out as `<.ws>` did ratchet, as did the same regex without the space:

```
$ raku -e 'say "!@\$" ~~ rx:r:s/.* "\$"/'      # ｢!@$｣, but should not match
$ raku -e 'say "!@\$" ~~ rx:r:s/.*<.ws>"\$"/'  # Nil
```

A `%%` trailing separator wraps its quantifier in a concat the same way, so `token { \w+ %% ' ' FALSE }` gave an iteration back even without sigspace.

This applies the ratchet to the atom's own node before the whitespace is attached, and to the quantifier inside the `%%` concat, so an atom ratchets the same with or without whitespace after it. An explicit `!`, `?` or `:` on the quantifier still wins, and the `%%` trailing separator follows whatever the quantifier does. Both frontends get the change. The legacy frontend overrides `quantified_atom` in `Perl6::RegexActions` rather than changing the nqp method, since that method also compiles the grammars of nqp and rakudo themselves.

This is a behavior change for grammars that relied on a rule backtracking across whitespace, e.g. `'DEFAULT' <id-string>? <value>` where `<id-string>` also matches what `<value>` needs (ASN::Grammar does this). The portable spelling is to ask for the backtracking explicitly with `<id-string>?! <value>`, which means the same thing on older rakudo. We should run Blin on this before merging.

Fixes #3890


ugexe:
This needs a Blin run to see the magnitude of the fallout since it may be too great at this point to proceed with this change

edit: it breaks ASN::Grammar, CSS::Grammar, CSS::Module, Grammar::Modelica, Math::Symbolic, SQL::Lexer

PR #6670, opened 2026-09-12
Better error when crawling files for "use lib" or RAKULIB fails, and warn about too big dir trees

timo:
Instead of giving just the underlying error message when crawling files fails for whatever reason, we output a hint that it's related to setting up a lib path.

For now, I just chose 10k files crawled as a reasonable "did you really mean to do this?" threshold.

Using "warn" is not my favourite way to do this, but I'm not sure if I can add a WORRY from this piece of code, since it may run at run-time as well.

```
timo@pebbsi:/var/home/timo/.var$ rakudo -I . -e 'say "hi"; use JSON::Fast; say "bye";'
Setting up lib path /var/home/timo/.var: after crawling 790729 files:
    Failed to stat file: no such file or directory
at -e:1
timo@pebbsi:/var/home/timo/.var$ cd app/org.mozilla.firefox/
timo@pebbsi:/var/home/timo/.var/app/org.mozilla.firefox$ rakudo -I . -e 'say "hi"; use JSON::Fast; say "bye";'
The path /var/home/timo/.var/app/org.mozilla.firefox has 59721 files in it (but no relevant files).
  Did you put it into rakudo's module search path by accident?
  in any statement_control at /var/home/timo/raku/prefix/share/perl6/lib/Perl6/Grammar.moarvm line 1
hi
bye
```

### DIFF
--- a/src/core.c/CompUnit/Repository/FileSystem.rakumod
+++ b/src/core.c/CompUnit/Repository/FileSystem.rakumod
@@ -333,16 +333,30 @@ class CompUnit::Repository::FileSystem
 
         # Set up hash of hashes of files found that could be modules.
         # Then select the most prominent one from there when done.
-        my %provides-exts = @!extensions.map(* => True);
-        my $provides-files := ls($!prefix.absolute).grep({ my $ext = $_.extension; %provides-exts{$ext} });
         my %provides;
-        %provides{
-          .subst(:g, /\//, "::")
-          .subst(:g, /\:\:+/, '::')
-          .subst(/^.*?'::'/, '')
-          .subst(/\..*/, '')
-        }{ $SPEC.extension($_) } = $_
-          for $provides-files.map(&to-relative);
+
+        try {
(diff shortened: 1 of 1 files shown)


timo:
I would also like to discuss if we should make either this code, or rakudo's internal DIR-RECURSE / DirRecurse ignore the kind of error I've seen, or handle it some other way.

The error I've encountered here is really a bit esoteric, as it relies on a filesystem that will give you a file entry when you read the dir, but then if you try to stat that exact path you got, it will tell you "no such file or directory"; There's a symlink somewhere in the folder I'm pointing `-I` at that links to `/run` or `/run/doc/$UID/` which is a virtual filesystem that the xdg-desktop-portal uses to expose stuff to a sandboxed application when a program shows you the portal's "open file" dialog to access a file from outside the sandbox.

timo:
oh, and I would also like to make either DirRecurse or at least CURFS not follow symlinks to outside of the starting path.

PR #6691, opened 2026-09-16
Add slanted equal to operator for comparison

habere-et-dispertire:
This allows for another equivalence:

* `>=`
* `≥`
* `⩾`


lizmat:
This would also at least need addition to the grammar.

